PR #5, opened 2015-10-03
Optimize clipper: SIMD, reuse allocations, use slices

Manishearth:
More to come, possibly

r? @glennw


Manishearth:
Not really sure if that second commit helps anything

Manishearth:
Added some benchmarks based on real world data. `rectangles.dat` contains rectangle clipping data from various sites (500-1000 each), `polygons.in` contains all the polygon clipping data for the same.

Before:

```
running 2 tests
test bench_clip_poly ... bench:     116,924 ns/iter (+/- 2,995)
test bench_clip_rect ... bench:      78,551 ns/iter (+/- 1,992)
```

After:

```
test bench_clip_poly ... bench:     109,496 ns/iter (+/- 7,831)
test bench_clip_rect ... bench:      62,761 ns/iter (+/- 18,283)
```

The poly test has large variance, but multiple runs confirm an improvement.

Manishearth:
With the shared clip buffers:

```
test bench_clip_poly ... bench:      99,667 ns/iter (+/- 6,136)
```

We still do some cloning. I think we can reduce some of that with some magic.

Manishearth:
With the shared result buffer:

```
test bench_clip_poly ... bench:      94,444 ns/iter (+/- 5,819)
```

Manishearth:
We should probably do some manual profiling of each step before merging. I'm not sure if all of these really end up optimizing it in actual usage. The numbers from testing on Servo looked promising, but they varied a lot.

glennw:
Thanks!

Manishearth:
Final numbers post rebase:

```
test bench_clip_poly ... bench:      88,945 ns/iter (+/- 11,819)
test bench_clip_rect ... bench:      57,166 ns/iter (+/- 3,722)
```


... I don't know how that happened. Perhaps SIMD got faster

PR #11, opened 2015-10-06
Implement partial support for `box-shadow` and add support for clipping out (rectangle subtraction) to the clipper.

pcwalton:
Known issues:

* Clipping out does not clip the uv coordinates properly.

* Clipping to rounded rectangle corners does not work yet.

* Box shadows with elliptical radii are unsupported (also in master
  Servo).

* Box shadows with heterogeneous radii for each corner are unsupported
  (also in master Servo).

This PR depends on some changes to Servo to support the new border
radius field in box shadow display items: servo/servo#7896

r? @glennw

### DIFF
--- a/src/clipper.rs
+++ b/src/clipper.rs
@@ -1,6 +1,6 @@
 use euclid::{Point2D, Rect};
 use internal_types::{ClipRectResult, WorkVertex};
-use types::ColorF;
+use types::{BoxShadowClipMode, ColorF};
 
 fn is_inside(a: &Point2D<f32>, b: &Point2D<f32>, c: &WorkVertex) -> bool {
     (a.x - c.x) * (b.y - c.y) > (a.y - c.y) * (b.x - c.x)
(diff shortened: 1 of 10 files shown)


src/util.rs
```diff
@@ -20,17 +20,18 @@ impl ProfileScope {
 
 impl Drop for ProfileScope {
     fn drop(&mut self) {
-        let t1 = precise_time_ns();
-        let ms = (t1 - self.t0) as f64 / 1000000f64;
-        //if ms > 0.1 {
+        if self.name.chars().next() != Some(' ') {
```
glennw: Was this just to reduce log spam?

glennw:
@pcwalton Nice, looks fairly straight forward. Just the one question / change above, then I'll merge it.

src/clipper.rs
```diff
@@ -86,3 +86,61 @@ pub fn clip_rect_pos_uv(pos: &Rect<f32>, uv: &Rect<f32>, clip_rect: &Rect<f32>)
...
+
+pub fn clip_out_rect_pos_uv(pos: &Rect<f32>, uv: &Rect<f32>, clip_rect: &Rect<f32>)
+                            -> Vec<ClipRectResult> {
+    let clip_rect = match pos.intersection(clip_rect) {
+        Some(clip_rect) => clip_rect,
+        None => return vec![ClipRectResult::from_rects(pos, uv)],
+    };
+
+    // FIXME(pcwalton): Clip the u and v too.
+    let mut result = vec![];
```
Manishearth: with_capacity

Manishearth: (My clipping PR has some reusable buffers that could be used here to be even more efficient)

PR #25, opened 2015-10-14
Implement the CSS `filter` property.

pcwalton:
The mechanism is largely the same as for `mix-blend-mode`.

Known issues:

* `drop-shadow` is unimplemented.

* `blur` does not properly blur the edges.

Both issues stem from the fact that we do not yet inflate bounding boxes for
filters and `mix-blend-mode`.

### DIFF
--- a/src/render_backend.rs
+++ b/src/render_backend.rs
@@ -31,7 +31,8 @@ use types::{BoxShadowCornerRasterOp, RectangleDisplayItem};
 use types::{Glyph, GradientStop, DisplayListMode, RasterItem, ClipRegion};
 use types::{GlyphInstance, ImageID, DrawList, ImageFormat, BoxShadowClipMode, DisplayItem};
 use types::{PipelineId, RenderNotifier, StackingContext, SpecificDisplayItem, ColorF, DrawListID};
-use types::{RenderTargetID, MixBlendMode, CompositeDisplayItem, BorderSide, BorderStyle, NodeIndex};
+use types::{RenderTargetID, MixBlendMode, CompositeDisplayItem, BorderSide, BorderStyle};
+use types::{NodeIndex, CompositionOp, FilterOp, LowLevelFilterOp, BlurDirection};
 use util;
 use util::MatrixHelpers;
 use scoped_threadpool;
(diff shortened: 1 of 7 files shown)


src/render_backend.rs
```diff
@@ -328,6 +329,12 @@ impl Scene {
         debug_assert!(self.render_target_stack.len() == 0);
         self.pipeline_epoch_map.clear();
 
+        println!("reset");
```
glennw: remove

src/render_backend.rs
```diff
@@ -579,7 +634,7 @@ impl Scene {
         // push all visible draw lists into aabb tree
         for (draw_list_index, flat_draw_list) in self.flat_draw_lists.iter_mut().enumerate() {
             for (item_index, item) in flat_draw_list.draw_list.items.iter_mut().enumerate() {
-                assert!(item.node_index.is_none());
+                //assert!(item.node_index.is_none());
```
glennw: Let's uncomment this - I'll fix it when I see it occur next

src/renderer.rs
```diff
@@ -637,25 +655,34 @@ impl Renderer {
...
+                                CompositionOp::MixBlend(MixBlendMode::Saturation) |
+                                CompositionOp::MixBlend(MixBlendMode::Color) |
+                                CompositionOp::MixBlend(MixBlendMode::Luminosity) |
+                                CompositionOp::Filter(LowLevelFilterOp::Blur(..)) |
+                                CompositionOp::Filter(LowLevelFilterOp::Brightness(_)) |
+                                CompositionOp::Filter(LowLevelFilterOp::Contrast(_)) |
+                                CompositionOp::Filter(LowLevelFilterOp::Grayscale(_)) |
+                                CompositionOp::Filter(LowLevelFilterOp::HueRotate(_)) |
+                                CompositionOp::Filter(LowLevelFilterOp::Invert(_)) |
+                                CompositionOp::Filter(LowLevelFilterOp::Opacity(_)) |
```
glennw: Can't opacity (and perhaps some of the others) be done without grabbing the framebuffer, just as a normal opengl blend op?

(We can still merge it if that's the case - but let's add a TODO noting that there is an optimization available for some of these modes to use the fixed function blend ops).

pcwalton:
Updated per comments. r? @glennw

PR #37, opened 2015-10-17
Use RGBA textures for glyphs.

pcwalton:
Improves batching and allows for colored glyphs like emoji.

Depends on https://github.com/glennw/webrender/pull/21 and https://github.com/servo/core-graphics-rs/pull/44


glennw:
This looks good to me - let me know when it's ready to merge  :)

PR #46, opened 2015-10-21
Rewrite much of the clipper to support rounded border corners in fuller generality and fix some bugs.

pcwalton:
Allows gradients to be clipped to rounded border corners, improving the look of
rounded elements' backgrounds.

The clipper no longer takes mask texture coordinates as input, although it
still supplies them as output. This is because it may need to manufacture mask
texture coordinates for rounded corner alpha masks. We were never relying on
this feature for anything, so it does not cause a problem.


glennw:
Nice!

Issue #30, opened 2015-10-15
Consider some form of DLBI for batch retention between paints


pcwalton:
If we could cache the batches on a per-bunch-of-display-item basis between paints somehow, then we could potentially get large wins when only small parts of the page change (say, hovering over links and buttons).

This would be different from traditional DLBI because the idea would be to cache VBOs (or parts of VBOs) rather than being spatial, though of course it could be combined with a form of spatial invalidation later (though given that we aren't ROP-bound, the latter seems unlikely to be that helpful).

We could reuse parts of my DLBI patch to Servo to compare and diff display items if we want to proceed down this path.

pcwalton:
One really nice thing about this approach compared to DLBI is that, because it's not spatial, we don't have to worry about properly handling stacking order differences like "element is sent to back", "element is brought to front", etc., as long as the VBOs end up stitched together properly.

pcwalton:
Actually, that is wrong—we do have to handle that, but at a different layer (possibly during layout, using Chromium's algorithm).

glennw:
I haven't really thought this through - but since webrender works at the granularity of display lists (where there is one generated roughly per fragment), if we could store the display list ID in the fragment, and only update this when layout changes - we would automatically get caching of display lists across frames. Would something like this even be feasible? And would it have any benefits over diffing the display lists?

pcwalton:
Well, the issue is that display items that originate on a fragment could get reshuffled into an arbitrary order, intermixed with those of other fragments.

pcwalton:
@glennw Is it possible to incrementally update the AABB tree when a new display list comes in by only adding items to the tree that are new and only throwing out items that have been removed, and achieve batch caching between paints that way?

glennw:
@pcwalton Can this be closed now?

pcwalton:
Yeah, let's close it

PR #54, opened 2015-10-24
Reuse batches if the display list didn't actually change.

pcwalton:
In general this results in something like a 2x-3x speedup when the test
passes, and the test usually early outs pretty quick if it's going to
fail.

Part of #30.


glennw:
This looks good to me, will merge once rebased.

pcwalton:
@glennw Rebased.

PR #60, opened 2015-10-27
Use OpenGL texture arrays where available to improve batching.

### LINKED ISSUE #38: Use OpenGL Texture Arrays to improve batching
A very common reason for batch breaking is switching textures. This could be mitigated by improved texture atlas allocation algorithms, but they won't fully solve the problem. But the OpenGL 3.0 texture array feature will.

To do this we'll need some way to conditionally opt-in to its use in the shader when GL 3.0 support is present.

pcwalton:
This takes an upgrade to OpenGL 3.x on the Mac. Because the Mac OpenGL
implementation does not support the Compatibility Profile as far as I'm
aware, the shaders have been rewritten to use some preprocessor magic to
work on both OpenGL ES 2.1 and OpenGL 3.x.

Memory usage will likely be somewhat suboptimal until we are dynamically
resizing texture arrays.

Closes #38.


pcwalton:
See https://github.com/glennw/servo/pull/8

res/blend.fs.glsl
```diff
@@ -8,8 +6,10 @@ uniform sampler2D sDiffuse;
 uniform sampler2D sMask;
 uniform vec4 uBlendParams;
 
-varying vec2 vColorTexCoord;
-varying vec2 vMaskTexCoord;
+IN_VARYING vec2 vColorTexCoord;
```
glennw: I think what we should do is introduce 2 new resource files, something like es2_common.glsl and gl3_common.glsl. These could define a function for sampling a texture, and also define all the attributes, varyings etc. Then the runtime code can prefix the shader compiles with the appropriate file. This means we won't need any preprocessor tricks, which I've found flaky in several real world GL drivers. It will also simplify the shaders since we'll have one place to declare all varyings etc (it doesn't matter if you declare unused varyings, attributes etc - they'll get stripped out).

res/quad.vs.glsl
```diff
@@ -1,27 +1,35 @@
...
+        vColorTexCoord = vec3(aColorTexCoord.xy / 65535.0, aColorTexIndex);
+        vMaskTexCoord = vec3(aMaskTexCoord.xy / 65535.0, aMaskTexIndex);
+    #else
+        vColorTexCoord = aColorTexCoord / 65535.0;
+        vMaskTexCoord = aMaskTexCoord / 65535.0;
+    #endif
 
     mat4 matrix = uMatrixPalette[int(aMatrixIndex.x)];
     vec4 pos = matrix * vec4(aPosition, 1.0);
-    pos.xy = floor(pos.xy * uDevicePixelRatio + 0.5) / uDevicePixelRatio;
```
glennw: You removed the hidpi scale here.

res/debug.fs.glsl
```diff
@@ -2,9 +2,11 @@
     precision mediump float;
```
glennw: I don't think this shader is used at all - we should just remove it I think.

src/device.rs
```diff
@@ -179,12 +183,26 @@ pub struct Device {
 
 #[cfg(not(any(target_os = "android", target_os = "gonk")))]
 fn shader_preamble() -> &'static str {
-    ""
```
glennw: This whole string should disappear with common shader include files.

src/device.rs
```diff
@@ -742,11 +1055,23 @@ impl Device {
                                   vertex_stride,
                                   16);
         gl::vertex_attrib_pointer(VertexAttribute::MatrixIndex as gl::GLuint,
-                                  4,
```
glennw: Some GPUs pay a per-attribute cycle fetch penalty. Let's make this one attribute of four bytes, with x=matrix index, y=color tex index etc.

pcwalton:
Rebased and updated per review.

PR #75, opened 2015-11-06
Reuse batches across paints on a per-node basis.

pcwalton:
This didn't really improve performance from what I saw, but it's a
cleaner and more general optimization than the one we had previously,
which only reused batches if the entire display list didn't change. In
any case, it's a precursor to more aggressive optimizations we could do
to reuse portions of batches within AABB tree nodes.


glennw:
Does this guarantee that the display item keys remain the same (the order of these is used to ensure correct paint order)?

glennw:
This seems too out of date now, closing for now.

PR #80, opened 2015-11-13
Implement tessellation for large border radii.

pcwalton:
Improves performance on the Moiré demo.

### DIFF
--- a/src/batch_builder.rs
+++ b/src/batch_builder.rs
@@ -14,6 +14,7 @@ use resource_cache::ResourceCache;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::hash_state::DefaultState;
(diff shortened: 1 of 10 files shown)


pcwalton:
I ended up not using the fancy algorithms I was coming up with, because (a) this trivial one works just as well, and (b) the fact that it tessellates into rectangles and the fact that boundaries of the rectangles can be easily computed individually end up being really useful properties.

glennw:
This looks good to me - could you just confirm that it still gets the triangle border corners correct for the acid2 ref test?

pcwalton:
Ah, this indeed broke the borders (in multiple places in fact!) I'll fix it up tomorrow.

pcwalton:
@glennw New PR fixes Acid2.

glennw:
Thanks!

PR #81, opened 2015-11-16
Texture cache tidy

glennw:


glennw:
This is a tidy up prior to modifying the texture cache to support bilinear filtering correctly.

PR #89, opened 2015-11-20
Remove use of texture arrays.

pcwalton:
The only real thing we need texture arrays for is to get around texture
size limitations, but we believe these are generally so high nowadays as
to not matter--at least on OpenGL 3.0+ hardware, which is the minimum
version required for texture arrays to begin with.


glennw:
Nice, thanks!

PR #90, opened 2015-11-21
Batch similar composite operations together.

pcwalton:
Known bugs and limitations:

* Only 32 composite operations can be batched, because that is the
  maximum size of the matrix palette.

* Nested composite operations aren't correctly handled at the moment.
  They use the wrong device pixel ratio and read from and write to the
  same texture, which is undefined behavior (although usually works).

* When a large number (>700 in my tests) of composite operations occur,
  there can be stray triangles drawn. I haven't been able to pin down
  why that occurs yet.

Please review this carefully if you get time; this is tricky stuff and I'm sure I've messed up :)

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -1,5 +1,10 @@
 use device::{ProgramId, TextureId};
-use internal_types::{AxisDirection, PackedVertex, PackedVertexForTextureCacheUpdate, Primitive};
+use fnv::FnvHasher;
+use internal_types::{AxisDirection, CompositeBatchInfo, CompositeInfo, CompositionOp};
+use internal_types::{PackedVertex, PackedVertexForTextureCacheUpdate, Primitive};
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use std::collections::hash_state::DefaultState;
 use std::sync::atomic::Ordering::SeqCst;
 use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT};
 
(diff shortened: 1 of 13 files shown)


glennw:
Unfortunately this breaks a lot of the ref tests (haven't checked against wpt/css yet). You can run ./mach test-ref --kind=wr --release to see the test failures.

glennw:
I just realised that I forgot to copy across the new shaders - but even after doing that I still see a lot of the ref tests fail, and I think I have updated them correctly this time. When those are fixed I'll get it merged as soon as possible.

pcwalton:
Rebased. Tests not fixed yet.

glennw:
@pcwalton I just found out that the mix-blend-mode shaders were broken (happened here https://github.com/glennw/webrender/commit/5e4493dc3545e8e45c5fe4df7adaa5bbdba4d24f).

I've pushed an update of the shaders to servo + WR.

It's possible that this was the main issue with the reftests + composite batching - you might want to try the reftests just rebased on top of master and see if they just work now :)

pcwalton:
Updated to fix the issues I could see.

src/batch.rs
```diff
@@ -133,6 +139,18 @@ impl<'a> BatchBuilder<'a> {
                          primitive: Primitive,
                          vertices: &mut [PackedVertex],
                          tile_params: Option<TileParams>) {
+        // FIXME(pcwalton): Awful hack to work around a panic.
+        if self.vertex_buffer.vertices.len() + vertices.len() > u16::MAX as usize {
```
glennw: Have you got any info on how to repro this? I would be keen to debug this and find the root cause if we have a repro (doesn't need to stop this merging though).

pcwalton:
@glennw Removed the hack.

PR #91, opened 2015-11-23
Switch to using ipc-channel.

glennw:


glennw:
Needed to be able to rebase to servo master after e10s landed.

Issue #73, opened 2015-11-04
Add WebGL support


ecoal95:
This is a meta-issue to discuss possible WebGL implementations with WebRender.

The first idea I had was keeping the WebGL task as-is, with the context sharing resources with the WebGL context, but that could be racy (that could be manually fixed though) as well as non-performant.

Another idea was to keep the same approach as now, but sending `NativeSurface`s to WebRender. The WebRender context would re-bind the surface to a texture, and paint using that texture. It seems to me that it will work, avoiding the roundrip to main memory (I think that at least in GLX the texture will get the contents of the surface, and probably in other platforms too).

Glenn also suggested using the WebGL task as a command proxy, and do all the paint work in WebRender, avoiding any resource sharing to avoid using context sharing (which is an uncommon code path). In my opinion this won't work out well, since it also adds a lot of complexity, since WebRender would be responsible of keeping each WebGL context *state* separately (that includes a lot of non-trivial state such as viewport, etc.), and restoring the WebRender state, wich will be a considerable overhead.

That got me to think though, that the latter approach, combined with resource sharing could be the best option: Sharing resources between the contexts, using the WebGL task as a proxy (or even taking rid of it), but keeping the contexts in WebRender, and only executing gl commands when required.

Since I probably didn't explain myself very well, this would be an sketch of the proposed implementation:

* When a WebGL context is (trying to be) created, the WebGL paint task requests WebRender to create a context.
* WebRender creates a new gl context, sharing resources with the main context, associates it with an unique id and returns that id (or the error, if that's the case) to the WebGL task.
* The WebGL task just serializes the gl commands received from script in a form like `(context_id, command)`.
* WebRender, before painting, takes that commands, sorts them by context id (to minimize context switching), and applies them to the given context. Note that there are special commands that need a reply from our side, like creating resources. When they arrive we could either do nothing until the next paint, when it will be processed, or flush the command buffer executing all the previous commands. The former is arguably simpler, but will make creating resources really slow, so I'd go for flushing the command buffer.
* WebRender does the actual paint, using the shared texture id's from the contexts.

I think this approach will be at least as performant as any other multithreaded approach, since most (if not all) the OpenGL implementations are single threaded, and this approach minimizes context switching, and prevents any kind of races that could be between WebRender and WebGL.

Note that the WebGL contexts should draw to a texture, but that's already done in the current implementation.

Also the WebGL task would be kind of naive with this scheme, so we could get rid of it easily I think.

[...]

PR #99, opened 2015-12-01
Add WebGL support

ecoal95:
For now it only has support for `clear`/`clearColor`, but adding support for more is easy (see `apply_webgl_command`).

It depends on a few more PRs (to servo and webrender_traits) and on the `webrender` branch of `offscreen_gl_context`, that currently lacks support for OS X since I don't own a Mac, but it should fail gracefully (just returning `null` on context creation).

It uses the last approach discussed in #73 

cc: @pcwalton @mrobinson

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -288,6 +288,7 @@ impl FBOId {
     fn bind(&self) {
         let FBOId(id) = *self;
         gl::bind_framebuffer(gl::FRAMEBUFFER, id);
+        gl::bind_framebuffer(gl::READ_FRAMEBUFFER, id);
     }
 }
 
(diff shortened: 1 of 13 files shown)


src/device.rs
```diff
@@ -288,6 +288,7 @@ impl FBOId {
     fn bind(&self) {
         let FBOId(id) = *self;
         gl::bind_framebuffer(gl::FRAMEBUFFER, id);
+        gl::bind_framebuffer(gl::READ_FRAMEBUFFER, id);
```
glennw: I don't think we need this - can it be removed?

glennw:
This looks good apart from the one comment.

src/device.rs
ecoal95: Yeah sure, It's been a rebase issue. Good catch, sorry!

PR #119, opened 2015-12-23
Partially fix blur filters.

pcwalton:
* The temporary framebuffer texture is now unused, since we use separate
  render targets instead of readback. This patch removes it.

* We now write the texture sizes into the mask UV coordinates properly.

* We allow composite filters to inflate the dimensions of their render
  targets. This means that blurs can draw outside their rects.

Remaining issues:

* Blur filters need to not read outside of the render target in the
  atlas.

* The vertical blur needs to blur the render target that the horizontal
  blur wrote to rather than the original render target.

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -733,27 +733,39 @@ impl<'a> SceneItemKind<'a> {
                                              info,
                                              context);
                 } else {
-                    let render_target_size = Size2D::new(overflow.size.width as u32,
-                                                         overflow.size.height as u32);
+                    let target_size = Size2D::new(overflow.size.width as i32,
+                                                  overflow.size.height as i32);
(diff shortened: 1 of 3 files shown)


pcwalton:
I'd be interested to know what you think the best solution for the second problem here is. A blur essentially needs to result in the creation of two nested render targets. At first I just tried to special case this in the composite batching code (creating a second render target for a blur) but that seemed inelegant, especially since we'll probably want to be able to batch them (i.e. blur many elements horizontally, then blur many elements vertically). So I figured it'd be better to hook this into the layers/render target infrastructure somehow—essentially to create two nested render targets for each blur. Thoughts on that?

pcwalton:
@glennw Fixed mix-blend-mode.

glennw:
Thanks!

PR #122, opened 2015-12-24
Fix a couple of correctness issues around border-radius.

pcwalton:
1. Masks for large border radii were not being tessellated properly.
This commit simply switches off tessellation for them. Eventually we may
want to go back and fix this properly.

2. Border corners were being rendered correctly when 0 < border radius <
border width. We needed to do a bit more tessellation than we were. This
commit fixes that and adds a diagram to `add_border_corner` to explain
what's going on.

These commits fix the `border_radius_clip_a.html` and
`border_radius_zero_sizes_a.html` reftests.


glennw:
Thanks!

PR #123, opened 2015-12-25
Support groove and ridge border styles.

pcwalton:
Fixes the `border_black_groove.html` and `border_black_ridge.html`
reftests.


glennw:
Thanks!

PR #131, opened 2016-01-07
First pass of GPU clipping support.

glennw:
* Add support for clip in/out rects on GPU.
* Remove CPU polygon clipping code.
* Fix all known vertex cracks and pixel sparkling.

There is still some code cleanup to do, and missing features, such as:
 * Nested complex clip regions.
 * Switching angle gradients to rectangles.
 * Complex clip regions + border radius.

These *should* be fairly simple to implement on top of these changes.


pcwalton:
I get `thread '<unnamed>' panicked at 'assertion failed: border_radius == 0.0', /Users/pcwalton/Source/servo-packages/webrender/src/batch_builder.rs:911` on http://github.com/servo/servo.

glennw:
@pcwalton That assert is because I haven't done the clip support for box-shadow + border-radius yet. I was going to do that as a follow up PR, so I've commented that assert out for now, if you're ok with doing that as a follow up?

PR #133, opened 2016-01-10
Sync core-graphics/core-text versions with Servo

frewsxcv:
Bumping these versions allows compilation with Servo HEAD


glennw:
Thanks!

PR #151, opened 2016-01-26
Free resources from the texture cache as appropriate.

pcwalton:
For now, this just unconditionally frees unused resources from the
previous frame after building a new frame. We may want to throttle this
back at some point.

Also, the cache used is potentially inefficient and can be easily
improved.

We don't free texture pages at the GL level yet.

### DIFF
--- a/src/batch_builder.rs
+++ b/src/batch_builder.rs
@@ -3,6 +3,7 @@ use batch::{BatchBuilder, TileParams};
 use device::TextureId;
 use euclid::{Rect, Point2D, Size2D};
 use fnv::FnvHasher;
+use frame::FrameId;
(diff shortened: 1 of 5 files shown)


src/freelist.rs
```diff
@@ -86,11 +86,12 @@ impl<T: FreeListItem> FreeList<T> {
 
     // TODO(gw): Actually free items from the texture cache!!
     #[allow(dead_code)]
-    pub fn free(&mut self, id: FreeListItemId) {
+    pub fn free(&mut self, id: FreeListItemId) -> &mut T {
```
glennw: It's not safe to do this - the linked list of free items are stored inside the freed items themselves (for memory / perf reasons), so you can't reference the item after it's freed. Can you change the calling code to get the item first, extract the needed data and then free the item?

PR #162, opened 2016-02-01
Support opt-in multisampled antialiasing.

pcwalton:
Multisampled render targets are not yet supported.


pcwalton:
Corresponding Servo WebRender branch PR: https://github.com/glennw/servo/pull/23

PR #164, opened 2016-02-02
Reimplement arbitrary rotation angles for linear gradients by clipping in the fragment shader.

### LINKED ISSUE #161: Gradient not painted correctly
```html
<style>
  div {
    width: 50px;
    height: 50px;
    background-color: #00f;
    margin: 20px;
  }
</style>

<div style="background-image: linear-gradient(135deg, #F00 50%, transparent 50%)"></div>
<div style="background-image: linear-gradient(135deg, #F00,     transparent 50%)"></div>
```

screen shot 2016-02-01 at 11 31 19

pcwalton:
This reintroduces the clip-in rect in the FS. Since we'll need it anyway
when we have arbitrary matrix/clip stacks, this seems harmless.

Color strips for non-axis-aligned linear gradients are submitted as
axis-aligned rects centered at the appropriate point, just as
axis-aligned ones are. However, we reuse one of the vertex fields to
store the rotation angle. The vertex shader detects this condition,
rotates the rectangle around its midpoint, and passes the clip rect on
to the fragment shader.

Additionally, this patch eliminates the ugly (-10000,10000) spans for
linear gradient strips that we had before. It was necessary to tighten
this up in order to avoid needless fragment shader invocations in the
non-axis-aligned case, so I went ahead and fixed it everywhere.

Closes #161.

### DIFF
--- a/res/quad.fs.glsl
+++ b/res/quad.fs.glsl
@@ -8,8 +8,13 @@ bool PointInRect(vec2 p, vec2 p0, vec2 p1)
 
 void main(void)
 {
-    // Clip out rect
+    // Clip out.
     if (PointInRect(vPosition, vClipOutRect.xy, vClipOutRect.zw)) {
+        discard;
+    }
+
+    // Clip in.
+    if (vClipInRect != vec4(0.0) && !PointInRect(vPosition, vClipInRect.xy, vClipInRect.zw)) {
         discard;
     }
 
--- a/src/node_compiler.rs
+++ b/src/node_compiler.rs
@@ -65,6 +65,8 @@ impl NodeCompiler for AABBTreeNode {
                             builder.push_clip_in_rect(clip_rect);
                             builder.push_complex_clip(&display_item.clip.complex);
 
+                            println!("clip rect={:?}", clip_rect);
+
                             match display_item.item {
                                 SpecificDisplayItem::WebGL(ref info) => {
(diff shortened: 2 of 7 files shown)


res/quad.fs.glsl
```diff
@@ -8,11 +8,16 @@ bool PointInRect(vec2 p, vec2 p0, vec2 p1)
...
 void main(void)
 {
-    // Clip out rect
+    // Clip out.
     if (PointInRect(vPosition, vClipOutRect.xy, vClipOutRect.zw)) {
         discard;
     }
 
+    // Clip in.
+    if (vClipInRect != vec4(0.0) && !PointInRect(vPosition, vClipInRect.xy, vClipInRect.zw)) {
```
glennw: I don't think you need this if () condition - I'm fairly sure it wasn't in the shader last time?

glennw: Oh I see what's happening - perhaps just set it to a large value in the VS to avoid a conditional in the FS?

src/node_compiler.rs
```diff
@@ -65,6 +65,8 @@ impl NodeCompiler for AABBTreeNode {
                             builder.push_clip_in_rect(clip_rect);
                             builder.push_complex_clip(&display_item.clip.complex);
 
+                            println!("clip rect={:?}", clip_rect);
```
glennw: Can be removed

PR #173, opened 2016-02-05
Add support for GPU generated raster ops to have texture atlas border edges.

glennw:
Raster ops are now generated to a single render target (per format) allocated at start up.

After a raster batch is rendered, it is then blitted into the atlas, along with border edges if requested.

This is necessary since it's not possible to read from a texture bound as a render target.

This fixes several artifacts that are often noticeable with box-shadow etc.

Also fix a couple of warnings.


pcwalton:
Looks good.

Issue #177, opened 2016-02-05
overflow issue: parent's transform not taken into account


paulrouget:
```html
<style>

  .green {
    background: green; 
    position: absolute;
    width: 100%;
    height: 100px;
    transform: translateX(-200px);
  }

  .red {
    background: red;
    position: absolute;
    width: 100px;
    height: 100px;
    right: -50px;
  }

</style>

<div class="green">
  <div class="red"></div>
</div>
```

<img width="905" alt="screen shot 2016-02-05 at 08 59 49" src="https://cloud.githubusercontent.com/assets/373579/12840870/2b0f0e40-cbe7-11e5-9c2c-8a97634c32d1.png">

^ non WR vs WR.

The red div is cut as if it was offscreen, even though its parent `transform:translate` makes it onscreen.

PR #180, opened 2016-02-09
Use stencil buffer for clipping iframes and aabb tree nodes.

glennw:
(This allows correct clipping for these cases when transforms are involved).

This is a partial fix for #177 (the other part relies on changes to layout).

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -64,6 +64,19 @@ impl MatrixHelpers for Matrix4 {
     }
 }
 
+pub trait RectHelpers {
+    fn contains_rect(&self, other: &Rect<f32>) -> bool;
+}
+
+impl RectHelpers for Rect<f32> {
+    fn contains_rect(&self, other: &Rect<f32>) -> bool {
+        self.origin.x <= other.origin.x &&
+        self.origin.y <= other.origin.y &&
+        self.max_x() >= other.max_x() &&
+        self.max_y() >= other.max_y()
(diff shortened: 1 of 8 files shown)


glennw:
r? @pcwalton 

(This relies on a related PR in servo before it can be merged).

pcwalton:
Could we file a followup to add the number of stencil buffer changes to `debug_render`? It would be nice to be able to diagnose the problem quickly if we see pages in the wild that thrash the stencil buffer.

src/util.rs
```diff
@@ -64,6 +64,19 @@ impl MatrixHelpers for Matrix4 {
...
+pub trait RectHelpers {
+    fn contains_rect(&self, other: &Rect<f32>) -> bool;
+}
+
+impl RectHelpers for Rect<f32> {
+    fn contains_rect(&self, other: &Rect<f32>) -> bool {
+        self.origin.x <= other.origin.x &&
+        self.origin.y <= other.origin.y &&
+        self.origin.x + self.size.width >= other.origin.x + other.size.width &&
+        self.origin.y + self.size.height >= other.origin.y + other.size.height
```
pcwalton: nit: Could use `max_x()` and `max_y()` I think

pcwalton:
Looks good.

glennw:
Done - https://github.com/glennw/webrender/issues/181

PR #187, opened 2016-02-18
Fix speed regressions in spheres demo

glennw:


glennw:
The first commit (to store the local raster origin) is the most important change, but both of these drastically reduce the number of box shadow raster ops in the spheres demo.

On my machine, the spheres demo easily runs at 60fps with these changes.

r? @pcwalton

pcwalton:
Thanks so much!

PR #199, opened 2016-02-25
Reenable border corner tessellation for monochromatic borders.

pcwalton:
Multicolored borders aren't tessellated for now; this can be a followup.

r? @glennw

PR #200, opened 2016-02-26
Start texture atlases off small, and allow them to grow.

### LINKED ISSUE #193: moires runs very slow
really amazing demo! I tried cloning servo and compiling it myself, but I'm pretty consistently getting 5 fps on the moire demo (on a pretty recent macbook pro).

 - I compiled with `./mach build --release`
 - I cloned moire.html from https://github.com/pcwalton/webrender-demos
 - I ran servo with `target/release/servo -Z wr-stats -w --resolution=1024x768 --pref dom.mozbrowser.enabled ../webrender-demos/moire.htm` (I copied it straight from the demo stream)

this is what I'm seeing: (note the 4 fps)

Interestingly, this seems to be limited to moires. Both spheres and transparent_rects run at ~60fps.

Any thoughts?

pcwalton:
Includes #199.

Closes #193.

r? @glennw

### DIFF
--- a/src/freelist.rs
+++ b/src/freelist.rs
@@ -97,4 +97,72 @@ impl<T: FreeListItem> FreeList<T> {
         item.set_next_free_id(self.first_free_index);
         self.first_free_index = Some(id);
     }
+
+    /// NB: This iterates over free items too!
+    pub fn iter_mut(&mut self) -> IterMut<T> {
+        let first_free_index = self.first_free_index;
+        let mut iterator = IterMut {
+            free_list: self,
+            next_index: Some(FreeListItemId(0)),
+            next_free_index: first_free_index,
+        };
+        iterator.advance_past_free_indices();
+        iterator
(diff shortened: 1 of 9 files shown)


src/freelist.rs
```diff
@@ -97,4 +99,9 @@ impl<T: FreeListItem> FreeList<T> {
         item.set_next_free_id(self.first_free_index);
         self.first_free_index = Some(id);
     }
+
+    /// NB: This iterates over free items too!
```
glennw: I'm uneasy about this since it exposes internal details - is there a way to avoid it?

pcwalton: Maybe I can yield indices instead—let me see.

pcwalton:
Fixed. r? @glennw

frewsxcv:
Just published heapsize_plugin 0.1.4, which *should* cause Travis to pass again.

PR #204, opened 2016-02-27
Fix documentation in batch_builder

nox:
Blocks annotated with `ignore` are not tested, but are still considered Rust code.


jdm:
Serde just released a new version :(

PR #206, opened 2016-02-28
Temporary fix for Windows

UK992:
Temporary solution to pass building on Windows. 
Related to https://github.com/servo/servo/issues/9775


larsbergstrom:
r? @glennw 

This would at least get us building until there is a "real" WebRender backend :-)

Thanks, @UK992!

larsbergstrom:
Gah, I think we're blocked on landing anything via CI until bincode is fixed (https://github.com/TyOverby/bincode/pull/57).

zeusintuivo:
I have researched and documented an effective way to have windows environments set up for 32 and 64 bits in Windows 7. I assume this will work in all windowses. 
Please see documentation I posted in this ticket. #12356

PR #207, opened 2016-02-29
Support for fixed layer per iframe

glennw:


src/frame.rs
```diff
@@ -1119,6 +1142,21 @@ impl Frame {
         if let Some(root_scroll_layer_id) = self.root_scroll_layer_id {
             self.update_layer_transform(root_scroll_layer_id, &Matrix4::identity());
         }
+
+        // Updated any fixed layers
```
pcwalton: typo: "update"

pcwalton:
One typo, otherwise looks good

PR #208, opened 2016-02-29
Resize uvs when growing textures.

pcwalton:
Fixes browser.html.

r? @glennw

### DIFF
--- a/src/texture_cache.rs
+++ b/src/texture_cache.rs
@@ -461,7 +461,6 @@ pub struct TextureCacheItem {
     pub user_data: TextureCacheItemUserData,
 
     // The texture coordinates for this item
-    pub uv_rect: RectUv<f32>,
     pub pixel_rect: RectUv<DevicePixel>,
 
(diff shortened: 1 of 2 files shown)


src/texture_cache.rs
```diff
@@ -817,7 +817,20 @@ impl TextureCache {
...
                             if item.texture_id == old_texture_id {
                                 item.texture_id = new_texture_id;
-                                item.texture_size = Size2D::new(texture_size, texture_size)
+                                let width_factor = texture_size as f32 /
+                                    item.texture_size.width as f32;
+                                let height_factor = texture_size as f32 /
+                                    item.texture_size.height as f32;
+                                item.texture_size = Size2D::new(texture_size, texture_size);
+
+                                item.uv_rect.top_left.x /= width_factor;
```
glennw: @pcwalton Is this going to introduce float accuracy issues after a number of resizes? Should we directly recalculate from the pixel rect each time instead?

pcwalton:
@glennw r?

PR #211, opened 2016-03-01
Switch to beta rust for travis, to allow us to land WR changes until …

glennw:
…serde/bincode fixes land.


larsbergstrom:
gah! WR uses `feature`, so can't work on beta. Try:
```
rust:
 - nightly-2016-02-22
```

PR #214, opened 2016-03-01
Fix loop count from 7b9c8e638

kevinmehall:
After updating WebRender, [this box shadow](https://github.com/kevinmehall/webrender-experiments/blob/master/src/main.rs#L127) rendered incorrectly.
`git bisect` pointed to 7b9c8e638, where I noticed that the replacement
code wasn't quite equivalent -- `self.vertices` vs `vertices`

Before:

After:


glennw:
Thanks!

PR #217, opened 2016-03-03
Implement support for overscrolling on the Mac.

pcwalton:
This generalizes scroll info to a separate structure, which paves the
way toward supporting zoom as well.

r? @glennw 
cc @paulrouget


glennw:
@pcwalton This looks good to me - will need a rebase.

PR #228, opened 2016-03-09
Allow core-graphics 0.3

nox:


nox:
@bors-servo r-

Didn't delete Cargo.lock.

PR #231, opened 2016-03-10
Flatten display list transfer into a separately-transferred byte array, and consolidate the multitude of display list messages into one.

pcwalton:
Along with the corresponding patches to other Servo repositories, this
improves display list building performance on small display lists by
about 3x.

r? @glennw


jdm:
Hmm, we pinned the nightly that travis uses but continue to use nightly serde, apparently.

PR #232, opened 2016-03-10
Try to make the GPU profiling code compile on Gonk.

pcwalton:
r? @glennw


pcwalton:
@bors-servo: retry

Travis cargo network failure

PR #237, opened 2016-03-12
Make the desktop no longer show through when alpha blending is involved for borderless windows on the Mac.

pcwalton:
This was because the blend function was
`GL_SRC_ALPHA`/`GL_ONE_MINUS_SRC_ALPHA` for both RGB and alpha, and that
was incorrect for alpha, since a destination alpha of 1.0 and a source
alpha < 1.0 should result in an alpha of 1.0.

r? @glennw

### DIFF
--- a/src/debug_render.rs
+++ b/src/debug_render.rs
@@ -168,7 +168,8 @@ impl DebugRenderer {
             gl::disable(gl::DEPTH_TEST);
             gl::enable(gl::BLEND);
             gl::blend_equation(gl::FUNC_ADD);
-            gl::blend_func(gl::SRC_ALPHA, gl::ONE_MINUS_SRC_ALPHA);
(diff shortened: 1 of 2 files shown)


pcwalton:
cc @paulrouget

paulrouget:
Appears to work well here.

paulrouget:
Fix https://github.com/servo/servo/issues/9877

glennw:
@pcwalton This sounds right - could we just check that running against the mozilla/ reftests dir and webrender-demos doesn't introduce any regressions?

pcwalton:
@glennw Checked those tests. They seem fine. (Some of them fail for me, but I inspected them manually and they look fine. It looks like a problem with the harness, probably pre-existing and maybe Mac-specific.)

PR #247, opened 2016-03-21
Use Linux font config on Windows platform

UK992:
It's slightly faster then normal, not much, but it works.
I did this same change few weeks ago and it didn't work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+Cargo.lock
 target/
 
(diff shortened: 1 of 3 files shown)


glennw:
Does this mean that freetype impl works on Windows? cc @larsbergstrom

UK992:
Yes, it works.

adamncasey:
Nice. What's needed to merge this?

PR #248, opened 2016-03-23
Add appveyor config for Windows CI support.

frewsxcv:
Fixes https://github.com/servo/webrender/issues/246.

PR #250, opened 2016-03-27
webgl: Return the GLLimits structure to the caller

emilio:
See https://github.com/servo/servo/pull/10224 and https://github.com/servo/webrender_traits/pull/26.

r? @glennw


frewsxcv:
Cherry-picked in https://github.com/servo/webrender/pull/248

PR #251, opened 2016-03-29
Handle nested transforms and clips when the transform consists solely of a translation.

pcwalton:
This is a partial fix for the underlying issue, sufficient to address
servo/servo#10164.

r? @glennw


paulrouget:
Can we cargo-update webrender to get that into servo? (there's some build issues with servo master + wr master)

emilio:
@paulroget You should be able to update normally once https://github.com/servo/servo/pull/10224 has landed

PR #252, opened 2016-03-30
Change compile_shader debug println to log

dporru:
This should fix https://github.com/servo/servo/issues/10273


paulrouget:
@glennw / @pcwalton can you guys look at this?

PR #255, opened 2016-03-31
Clip iframes to clipping regions that consist of single rectangles.

pcwalton:
This is a prerequisite to solving servo/servo#9983.

Requires servo/webrender_traits#28.

r? @glennw


glennw:
@pcwalton Looks good to me once rebased

PR #256, opened 2016-03-31
Implement a less broken fix for nested transforms and clips.

pcwalton:
The fix in PR #251 was broken.

Addresses servo/servo#10307.

r? @glennw


glennw:
@pcwalton Could you check that the wpt/mozilla tests still pass with this?

pcwalton:
@glennw Verified that the Mozilla tests pass. The CSSWG tests kernel panic my system, so I couldn't complete the run of those, but they didn't look out of the ordinary.

PR #257, opened 2016-04-01
Brand each payload message with the stacking context ID and epoch to avoid races.

pcwalton:
Addresses servo/servo#10256.

Needs servo/webrender_traits#31.

r? @glennw


glennw:
@pcwalton This looks good to me once updated to include the traits change.

PR #259, opened 2016-04-14
Update lazy_static.

Ms2ger:


jdm:
Aster is broken again.

PR #260, opened 2016-04-15
Update Rust version in travis configuration

mrobinson:
Servo has updated its rust version, so we need to update the travis
configuration in WebRender.


mrobinson:
Closing this as it needs to be tested in tandem with updates to handle recent changes in webrender_traits: https://github.com/servo/webrender/pull/261

PR #271, opened 2016-05-03
Use the new `webrender_traits` types to avoid copying buffers when new frames come in.

pcwalton:
This improves performance.

Depends on servo/webrender_traits#46.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/271)

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -389,12 +389,13 @@ impl RenderTarget {
     }
 }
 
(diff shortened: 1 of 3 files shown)


glennw:
@pcwalton Is this still needed?

glennw:
@pcwalton What's the status of this?

PR #272, opened 2016-05-04
Send back the corresponding pipeline ID when translating a point to layer space.

pcwalton:
Fixes mouse events in browser.html.

Requires servo/webrender_traits#47.

r? @glennw


pcwalton:
@bors-servo: retry

Travis didn't seem to pick up the traits change

PR #276, opened 2016-05-24
Update texture size properly when grow texture

kilobtye:
This should fix servo/servo#11323


pcwalton:
@bors-servo: r+

Awesome!

PR #277, opened 2016-05-25
Add functionality to enable Servo to fetch scroll positions for all stacking contexts.

pcwalton:
Requires servo/webrender_traits#55.

Partially addresses servo/servo#11108.

r? @glennw


glennw:
@pcwalton This part looks good to me - just a comment / question on the related servo PR.

PR #279, opened 2016-05-28
Add some simple optimizations to skip clears and stencil buffer operations.

pcwalton:
This series of optimizations speeds up painting of browser.html by 50% or more on my MacBook Pro/Intel Iris. It allows us to reach 60 FPS in some cases (though not reliably) even when YouTube is playing and visible and a transparent blurry Terminal is in the background.

r? @glennw

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -1447,18 +1447,3 @@ impl Frame {
     }
 }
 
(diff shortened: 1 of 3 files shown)


glennw:
Nice one!

PR #284, opened 2016-06-02
Track visible areas of composited layers and cull them out if off screen.

pcwalton:
Reduces the number of draw calls on the Google SERPs.

r? @glennw

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use batch::{RasterBatch, VertexBufferId};
+use bit_set::BitSet;
 use debug_render::DebugRenderer;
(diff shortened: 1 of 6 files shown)


src/renderer.rs
```diff
@@ -1422,269 +1480,291 @@ impl Renderer {
...
+                        draw_simple_triangles(&mut self.simple_triangles_vao,
+                                              &mut self.device,
+                                              &mut self.profile_counters,
+                                              &indices[..],
+                                              &vertices[..],
+                                              info.texture_id);
+                    } else {
+                        for job in &info.jobs {
+                            let ChildLayerIndex(child_layer_index) = job.child_layer_index;
+                            let src_target = &layer.child_layers[child_layer_index as usize];
```
jdm: What am I missing here? What is this loop doing?

pcwalton: Uh...nothing! Oops.

PR #285, opened 2016-06-07
Inset box shadow fixes

### LINKED ISSUE #183: Simple inset box shadows don't draw anything
```
<style>
  html, body {
    margin: 0;
  }
  .b {
    width: 300px;
    height: 20px;
    margin: 50px;
    background: green;
    box-shadow: inset 0 0 0 2px #0F0;
  }

</style>


  <div class="b">
  </div>
```

anderco:
Hi,

I found this project interesting and figured I try to contribute, so I looked at the open issues and tried to fix #183. I saw a couple more issues with inset box shadows, so here's an attempt to fix those.

I'm not exactly sure what's the best way to test these changes. I did manual testing, but pointers for a better method are appreciated.

Thanks,
Ander

### DIFF
--- a/src/batch_builder.rs
+++ b/src/batch_builder.rs
@@ -819,6 +819,10 @@ impl<'a> BatchBuilder<'a> {
                                                                    border_radius,
                                                                    clip_mode);
 
+        if clip_mode == BoxShadowClipMode::Inset && blur_radius == 0.0 {
+            return;
+        }
+
         // Draw the sides.
         //
         //      +--+------------------+--+
--- a/src/batch_builder.rs
+++ b/src/batch_builder.rs
@@ -642,7 +642,7 @@ impl<'a> BatchBuilder<'a> {
                           clip_mode: BoxShadowClipMode,
                           resource_cache: &ResourceCache,
                           frame_id: FrameId) {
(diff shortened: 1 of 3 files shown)


glennw:
Thanks! r? @pcwalton

src/batch_builder.rs
```diff
@@ -1884,12 +1884,19 @@ impl BoxShadowMetrics {
...
-                               spread_radius: f32)
+                               spread_radius: f32,
+                               clip_mode: BoxShadowClipMode)
                                -> Rect<f32> {
     let mut rect = (*box_bounds).clone();
     rect.origin.x += box_offset.x;
     rect.origin.y += box_offset.y;
-    rect.inflate(spread_radius, spread_radius)
+
+    let _spread = match clip_mode {
```
pcwalton: I would just make this:

    if clip_mode == BoxShadowClipMode::Inset {
        spread_radius = -spread_radius
    }

and change `spread_radius` in the function signature above to `mut spread_radius`.

pcwalton:
Looks good other than that one nit. Thanks a lot!

src/batch_builder.rs
anderco: Makes sense. I'll update.

PR #287, opened 2016-06-10
Rewrite the flatten transform/clip code to more clearly separate local transform/perspective and world transform/perspective.

pcwalton:
Addresses servo/servo#11650.

r? @glennw


glennw:
@pcwalton r+

Issue #166, opened 2016-02-02
text antialiasing issue with opacity < 1 


paulrouget:
I guess it's an antialiasing issue. The top text here has a dark outline. That makes text opacity animations very ugly.

<img width="1095" alt="screen shot 2016-02-02 at 08 39 41" src="https://cloud.githubusercontent.com/assets/373579/12742761/dec7e094-c988-11e5-8ba4-602270626cdb.png">

```html
<style>
  body {
    padding: 20px;
    background: rgb(65, 149, 250);
  }
</style>

<div style="opacity: 0.8; color: white">FOOBAR</div>
<div style="opacity: 1.0; color: white">FOOBAR</div>
```

pcwalton:
$10 says it's one of those annoying `glBlendFunc()`/premultiplied alpha issues.

glennw:
Yep, this is occurring on Linux as well, so probably an issue like @pcwalton mentioned above rather than an issue with the platform font rasterizing.

glennw:
@pcwalton Changing https://github.com/glennw/webrender/blob/master/src/renderer.rs#L988 from 0,0,0,0 -> 1,1,1,0 fixes this issue. However, it introduces artifacts into the box shadows in the ref tests.

I think this is a correct fix - any thoughts on what effect that would have on box shadow and if it can be worked around? I'll investigate further - just posting this in case you know straight away what the issue might be?

glennw:
Interestingly the blur on text_shadow_blur_a.html remains the same with that change (which is a text-shadow), but the blur on text_shadow_blur_ref.html (which is a box shadow) is what starts to show artifacts.

pcwalton:
@glennw Hmm, but I think our blend mode is wrong if we're relying on (1,1,1,0) instead of (0,0,0,0).

glennw:
I'm not sure what blend mode you would use.

A traditional alpha blend is:

Result.rgb = Dest.rgb*(1-Source.alpha) + Source.rgb * Source.alpha.

Pre-multipled alpha blend is:

Result.rgb = Dest.rgb*(1-Source.alpha) + Source.rgb.

In both cases, the destination color is going to be modulated by the inverse source alpha, so you will get black fringing if drawing text with alpha on a black FBO, which is the case for text with opacity at the moment.

Or is there another alternative?

glennw:
Hmmm, perhaps we could make of separate blend func to store the alpha but just use the color. I'll try that.

glennw:
[...]

PR #288, opened 2016-06-10
Enable blending on render targets.

pcwalton:
It was disabled as a fix for #166, but that caused regressions on pages
that depend on blending. The issues related to the blur on box shadows
discussed in that issue seem to be resolved now, so the clear-color
solution works fine.

Addresses browserhtml/browserhtml#1068.

r? @glennw


glennw:
The change itself looks fine to me - but it's hard to think of all situations that this might break - so long as this has been tested on a variety of sites / pages with blends, r=me.

pcwalton:
I ran it through the Mozilla CSS tests and it seemed OK, for what it's worth.

PR #289, opened 2016-06-10
Minor fix of box-shadow

kilobtye:
Avoid wrong clipping when box shadow have offsets


pcwalton:
@bors-servo: r+

Thanks!

PR #291, opened 2016-06-11
Remove TranslatePointToLayerSpace handler

aneeshusa:
As of servo/servo#11537 this is no longer used, as Servo now uses
a unified code path for mouse events whether WebRender is in use or not.
(Mouse events are always dispatched to the root pipeline, removing the
need to translate into layer space.)

Requires servo/webrender_traits#56. Part of servo/servo#10808.

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -1496,28 +1496,5 @@ impl Frame {
         }
         layers_bouncing_back
     }
-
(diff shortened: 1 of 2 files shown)


glennw:
@aneeshusa What's the status of this?

aneeshusa:
I don't have time to work on this right now.

PR #292, opened 2016-06-15
resource-cache: Allow deallocating images

emilio:
This will allow us leaking a lot of canvas memory while using readback.

Also we should probably audit where are these used and when can they be
deallocated.

Depends on: https://github.com/servo/webrender_traits/pull/57

r? @glennw

### DIFF
--- a/src/render_backend.rs
+++ b/src/render_backend.rs
@@ -114,6 +114,9 @@ impl RenderBackend {
                                                                       format,
                                                                       bytes);
                         }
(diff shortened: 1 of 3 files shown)


glennw:
@emilio We'll need to remove it from the texture cache too.

emilio:
@glennw: Don't they get automatically removed on `expire_old_resources`?

I don't think I can figure out a way to obtain the texture id otherwise, plus it could cause races.

PR #294, opened 2016-06-17
Truncate glyph positions to whole device pixels

mbrubeck:
This addresses servo/servo#11751: The quad shader rounds rect boundaries to the nearest pixel.  This can cause glyphs' relative positions to change if one glyph is shifted left while the next is shifted right.  Using `floor()` ensures that all glyphs are shifted in the same direction.

r? @glennw

### DIFF
--- a/src/batch_builder.rs
+++ b/src/batch_builder.rs
@@ -436,8 +436,8 @@ impl<'a> BatchBuilder<'a> {
             glyph_key.index = glyph.index;
             let image_info = resource_cache.get_glyph(&glyph_key, frame_id);
             if let Some(image_info) = image_info {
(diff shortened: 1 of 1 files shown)


mbrubeck:
Note: The root cause of this is the behavior of the GLSL `round` function on some implementations:

>  **round** returns a value equal to the nearest integer to *x*. The fraction 0.5 will round in a direction chosen by the implementation, presumably the direction that is fastest. This includes the possibility that **round**(*x*) returns the same value as **roundEven**(*x*) for all values of *x*. 

—https://www.opengl.org/sdk/docs/man4/html/round.xhtml

After talking about this some more with @pcwalton, I think it might make sense to change the quad shader to not use `round` (but instead do something like `floor(x + 0.5)`), since the `roundEven` behavior allowed for `round` is likely to have negative effects on items besides font glyphs too.

PR #297, opened 2016-06-28
Remove all compilation warnings

nox:


src/renderer.rs
```diff
@@ -1479,282 +1479,279 @@ impl Renderer {
...
+                                        gl::disable(gl::BLEND);
+                                        (6.0, amount.to_f32_px(), 0.0, 0.0)
+                                    }
+                                    LowLevelFilterOp::Brightness(_) |
+                                    LowLevelFilterOp::Opacity(_) => {
+                                        // Expressible using GL blend modes, so not handled
+                                        // here.
+                                        unreachable!()
+                                    }
+                                };
```
pcwalton: These are just indentation changes, right?

nox: Yes I removed the dead code and then realised I could just do an early `continue`.

See https://github.com/servo/webrender/pull/297/files/baf8a0f94ec2acf3903e9d1bd924b1a6986c20bf..ddf0f25dca6d3b1c3031033158e5da80695d6129?w=1#r68850332.

PR #299, opened 2016-06-29
Omit nested clip regions if internal one is their subregion

splav:
In case of nested clipping regions it is common that external ones do contain the most internal clipping region and essentially only that internal one should be used.
I'm not sure that this is the correct path, but it's rather common ([servo/issues/11562](https://github.com/servo/servo/issues/11562), [servo/issues/11857](https://github.com/servo/servo/issues/11857)) and at least can be used as a fast path if the supposed full of nested clipping regions is rather heavy.

This PR implements this fast path. It relies on the fact that int nested regions list the last is the innermost.

While not sure that it's a good idea at all, looks like a hack, but still...

If there are any thought hot to implement full nested regions support I'll try to implement those (Or may be it will be done in WR2 and there is no need to do this right now).

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -16,7 +16,7 @@ use scene::{SceneStackingContext, ScenePipeline, Scene, SceneItem, SpecificScene
 use scoped_threadpool;
 use std::collections::{HashMap, HashSet};
 use std::hash::BuildHasherDefault;
-use tiling::{Clip, FrameBuilder, FrameBuilderConfig};
+use tiling::{Clip, FrameBuilder, FrameBuilderConfig, InsideTest};
 use util::{MatrixHelpers};
 use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLayerId};
 use webrender_traits::{StackingContext, FilterOp, MixBlendMode};
@@ -495,10 +495,17 @@ impl Frame {
 
                     for item in &draw_list.items {
                         let clips = auxiliary_lists.complex_clip_regions(&item.clip.complex);
-                        let clip = if clips.is_empty() {
-                            None
-                        } else {
-                            Some(Box::new(Clip::from_clip_region(&clips[0])))
+                        let clip = match clips.len() {
+                            0 => None,
+                            1 => Some(Box::new(Clip::from_clip_region(&clips[0]))),
+                            _ => {
+                                let internal_clip = clips.last().unwrap();
+                                if clips.iter().all(|current_clip| current_clip.might_contain(internal_clip)) {
+                                    Some(Box::new(Clip::from_clip_region(internal_clip)))
+                                } else {
+                                    Some(Box::new(Clip::from_clip_region(&clips[0])))
+                                }
+                            },
                         };
 
                         match item.item {
@@ -933,4 +940,3 @@ impl Frame {
         }
     }
 }
-
--- a/src/tiling.rs
+++ b/src/tiling.rs
@@ -2934,3 +2934,31 @@ impl PackedPrimitiveCache {
     }
 }
 
+//Test for one clip region contains another
+pub trait InsideTest<T> {
+    fn might_contain(&self, clip: &T) -> bool;
+}
+
+impl InsideTest<ComplexClipRegion> for ComplexClipRegion {
+    // Returns true if clip is inside self, can return false negative
+    fn might_contain(&self, clip: &ComplexClipRegion) -> bool {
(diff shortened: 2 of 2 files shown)


src/batch.rs
```diff
@@ -453,3 +460,30 @@ impl RasterBatch {
         false
     }
 }
+
+trait InsideTest<T> {
+    fn contains_may_be_false_negative(&self, clip: &T) -> bool;
```
pcwalton: I would just call this `might_contain` and clarify in the comments that this may be a false negative.

src/batch.rs
```diff
@@ -453,3 +460,30 @@ impl RasterBatch {
...
+}
+
+impl InsideTest<ComplexClipRegion> for ComplexClipRegion {
+    fn contains_may_be_false_negative(&self, clip: &ComplexClipRegion) -> bool {
+        print!("contains_may_be_false_negative: {:?} {:?}\n", self, clip);
```
pcwalton: Remove this debugging code.

src/batch.rs
```diff
@@ -453,3 +460,30 @@ impl RasterBatch {
...
+        let delta_left = clip.rect.origin.x - self.rect.origin.x;
+        let delta_top = clip.rect.origin.y - self.rect.origin.y;
+        let delta_right = self.rect.origin.x + self.rect.size.width - (clip.rect.origin.x + clip.rect.size.width);
```
pcwalton: Could use `self.rect.max_x()` and `clip.rect.max_x()` here for brevity

src/batch.rs
```diff
@@ -453,3 +460,30 @@ impl RasterBatch {
...
+        let delta_bottom = self.rect.origin.y + self.rect.size.height - (clip.rect.origin.y + clip.rect.size.height);
```
pcwalton: Could use `self.rect.max_y()` and `clip.rect.max_y()` here for brevity

src/batch.rs
```diff
@@ -291,10 +291,17 @@ impl<'a> BatchBuilder<'a> {
...
         // TODO(gw): Handle nested complex clips!
-        if clip.len() > 0 {
-            self.complex_clip = Some(clip[0]);
-        } else {
-            self.complex_clip = None;
+        self.complex_clip = match clip.len() {
+            0 => None,
+            n if n > 1 => {
+                let internal_clip = clip.last().unwrap();
+                if clip[..clip.len()].iter().all(|current_clip| current_clip.contains_may_be_false_negative(internal_clip)) {
```
pcwalton: Why not just `clip.iter().all(...)`?

src/batch.rs
```diff
@@ -291,10 +291,17 @@ impl<'a> BatchBuilder<'a> {
...
+                    Some(*internal_clip)
+                } else {
+                    Some(clip[0])
+                }
+            },
+            _ => Some(clip[0]),
```
pcwalton: Could put this one before the above match and say `1 => Some(clip[0])`, then remove the guard

src/batch.rs
glennw: Should be able to just us clip.iter()... I think

src/batch.rs
```diff
@@ -291,10 +291,17 @@ impl<'a> BatchBuilder<'a> {
 
     pub fn push_complex_clip(&mut self, clip: &[ComplexClipRegion]) {
         // TODO(gw): Handle nested complex clips!
-        if clip.len() > 0 {
-            self.complex_clip = Some(clip[0]);
-        } else {
-            self.complex_clip = None;
+        self.complex_clip = match clip.len() {
+            0 => None,
+            n if n > 1 => {
```
glennw: Probably clearer to add a condition for 1 => above this and remove the _ => condition.

src/batch.rs
glennw: Remove print

src/batch.rs
```diff
@@ -453,3 +460,30 @@ impl RasterBatch {
         false
     }
 }
+
+trait InsideTest<T> {
```
glennw: Could you add a comment here describing exactly what it's checking for?

glennw:
I think this is still relevant but would need to be ported to the wr2 code - thoughts @splav  ?

splav:
@glennw, sure, I'll port it.

glennw:
Thanks!

splav:
@glennw not at all!

larsbergstrom:
Passed: https://travis-ci.org/servo/webrender/builds/150772527

PR #300, opened 2016-06-29
Fix scrolling on DDG HTML results page.

glennw:
There are a number of fixes / bandaids here:
 - get_scroll_layer() only returns a layer that is actually scrollable.
 - update the viewport transform when scrolling.
 - propagate the scrolling offset to scrollable child layers.


glennw:
@bors-servo retry

 * CI issues.

PR #302, opened 2016-06-29
 	Only send new scroll frames to the compositor if a scroll event actually resulted in motion, and fix some overscroll problems.

pcwalton:
Requires servo/webrender_traits#64.

r? @glennw


glennw:
@pcwalton This part looks fine once the other bits are reviewed.

PR #303, opened 2016-06-30
Fix twitter login crash.

glennw:
This allows texture allocation to try multiple times, growing or adding texture pages as required.

### DIFF
--- a/src/texture_cache.rs
+++ b/src/texture_cache.rs
@@ -799,100 +799,87 @@ impl TextureCache {
         let requested_size = Size2D::new(requested_width, requested_height);
         let allocation_size = Size2D::new(requested_width + border_size * 2,
                                           requested_height + border_size * 2);
-        let location = page_list.last_mut().and_then(|last_page| {
-            last_page.allocate(&allocation_size, filter)
-        });
-        let location = match location {
-            Some(location) => location,
-            None => {
-                if !page_list.is_empty() && page_list.last().unwrap().can_grow() {
-                    let last_page = page_list.last_mut().unwrap();
-                    // Grow the texture.
-                    let texture_size = cmp::min(last_page.texture_size * 2,
-                                                max_texture_size());
-                    self.pending_updates.push(TextureUpdate {
-                        id: last_page.texture_id,
-                        op: texture_grow_op(texture_size, format, mode),
-                    });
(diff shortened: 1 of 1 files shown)


src/texture_cache.rs
```diff
@@ -799,100 +799,86 @@ impl TextureCache {
...
+        assert!(allocation_size.width < max_texture_size());
+        assert!(allocation_size.height < max_texture_size());
+
+        loop {
+            let location = page_list.last_mut().and_then(|last_page| {
+                last_page.allocate(&allocation_size, filter)
+            });
+
+            match location {
+                Some(location) => {
```
pcwalton: nit: How about `if let Some(location) = location { ... }`? Since there's a `return` at the end, this should be equivalent, and then you can eliminate lots of rightward drift.

src/texture_cache.rs
```diff
@@ -799,100 +799,86 @@ impl TextureCache {
...
-                                                &mut free_texture_levels,
-                                                texture_size,
-                                                format,
-                                                mode);
+
+        // TODO(gw): Handle this sensibly (support failing to render items that can't fit?)
+        assert!(allocation_size.width < max_texture_size());
+        assert!(allocation_size.height < max_texture_size());
+
+        loop {
```
pcwalton: nit: I'd add a comment saying that you're looping to find a spot for the allocation, growing or adding pages as necessary.

src/texture_cache.rs
```diff
@@ -799,100 +799,86 @@ impl TextureCache {
...
+                            id: last_page.texture_id,
+                            op: texture_grow_op(texture_size, format, mode),
+                        });
+                        last_page.grow(texture_size);
+
+                        self.items.for_each_item(|item| {
+                            if item.texture_id == last_page.texture_id {
+                                item.texture_size = Size2D::new(texture_size, texture_size);
+                            }
+                        });
```
pcwalton: nit: you could add `continue` here and then eliminate even more rightward drift by removing the `else` clause :)

metajack:
Will this fix the box shadow thing on Windows too?

glennw:
@metajack You mean where it asserts about not getting a standalone alloc? Yes, it should fix that.

PR #308, opened 2016-07-05
modified config lines to specify platforms

alanamramjit:
the previous PR had a merge conflict in device.rs but this should work
r? @larsbergstrom

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -19,13 +19,13 @@ use std::mem;
 //use std::thread;
 use webrender_traits::ImageFormat;
 
-#[cfg(not(any(target_os = "android", target_os = "gonk")))]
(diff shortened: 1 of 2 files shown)


metajack:
Patch looks fine, but I'm curious why you only changed some of these? For example, in `src/renderer.rs` `enable_msaa` has the same cfg line but you didn't modify it. I found a few in the other file as well.

alanamramjit:
I'm sorry, I'm not seeing the lines you're talking about--can you give me a line no? I only changed lines that were targeting platforms that were defined as "neither gonk nor android"  and I don't see any of those left in the patches I submitted. I left the cfg lines that *were* targeting gonk and android alone because that's what Glenn had done originally

metajack:
@bors-servo r+

Sorry, I misread those lines.

PR #311, opened 2016-07-06
Remove gonk specific configuration lines

### LINKED ISSUE #310: remove gonk support
Now that we no longer have a gonk port, the gonk specific target configuration can be removed.

alanamramjit:
fixes #310


metajack:
Can you remove `any` now that there is only one option in those `cfg` lines? r=me with that change.

alanamramjit:
r=metajack

Issue #262, opened 2016-04-16
Display items with off-screen positions that are translated to be on-screen incorrectly culled out


pcwalton:
Test case:

    <!DOCTYPE html>
    <div style="transform: translateX(-2000px)">
        <div style="position: absolute; top: 0; left: 2000px;">X</div>
    </div>

This breaks http://apple.com/.

daoshengmu:
I can help investigate. I notice the below test case is wrong as well. 

```
<div style="transform: translateX(0px)">
    <div style="position: absolute; top: 0; left: 0px;">X</div>
</div>
```

daoshengmu:
I think I have fixed the first test case via this way at `frame::add_items_to_target()`:

```
let rect = info.world_transform.transform_rect(&item.rect
                  .translate(&info.offset_from_current_layer));
```

Because it forgets to append the world transform from it parent. I have checked http://apple.com/ that works well.

The test case of ```left: 0px;``` is another case that the letter `x` is clipped. Probably there is something wrong on the scissor test. I will follow up for this.

@pcwalton Do you think it is ok for me to send a PR?

tschneidereit:
@daoshengmu, if you already have a patch that fixes the issue for you, it's best to just open a PR. If something's wrong about it, then that can be discussed in the PR itself instead of here. This way, you don't need to wait for a response here and then again in the PR after you opened it.

glennw:
No longer an issue in WR2.

PR #313, opened 2016-07-14
The transform rect of draw list items needs to append parent's world transform.

daoshengmu:
Fix bug #262. Before adding item to aabb tree node, the item's rect needs to append parent's world transform. Otherwise, it would cull display items that are off-screen position. And I have confirmed this patch resolves http://apple.com case.

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -946,8 +946,8 @@ impl Frame {
                     let layer = self.layers.get_mut(&info.actual_scroll_layer_id).unwrap();
                     for (item_index, item) in draw_list.items.iter().enumerate() {
                         let item_index = DrawListItemIndex(item_index as u32);
-                        let rect = item.rect
-                                       .translate(&info.offset_from_current_layer);
+                        let rect = info.world_transform.transform_rect(&item.rect
+                                       .translate(&info.offset_from_current_layer));
                         layer.insert(rect,
                                      draw_list_group_id,
                                      draw_list_id,


glennw:
I don't think this is the correct fix. When originally implemented, the idea was that the AABB tree is stored in scroll-layer-local space. I suspect the actual problem is that the viewport matrix / rect used during culling check needs to be transformed into layer local space - does this make sense? Thanks for looking into it!

daoshengmu:
In this case, it only has one layer. Its world_origin is (0, 0), local_transform is identity, and world_transform is NONE. The aabb origin of this item is (-2000, 0), so it is culled.

I am not sure I understand it correctly. The viewport matrix has already transformed to layer's local space at Layer::Cull()

```
let adjusted_viewport = viewport_rect.translate(&-self.world_origin)
                                             .translate(&-self.scrolling.offset);
```

The problem is why the self.world_origin is (0,0). Even though we transform the viewport_rect to (-2000, 0). It will affect other items in this layer. Unless, we prepare to separate them into another layer.

Therefore, my approach is making its aabb has right transform to let it can be visible after culling.

daoshengmu:
In summary, this scene item is ```(ScrollPolicy::Scrollable, None)```. It will be added into an existing scroll layer. The scroll layer's world transform is None, the local transform is identity, and its world origin is (0, 0). Therefore, it would not give any help to transform the viewport to layer's local space. Besides, the viewport rect size is (1024x740), but the translate of this item is 2000, so it makes sense why this item is culled. If I let it is translated less than 1030, it would can be seen. 

Then, this item is going to call ```add_items_to_target()``` function, and it will be added to the scroll layer as an aabb node.

glennw:
@daoshengmu The scroll layer world transform should only be None initially - it should get set by update_layer_transforms() - could you check the value of the world transform after that function runs?

Looking at that code above that calculates adjusted_viewport, it only applies translations - it doesn't seem to apply an inverse matrix to move from world space into layer local space. I suspect that the bug is related to the calculation of the adjusted viewport.

daoshengmu:
@glennw I have checked the value of the world / local transform. They are all identity.

There are two layers, root layer and scroll layer. 

Their hierarchy looks like this:
root layer --- scroll layer
******************aabb node (scene_item at (11.55×15.983334 at (2000,1)))

Probably, this scene item has to be put to a new layer and transform the viewport to this new layer's local space...

glennw:
@daoshengmu I will do some testing with this now that WR2 has landed - but I suspect it's no longer an issue.

daoshengmu:
Cool~ That would be great!

glennw:
OK, I've confirmed that this is working correctly with WR2, so I'll close this. Thanks for looking into it!

PR #314, opened 2016-07-18
Fixes allocating textures that use nearest sampling.

glennw:
This code was accidentally removed during fixes for the tech preview demo.

Fixes an assert in the image_rendering_pixelated reftest.


glennw:
r? @pcwalton 

cc @Ms2ger

PR #317, opened 2016-07-31
Document the public api

MSleepyPanda:
While working with the API, i couldn't help but notice the lack of documentation. I talked with @glennw on IRC and he explained me a lot of details which i wrote down, so that new contributors will learn from my mistakes. I also tried to explain the relationship between the different parts of the API.

It's still a bit rough but it's something.
There's an equivalent PR for webrender_traits as well.

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -2,6 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//! The webrender API.
+//!
+//! The `webrender::renderer` module provides the interface to webrender, which
+//! is accessible through [`Renderer`][renderer]
+//!
+//! [renderer]: struct.Renderer.html
+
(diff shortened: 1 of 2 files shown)


src/renderer.rs
```diff
@@ -1877,4 +1915,3 @@ struct VertexBufferAndOffset {
     buffer: VertexBuffer,
     offset: u32,
 }
-
```
KiChjang: nit: Re-add this newline.

emilio:
Thanks for doing this! It's awesome! I have skimmed over it just a bit, and it looks good to me, but Glenn will have to sanity-check and sign-off this.

r? @glennw 

Thanks again for doing this! :)

MSleepyPanda:
Thanks for the encouraging words!

If you'd like coverage on other parts of the api, please let me know and i'll wrap my mind around it :) Personally, the pipeline part still bugs me a bit, but i'm getting closer

glennw:
@MSleepyPanda Fantastic, thanks for taking the time to do this!

PR #326, opened 2016-08-04
clean up a compile error in the border fragment shader

changm:
This caused the shaders to fail to compile on windows.


larsbergstrom:
Confirmed it worked in: https://travis-ci.org/servo/webrender/builds/149828478

Had to manually merge due to: https://github.com/servo/servo/issues/12739

PR #329, opened 2016-08-05
Supports border-style: double.

mephisto41:
Add support for border-style: double.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/329)

### DIFF
--- a/res/ps_border.fs.glsl
+++ b/res/ps_border.fs.glsl
@@ -56,6 +56,57 @@ vec4 draw_dotted_edge() {
   return mix(white, circleColor, circleColor.a);
 }
 
+vec4 draw_double_edge(float pos, float len) {
+  // Devided border to 3 parts, draw color on first and third part,
+  // leave second part blank.
+  float one_third_len = len / 3.0;
+
+  float in_first_part = step(pos, one_third_len);
+  float in_third_part = step(len - one_third_len, pos);
+
+  // The result of this should be 1.0 if we're in the 1st or 3rd part.
+  // And 0.0 for the blank part.
+  float should_fill = in_first_part + in_third_part;
+
+  float color_weight = step(0.0, vF);
+  vec4 color = mix(vHorizontalColor, vVerticalColor, color_weight);
+
+  vec4 white = vec4(1.0, 1.0, 1.0, 1.0);
+  return mix(white, color, should_fill);
+}
+
+vec4 draw_double_edge_vertical() {
+  // Get our position within this specific segment
+  float position = vLocalPos.x - vLocalBorders.x;
+  return draw_double_edge(position, vLocalBorders.z);
+}
+
+vec4 draw_double_edge_horizontal() {
+  // Get our position within this specific segment
+  float position = vLocalPos.y - vLocalBorders.y;
+  return draw_double_edge(position, vLocalBorders.w);
+}
+
+vec4 draw_double_edge_with_radius() {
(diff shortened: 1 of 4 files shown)


glennw:
@mephisto41 Thanks! Could you take a look over this please @changm ?

res/ps_border.fs.glsl
```diff
@@ -56,6 +56,59 @@ vec4 draw_dotted_edge() {
...
+  // leave second part blank.
+  float one_third_len = len / 3.0;
+
+  // The original algorithm should be check if pos in the blank part.
+  // That is, checking (1/3 * len) < pos < (2/3 * len)
+  //
+  // The idea here is, offset pos by (1/3 * len) then mod it by len.
+  // And if this value is lesser than (2/3 * len) then we're in
+  // blank part. Thus, we can use only 1 check instead of 2.
+  float compared_pos = mod(pos + one_third_len, len);
```
changm: How does this work? Let's say I have a length of 100 and I want to see of pos 50 is in the blank part.

So 0- 33 - Filled
33 - 66 - Blank
66 - 100 filled

(1/3) * len = 33 here

So offset 50 would be:
50 + 33 = 83
83 % 100 = 83

This doesn't match the comment, but it does match the code. In the code you're checking that compared_pos is greater than 2/3, which if true, means its in the blank part. Please fix the comment. I think should be "greater than (2/3 * len).

nit: I'd rather do the most obvious thing here and just do two checks.

Also, I should fix the other borders too, but please do the calculations in local space rather than device space. Thanks!

mephisto41: I change my implementation so it is easier to understand. And also calculations in local space. Would you check it again? Thanks very much.

res/ps_border.fs.glsl
```diff
@@ -56,6 +56,57 @@ vec4 draw_dotted_edge() {
...
+  float len = vRadii.x - vRadii.z;
+  return draw_double_edge(position, len);
+}
+
+vec4 draw_double_edge_corner() {
+  if (vRadii.x > 0) {
+    return draw_double_edge_with_radius();
+  }
+
+  bool is_vertical = (vBorderPart == PST_TOP_LEFT) ? vF < 0 : vF >= 0;
```
changm: How does this work for the bottom left corner? Seems like the vertical case shouldn't work here?

mephisto41: The vF calculation is quiet confuse now. For the vertical case, only the vF on top left corner is lesser than 0. All other 3 corners are greater than 0. I just tweak the vF calculation to make it more consistent. After my tweak, the vF stand for vertical edge if vF > 0 and for horizontal edge if vF < 0 for all corners, without exception. So this `is_vertical` check would be more clear.

src/tiling.rs
```diff
@@ -863,8 +863,8 @@ impl Primitive {
                                                        details.tl_inner.x,
                                                        details.tl_inner.y),
                     },
-                    vertical_color: details.top_color,
```
changm: This will break inset borders. Let's just revert the colors back and keep the top left as is. Sorry for reverting the change. The previous commit works fine. Thanks again!

mephisto41:
I have reverted to previous commit. Thanks!

changm:
LGTM! Thanks

@glennw care to take a look or merge if you think it's good. Thanks!

PR #331, opened 2016-08-08
Abstract the common transform related shader code into prim_shared.glsl.

glennw:
This is the first step in adding transform support for other primitive types,
as well as optimizing the performance of the 3d transform shader.


larsbergstrom:
Passed: https://travis-ci.org/servo/webrender/builds/150744613

PR #332, opened 2016-08-09
Update readme on how to use a custom webrender

changm:
Since this is documented in a non-obvious way in cargo.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/332)


larsbergstrom:
Passed: https://travis-ci.org/servo/webrender/builds/151010505

PR #333, opened 2016-08-09
Correct inner radii for the border

msreckovic:
Inner radius values in X are wrong - they are using the values from X direction instead.
- [ X] `./mach build -d` does not report any errors
- [X ] `./mach test-tidy` does not report any errors

- [X ] There are tests for these changes (the errors are masked until we implement elliptical corners)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/333)


glennw:
Thanks!

emilio:
Pass: https://travis-ci.org/servo/webrender/builds/151056296

PR #334, opened 2016-08-09
Edge-distance anti-aliasing for transformed rectangles

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/334)


mrobinson:
Going to try to retry this, just to ensure it isn't homu flakiness.

PR #335, opened 2016-08-10
Require euclid 0.8 and update code to support it

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/335)


mrobinson:
@nox, @larsbergstrom, or @glennw r?

PR #339, opened 2016-08-11
Switch 3d transform shaders to be features, rather than separate shaders.

glennw:
This doesn't change any existing functionality, but makes adding 3d transform
support for other primitives much simpler and more maintainable.

Also clean up the renderer code for each primitive type into a generic method.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/339)


src/renderer.rs
```diff
@@ -1223,6 +1235,36 @@ impl Renderer {
         }
     }
 
+    fn draw_ubo_batch<T>(&mut self,
+                         ubo_data: &Vec<T>,
```
emilio: nit: you can remove one level of indirection if you just use `&[T]`.

glennw:
@emilio Good point - fixed

PR #343, opened 2016-08-14
Implement a number of basic optimizations for WR2.

glennw:
* Retain frame builder during scrolling.
* Move primitive packing to per-primitive type, and only run on newly visible primitives.
* Add groundwork for tile fast paths (none are implemented yet).
* Reduce hash table lookups for glyphs.
* Cache packed primitives between scrolls.
* Add basic occlusion.
* Handle narrow phase collision test for tiles inside simple borders.

There is a lot of optimization work remaining, but these changes typically improve CPU/GPU time by around 25-50%.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/343)

### DIFF
--- a/src/tiling.rs
+++ b/src/tiling.rs
@@ -897,140 +897,135 @@ impl Primitive {
             (&mut PrimitiveBatchData::Composite(..), _) => return false,
             (&mut PrimitiveBatchData::Rectangles(ref mut data),
              &PrimitiveDetails::Rectangle(ref rectangle)) => {
-                match self.complex_clip {
-                    Some(..) => return false,
-                    None => {
-                        data.push(PackedRectanglePrimitive {
-                            common: PackedPrimitiveInfo {
-                                padding: 0,
-                                tile_index: tile_index_in_ubo,
-                                layer_index: layer_index_in_ubo,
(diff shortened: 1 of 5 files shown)


src/tiling.rs
```diff
@@ -745,79 +758,219 @@ impl Primitive {
...
-                        let texture_id = ctx.resource_cache.get_webgl_texture(&context_id);
-                        let uv = RectUv {
-                            top_left: Point2D::new(0.0, 1.0),
-                            top_right: Point2D::new(1.0, 1.0),
-                            bottom_left: Point2D::zero(),
-                            bottom_right: Point2D::new(1.0, 0.0),
+            (&mut PrimitiveBatchData::RectanglesClip(..), _) => return false,
+            (&mut PrimitiveBatchData::Image(ref mut data),
+             &PrimitiveDetails::Image(ref image)) => {
+                match self.complex_clip {
```
pcwalton: How about just `if self.complex_clip.is_some() { return false }`?

pcwalton:
This looks good, r=me with that nit fixed. Nice to see that this doesn't increase the code size much either!

pcwalton:
Looks good, squash if you'd like and r=me

PR #347, opened 2016-08-17
Use shared memory to pass around font data

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/347)


nox:
Blocked by https://github.com/servo/webrender_traits/pull/72.

nox:
The font cache needs an overhaul.

PR #348, opened 2016-08-17
Add edge-distance anti-aliasing support for rounded borders

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/348)


mrobinson:
I've updated the branch. Support for transformed borders is still missing as well as double, inset, and outset borders.

PR #349, opened 2016-08-17
run glyph raster jobs in parallel

alanamramjit:
preliminary tests show a ~31% improvement on glyph rasterization time of initial frames on wikipedia/nyt

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/349)

### DIFF
--- a/src/resource_cache.rs
+++ b/src/resource_cache.rs
@@ -11,6 +11,7 @@ use freelist::FreeList;
 use internal_types::{FontTemplate, GlyphKey, RasterItem};
 use internal_types::{TextureUpdateList, DrawListId, DrawList};
 use platform::font::{FontContext, RasterizedGlyph};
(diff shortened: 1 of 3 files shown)


alanamramjit:
I'll run some tests changing the size of the thread pool and post them here later

src/resource_cache.rs
```diff
@@ -400,9 +401,8 @@ fn run_raster_jobs(pending_raster_jobs: &mut Vec<GlyphRasterJob>,
     if pending_raster_jobs.is_empty() {
         return
     }
-
     // TODO(gw): Run raster jobs in parallel again
```
Manishearth: nit: this comment should be removed

alanamramjit: oops! fixed

PR #353, opened 2016-08-18
Add Api to access glyph dimensions

MSleepyPanda:
With this, standalone applications will be able to handle glyph spacing and/or positioning. Currently the workaround would be setting up a font renderer by yourself, but this would involve use of plattform dependent libraries and introduce overhead, since the texture cache already knows about the glyph dimensions.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/353)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -15,7 +15,7 @@ use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Sender;
 use texture_cache::TextureCache;
 use webrender_traits::{ApiMsg, AuxiliaryLists, BuiltDisplayList, IdNamespace};
-use webrender_traits::{PipelineId, RenderNotifier, WebGLContextId};
+use webrender_traits::{GlyphDimensions, PipelineId, RenderNotifier, WebGLContextId};
 use batch::new_id;
 use device::TextureId;
 use tiling::FrameBuilderConfig;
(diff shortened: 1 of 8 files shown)


webrender/src/render_backend.rs
```diff
@@ -92,6 +92,15 @@ impl RenderBackend {
                             self.resource_cache
                                 .add_font_template(id, FontTemplate::Native(native_font_handle));
                         }
+                        ApiMsg::GetGlyphDimensions(font_key, size, blur_radius, index, tx) => {
+                            let glyph_key = GlyphKey::new(font_key, size, blur_radius, index);
+                            // Get the next frame id, so it'll remain in the texture cache.
+                            // This assumes that the glyph will be used in the next frame.
+                            let frame_id = self.frame.next_frame_id();
+                            let glyph_dim = self.resource_cache.get_glyph_dimensions(glyph_key,
+                                                                                    frame_id);
```
glennw: nit: indentation

webrender/src/resource_cache.rs
```diff
@@ -349,6 +354,25 @@ impl ResourceCache {
         image_id.map(|image_id| self.texture_cache.get(image_id))
     }
 
+    pub fn get_glyph_dimensions(&mut self,
+                                glyph_key: GlyphKey,
+                                frame_id: FrameId)
+                                -> Option<GlyphDimensions> {
+        self.create_raster_job(&glyph_key, frame_id);
+        self.raster_pending_glyphs(frame_id);
+        if let Some(cached_glyph) = self.get_glyph(&glyph_key, frame_id) {
```
glennw: Let's use map() here instead of if let

webrender_traits/src/api.rs
```diff
@@ -67,6 +69,15 @@ impl RenderApi {
         key
     }
 
+    ///
+    pub fn get_glyph_dimensions(&self, font_key: FontKey, size: Au, blur_radius: Au, index: u32)
```
glennw: I wonder if this should be able to pass multiple glyphs to get in one batch, to reduce the overhead of an IPC message per glyph?

glennw:
@MSleepyPanda Nice! A couple minor issues, but generally looks good.

webrender_traits/src/api.rs
MSleepyPanda: Thought about that too, but since GlyphKey is internal, we would either have to create a vector of tuples or a vector with an equivalent struct. Your thoughts?

glennw: I think it's probably fine to expose GlyphKey (or similar) as a public type.

glennw:
@MSleepyPanda Thanks, could you squash those two commits and then this is ready to merge.

MSleepyPanda:
Done, thx for feedback!

PR #354, opened 2016-08-18
updated recording branch

alanamramjit:
updated recording infrastructure for #6

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/354)

### DIFF
--- a/webrender/src/record.rs
+++ b/webrender/src/record.rs
@@ -1,0 +1,32 @@
+use bincode::serde::serialize;
+use bincode;
+use std::fs::OpenOptions;
+use std::io::Write;
+use webrender_traits::ApiMsg;
+use byteorder::{LittleEndian, WriteBytesExt};
+
+pub fn write_data(frame:u32, data: &Vec<u8>){
+       let filename = format!("record/frame_{}.bin", frame);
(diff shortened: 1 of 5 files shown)


webrender/src/record.rs
```diff
@@ -0,0 +1,34 @@
...
+               ref msg @ &ApiMsg::AddRawFont(..) | 
+               ref msg @ &ApiMsg::AddNativeFont(..) |
+               ref msg @ &ApiMsg::AddImage(..) |
+               ref msg @ &ApiMsg::SetRootPipeline(..) |
+               ref msg @ &ApiMsg::UpdateImage(..) |
+               ref msg @ &ApiMsg::Scroll(..)|
+               ref msg @ &ApiMsg::TickScrollingBounce|
+               ref msg @ &ApiMsg::DeleteImage(..)|
+               ref msg @ &ApiMsg::SetRootStackingContext(..) =>{
+                       let filename = format!("record/frame_{}.bin", frame);
```
glennw: Can this share code with or directly use the write_data function above? (Just so the filename isn't declared in two locations etc).

glennw:
@alanamramjit One minor issue, then it should be good to go.

alanamramjit:
You're right; this should probably create a RecordMsg enum so that it's easier to deserialize anyway.

PR #355, opened 2016-08-19
Fix a number of bugs and issues with scroll roots and transforms.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/355)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -7,16 +7,15 @@ use euclid::{Matrix4D, Point2D, Point3D, Point4D, Rect, Size2D};
 use fnv::FnvHasher;
 use geometry::ray_intersects_rect;
 use internal_types::{ANGLE_FLOAT_TO_FIXED, AxisDirection};
-use internal_types::{CompositionOp};
-use internal_types::{LowLevelFilterOp, max_rect};
+use internal_types::{CompositionOp, DevicePixel};
+use internal_types::{LowLevelFilterOp};
 use internal_types::{RendererFrame};
 use layer::{Layer, ScrollingState};
 use resource_cache::ResourceCache;
 use scene::{SceneStackingContext, ScenePipeline, Scene, SceneItem, SpecificSceneItem};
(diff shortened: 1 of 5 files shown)


glennw:
r? @pcwalton 

You have already reviewed the first commit in this patch set. I'm including it here (and closing the PR for that patch) because bors is not working at the moment and won't merge that PR. So only the scroll roots commit in this PR needs to be reviewed.

Note that this also requires a small patch in servo (https://github.com/servo/servo/pull/12932), so WR shouldn't be updated in Servo without that patch.

webrender/src/frame.rs
```diff
@@ -673,203 +467,313 @@ impl Frame {
...
+                                          .get(&pipeline_id)
+                                          .expect("No auxiliary lists?!");
+                stacking_context.composition_operations(auxiliary_lists)
+            };
+
+            // Detect composition operations that will make us invisible.
+            for composition_operation in &composition_operations {
+                match *composition_operation {
+                    CompositionOp::Filter(LowLevelFilterOp::Opacity(Au(0))) => return,
+                    _ => {}
```
pcwalton: nit: Could be `if let` I guess

webrender/src/frame.rs
```diff
@@ -673,203 +467,313 @@ impl Frame {
...
-            match (stacking_context.scroll_policy, stacking_context.scroll_layer_id) {
-                (ScrollPolicy::Scrollable, Some(ScrollLayerId {
-                    info: ScrollLayerInfo::Scrollable(index),
-                    ..
-                })) if index > 0 => {
-                    Some(stacking_context.transform
-                                         .invert()
-                                         .transform_rect(&stacking_context.overflow))
+        let scene_items = scene_item.collect_scene_items(&context.scene);
+        if !scene_items.is_empty() {
```
pcwalton: Can this be `if scene_items.is_empty() { return }`?

pcwalton:
Looks good, r=me

PR #361, opened 2016-08-23
Switch the layer viewport rect to local space.

glennw:
This allows iframes and scroll roots to correctly clip to the viewport when they have
a 3d transform.

This still doesn't solve the nested transform within rotated
transform case, but it's a big improvement on the current situation.

Also add transform feature to various primitive shaders.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/361)


glennw:
This contains the required fixes for https://github.com/servo/servo/issues/12286

PR #362, opened 2016-08-23
Add edge-distance anti-aliasing for inset and outset borders

mrobinson:
This change also fixes a rendering problem where the corner colors for
inset and outset borders was reversed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/362)


mrobinson:
@glennw r?

PR #364, opened 2016-08-23
Split text primitives up into runs of 8 glyphs each.

pcwalton:
Statistics on http://en.wikipedia.org/wiki/Servomechanism (MBP, Haswell, Intel Iris) before and after:

* Vertices: 123,150 → 53,802 (2.28x fewer)
* CPU backend time: 2.74ms → 2.48ms (10% improvement)
* CPU compositor time: 1.19ms → 0.88ms (35% improvement)
* GPU time: 2.53ms → 1.60ms (58% improvement)

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/364)

### DIFF
--- a/webrender/res/ps_text_run.fs.glsl
+++ b/webrender/res/ps_text_run.fs.glsl
@@ -1,0 +1,8 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+void main(void) {
+    float a = texture(sDiffuse, vUv).a;
+    oFragColor = vec4(vColor.rgb, vColor.a * a);
+}
--- a/webrender/res/ps_text_run.glsl
+++ b/webrender/res/ps_text_run.glsl
@@ -1,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+flat varying vec4 vColor;
+varying vec2 vUv;
+
+#ifdef WR_FEATURE_TRANSFORM
+varying vec3 vLocalPos;
+flat varying vec4 vLocalRect;
+#endif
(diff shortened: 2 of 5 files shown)


glennw:
Reviewed 5 of 5 files at r1.

---

*[res/ps_text_run.vs.glsl, line 22 \[r1\]](https://reviewable.io:443/reviews/servo/webrender/364#-KPtYlGsHAivLLN04g33:-KPtYlGsHAivLLN04g34:b-h2zdxo) ([raw file](https://github.com/servo/webrender/blob/ebf3b38ffc65d05c5874913bff54e076bc9caac5/res/ps_text_run.vs.glsl#L22)):*
> ```GLSL
> 
> void main(void) {
>     TextRun text_run = text_runs[gl_InstanceID / WR_GLYPHS_PER_TEXT_RUN];
> ```

We should also add the transform feature support (should be easy enough to port from ps_text shaders).

---

*[webrender/src/tiling.rs, line 1561 \[r1\]](https://reviewable.io:443/reviews/servo/webrender/364#-KPtYYwJZ2ZwhS4exSno:-KPtYYwKOjvTaauBYEOZ:b-usrf42) ([raw file](https://github.com/servo/webrender/blob/ebf3b38ffc65d05c5874913bff54e076bc9caac5/webrender/src/tiling.rs#L1561)):*
> ```Rust
>              &PrimitiveDetails::Text(ref text)) => {
>                 let cache = text.cache.as_ref();
>                 // FIXME(pcwalton): Why is this not always true??
> ```

Can we find out why this is happening? It seems like a bug for this to occur?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/364)*

glennw:
@pcwalton The shaders are also in the wrong location, they need to be moved from res/ to webrender/res/ now that the merge with wr_traits has happened.

jrmuizel:
How is this approach different from what was happening before?

glennw:
Previously an entire text run was being submitted to every tile that it touches. This is quite wasteful since it means a lot of glyphs end up getting submitted to tiles that they don't hit, and are culled in the vertex shader. This splits text runs up into smaller blocks so that they typically hit fewer tiles.

pcwalton:
Comments addressed. 

I added a comment to specify when the `cache` of text primitive items is `None`: it's when the resource cache didn't successfully rasterize the glyph (probably because the font was missing that glyph).

r? @glennw

PR #365, opened 2016-08-24
Properly mix border colors to avoid jagged edges

mrobinson:
When border colors meet, properly mix them instead of selecting one or
the other. This creates smooth boundaries between border colors.
Currently transformed borders are not handled.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/365)


mrobinson:
before
after

Before and after images are above. I think this greatly improves where border colors meet.

PR #366, opened 2016-08-24
Edge-distance AA for transformed border corners

mrobinson:
Extend edge-distance anti-aliasing to transformed border corner,
including border corner edges and color mixing areas.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/366)


mrobinson:
@glennw r?

glennw:
Nice!

PR #370, opened 2016-08-25
Add edge-distance anti-aliasing for double borders

mrobinson:
This includes anti-aliasing for rounded corners as well as for the
interior border edges which may fall on subpixel boundaries.
Additionally, this fixes a bug where the central gap was rendered with
white instead of transparent.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/370)

### DIFF
--- a/webrender/res/ps_border.fs.glsl
+++ b/webrender/res/ps_border.fs.glsl
@@ -19,6 +19,39 @@ vec4 get_fragment_color(float distanceFromMixLine, float pixelsPerFragment) {
   return mix(vHorizontalColor, vVerticalColor, colorMix);
 }
 
+float alpha_for_solid_border(float distance_from_ref,
+                             float inner_radius,
+                             float outer_radius,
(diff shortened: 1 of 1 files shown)


webrender/res/ps_border.fs.glsl
```diff
@@ -13,12 +13,45 @@ void discard_pixels_in_rounded_borders(vec2 local_pos) {
 
 vec4 get_fragment_color(float distanceFromMixLine, float pixelsPerFragment) {
   // Here we are mixing between the two border colors. We need to convert
-  // distanceFromMixLine it to pixel space to properly anti-alias and then push
+  // distanceFromMixLine it to pixel space to properly anti-alias and them push
```
samlh: Intended?

mrobinson: Oops. Not at all. I will push a new version of the branch.

Issue #372, opened 2016-08-26
Batching changes have broken some reftests


glennw:
opacity_simple_a.html doesn't render correctly with the batching changes.

There are some other failures, but I'm not sure if they are related to the batching changes or a different PR.

glennw:
cc @pcwalton 

We'll need to get this fixed, so we can update WR into servo.

glennw:
Ah, found the bug - fix incoming shortly.

glennw:
This got fixed.

PR #373, opened 2016-08-26
Fix reftest failure in opacity_simple test.

glennw:
Blend/composite items were being added to the first batch twice,
resulting in incorrect blending in the first tile of a new batch.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/373)


glennw:
This fixes #372

PR #376, opened 2016-08-28
Bump wr_traits version, to pick the last PR.

emilio:
I don't know if right now we're supposed to use it, though. Isn't it included
via path by WR?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/376)

### DIFF
--- a/webrender_traits/Cargo.toml
+++ b/webrender_traits/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "webrender_traits"
-version = "0.5.1"
+version = "0.5.2"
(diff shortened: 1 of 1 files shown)


nox:
Yes we can just `./mach cargo-update` it. If you do a bump, please bump both crates though.

nox:
Last PR was a breaking change though. Adding new variants to a public enum is a breaking change.

glennw:
@nox @emilio  Is this PR needed / wanted?

nox:
As long as we don't publish it, not really.

emilio:
Nope, it's not.

PR #381, opened 2016-08-30
Switch primitives to make use of the caching available in prepare_render.

glennw:
This is a slight performance win, but is mostly to make it easier to switch UBOs to SoA style. When this is done, these changes allow a much greater performance improvement (since primitives that hit multiple tiles are added only once).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/381)

### DIFF
--- a/webrender/res/ps_image.vs.glsl
+++ b/webrender/res/ps_image.vs.glsl
@@ -5,8 +5,8 @@
 
 struct Image {
     PrimitiveInfo info;
-    vec4 st_rect;       // Location of the image texture in the texture atlas.
-    vec4 stretch_size;  // Size of the actual image.
+    vec4 st_rect;               // Location of the image texture in the texture atlas.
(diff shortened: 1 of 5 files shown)


webrender/res/ps_image.vs.glsl
```diff
@@ -20,13 +20,27 @@ void main(void) {
...
+    vec2 st0 = image.st_rect.xy;
+    vec2 st1 = image.st_rect.zw;
+
+    switch (uint(image.stretch_size_uvkind.z)) {
+        case UV_NORMALIZED:
+            break;
+        case UV_PIXEL: {
+                vec2 texture_size = textureSize(sDiffuse, 0);
+                st0 /= texture_size;
+                st1 /= texture_size;
```
pcwalton: Can this divide by zero?

glennw: I don't believe so, the sampler is set to the texture cache id, so it should always be a valid size.

PR #382, opened 2016-08-30
Early out for text runs with size <= 0.

glennw:
This fixes a few dozen css tests that provide negative font size.


emilio:
I don't think if negative font sizes are allowed to get out of the style system. Let me double check and provide a proper fix if that's the case. Otherwise r=me

emilio:
> gw: I think it should be fixed elsewhere, but it also makes sense to check it there for non-servo users of WR :)

Fair enough :P 

@bors-servo: r+

PR #384, opened 2016-08-31
Add anti-aliasing for dashed borders

mrobinson:
Add anti-aliasing for dashed border segments that are not corners. This
is important for transformed dashed borders.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/384)


mrobinson:
@glennw r?

PR #385, opened 2016-09-01
Switch the UBO data to be untyped.

glennw:
This is an interim workaround for how slow shader compilation is on
Linux when using AoS style declarations. We need this change in order
to enable CI testing of WR, otherwise the test pass is too slow.

It brings the shader compile time on Linux down from around 1.8s to
around 0.2 seconds.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/385)


pcwalton:
@bors-servo: r+

Yuck but OK.

PR #387, opened 2016-09-02
Add anti-aliasing support for dotted borders

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/387)


mrobinson:
@glennw r?

PR #388, opened 2016-09-05
Store the last rendered epoch for all pipelines.

glennw:
Instead of storing the last rendered epoch for each pipeline from
the previous frame, accumulate this information for each frame.
This is a workaround for the way servo's compositor keeps iframes
in the frame tree when they are invisible.

Long term, we should find a better solution for this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/388)


mrobinson:
@glennw This looks good to me, so r=mrobinson unless you explicitly want @pcwalton to look at this.

pcwalton:
Could you add a link to the Servo issue number in the comment?

glennw:
@pcwalton Rebased and added comment.

PR #390, opened 2016-09-06
Switch layers to use VTF instead of UBO.

glennw:
Using vertex texture fetch simplifies the CPU-side batching code,
and is also a performance win.

Also clean up the texture binding APIs / code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/390)


pcwalton:
Interesting. I'm surprised this was a win, but I believe you!

@bors-servo: r+

PR #393, opened 2016-09-07
Mac drivers seem unable to use texel fetch offsets >8.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/393)


pcwalton:
r=me with a comment explaining why this is done

PR #394, opened 2016-09-07
Simplify border shader code

mrobinson:
Eliminate many of the #ifdef'd sections for the transformation shader,
which are now duplicated. Additionally, combine the functions that
draw dashed and dotted borders, since the corners for these border types
are drawn in the same way.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/394)


mrobinson:
@glennw r?

PR #395, opened 2016-09-07
Add anti-aliasing support for rounded rects

mrobinson:
Do edge distance anti-aliasing on rectangles (solid color, images, and
gradients) with clips that are rounded rects.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/395)


mrobinson:
@glennw r?

PR #396, opened 2016-09-08
Change packed tiles to use vertex textures, and extend to generic render task data.

glennw:
This is a performance win (by removing all the CPU code that handles batching
tiles) and also implements the framework for adding different types of
render tasks. For example, a prim cache or blur render task may provide
some different data to the shaders than the alpha batcher render task.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/396)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -35,8 +35,8 @@ use std::sync::{Arc, Mutex};
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::thread;
 use texture_cache::{BorderType, TextureCache, TextureInsertOp};
-use tiling::{self, Frame, FrameBuilderConfig, GLYPHS_PER_TEXT_RUN, PrimitiveBatchData, PackedTile};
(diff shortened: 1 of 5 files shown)


webrender/src/renderer.rs
```diff
@@ -35,19 +35,65 @@ use std::sync::{Arc, Mutex};
...
+            id: id,
+        }
+    }
+
+    fn init<T: Default>(&mut self,
+                        device: &mut Device,
+                        data: &mut Vec<T>) {
+        let item_size = mem::size_of::<T>();
+        debug_assert!(item_size % 16 == 0);
+        let vecs_per_item = item_size / 16;
```
pcwalton: What's the significance of 16 here?

glennw: In a RGBAF32 texture, 16 bytes is one texture fetch / texel. All our current data matches this requirement (which was the 16 byte UBO alignment requirement).

pcwalton: Ah, yes. That makes sense!

PR #397, opened 2016-09-08
Add basic overlay scrollbar support.

glennw:
This adds the infrastructure for adding scrollbars to any scroll root,
however for now it only draws a scrollbar for the root frame. Once we
are happy with the implementation and design, we can easily enable it
for iframes and other scroll roots.

The idea is to insert a primitive of any kind into the primitive tree
and tag it as being a scroll bar primitive for a given layer. This
means we can allow styling of the scrollbar in the future, making use
of any of the display item features that are supported (e.g. border
corners, gradients, images, anti-aliasing).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/397)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -15,7 +15,7 @@ use resource_cache::ResourceCache;
 use scene::{SceneStackingContext, ScenePipeline, Scene, SceneItem, SpecificSceneItem};
 use std::collections::{HashMap, HashSet};
 use std::hash::BuildHasherDefault;
-use tiling::{Clip, FrameBuilder, FrameBuilderConfig, InsideTest};
+use tiling::{Clip, FrameBuilder, FrameBuilderConfig, InsideTest, PrimitiveFlags};
 use util::MatrixHelpers;
 use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLayerId};
 use webrender_traits::{ColorF, StackingContext, FilterOp, MixBlendMode};
@@ -29,6 +29,8 @@ const CAN_OVERSCROLL: bool = false;
 
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
 pub struct FrameId(pub u32);
+
+static DEFAULT_SCROLLBAR_COLOR: ColorF = ColorF { r: 0.3, g: 0.3, b: 0.3, a: 0.6 };
 
 struct FlattenContext<'a> {
     resource_cache: &'a mut ResourceCache,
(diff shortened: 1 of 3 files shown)


glennw:
scrollbars

webrender/src/frame.rs
```diff
@@ -724,6 +726,17 @@ impl Frame {
             }
         }
 
+        if level == 0 && self.frame_builder_config.enable_scrollbars {
+            let scrollbar_rect = Rect::new(Point2D::zero(),
+                                           Size2D::new(10.0, 70.0));
+            context.builder.add_solid_rectangle(&scrollbar_rect,
+                                                &scrollbar_rect,
+                                                None,
+                                                &ColorF::new(0.3, 0.3, 0.3, 0.6),
```
pcwalton: Factor this out into a static.

webrender/src/tiling.rs
```diff
@@ -3206,6 +3276,50 @@ impl FrameBuilder {
...
+        let distance_from_edge = 8.0;
+
+        for scrollbar_prim in &self.scrollbar_prims {
+            let prim = &mut self.prim_store[scrollbar_prim.prim_index.0];
+            let scroll_layer = &layer_map[&scrollbar_prim.scroll_layer_id];
+
+            let scrollable_distance = scroll_layer.content_size.height - scroll_layer.local_viewport_rect.size.height;
+
+            if scrollable_distance <= 0.0 {
+                prim.local_clip_rect.size = Size2D::zero();
```
pcwalton: suggest `continue` here to avoid rightward drift

webrender/src/tiling.rs
```diff
@@ -3206,6 +3276,50 @@ impl FrameBuilder {
...
+                            scroll_layer.scrolling.offset.y -
+                            prim.rect.size.height -
+                            distance_from_edge;
+
+                prim.rect.origin.x = scroll_layer.local_viewport_rect.origin.x +
+                                     scroll_layer.local_viewport_rect.size.width -
+                                     prim.rect.size.width -
+                                     distance_from_edge;
+
+                prim.rect.origin.y = min_y + f * (max_y - min_y);
```
pcwalton: Isn't there a `lerp` function for this?

webrender/src/tiling.rs
```diff
@@ -3206,6 +3276,50 @@ impl FrameBuilder {
...
+                prim.local_clip_rect.origin = prim.rect.origin;
+                prim.local_clip_rect.size = prim.rect.size;
```
pcwalton: Could just be `prim.local_clip_rect = prim.rect;`

PR #399, opened 2016-09-09
Transformation support for rounded rectangles

mrobinson:
Add transformation support for rounded solid color and image
rectangles. This already exists for gradients.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/399)


mrobinson:
@glennw r?

PR #400, opened 2016-09-12
Do not repeat textures in transformed image rects

mrobinson:
The way that texture coordinates were calculated meant that textures
were repeated outside the bounds of transformed image primitives. Clamp
to local rect boundaries in order to simulate texture clamping here.
This is important for anti-aliased edges, which have subpixels
outside texture boundaries.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/400)


mrobinson:
@glennw r?

glennw:
Nice fix!

PR #408, opened 2016-09-15
Add ARM support

mmatyas:
This PR adds ARM support to Webrender:

- Fixes int-to-float conversions, causing shader compile errors
- Instead of using OS name/`android` to determine OpenGL/GLES version, uses the architecture; use GLES if the target arch is ARM or AArch64

A screenshot of Servo running on Odroid XU3:

servo-wr2-arm

(running as `./servo -G es2 -w -Z wr-stats https://en.wikipedia.org/wiki/Rust`)

As you can see, there are still some bugs, eg. missing hover effects or font drawing problems of `wr-stats`, but the rendering itself seems ok, and the performance is really good.
Also I'm not 100% sure about the GL parameters like `GL_FORMAT_A` and the `GpuProfile`, I'd be happy if someone could review them.

CC @glennw, @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/408)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -19,10 +19,10 @@ use std::mem;
 //use std::thread;
 use webrender_traits::ImageFormat;
 
-#[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
+#[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
 const GL_FORMAT_A: gl::GLuint = gl::RED;
 
-#[cfg(target_os = "android")]
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
 const GL_FORMAT_A: gl::GLuint = gl::ALPHA;
 
 #[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
@@ -31,16 +31,10 @@ const GL_FORMAT_BGRA: gl::GLuint = gl::BGRA;
 #[cfg(target_os = "android")]
 const GL_FORMAT_BGRA: gl::GLuint = gl::BGRA_EXT;
 
-#[cfg(target_os = "linux")]
+#[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
 const SHADER_VERSION: &'static str = "#version 150\n";
 
-#[cfg(target_os = "macos")]
-const SHADER_VERSION: &'static str = "#version 150\n";
-
-#[cfg(target_os = "windows")]
-const SHADER_VERSION: &'static str = "#version 150\n";
-
-#[cfg(target_os = "android")]
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
 const SHADER_VERSION: &'static str = "#version 300 es\n";
 
 static SHADER_PREAMBLE: &'static str = "shared.glsl";
@@ -578,19 +572,19 @@ pub struct VBOId(gl::GLuint);
 #[derive(PartialEq, Eq, Hash, Debug, Copy, Clone)]
 struct IBOId(gl::GLuint);
 
-#[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
+#[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
 pub struct GpuProfile {
     next_query: usize,
     qids: Vec<gl::GLuint>,
 }
 
-#[cfg(target_os = "android")]
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
 pub struct GpuProfile;
 
 const QUERY_COUNT: i32 = 4;
 
 impl GpuProfile {
-    #[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
+    #[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
     pub fn new() -> GpuProfile {
         let queries = gl::gen_queries(QUERY_COUNT);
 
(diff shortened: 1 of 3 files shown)


mmatyas:
To enable using it in Servo, the OpenGL version handling will also need an update (in `ports/glutin/window.rs`), I'll send a follow-up PR if this patch lands.

webrender/src/device.rs
```diff
@@ -19,28 +19,22 @@ use std::mem;
 //use std::thread;
 use webrender_traits::ImageFormat;
 
-#[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
```
glennw: These seem to be inverted from what they were previously?

webrender/src/device.rs
```diff
@@ -19,28 +19,22 @@ use std::mem;
...
 const GL_FORMAT_A: gl::GLuint = gl::RED;
 
-#[cfg(target_os = "android")]
+#[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
 const GL_FORMAT_A: gl::GLuint = gl::ALPHA;
 
-#[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
```
glennw: These also seem to be inverted?

mmatyas:
Thanks, I've fixed and rebased the patch. I've left the `GL_FORMAT_BGRA` lines as they were before, because changing that caused a surprising number of build errors in Servo, requiring changes in other dependencies (skia, gleam, ...), and even then it caused a shader error.
However, I couldn't test the PR with the rebase, because #410 causes a `gl function was not loaded` error for me...

glennw:
Interesting - nothing in that PR calls a new function, so I'm not sure why it would cause that error. Could you get a backtrace of that error, which should identify which function isn't being loaded?

mmatyas:
Sorry for the late reply, here's the backtrace:

```
$ RUST_BACKTRACE=1 DISPLAY=:0 ./servo -G es2 https://en.wikipedia.org/wiki/Rust
gl function was not loaded (thread main, at /home/mmatyas/Servo/servo-arm/servo/.cargo/registry/src/github.com-1ecc6299db9ec823/gleam-0.2.22/src/lib.rs:16)
stack backtrace:
   0: 0xb2c128e3 - backtrace::backtrace::libunwind::trace
                at /home/mmatyas/Servo/servo-arm/servo/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.2.3/src/backtrace/mod.rs:82
                 - backtrace::backtrace::trace<closure>
                at /home/mmatyas/Servo/servo-arm/servo/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.2.3/src/backtrace/mod.rs:70
   1: 0xb2c136ab - backtrace::capture::{{impl}}::new
                at /home/mmatyas/Servo/servo-arm/servo/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.2.3/src/lib.rs:96
   2: 0xb1f1a7b7 - servo::main::{{closure}}
                at /home/mmatyas/Servo/servo-arm/servo/components/servo/main.rs:121
   3: 0xb632badb - std::panicking::rust_panic_with_hook::hbcdd76ca45e6ae3d
   4: 0xb5c8663f - std::panicking::begin_panic<&str>
                at /buildslave/rust-buildbot/slave/nightly-dist-rustc-cross-host-linux/build/src/libstd/panicking.rs:383
   5: 0xb5c8b63f - gleam::ffi::missing_fn_panic
                at /home/mmatyas/Servo/servo-arm/servo/.cargo/registry/src/github.com-1ecc6299db9ec823/gleam-0.2.22/src/lib.rs:16
   6: 0xb1f40aab - gleam::ffi::ClearColor
                at /home/mmatyas/Servo/servo-arm/servo/target/arm-unknown-linux-gnueabihf/debug/build/gleam-3321b7699971b81b/out/gl_bindings.rs:1086
   7: 0xb1f409f7 - gleam::gl::clear_color
                at /home/mmatyas/Servo/servo-arm/servo/.cargo/registry/src/github.com-1ecc6299db9ec823/gleam-0.2.22/src/gl.rs:1298
   8: 0xb1f4dddf - glutin_app::window::{{impl}}::new
                at /home/mmatyas/Servo/servo-arm/servo/ports/glutin/window.rs:191
   9: 0xb1f546e3 - glutin_app::create_window
                at /home/mmatyas/Servo/servo-arm/servo/ports/glutin/lib.rs:48
  10: 0xb1f11007 - servo::main
                at /home/mmatyas/Servo/servo-arm/servo/components/servo/main.rs:138
  11: 0xb6331f5b - __rust_maybe_catch_panic
  12: 0xb632a94b - std::rt::lang_start::h6d20aa8156b6f2f5
  13: 0xb1f1ab17 - main
  14: 0xb16bf771 - __libc_start_main
```

It seems `ClearColor` fails?

The other dependencies (eg. gleam) were unchanged, maybe this needs some tweaking in them too.

glennw:
It looks like you are missing the -w (enable webrender) flag in the last comment?

mmatyas:
Ah yes, you're right, but unfortunately Servo still produces the same error. I've also noticed that I forgot to update two `"target_os="android"` lines when I did the rebase, so I've just fixed that, but it didn't help either.

glennw:
The missing function is glClearColor - this is the first GL function that is ever called, so it probably signals that the GL functions are all failing to load for some reason. Sometimes this also happens when the GL library loader gets linked twice (e.g. if there are two different versions of gleam present in the cargo.lock file).

Could you check what version(s) of gleam are present in the cargo.lock file?

mmatyas:
In `./components/servo/Cargo.lock` I only see references to gleam 0.2.22 (`"checksum gleam 0.2.22 (registry+https://github.com/rust-lang/crates.io-index)" = "e299fc6b34e698955c6e7cbe7afef2b89774124c82d3636ab85c540edd9ad567"`).

Here's the linking step that rustc calls: http://pastebin.com/97fxDELF; I've tried things like adding `-lGLESv2` manually, or leave only one occurrence of GL, EGL and X11, but that didn't help, I get the same ClearColor error.

mmatyas:
Looks like the problem was on my side: I made a completely clean build, and it seems to be working now; sorry for the inconvenience. I'll do some more testing and benchmarking next week.

glennw:
Thanks!

PR #409, opened 2016-09-15
Change shaders to compile lazily.

glennw:
This makes a huge improvement to startup time by compiling shaders
the first time they are used instead of up front.

This improves iteration time during development, but most importantly
it drastically improves the CI time when using webrender, since many
tests only need to compile a small subset of the shaders that are
available.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/409)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -94,10 +94,68 @@ impl VertexDataTexture {
     }
 }
 
-#[derive(Copy, Clone)]
+const TRANSFORM_FEATURE: &'static [&'static str] = &["TRANSFORM"];
+
+enum ShaderKind {
+    Primitive,
+    Clear,
+}
+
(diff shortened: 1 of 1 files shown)


webrender/src/renderer.rs
```diff
@@ -94,10 +94,70 @@ impl VertexDataTexture {
...
+                    }
+                    ShaderKind::Primitive => {
+                        create_prim_shader(self.name,
+                                           device,
+                                           self.max_ubo_vectors,
+                                           self.features)
+                    }
+                };
+                self.id = Some(id);
+                id
```
pcwalton: How about putting this at the top of the function as:

    if self.id.is_none() {
        ...
    }

And then just return `self.id.unwrap()` out of this function.

glennw: Updated

PR #410, opened 2016-09-15
Don't deinit textures each frame.

pcwalton:
The calls to `glTexImage2D()` cause driver stalls on Mac.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/410)


pcwalton:
Also, this deinitializes textures using the same format the texture was created in

PR #414, opened 2016-09-19
Implementation of background-repeat space

mrobinson:
Add a new parameter to image shaders which specifies the spacing
between tiles, when an image is repeated.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/414)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -375,38 +375,44 @@ PrimitiveInfo fetch_text_run_glyph(int index, out vec4 color, out vec4 uv_rect) 
 
 struct Image {
     PrimitiveInfo info;
-    vec4 st_rect;               // Location of the image texture in the texture atlas.
-    vec4 stretch_size_uvkind;   // Size of the actual image.
+    vec4 st_rect;                        // Location of the image texture in the texture atlas.
+    vec4 stretch_size_and_tile_spacing;  // Size of the actual image and amount of space between
+                                         //     tiled instances of this image.
+    vec4 uvkind;                         // Type of texture coordinates.
 };
 
 Image fetch_image(int index) {
     Image image;
 
-    int offset = index * 5;
+    int offset = index * 6;
 
     image.info = unpack_prim_info(offset);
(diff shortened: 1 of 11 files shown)


webrender/res/prim_shared.glsl
```diff
@@ -376,37 +376,45 @@ PrimitiveInfo fetch_text_run_glyph(int index, out vec4 color, out vec4 uv_rect)
 struct Image {
     PrimitiveInfo info;
     vec4 st_rect;               // Location of the image texture in the texture atlas.
-    vec4 stretch_size_uvkind;   // Size of the actual image.
+    vec2 stretch_size;          // Size of the actual image.
+    vec2 tile_spacing;          // Amount of space between tiled instances of this image.
+    float uvkind;
```
glennw: This will need to be 2 vec4s. It's the only reliable way I've found of making all the shader layouts work correctly between linux and mac implementations.

webrender/res/prim_shared.glsl
```diff
@@ -376,37 +376,45 @@ PrimitiveInfo fetch_text_run_glyph(int index, out vec4 color, out vec4 uv_rect)
...
     return image;
 }
 
 struct ImageClip {
     PrimitiveInfo info;
     vec4 st_rect;               // Location of the image texture in the texture atlas.
-    vec4 stretch_size_uvkind;   // Size of the actual image.
+    vec2 stretch_size;          // Size of the actual image.
+    vec2 tile_spacing;          // Amount of space between tiled instances of this image.
+    float uvkind;
```
glennw: Same vec4 layout issue.

webrender/res/prim_shared.glsl
mrobinson: Thanks for pointing this out. I think I have eliminated the issue in the latest version of the branch.

webrender/res/prim_shared.glsl
mrobinson: As well, as this one.

glennw:
Thanks!

PR #416, opened 2016-09-20
Cleanups and optimizations for composite ops, tile cmd lists.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/416)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -1952,7 +1952,7 @@ struct PackedCompositeInfo {
 }
 
 impl PackedCompositeInfo {
-    fn new(ops: &Vec<CompositionOp>) -> PackedCompositeInfo {
+    fn new(ops: &[CompositionOp]) -> PackedCompositeInfo {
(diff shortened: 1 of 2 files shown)


glennw:
r? @pcwalton 

These are a small performance optimization, but the main point is to remove the simple cmd list code, so there is only one path for adding primitives (this will be useful once I start making use of prim caching).

webrender/src/tiling.rs
```diff
@@ -2681,7 +2625,7 @@ impl FrameBuilder {
                       transform: Matrix4D<f32>,
                       pipeline_id: PipelineId,
                       scroll_layer_id: ScrollLayerId,
-                      composition_operations: Vec<CompositionOp>) {
+                      composition_operations: &Vec<CompositionOp>) {
```
pcwalton: nit: could just be `composition_operations: &[CompositionOp]`

pcwalton:
r=me with that nit if you like

PR #417, opened 2016-09-22
Add a number of changes to support running tests on OSMesa.

glennw:
* Add a wrapper type for webgl contexts that allows selectinng
  between a native (hw accel) or osmesa (software) context at
  runtime.

* Change request_webgl_context behaviour to try creating a readback
  (non-shared) context if shared context creation fails. This means
  that the webgl code in servo doesn't have to be aware of whether
  it should be creating native or osmesa contexts in fallback mode.

* Flush each webgl context at the start of the frame to ensure
  contexts are painted correctly when sharing osmesa contexts.

* Add init option to specify what kind of renderering context
  is being used.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/417)

### DIFF
--- a/webrender/src/internal_types.rs
+++ b/webrender/src/internal_types.rs
@@ -43,33 +43,20 @@ impl GLContextHandleWrapper {
 
     pub fn new_context(&self,
                        size: Size2D<i32>,
-                       attributes: GLContextAttributes,
-                       shared: bool) -> Result<GLContextWrapper, &'static str> {
+                       attributes: GLContextAttributes) -> Result<GLContextWrapper, &'static str> {
         match *self {
             GLContextHandleWrapper::Native(ref handle) => {
-                let shared_handle = if shared {
-                    Some(handle)
-                } else {
-                    None
-                };
-
                 let ctx = GLContext::<NativeGLContext>::new(size,
                                                             attributes,
                                                             ColorAttachmentType::Texture,
-                                                            shared_handle);
+                                                            Some(handle));
                 ctx.map(GLContextWrapper::Native)
             }
             GLContextHandleWrapper::OSMesa(ref handle) => {
-                let shared_handle = if shared {
-                    Some(handle)
-                } else {
-                    None
-                };
-
                 let ctx = GLContext::<OSMesaContext>::new(size,
                                                           attributes,
                                                           ColorAttachmentType::Texture,
-                                                          shared_handle);
+                                                          Some(handle));
                 ctx.map(GLContextWrapper::OSMesa)
             }
         }
--- a/webrender/src/internal_types.rs
+++ b/webrender/src/internal_types.rs
@@ -9,6 +9,10 @@ use euclid::{Matrix4D, Point2D, Rect, Size2D};
 use fnv::FnvHasher;
 use freelist::{FreeListItem, FreeListItemId};
 use num_traits::Zero;
+use offscreen_gl_context::{NativeGLContext, NativeGLContextHandle};
+use offscreen_gl_context::{GLContext, NativeGLContextMethods};
+use offscreen_gl_context::{OSMesaContext, OSMesaContextHandle};
+use offscreen_gl_context::{ColorAttachmentType, GLContextAttributes, GLLimits};
 use profiler::BackendProfileCounters;
 use std::collections::{HashMap, HashSet};
 use std::f32;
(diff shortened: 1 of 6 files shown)


glennw:
DO NOT MERGE! (This shouldn't be merged until we're ready to land the headless changes in servo, due to the behavior change when creating webgl contexts).

r? @emilio

webrender/src/render_backend.rs
```diff
@@ -310,7 +318,21 @@ impl RenderBackend {
...
+        // When running in OSMesa mode with texture sharing,
+        // a flush is required on any GL contexts to ensure
+        // that read-back from the shared texture returns
+        // valid data! This should be fine to have run on all
+        // implementations - a single flush for each webgl
+        // context at the start of a render frame should
+        // incur minimal cost.
+        for (_, webgl_context) in &self.webgl_contexts {
+            webgl_context.make_current();
+            gl::flush();
```
emilio: I'm fine with this, but are you sure the `if let Some(id)` before this loop isn't a better place to do it? I expect a context change to trigger a flush. Indeed, from a quick look at mesa's source:

```c
   if (curCtx && 
       (curCtx->WinSysDrawBuffer || curCtx->WinSysReadBuffer) &&
       /* make sure this context is valid for flushing */
       curCtx != newCtx &&
       curCtx->Const.ContextReleaseBehavior ==
       GL_CONTEXT_RELEASE_BEHAVIOR_FLUSH)
      _mesa_flush(curCtx);
```

It seems to be the case (the `WinSysDrawBuffer` and `WinSysReadBuffer` are set on `make_current`, and `GL_CONTEXT_RELEASE_BEHAVIOR_FLUSH` is the default one (and we're not changing it anywhere). So the problem is we're not flushing on `unbind` on the OSMesa back-end.

I'm sending a patch to allow unbinding a context in OSMesa to the mesa-dev mailing list, meanwhile I think the cleanest solution is to call `gl::flush` in `OSMesaContext::unbind`, what do you think?

emilio: Mesa patch: https://www.mail-archive.com/mesa-dev@lists.freedesktop.org/msg128408.html

webrender/src/internal_types.rs
```diff
@@ -20,7 +24,130 @@ use std::sync::Arc;
...
+use webrender_traits::{ScrollLayerId, WebGLCommand};
+
+pub enum GLContextHandleWrapper {
+    Native(NativeGLContextHandle),
+    OSMesa(OSMesaContextHandle),
+}
+
+impl GLContextHandleWrapper {
+    pub fn current_native_handle() -> Option<GLContextHandleWrapper> {
+        NativeGLContext::current_handle().map(|handle| {
```
glennw: Fixed

webrender/src/internal_types.rs
```diff
@@ -20,7 +24,130 @@ use std::sync::Arc;
...
+            GLContextHandleWrapper::Native(handle)
+        })
+    }
+
+    pub fn current_osmesa_handle() -> Option<GLContextHandleWrapper> {
+        OSMesaContext::current_handle().map(|handle| {
```
glennw: Fixed

webrender/src/internal_types.rs
```diff
@@ -20,7 +24,130 @@ use std::sync::Arc;
...
+                    Some(handle)
+                } else {
+                    None
+                };
+
+                let ctx = GLContext::<NativeGLContext>::new(size,
+                                                            attributes,
+                                                            ColorAttachmentType::Texture,
+                                                            shared_handle);
+                ctx.map(|ctx| {
```
glennw: Fixed

webrender/src/render_backend.rs
```diff
@@ -261,24 +260,33 @@ impl RenderBackend {
                               .unwrap()
                         }
                         ApiMsg::RequestWebGLContext(size, attributes, tx) => {
-                            if let Some(ref handle) = self.webrender_context_handle {
-                                match GLContext::<NativeGLContext>::new(size, attributes, ColorAttachmentType::Texture, Some(handle)) {
+                            if let Some(ref wrapper) = self.webrender_context_handle {
+                                // Try to create a shared context first.
+                                let mut shared = true;
```
glennw: Yes - your thinking is exactly right. The corresponding servo code is https://github.com/servo/servo/pull/13382/commits/925c8fca9bc1352576acde7b3ce21ea0600f382a - it just saves servo having to worry about what type of context to create.

webrender/src/render_backend.rs
```diff
@@ -310,7 +318,21 @@ impl RenderBackend {
         let mut new_pipeline_sizes = HashMap::new();
 
         if let Some(id) = self.current_bound_webgl_context_id {
-            self.webgl_contexts.get(&id).unwrap().unbind().unwrap();
+            self.webgl_contexts.get(&id).unwrap().unbind();
```
glennw: Fixed

webrender/src/render_backend.rs
glennw: The intent was to ensure that flush is manually called on every context, not just the currently bound one. This seems conservative but safe - perhaps we land it like this, but we could test switching to the loop above once it's all landed and testing?

emilio: Sure, sounds fine, r=me with a comment about being able to do it less conservatively :)

glennw:
@emilio OK, changed this so the readback path is created on the servo side, as it was previously.

emilio:
r=me when it's ready, with or without an squash :)

glennw:
This is safe to merge now, since we changed the webgl behaviour to match current semantics.

PR #420, opened 2016-09-26
Fix upside down WebGL

MortimerGoro:
Fix WebGL texture UVs: WebGL canvas is renderered upside down in servo when webrender is enabled. glReadPixels is returning the correct image.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/420)


emilio:
Hmm... I remember this was eventually correct, but I've verified this is happening right now so...

@bors-servo: r+

Can't wait enough for https://github.com/servo/servo/pull/13382!

PR #430, opened 2016-10-04
Switch to serde_derive

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/430)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 name = "webrender"
-version = "0.5.1"
+version = "0.6.0"
 authors = ["Glenn Watson <gw@intuitionlibrary.com>"]
 license = "MPL-2.0"
 repository = "https://github.com/servo/webrender"
 
 [features]
-default = ["webrender_traits/serde_codegen"]
-serde_macros = ["webrender_traits/serde_macros"]
+default = ["webrender_traits/codegen"]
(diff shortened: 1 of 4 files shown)


nox:
@alexcrichton There is an issue with default features again, somehow running `cargo build --no-default-features --features serde_derive` enables the `with-syntex` feature of `serde_codegen`. Any idea whether this is fixable?

alexcrichton:
@nox features are unioned amongst crates, so it's likely that some other crate is enabling the `with-syntex` feature of `serde_codegen`. As to debugging *which* crate is enabling that, unfortunately we don't have a lot of tooling for that right now.

nox:
> so it's likely that some other crate is enabling the with-syntex feature of serde_codegen

AFAIK no, if I comment the `default` and `codegen` features of `webrender_traits`, `with-syntex` isn't enabled anymore. Will look again though.

nox:
@alexcrichton Did you check the Cargo.lock file? I just noticed that while syntex_syntax is present in it, I see no "Compiling syntex_syntax" message.

webrender/Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [package]
 name = "webrender"
-version = "0.5.1"
+version = "0.6.0"
```
Manishearth: I don't see why this version bump is necessary.

nox:
@bors-servo r=Manishearth

As discussed on IRC.

PR #432, opened 2016-10-05
Implement GLContextDispatcher trait for NativeGLContexts

MortimerGoro:
Implement GLContextDispatcher trait for NativeGLContexts. Required to allow GLContext sharing in some implementations like WGL.

See https://github.com/emilio/rust-offscreen-rendering-context/pull/69 for more info.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/432)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -16,12 +16,13 @@ use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Sender;
 use texture_cache::TextureCache;
 use webrender_traits::{ApiMsg, AuxiliaryLists, BuiltDisplayList, IdNamespace};
-use webrender_traits::{RenderNotifier, WebGLContextId};
+use webrender_traits::{RenderNotifier, WebGLContextId, RenderDispatcher};
 use batch::new_id;
 use device::TextureId;
 use record;
 use tiling::FrameBuilderConfig;
 use gleam::gl;
+use offscreen_gl_context::GLContextDispatcher;
 
 pub struct RenderBackend {
     api_rx: IpcReceiver<ApiMsg>,
@@ -42,6 +43,7 @@ pub struct RenderBackend {
     webgl_contexts: HashMap<WebGLContextId, GLContextWrapper>,
     current_bound_webgl_context_id: Option<WebGLContextId>,
     enable_recording: bool,
+    main_thread_dispatcher: Arc<Mutex<Option<Box<RenderDispatcher>>>>
 }
 
 impl RenderBackend {
@@ -56,7 +58,8 @@ impl RenderBackend {
                webrender_context_handle: Option<GLContextHandleWrapper>,
                config: FrameBuilderConfig,
                debug: bool,
-               enable_recording:bool) -> RenderBackend {
+               enable_recording:bool,
+               main_thread_dispatcher:  Arc<Mutex<Option<Box<RenderDispatcher>>>>) -> RenderBackend {
         let resource_cache = ResourceCache::new(texture_cache,
                                                 device_pixel_ratio,
                                                 enable_aa);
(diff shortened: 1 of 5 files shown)


webrender/src/render_backend.rs
```diff
@@ -42,6 +43,7 @@ pub struct RenderBackend {
     webgl_contexts: HashMap<WebGLContextId, GLContextWrapper>,
     current_bound_webgl_context_id: Option<WebGLContextId>,
     enable_recording: bool,
+    main_thread_dispatcher: Arc<Mutex<Option<Box<RenderDispatcher>>>>
```
emilio: Probably an option of an arc is a better Idea?

webrender/src/render_backend.rs
```diff
@@ -261,7 +265,10 @@ impl RenderBackend {
                         }
                         ApiMsg::RequestWebGLContext(size, attributes, tx) => {
                             if let Some(ref wrapper) = self.webrender_context_handle {
-                                let result = wrapper.new_context(size, attributes);
+                                let dispatcher = Box::new(WebRenderGLDispatcher {
```
emilio: Do this only on windows?

emilio:
I need to review this really carefully. May you open also a PR with your servo-related changes so it's easy to see the code flow?

webrender/src/render_backend.rs
MortimerGoro: I used the same type pattern as the one used in RenderNotifier setter

emilio: Hmm... I see why you need this, but it seems unfortunate.

MortimerGoro:
Still have to rebase the latest servo master, I'll do tomorrow. In the meantime you can check this: https://github.com/MortimerGoro/servo/commit/64f92be79ad5cc2df2fd207a5d7fc085fbb1c6a9, it will be the same code, just changing Dispatcher name to RenderDispatcher.

P.D.: I created another trait called RenderDispatcher in order to not expose internal members of webrender. I assumed that you want that  webrender library users don't need to be aware that offscreen_gl_context is used internally. Servo now uses offscreen_gl_context includes, but I thought that might not be true in the future when webrender is set up as default and skia renderer dropped.

webrender/src/render_backend.rs
MortimerGoro: Yeah, I'll add a macro. It will avoid unneeded allocations in the platforms that don't use this.

MortimerGoro: Done

MortimerGoro:
Any news on this? 

I have a another idea to implement this. We could create and pass the dispatcher directly from WebGLPaintThred in servo, in this call.

` wr_api.request_webgl_context(&size, attrs)`

Let me know what you think

webrender/src/render_backend.rs
```diff
@@ -261,7 +265,16 @@ impl RenderBackend {
                         }
                         ApiMsg::RequestWebGLContext(size, attributes, tx) => {
                             if let Some(ref wrapper) = self.webrender_context_handle {
-                                let result = wrapper.new_context(size, attributes);
+                                let dispatcher: Option<Box<GLContextDispatcher>>;
+                                if cfg!(target_os = "windows") {
+                                    dispatcher = Some(Box::new(WebRenderGLDispatcher {
```
emilio: nit:

```rust
let dispatcher = if cfg!(target_os = "windows") {
    Some(...)
} else {
    None
};

File: webrender/src/renderer.rs [hunk 4]
```diff
@@ -691,6 +696,15 @@ impl Renderer {
         *notifier_arc = Some(notifier);
     }
 
+    /// Sets the new MainThreadDispatcher.
+    ///
+    /// Used to dispatch functions to the main thread's event loop.
+    /// Required to allow GLContext sharing in some implementations like WGL.
```
emilio: probably moving the last part of this comment to the `main_thread_dispatcher` member definition would make it a bit more helpful.

MortimerGoro:
Done :)

webrender/src/render_backend.rs
MortimerGoro: I need to specify the type: Option<Box<GLContextDispatcher>> or rust compiler doesn't have the enough magic to coerce

PR #433, opened 2016-10-06
Add fine grained (per-draw call) GPU profiling.

glennw:
This provides a graph that shows draw call timing for the
previous 20 frames. The colors identify the type of primitive
batch. The lengths identify the GPU time of each batch, relative
to the maximum frame time of the current graph.

This allows us to see visually:
 - How many draw calls are occurring per frame.
 - Why batches are being split.
 - Which draw calls are being reported as frame spikes.
 - Which shaders are the most expensive for a given scene.
 - How batching changes during scrolling / reflows.
 - How long it takes to execute vertex texture uploads, target switches etc.

This makes it much easier to see the effects of experimenting with
new features and optimizations, such as batcher improvements,
primitive caching etc without having to use apitrace.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/433)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -3,11 +3,25 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use debug_render::DebugRenderer;
+use device::GpuSample;
 use euclid::{Point2D, Size2D, Rect};
 use std::collections::vec_deque::VecDeque;
 use std::f32;
+use std::mem;
 use webrender_traits::ColorF;
 use time::precise_time_ns;
+
+const GRAPH_WIDTH: f32 = 1024.0;
+const GRAPH_HEIGHT: f32 = 320.0;
+const GRAPH_PADDING: f32 = 8.0;
+const GRAPH_FRAME_HEIGHT: f32 = 16.0;
+const PROFILE_PADDING: f32 = 10.0;
+
+#[derive(Debug, Clone)]
+pub struct GpuProfileTag {
+    pub label: &'static str,
+    pub color: ColorF,
+}
 
 trait ProfileCounter {
     fn description(&self) -> &'static str;
@@ -195,9 +209,8 @@ pub struct RendererProfileCounters {
 
 pub struct RendererProfileTimers {
     pub cpu_time: TimeProfileCounter,
-    pub gpu_time_paint: TimeProfileCounter,
-    pub gpu_time_composite: TimeProfileCounter,
-    pub gpu_time_total: TimeProfileCounter,
+    pub gpu_time: TimeProfileCounter,
+    pub gpu_samples: Vec<GpuSample<GpuProfileTag>>,
 }
 
 impl RendererProfileCounters {
@@ -221,9 +234,8 @@ impl RendererProfileTimers {
     pub fn new() -> RendererProfileTimers {
         RendererProfileTimers {
             cpu_time: TimeProfileCounter::new("Compositor CPU Time", false),
-            gpu_time_paint: TimeProfileCounter::new("GPU Time (paint)", false),
-            gpu_time_composite: TimeProfileCounter::new("GPU Time (composite)", false),
-            gpu_time_total: TimeProfileCounter::new("GPU Time (total)", false),
+            gpu_samples: Vec::new(),
+            gpu_time: TimeProfileCounter::new("GPU Time", false),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


glennw:
gpu_profiler

webrender/src/profiler.rs
```diff
@@ -352,6 +359,88 @@ impl ProfileGraph {
...
+        });
+    }
+}
+
+impl GpuFrameCollection {
+    fn draw(&self,
+            x: f32,
+            y: f32,
+            debug_renderer: &mut DebugRenderer) -> Rect<f32> {
+        let graph_size = Size2D::new(1024.0, 320.0);
```
pcwalton: Factor this out into a constant?

webrender/src/profiler.rs
```diff
@@ -352,6 +359,88 @@ impl ProfileGraph {
...
+        let bounding_rect = Rect::new(Point2D::new(x, y),
+                                      Size2D::new(graph_size.width + 16.0,
+                                                  graph_size.height + 16.0));
+        let graph_rect = bounding_rect.inflate(-8.0, -8.0);
```
pcwalton: I think I'd prefer this padding to be a constant too

webrender/src/profiler.rs
```diff
@@ -352,6 +359,88 @@ impl ProfileGraph {
...
+        let mut y0 = graph_rect.origin.y;
+
+        let mut max_time = 0;
+        for frame in &self.frames {
+            max_time = cmp::max(max_time, frame.total_time);
+        }
+        let max_time = max_time as f32;
+
+        for frame in &self.frames {
+            let h = 16.0;
```
pcwalton: Constantify this

webrender/src/profiler.rs
```diff
@@ -475,20 +566,29 @@ impl Profiler {
...
         let rect = self.backend_time.draw_graph(self.x_left, self.y_left, "CPU (backend)", debug_renderer);
         self.y_left += rect.size.height + 10.0;
         let rect = self.compositor_time.draw_graph(self.x_left, self.y_left, "CPU (compositor)", debug_renderer);
         self.y_left += rect.size.height + 10.0;
         let rect = self.gpu_time.draw_graph(self.x_left, self.y_left, "GPU", debug_renderer);
         self.y_left += rect.size.height + 10.0;
+        let rect = self.gpu_frames.draw(self.x_left,
+                                        self.y_left,
+                                        debug_renderer);
+        self.y_left += rect.size.height + 10.0;
```
pcwalton: While you're here, could you move 10.0 to being a constant?

We'll appreciate this if we port to different sized screens :)

webrender/src/renderer.rs
```diff
@@ -47,6 +47,43 @@ pub const MAX_VERTEX_TEXTURE_WIDTH: usize = 1024;
...
+// Yellow / dark yellow
+const GPU_TAG_PRIM_ALIGNED_GRADIENT: GpuProfileTag = GpuProfileTag { label: "AlignedGradient", color: ColorF { r: 1.0, g: 1.0, b: 0.0, a: 1.0 } };
+const GPU_TAG_PRIM_ANGLE_GRADIENT: GpuProfileTag = GpuProfileTag { label: "AngleGradient", color: ColorF { r: 0.7, g: 0.7, b: 0.0, a: 1.0 } };
+
+// Cyan
+const GPU_TAG_PRIM_BOX_SHADOW: GpuProfileTag = GpuProfileTag { label: "BoxShadow", color: ColorF { r: 0.0, g: 1.0, b: 1.0, a: 1.0 } };
+
+// Orange
+const GPU_TAG_PRIM_BORDER: GpuProfileTag = GpuProfileTag { label: "Border", color: ColorF { r: 1.0, g: 0.5, b: 0.0, a: 1.0 } };
+
```
pcwalton: It would be nice to have this in the wiki or something instead of just in the code, but you don't have to do this now.

webrender/src/renderer.rs
```diff
@@ -730,11 +765,14 @@ impl Renderer {
 
                 // Block CPU waiting for last frame's GPU profiles to arrive.
                 // In general this shouldn't block unless heavily GPU limited.
-                let paint_ns = self.gpu_profile_paint.get();
-                let composite_ns = self.gpu_profile_composite.get();
+                if let Some(samples) = self.gpu_profile.build_samples() {
+                    profile_timers.gpu_samples = samples.clone();
```
pcwalton: Why the clone?

webrender/src/profiler.rs
```diff
@@ -475,20 +566,29 @@ impl Profiler {
...
             &backend_profile.total_time,
             &renderer_timers.cpu_time,
-            &renderer_timers.gpu_time_paint,
-            &renderer_timers.gpu_time_composite,
-            &renderer_timers.gpu_time_total,
+            &renderer_timers.gpu_time,
         ], debug_renderer, false);
 
+        let mut gpu_time = 0;
+        let gpu_samples = mem::replace(&mut renderer_timers.gpu_samples, Vec::new());
```
pcwalton: Can you use `.sum()` here?

webrender/src/profiler.rs
```diff
@@ -352,6 +359,88 @@ impl ProfileGraph {
...
+                                bounding_rect.origin.y + bounding_rect.size.height,
+                                &ColorF::new(0.1, 0.1, 0.1, 0.8),
+                                &ColorF::new(0.2, 0.2, 0.2, 0.8));
+
+        let w = graph_rect.size.width;
+        let mut y0 = graph_rect.origin.y;
+
+        let mut max_time = 0;
+        for frame in &self.frames {
+            max_time = cmp::max(max_time, frame.total_time);
```
pcwalton: There's a `max_by_key` method for this: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by_key

webrender/src/profiler.rs
glennw: Done

webrender/src/profiler.rs
glennw: Done

webrender/src/profiler.rs
glennw: Done

webrender/src/profiler.rs
glennw: Done

webrender/src/renderer.rs

Issue #336, opened 2016-08-10
webrender does not build with stable rust


jrmuizel:
We need this to able to put it in Gecko.

emilio:
cc @nox, who did a tom of dependency work, how much work is needed for this?

nox:
AFAICT the only thing needed is https://github.com/servo/ipc-channel/commit/0e11b11900e2f0a07ea10c92741d6aa1f3648bb3, but that comes with the serde bump too, which depends on bumping hyper otherwise we get a serde split in servo. I can also just put that on an v0.4.x branch in ipc-channel if we are that hurried. WR and WR_traits themselves don't use fancy unstable features by default.

nox:
Also for Gecko we probably need an implementation of ipc-channel for Windows that actually work. Cc @metajack @vvuk.

drawoc:
Current master (68824209) builds on stable rust (1.12) for me.

nox:
Did you try on Windows?

drawoc:
Uh, no. I'm on 64-bit linux. I thought it might be worth noting that it builds on stable on at least one platform. I wouldn't know about windows.

nox:
Yes it's definitely good news!

glennw:
@nox Should we close this bug now, since I think WR itself builds on stable (we could open bugs in upstream dependencies if we need to)?

nox:
NOT TRUE ANYMORE WOOOOOOOOOOOOOOO

PR #438, opened 2016-10-10
Stop using step_by, which is unstable.

drawoc:
step_by on ranges isn't currently stable (see rust-lang/rust#27741), so stop using that function.

This is enough for webrender to compile on stable rust (1.12) on my machine, so it probably closes issue #336.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/438)


glennw:
Thanks!

PR #443, opened 2016-10-14
Fixed do_clip() undefined behavior

kvark:
According to GLSL docs:
> `genType smoothstep( 	genType edge0, genType edge1, genType x);`
> `smoothstep` performs smooth Hermite interpolation between 0 and 1 when `edge0 < x < edge1`
> Results are undefined if edge0 ≥ edge1. 

Our usage was:
`smoothstep(1.0, 0.0, distance_from_border)`

Also vectorized a bit of the distance calculation code.

P.S. This change has actually fixed the behavior of the clip shader for my case, possibly because `radius == 0.0`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/443)


glennw:
@kvark This can be closed now - I've r+'ed the follow up PR that includes this patch.

kvark:
@glennw thanks, closing

Issue #405, opened 2016-09-14
Add support for masking to images


jrmuizel:
This will be needed for SVG support, css mask-image and will be helpful for us in transitioning Gecko to webrender as we'll often have a mask image instead of the higher level primitive.

kvark:
Looking into this

PR #444, opened 2016-10-14
Shared clip source for the shaders

kvark:
This will make it easier to introduce clipped primitives outside of the currently supported rectangles and images (related to #405). Also includes #443

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/444)


glennw:
Thanks!

PR #445, opened 2016-10-17
Image mask support

### LINKED ISSUE #405: Add support for masking to images

kvark:
Closes #405 

I'm not completely sure if this is the right approach, hoping to get some review/feedback to proceed. The API and logic works for me with `wr_sample`, although the implementation lacks a lot of features to support, like image repeating and layout transformations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/445)

### DIFF
--- a/webrender/res/clip_shared.glsl
+++ b/webrender/res/clip_shared.glsl
@@ -5,14 +5,19 @@
 
 flat varying vec4 vClipRect;
 flat varying vec4 vClipRadius;
+flat varying vec4 vClipMaskUvRect;
+flat varying vec4 vClipMaskLocalRect;
 
 #ifdef WR_VERTEX_SHADER
-void write_clip(Clip clip) {
+void write_clip(ClipInfo clip) {
     vClipRect = vec4(clip.rect.rect.xy, clip.rect.rect.xy + clip.rect.rect.zw);
     vClipRadius = vec4(clip.top_left.outer_inner_radius.x,
                        clip.top_right.outer_inner_radius.x,
                        clip.bottom_right.outer_inner_radius.x,
                        clip.bottom_left.outer_inner_radius.x);
+    //TODO: interpolate the final mask UV
+    vClipMaskUvRect = clip.mask_info.uv_rect;
+    vClipMaskLocalRect = clip.mask_info.local_rect; //TODO: transform
 }
 #endif
 
@@ -29,23 +34,31 @@ float do_clip(vec2 pos) {
     float d_bl = distance(pos, ref_bl);
 
     float pixels_per_fragment = length(fwidth(pos.xy));
-    // TODO: compute the `nudge` separately for X and Y
     float nudge = 0.5 * pixels_per_fragment;
-
-    bool out0 = pos.x < ref_tl.x && pos.y < ref_tl.y && d_tl > vClipRadius.x - nudge;
-    bool out1 = pos.x > ref_tr.x && pos.y < ref_tr.y && d_tr > vClipRadius.y - nudge;
-    bool out2 = pos.x > ref_br.x && pos.y > ref_br.y && d_br > vClipRadius.z - nudge;
-    bool out3 = pos.x < ref_bl.x && pos.y > ref_bl.y && d_bl > vClipRadius.w - nudge;
-
     vec4 distances = vec4(d_tl, d_tr, d_br, d_bl) - vClipRadius + nudge;
-    float distance_from_border = dot(vec4(out0, out1, out2, out3), distances);
+
+    bvec4 is_out = bvec4(pos.x < ref_tl.x && pos.y < ref_tl.y,
+                         pos.x > ref_tr.x && pos.y < ref_tr.y,
+                         pos.x > ref_br.x && pos.y > ref_br.y,
+                         pos.x < ref_bl.x && pos.y > ref_bl.y);
+
+    float distance_from_border = dot(vec4(is_out),
+                                     max(vec4(0.0, 0.0, 0.0, 0.0), distances));
 
     // Move the distance back into pixels.
     distance_from_border /= pixels_per_fragment;
-
     // Apply a more gradual fade out to transparent.
     //distance_from_border -= 0.5;
 
-    return 1.0 - smoothstep(0.0, 1.0, distance_from_border);
+    float border_alpha = 1.0 - smoothstep(0.0, 1.0, distance_from_border);
+
+    bool repeat_mask = false; //TODO
+    vec2 vMaskUv = (pos - vClipMaskLocalRect.xy) / vClipMaskLocalRect.zw;
+    vec2 clamped_mask_uv = repeat_mask ? fract(vMaskUv) :
+        clamp(vMaskUv, vec2(0.0, 0.0), vec2(1.0, 1.0));
+    vec2 source_uv = clamped_mask_uv * vClipMaskUvRect.zw + vClipMaskUvRect.xy;
+    float mask_alpha = texture(sMask, source_uv).r; //careful: texture has type A8
+
+    return border_alpha * mask_alpha;
 }
 #endif
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -257,7 +257,6 @@ PrimitiveInstance fetch_instance(int index) {
 
     return pi;
 }
-
 struct Primitive {
     Layer layer;
     Tile tile;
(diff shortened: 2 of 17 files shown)


glennw:
@kvark Thanks! The general idea here seems good - the public API looks fine to me.

In terms of implementation, I try to avoid using uniform variables whenever possible, to simplify how the batching code works and reduce batch breaks. Instead, what I'd suggest is adding fields to the Clip struct in tiling.rs, and modifying the code in prim_shared to fetch the data from there. That way, there shouldn't be any need to change the batching code at all.

Let me know if that doesn't make sense or you foresee any issues with that :)

kvark:
@glennw thanks for taking a look!
I've actually considered that (passing via `Clip`) but then thought that since the image itself is being passed as a uniform sampler, it doesn't make sense to pass the associated data per batch.
Now that I think of it though, the image we pass in is the atlas, so it makes perfect sense to have different source/dest rectangles per batch given that these refer to different sub-images within the atlas.
I'll implement the proposed change and get back to you ;)

glennw:
@kvark Yep, that's right - we put (almost) everything in an atlas, so it should batch together nicely :)

kvark:
Rebased now, the implementation follows the route you laid out - passing the clip rectangles via the batch data instead of uniforms.

webrender/res/clip_shared.glsl
```diff
@@ -5,6 +5,8 @@
 
 flat varying vec4 vClipRect;
 flat varying vec4 vClipRadius;
+flat varying vec4 vClipMaskUvRect;
+flat varying vec4 vClipMaskScreenRect;
```
glennw: We should probably rename this to be something like vClipMaskLocalRect, since (I think) it's in the local space of the primitive rectangle?

webrender/res/prim_shared.glsl
```diff
@@ -298,7 +297,24 @@ ClipRect fetch_clip_rect(int index) {
...
 
     rect.rect = texelFetchOffset(sData32, uv, 0, ivec2(0, 0));
-    rect.dummy = texelFetchOffset(sData32, uv, 0, ivec2(1, 0));
+    //rect.dummy = texelFetchOffset(sData32, uv, 0, ivec2(1, 0));
+    rect.dummy = vec4(0.0, 0.0, 0.0, 0.0);
+
+    return rect;
+}
+
+struct MaskRect {
```
glennw: Could we rename these fields to be a bit more descriptive? Perhaps uv_rect and local_mask_rect or something like that?

webrender/src/prim_store.rs
```diff
@@ -206,12 +209,19 @@ impl ClipCorner {
 }
 
 #[derive(Debug, Clone)]
+pub struct MaskRect {
```
glennw: Perhaps rename this to ImageMaskInfo or something like that?

webrender/src/renderer.rs
```diff
@@ -552,6 +553,11 @@ impl Renderer {
         let data64_texture = VertexDataTexture::new(&mut device);
         let data128_texture = VertexDataTexture::new(&mut device);
 
+        let dummy_mask_texture = device.create_texture_ids(1)[0];
```
glennw: I wonder if this will cause unnecessary batch breaks when there are clips that don't have masks. We could probably insert a dummy mask into the atlas instead of creating a separate texture. But that doesn't matter for now, we can do it as a follow up if it's an issue.

webrender/src/tiling.rs
```diff
@@ -702,6 +704,23 @@ impl RenderTask {
 pub const SCREEN_TILE_SIZE: i32 = 64;
 pub const RENDERABLE_CACHE_SIZE: DevicePixel = DevicePixel(2048);
 
+/// Per-batch clipping info merged with the mask image.
+#[derive(Clone, Debug)]
+pub struct MaskedClip {
```
glennw: Instead of introducing a new structure here, could we just add the fields/struct to the Clip structure? The Clip structure already handles simple / complex clips, it seems like it's probably reasonable to make the Clip structure able to handle any number of clip types. If we do that, we won't need to modify the API of all the FrameBuilder methods.

webrender/res/clip_shared.glsl
kvark: fixed

webrender/res/prim_shared.glsl
kvark: fixed

webrender/src/prim_store.rs
kvark: fixed

webrender/src/renderer.rs
kvark: I had a hard time figuring out how to properly assign a default masking texture for this case. Ended up introducing `DummyResources` struct to store the white texture as well the opaque one (which were previously forgotten but still created). Let me know if there is a better solution please.

webrender/src/tiling.rs
kvark: `MaskedClip` is a part of the `Clip`, so moving the fields in there direction would break the API as much regardless. The reason I even introduced this was to pack both rectangles into a single 32 byte gpu block, instead of having a 16 byte pad after each.

webrender/res/ps_gradient.fs.glsl
```diff
@@ -0,0 +1,20 @@
...
+void main(void) {
+#ifdef WR_FEATURE_TRANSFORM
+    float alpha = 0.0;
+    vec2 local_pos = init_transform_fs(vLocalPos, vLocalRect, alpha);
+#else
+    float alpha = 1.0;
+    vec2 local_pos = vPos;
+#endif
+
+    alpha = min(alpha, do_clip(local_pos));
```
glennw: This contains a do_clip() call which fails to compile.

webrender/src/renderer.rs
```diff
@@ -676,7 +693,7 @@ impl Renderer {
     fn enable_msaa(&self, _: bool) {
     }
 
-    #[cfg(any(target_os = "windows", unix))]
+    #[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
```
glennw: Mistake during rebase to change this?

webrender/src/tiling.rs
```diff
@@ -680,6 +682,29 @@ impl RenderTask {
...
 
+#[derive(Clone, Copy, Debug)]
+pub enum MaskImageSource {
+    User(ImageKey),
+    Renderer(TextureId),
+}
+
+/// Per-batch clipping info merged with the mask image.
+#[derive(Clone, Debug)]
+pub struct MaskedClip {
```
glennw: I'm still not sure about having this struct. Could we have something like:

```
struct Clip {
  // existing clip rect and corners
  pub mask: Option<MaskImageSource>,
}

And then use that to construct the data to be put in the gpu blocks during prim prepare / cache?

I might be missing something as I haven't actually tried this - but it sounds like it might be a simpler API if it works.

File: webrender/src/renderer.rs (hunk 7 above)
Comment (kvark) [inline webrender/src/renderer.rs]: weird indeed, fixed

File: webrender/res/ps_gradient.fs.glsl (hunk 6 above)
Comment (kvark) [inline webrender/res/ps_gradient.fs.glsl]: oh, nicely spotted!

File: webrender/res/ps_gradient.fs.glsl [hunk 9]
```diff
@@ -11,10 +11,5 @@ void main(void) {
     vec2 local_pos = vPos;
 #endif
 
-    alpha = min(alpha, do_clip(local_pos));
     oFragColor = vColor * vec4(1, 1, 1, alpha);
-
-#ifdef WR_FEATURE_TRANSFORM
```
kvark: @glennw 
Looks like we were multiplying the alpha twice. I assume this was a mistake.

webrender/src/tiling.rs
glennw: I can see your point about the consistency aspect of the structure internals, but I think having the public API type for FrameBuilder be called Clip is important. How about as a compromise we keep the structure layout as is, but rename Clip -> ClipGpu and make it private, and rename MaskedClip -> Clip?

kvark: agreed, fixed now

glennw:
@kvark Looks good! Let's squash and get this merged :)

kvark:
Do you want a single commit? Or 2-3?

glennw:
@kvark The smallest number of commits that make sense as individual commits. 2-3 is fine if that makes sense.

PR #446, opened 2016-10-18
Bring the border radius back on Linux.

emilio:
I just noticed while reading over #444 that this same line is commented for
clipping. This was making the alpha component of the border radius being always
zero.

This brings back the border radius on Linux. I didn't notice any other issue.
Not totally sure why this only happens on Intel cards though, probably is a
floating point math issue?

Also, this improves the appearance with `LIBGL_ALWAYS_SOFTWARE=1` too, which I guess is the canonical representation.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/446)

### DIFF
--- a/webrender/res/ps_border.fs.glsl
+++ b/webrender/res/ps_border.fs.glsl
@@ -39,7 +39,7 @@ float alpha_for_solid_border(float distance_from_ref,
   distance_from_border /= pixels_per_fragment;
 
   // Apply a more gradual fade out to transparent.
-  distance_from_border -= 0.5;
+  // distance_from_border -= 0.5;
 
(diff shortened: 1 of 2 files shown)


emilio:
r? @glennw 

The two last commits are just cleanups though.

glennw:
@mrobinson Does this make sense as a fix? (It seems like the clip/non-clip shader should be consistent with use of that distance_from_line anyway)?

mrobinson:
This change looks okay to me, especially if it fixes an issue. I didn't mean to include the gradual fade out here (I meant to delete the commented-out line), but clearly something is wrong. At some point after this lands, I plan to take a look the bug again in more detail.

PR #449, opened 2016-10-18
Ensure that RGB8 texture cache pages are backed by an FBO.

glennw:
Previously, only RGBA8 and A8 atlas pages were backed by FBO, since
they are the only formats that are blitted to by dynamic resources.
However, texture cache resize makes use of FBO, so ensure that RGB8
pages have an FBO attached too.

In the future, it might make sense to consider only creating the
backing FBO when/as it's needed for resize operations, since they are
very rarely used (we should profile to see if this makes any difference).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/449)


glennw:
r? @vvuk or @pcwalton 

(This fixes the texture cache resize panic in wr-pad).

PR #467, opened 2016-10-24
Replace ServoStackingContextId with ServoScrollRootId

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/467)

### DIFF
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -405,13 +405,13 @@ pub struct ScrollLayerId {
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub enum ScrollLayerInfo {
     Fixed,
-    Scrollable(usize)
+    Scrollable(usize, ServoScrollRootId)
 }
 
 #[derive(Clone, Deserialize, Serialize)]
 pub struct ScrollLayerState {
     pub pipeline_id: PipelineId,
-    pub stacking_context_id: ServoStackingContextId,
+    pub scroll_root_id: ServoScrollRootId,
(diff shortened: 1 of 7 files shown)


mrobinson:
@glennw r?

webrender_traits/src/types.rs
```diff
@@ -397,13 +397,13 @@ pub struct ScrollLayerId {
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub enum ScrollLayerInfo {
     Fixed,
-    Scrollable(usize)
+    Scrollable(usize, ServoScrollRootId)
```
glennw: Should we just call this ScrollRootId instead?

glennw:
@mrobinson OK, the WR update in master has landed. So feel free to merge this when you have the corresponding servo update ready.

webrender_traits/src/types.rs
mrobinson: Hrm. I'm not sure. I kept the same naming scheme as ServoStackingContextId in the off-chance that WebRender also grew an understanding (and an id) for ScrollRoots. I guess we can land this PR and then investigate later once my first round of work is complete.

mrobinson:
The Servo part of this change is here: https://github.com/mrobinson/servo/tree/scroll_id

glennw:
@mrobinson The CI-related issues have been resolved, but this will need to be rebased then it should be fine to land.

mrobinson:
@glennw Thanks. I'll try this one more time.

mrobinson:
@emilio Thanks!

PR #468, opened 2016-10-24
Added shader base name to the source

kvark:
Helps debugging when all you got is a shader source (`apitrace`).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/468)


anholt:
I also highly recommend using glObjectLabel so that debugging you get from inside your GL driver gets the name attached.

glennw:
@anholt Added https://github.com/servo/webrender/issues/469

PR #471, opened 2016-10-25
Add time-averaging profile counter, use it for FPS

vvuk:
The profiler often has such quickly changing values that it's difficult to see what the actual value is.  This adds a new counter type that displays an average value over a specified time.  It's used for the FPS, but could be used elsewhere (or made configurable etc.)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/471)


vvuk:
Fixed, had 3 too many 0's in ONE_SECOND_NS... last minute untested cleanup :)

PR #466, opened 2016-10-22
Change linux to unix

myfreeweb:
[Works on FreeBSD](https://unrelenting.technology/notes/2016-10-21-21-16-04)!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/466)

PR #473, opened 2016-10-25
Fix Android build

mmatyas:
#466 (c0b0ac08f4c829a979814cb1733262288a30e844) breaks the Android build, because both `unix` and `target_os = "android"` becomes true, so you get duplicate definition errors.

(This is a fix for servo/servo#13154)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/473)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -25,7 +25,7 @@ const GL_FORMAT_A: gl::GLuint = gl::RED;
 #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
 const GL_FORMAT_A: gl::GLuint = gl::ALPHA;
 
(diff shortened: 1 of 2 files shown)


mmatyas:
```
/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/rayon-0.4.3/src/par_iter/internal.rs:96:47: 96:48 error: no rules expected the token `;`

/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/rayon-0.4.3/src/par_iter/internal.rs:96         thread_local!{ static ID: bool = false; }
                                                                                                                                                    ^
```

I guess that's not my fault?

PR #475, opened 2016-10-25
Update nightly version

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/475)


glennw:
Let's see if travis passes now with this.

PR #478, opened 2016-10-25
Use texture layers for render target allocations, remove render phases.

glennw:
Instead of splitting the scene into multiple phases, use texture
layers to allocate slices as needed for each pass of the render
frame. These are pooled and only allocated when the frame render
target configuration changes.

Previously, the code asserted if it was unable to fit the render
tasks into a single render target for a single tile. Now, this
(pathological) case is handled correctly.

This makes the upcoming dynamic primitive cache support a lot
simpler to add. It also means that we don't use a ping-pong
allocation style, which typically performs badly on mobile / tiled
rendering architectures.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/478)

### DIFF
--- a/webrender/res/ps_blend.vs.glsl
+++ b/webrender/res/ps_blend.vs.glsl
@@ -13,12 +13,13 @@ void main(void) {
                        src.screen_origin_task_origin.xy;
 
     vec2 local_pos = mix(dest_origin,
-                         dest_origin + src.size.xy,
+                         dest_origin + src.size_target_index.xy,
                          aPosition.xy);
 
-    vec2 st0 = vec2(src.screen_origin_task_origin.zw) / 2048.0;
-    vec2 st1 = vec2(src.screen_origin_task_origin.zw + src.size.xy) / 2048.0;
-    vUv = mix(st0, st1, aPosition.xy);
+    vec2 texture_size = vec2(textureSize(sCache, 0));
+    vec2 st0 = src.screen_origin_task_origin.zw / texture_size;
+    vec2 st1 = (src.screen_origin_task_origin.zw + src.size_target_index.xy) / texture_size;
+    vUv = vec3(mix(st0, st1, aPosition.xy), src.size_target_index.z);
 
     vOp = blend.src_id_target_id_op_amount.z;
     vAmount = blend.src_id_target_id_op_amount.w / 65535.0;
--- a/webrender/res/ps_composite.vs.glsl
+++ b/webrender/res/ps_composite.vs.glsl
@@ -10,20 +10,21 @@ void main(void) {
     Tile dest = fetch_tile(composite.src0_src1_target_id_op.z);
 
     vec2 local_pos = mix(dest.screen_origin_task_origin.zw,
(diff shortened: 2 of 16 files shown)


glennw:
@pcwalton btw I tested this on mac mini hardware, since I know we've had troubles with driver stalls around texture / fbo allocs before. It seemed to be working nicely on my mac mini (and also on linux + osmesa).

webrender/res/ps_blend.vs.glsl
```diff
@@ -13,12 +13,13 @@ void main(void) {
...
     vec2 local_pos = mix(dest_origin,
-                         dest_origin + src.size.xy,
+                         dest_origin + src.size_target_index.xy,
                          aPosition.xy);
 
-    vec2 st0 = vec2(src.screen_origin_task_origin.zw) / 2048.0;
-    vec2 st1 = vec2(src.screen_origin_task_origin.zw + src.size.xy) / 2048.0;
-    vUv = mix(st0, st1, aPosition.xy);
+    vec2 texture_size = vec2(textureSize(sCache, 0));
+    vec2 st0 = vec2(src.screen_origin_task_origin.zw) / texture_size;
```
pcwalton: nit: Do you need the `vec2()` cast here and below?

webrender/res/ps_composite.vs.glsl
```diff
@@ -10,20 +10,21 @@ void main(void) {
...
-                         dest.screen_origin_task_origin.zw + dest.size.xy,
+                         dest.screen_origin_task_origin.zw + dest.size_target_index.xy,
                          aPosition.xy);
 
-    vec2 st0 = vec2(src0.screen_origin_task_origin.zw) / 2048.0;
-    vec2 st1 = vec2(src0.screen_origin_task_origin.zw + src0.size.xy) / 2048.0;
-    vUv0 = mix(st0, st1, aPosition.xy);
+    vec2 texture_size = vec2(textureSize(sCache, 0));
+    vec2 st0 = vec2(src0.screen_origin_task_origin.zw) / texture_size;
+    vec2 st1 = vec2(src0.screen_origin_task_origin.zw + src0.size_target_index.xy) / texture_size;
```
pcwalton: Same comment about `vec2()` as above

pcwalton:
r=me with a couple of nits

webrender/src/device.rs
```diff
@@ -1075,26 +1099,69 @@ impl Device {
...
 
-            gl::framebuffer_texture_2d(gl::FRAMEBUFFER,
-                                       gl::COLOR_ATTACHMENT0,
-                                       gl::TEXTURE_2D,
-                                       texture_id.0,
-                                       0);
+                let (internal_format, gl_format) = gl_texture_formats_for_image_format(texture.format);
+                let type_ = gl_type_for_texture_format(texture.format);
+
+                gl::tex_image_3d(texture_id.target,
```
kvark: This is concerning. I don't see the previous implementation actually reallocating the texture, like this one. Neither does the method name imply that the texture is going to be changed (`create_fbo_for_texture_if_necessary`).

Issue #407, opened 2016-09-15
Consider using gfx-rs for rendering


kvark:
[GFX](https://github.com/gfx-rs/gfx) has been around for quite a long time, and is still actively developed. We have OpenGL/GLES, DX11 backends working with Metal finishing up and Vulkan in the works. Our [programming model](http://gfx-rs.github.io/2016/09/14/programming-model.html) uses command buffers, allowing WebRender to parallelize. Surely, we could benefit from some cooperation and avoid re-implementing all the backand logic for different graphics APIs.

samlh:
I agree alternate backends may be useful, especially for D3D.

It looks like most of the gl specific code is in [renderer.rs](https://github.com/servo/webrender/blob/master/webrender/src/renderer.rs), [device.rs](https://github.com/servo/webrender/blob/master/webrender/src/device.rs), and (of course) [webgl.rs](https://github.com/servo/webrender/blob/master/webrender_traits/src/webgl.rs). In addition, there are the shaders in [the resource directory](https://github.com/servo/webrender/tree/master/webrender/res) that would need to be ported.

Alas, I think adding an abstraction layer wouldn't help much, and may hurt. A lot of the perf work I've been seeing recently has been in reducing the complexity of the code, and using the underlying apis in ways that avoid triggering driver perf or correctness bugs. Gfx-rs [is comparable in size](https://api.github.com/repos/gfx-rs/gfx/languages) to [the entirety of webrender](https://api.github.com/repos/servo/webrender/languages). The less code used, the fewer places inefficiencies or bugs may hide.

However, I am entirely unaffiliated with both projects, and thus have no say in the matter :).

glennw:
gfx-rs does look pretty awesome, nice work! I haven't spent a huge amount of time looking at the exact APIs, so please forgive my ignorance on the details of gfx-rs.

I will admit to being wary of using libraries that attempt to abstract the differences between rendering APIs to a common interface. I'll try to explain why below :)

In WR, the majority of the code itself is not GL calls. In fact, the core GL rendering loop code, minus boilerplate, is only a couple of hundred lines of code. However, the rest of the code that doesn't call GL and is preparing the data for submission to GL is very tailored towards the *capabilities* of the rendering API. For example, the way that culling, batching, texture atlas allocation, render target management, threading etc is handled is platform independent code, but is highly specific to the *capabilities* of the base GL3 API.

[...]

PR #480, opened 2016-10-26
Isolate GL calls to the device only

kvark:
Also used by the WebGL module, but cutting that tie off would be a much bigger task.
This PR makes the code a bit cleaner, also paves the way to #407.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/480)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -39,7 +39,7 @@ const SHADER_VERSION: &'static str = "#version 300 es\n";
 
 static SHADER_PREAMBLE: &'static str = "shared.glsl";
 
-pub type Buffer = u32;
+pub type ViewportDimensions = [u32; 2];
 
 lazy_static! {
     pub static ref MAX_TEXTURE_SIZE: gl::GLint = {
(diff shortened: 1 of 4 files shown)


webrender/src/device.rs
```diff
@@ -1179,7 +1213,7 @@ impl Device {
         self.init_texture(temp_texture_id, old_width, old_height, format, filter, mode, None);
         self.create_fbo_for_texture_if_necessary(temp_texture_id, None);
 
-        self.bind_render_target(Some((texture_id, 0)));
+        self.bind_render_target(Some((texture_id, 0, [1, 1])));
```
glennw: This adds an extra glViewport call here. It should be fine, I think - but perhaps worth a comment just noting why it's ok to pass [1, 1] here.

webrender/src/device.rs
```diff
@@ -1194,7 +1228,7 @@ impl Device {
         self.deinit_texture(texture_id);
         self.init_texture(texture_id, new_width, new_height, format, filter, mode, None);
         self.create_fbo_for_texture_if_necessary(texture_id, None);
-        self.bind_render_target(Some((temp_texture_id, 0)));
+        self.bind_render_target(Some((temp_texture_id, 0, [1, 1])));
```
glennw: Same here

webrender/src/renderer.rs
```diff
@@ -664,7 +658,7 @@ impl Renderer {
             profile_counters: RendererProfileCounters::new(),
             profiler: Profiler::new(),
             enable_profiler: options.enable_profiler,
-            enable_msaa: options.enable_msaa,
+            _enable_msaa: options.enable_msaa,
```
glennw: We could possibly just remove the MSAA related code altogether, since it's unused. But it's fine to leave here for now if you like, we could do that as a follow up.

kvark: Ok, I removed some of the MSAA logic, but left the `RenderOptions` (to avoid breaking the API, who knows if we'll need it back soon?) as well as the device's capability field.

kvark:
Should be ready now.

PR #483, opened 2016-10-26
Move the resources folder to the repo root

### LINKED ISSUE #481: Move the resources folder to the repo root
It would look much better as `root/shaders` instead of `root/webrender/res`

Nachozombie:
closes #481 

This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/483)


glennw:
Thanks for fixing that issue so quickly! I'm not sure this task is a good idea though - see https://github.com/servo/webrender/issues/481#issuecomment-256511797 for more info.

PR #484, opened 2016-10-26
Improved/fixed the command recording

kvark:
This PR changes the recorded data stream. Instead of relying on the `deserialize` to fail for the payload data, I'm sending the message of size 0, followed by an actual payload size to disambiguate it.
A PR to update the [wr-replay](https://github.com/alanamramjit/replay) tool will follow.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/484)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -89,15 +89,15 @@ impl RenderBackend {
 
     pub fn run(&mut self) {
         let mut profile_counters = BackendProfileCounters::new();
-        let mut frame_counter:u32 = 0;
-        if self.enable_recording{
-                fs::create_dir("record").ok();
+        let mut frame_counter: u32 = 0;
+        if self.enable_recording {
+            fs::create_dir("record").ok();
         }
         loop {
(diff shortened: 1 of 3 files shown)


webrender/src/render_backend.rs
```diff
@@ -233,7 +235,10 @@ impl RenderBackend {
                                 self.render()
                             });
 
-                            self.publish_frame(frame, &mut profile_counters);
+                            if self.scene.root_pipeline_id.is_some() {
```
glennw: I think this if() is unnecessary, right? (since by definition this method is going to result in the root pipeline id being set to a value).

kvark: This `if` was intentional, for consistency only

glennw:
@kvark Oops, typo causing compiler error - fix that up when you have time and I'll merge it.

kvark:
@glennw ah, that's what happens when you hit "Ctrl+S" with the wrong Ctrl.

kvark:
Yes, I put it intentionally there for consistency.

PR #485, opened 2016-10-27
Add framework for dynamic/cache render tasks, and use for box shadows.

glennw:
This adds the framework for tiles and primitives to add dependent
render tasks. The proof of concept in this patch uses them to
speed up box shadow rendering, but the dynamic render tasks can
be used for many things in the future. For example, we can use
this system to render rotated clip masks, or cache text runs
to a texture.

The box shadows are rendered by adding a child task that draws the
upper left corner of a box shadow. The primitive shader for box
shadows then applies this rendered patch over the box shadow by
clamping and mirroring the texture coordinates in the fragment
shader.

Dynamic render tasks are added to passes via a cache key that is
defined based on the type of item being cached. This means that
dynamic render tasks are du-duplicated during allocation and
assignment to render passes and targets. This means that if different
tiles require the same cache primitive, it will only get rendered
once. It also means that if two primitive exist that require the
same resource type (e.g. two box shadows with different colors but
same size and blur radius) then the cache item will only be drawn
once to the child render target.

With this change, GPU time on transparent_rects demo drops from
16ms down to 9ms on my test setup.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/485)


webrender/src/tiling.rs
```diff
@@ -515,12 +602,16 @@ impl RenderPass {
...
+            render_tasks.add(&task, self.pass_index);
+            self.targets.last_mut().unwrap().add_task(task,
+                                                      ctx,
+                                                      render_tasks,
+                                                      self.pass_index);
         }
 
         for target in &mut self.targets {
-            target.build(ctx, render_tasks);
+            let child_pass_index = RenderPassIndex(self.pass_index.0 - 1);
```
kvark: is this new index guaranteed to exist?
perhaps, we can have `let child_pass_index = self.pass_index.get_child().unwrap();` instead

webrender/res/cs_box_shadow.fs.glsl
```diff
@@ -0,0 +1,148 @@
...
+    //          ("colorCutoutTop"_sigma(x_{0_l}, x_{0_r}, y_{0_t}, a, b) +
+    //           "colorCutoutBottom"_sigma(x_{0_l}, x_{0_r}, y_{0_b}, a, b))
+    float cCutoutTop = colorCutoutTop(x0l, x0r, y0t, radii, sigma);
+    float cCutoutBottom = colorCutoutBottom(x0l, x0r, y0b, radii, sigma);
+    return cRect - (cCutoutTop + cCutoutBottom);
+}
+
+void main(void) {
+    vec2 pos = vPos.xy;
+    vec2 p0Rect = vBoxShadowRect.xy, p1Rect = vBoxShadowRect.zw;
```
pcwalton: Yeah, we didn't bother to establish a convention. This is my fault. :(

I think we should pick one and stick with it.

webrender/res/cs_box_shadow.vs.glsl
```diff
@@ -0,0 +1,23 @@
...
+
+    vec2 pos = mix(p0, p1, aPosition.xy);
+
+    vBorderRadii = bs.border_radius_edge_size_blur_radius_inverted.xx;
+    vBlurRadius = bs.border_radius_edge_size_blur_radius_inverted.z;
+    vInverted = bs.border_radius_edge_size_blur_radius_inverted.w;
+    vBoxShadowRect = vec4(bs.bs_rect.xy, bs.bs_rect.xy + bs.bs_rect.zw);
+    vPos = (pos - 1.0 - p0) / uDevicePixelRatio + bs.bs_rect.xy - vec2(2.0 * vBlurRadius);
+
+    gl_Position = uTransform * vec4(pos, 0, 1);
```
pcwalton: Shouldn't this be `vec4(pos, 0.0, 1.0)`? I always forget how strict shaders are but I remember getting burned by stuff like this…

webrender/src/prim_store.rs
```diff
@@ -489,13 +518,26 @@ impl PrimitiveStore {
...
 
                 self.cpu_gradients.push(gradient_cpu);
                 metadata
             }
-            PrimitiveContainer::BoxShadow(box_shadow, instance_rects) => {
-                let gpu_prim_address = self.gpu_data64.push(box_shadow);
+            PrimitiveContainer::BoxShadow(box_shadow_gpu, instance_rects) => {
+                // TODO(gw): Account for zoom factor!
+                let edge_size = box_shadow_gpu.edge_size.ceil() * self.device_pixel_ratio;
+                let edge_size = edge_size as i32 + 2;   // Account for bilinear filtering
```
pcwalton: Could you elaborate on what this does? I'm a bit confused…

webrender/src/tiling.rs
```diff
@@ -236,37 +246,69 @@ pub enum PrimitiveFlags {
...
-    fn add(&mut self, task: &RenderTask) {
+    fn add(&mut self, task: &RenderTask, pass: RenderPassIndex) {
         match task.id {
             RenderTaskId::Static(index) => {
                 self.render_task_data[index.0] = task.write_task_data();
             }
+            RenderTaskId::Dynamic(key) => {
+                let index = RenderTaskIndex(self.render_task_data.len());
+                let key = (key, pass);
+                debug_assert!(self.dynamic_tasks.contains_key(&key) == false);
```
pcwalton: I assume you said `== false` here to avoid the double `!` being confusing?

webrender/src/tiling.rs
```diff
@@ -1582,41 +1703,49 @@ impl FrameBuilder {
...
+            };
 
-        self.add_primitive(&prim_rect,
-                           clip_rect,
-                           clip,
-                           PrimitiveContainer::BoxShadow(prim, instance_rects));
+            self.add_primitive(&prim_rect,
+                            clip_rect,
+                            clip,
+                            PrimitiveContainer::BoxShadow(prim_gpu, instance_rects));
```
pcwalton: nit: weird indentation here

webrender/res/cs_box_shadow.vs.glsl
leeoniya: https://github.com/servo/webrender/pull/401

webrender/res/cs_box_shadow.fs.glsl
glennw: I think we need to be able to have both - depending on what the FS does, it may be far more efficient to have one or the other representation. But we should definitely at least have a naming convention for these cases.

webrender/res/ps_box_shadow.vs.glsl
```diff
@@ -13,11 +13,19 @@ void main(void) {
...
+    // Constant offsets to inset from bilinear filtering border.
+    vec2 patch_origin = child_task.data0.xy + vec2(1.0);
+    vec2 patch_size_device_pixels = child_task.data0.zw - vec2(2.0);
+    vec2 patch_size = patch_size_device_pixels / uDevicePixelRatio;
+
+    vUv.xy = (vi.local_clamped_pos - prim.local_rect.xy) / patch_size;
+    vMirrorPoint = 0.5 * prim.local_rect.zw / patch_size;
+
+    vec2 texture_size = vec2(textureSize(sCache, 0));
+    vCacheUvRectCoords = vec4(patch_origin, patch_origin + patch_size_device_pixels) / texture_size.xyxy;
```
glennw: The textures are size in device pixels.

webrender/src/tiling.rs
glennw: Yes, this pass index is guaranteed to exist. If it doesn't it indicates a very serious bug! :)

webrender/src/tiling.rs
glennw: Yep - I find it easier to read. Can change if you prefer though?

glennw:
@pcwalton Rebased, addressed comments and pushed updates.

webrender/src/tiling.rs
pcwalton: Nope, the way you have it works for me!

PR #486, opened 2016-10-27
Add some information about coordinate systems in webrender.

glennw:
This is just a start - we need to document each API that
deviates from the conventions listed here, and also investigate
using strongly typed units where it makes sense to.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/486)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # webrender
-A very incomplete proof of concept GPU renderer for Servo
+A somewhat incomplete proof of concept GPU renderer for Servo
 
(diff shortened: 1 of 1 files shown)


mrobinson:
@glennw Looks good to me, but feel free to wait for @pcwalton.

staktrace:
Just to confirm - when you say CSS pixels, I'm assuming you mean it doesn't include CSS transforms, and instead those transforms are put into the `transform` field of the stacking context?

glennw:
@staktrace Yep, that's right.

PR #492, opened 2016-10-28
Fix Android build

mmatyas:
`gl::MULTISAMPLE` is not defined on Android.

Related: servo/servo#13154

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/492)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -1752,6 +1752,11 @@ impl Device {
         gl::delete_buffers(&[buffer.0]);
     }
 
+    #[cfg(target_os = "android")]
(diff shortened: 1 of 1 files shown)


kvark:
Looks trivial enough for me to approve, assuming bors listens to me :)
@bors-servo r+

webrender/src/device.rs
```diff
@@ -1752,6 +1752,11 @@ impl Device {
         gl::delete_buffers(&[buffer.0]);
     }
 
+    #[cfg(target_os = "android")]
+    pub fn set_multisample(&self, enable: bool) {
```
kvark: note that this would add a warning for unused `enable`

PR #506, opened 2016-10-31
Add some new features to the freetype font platform implementation.

glennw:
* Add support for getting glyph dimensions without having to
  rasterize the glyph first.
* Add support for rasterizing with subpixel antialiasing.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/506)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -133,13 +133,12 @@ impl FontContext {
         })
     }
 
-    pub fn get_glyph(&mut self,
-                     font_key: FontKey,
-                     size: Au,
-                     character: u32,
-                     device_pixel_ratio: f32,
-                     render_mode: FontRenderMode)
-                     -> Option<RasterizedGlyph> {
+    pub fn rasterize_glyph(&mut self,
+                           font_key: FontKey,
+                           size: Au,
+                           character: u32,
+                           device_pixel_ratio: f32,
+                           render_mode: FontRenderMode) -> Option<RasterizedGlyph> {
         match self.get_ct_font(font_key, size, device_pixel_ratio) {
             Some(ref ct_font) => {
                 let glyph = character as CGGlyph;
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -44,6 +44,33 @@ impl RasterizedGlyph {
     }
 }
 
+struct GlyphMetrics {
+    rasterized_left: i32,
+    rasterized_descent: i32,
+    rasterized_ascent: i32,
+    rasterized_width: u32,
+    rasterized_height: u32,
+}
+
+fn get_glyph_metrics(ct_font: &CTFont, glyph: CGGlyph) -> GlyphMetrics {
+    let bounds = ct_font.get_bounding_rects_for_glyphs(kCTFontDefaultOrientation, &[glyph]);
+
+    let rasterized_left = bounds.origin.x.floor() as i32;
+    let rasterized_width =
+        (bounds.origin.x - (rasterized_left as f64) + bounds.size.width).ceil() as u32;
+    let rasterized_descent = (-bounds.origin.y).ceil() as i32;
+    let rasterized_ascent = (bounds.size.height + bounds.origin.y).ceil() as i32;
+    let rasterized_height = (rasterized_descent + rasterized_ascent) as u32;
+
+    GlyphMetrics {
+        rasterized_ascent: rasterized_ascent,
+        rasterized_descent: rasterized_descent,
+        rasterized_left: rasterized_left,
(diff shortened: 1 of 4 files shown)


webrender/src/platform/unix/font.rs
```diff
@@ -82,82 +87,135 @@ impl FontContext {
...
+                    match bitmap_mode {
+                        FT_PIXEL_MODE_MONO => {
+                            // This is not exactly efficient... but it's only used by the
+                            // reftest pass when we have AA disabled on glyphs.
+                            for y in 0..bitmap.rows {
+                                for x in 0..bitmap.width {
+                                    let byte_index = (y * bitmap.pitch) + (x >> 3);
+                                    let bit_index = x & 7;
+                                    let byte_ptr = bitmap.buffer.offset(byte_index as isize);
+                                    let bit = (*byte_ptr & (128 >> bit_index)) != 0;
```
pcwalton: nit: I think this would be a bit clearer as `0x80` instead of `128`

webrender/src/platform/unix/font.rs
```diff
@@ -82,82 +87,135 @@ impl FontContext {
...
-            let result =  FT_Load_Glyph(face.face, character as FT_UInt, flags);
+            let result =  FT_Load_Glyph(face.face, character as FT_UInt, 0);
             if result.succeeded() {
-
                 let void_glyph = (*face.face).glyph;
-                let slot: FT_GlyphSlot = mem::transmute(void_glyph);
-                assert!(!slot.is_null());
+                let slot_ptr: FT_GlyphSlot = mem::transmute(void_glyph);
+                assert!(!slot_ptr.is_null());
+                slot = Some(slot_ptr);
```
pcwalton: This is kinda weird. Why not just `return Some(slot_ptr)`…

webrender/src/platform/unix/font.rs
```diff
@@ -82,82 +87,135 @@ impl FontContext {
...
+            }
+        }
 
-                let bitmap = &(*slot).bitmap;
-                debug_assert!((enable_aa && bitmap.pixel_mode as i8 == FT_PIXEL_MODE_GRAY as i8) ||
-                              (!enable_aa && bitmap.pixel_mode as i8 == FT_PIXEL_MODE_MONO as i8));
+        slot
```
pcwalton: …and then change this to `None`? Then you can remove `slot`.

webrender/src/platform/unix/font.rs
glennw: Fixed

webrender/src/platform/unix/font.rs
glennw: Fixed

glennw:
@pcwalton Addressed review comments, and added a mac implementation of get_glyph_dimensions.

webrender/src/platform/macos/font.rs
```diff
@@ -72,76 +100,96 @@ impl FontContext {
...
+                                character: u32,
+                                device_pixel_ratio: f32) -> Option<(Au, Au)> {
+        self.get_ct_font(font_key, size, device_pixel_ratio).map(|ref ct_font| {
+            let glyph = character as CGGlyph;
+            let metrics = get_glyph_metrics(ct_font, glyph);
+            (Au::from_px(metrics.rasterized_width as i32), Au::from_px(metrics.rasterized_height as i32))
+        })
+    }
+
+    pub fn get_glyph(&mut self,
```
kvark: Perhaps, `rasterize_glyph`? to reflect that this is a heavy operation (contrary to most of the get_* routines being just accessors)

webrender/src/platform/unix/font.rs
```diff
@@ -98,66 +101,120 @@ impl FontContext {
...
-                            let bit_index = x & 7;
-                            let byte_ptr = bitmap.buffer.offset(byte_index as isize);
-                            let bit = (*byte_ptr & (128 >> bit_index)) != 0;
-                            let byte_value = if bit {
-                                0xff
-                            } else {
-                                0
-                            };
-                            final_buffer.extend_from_slice(&[ 0xff, 0xff, 0xff, byte_value ]);
+            unsafe {
```
kvark: perhaps, reduce the unsafe scope here?

webrender/src/platform/macos/font.rs
pcwalton: Yeah, I think this proposed name change is a good idea.

webrender/src/platform/unix/font.rs
glennw: Good idea - how about I do that as a follow up PR today (I have some more changes to the font code coming).

kvark: sounds good! Even better - let's make an easy "help wanted" kind of task for this

PR #508, opened 2016-10-31
Add support for specific blend modes per batch.

glennw:
Also track the previous blend mode while drawing batches
and only set it when changed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/508)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -88,6 +88,12 @@ const GPU_TAG_PRIM_BOX_SHADOW: GpuProfileTag = GpuProfileTag { label: "BoxShadow
 // Orange
 const GPU_TAG_PRIM_BORDER: GpuProfileTag = GpuProfileTag { label: "Border", color: ColorF { r: 1.0, g: 0.5, b: 0.0, a: 1.0 } };
 
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum BlendMode {
+    None,
+    Alpha,
+}
+
 struct VertexDataTexture {
     id: TextureId,
 }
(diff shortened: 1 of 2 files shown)


webrender/src/renderer.rs
```diff
@@ -1390,14 +1396,25 @@ impl Renderer {
...
-            self.device.set_blend(blending_enabled);
+            if batch.key.blend_mode != prev_blend_mode {
+                match batch.key.blend_mode {
+                    BlendMode::None => {
+                        self.device.set_blend(false);
+                    }
+                    BlendMode::Alpha => {
+                        self.device.set_blend(true);
+                    }
+                }
```
pcwalton: nit: could just be `self.device.set_blend(batch.key.blend_mode == BlendMode::Alpha)`

webrender/src/tiling.rs
```diff
@@ -862,11 +871,9 @@ pub struct AlphaBatchKeyFlags(u8);
 
 impl AlphaBatchKeyFlags {
```
pcwalton: Should we use `bitflags!` for this at some point? (Doesn't have to be now)

webrender/src/renderer.rs

webrender/src/tiling.rs
glennw: https://github.com/servo/webrender/issues/509

glennw:
@pcwalton Addressed review comments

webrender/src/renderer.rs
```diff
@@ -88,6 +88,12 @@ const GPU_TAG_PRIM_BOX_SHADOW: GpuProfileTag = GpuProfileTag { label: "BoxShadow
 // Orange
 const GPU_TAG_PRIM_BORDER: GpuProfileTag = GpuProfileTag { label: "Border", color: ColorF { r: 1.0, g: 0.5, b: 0.0, a: 1.0 } };
 
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum BlendMode {
+    None,
```
kvark: Isn't this non-idiomatic? I'd think an `Option<BlendMode>` would be appropriate, where the actual `BlendMode` variants are all blend-enabled, even if the equation is trivial.

PR #510, opened 2016-11-01
Merged all the projects under the Cargo workspace

kvark:
This will speed up the builds a little. Also added CI testing for the sample/replay binaries.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/510)


Cargo.toml
```diff
@@ -0,0 +1,2 @@
+[workspace]
+members = ["webrender", "webrender_traits", "sample", "replay"]
```
larsbergstrom: Total nit: typically, we would put each of these on their own line and sort them. e.g.:
```
[workspace]
members = [
  "replay",
  "sample",
  "webrender",
  "webrender_traits",
]
```

kvark:
Thanks @larsbergstrom for your points! Should be good now.

PR #511, opened 2016-11-01
Add get_image_properties() to resource cache.

glennw:
This allows querying image opacity before an image is added to
the texture cache.

This doesn't change functionality, but it's prep work for some
of the upcoming interface changes in the resource cache that are
needed for subpixel glyph functionality.

This is arguably a much cleaner interface as well, it never really
made sense to have this accessible only via the texture cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/511)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -27,6 +27,11 @@ use webrender_traits::{GlyphDimensions, PipelineId, WebGLContextId};
 
 thread_local!(pub static FONT_CONTEXT: RefCell<FontContext> = RefCell::new(FontContext::new()));
 
+pub struct ImageProperties {
+    pub format: ImageFormat,
+    pub is_opaque: bool,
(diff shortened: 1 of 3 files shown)


webrender/src/resource_cache.rs
```diff
@@ -440,3 +457,23 @@ impl Resource for CachedImageInfo {
...
+            for i in 0..(bytes.len() / 4) {
+                if bytes[i * 4 + 3] != 255 {
+                    is_opaque = false;
+                    break;
+                }
+            }
+            is_opaque
+        }
+        ImageFormat::RGB8 => true,
+        ImageFormat::A8 => false,
```
pcwalton: Probably worth documenting that it doesn't actually check A8

glennw: Fixed

Issue #498, opened 2016-10-28
Support for transformed clip regions


kvark:
Can be implemented via the image masking (#445) and temporary cached targets (#485) to record the transformed clip region into the mask image.
Related to #154 and #427. Also - https://github.com/servo/servo/issues/13578

The implementation is split into 3 parts:
- [x] refactor the clip handling (#512)
- [x] off-screen clip mask generation (#556)
- [x] removal of the clip shaders (#556)
- [x] API upgrade to actually support the transformed clips (#595)

kvark:
Done by #595. Further support/improvement will go as separate issues.

PR #512, opened 2016-11-01
Major refactoring of how clips are processed down the pipeline

kvark:
This moves the clipping hack down the road (from `Frame` to `FrameBuilder`), allowing to eventually implement multiple transformed clips properly (#498), as discussed on IRC with @glennw .

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/512)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -11,10 +11,10 @@ use renderer::BLUR_INFLATION_FACTOR;
 use resource_cache::ResourceCache;
 use std::mem;
 use std::usize;
-use tiling::{Clip, MaskImageSource};
+use texture_cache::TextureCacheItem;
 use util::TransformedRect;
 use webrender_traits::{AuxiliaryLists, ColorF, ImageKey, ImageRendering, WebGLContextId};
-use webrender_traits::{FontKey, ItemRange, ComplexClipRegion, GlyphKey};
+use webrender_traits::{ClipRegion, FontKey, ItemRange, ComplexClipRegion, GlyphKey};
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct SpecificPrimitiveIndex(pub usize);
@@ -56,13 +56,20 @@ pub struct PrimitiveCacheInfo {
     pub key: PrimitiveCacheKey,
 }
 
+#[derive(Debug)]
+pub enum PrimitiveClipSource {
+    NoClip,
+    Complex(Rect<f32>, f32),
+    Region(ClipRegion),
+}
+
 // TODO(gw): Pack the fields here better!
 #[derive(Debug)]
 pub struct PrimitiveMetadata {
     pub is_opaque: bool,
     pub mask_texture_id: TextureId,
-    pub mask_image: Option<ImageKey>,
     pub clip_index: Option<GpuStoreAddress>,
+    pub clip_source: Box<PrimitiveClipSource>,
     pub prim_kind: PrimitiveKind,
     pub cpu_prim_index: SpecificPrimitiveIndex,
     pub gpu_prim_index: GpuStoreAddress,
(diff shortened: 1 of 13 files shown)


kvark:
Now that I think of it, this might be incorrect. I'm borrowing the `scene.pipeline_auxiliary_lists` during both the frame building and rendering (where it used to `clone()` them). Bad things will happen if these lists are changed in-between.

kvark:
I rewrote the code to solve the [concern](https://github.com/servo/webrender/pull/512#issuecomment-257717923) by cloning the aux lists into the `FrameBuilder`. Should be ready now.

glennw:
At the moment, I try do avoid passing the auxiliary lists around to many APIs, and instead resolve them all at one part of the frame. This has the added advantage that we only resolve items from the auxiliary lists when they first come on screen, avoiding a large upfront cost for a new scene. Is it possible do just make clips work the same way as all the other primitives, or does something prevent that?

kvark:
Redesigned the whole thing after talking to @glennw . The auxiliary lists are no longer accessed at all during the primitive population, and are only passed down (by reference, no cloning) during the frame building. This is also an optimization, since we only resolve the aux lists for visible primitives now.
There is a lot of code simplification here as well.

webrender/src/prim_store.rs
```diff
@@ -58,15 +58,38 @@ pub struct PrimitiveCacheInfo {
     pub key: PrimitiveCacheKey,
 }
 
+#[derive(Debug)]
+pub enum PrimitiveClipInfo {
```
glennw: Is the idea here that it gets resolved from Complex / Region to GpuAddress? Would it make more sense to split into a "source" and "resolved" field perhaps?

webrender/src/prim_store.rs
```diff
@@ -58,15 +58,38 @@ pub struct PrimitiveCacheInfo {
...
+    NoClip,
+    Complex(Rect<f32>, f32),
+    Region(ClipRegion),
+    GpuAddress(GpuStoreAddress),
+}
+
+impl PrimitiveClipInfo {
+    pub fn is_clipped(&self) -> bool {
```
glennw: Will this return true even for a rectangular clip only? (If so, that will prevent occlusion culling where it is being used).

webrender/src/prim_store.rs
```diff
@@ -677,23 +662,55 @@ impl PrimitiveStore {
...
                                    device_pixel_ratio: f32,
+                                   dummy_mask_cache_item: &TextureCacheItem,
                                    auxiliary_lists: &AuxiliaryLists) -> bool {
         let metadata = &mut self.cpu_metadata[prim_index.0];
         debug_assert!(metadata.need_to_build_cache);
         metadata.need_to_build_cache = false;
 
-        if let Some(mask_key) = metadata.mask_image {
-            let tex_cache = resource_cache.get_image(mask_key, ImageRendering::Auto, frame_id);
+        let clip_data_mask = match metadata.clip_info.as_ref() {
```
glennw: Does this mean that we'll end up with a complex clip region enabled when the clip region is a simple rect with zero radius? If so, that will introduce a lot more expensive clip shaders than needed.

webrender/src/prim_store.rs
kvark: You are right, this part doesn't seem right. If it's called after after the `prepare_*` body (and I think it does), the only variant here would be `GpuAddress`. Will be fixed in a minute.

glennw:
@kvark I ran the mozilla css tests subset with this patch - border_radius_clip_a.html fails but everything else passes. Could you take a look at that failure when you have time?

webrender/src/prim_store.rs
kvark: I don't think we handled the zero radius anywhere except this piece:
```diff
             if scrollbar_prim.border_radius == 0.0 {
-                self.prim_store.set_complex_clip(scrollbar_prim.prim_index, None);
+                self.prim_store.unset_complex_clip(scrollbar_prim.prim_index)
             } else {
```
The general case of no `ComplexClipRegion` given is handled below, here:
```rust
let clip_data = match clips.len() {
                    0 if clip_region.image_mask.is_none() => None,
```
So I don't think this PR changes the logic in a way that makes the shaders more expensive.

kvark:
@glennw I was also checking the tests, saw that one, and confirmed it happens without this PR equally.

webrender/src/prim_store.rs
glennw: Yup, that should handle it.

webrender/src/prim_store.rs
glennw: OK, I think that's fine - we might like to make some of the conditions marked as unreachable!() or something like that (e.g. if the non-resolved type gets encountered after it should always have been resolved). But no need to do that now.

glennw:
Generally looks good to me - this fits in quite well with the rest of the items that use ItemRange. Just need to get the test failures sorted and then this should be good to go.

glennw:
Looks good, thanks!

PR #517, opened 2016-11-03
Support for non-stacking-context scrolling areas

mrobinson:
When we encounter a stacking context which scrolls its contents, we
treat it as a special non-stacking-context scrolling area. All
scrolling areas with the same servo scroll root id will be scrolled in
tandem. Later changes will give these fake stacking contexts their own
display item.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/517)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -301,65 +301,86 @@ impl Frame {
             None => return false,
         };
 
-        let layer = self.layers.get_mut(&scroll_layer_id).unwrap();
-        if layer.scrolling.started_bouncing_back && phase == ScrollEventPhase::Move(false) {
-            return false
-        }
+        let scroll_root_id = match scroll_layer_id.info {
+            ScrollLayerInfo::Scrollable(_, scroll_root_id) => scroll_root_id,
+            ScrollLayerInfo::Fixed => unreachable!("Tried to scroll a fixed position layer."),
+        };
 
-        let overscroll_amount = layer.overscroll_amount();
-        let overscrolling = CAN_OVERSCROLL && (overscroll_amount.width != 0.0 ||
-                                               overscroll_amount.height != 0.0);
-        if overscrolling {
-            if overscroll_amount.width != 0.0 {
-                delta.x /= overscroll_amount.width.abs()
+        let mut scrolled_a_layer = false;
+        for (layer_id, layer) in self.layers.iter_mut() {
+            if layer_id.pipeline_id != scroll_layer_id.pipeline_id {
+                continue;
             }
-            if overscroll_amount.height != 0.0 {
-                delta.y /= overscroll_amount.height.abs()
+
+            match layer_id.info {
+                ScrollLayerInfo::Scrollable(_, id) if id != scroll_root_id => continue,
+                ScrollLayerInfo::Fixed => continue,
+                _ => {}
             }
-        }
 
-        let is_unscrollable = layer.content_size.width <= layer.local_viewport_rect.size.width &&
-            layer.content_size.height <= layer.local_viewport_rect.size.height;
-
-        let original_layer_scroll_offset = layer.scrolling.offset;
-
-        if layer.content_size.width > layer.local_viewport_rect.size.width {
-            layer.scrolling.offset.x = layer.scrolling.offset.x + delta.x;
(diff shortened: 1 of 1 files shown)


mrobinson:
@glennw r?

webrender/src/frame.rs
```diff
@@ -301,65 +301,86 @@ impl Frame {
             None => return false,
         };
 
-        let layer = self.layers.get_mut(&scroll_layer_id).unwrap();
-        if layer.scrolling.started_bouncing_back && phase == ScrollEventPhase::Move(false) {
-            return false
-        }
+        let scroll_root_id = match scroll_layer_id.info {
+            ScrollLayerInfo::Scrollable(_, scroll_root_id) => scroll_root_id,
+            ScrollLayerInfo::Fixed => unreachable!("Tried to scroll a fixed position layer."),
```
kvark: should it do `return false` instead?

webrender/src/frame.rs
```diff
@@ -301,65 +301,86 @@ impl Frame {
...
+        };
 
-        let overscroll_amount = layer.overscroll_amount();
-        let overscrolling = CAN_OVERSCROLL && (overscroll_amount.width != 0.0 ||
-                                               overscroll_amount.height != 0.0);
-        if overscrolling {
-            if overscroll_amount.width != 0.0 {
-                delta.x /= overscroll_amount.width.abs()
+        let mut scrolled_a_layer = false;
+        for (layer_id, layer) in self.layers.iter_mut() {
```
kvark: can be rewritten as `let scrolled_layer = match self.layers.find_mut(...) { Some(layer) => layer, None => return false };`, followed by the code currently sitting in the `for` body. That would leave some code unchanged.

webrender/src/frame.rs
glennw: I think in this case, get_scroll_layer is guaranteed to never return a position: fixed layer, so the unreachable!() is probably correct here.

webrender/src/frame.rs
```diff
@@ -496,13 +517,26 @@ impl Frame {
...
-                                                            .pre_mul(&stacking_context.transform)
-                                                            .pre_mul(&stacking_context.perspective);
+        // Stacking contexts with scroll roots are currently not "real" stacking contexts,
+        // but are currently represented as stacking contexts in the display list until they
+        // can get their own display item to represent them. Thus we do not adjust the transform
+        // to account for them and we expand the overflow region to include the area above
+        // their origin in the parent context.
+        let (transform, overflow) = if stacking_context.scroll_layer_id.is_none() {
+            (parent_info.layer_relative_transform.pre_translated(stacking_context.bounds.origin.x,
+                                                                stacking_context.bounds.origin.y,
```
glennw: nit: indentation

webrender/src/frame.rs
mrobinson: Yep, @glennw is correct here.

webrender/src/frame.rs
mrobinson: Hrm. The issue here is that I want to scroll all layers that have the same scroll root id, instead of just the first one. Otherwise, perhaps I am misunderstanding your suggestion?

webrender/src/frame.rs
mrobinson: Okay. Will fix this one.

glennw:
@mrobinson r=me - feel free to land when it suits you. Thanks!

kvark:
This PR breaks some tests:
> FAIL [expected PASS] /_mozilla/css/block_formatting_context_negative_margins_a.html
> FAIL [expected PASS] /_mozilla/css/overflow_auto.html
> FAIL [expected PASS] /_mozilla/css/overflow_scroll.html
> FAIL [expected PASS] /_mozilla/css/textarea_space_calculation.html
> FAIL [expected PASS] /_mozilla/css/transform_scroll_layer.html

[wpt.log.zip](https://github.com/servo/webrender/files/572020/wpt.log.zip)

mrobinson:
The Servo half of this PR hasn't landed yet. It's currently at https://github.com/mrobinson/servo/tree/scroll_root_2. I'm doing some final testing and should have the PR posted today.

PR #518, opened 2016-11-03
Remove the unused StackingContext has_stacking_contexts field

mrobinson:
It seems that this field has been unused since the switch to WR2.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/518)


mrobinson:
@glennw r?

PR #519, opened 2016-11-03
Implement WebGLContext resize

MortimerGoro:
Implement WebGLContext resize (canvas.width & canvas.height). I have tested: 
- Fixes WebGL apps that initialize canvas size after calling canvas.getContext("webgl")
- Support WebGL apps that change the canvas size & viewport on window.onresize

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/519)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -216,6 +216,18 @@ impl ResourceCache {
         self.texture_cache.add_raw_update(texture_id, size);
     }
 
+    pub fn update_webgl_texture(&mut self, id: WebGLContextId, texture_id: TextureId, size: Size2D<i32>) {
+        let prev_texture_id = *self.webgl_textures.get(&id).unwrap();
+
+        // Remove existing cache if texture id has changed
+        if prev_texture_id != texture_id {
+            self.texture_cache.add_raw_remove(prev_texture_id);
+        }
+        // Update new texture id and size
+        self.webgl_textures.insert(id, texture_id);
+        self.texture_cache.add_raw_update(texture_id, size);
+    }
+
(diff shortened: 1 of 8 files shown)


webrender/src/texture_cache.rs
```diff
@@ -779,6 +779,13 @@ impl TextureCache {
         })
     }
 
+    pub fn add_raw_remove(&mut self, id: TextureId) {
```
emilio: `add_raw_remove` sounds a bit weird, though it follows the convention. I can't think of an immediately better name now, so I guess this is ok.

webrender/src/resource_cache.rs
```diff
@@ -216,6 +216,20 @@ impl ResourceCache {
         self.texture_cache.add_raw_update(texture_id, size);
     }
 
+    pub fn update_webgl_texture(&mut self, id: WebGLContextId, texture_id: TextureId, size: Size2D<i32>) {
+        { 
+            let prev_texture_id = self.webgl_textures.get(&id).unwrap();
+
+            // Remove existing cache if texture id has changed
+            if !prev_texture_id.eq(&texture_id) {
```
kvark: any reason `!eq()` and `clone()` are used instead of `!=` and `*` (since this is `Copy`)?

webrender/src/texture_cache.rs
kvark: yeah, it does follow the convention. Perhaps, the convention should be revised? as a separate PR

webrender/src/resource_cache.rs
MortimerGoro: I didn't realize that it is `Copy `;)

PR #523, opened 2016-11-03
Remove ResourceList structure and usage.

glennw:
Previously, a resource list would be built at the start of the
frame. This would allow glyphs missing from the texture cache
to be rasterized.

However, sometimes (e.g. with subpixel antialiasing) it's not possible
to know right at the start of the frame exactly what needs to be
rasterized yet.

Instead, we now allow the resource cache to add requests during
the prepare_prim_for_render() stage. Primitives that request
resources are marked as needing a resolve operation. This allows
primitives that request resources to update their UV coords etc
before batching occurs (since texture IDs for resources must be
known before batching can occur).

This also lays most of the groundwork for running the resource
cache rasterizer as a separate thread in the future. This will be
used to ensure that if too many glyphs are requested in one frame,
WR can continue running without blocking, using older glyphs and
then re-render the scene when newer, high resolution glyphs are
available.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/523)


glennw:
Note: Let's get https://github.com/servo/webrender/pull/512 reviewed and merged before landing this (and I'll rebase mine on top of it). This can still be reviewed now though :)

webrender/src/prim_store.rs
```diff
@@ -580,6 +566,74 @@ impl PrimitiveStore {
...
+        for prim_index in self.prims_to_resolve.drain(..) {
+            let metadata = &mut self.cpu_metadata[prim_index.0];
+
+            if let Some(mask_key) = metadata.mask_image {
+                let tex_cache = resource_cache.get_image(mask_key, ImageRendering::Auto);
+                metadata.mask_texture_id = tex_cache.texture_id;
+                if let Some(address) = metadata.clip_index {
+                    let clip = self.gpu_data32.get_slice_mut(address, 6);
+                    let old = clip[5].data; //TODO: avoid retaining the screen rectangle
+                    clip[5] = GpuBlock32::from(ImageMaskInfo {
```
pcwalton: At some point I'd like to factor all these indices out into constants. (You don't have to do this now)

Maybe we could even make macros to make this stuff easier.

webrender/src/prim_store.rs
```diff
@@ -580,6 +566,74 @@ impl PrimitiveStore {
...
+                    let dest_glyphs = self.gpu_data32.get_slice_mut(metadata.gpu_data_address,
+                                                                    text.glyph_range.length);
+
+                    let texture_id = resource_cache.get_glyphs(text.font_key,
+                                                               text.font_size,
+                                                               text.blur_radius,
+                                                               &text.glyph_indices, |index, uv0, uv1| {
+                        let dest_glyph = &mut dest_glyphs[index];
+                        let dest: &mut GlyphPrimitive = unsafe {
+                            mem::transmute(dest_glyph)
```
pcwalton: What is this transmute for?

webrender/src/resource_cache.rs
```diff
@@ -27,11 +26,24 @@ use webrender_traits::{GlyphDimensions, PipelineId, WebGLContextId};
 
 thread_local!(pub static FONT_CONTEXT: RefCell<FontContext> = RefCell::new(FontContext::new()));
 
+pub struct CacheItem {
+    pub texture_id: TextureId,
+    pub uv0: Point2D<f32>,
```
kvark: would be great to specify that this is in texels, not actual normalized UVs, either by better naming, typing, or just comments

webrender/src/resource_cache.rs
```diff
@@ -117,24 +129,43 @@ impl<K,V> ResourceClassCache<K,V> where K: Clone + Hash + Eq + Debug, V: Resourc
...
+
+struct TextRunResourceRequest {
+    key: FontKey,
+    size: Au,
+    blur_radius: Au,
+    glyph_indices: Vec<u32>,
+}
+
+enum ResourceRequest {
+    Image(ImageResourceRequest),
```
kvark: I think it would be clearer to just have `Image(ImageKey, ImageRendeing)`, since `ImageResourceRequest` is not used in any other context anyway

webrender/src/resource_cache.rs
```diff
@@ -117,24 +129,43 @@ impl<K,V> ResourceClassCache<K,V> where K: Clone + Hash + Eq + Debug, V: Resourc
...
+    TextRun(TextRunResourceRequest),
+}
+
 pub struct ResourceCache {
     cached_glyphs: ResourceClassCache<GlyphKey, Option<TextureCacheItemId>>,
     cached_images: ResourceClassCache<(ImageKey, ImageRendering), CachedImageInfo>,
 
     // TODO(pcwalton): Figure out the lifecycle of these.
-    webgl_textures: HashMap<WebGLContextId, TextureId, BuildHasherDefault<FnvHasher>>,
+    webgl_textures: HashMap<WebGLContextId, (TextureId, Size2D<i32>), BuildHasherDefault<FnvHasher>>,
```
kvark: This asks for a new type

webrender/src/resource_cache.rs
```diff
@@ -212,79 +246,35 @@ impl ResourceCache {
...
-        });
-
-        // Update texture cache with any newly rasterized glyphs.
-        resource_list.for_each_glyph(|glyph_key| {
-            self.create_raster_job(glyph_key, frame_id);
-        });
+    pub fn request_image(&mut self,
+                         key: ImageKey,
+                         rendering: ImageRendering) {
+        debug_assert!(self.state == State::AddResources);
```
kvark: I assume this will be converted to a type system rule/restriction later on

webrender/src/resource_cache.rs
```diff
@@ -347,10 +337,32 @@ impl ResourceCache {
...
+            glyph_key.index = *glyph_index;
+            let image_id = self.cached_glyphs.get(&glyph_key, self.current_frame_id);
+            let cache_item = image_id.map(|image_id| self.texture_cache.get(image_id));
+            if let Some(cache_item) = cache_item {
+                let uv0 = Point2D::new(cache_item.pixel_rect.top_left.x as f32,
+                                       cache_item.pixel_rect.top_left.y as f32);
+                let uv1 = Point2D::new(cache_item.pixel_rect.bottom_right.x as f32,
+                                       cache_item.pixel_rect.bottom_right.y as f32);
+                f(loop_index, uv0, uv1);
+                texture_id = cache_item.texture_id;
```
kvark: is this `texture_id` guaranteed to be the same for all glyphs?

webrender/src/tiling.rs
```diff
@@ -456,12 +474,17 @@ impl AlphaBatcher {
...
 
-struct RenderTargetContext<'a> {
+struct CompileTileContext<'a> {
     layer_store: &'a Vec<StackingContext>,
     prim_store: &'a PrimitiveStore,
     render_task_id_counter: AtomicUsize,
 }
 
+struct RenderTargetContext<'a> {
+    layer_store: &'a Vec<StackingContext>,
```
kvark: Any reason not to use `&'a [StackingContext]`? I think that would reduce the indirections on access

webrender/src/tiling.rs
```diff
@@ -2098,75 +2115,79 @@ impl FrameBuilder {
...
-                            &CompiledScreenTileInfo::ComplexAlpha(cmd_count, prim_count) => {
-                                (format!("{}|{}", cmd_count, prim_count), ColorF::new(1.0, 0.0, 0.0, 1.0))
-                            }
-                        };
-                        debug_rects.push(DebugRect {
-                            label: label,
-                            color: color,
+            // Build list of passes, target allocs that each tile needs.
+            for screen_tile in screen_tiles {
+                let rect = screen_tile.rect;        // TODO(gw): Remove clone here
```
kvark: where is the clone?

webrender/src/prim_store.rs
glennw: The existing data is a GpuBlock32 - I need to access it as the original type, in order to modify some fields of it. Is there a better way to express this (the From trait is used to create the GpuBlock32 initially)?

webrender/src/resource_cache.rs
glennw: Added comment explaining that it is in texels, and why.

webrender/src/resource_cache.rs
glennw: Fixed

webrender/src/resource_cache.rs
glennw: Fixed

webrender/src/resource_cache.rs
glennw: Yes.

webrender/src/tiling.rs
glennw: Fixed

webrender/src/tiling.rs
glennw: Removed comment

glennw:
@pcwalton @kvark  OK, addressed those review comments. I rebased kvark's PR on top of this locally, and the conflicts looked straightforward. So it probably doesn't matter which one lands first.

kvark:
In that case, you may add an assert to enforce this

PR #526, opened 2016-11-04
Add API for tweaking debug options.

glennw:
Right now, it's possible to toggle the profiler. In the future
this can include options for subpixel AA, shader hot reload etc.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/526)


pcwalton:
This is an uber-nit, but do you think that toggling is really the best API here? It requires clients to know the state and doesn't allow multiple clients to interact with the system…

Issue #230, opened 2016-03-09
Text is not gamma corrected


jrmuizel:
On Windows and OS X text is blended in a way that tries to be more correct gamma.

For example, OS X blends using a gamma of 2.0. CoreGraphics implements this by reading from the destination framebuffer while shading (this is supported on all platforms that CoreGraphics runs on). Skia has a pre-blend step that tries to mimic this without needing to read from the framebuffer. See SkTMaskGamma_build_correcting_lut near https://skia.googlesource.com/skia.git/+/master/src/core/SkMaskGamma.cpp#78

pcwalton:
How does Core Graphics read from the destination framebuffer while shading? I thought that was unsupported on all desktop GPUs…

jrmuizel:
GL_NV_texture_barrier lets you read from the destination framebuffer if you play nice.

pcwalton:
Very interesting. I do have that extension on my Intel.

This yields another alternative approach to that of WR2.

cc @glennw

jrmuizel:
I think the best approach here is probably to do what Skia does and preblend. This approach will across all of the platforms we need to support and the fact that Skia does it suggests it good enough.

Basically it builds a lookup table for 1 << MAX_LUM_BITS (8) different luminance values using SkTMaskGamma_build_correcting_lut.

these lookup tables are then used to adjust the mask so that when composited using linear blending the result is somewhat similar to a gamma corrected results.

The larger impact of this is that we need to include the luminance value of the text color in the glyph cache key.

glennw:
@jrmuizel That seems like a good approach.

kvark:
If we don't aim to be precisely matching the native OS text rendering, we might also consider just using [sRGB framebuffer](https://www.opengl.org/registry/specs/ARB/framebuffer_sRGB.txt) for the GPU text anti-aliased rendering:
> When blending is enabled along with the new sRGB update mode, red, green, and blue framebuffer color components are treated as sRGB values that are converted to linear color values, blended with the high-precision color values generated by fragment coloring, and then the blend result is encoded for sRGB just prior to being written into the framebuffer

samlh:
Fair warning, that extension has had some confusing specification history - see e.g. https://lists.freedesktop.org/archives/mesa-dev/2016-August/125215.html

jrmuizel:
I've ported the skia preblend code to rust here: https://github.com/jrmuizel/mac-font-stuff/blob/master/src/lib.rs#L57

[...]

PR #527, opened 2016-11-04
Add basic subpixel AA text support on Linux.

glennw:
This is some preliminary support for subpixel AA. Specifically:
 * Only on Linux for now.
 * No gamma correction.
 * Disabled by default (use -Z subpixel-aa to test).
 * Doesn't deal with rotations, subpixel positioning etc.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/527)

### DIFF
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -3,8 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use app_units::Au;
-use internal_types::FontRenderMode;
-use webrender_traits::{FontKey, GlyphDimensions, NativeFontHandle};
+use webrender_traits::{FontKey, FontRenderMode, GlyphDimensions, NativeFontHandle};
 
 use freetype::freetype::{FTErrorMethods, FT_PIXEL_MODE_GRAY, FT_PIXEL_MODE_MONO, FT_PIXEL_MODE_LCD};
 use freetype::freetype::{FT_Done_FreeType, FT_RENDER_MODE_LCD, FT_Library_SetLcdFilter};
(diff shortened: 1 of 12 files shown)


jrmuizel:
Re: gamma correction. I've elaborated some of my thoughts in #230.

webrender/src/tiling.rs
```diff
@@ -105,6 +106,25 @@ impl AlphaBatchHelpers for PrimitiveStore {
...
+                    FontRenderMode::Subpixel => BlendMode::Subpixel(text_run_cpu.color),
+                    FontRenderMode::Alpha | FontRenderMode::Mono => BlendMode::Alpha,
+                }
+            }
+            _ => {
+                if needs_blending {
+                    BlendMode::Alpha
+                } else {
+                    BlendMode::None
+                }
```
pcwalton: nit: could be

    _ if needs_blending => BlendMode::Alpha,
    _ => BlendMode::None

Up to you

webrender/src/platform/unix/font.rs
```diff
@@ -215,6 +217,9 @@ impl FontContext {
                             }
                         }
                         FT_PIXEL_MODE_LCD => {
+                            // Extra subpixel on each side of the glyph.
+                            glyph_width += 2;
+
```
pcwalton: Ugly. :( Oh well, probably can't be helped.

webrender/src/renderer.rs
```diff
@@ -92,6 +92,7 @@ const GPU_TAG_PRIM_BORDER: GpuProfileTag = GpuProfileTag { label: "Border", colo
 pub enum BlendMode {
     None,
     Alpha,
+    Subpixel(ColorF),
```
pcwalton: Could you add a comment saying that the `ColorF` is the color of the text?

webrender/src/tiling.rs
glennw: I've left as is for now - will tidy up when I expand this function today for a different PR.

webrender/src/renderer.rs

PR #529, opened 2016-11-04
Fixed shader error in ps_image_clip

kvark:
>Failed to compile shader: "../webrender/res/ps_image_clip.vs.glsl"
>0:33(6): error: cannot access field `has_pixel_coords' of structure
>0:33(6): error: type mismatch
>0:33(6): error: if-statement condition must be scalar boolean

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/529)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -138,7 +138,7 @@ fn main() {
         enable_recording: false,
         enable_scrollbars: false,
         debug: true,
-        precache_shaders: false,
+        precache_shaders: true,
         renderer_kind: RendererKind::Native,
     };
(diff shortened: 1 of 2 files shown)


sample/src/main.rs
```diff
@@ -138,7 +138,7 @@ fn main() {
         enable_recording: false,
         enable_scrollbars: false,
         debug: true,
-        precache_shaders: false,
+        precache_shaders: true,
```
glennw: We need to not precache shaders by default (it makes CI runs too slow).

PR #530, opened 2016-11-04
[WIP] Image masked screen rectangle

kvark:
Includes #529
Renames `local_rect` -> `screen_rect` for the `ImageMaskData`, since the idea now is that the mask is always screen-aligned, and we are baking anything non-aligned into the off-screen textures (#498).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/530)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -138,7 +138,7 @@ fn main() {
         enable_recording: false,
         enable_scrollbars: false,
         debug: true,
-        precache_shaders: false,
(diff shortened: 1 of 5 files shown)


kvark:
Discussed with Glenn. It appears that we first need the image masking shader logic to work properly in local space. That's going to be a small change. And then, we'll be introducing the masking tiles (in screen space) to either replace the local image masking or be optional alternative to it.

PR #533, opened 2016-11-06
Implement scroll handoff

gterzian:
WIP for servo/servo#13249

To summarize what was discussed in the issue:

1. "If a new scroll gesture starts while you're already scrolled all the way to the edge of the scrollable element under the mouse, then the scroll will be handed off to the next outer scrollable element which is actually scrollable in that direction" servo/servo#13249 (comment)
2. "If you scroll a long page, and somewhere on that page is a smaller scrollable element, you don't want that element to swallow the rest of your scroll gesture if it suddenly happens to move under your mouse. You want the current scroll gesture to keep scrolling the outer page." servo/servo#13249 (comment)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/533)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -331,7 +331,7 @@ impl Frame {
 
     /// Returns true if any layers actually changed position or false otherwise.
     pub fn scroll(&mut self,
-                  mut delta: Point2D<f32>,
+                  delta: Point2D<f32>,
                   cursor: Point2D<f32>,
                   phase: ScrollEventPhase)
                   -> bool {
@@ -345,13 +345,49 @@ impl Frame {
             None => return false,
         };
 
-        let scroll_root_id = match scroll_layer_id.info {
-            ScrollLayerInfo::Scrollable(_, scroll_root_id) => scroll_root_id,
-            ScrollLayerInfo::Fixed => unreachable!("Tried to scroll a fixed position layer."),
+        let non_root_overscroll = if scroll_layer_id != root_scroll_layer_id {
+            // true if the current layer is overscrolling,
+            // and it is not the root scroll layer.
+            let child_layer = self.layers.get(&scroll_layer_id).unwrap();
+            let overscroll_amount = child_layer.overscroll_amount();
+            overscroll_amount.width != 0.0 || overscroll_amount.height != 0.0
+        } else {
+            false
+        };
+
+        let switch_layer = match phase {
+            ScrollEventPhase::Start => {
+                // if this is a new gesture, we do not switch layer,
+                // however we do save the state of non_root_overscroll,
+                // for use in the subsequent Move phase.
+                let mut current_layer = self.layers.get_mut(&scroll_layer_id).unwrap();
+                current_layer.scrolling.should_handoff_scroll = non_root_overscroll;
+                false
+            },
+            ScrollEventPhase::Move(_) => {
+                // Switch layer if movement originated in a new gesture,
+                // from a non root layer in overscroll.
+                let current_layer = self.layers.get_mut(&scroll_layer_id).unwrap();
+                current_layer.scrolling.should_handoff_scroll
+            },
+            ScrollEventPhase::End => {
+                // clean-up when gesture ends.
+                let mut current_layer = self.layers.get_mut(&scroll_layer_id).unwrap();
+                current_layer.scrolling.should_handoff_scroll = false;
+                false
+            },
+        };
+
+        let scroll_root_id = match (switch_layer, scroll_layer_id.info, root_scroll_layer_id.info) {
+            (true, _, ScrollLayerInfo::Scrollable(_, scroll_root_id)) |
+            (true, ScrollLayerInfo::Scrollable(_, scroll_root_id), ScrollLayerInfo::Fixed) |
+            (false, ScrollLayerInfo::Scrollable(_, scroll_root_id), _) => scroll_root_id,
+            (_, ScrollLayerInfo::Fixed, _) => unreachable!("Tried to scroll a fixed position layer."),
         };
 
         let mut scrolled_a_layer = false;
         for (layer_id, layer) in self.layers.iter_mut() {
+            let mut layer_delta = delta;
             if layer_id.pipeline_id != scroll_layer_id.pipeline_id {
                 continue;
             }
@@ -371,10 +407,10 @@ impl Frame {
                                                    overscroll_amount.height != 0.0);
             if overscrolling {
                 if overscroll_amount.width != 0.0 {
-                    delta.x /= overscroll_amount.width.abs()
+                    layer_delta.x /= overscroll_amount.width.abs()
                 }
                 if overscroll_amount.height != 0.0 {
-                    delta.y /= overscroll_amount.height.abs()
+                    layer_delta.y /= overscroll_amount.height.abs()
                 }
             }
 
@@ -385,7 +421,7 @@ impl Frame {
             let original_layer_scroll_offset = layer.scrolling.offset;
 
             if layer.content_size.width > layer.local_viewport_rect.size.width {
-                layer.scrolling.offset.x = layer.scrolling.offset.x + delta.x;
+                layer.scrolling.offset.x = layer.scrolling.offset.x + layer_delta.x;
                 if is_unscrollable || !CAN_OVERSCROLL {
                     layer.scrolling.offset.x = layer.scrolling.offset.x.min(0.0);
                     layer.scrolling.offset.x =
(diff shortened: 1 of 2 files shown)


gterzian:
@glennw  @mrobinson Any input appreciated, following the discussions in servo/servo#13249, I am first trying to implement the 'handoff of scroll from inner to outer element" see 1 above. 

I also moved the "scroll" logic onto `Layer`, because it was getting a bit too much in that one function I felt, and also because currently I am actually 'scrolling' one particular layer outside of the loop that @mrobinson recently implemented. 

Do you guys see what I am trying to do here?

glennw:
@mrobinson is probably best to comment on this stuff.

mrobinson:
@gterzian You'll need to scroll all the layers that have the same scroll root. Otherwise the contents of scrollable divs will be out of sync. I'll be on #servo if you want to discuss in real-time.

gterzian:
@glennw OK I have taken @mrobinson feedback into account, should we ask for a review?

webrender/src/frame.rs
```diff
@@ -298,9 +298,36 @@ impl Frame {
...
+                let mut current_layer = self.layers.get_mut(&scroll_layer_id).unwrap();
+                current_layer.scrolling.non_root_overscroll = false;
+                false
+            },
+        };
+
+        let scroll_root_id = match (switch_layer, scroll_layer_id.info, root_scroll_layer_id.info) {
+            (true, _, ScrollLayerInfo::Scrollable(_, scroll_root_id)) => scroll_root_id,
+            (true, ScrollLayerInfo::Scrollable(_, scroll_root_id), ScrollLayerInfo::Fixed) => scroll_root_id,
+            (false, ScrollLayerInfo::Scrollable(_, scroll_root_id), _) => scroll_root_id,
```
pcwalton: Combine this with the first in an or-pattern, perhaps?

gterzian:
@pcwalton I have noticed, that when I 'switch layer', the overscrolling layer that I am switching out 'jumps' up and comes down only when the scroll has ended, perhaps because I am interfering with some of the overscroll behavior?

glennw:
@gterzian Sorry this has languished for a bit - Unfortunately the overscrolling is on mac only (and the only mac I have available is a mac mini without trackpad). Do you know if that intermittent behaviour you mention above also happens without your patch? If it does, it's probably fine to merge this.

gterzian:
@glennw I finally found what caused that jumping behavior, it was the mutation of `delta` that was spilling over from one loop iteration to the other, so I now switched to copying the delta for each iteration and it has been fixed. (I think it was intermittend previously because the order in which the layers would come up in the iteration would vary).

Since the `scroll` function is called only once with a given `delta`, and it is mutated to scroll a layer, I think that since we now have a loop that can scroll different layers, it makes sense in any case to not have those mutations spill over.

webrender/src/frame.rs
```diff
@@ -302,18 +304,52 @@ impl Frame {
...
-            None => return false,
+        let scroll_layer_id = match (phase, self.get_scroll_layer(&cursor, root_scroll_layer_id),
+            self.current_scroll_layer_id) {
+            (ScrollEventPhase::Start, Some(scroll_layer_id), _) => {
+                self.current_scroll_layer_id = Some(scroll_layer_id);
+                scroll_layer_id
+            },
+            (ScrollEventPhase::Start, None, _) => return false,
+            (_, _, Some(scroll_layer_id)) => scroll_layer_id,
+            (_, _, None) => return false,
```
glennw: We should probably have a condition here that sets the current scroll layer back to None at the end of the scroll event?

webrender/src/frame.rs
```diff
@@ -302,18 +304,52 @@ impl Frame {
...
+        let scroll_root_id = match (switch_layer, scroll_layer_id.info, root_scroll_layer_id.info) {
+            (true, _, ScrollLayerInfo::Scrollable(_, scroll_root_id)) |
+            (true, ScrollLayerInfo::Scrollable(_, scroll_root_id), ScrollLayerInfo::Fixed) |
+            (false, ScrollLayerInfo::Scrollable(_, scroll_root_id), _) => scroll_root_id,
+            (_, ScrollLayerInfo::Fixed, _) => unreachable!("Tried to scroll a fixed position layer."),
         };
 
         let mut scrolled_a_layer = false;
         for (layer_id, layer) in self.layers.iter_mut() {
+            let mut inner_delta = delta;
```
glennw: Perhaps call this layer_delta? And we should be able to remove the ```mut``` from the function parameter for delta, I think?

webrender/src/frame.rs
```diff
@@ -380,12 +416,11 @@ impl Frame {
             if CAN_OVERSCROLL {
                 layer.stretch_overscroll_spring();
             }
-
+
```
glennw: nit: extra space

webrender/src/frame.rs
```diff
@@ -302,18 +304,52 @@ impl Frame {
...
+            let overscroll_amount = child_layer.overscroll_amount();
+            overscroll_amount.width != 0.0 || overscroll_amount.height != 0.0
+        } else {
+            false
         };
 
-        let scroll_root_id = match scroll_layer_id.info {
-            ScrollLayerInfo::Scrollable(_, scroll_root_id) => scroll_root_id,
-            ScrollLayerInfo::Fixed => unreachable!("Tried to scroll a fixed position layer."),
+        let switch_layer = match phase {
```
glennw: I don't completely understand what the non_root_overscroll is doing - could you add some comments explaining?

gterzian:
@glennw thanks for the review! I will address those suggestions in the coming days...

webrender/src/frame.rs
gterzian: I have tried this, and on Mac it makes the scroll stop right in it's track when you stop scrolling, as opposed to the default 'natural continuous scroll'. I think we can consider leaving things as they are, to retain the default behavior on Mac.

webrender/src/frame.rs
gterzian: good one, done.

glennw:
I was just doing some testing with this, but it seems to break scrolling on Linux altogether.

I'm not sure if that's because of the Linux scroll implementation, or because I rebased this on top of master before testing.

Are you able to test this on a Linux machine? (You may be able to repro it on a mac by switching the CAN_OVERSCROLL constant, which is set based on mac/linux).

gterzian:
@glennw thanks for testing. I don't have a Linux machine readily available, however I might be able to run it into a VM of some sorts, given more time. 

I did already follow your suggestion of setting `CAN_OVERSCROLL` to false on Mac, results:

* Overall scrolling seemed fine.
* There is one small issue, which is that since there is no bounce back of an ovescrolling layer, the child layer that we switch away from remains in 'should_handoff_scroll' mode, since it is still scrolled to the end, and it takes at least one extra gesture and some scrolling with the parent layer to start scrolling that child layer again. I've fixed that by only switching layer if `current_layer.scrolling.should_handoff_scroll && non_root_overscroll`.
* I also noticed that sometimes the last phase would be Move(false), as opposed to End, so I added a case to the match for proper cleanup of `layer.should_handoff_scroll`.

I haven't tried yet to rebase off the latest. 

In the meantime can I rebase/edit and push those? I guess it should not create any problems unless you were to work off the branch that you pulled for testing...

gterzian:
UPDATE: I have now also tested after a rebase off master (using https://github.com/servo/servo/pull/14286 on the servo side), and it seems to work on Mac, even with overscrolling turned off. 

@glennw could you describe the Linux problems in more details? Does the scrolling simply not work at all?

glennw:
@gterzian Yep - the scrolling was completely broken. Even on a really simple page with a single scroll layer (like https://news.ycombinator.com) I was unable to scroll at all. But when I reverted the commits in this branch and ran, I could scroll again. That's quite strange that changing CAN_OVERSCROLL doesn't reproduce it - I'm not aware of any other obvious differences between the linux/mac scrolling implementations...

gterzian:
@glennw thanks for the update! Ok, I guess I will need to debug that using linux on virtualbox or something...

gterzian:
@glennw Due to resource constraints on my current machine, I'm having some troubles building and running Servo in a Linux VM for testing. 

In an attempt to try to move things forward while I find a solution, I've broken up this PR into two ones, since we're actually doing two separate things here, and I have a feeling it's only one of the two that has the potential to completely break scrolling on Linux. 

Could I ask you to please test this one on your machine: https://github.com/servo/webrender/pull/599

I am hopeful it will work on Linux, while I will have to further debug this one later: https://github.com/servo/webrender/pull/600

glennw:
@gterzian Sure, I'll test those patches on Linux today and report back the results.

PR #534, opened 2016-11-07
[WIP] Simplify texture names and support more than two textures in the shaders.

nical:
WebRender currently mostly assumes two reusable texture slots: the color and mask textures, which are sometimes used as color and color, and will soon be also used for other purposes like the Y, U and V planes (requiring an extra slot).
The way some shaders refer to the mask texture (sMask) as an extra color texture is confusing, and while it is most often used as a mask, the fact that it is sometimes not used as such means one can never know for sure until careful review of the corresponding shader code.
We also need more samplers for some effects such as YUV images.

I propose that the reusable samplers have generic names (sTexture0, sTexture1, etc.) which are used in the generic places (for example the device mostly does not know whether a texture is used as a color, a mask or a YUV plane), with a collection of specific name aliases (COLOR_TEXTURE_0, MASK_TEXTURE, Y_TEXTURE, etc.) for the code manipulating the samplers for specific purposes. This applies to both the shaders and the various data structures generated in the render backend.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/534)


nical:
This is a work in progress (do not land), I haven't yet figured out all of the places where to use generic names versus specific (for example I have a hunch  that mask_texture_id in prim_store.rs may used as a non-mask texture but I am not sure).
I'd also like to use this refactoring to add extra samplers where they'll be needed for yuv images and rebase my video work on top of this if the approach is good.

Feedback welcome!

kvark:
This redirection would allow us to use less texture slots, so can be useful if we see the problem with the texture slots on the horizon. Alternatively, If we don't anticipate the problem, we can continue using pre-defined slots for stuff, just make it more clean (i.e. add `sColorExtra` instead of re-using `sMask`), which has a benefit of simpler code.

nical:
I'm all for not aliasing the sampler names if we have the budget to do so. As long as I can trust the variable names when I read the code I am happy.

we'd have:
- sColor
- sSecondaryColor (for blending)
- sMask
- sYTexture
- sUTexture
- sVTexture

It would not be too bad to reuse "color" for the yuv planes so we could have instead:

- sColor0 (most things, inclusing Y plane)
- sColor1 (blending, U plane)
- sColor2 (V plane)
- sMask

That's for the shaders. The rust side of things has the same naming issue (even harder to understand in my opinion) which my patch doesn't properly address yet.

glennw:
I'd prefer not to alias the names right now - we can deal with that if/when we hit it. Having sColor0-2 and sMask seems good to me.

nical:
Sounds good to me, I'll update the PR.

nical:
The latest version is still a work in progress. Shaders can receive up to 3 color textures plus 1 mask. AlphaBatchKey specifies the four textures (instead of a color and something that is either a mask or a color but is called mask), I haven't gotten to the primitive store yet, but something similar needs to happen there.

Note that these four SourceTexture enums end up taking a lot more space that two texture ids. If that's an issue we can optimize the SourceTexture data structure (for example encode the external key in the TextureId using some of the spare bits in the target member to differentiate between the two).

nical:
Things are starting to come together. I have removed the SourceTexture/ExternalImage part, which I'll come back with in a separate pull request.

This PR now does the following things:
- Stop reusing texture names for other things (for instance using mask_texture to store a secondary color), both in the shaders and the rust code.
- Support up to three color textures, in preparation for the YUV image work.

webrender/res/clip_shared.glsl
```diff
@@ -58,7 +58,7 @@ float do_clip(vec2 pos) {
     vec2 clamped_mask_uv = repeat_mask ? fract(vMaskUv) :
         clamp(vMaskUv, vec2(0.0, 0.0), vec2(1.0, 1.0));
     vec2 source_uv = clamped_mask_uv * vClipMaskUvRect.zw + vClipMaskUvRect.xy;
-    float mask_alpha = texture(sMask, source_uv).r; //careful: texture has type A8
+    float mask_alpha = texture(sMask,  source_uv).r; //careful: texture has type A8
```
kvark: accidental spacing?

webrender/src/device.rs
```diff
@@ -341,6 +341,8 @@ impl TextureId {
             target: gl::TEXTURE_2D,
         }
     }
+
+    pub fn is_invalid(&self) -> bool { *self == TextureId::invalid() }
```
kvark: while certainly useful, I think it would be better to avoid double-negative, and have `is_valid()` method instead

webrender/src/internal_types.rs
```diff
@@ -180,6 +182,39 @@ pub enum TextureSampler {
...
+    pub fn no_texture() -> Self {
+        PrimitiveBatchTextures {
+            colors: [TextureId::invalid(); 3],
+            mask: TextureId::invalid(),
+        }
+    }
+}
+
+// In some places we need to temporarily bind a texture to any slot.
+pub const DEFAULT_SAMPLER: TextureSampler = TextureSampler::Color0;
```
kvark: Perhaps, we can lean more to the "texture" part of the concept, since eventually we'll have separate samplers from textures, and this term (`DEFAULT_SAMPLER`) will be misleading. Maybe `DEFAULT_TEXTURE`?

webrender/src/internal_types.rs
```diff
@@ -180,6 +182,39 @@ pub enum TextureSampler {
...
+                panic!("There are only 3 color samplers.");
+            }
+        }
+    }
+}
+
+/// Optional textures that can be used as a source in the shaders.
+/// Textures that are not used by the batch are equal to TextureId::invalid().
+#[derive(Copy, Clone, Debug)]
+pub struct PrimitiveBatchTextures {
```
kvark: that's nice! Although, FYI the `mask` one may be gone soon.

webrender/src/prim_store.rs
```diff
@@ -93,7 +93,7 @@ pub enum ImagePrimitiveKind {
 #[derive(Debug)]
 pub struct ImagePrimitiveCpu {
     pub kind: ImagePrimitiveKind,
-    pub color_texture_id: TextureId,
+    pub color_texture: TextureId,
```
kvark: any reason you didn't like the `_id` at the end?

webrender/src/renderer.rs
```diff
@@ -1319,14 +1322,18 @@ impl Renderer {
...
+                         textures: &PrimitiveBatchTextures,
                          max_prim_items: usize,
                          projection: &Matrix4D<f32>) {
         self.device.bind_program(shader, &projection);
         self.device.bind_vao(self.quad_vao_id);
-        self.device.bind_texture(TextureSampler::Color, color_texture_id);
-        self.device.bind_texture(TextureSampler::Mask, mask_texture_id);
+
+        // TODO(nical): This will force all unused slots to be bound to the invalid
+        // TextureId. Should we just leave them out?
```
kvark: yeah, no need to touch what is not used
if a texture is used and not provided (thus forcing us to bind `invalid()`), that's a bug we should be chasing and fixing

webrender/src/internal_types.rs
```diff
@@ -180,6 +182,39 @@ pub enum TextureSampler {
...
+    pub colors: [TextureId; 3],
+    pub mask: TextureId,
+}
+
+impl PrimitiveBatchTextures {
+    pub fn no_texture() -> Self {
```
kvark: perhaps just `empty()` ?

webrender/src/renderer.rs
```diff
@@ -1425,34 +1429,27 @@ impl Renderer {
...
-                        let ubo = self.device.create_ubo(&chunk, UBO_BIND_DATA);
-
-                        self.device.draw_indexed_triangles_instanced_u16(6, chunk.len() as i32);
-                        self.profile_counters.vertices.add(6 * chunk.len());
-                        self.profile_counters.draw_calls.inc();
-
-                        self.device.delete_buffer(ubo);
-                    }
+                    let max_prim_items = self.max_prim_blends;
+                    self.draw_ubo_batch(ubo_data, shader,
```
kvark: interesting. @glennw do you remember why `draw_ubo_batch` wasn't used here in the first place?

webrender/src/tiling.rs
```diff
@@ -861,34 +866,35 @@ impl AlphaBatchKey {
...
 
     fn is_compatible_with(&self, other: &AlphaBatchKey) -> bool {
         self.kind == other.kind &&
             self.flags == other.flags &&
             self.blend_mode == other.blend_mode &&
-        (self.color_texture_id == TextureId::invalid() || other.color_texture_id == TextureId::invalid() ||
-             self.color_texture_id == other.color_texture_id) &&
-            (self.mask_texture_id == TextureId::invalid() || other.mask_texture_id == TextureId::invalid() ||
-             self.mask_texture_id == other.mask_texture_id)
+            (self.textures.colors[0].is_invalid() || other.textures.colors[0].is_invalid() ||
```
kvark: could use some iteration here for DRY

webrender/src/renderer.rs
nical: FWIW we already explicitly bind to TextureId::invalid in some places such as https://github.com/servo/webrender/blob/74a21e9175a8eb0fc8ce1a18b96c7fb966efb3cb/webrender/src/renderer.rs#L1157
The renderer currently also does that for the color and mask textures every time.

In practice it unbinds the texture since the "invalid" gl handle is 0 under the hood.

I don't know if unbinding an unused texture slot after a pass that used it has any noticeable overhead so I added the comment bu it's probably not a big deal.

webrender/src/prim_store.rs
nical: that's an artifact from when I had replaced some of the TextureId by the SourceTexture enum (and thus removed the _id suffix) before splitting that out for a future PR. I'll correct that.

webrender/src/device.rs
nical: I originally had is_valid, but all call sites were negating it ```if !texture.is_valid() { ...``` so I ended up bundling the negation in the function. I don't mind either way, whichever you prefer works for me.

kvark: Perf-wise, marking it `#[inline]` makes both equally efficient, and `!texture.is_valid()` takes even 1 less char than `texture.is_invalid()`, so it's not like we are sacrificing compactness for clarity here :)

webrender/src/renderer.rs
glennw: Hmmm, I'm not sure - possibly just an oversight.

webrender/src/tiling.rs
```diff
@@ -861,34 +866,35 @@ impl AlphaBatchKey {
...
 
     fn is_compatible_with(&self, other: &AlphaBatchKey) -> bool {
         self.kind == other.kind &&
             self.flags == other.flags &&
             self.blend_mode == other.blend_mode &&
-        (self.color_texture_id == TextureId::invalid() || other.color_texture_id == TextureId::invalid() ||
-             self.color_texture_id == other.color_texture_id) &&
-            (self.mask_texture_id == TextureId::invalid() || other.mask_texture_id == TextureId::invalid() ||
-             self.mask_texture_id == other.mask_texture_id)
+            (!self.textures.colors[0].is_valid() || !other.textures.colors[0].is_valid() ||
```
glennw: Can we add a (inline) function to check if two textures are compatible? That would make this function a lot easier to read.

glennw:
@nical BTW, I've finished up the work I was doing for text-shadow to allow me to work on the texture cache. This is mostly related to improvements for handling the glyph cache, but is likely to overlap quite a bit with the SourceTexture / ExternalTexture stuff you mentioned - so we should discuss that at some point. In the meantime, let's get this PR finished up and merged as soon as we can :)

nical:
I rebased everything on top of current master (which was epic, I should have squashed the commits before rather than after doing that), and squashed it down to one commit.

PR #535, opened 2016-11-07
Add stride data to add image

changm:
This adds support for a custom stride versus tightly packed image data. A couple of questions though:

1) In get_texture_info, please review this carefully. I want to make sure I used Cow right and got the lifetimes right.
2) The current iteration requires stride data on ApiMsg::AddImage. I'm not sure this will break Servo, a second patch for the gecko-wr integration fixes this up. Alternatively, we can add an ApiMsg::AddImageWithStride if you'd prefer.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/535)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -15,6 +15,8 @@ use std::hash::BuildHasherDefault;
 use std::io::Read;
 use std::path::PathBuf;
 use std::mem;
+use std::borrow::Cow;
+use std::borrow::Borrow;
 //use std::sync::mpsc::{channel, Sender};
 //use std::thread;
 use webrender_traits::ImageFormat;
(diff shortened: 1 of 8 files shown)


webrender/src/device.rs
```diff
@@ -1532,36 +1534,67 @@ impl Device {
...
-        debug_assert!(self.inside_frame);
-
-        let mut expanded_data = Vec::new();
-
-        let (gl_format, bpp, data) = match self.textures.get(&texture_id).unwrap().format {
+    fn get_texture_info<'a>(&mut self, texture_id: TextureId, data: &'a [u8]) -> (u32, u32, Cow<'a, [u8]>) {
+        match self.textures.get(&texture_id).unwrap().format {
             ImageFormat::A8 => {
                 if cfg!(any(target_arch="arm", target_arch="aarch64")) {
+                    // Since we can allocate data here, i don't think we can
```
glennw: I'm not sure why we need this - can we just have a single update_texture function which always expects stride? In which case there's no need for the get_texture_info() function, and thus no need for Cow.

webrender/src/internal_types.rs
```diff
@@ -335,6 +335,7 @@ pub enum RenderTargetMode {
 pub enum TextureUpdateDetails {
     Raw,
     Blit(Vec<u8>),
+    BlitWithStride(Vec<u8>, u32),
```
glennw: Instead of introducing a new type here, could we just add the stride to the Blit enum variant and always expect stride?

webrender/src/resource_cache.rs
```diff
@@ -209,12 +210,14 @@ impl ResourceCache {
...
                               width: u32,
                               height: u32,
+                              stride: u32,
                               format: ImageFormat,
                               bytes: Vec<u8>) {
         let resource = ImageResource {
             is_opaque: is_image_opaque(format, &bytes),
             width: width,
             height: height,
+            stride: width, // This is an invalid stride for now, fixup
```
glennw: Why is this?

PR #538, opened 2016-11-08
add stride data as an optional field to upload textures

changm:
Thanks for the idea of using option! That worked quite well :)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/538)


webrender/src/device.rs
```diff
@@ -1562,7 +1563,19 @@ impl Device {
...
 
-        assert!(data.len() as u32 == bpp * width * height);
+        let row_length = match stride {
+            Some(value) => value / bpp,
+            None => width,
+        };
+
+        assert!(data.len() as u32 == bpp * row_length * height);
+
+        match stride {
```
glennw: The normal way to write this would be:

```
if let Some(stride) = stride {
 ...
}
```

Then you won't need the empty match case.

webrender/src/device.rs
```diff
@@ -1572,6 +1585,12 @@ impl Device {
                                          height as gl::GLint,
                                          gl_format,
                                          data);
+
+        // Reset row length to 0, otherwise the stride would apply to all texture uploads.
+        match stride {
```
glennw: Same here

glennw:
@changm Looks good! Just a minor code style change, then we'll get this merged.

changm:
Thanks! PR fixed.

glennw:
@changm Ah, you'll need to fix up the calls in the wr-sample/ and replay/ dirs. You can confirm these are working by running 'cargo build' in those directories.

Issue #509, opened 2016-10-31
Switch AlphaBatchKeyFlags to use bitflags! crate.


glennw:

moriturus:
I'm going to try it.

dignifiedquire:
It looks like https://github.com/servo/webrender/pull/541 was merged, which supposedly does this so can this be closed?

glennw:
Yup, this happened.

PR #539, opened 2016-11-08
Switch AlphaBatchKeyFlags to use bitflags! crate. 

moriturus:
servo/webrender#509

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/539)


webrender/src/tiling.rs
```diff
@@ -818,6 +818,42 @@ enum AlphaBatchKind {
     BoxShadow,
 }
 
+bitflags! {
+    pub flags AlphaBatchKeyFlags: u8 {
+        const AXIS_ALIGNED    = 0b00000010,
+        const COMPLEX         = 0b00000100,
+        const NEEDS_CLIPPLING = 0b00000001,
```
emilio: Maybe sorting them from smallest to largest?

webrender/src/tiling.rs
```diff
@@ -818,6 +818,42 @@ enum AlphaBatchKind {
...
+    }
+}
+
+impl AlphaBatchKeyFlags {
+    fn new(transform_kind: TransformedRectKind,
+        needs_clipping: bool) -> AlphaBatchKeyFlags {
```
emilio: nit: align the arguments, please :)

emilio: Though now seems easier to just use the literals and or them as you've done below, so we might as well get rid of this function entirely.

webrender/src/tiling.rs
mstange: There's an extra L in the word CLIPPLING. It should be CLIPPING.

webrender/src/tiling.rs
```diff
@@ -818,6 +825,28 @@ enum AlphaBatchKind {
     BoxShadow,
 }
 
+bitflags! {
+    pub flags AlphaBatchKeyFlags: u8 {
+        const NEEDS_CLIPPING  = 0b00000001,
+        const AXIS_ALIGNED    = 0b00000010,
+        const COMPLEX         = 0b00000100,
```
glennw: I think we can remove the COMPLEX flag altogether, since it's mutually exclusive with AXIS_ALIGNED and isn't actually used.

glennw: The COMPLEX flag is set there, but it's never actually tested against - the fn transform_kind() function just checks the valule of AXIS_ALIGNED (since the presence of one implies the absence of the other). Something seems to have gotten confused with the merge / rebase too - it looks like there is part of another commit in this PR now?

moriturus:
Oops, I accidentally remove this PR's branch. I will re-open My PR.

PR #540, opened 2016-11-08
Implement Home end key scroll

samuknet:
Addressing issue https://github.com/servo/servo/issues/13082.
Supporting https://github.com/servo/servo/pull/14141.

* Adds `ScrollLocation` type in `webrender_traits`.
* Refactor api to use new `ScrollLocation`
* Implement home/end scrolling in `webrender/src/frame.rs` `fn scroll(...)` using new `ScrollLocation` struct passed in.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/540)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -19,7 +19,7 @@ use tiling::{AuxiliaryListsMap, FrameBuilder, FrameBuilderConfig, PrimitiveFlags
 use util::MatrixHelpers;
 use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLayerId};
 use webrender_traits::{ClipRegion, ColorF, DisplayItem, StackingContext, FilterOp, MixBlendMode};
-use webrender_traits::{ScrollEventPhase, ScrollLayerInfo, SpecificDisplayItem, ScrollLayerState};
+use webrender_traits::{ScrollEventPhase, ScrollLayerInfo, ScrollLocation, SpecificDisplayItem, ScrollLayerState};
 use webrender_traits::{LayerRect, LayerPoint, LayerSize};
 use webrender_traits::{ServoScrollRootId, ScrollLayerRect, as_scroll_parent_rect, ScrollLayerPixel};
 use webrender_traits::WorldPoint4D;
@@ -331,10 +331,11 @@ impl Frame {
 
     /// Returns true if any layers actually changed position or false otherwise.
     pub fn scroll(&mut self,
-                  mut delta: Point2D<f32>,
+                  scroll_location: ScrollLocation,
                   cursor: Point2D<f32>,
                   phase: ScrollEventPhase)
                   -> bool {
+
         let root_scroll_layer_id = match self.root_scroll_layer_id {
             Some(root_scroll_layer_id) => root_scroll_layer_id,
             None => return false,
@@ -350,6 +351,37 @@ impl Frame {
             ScrollLayerInfo::Fixed => unreachable!("Tried to scroll a fixed position layer."),
         };
 
+        let mut delta:Point2D<f32> = match scroll_location {
+            ScrollLocation::Delta(delta) => delta,
+            ScrollLocation::Start => {
+                if layer.scrolling.offset.y.round() == 0.0 {
+                    // Nothing to do.
+                    return false;
+                }
+
+                layer.scrolling.offset.y = 0.0;
+                return true;
+            },
+            ScrollLocation::End => {
+                let end_pos = -layer.content_size.height +
+                                                 (layer.local_viewport_rect.size.height);
+
+                if layer.scrolling.offset.y.round() == end_pos {
+                    // Nothing to do.
+                    return false;
+                }
+                
+                layer.scrolling.offset.y = end_pos;
+                return true;
+            },
+        };
+
+        let overscroll_amount = layer.overscroll_amount();
+        let overscrolling = CAN_OVERSCROLL && (overscroll_amount.width != 0.0 ||
+                                               overscroll_amount.height != 0.0);
+        if overscrolling {
+            if overscroll_amount.width != 0.0 {
+                delta.x /= overscroll_amount.width.abs()
         let mut scrolled_a_layer = false;
         for (layer_id, layer) in self.layers.iter_mut() {
             if layer_id.pipeline_id != scroll_layer_id.pipeline_id {
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -19,15 +19,11 @@ use tiling::{AuxiliaryListsMap, FrameBuilder, FrameBuilderConfig, PrimitiveFlags
 use util::MatrixHelpers;
 use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLayerId};
 use webrender_traits::{ClipRegion, ColorF, DisplayItem, StackingContext, FilterOp, MixBlendMode};
-<<<<<<< 22aefa1b9daa4ae0b8e8d161d9d0173bb2117c9b
 use webrender_traits::{ScrollEventPhase, ScrollLayerInfo, ScrollLocation, SpecificDisplayItem, ScrollLayerState};
 use webrender_traits::{LayerRect, LayerPoint, LayerSize};
 use webrender_traits::{ServoScrollRootId, ScrollLayerRect, as_scroll_parent_rect, ScrollLayerPixel};
 use webrender_traits::WorldPoint4D;
 use webrender_traits::{LayerTransform, LayerToScrollTransform, ScrollToWorldTransform};
-=======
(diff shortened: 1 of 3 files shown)


webrender/src/frame.rs
```diff
@@ -282,9 +282,9 @@ impl Frame {
         result
     }
 
-    /// Returns true if any layers actually changed position or false otherwise.
+      /// Returns true if any layers actually changed position or false otherwise.
```
glennw: nit: indentation

webrender/src/frame.rs
```diff
@@ -378,7 +404,7 @@ impl Frame {
         }
 
         scrolled_a_layer
-    }
+}
```
glennw: nit: indentation

webrender/src/frame.rs
```diff
@@ -319,6 +319,32 @@ impl Frame {
                 continue;
             }
 
+           let mut delta:Point2D<f32> = match scroll_location {
```
glennw: nit: indentation of this block

webrender/src/frame.rs
```diff
@@ -319,6 +319,32 @@ impl Frame {
...
+                    // Nothing to do.
+                    return false;
+                }
+
+                layer.scrolling.offset.y = 0.0;
+                return true;
+            },
+            ScrollLocation::End => {
+                let end_pos = -layer.content_size.height +
+                                                 (layer.local_viewport_rect.size.height);
```
glennw: This is probably clearer as layer.local_viewport_rect.size.height - layer.content_size.height

webrender/src/frame.rs
```diff
@@ -319,6 +319,32 @@ impl Frame {
                 continue;
             }
 
+           let mut delta:Point2D<f32> = match scroll_location {
+            ScrollLocation::Delta(delta) => delta,
+            ScrollLocation::Start => {
+                if layer.scrolling.offset.y.round() == 0.0 {
+                    // Nothing to do.
+                    return false;
```
glennw: I *think* that we don't want to return straight away here (or below) due to the recent scroll root changes that have been made. Is that correct @mrobinson  ?

samuknet:
@glennw thanks 😃  Just pushed those indentation fixes.

webrender/src/frame.rs
```diff
@@ -319,6 +319,31 @@ impl Frame {
                 continue;
             }
 
+            let mut delta:Point2D<f32> = match scroll_location {
```
kvark: is the explicit type required here?

webrender/src/frame.rs
```diff
@@ -319,6 +319,31 @@ impl Frame {
...
+                    if layer.scrolling.offset.y.round() == 0.0 {
+                        // Nothing to do.
+                        return false;
+                    }
+
+                    layer.scrolling.offset.y = 0.0;
+                    return true;
+                },
+                ScrollLocation::End => {
+                    let end_pos = -layer.content_size.height +
```
kvark: weird indentation and brackets

webrender/src/frame.rs
```diff
@@ -319,6 +319,31 @@ impl Frame {
...
+                                                 (layer.local_viewport_rect.size.height);
+
+                    if layer.scrolling.offset.y.round() == end_pos {
```
kvark: should it be `>=` instead of `==`?

webrender/src/frame.rs
```diff
@@ -319,6 +319,31 @@ impl Frame {
...
+            let mut delta:Point2D<f32> = match scroll_location {
+                ScrollLocation::Delta(delta) => delta,
+                ScrollLocation::Start => {
+                    if layer.scrolling.offset.y.round() == 0.0 {
+                        // Nothing to do.
+                        return false;
+                    }
+
+                    layer.scrolling.offset.y = 0.0;
+                    return true;
```
kvark: So if it's `Start` or `End` you don't want to even proceed to the rest of the function body, even if the scroll happened?

webrender_traits/src/types.rs
```diff
@@ -439,6 +439,13 @@ pub enum ScrollPolicy {
     Fixed,
 }
 
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub enum ScrollLocation {
+    Delta(Point2D<f32>), // Scroll by a certain amount.
```
kvark: please move those comments atop of the elements, preceded by `///` for some real documentation

webrender/src/frame.rs
gterzian: @glennw and @samuknet from my own discussions with @mrobinson with regards to https://github.com/servo/webrender/pull/533 I am pretty much certain that, as @glennw suggested, those `return` statements should be turned into `continue`, because the loop would still need to scroll any other layer with the same `scroll_root_id`

webrender/src/frame.rs
```diff
@@ -319,6 +319,31 @@ impl Frame {
                 continue;
             }
 
+            let mut delta:Point2D<f32> = match scroll_location {
+                ScrollLocation::Delta(delta) => delta,
+                ScrollLocation::Start => {
+                    if layer.scrolling.offset.y.round() == 0.0 {
+                        // Nothing to do.
+                        return false;
```
gterzian: from my own discussions with @mrobinson with regards to https://github.com/servo/webrender/pull/533 I am pretty much certain that, as @glennw suggested in https://github.com/servo/webrender/pull/540#discussion_r87093181, those `return` statements should be turned into `continue`, because the loop would still need to scroll any other layer with the same `scroll_root_id`. 

So instead of returning a boolean, you could do something like `scrolled_a_layer = scrolled_a_layer || your_boolean`, followed by `continue;`

webrender/src/frame.rs
samuknet: Yes, I think `>=` makes sense

webrender/src/frame.rs
samuknet: okay, sure :)

I was thinking something like this would work for these blocks:

```rust 
ScrollLocation::Start => {
  if layer.scrolling.offset.y.round() == 0.0 {
    // Nothing to do on this layer.
    continue;
  }

  layer.scrolling.offset.y = 0.0;
  scrolled_a_layer = true;
  continue;
},
```

What do you think? I have put these changes in and updated with latest WR master in the latest push

webrender/src/frame.rs
```diff
@@ -328,6 +328,33 @@ impl Frame {
...
+                        continue;
+                    }
+
+                    layer.scrolling.offset.y = 0.0;
+                    scrolled_a_layer = true;
+                    continue;
+                },
+                ScrollLocation::End => {
+                    let end_pos = -layer.content_size.height +
+                                  layer.local_viewport_rect.size.height;
```
glennw: Let's make this layer.local_viewport_rect.size.height - layer.content_size.height (just reads a bit better switching the order around).

webrender/src/frame.rs
```diff
@@ -328,6 +328,33 @@ impl Frame {
...
+                ScrollLocation::Start => {
+                    if layer.scrolling.offset.y.round() <= 0.0 {
+                        // Nothing to do on this layer.
+                        continue;
+                    }
+
+                    layer.scrolling.offset.y = 0.0;
+                    scrolled_a_layer = true;
+                    continue;
+                },
```
glennw: The comma here and in the section below aren't necessary.

glennw:
@samuknet Ah, I hadn't realised this was already rebased. Let's fix those last couple of minor nits, squash in to one commit, and then r=me.

glennw:
@samuknet Have you got time to rebase this and fix those last couple of nits?

samuknet:
Sure! One issue I was having, is getting Servo to compile once the Webrender version has changed.  I noticed webrender has had changes to its API/type signatures which means I can no longer compile and test Servo with the new scrolling code.  Is there an easy way to check that everything still works together?

gterzian:
@samuknet Have you considered rebasing your branch of the uptream master? It seems a bunch of traits file have indeed changed, so it could help solve the problems you describe. If that isn't enough, you could also try to rebase the servo repository as well...

samuknet:
I notice in `webrender_traits/types.rs` the previous `Scroll` message has now been changed to ` ScrollLayersWithScrollId(Point2D<f32>, PipelineId, ServoScrollRootId)`.

 And that now this messages makes a call to:

```rust
pub fn scroll_layers(&mut self,
                         origin: Point2D<f32>,
                         pipeline_id: PipelineId,
                         scroll_root_id: ServoScrollRootId)

```

Is this effectively the new `fn scroll_layers`, i.e. should I refactor the existing home/end key scrolling code in `fn` scroll into `fn scroll_layers`?

glennw:
The scroll_layers interface is mostly used for scrolling via JS, so it's fine to leave that as is.

glennw:
Thanks!

samuknet:
@glennw Thank you!

PR #541, opened 2016-11-08
Switch AlphaBatchKeyFlags to use bitflags! crate.

moriturus:
Sorry about my mistake on #539 then I re-open this PR.

This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/541)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -432,9 +432,16 @@ impl AlphaBatcher {
                         let needs_blending = transform_kind == TransformedRectKind::Complex ||
                                              !prim_metadata.is_opaque ||
                                              needs_clipping;
-                        let flags = AlphaBatchKeyFlags::new(transform_kind,
-                                                            needs_clipping);
                         let blend_mode = ctx.prim_store.get_blend_mode(needs_blending, prim_metadata);
+                        let needs_clipping_flag = if needs_clipping {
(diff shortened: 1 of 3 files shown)


webrender/src/tiling.rs
```diff
@@ -432,9 +432,13 @@ impl AlphaBatcher {
...
                                              needs_clipping;
-                        let flags = AlphaBatchKeyFlags::new(transform_kind,
-                                                            needs_clipping);
                         let blend_mode = ctx.prim_store.get_blend_mode(needs_blending, prim_metadata);
+                        let needs_clipping_flag = if needs_clipping {
+                            NEEDS_CLIPPING
+                        } else {
+                            AlphaBatchKeyFlags::empty()
+                        };
+                        let flags = AXIS_ALIGNED | needs_clipping_flag;
```
glennw: Sorry, my last comment was not very well described. We still need to check if transform_kind == AxisAligned, and only set the AXIS_ALIGNED flag if that's true. When that's done, the fn transform_kind() function will work correctly, as it will return AxisAligned if the AXIS_ALIGNED bit was set, and Complex otherwise.

glennw:
@moriturus Looks good, thanks! Would you mind squashing these into one commit and then I'll get it merged.

glennw:
Thanks!

PR #543, opened 2016-11-09
Update `ipc-channel` dependency to 0.6

antrik:
Doing a tree-wide update, mostly to make sure nobody accidentally breaks
the new interface, where `ipc_channel::Sender` is no longer `Sync`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/543)


glennw:
@antrik I think you'll need to do a 'cargo build' in the wr-sample and/or replay directories to update the cargo.lock file for those binaries.

After that, looks good to me - once this is merged, is it going to be possible to update Servo straight away, or should I wait until other Servo dependencies have updated before merging this?

antrik:
@glennw good catch: I failed to account for the fact that the package comes with executable programs as well... Fixed now.

(The update command is just the usual `cargo update -p ipc-channel`; no need to build anything -- except for testing purposes, I guess...)

As for Servo, I'll finalise and push my PR once this has landed here.

PR #546, opened 2016-11-09
Change text-shadow to blur entire text run, and clean up texture cache.

glennw:
* Add support for text-shadow to be drawn via render tasks on text runs.
* Add support for blur render tasks. Uses standard 2 pass separable technique.
* Remove a large amount of code that was used only for old glyph blur support.
* Remove blur_radius from GlyphKey, resource cache, texture cache etc.
* Add gaussian blur shader (cs_blur).
* Add text run cache shader (cs_text_run).
* Add generic primitive shader for drawing cache items (ps_cache_image).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/546)

### DIFF
--- a/webrender/res/cs_blur.fs.glsl
+++ b/webrender/res/cs_blur.fs.glsl
@@ -1,0 +1,39 @@
+#line 1
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// TODO(gw): Write a fast path blur that handles smaller blur radii
+//           with a offset / weight uniform table and a constant
+//           loop iteration count!
+
+// TODO(gw): Make use of the bilinear sampling trick to reduce
+//           the number of texture fetches needed for a gaussian blur.
+
+float gauss(float x, float sigma) {
+    return (1.0 / sqrt(6.283185307179586 * sigma * sigma)) * exp(-(x * x) / (2.0 * sigma * sigma));
+}
+
+void main(void) {
+    vec4 color = texture(sCache, vUv) * gauss(0.0, vSigma);
+
+    for (int i=1 ; i < vBlurRadius ; ++i) {
+        vec2 offset = vec2(float(i)) * vOffsetScale;
+
+        vec2 st0 = clamp(vUv.xy + offset, vUvRect.xy, vUvRect.zw);
+        vec4 color0 = texture(sCache, vec3(st0, vUv.z));
+
+        vec2 st1 = clamp(vUv.xy - offset, vUvRect.xy, vUvRect.zw);
+        vec4 color1 = texture(sCache, vec3(st1, vUv.z));
+
+        // Alpha must be premultiplied in order to properly blur the alpha channel.
+        float weight = gauss(float(i), vSigma);
+        color += vec4(color0.rgb * color0.a, color0.a) * weight;
+        color += vec4(color1.rgb * color1.a, color1.a) * weight;
+    }
+
+    // Unpremultiply the alpha.
+    color = vec4(color.rgb / color.a, color.a);
+
+    oFragColor = color;
+}
--- a/webrender/res/cs_blur.fs.glsl
+++ b/webrender/res/cs_blur.fs.glsl
@@ -15,7 +15,8 @@ float gauss(float x, float sigma) {
 }
 
 void main(void) {
-    vec4 color = texture(sCache, vUv) * gauss(0.0, vSigma);
+    vec4 sample = texture(sCache, vUv);
+    vec4 color = vec4(sample.rgb * sample.a, sample.a) * gauss(0.0, vSigma);
 
     for (int i=1 ; i < vBlurRadius ; ++i) {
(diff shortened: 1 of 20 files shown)


glennw:
r? @pcwalton 

I'm sorry that this ended up as one commit. I can separate them into a couple of commits if you would prefer?

webrender/src/prim_store.rs
```diff
@@ -76,7 +71,15 @@ pub struct PrimitiveMetadata {
...
-    pub cache_info: Option<PrimitiveCacheInfo>,
+    // An optional render task that is a dependency of
+    // drawing this primitive. For instance, box shadows
+    // use this to draw a portion of the box shadow to
+    // a render target to reduce the number of pixels
+    // that the box-shadow shader needs to run on. For
+    // text-shadow, this creates a render task chain
+    // that implements a 2-pass separable blur on a
+    // text run.
+    pub render_task: Option<RenderTask>,
```
kvark: I wonder if we should turn it into `Vec<RenderTask>`. Do we ever need more than one task?..

webrender/src/tiling.rs
```diff
@@ -718,23 +836,72 @@ impl RenderTask {
...
+                                                         prim_index);
+
+        let blur_target_size = size + DeviceSize::new(2 * blur_radius.0,
+                                                      2 * blur_radius.0);
+
+        let blur_task_v = RenderTask {
+            id: RenderTaskId::Dynamic(RenderTaskKey::VerticalBlur(blur_radius.0, prim_index)),
+            children: vec![prim_cache_task],
+            location: RenderTaskLocation::Dynamic(None, blur_target_size),
+            kind: RenderTaskKind::VerticalBlur(blur_radius, prim_index),
```
kvark: this seems slightly redundant, as the same info is already stored with the `RenderTaskKey`

webrender/res/cs_blur.fs.glsl
```diff
@@ -0,0 +1,39 @@
...
+
+// TODO(gw): Make use of the bilinear sampling trick to reduce
+//           the number of texture fetches needed for a gaussian blur.
+
+float gauss(float x, float sigma) {
+    return (1.0 / sqrt(6.283185307179586 * sigma * sigma)) * exp(-(x * x) / (2.0 * sigma * sigma));
+}
+
+void main(void) {
+    vec4 color = texture(sCache, vUv) * gauss(0.0, vSigma);
```
glennw: Fixed

webrender/res/cs_blur.fs.glsl
```diff
@@ -0,0 +1,39 @@
...
+        vec4 color1 = texture(sCache, vec3(st1, vUv.z));
+
+        // Alpha must be premultiplied in order to properly blur the alpha channel.
+        float weight = gauss(float(i), vSigma);
+        color += vec4(color0.rgb * color0.a, color0.a) * weight;
+        color += vec4(color1.rgb * color1.a, color1.a) * weight;
+    }
+
+    // Unpremultiply the alpha.
+    color = vec4(color.rgb / color.a, color.a);
```
glennw: Fixed

webrender/res/cs_blur.vs.glsl
```diff
@@ -0,0 +1,52 @@
...
+
+void main(void) {
+    BlurCommand cmd = fetch_blur(gl_InstanceID);
+    RenderTaskData task = fetch_render_task(cmd.task_id);
+    RenderTaskData src_task = fetch_render_task(cmd.src_task_id);
+
+    vec4 local_rect = task.data0;
+
+    vec2 pos = mix(local_rect.xy,
+    			   local_rect.xy + local_rect.zw,
```
glennw: Fixed

webrender/res/cs_blur.vs.glsl
```diff
@@ -0,0 +1,52 @@
...
+    			   aPosition.xy);
+
+    vec2 texture_size = textureSize(sCache, 0).xy;
+    vUv.z = src_task.data1.x;
+    vBlurRadius = int(task.data1.y);
+    vSigma = task.data1.y * 0.5;
+
+    vec4 src_rect;
```
glennw: Fixed

webrender/res/cs_blur.vs.glsl
```diff
@@ -0,0 +1,52 @@
...
+    vec2 uv0, uv1;
+
+    switch (cmd.dir) {
+        case DIR_HORIZONTAL:
+            vOffsetScale = vec2(1.0 / texture_size.x, 0.0);
+            vUvRect = vec4(src_task.data0.xy, src_task.data0.xy + src_task.data0.zw);
```
glennw: Fixed

webrender/src/prim_store.rs
glennw: We might do in the future - it's an easy change if/when we do need it.

webrender/src/tiling.rs
glennw: Yea - it makes it a bit easier to extract the info (since you don't have to match on all the key kinds). But I do agree it's a bit redundant and we might want to consider changing how this works in the future.

glennw:
@kvark Thanks for the review, comments addressed.

Issue #524, opened 2016-11-03
Support external image buffer for add_image()


JerryShih:
Right now, there are some copy in the path from add_image() interface to the final gpu texture uploading in texture_cache module.
The buffer is serialized(so, here is a copy) through the ipc-channel at [1].
If we are playing a video, it's a big performance impact for every frame's copy-op.

In quantum project, I'm trying to pass raw_buffer_ptr through FFI and serialize the raw_ptr using my custom serializer.

The data enum will like:
```
pub enum ExternalImage {
    // shmem or memory buffer
    MemoryBuff {
        buff: *const u8,  // the raw_buffer ptr from c++
        size: usize,
    }
}
```

The serialization/deserialization code is at [2].
With this enum, we could have the new interface "AddExternalImage()" in WR and pass the "buff: *const u8" to gl texture with the minimizing copy.
There will have life-time and sync problem for this raw-buffer. If we can handle the the persistence and thread problem in gecko for this buffer,  how about this idea if we try to use WR in gecko or other project?

```
pub enum ApiMsg {
   ....
  AddExternalImage(ImageKey, u32, u32, ImageFormat, ExternalImage),
}
```

[1]
https://github.com/servo/webrender/blob/58b9e983a5e74ac1670fcf67f9c2ba68740ab2cc/webrender_traits/src/api.rs#L99
[2]
https://gist.github.com/JerryShih/2d4b4738fe0d9d399f90300cdbe6c95d

glennw:
@JerryShih We definitely want / need to improve this path to reduce extra copies of data. I think your idea above is possible. I have some other thoughts on possible API though, mentioned below. 

In your example above, instead of passing a raw pointer to WR, what if you could register a callback with WR though the public API. This callback would have a signature something like:

```
fn WRImageRequest(key: ImageKey) -> &[u8]
```

WR would call this from the compositor thread at the exact instance it needs the image data and then pass that slice of data straight to GL.

This would mean the client managing thread-safe access to the data, but would have the benefit of WR never holding on to a pointer.

One downside of such an API is that we need to work out how to handle device lost situations (e.g. on android), but that's somewhat orthogonal to this discussion.

Thoughts?

jrmuizel:
With WRImageRequest(key: ImageKey) the callee wouldn't have a great way to know how long to keep the data alive for so we should probably at least have a matching WRImageRelease() or some such thing.

glennw:
@jrmuizel Yup, agreed.

[...]

PR #547, opened 2016-11-09
Support external image texture/buffer

JerryShih:
This patch try to support external image in WR (#524).

Then, we could use these two callback function to get/release the external buffer/image from c++.
```
pub get_func: GetExternalImageCallback
pub release_func: ReleaseExternalImageCallback
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/547)

### DIFF
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -743,3 +743,55 @@ pub enum WebGLShaderParameter {
     Bool(bool),
     Invalid,
 }
+
+#[repr(C)]
+pub struct ExternalImageKey(pub u32);
+
+pub enum ExternalImage<'i> {
+    TextureHandle {
+        handle: u32,
+    },
+    MemoryBuff {
+        buff: &'i [u8],
+    }
+}
+
+#[repr(C)]
+pub enum ExternalImageType {
+    TEXTURE_HANDLE,
+    MEM_OR_SHMEM,
+}
+
+#[repr(C)]
+pub struct ExternalImageStruct {
+    image_type: ExternalImageType,
+
+    // external buffer handle
+    handle: u32,
+
+    // shmem or memory buffer
+    buff: *const u8,
+    size: usize,
+}
+
+impl ExternalImageStruct {
+    pub fn get_image(&self) -> ExternalImage {
+        match self.image_type {
+            ExternalImageType::TEXTURE_HANDLE =>
+                ExternalImage::TextureHandle { handle: self.handle },
+            ExternalImageType::MEM_OR_SHMEM =>
+                ExternalImage::MemoryBuff { buff: unsafe { std::slice::from_raw_parts(self.buff, self.size)} },
+        }
+    }
+}
+
+pub type GetExternalImageCallback = fn(ExternalImageKey) -> ExternalImageStruct;
+
+pub type ReleaseExternalImageCallback = fn(ExternalImageKey);
+
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+pub struct ExternalImageCallback {
+    pub get_func: GetExternalImageCallback,
+    pub release_func: ReleaseExternalImageCallback,
+}
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -42,6 +42,7 @@ use time::precise_time_ns;
 use util::TransformedRectKind;
 use webrender_traits::{ColorF, Epoch, PipelineId, RenderNotifier, RenderDispatcher};
 use webrender_traits::{ImageFormat, RenderApiSender, RendererKind};
+use webrender_traits::{ExternalImage, ExternalImageCallback, ExternalImageKey};
 
 pub const BLUR_INFLATION_FACTOR: u32 = 3;
 pub const MAX_RASTER_OP_SIZE: u32 = 2048;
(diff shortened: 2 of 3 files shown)


JerryShih:
@nical @glennw 
How about this for the external image problem?
Then, we could create a new display item type "ExternalImageItem" with the external image key. When the WR tries to render that ExternalImageItem item, WR could use the GetExternalImageCallback() to get the external buffer slice or gl_texture id for painting.

JerryShih:
@glennw 
Could we have a default value for the struct's member?
Like:
```
pub struct RendererOptions {
     ....
     pub renderer_kind: RendererKind,
     pub external_image_callback: Option<ExternalImageCallback> = None,
 }
```

webrender_traits/src/types.rs
```diff
@@ -743,3 +743,55 @@ pub enum WebGLShaderParameter {
...
+}
+
+#[repr(C)]
+pub enum ExternalImageType {
+    TEXTURE_HANDLE,
+    MEM_OR_SHMEM,
+}
+
+#[repr(C)]
+pub struct ExternalImageStruct {
```
nical: How about:

```rust
struct RgbImage {
    size: Size2D<u32>,
    stride: u32,
    format: ImageFormat,
    buffer: *const u8,
}

struct GlHandle {
     handle: u32,
}

// YuvImage, DxgiHandle, etc...

enum ExternalImage {
    RgbMemory(RgbImage),
    YuvMemory(YuvImage),
    Gl(GlHandle),
    Dxgi(DxgiHandle),
    // etc..
}
```

Using enums this way is a lot more idiomatic in rust,  and I think that it'll make it easier to work with internally

glennw:
Similarly to what @nical said, I think having the source data as a rust enum makes sense. If we do this, we won't even need to add a new public API for adding external images - we could change the last parameter of the existing add_image() function from taking a byte array to take a rust enum, something roughly like:

```
pub enum ExternalImageHandle {
    DXGI(dxgi_handle),
    GLTexture(i32),
    Raw(usize),
}

pub enum ImageData {
    Bytes(Vec<u8>),
    External(ExternalImageHandle),
}

fn add_image(&mut self, width: u32, height: u32, format: ImageFormat, data: ImageData) -> ImageKey;
```

And then WR would invoke the callback you've added when it encounters an image where the data is ImageData::External. The callback signature would take the ExternalImageHandle as a parameter.

Thoughts?

JerryShih:
@glennw @nical 

If we have:
```
struct RgbImage {
    size: Size2D<u32>,
    stride: u32,
    format: ImageFormat,
    buffer: *const u8,
}

struct GlHandle {
     handle: u32,
}

// YuvImage, DxgiHandle, etc...

enum ExternalImage {
    RgbMemory(RgbImage),
    YuvMemory(YuvImage),
    Gl(GlHandle),
    Dxgi(DxgiHandle),
    // etc..
}
```

What's the corresponding c++ struct type for rust "repr(C)" ExternalImage enum type?
If we have that, the callback will become clear like:


```
// return a rust enum
pub type GetExternalImageCallback = fn(ExternalImageKey) -> ExternalImage
```

Otherwise, I will still need to have a big temporary struct ExternalImageStruct which contain all fields from all enum types and convert ExternalImageStruct to ExternalImage in binding code like (this PR already has this approach):

```
// get ExternalImageStruct struct instead of ExternalImage enum from c++
pub type GetExternalImageCallback = fn(ExternalImageKey) -> ExternalImageStruct

#[repr(C)]
pub struct ExternalImageStruct {
    image_type: ExternalImageType,

    // external texture handle for gl or dx
    handle: u32,
    target: u32, // 

    // rgb buffer or yuv buffer's size and format
    width: usize,
    height: usize,
    format: u32,

    // rgb or y channel
    buff0: *const u8,
    stride0: usize,
    size0: usize,

    // u channel
    buff1: *const u8,
    stride1: usize,
    size1: usize,

    //v channel
    buff2: *const u8,
    stride2: usize,
    size2: usize,
}

// convert ExternalImageStruct to ExternalImage enum
impl ExternalImageStruct {
    pub fn get_image(&self) -> ExternalImage {
        match self.image_type {
            ExternalImageType::GL_HANDLE =>
                ExternalImage::Gl {....},
            ExternalImageType::RGB =>
                ExternalImage::RgbMemory { .... },
            ExternalImageType::YUB =>
                ExternalImage::YuvMemory { .... },
            .........
        }
    }
}
```

nical:
@JerryShih and I talked about this offline and we think that as a first step it would be good to simplify this down to only exposing TextureId in the callbacks (we'd do the eventual texture uploads / dxgi glue, etc in gecko for now). This would simplify the callback interface into something that easy to express in both C++ and rust and avoid polluting the API with structures that would be better expressed as rust enums. Also, it would give us more time to figure out a more elaborate API to deal with the external image interface.

We should have a vidyo chat about this to figure out what servo and gecko need in the longer term.

nical:
@glennw I don't think we can / want to have the handles (GL/DXGI/etc.) stored in the display list. The thread that produces the display list does not know the value of this handle, which will be created on the compositor/renderer thread later. The idea behind ExternalImageKey is to have something that we can refer to (when creating the display list and in the render backend) before we create the actual handles (on the compositor/render thread). It also has the benefit of letting us change the texture handle associated to a key without rebuilding the frame if only the content of the current image has changed, which is pretty common when watching videos).

glennw:
@JerryShih @nical Should we close this PR? The basic external native texture support has landed now.

JerryShih:
Thanks. I'm trying to use that interface in Gecko.

PR #549, opened 2016-11-09
DirectWrite font rendering for Windows

vvuk:
This PR adds DirectWrite font rendering for Windows, using a thin native-rust dwrite wrapper library.  There's a corresponding PR for Servo that adds support for DirectWrite to their font backend as well.

Things work, though right now there is no support for web fonts.  This isn't hard to add and should be straightforward; @metajack already has rust code for implementing the required COM interfaces to do so, and this just needs to be done in https://github.com/vvuk/dwrote-rs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/549)


webrender/src/lib.rs
```diff
@@ -95,9 +101,12 @@ extern crate core_graphics;
 #[cfg(target_os="macos")]
 extern crate core_text;
 
-#[cfg(not(target_os="macos"))]
+#[cfg(any(target_os="linux", target_os="android"))]
```
glennw: This should be all(unix, not(target_os = "macos")) I think, since WR builds and runs on *BSD now.

webrender/src/platform/windows/font.rs
```diff
@@ -0,0 +1,147 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use app_units::Au;
+use std::collections::HashMap;
+use webrender_traits::{FontKey, FontRenderMode, GlyphDimensions};
+
+use dwrote;
+
+pub struct FontContext {
+    // FIXME do I need to protect this in a mutex?
```
glennw: No - FontContexts are created per-thread and stored in TLS.

webrender/src/platform/windows/font.rs
```diff
@@ -0,0 +1,147 @@
...
+            main_display_rendering_params: dwrote::RenderingParams::create_for_primary_monitor(),
+        }
+    }
+
+    pub fn add_raw_font(&mut self, font_key: &FontKey, _: &[u8]) {
+        if self.fonts.contains_key(font_key) {
+            return
+        }
+
+        panic!("DWrite can't do raw fonts yet");
```
glennw: Does this break a lot of websites? Should we consider making it non-fatal for now?

webrender_traits/src/types.rs
```diff
@@ -371,10 +372,13 @@ pub enum MixBlendMode {
 pub type NativeFontHandle = CGFont;
 
 /// Native fonts are not used on Linux; all fonts are raw.
-#[cfg(not(target_os = "macos"))]
-#[cfg_attr(not(target_os = "macos"), derive(Clone, Serialize, Deserialize))]
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
glennw: Same here as the above comment about *BSD.

webrender/src/platform/windows/font.rs
```diff
@@ -0,0 +1,147 @@
...
+        };
+
+        // if we weren't asked to rasterize, we're done
+        if render_mode.is_none() {
+            return (Some(dims), None);
+        }
+
+        // don't rasterize blank glyphs
+        if width_u == 0 || height_u == 0 {
+            return (Some(dims), Some(RasterizedGlyph {
```
glennw: WR expects the font platform to return None when the glyph has zero dimensions.

webrender_traits/src/types.rs
```diff
@@ -371,10 +372,13 @@ pub enum MixBlendMode {
 pub type NativeFontHandle = CGFont;
 
 /// Native fonts are not used on Linux; all fonts are raw.
-#[cfg(not(target_os = "macos"))]
-#[cfg_attr(not(target_os = "macos"), derive(Clone, Serialize, Deserialize))]
+#[cfg(all(unix, not(target_os="macos")))]
```
glennw: @vvuk This is resulting in Clone *not* being derived on Linux, causing a compile error. I'm not sure why - it looks correct to me. I guess we could try make it not(windows or mac) or something like that?

nox: I don't understand what that change is supposed to do, but please do try not(windows or mac).

vvuk: (@nox - @glennw left out a second added line which was the cfg_attr with the same conditional as the cfg)

Yeah, this is definitely a serde bug.  In the generated `types.rs`, this is expanding out to:
```
#[cfg(target_os = "macos")]
pub type NativeFontHandle = CGFont;
#[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
#[cfg(all(unix, not(target_os = "macos")))]
const _IMPL_DESERIALIZE_FOR_NativeFontHandle: () =
...
#[cfg(target_os = "macos")]
pub type NativeFontHandle = CGFont;
#[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
#[cfg(all(unix, not(target_os = "macos")))]
const _IMPL_DESERIALIZE_FOR_NativeFontHandle: () =
    {
...
    };
/// Native fonts are not used on Linux; all fonts are raw.
#[cfg(all(unix, not(target_os = "macos")))]
pub struct NativeFontHandle;
```
Note that all mention of `derive(Clone)` was lost.  Even if I add a second `cfg_attr` with just `derive(Clone)` it gets eaten as well.  Changing the conditional does not change the behaviour.  Forcing serde to 0.8.16 doesn't help.  Hrm.

vvuk:
I pushed a new commit by itself that should fix this...  hopefully someone from serde can look at it and see what's up with the other version.

PR #552, opened 2016-11-10
Flatten the display list in the API

mrobinson:
This changes brings the WebRender API more into line with the Servo
display list and also the frames generated inside WebRender itself.
Most intermediate display list representations are removed as well as
their place on the resource cache. Perhaps most notably, stacking
contexts are started via a PushStackingContext display item and ended
via a PopStackingContext item. Additionally, the flatten pass has been
significantly simplified to account for these changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/552)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -12,13 +12,13 @@ use internal_types::{LowLevelFilterOp};
 use internal_types::{RendererFrame};
 use layer::{Layer, ScrollingState};
 use resource_cache::{DummyResources, ResourceCache};
-use scene::{SceneStackingContext, ScenePipeline, Scene, SceneItem, SpecificSceneItem};
+use scene::Scene;
 use std::collections::{HashMap, HashSet};
 use std::hash::BuildHasherDefault;
 use tiling::{AuxiliaryListsMap, FrameBuilder, FrameBuilderConfig, PrimitiveFlags};
 use util::MatrixHelpers;
 use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLayerId};
-use webrender_traits::{ClipRegion, ColorF, StackingContext, FilterOp, MixBlendMode};
+use webrender_traits::{ClipRegion, ColorF, DisplayItem, StackingContext, FilterOp, MixBlendMode};
 use webrender_traits::{ScrollEventPhase, ScrollLayerInfo, SpecificDisplayItem, ScrollLayerState};
 
 #[cfg(target_os = "macos")]
@@ -33,21 +33,9 @@ pub struct FrameId(pub u32);
 static DEFAULT_SCROLLBAR_COLOR: ColorF = ColorF { r: 0.3, g: 0.3, b: 0.3, a: 0.6 };
 
 struct FlattenContext<'a> {
-    resource_cache: &'a mut ResourceCache,
     scene: &'a Scene,
     pipeline_sizes: &'a mut HashMap<PipelineId, Size2D<f32>>,
     builder: &'a mut FrameBuilder,
-}
-
-#[derive(Debug)]
-struct FlattenInfo {
-    // Pipeline this part of the tree belongs to.
-    pipeline_id: PipelineId,
-
-    current_scroll_layer_id: ScrollLayerId,
-    current_fixed_layer_id: ScrollLayerId,
-
-    layer_relative_transform: Matrix4D<f32>,
 }
 
 pub type LayerMap = HashMap<ScrollLayerId, Layer, BuildHasherDefault<FnvHasher>>;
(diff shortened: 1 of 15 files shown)


mrobinson:
@glennw r?

mrobinson:
The Servo part of this patch is here, which can be ready to go soon after this lands: https://github.com/mrobinson/servo/tree/display-list

glennw:
@mrobinson Wow, this is awesome! Will review today :)

emilio:
@mrobinson I think you'll need to update the lockfiles, and potentially the callers in the other crates I guess, Travis is complaining :(

glennw:
@mrobinson I've had a reasonable look over this - it looks great! I ran all the WPT tests on it locally, and I get the failures below. Once those are fixed up we'll try to finish up the review and get it merged before it bitrots too badly.

```

  ▶ FAIL [expected PASS] /_mozilla/css/iframe/hide_and_show.html
  └   → /_mozilla/css/iframe/hide_and_show.html 1b322aa260e8efcc96823b3616dec4048e764fd7
/_mozilla/css/iframe/hide_and_show_ref.html a588759663fe2c5184aba85f7ea89f48a5c2b38f
Testing 1b322aa260e8efcc96823b3616dec4048e764fd7 == a588759663fe2c5184aba85f7ea89f48a5c2b38f

  ▶ FAIL [expected PASS] /_mozilla/css/stacked_layers.html
  └   → /_mozilla/css/stacked_layers.html 47267c953cbf670c595a258334dff478f5c24369
/_mozilla/css/stacked_layers_ref.html 6955d0a11b580828a5e9a47926e22b577c48ee5f
Testing 47267c953cbf670c595a258334dff478f5c24369 == 6955d0a11b580828a5e9a47926e22b577c48ee5f
```

mrobinson:
@glennw Ack! Thanks for taking a look. I'll try to get the ref tests sorted today. Sorry for letting those slip by.

mrobinson:
@glennw Okay. Sorry for the failures. Things should be working now.

webrender/src/frame.rs
```diff
@@ -66,45 +54,21 @@ pub struct Frame {
...
+            Some(item) => item,
+            None => return None,
         };
 
-        for display_list_id in &stacking_context.display_lists {
-            let display_list = &scene.display_list_map[display_list_id];
-            for item in &display_list.items {
-                result.push(item.clone());
-            }
+        match first_item.item {
```
kvark: nit: could combine these matches into one

webrender/src/frame.rs
```diff
@@ -190,6 +154,57 @@ impl StackingContextHelpers for StackingContext {
     }
 }
 
+pub struct DisplayListTraversal<'a> {
+    pub display_list: &'a Vec<DisplayItem>,
```
kvark: do these have to be public?

webrender/src/frame.rs
```diff
@@ -190,6 +154,57 @@ impl StackingContextHelpers for StackingContext {
...
+        for item in self {
+            if item.item == SpecificDisplayItem::PopStackingContext {
+                return;
+            }
+        }
+    }
+
+    pub fn current_stacking_context_empty(&self) -> bool {
+        match self.peek() {
+            Some(item) if item.item == SpecificDisplayItem::PopStackingContext => true,
```
kvark: nit: could do `Some(item) => item.item == SpecificDisplayItem::PopStackingContext,`

webrender/src/frame.rs
```diff
@@ -757,6 +633,147 @@ impl Frame {
...
+        let iframe_fixed_layer_id = ScrollLayerId::create_fixed(pipeline_id);
+        self.layers.insert(iframe_fixed_layer_id,
+                           Layer::new(&iframe_rect,
+                                      iframe_stacking_context.overflow.size,
+                                      &transform,
+                                      pipeline_id));
+
+        let iframe_scroll_layer_id = iframe_stacking_context.scroll_layer_id.unwrap();
+        self.layers.insert(iframe_scroll_layer_id,
+                           Layer::new(&iframe_rect,
```
kvark: nit: looks like the same layer as above, could `clone()` it

webrender/src/scene.rs
```diff
@@ -4,25 +4,37 @@
...
+use webrender_traits::{AuxiliaryLists, BuiltDisplayList, PipelineId, Epoch, ColorF};
+use webrender_traits::{DisplayItem, SpecificDisplayItem, StackingContext};
+
+trait DisplayListHelpers {
+    fn starting_stacking_context<'a>(&'a self) -> Option<&'a StackingContext>;
+}
+
+impl DisplayListHelpers for Vec<DisplayItem> {
+    fn starting_stacking_context<'a>(&'a self) -> Option<&'a StackingContext> {
+        let first_item = match self.first() {
```
kvark: nit: same as the other implementation, could combine the two matches into one:
```
Some(SomethingSomething { item: SpecificDisplayItem::PushStackingContext(ref item), .. }) => ...
```

webrender/src/frame.rs
emilio: As a uber-nit: this can probably be `&'a [DisplayItem]` to save an extra level of indirection.

mrobinson: Okay!

webrender/src/frame.rs
mrobinson: Okay. I've made Layer derive from Clone and done this. It made things quite a bit simpler.

webrender/src/frame.rs
kvark: That's really minor and doesn't need to be fixed! Could be done in a single match (like I specified in another comment), in nested matches, or even with `and_then` construct:
```
self.first().and_then(|item| match item.item {...})
```

glennw:
@mrobinson r=me. I haven't merged this yet to minimize the amount of time that we have this merged in WR without the corresponding Servo PR. But feel free to merge this when the Servo PR is ready to go!

mrobinson:
@glennw Thanks. I'll get the Servo bit up ASAP.

PR #553, opened 2016-11-10
Revert "Update `ipc-channel` dependency to 0.6"

glennw:
This reverts commit 7c4f0af092c6b3527b045e7d03be841b799b2909.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/553)


glennw:
This is a temporary revert until a bug fix for ipc-channel is made on mac (until that happens we can't land WR updates in Servo without this revert).

PR #554, opened 2016-11-11
Introduce SourceTexture, an abstraction for location of source textures.

glennw:
This allows the backend thread to work with textures that have
been allocation by the texture cache, via webgl or an external
source transparently. The native texture ID is resolved by the
rendering thread as requited.

Specifically:

* Add SourceTexture, an abstraction for the location of a source texture.
* Add CacheTextureID, an ID for a texture cache managed texture.
    * This removes the need to allocate a pool of textures up front.
* Remove unused batch.rs file.
* Remove the concept of "raw textures" from device.rs
    * The native texture ID is stored inside the SourceTexture enum now.
* Remove some unused functions in device.rs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/554)


glennw:
I figured I'd prototype adding the source texture -> resolving idea I had today, and it worked out quite well I think. Thoughts @nical @kvark @JerryShih ?

This removes the use of TextureId from the backend thread, using SourceTexture instead, which is resolved to TextureId in the rendering thread. I haven't implemented the external texture support, but I've left a couple of TODO notes where I would expect it to fit in.

This also doesn't handle the idea of getting UV rects in the renderer thread, but that can be sorted out when we do the actual external image support. Does this look reasonable to you guys?

glennw:
This is basically taking @nical 's external texture branch, and extending it to handle webgl textures, and fixing up the texture cache ID pool at the same time.

glennw:
@nical @JerryShih @kvark I have pushed a WIP follow up commit to this https://github.com/servo/webrender/commit/3fe9214ec6287c7de4ec2572f3b3c58425760478 here.

This commit is a messy, untested, but mostly complete implementation of how I think external images + custom UV rects could work.

The first part of the idea is that primitives no longer store their UV rect directly - there is a GpuDataStore for all UV rectangles, just like there is for many other GPU types. This means that vertex shaders fetch their UV rect from here, and primitives allocate a slot inside this array when they are first added.

Next, the add_image() API gets extended to support adding different types of images (raw bytes, or external, with some arbitrary key).

Now, when we resolve images from the texture cache to get their UV rect, if it finds an external texture, instead of storing a UV rect directly, it creates a DeferredResolve item, noting the external source texture and the index in the UV rects array where the UV rect actually belongs.

Finally, that array of DeferredResolves are passed to the render thread with each frame. The first thing the renderer does is iterate that list, invoking the user callback to get the uv rect and / or texture id. Once that is done, it can patch the resource rects array, just before it uploads to the GPU.

The part I haven't implemented is the actual callbacks (there are two panics where the callbacks would be invoked).

Apart from that, I think the ideal should work. Any thoughts on whether this seems like a reasonable API / implementation from your perspective?

glennw:
The nice thing about this approach is that the render thread doesn't really have to know much more about the backend thread than it already does. The only real new functionality on the render thread is to run through a list of deferred resolve structures, run a callback, and patch an array with the result.

nical:
Looks good! (both this PR and the draft deferred UV implementation).

glennw:
@kvark Thanks for the review, comments addressed and pushed.

r? @pcwalton

webrender/src/renderer.rs
```diff
@@ -841,10 +866,25 @@ impl Renderer {
...
+                        let CacheTextureId(cache_texture_index) = update.id;
+                        if self.cache_texture_id_map.len() == cache_texture_index {
+                            // It was a new texture, so add to end of the map.
+                            self.cache_texture_id_map.push(Some(texture_id));
+                        } else {
+                            // It was re-using an item from the free-list, so store
+                            // the new ID there.
+                            debug_assert!(self.cache_texture_id_map[cache_texture_index].is_none());
+                            self.cache_texture_id_map[cache_texture_index] = Some(texture_id);
+                        }
```
pcwalton: This could use `Vec::grow_set()` to make this a one-liner if you wanted, but up to you

glennw: It looks like grow_set was deprecated (https://github.com/rust-lang/rust/pull/17029) - it doesn't seem to be available in stable rust.

PR #556, opened 2016-11-11
Off-screen clip mask generation

kvark:
This is the second major step towards #498
Edit: now actually includes the third step as well (removal of `*_clip` shaders).

The PR makes all the clip masks to be generated via the cached rendering tasks (of the new kind). These task work on the area of intersection between all the clip items. The resulting draw calls are being batched via the new `ClipBatcher`.

It supports everything that we currently support, plus the actual handling of arbitrary number of clips.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/556)


Cargo.lock
```diff
@@ -314,7 +314,7 @@ dependencies = [
 
 [[package]]
 name = "ipc-channel"
-version = "0.6.0"
+version = "0.5.1"
```
frewsxcv: Seems unfortunate this is downgraded

glennw: There's a bug in ipc-channel 0.6 preventing us from landing the update in Servo, so we had to revert the update in WR (for now).

webrender/src/clip_stack.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+#[derive(Debug)]
+pub struct MaskCacheInfo {
+    pub key: MaskCacheKey,
+    // this is needed to update the ImageMaskData after the
+    // ResourceCache allocates/load the actual data
+    // will be simplified after the TextureCache upgrade
+    pub image: Option<ImageMask>,
+    pub device_rect: DeviceRect,
```
kvark: One problem that I see with this change is the computation of `device_rect`. Since layers are supposed to move, this rectangle needs to be re-computed, but there is not tracking of the layers at this point. Is there an example where the layers state is being cached and updated upon any movement? Alternatively, I can re-compute the rectangle every frame.

Cargo.lock
glennw: The revert to 0.5.1 is landed in WR master, so if you rebase all the changes related to ipc-channel should disappear.

webrender/Cargo.toml
```diff
@@ -18,7 +18,7 @@ byteorder = "0.5"
 euclid = "0.10"
 fnv="1.0"
 gleam = "0.2"
-ipc-channel = "0.6"
+ipc-channel = "0.5"
```
glennw: Remove

webrender/res/clip_shared.glsl
```diff
@@ -3,63 +3,55 @@
...
-float do_clip(vec2 pos) {
-    vec2 ref_tl = vClipRect.xy + vec2( vClipRadius.x,  vClipRadius.x);
-    vec2 ref_tr = vClipRect.zy + vec2(-vClipRadius.y,  vClipRadius.y);
-    vec2 ref_br = vClipRect.zw + vec2(-vClipRadius.z, -vClipRadius.z);
-    vec2 ref_bl = vClipRect.xw + vec2( vClipRadius.w, -vClipRadius.w);
+struct CacheClipInstance {
+    int render_task_index;
+    int layer_index;
+    int data_index;
+    int pad;
```
glennw: The pad field can be removed since this is a VS only structure.

webrender/res/cs_clip_clear.fs.glsl
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+void main(void) {
+    oFragColor = vec4(1, 1, 1, 1);
```
glennw: 1 -> 1.0 for strict GLESv3 compilers.

webrender/src/clip_stack.rs
```diff
@@ -0,0 +1,178 @@
...
+use prim_store::{ClipData, GpuBlock32, PrimitiveClipSource, PrimitiveStore};
+use prim_store::{CLIP_DATA_GPU_SIZE, MASK_DATA_GPU_SIZE};
+use tiling::StackingContextIndex;
+use util::TransformedRect;
+use webrender_traits::{AuxiliaryLists, ImageMask};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct ClipAddressRange {
+    pub start: GpuStoreAddress, // start GPU address
+    pub count: u32, // number of items, not bytes
```
glennw: Perhaps rename to item_count?

webrender/src/render_backend.rs
```diff
@@ -36,7 +36,7 @@ pub struct RenderBackend {
     next_namespace_id: IdNamespace,
 
     resource_cache: ResourceCache,
-    dummy_resources: DummyResources,
+    _dummy_resources: DummyResources,
```
glennw: Can this be removed completely?

webrender/src/renderer.rs
```diff
@@ -1047,18 +1092,61 @@ impl Renderer {
...
                                 shader,
                                 1,
                                 &BatchTextures::no_texture(),
-                                max_cache_instances,
+                                max_prim_items,
                                 &projection);
         }
 
+        // Draw the clip items into the tiled alpha mask.
+        if true {
```
glennw: Remove condition

webrender/src/tiling.rs
```diff
@@ -567,11 +645,14 @@ struct CompileTileContext<'a> {
 struct RenderTargetContext<'a> {
     layer_store: &'a [StackingContext],
     prim_store: &'a PrimitiveStore,
+    resource_cache: &'a ResourceCache,
+    _clip_stack: &'a ClipRegionStack,
```
glennw: Unused?

glennw:
@kvark What was the reason for having to cover the entire tile with each clip primitive again? I think we'll probably need to find a solution that doesn't require this - I think the performance overhead will probably be too high otherwise. I'll pull the branch today and do some performance tests though to check that.

kvark:
The name became slightly misleading. The clip tile is not the tile a clip belongs to, it's actually just representing the screen space boundary of thr intersection between all clip instances, so the performance is not expected to suffer.

webrender/res/clip_shared.glsl
```diff
@@ -3,63 +3,55 @@
...
+    cci.pad = 0;
 
-    bvec4 is_out = bvec4(pos.x < ref_tl.x && pos.y < ref_tl.y,
-                         pos.x > ref_tr.x && pos.y < ref_tr.y,
-                         pos.x > ref_br.x && pos.y > ref_br.y,
-                         pos.x < ref_bl.x && pos.y > ref_bl.y);
+    return cci;
+}
+
+// The transformed vertex function that always covers the whole tile with the primitive
```
glennw: Let's specify that the whole tile means the intersection of the clip instances here.

glennw:
Doing some testing with this today, and ran into something which I assume is a bug - unless I'm misunderstanding something:

I have a test case that contains a single div, with the follow style:

```
  position: absolute;
  top: 100px;
  left: 100px;
  width: 100px;
  height: 100px;
  background: red;
  border-radius: 32px;
```
With some logging so show what the renderer is submitting for clip cache draws, I see:

```
draw_target Some((TextureId { name: 1033, target: 35866 }, 0))
clear [CacheClipInstance { task_id: 12, layer_index: 0, address: GpuStoreAddress(0), pad: 0 }, 
CacheClipInstance { task_id: 13, layer_index: 0, address: GpuStoreAddress(0), pad: 0 }]
rect [CacheClipInstance { task_id: 13, layer_index: 0, address: GpuStoreAddress(0), pad: 0 }]

draw_target None
clear [CacheClipInstance { task_id: 14, layer_index: 0, address: GpuStoreAddress(0), pad: 0 }]
```

Which suggests that it's issuing two clear draw calls for the cache render render, and also a clear on the main render target. I was expecting to see 1 clear + 1 rect in the render target, and no clip items in the main frame buffer target.

(I ensured the div is positioned completely inside one tile, just to rule out any bugs related to crossing tile boundaries).

glennw:
Ah, those extra clears are related to the dummy mask item - I guess that always gets created at the moment, even if not needed?

kvark:
Yeah, and it's also incomplete/buggy. I've been looking into fixing it, hopefully done by tomorrow. Note: it should be just one draw, I'll have a closer look tomorrow about the rest.

webrender/res/prim_shared.glsl
```diff
@@ -49,7 +52,6 @@ uniform sampler2DArray sCache;
 uniform sampler2D sLayers;
 uniform sampler2D sRenderTasks;
 uniform sampler2D sPrimGeometry;
-uniform sampler2D sClips;
```
glennw: We could also remove Clips from the TextureSampler enum, and associated code.

webrender/res/prim_shared.glsl
```diff
@@ -695,4 +636,12 @@ vec2 init_transform_fs(vec3 local_pos, vec4 local_rect, out float fragment_alpha
 
     return pos;
 }
-#endif
+
+float do_clip() {
+    // anything outside of the mask is considered transparent
+    bvec4 inside = lessThanEqual(
```
glennw: It's unfortunate to add the cost of this conditional to every shader - is it actually needed? Shouldn't we always be able to just sample from the clip mask?

webrender/res/cs_clip_clear.vs.glsl
```diff
@@ -0,0 +1,14 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+void main(void) {
+    CacheClipInstance cci = fetch_clip_item(gl_InstanceID);
+    Tile tile = fetch_tile(cci.render_task_index);
+
+    vec2 final_pos = tile.screen_origin_task_origin.zw +
+                     tile.size_target_index.xy * aPosition.xy;
+
+    gl_Position = uTransform * vec4(final_pos, 0, 1);
```
glennw: Use float constants

webrender/res/cs_clip_image.fs.glsl
```diff
@@ -0,0 +1,16 @@
...
+    float alpha = 1.f;
+    vec2 local_pos = init_transform_fs(vPos, vLocalRect, alpha);
+
+    bool repeat_mask = false; //TODO
+    vec2 clamped_mask_uv = repeat_mask ? fract(vClipMaskUv.xy) :
+        clamp(vClipMaskUv.xy, vec2(0.0, 0.0), vec2(1.0, 1.0));
+    vec2 source_uv = clamped_mask_uv * vClipMaskUvRect.zw + vClipMaskUvRect.xy;
+    float clip_alpha = texture(sMask, source_uv).r; //careful: texture has type A8
+
+    oFragColor = vec4(1, 1, 1, min(alpha, clip_alpha));
```
glennw: Use float constants

webrender/res/cs_clip_rectangle.fs.glsl
```diff
@@ -0,0 +1,44 @@
...
+}
+
+
+void main(void) {
+    float alpha = 1.f;
+    vec2 local_pos = init_transform_fs(vPos, vLocalRect, alpha);
+
+    float clip_alpha = rounded_rect(local_pos);
+
+    oFragColor = vec4(1, 1, 1, min(alpha, clip_alpha));
```
glennw: Use float constants

webrender/res/ps_gradient_clip.fs.glsl
```diff
@@ -11,6 +11,7 @@ void main(void) {
     vec2 local_pos = vPos;
 #endif
 
-    alpha = min(alpha, do_clip(local_pos));
+    //alpha = min(alpha, do_clip(local_pos));
+    alpha = min(alpha, do_clip());
     oFragColor = vColor * vec4(1, 1, 1, alpha);
```
glennw: Use float constants

webrender/src/clip_stack.rs
```diff
@@ -0,0 +1,178 @@
...
+
+struct LayerInfo {
+    world_transform: Matrix4D<f32>,
+    _transformed_rect: TransformedRect,
+    local_rect: Rect<f32>,
+    parent_id: Option<StackingContextIndex>,
+}
+
+pub struct ClipRegionStack {
+    layers: HashMap<StackingContextIndex, LayerInfo>,
```
glennw: It would be good if we could get rid of this layers hash map - the LayerInfo struct doesn't really seem to store anything that's not already stored in the main Layer Store. Could we instead just pass in a borrow of the Layer Store when it's needed in generate(), including the current layer id? That way, we guarantee that there is one source of truth for the stacking context transform, that includes scroll offsets etc.

webrender/src/clip_stack.rs
```diff
@@ -0,0 +1,178 @@
...
+    image_masks: Vec<ImageMask>,
```
glennw: image_masks seems to be unused - remove?

webrender/src/gpu_store.rs
```diff
@@ -5,7 +5,7 @@
 use renderer::MAX_VERTEX_TEXTURE_WIDTH;
 use std::mem;
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, Eq, Hash, PartialEq)]
```
glennw: I'm unsure about this - maybe it's necessary but it just doesn't seem quite right. Perhaps in the cache key we could store the ImageKey instead of the actual GpuStoreAddress?

webrender/src/clip_stack.rs
```diff
@@ -0,0 +1,178 @@
...
+            Some(lid) => lid,
+            None => return None,
+        });
+        let layer = &self.layers[&clip_key.layer_id];
+        let mut local_rect = Some(layer.local_rect);
+
+        let image = match source {
+            &PrimitiveClipSource::NoClip => None,
+            &PrimitiveClipSource::Complex(rect, radius) => {
+                let address = clip_store.alloc(CLIP_DATA_GPU_SIZE);
```
glennw: Ideally we don't want to be calling alloc() on gpu stores during generate. The reason is that the gpu stores are not re-created during scrolling, so this will end up growing the gpu stores during scrolling. What most of the primitives do is a gpu store alloc when the prim is first added (if needed), and then just get a (mutable) reference to that address when they need to update that data. I think that would work OK in this case?

webrender/src/clip_stack.rs
glennw: Doesn't this get re-computed every frame anyway, since the generate() call is done during cull_layers()?

kvark: generate() is currently only done `if metadata.clip_cache_info.is_none()`

webrender/src/tiling.rs
kvark: removed now

webrender/res/clip_shared.glsl
kvark: reworked the whole tile confusion, at last

webrender/res/prim_shared.glsl
kvark: We'd have to clamp to UV borders, and even then - we don't currently guarantee that those borders contain opaque data. So - possible, but non-trivial.
I don't see it as a big deal though, just a few comparisons with masking the fetch result.

webrender/src/clip_stack.rs
kvark: done

webrender/src/gpu_store.rs
kvark: we may be able to store `ImageKey` instead of the gpu address in the cache key, but there is also `ClipAddressRange` in there that uses `GpuStoreAddress`

webrender/src/clip_stack.rs
kvark: done

kvark:
@glennw thank you for the fantastic review!
I've addressed most of your concerns (and hope to chat about the rest few). The `ClipRegionStack` is no more. You can see the code is much more even now (+889 −676 loc), and layers scrolling should be properly supported (running the tests ATM, unsure if it's covered though).
The shader side of things is also clear now - with the distinct `ClipArea` struct used instead of piggy-backing on the `Tile`. The dummy task renders nothing, is only added on the first opaque primitive for the tile, and works correctly.

kvark:
Fixed the dummy task now some more and confirmed the mozilla test suite passes for servo.

glennw:
@kvark Thanks! I'll take another look at this today.

I did some testing, and in general the performance seems quite good. A good page to do some testing with is https://github.com/servo/servo - the performance goes off a cliff as you scroll around halfway down the page. The GPU profiler shows that it is spending the majority of its time generating clip masks, so that would be worth investigating. There also seems to be some clipping artifacts when scrolling on some of the buttons.

It's possible those slowdowns / artifacts are unrelated to this change, but I haven't noticed them previously.

webrender/src/tiling.rs
```diff
@@ -156,8 +157,13 @@ impl AlphaBatchHelpers for PrimitiveStore {
             PrimitiveKind::TextRun |
             PrimitiveKind::Image |
             PrimitiveKind::Gradient |
-            PrimitiveKind::BoxShadow => true,
-
+            PrimitiveKind::BoxShadow => {
```
glennw: Let's do this test outside the match statement, so it handles borders too.

webrender/src/tiling.rs
```diff
@@ -1601,13 +1719,24 @@ impl ScreenTile {
...
                             current_task.as_alpha_batch().items.clear();
                         }
                     }
 
+                    // Add a task to render the updated image mask
+                    if let Some(ref clip_info) = prim_metadata.clip_cache_info {
+                        if let Some(mask_task) = RenderTask::new_mask(self.rect, clip_info) {
+                            current_task.children.push(mask_task);
+                        } else {
+                            // The primitive has clip items but their intersection is empty
```
glennw: Ideally we'd catch this earlier, and not even mark the primitive as visible in this case. But we can do that as a follow up (let's just add a TODO comment here about it).

kvark:
@glennw I checked the servo page and haven't found anything outstanding in terms of what the implementation is doing. There is quite a large of an area associated with the clips (~80% screen space), so that's where the slowdown can come from. Frankly, I do not observe the "going off the cliff" effect, the FPS drops from 55 to 51 when scrolled to the middle of the document.

Given that the rounded corner radius is typically about 2-3px, up to 10, and the rectangles are pretty big, I can think of a way to drastically reduce the time spent on clip generation:
  1. make sure the target is cleared to 1.0 (opaque). This may require a separate target from the general RT cache since that one is cleared (intentionally) to 0.0.
  2. draw only 4 corners instead of the whole thing.

I'm going to rebase and resolve the last concerns.

PR #560, opened 2016-11-15
Add a flush notification api

changm:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/560)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -17,7 +17,7 @@ use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Sender;
 use texture_cache::TextureCache;
 use webrender_traits::{ApiMsg, AuxiliaryLists, BuiltDisplayList, IdNamespace};
-use webrender_traits::{RenderNotifier, RenderDispatcher, WebGLCommand, WebGLContextId};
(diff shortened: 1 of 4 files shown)


glennw:
@changm Let's add a doc comment to the public API to this, saying that it's likely to be removed in the future and that it shouldn't be relied on to stay as an API. Then this should be fine to merge.

webrender/src/render_backend.rs
```diff
@@ -132,6 +136,10 @@ impl RenderBackend {
                                                                    format,
                                                                    bytes);
                         }
+                        ApiMsg::Flush => {
+                            let mut flush_notifier = self.flush_notifier.lock();
+                            flush_notifier.as_mut().unwrap().as_mut().unwrap().all_messages_flushed();
```
kvark: oh wow

kvark: @changm I see, thanks for the explanation.

> Finally, the flush here is to flush the ApiMsg pipeline.

I understand how you use it and how it works. I'm saying there is no actual "flush" action associated with it, so it might be called something more general. But it doesn't matter since the change is temporary anyway.

changm:
@glennw Alright, hows the comment look?

PR #561, opened 2016-11-15
Add an API for providing external images.

glennw:
This allows applications to provide images without supplying
the image bytes up front. This can be useful if the application
doesn't have the bytes available immediately (e.g. video decoding)
or wants to manage the buffer allocation and reduce memory
copies.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/561)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -57,6 +57,7 @@ uniform sampler2D sData16;
 uniform sampler2D sData32;
 uniform sampler2D sData64;
 uniform sampler2D sData128;
+uniform sampler2D sResourceRects;
 
 ivec2 get_fetch_uv(int index, int vecs_per_item) {
     int items_per_row = WR_MAX_VERTEX_TEXTURE_WIDTH / vecs_per_item;
@@ -208,16 +209,14 @@ GradientStop fetch_gradient_stop(int index) {
 
 struct Glyph {
     vec4 offset;
-    vec4 uv_rect;
 };
 
 Glyph fetch_glyph(int index) {
     Glyph glyph;
 
-    ivec2 uv = get_fetch_uv_2(index);
+    ivec2 uv = get_fetch_uv_1(index);
 
-    glyph.offset = texelFetchOffset(sData32, uv, 0, ivec2(0, 0));
-    glyph.uv_rect = texelFetchOffset(sData32, uv, 0, ivec2(1, 0));
+    glyph.offset = texelFetchOffset(sData16, uv, 0, ivec2(0, 0));
 
     return glyph;
 }
@@ -324,19 +323,22 @@ struct CachePrimitiveInstance {
     int specific_prim_index;
     int render_task_index;
     int sub_index;
+    ivec4 user_data;
 };
 
 CachePrimitiveInstance fetch_cache_instance(int index) {
     CachePrimitiveInstance cpi;
 
-    int offset = index * 1;
+    int offset = index * 2;
 
     ivec4 data0 = int_data[offset + 0];
+    ivec4 data1 = int_data[offset + 1];
 
     cpi.global_prim_index = data0.x;
     cpi.specific_prim_index = data0.y;
     cpi.render_task_index = data0.z;
     cpi.sub_index = data0.w;
+    cpi.user_data = data1;
 
     return cpi;
 }
@@ -536,6 +538,20 @@ TransformVertexInfo write_transform_vertex(vec4 instance_rect,
 
 #endif //WR_FEATURE_TRANSFORM
 
+struct ResourceRect {
+    vec4 uv_rect;
+};
+
+ResourceRect fetch_resource_rect(int index) {
+    ResourceRect rect;
+
+    ivec2 uv = get_fetch_uv_1(index);
+
+    rect.uv_rect = texelFetchOffset(sResourceRects, uv, 0, ivec2(0, 0));
+
+    return rect;
+}
+
 struct Rectangle {
     vec4 color;
 };
@@ -565,7 +581,6 @@ TextRun fetch_text_run(int index) {
 }
 
 struct Image {
-    vec4 st_rect;                        // Location of the image texture in the texture atlas.
     vec4 stretch_size_and_tile_spacing;  // Size of the actual image and amount of space between
                                          //     tiled instances of this image.
 };
@@ -573,10 +588,9 @@ struct Image {
 Image fetch_image(int index) {
     Image image;
 
-    ivec2 uv = get_fetch_uv_2(index);
+    ivec2 uv = get_fetch_uv_1(index);
 
-    image.st_rect = texelFetchOffset(sData32, uv, 0, ivec2(0, 0));
-    image.stretch_size_and_tile_spacing = texelFetchOffset(sData32, uv, 0, ivec2(1, 0));
+    image.stretch_size_and_tile_spacing = texelFetchOffset(sData16, uv, 0, ivec2(0, 0));
 
     return image;
 }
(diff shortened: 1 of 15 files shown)


glennw:
cc @nical @kvark @JerryShih 

How does this look as an API?

glennw:
Also CC @jrmuizel

JerryShih:
@glennw 

How about the external "gl texture"?
I think we need both "external buffer" and "external texture".

Here is my patch to resolve the external texture using the callback function:
https://github.com/JerryShih/gecko-dev/commit/56ab6b76b209bebc11613fecadf77f35670cc412#diff-8f929d6c300729f5810b7f5ae7d3233fR790

glennw:
@JerryShih Added basic support for using a native texture as an external image. I haven't added support for the release callback of native textures yet (it should work for cpu external images though). If we're happy in general with the API, I think it's fine to get this merged, and I'll do the native texture release code in a follow up PR.

glennw:
The semantics of the get/release calls are a bit confusing probably. Perhaps there should be a separate callback to just get the timestamp instead of calling get() each frame to check if the timestamp has changed? Thoughts?

JerryShih:
@glennw 
When should we use the timestamp?
If there is a video playback case, I think we could just call update_image() with the new "external key" for the new buffer and the exist image key which we want to update.
I think the timestamp logic(e.g. skip video frame) could put at c++ side. It makes WR much clear. When we call update_image() or add_image(), WR should just do the texture uploading with no condition.

glennw:
@JerryShih The reason for the timestamp was based on a discussion with @nical.

He suggested that we'd want to be able to just invoke the render thread to re-draw the frame in the case of a video where nothing else in the layout has changed, but the video is advancing (for performance and power reasons).

If we go via the update_image() API, that involves waking up the backend thread, processing the frame and pushing that to the render thread.

Personally, I'm happy with either solution - but hopefully that explains the logic of having the timestamp field, and we can discuss here the preferred option.

nical:
The timestamped video frames logic lives in the gecko/compositor side. I don't think we need to expose it to webrender, as long as the latter invokes the the callback every frame. The gecko implementation of the callback can query the composition timestamp from the Compositor and select the texture accordingly.

As far as gecko is concerned we don't want to use update_image() for video frames because it wakes up the backend thread and rebuilds the frame, which we don't want to happen every frame during video playback unless the geometry actually changed. We only want to swap out the the texture id and uv coordinates and submit the draw calls again.

webrender/src/renderer.rs
```diff
@@ -1267,6 +1299,100 @@ impl Renderer {
...
+        // The first thing we do is run through any pending deferred
+        // resolves, and use a callback to get the UV rect for this
+        // custom item. Then we patch the resource_rects structure
+        // here before it's uploaded to the GPU.
+        if !frame.deferred_resolves.is_empty() {
+            let handler = self.external_image_handler
+                              .as_mut()
+                              .expect("Found external image, but no handler set!");
+
+            for deferred_resolve in &frame.deferred_resolves {
```
kvark: is this guaranteed to run once only? and if not, will it behave properly for multiple invocations (scanning the `deferred_resolves` list)?

JerryShih:
@nical 
I think it's fine for "external texture id", but how about the "external buffer" ?
For external "texture id", the flow will be:
1) call add_image() with the external key at the first time
2) update the corresponding texture id for the external key
3) call wr_render() to do the rendering without any layout task

For external "buffer":
1) call add_image() with the external key at the first time
2) update the corresponding texture id for the external key
3) call wr_render(). At this time, Should we upload the buffer into gl texture here? What's the uploading condition for the external buffer? In this PR, @glennw uses a timestamp id to make sure WR only uploads the external buffer once. 

I think we still need to have a function to notify WR that here comes a new image. Then, WR could do the texture uploading if the image is an "external buffer".

Could update_image() only update the TextureSource, TextureId in the frame? It will still wake up the backend thread as other rendering api does, but just update the TextureSource or TextureId with for the new external key. Frame structure building should be skip in this case.

webrender/res/prim_shared.glsl
```diff
@@ -55,6 +55,7 @@ uniform sampler2D sData16;
 uniform sampler2D sData32;
 uniform sampler2D sData64;
 uniform sampler2D sData128;
+uniform sampler2D sResourceRects;
```
glennw: In the future, I expect this will be an array managed by the texture cache on a thread that is running async to the backend thread - so I've made it a separate array because of that.

webrender/src/prim_store.rs
```diff
@@ -16,6 +16,39 @@ use webrender_traits::{AuxiliaryLists, ColorF, ImageKey, ImageRendering};
 use webrender_traits::{FontRenderMode, WebGLContextId};
 use webrender_traits::{ClipRegion, FontKey, ItemRange, ComplexClipRegion, GlyphKey};
 
+/// Stores two coordinates in texel space. The coordinates
+/// are stored in texel coordinates because the texture atlas
+/// may grow. Storing them as texel coords and normalizing
```
glennw: Right now, we don't do that - we resize via FBO. But in the future that's possible, yes.

webrender/src/render_backend.rs
```diff
@@ -123,14 +123,19 @@ impl RenderBackend {
                             };
                             tx.send(glyph_dimensions).unwrap();
                         }
-                        ApiMsg::AddImage(id, width, height, stride, format, bytes) => {
-                            profile_counters.image_templates.inc(bytes.len());
+                        ApiMsg::AddImage(id, width, height, stride, format, data) => {
+                            match data {
+                                ImageData::Raw(ref bytes) => {
+                                    profile_counters.image_templates.inc(bytes.len());
```
glennw: Fixed

webrender/src/renderer.rs
```diff
@@ -1267,6 +1299,100 @@ impl Renderer {
     fn draw_tile_frame(&mut self,
                        frame: &mut Frame,
                        framebuffer_size: &Size2D<u32>) {
+        // The first thing we do is run through any pending deferred
```

webrender/src/renderer.rs
glennw: It can run once per scroll, but this is expected behaviour.

webrender/src/renderer.rs
```diff
@@ -1394,6 +1522,36 @@ impl Renderer {
     }
 }
 
+pub enum ExternalImageData {
```

glennw:
Rebased over recent changes and addressed most of the review comments.

I think @nical mentioned that a first test of this case would be using native textures? So I have removed the external buffer (and timestamp related code) until we work out a suitable API for them.

Hopefully that means we can land this part of the external image support (once any further API changes / review comments are sorted), which will give us external texture support, and then work on the external buffer support as a follow up API.

It now requests the external texture ID at the start of the frame, and releases after issuing all the draw calls. The calling application doesn't *have* to free the texture at that point (e.g. it may know it's going to use it again next frame), but it knows that WR will not issue another draw call referencing that texture without calling get() again first.

glennw:
@JerryShih @nical @kvark How does the above sound?

glennw:
I pushed a simple test case / sample of using the external image function to draw an image which is the output of an FBO clear - https://github.com/glennw/webrender/commit/481a6b1ba7aa2aca4ada005648172bcae2ce672c

nical:
Sounds good to me.
As far as gecko is concerned, external texture ids are enough to get the functionality we need. If we need external buffers in the future (or for the sake of completeness), we can add them later. There may be more efficient ways to upload textures if it is handled by webrender so we may want to revisit that. In the mean time, texture ids are simple to express at the ffi boundary and I think that getting it to work in this common scenario and iterating on the lifecycles and synchronization of external textures is what's most important at the moment.

@JerryShih perhaps I misunderstand what you mean by update_image(). This API currently just sends a message to the backend thread which rebuilds the frame with the updated information. For externally allocated textures all we need to happen is on the renderer thread, so it would make sense to have a separate API which interacts with the renderer directly rather than the backend.
Sorry for the confusion about timestamps in the PR, I thought you were referring to something else.

pcwalton:
r=me, squash if you want

PR #562, opened 2016-11-16
Remove establishes_3d_context field from StackingContext

mrobinson:
This field is currently unused.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/562)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -128,7 +128,6 @@ fn main() {
                                                0,
                                                &Matrix4D::identity(),
                                                &Matrix4D::identity(),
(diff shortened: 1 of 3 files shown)


mrobinson:
@glennw r?

kvark:
is it the same as checking for the `transform` to use Z?

mrobinson:
This used to mean a bunch of different things in Servo and, if I remember correctly, was used primarily by rust-layers. Now this is more or less hard-coded to
```
let establishes_3d_context = transform_style == transform_style::T::flat;
```

PR #564, opened 2016-11-17
Fix a bug with clipping when scrolling after a layout.

glennw:
The layer local rect is clipped against the root viewport
in local space, so it can't be used in a way that it was.

We don't actually need this here anyway. The local clip rect
of the layer is currently handled by write_vertex, and in the
future will be handled when stacking contexts write their
own clip mask.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/564)

### DIFF
--- a/webrender/src/mask_cache.rs
+++ b/webrender/src/mask_cache.rs
@@ -8,8 +8,10 @@ use internal_types::DeviceRect;
 use prim_store::{ClipData, GpuBlock32, PrimitiveClipSource, PrimitiveStore};
 use prim_store::{CLIP_DATA_GPU_SIZE, MASK_DATA_GPU_SIZE};
 use tiling::StackingContextIndex;
-use util::TransformedRect;
+use util::{rect_from_points_f, TransformedRect};
 use webrender_traits::{AuxiliaryLists, ImageMask};
+
+const MAX_COORD: f32 = 1.0e+16;
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub struct ClipAddressRange {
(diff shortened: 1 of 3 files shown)


webrender/src/mask_cache.rs
```diff
@@ -95,15 +97,14 @@ impl MaskCacheInfo {
...
                   source: &PrimitiveClipSource,
                   transform: &Matrix4D<f32>,
-                  clip_rect: &Rect<f32>,
                   clip_store: &mut GpuStore<GpuBlock32>,
                   device_pixel_ratio: f32,
                   aux_lists: &AuxiliaryLists) {
 
         if self.local_rect.is_none() {
-            let mut local_rect = Some(clip_rect.clone());
+            let mut local_rect = Some(rect_from_points_f(-MAX_COORD, -MAX_COORD, MAX_COORD, MAX_COORD));
```
kvark: oh, I was thinking if we have an omni-covering rectangle on multiple occasions :)

webrender/src/mask_cache.rs
```diff
@@ -95,15 +97,14 @@ impl MaskCacheInfo {
...
             match source {
-                &PrimitiveClipSource::NoClip => (),
+                &PrimitiveClipSource::NoClip => unreachable!(),
```
kvark: nice!

webrender/src/prim_store.rs
```diff
@@ -719,7 +719,6 @@ impl PrimitiveStore {
                                    prim_index: PrimitiveIndex,
                                    resource_cache: &mut ResourceCache,
                                    layer_transform: &Matrix4D<f32>,
-                                   layer_combined_local_clip_rect: &Rect<f32>,
```
kvark: glad to see this removed

PR #565, opened 2016-11-17
Introduce a glyph cache thread.

glennw:
This is the first step towards making the glyph cache more efficient
and also run asynchronously to allow rendering to continue if the
glyph rasterizing becomes a bottleneck (e.g. during animated zooms).

Previously, we collected the list of glyphs that were required for
a frame, and ran a rayon loop to rasterize any that weren't in the
glyph cache. However, this meant that the glyph rasterizion could
not start until we knew the list of all required glyphs.

Now, all requests for text run glyphs are sent immediately to the
glyph cache thread. The glyph cache thread uses a worker thread
pool to immediately begin rasterizing any uncached glyphs as soon
as they are requested.

In the future, the idea is that instead of waiting at the block()
method for all pending glyphs to be rasterized, the backend thread
will instead receive small numbers of glyphs as they are rasterized.
This will allow the backend thread to make a decision that glyph
rasterizing is taking too long for this frame, and that we should
just draw a frame with the currently cached glyph sizes available.
When this occurs, the glyph cache thread will signal that a new frame
should be rendered when it completes rasterizing any pending glyphs.
This will allow us to keep rendering at 60fps during zoom, even
if the glyph rasterizing threads can't keep up (at a quality
tradeoff of using lower resolution glyphs).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/565)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -8,21 +8,43 @@ use fnv::FnvHasher;
 use frame::FrameId;
 use internal_types::{FontTemplate, SourceTexture, TextureUpdateList};
 use platform::font::{FontContext, RasterizedGlyph};
-use rayon::prelude::*;
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::collections::hash_map::Entry::{self, Occupied, Vacant};
 use std::fmt::Debug;
 use std::hash::BuildHasherDefault;
 use std::hash::Hash;
-use std::sync::Arc;
+use std::sync::{Arc, Barrier};
+use std::sync::mpsc::{channel, Receiver, Sender};
+use std::thread;
 use texture_cache::{TextureCache, TextureCacheItemId};
 use webrender_traits::{Epoch, FontKey, GlyphKey, ImageKey, ImageFormat, ImageRendering};
 use webrender_traits::{FontRenderMode, ImageData, GlyphDimensions, WebGLContextId};
 use webrender_traits::{DevicePoint, DeviceIntSize};
 use webrender_traits::ExternalImageId;
+use threadpool::ThreadPool;
 
 thread_local!(pub static FONT_CONTEXT: RefCell<FontContext> = RefCell::new(FontContext::new()));
+
+type GlyphCache = ResourceClassCache<RenderedGlyphKey, Option<TextureCacheItemId>>;
+
+/// Message sent from the resource cache to the glyph cache thread.
+enum GlyphCacheMsg {
+    /// Begin the frame - pass ownership of the glyph cache to the thread.
+    BeginFrame(FrameId, GlyphCache),
+    /// Add a new font.
+    AddFont(FontKey, FontTemplate),
+    /// Request glyphs for a text run.
+    RequestGlyphs(FontKey, Au, Vec<u32>, FontRenderMode),
+    /// Finished requesting glyphs. Reply with new glyphs.
+    EndFrame,
+}
+
+/// Results send from glyph cache thread back to main resource cache.
+enum GlyphCacheResultMsg {
+    /// Return the glyph cache, and a list of newly rasterized glyphs.
+    EndFrame(GlyphCache, Vec<GlyphRasterJob>),
+}
 
 // These coordinates are always in texels.
 // They are converted to normalized ST
@@ -81,11 +103,6 @@ struct ImageResource {
     format: ImageFormat,
     epoch: Epoch,
     is_opaque: bool,
-}
-
-struct GlyphRasterJob {
-    glyph_key: RenderedGlyphKey,
-    result: Option<RasterizedGlyph>,
 }
 
 struct CachedImageInfo {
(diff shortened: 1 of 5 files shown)


pcwalton:
Does rayon not have a task API (like a job queuing kind of thing)? I thought it did.

pcwalton:
cc @nikomatsakis

pcwalton:
Niko says we do offer that, but it requires `--feature unstable` in Rayon. How about using that?

pcwalton:
https://github.com/nikomatsakis/rayon/blob/master/src/scope/mod.rs#L14

pcwalton:
https://github.com/nikomatsakis/rayon/blob/master/src/scope/mod.rs#L35-L226

emilio:
We use that in servo and stylo, so presumably that's ok for WR too.

nikomatsakis:
You would just do this:

```
rayon::scope(|s| {
    s.spawn(|s| /* parallel task */)
}
```

Each task gets a `s` argument that can be used to spawn further tasks into the same scope.

glennw:
I'm not sure that will work for the future use case described in the commit, since from what I understand the thread creating that scope will be blocked until all those job(s) complete? I'll have a play with it today, maybe I'm misunderstanding how the API works.

glennw:
@pcwalton @nikomatsakis I don't really see how that API fits the design of this PR.

I have a thread that handles the event loop (glyph requests etc) and I want that to be able to fire and forget about jobs, and then go back to the event loop and be able to do other work in the meantime. The idea is that at some point later in the frame, the event loop gets a message to block until those jobs are complete, but until that time its free to do other work.

But with the rayon scope, that event loop thread would be blocked until all child jobs complete, right? Is there an easy way to express that behaviour with rayon?

nikomatsakis:
@glennw 

The scope API doesn't seem like exactly what you want, I agree. That said, it *would* be pretty easy to add an API that lets you "fork" a job outside of a scope. You could then have this job send a signal to a port or something when it is done (one could easily build a future etc atop this, if you wanted). 

Another option would be to run the event loop itself as the scope, but the current APIs aren't really designed for that. I've been wanting to explore integrating as an "executor" in the futures library -- and then it might be a better fit -- but I haven't had time for that yet. (Similarly, I've been wanting to explore things like TBB's 'graph' abstraction, though I think there is a lot of overlap between that and futures.) 

Anyway, if you would like me to add a "fork" API like the one I described I could do that quite readily. All the machinery is there, it's just not wrapped in a public, safe API.

Caveat: I've not read this current PR. =)

(I've also been wanting to explore integrating Rayon with the futures library, but haven't had time to do that.)

nikomatsakis:
I should add: one advantage of using Rayon uniformly across Servo is that you have the option all sharing a global thread pool, which tends to be a win, since you don't have multiple threadpools competing for resources. (Though it could cause longer latency, so you may not want that.)

nikomatsakis:
OK, I skimmed the PR. It seems like the API I proposed would act more-or-less precisely the same as the `thread_pool` crate.

glennw:
@nikomatsakis Yep, if we had a rayon API that could work that way, it'd be great to be able to switch this back to rayon and make use of the shared thread pool.

nikomatsakis:
@glennw can definitely do that...

nikomatsakis:
@glennw https://github.com/nikomatsakis/rayon/pull/161

glennw:
@nikomatsakis Would it make sense to land this PR now and switch it over to use rayon once that PR lands? (I guess what I'm really asking is how long will it be until that API is available in rayon for use?).

nikomatsakis:
@glennw Either way really. I'd probably be happy to land that patch as is, though I would expect to update the API slightly in the future (which might or might not be a breaking change for you). But if you wanted to land this PR and then compare a rayon-based impl when it becomes available, that's fine too.

I'm curious, is it easy for you to do some benchmarks now? (Or would it be, if I landed that PR?)

glennw:
@nikomatsakis My preference would probably be to get this reviewed and landed, and then switch it to rayon as a follow up (then you won't need to rush to merge your PR either). Benchmarking between the two implementations should be easy to do. Sound ok, @pcwalton ?

pcwalton:
Yup, sounds good!

glennw:
Rebased and pushed. r? @pcwalton

kvark:
There seem to be quite a few thread barriers on the glyph way to rasterization: RenderApi -> RenderBackend -> GlyphCacheThread -> WorkerThread. We might want to look (at some point later on) into reducing this depth.

webrender/src/resource_cache.rs
```diff
@@ -9,21 +9,43 @@ use fnv::FnvHasher;
...
 
 thread_local!(pub static FONT_CONTEXT: RefCell<FontContext> = RefCell::new(FontContext::new()));
 
+type GlyphCache = ResourceClassCache<RenderedGlyphKey, Option<TextureCacheItemId>>;
+
+/// Message sent from the resource cache to the glyph cache thread.
+enum GlyphCacheMsg {
+    /// Begin the frame - pass ownership of the glyph cache to the thread.
+    BeginFrame(FrameId, GlyphCache),
+    /// Add a new font
```
mrobinson: Super-super-nit: this comment doesn't have a period, but all the others do.

mrobinson:
This looks good to me as far as I understand it.

glennw:
@mrobinson Thanks, fixed up the nit and rebased.

PR #566, opened 2016-11-17
Use atomic reference counting to store images and avoid copying them.

nical:
We currently clone the bytes of images and glyphs before inserting them in the texture cache, because we need to keep a copy around in case the device is reset or the image is evicted from the cache. We want the images and glyphs to be managed on the backend thread while the buffers need to be read on the renderer thread when they get uploaded.
This PR removes the clone of the bytes by sharing the buffers using atomic reference counting.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/566)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -133,7 +133,7 @@ fn main() {
 
     let clip_region = {
         let mask = webrender_traits::ImageMask {
-            image: api.add_image(2, 2, None, ImageFormat::A8, ImageData::Raw(vec![0,80, 180, 255])),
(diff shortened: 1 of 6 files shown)


nical:
Rebased on top of the external image stuff. The fact that we still use ```bytes.clone()``` with ```Arc<Vec<u8>>``` instead of ```Vec<u8>``` with a totally different meaning makes me cringe, but Arc doesn't seem to have an equivalent to clone that doesn't read like expensive allocations (something like add_ref would be nice).

PR #572, opened 2016-11-17
Make ipc-channel an optional feature.

### LINKED ISSUE #558: Support using WebRender without IPC
It looks like we're going to be using Gecko IPC with WebRender instead of the Servo IPC. In this situation we don't want to be eating the overhead of an additional IPC serialization/copy.

glennw:
This allows clients to opt-in to IPC for communication if required.

Otherwise, in-process rust channels are used, avoiding serialization.

This fixes #558.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/572)


glennw:
cc @jrmuizel

webrender_traits/src/channel.rs
```diff
@@ -0,0 +1,148 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use serde::{Deserialize, Serialize};
+use std::io::Error;
+
+#[cfg(feature = "ipc")]
```
kvark: given that everything in this module is split between `ipc` and non-ipc, I'd rather see separate files `channel.rs` and `channel_ipc.rs` with a single switch upon including them.

glennw: Good idea - https://github.com/servo/webrender/issues/573

PR #582, opened 2016-11-21
Export the servo-freetype-sys feature from webrender

staktrace:
See commit message for why this is needed.

I tested this to ensure that in the Firefox build we can opt-in to the codegen feature but NOT freetype-lib. The servo build seems to also be ok with this change (I cherry-picked this patch to the version of webrender that servo is currently using, and used [replace] to make servo use the modified version with this patch). The sample library in webrender also builds fine with this change, since servo-freetype-sys is enabled in the default configuration. If there's more testing needed please let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/582)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -8,7 +8,9 @@ build = "build.rs"
 workspace = ".."
 
 [features]
-default = ["webrender_traits/codegen"]
+default = ["codegen", "freetype-lib"]
+codegen = ["webrender_traits/codegen"]
+freetype-lib = ["freetype/servo-freetype-sys"]
 serde_derive = ["webrender_traits/serde_derive"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


glennw:
@staktrace Just to confirm, the Servo test you did with this was on Linux? Assuming that's the case, this looks fine to me.

staktrace:
@glennw Yes, it was on Linux. For the record, after poking around in the servo Cargo.toml files, it looks like servo includes webrender *without* default features, and so with my change it should actually fail to build, unless I explicitly add the new feature. However, servo also depends on azure, which includes freetype with default features, so I think the dependency is getting satisfied via that route. It might be a good idea to update servo's Cargo.toml files that pull in webrender to explicitly request the freetype-lib feature, once this patch has landed and servo is using a version of WR that includes this patch.

glennw:
@staktrace Sounds good, thanks for checking!

PR #583, opened 2016-11-22
Change font platforms to work with device pixels directly.

glennw:
This means that glyph keys are correctly hashed and
cached when the device pixel ratio is changed dynamically.

This is a step towards support zoom functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/583)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -121,8 +121,7 @@ impl FontContext {
 
     fn get_ct_font(&mut self,
                    font_key: FontKey,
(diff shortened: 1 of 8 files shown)


glennw:
r? @pcwalton and also @vvuk for the Windows parts.

I have built and run this on Linux, Mac, Windows and it all looks correct.

vvuk:
r+ on the windows parts

pcwalton:
@bors-servo: r+

Nice! This simplifies code too :)

PR #590, opened 2016-11-23
Add API for scrolling individual layers

mrobinson:
This will allow Servo to reimplement scroll_to_fragment.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/590)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -21,7 +21,7 @@ use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLa
 use webrender_traits::{ClipRegion, ColorF, DisplayItem, StackingContext, FilterOp, MixBlendMode};
 use webrender_traits::{ScrollEventPhase, ScrollLayerInfo, SpecificDisplayItem, ScrollLayerState};
 use webrender_traits::{LayerRect, LayerPoint, LayerSize};
-use webrender_traits::{ScrollLayerRect, as_scroll_parent_rect, ScrollLayerPixel};
+use webrender_traits::{ServoScrollRootId, ScrollLayerRect, as_scroll_parent_rect, ScrollLayerPixel};
 use webrender_traits::WorldPoint4D;
 use webrender_traits::{LayerTransform, LayerToScrollTransform, ScrollToWorldTransform};
 
@@ -52,6 +52,7 @@ pub struct Frame {
                                           AuxiliaryLists,
                                           BuildHasherDefault<FnvHasher>>,
     pub root_scroll_layer_id: Option<ScrollLayerId>,
+    pending_scroll_offsets: HashMap<(PipelineId, ServoScrollRootId), LayerPoint>,
     id: FrameId,
     debug: bool,
     frame_builder_config: FrameBuilderConfig,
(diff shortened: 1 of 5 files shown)


mrobinson:
@glennw r?

webrender/src/frame.rs
```diff
@@ -296,6 +298,40 @@ impl Frame {
...
+            }
+
+            match layer_id.info {
+                ScrollLayerInfo::Scrollable(_, id) if id != scroll_root_id => continue,
+                ScrollLayerInfo::Fixed => continue,
+                _ => {}
+            }
+
+            found_layer = true;
+            scrolled_a_layer = layer.set_scroll_origin(&origin);
```
emilio: probably you want this to be `scrolled_a_layer |= layer.set_scroll_origin(&origin)`, right?

Otherwise (if we can just scroll one layer) we might as well break from the loop when that happens, right?

webrender/src/frame.rs
```diff
@@ -296,6 +298,40 @@ impl Frame {
...
+        }
+
+        if !found_layer {
+            let scroll_offsets =
+                self.pending_scroll_offsets.entry(pipeline_id).or_insert(HashMap::new());
```
emilio: nit: `or_insert_with(HashMap::new)` doesn't call `HashMap::new()` unconditionally, though I don't know if that allocates so it could be meaningless.

webrender/src/frame.rs
mrobinson: Nice catch! I will fix this.

In practice this doesn't cause a problem because all layers with the same ScrollRootId should be be in sync. On the other hand, it is good to write resilient code, which is what I tried to do here, but failed.

webrender/src/frame.rs
mrobinson: ```or_insert_with``` is much better here. Thanks!

webrender/src/frame.rs
```diff
@@ -52,6 +52,7 @@ pub struct Frame {
                                           AuxiliaryLists,
                                           BuildHasherDefault<FnvHasher>>,
     pub root_scroll_layer_id: Option<ScrollLayerId>,
+    pending_scroll_offsets: HashMap<PipelineId, HashMap<ServoScrollRootId, LayerPoint>>,
```
kvark: wouldn't it be more efficient to have a single `HashMap<(PipelineId, ServoScrollRootId), LayerPoint>`?

webrender/src/layer.rs
```diff
@@ -86,6 +86,26 @@ impl Layer {
         LayerSize::new(overscroll_x, overscroll_y)
     }
 
+    pub fn set_scroll_origin(&mut self, origin: &LayerPoint) -> bool{
```
kvark: nit: space missing before `{`

glennw:
Looks good once @kvark's comments are addressed.

webrender/src/frame.rs
mrobinson: Yeah, that is much nicer!

webrender/src/layer.rs
mrobinson: Okay. Will fix.

mrobinson:
@bors-servo r=glennw

Thanks everyone for the reviews!

PR #591, opened 2016-11-24
Add specific shaders for rectangles with CLIP feature.

glennw:
Having the base rectangle shader include clip makes the software
mesa implementation (somewhat) slower for reftests. This is enough
of a timing difference that it triggers a lot of intermittent
test failures in Servo.

Instead, include a simple rectangle shader for workloads that
don't have a clip mask.

The other shaders are typically more expensive and not used as
frequently by the reftests, so hopefully this change is enough
to work around the Servo intermittent bugs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/591)


mrobinson:
@bors-servo r+

Hopefully this fixes the issues with the Servo tests.

PR #595, opened 2016-11-25
Clip stack collapse path

kvark:
Based on #581
Implements the 3rd option explained in https://github.com/servo/webrender/pull/581#issuecomment-262657539 as an optional code path, controlled by `CLIP_TASK_COLLAPSE`, enabled by default.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/595)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -102,15 +102,11 @@ fn main() {
 
     let bounds = Rect::new(Point2D::new(0.0, 0.0), Size2D::new(width as f32, height as f32));
     let clip_region = {
-        let radius = webrender_traits::BorderRadius::uniform(20.0);
         let complex = webrender_traits::ComplexClipRegion::new(
             Rect::new(Point2D::new(50.0, 50.0), Size2D::new(100.0, 100.0)),
-            radius);
+            webrender_traits::BorderRadius::uniform(20.0));
 
-        webrender_traits::ClipRegion::new(&bounds,
-                                          vec![complex],
-                                          None,
-                                          &mut auxiliary_lists_builder)
+        builder.new_clip_region(&bounds, vec![complex], None)
     };
 
     builder.push_stacking_context(webrender_traits::ScrollPolicy::Scrollable,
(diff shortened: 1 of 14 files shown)


glennw:

---

*[webrender/src/tiling.rs, line 616 at r1](https://reviewable.io:443/reviews/servo/webrender/595#-KXd8M3BsVBvUV6QRiTI:-KXd8M3CgqKoZ7Vd2hxA:bveidhm) ([raw file](https://github.com/servo/webrender/blob/8db149b2be80504ed78dce4153b60e8be2f63fc5/webrender/src/tiling.rs#L616)):*
> ```Rust
>                         let mask_task_index = match ctx.layer_masks_tasks.get(&(task.tile_id, sc_index)) {
>                             Some(ref mask_task_id) => render_tasks.get_task_index(mask_task_id, child_pass_index),
>                             None => RenderTaskIndex(i32::MAX as usize), // special sentinel value recognized by the shader
> ```

This should be OPAQUE_TASK_INDEX I think.

---

*[webrender/src/tiling.rs, line 1750 at r1](https://reviewable.io:443/reviews/servo/webrender/595#-KXd92vnHfQiPgk8bCAI:-KXd92vocDRj1Q5rVvnO:bt4caig) ([raw file](https://github.com/servo/webrender/blob/8db149b2be80504ed78dce4153b60e8be2f63fc5/webrender/src/tiling.rs#L1750)):*
> ```Rust
> 
>     fn build(self, passes: &mut Vec<RenderPass>) {
>         //println!("{:#?}", self.main_render_task);
> ```

Can be removed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/595)*

glennw:
@kvark This looks great! Two very minor nits, then it looks good to me.

Before merging, I'd like to just get @mrobinson to make sure the stacking context interface change is OK. Instead of having an overflow rect in the stacking context, the method to push a new stacking context now takes a clip region, which can be a normal rect, but could include an image mask etc. Does this interface change have any implications for the scroll root code?

kvark:

---

*[webrender/src/tiling.rs, line 616 at r1](https://reviewable.io:443/reviews/servo/webrender/595#-KXd8M3BsVBvUV6QRiTI:-KXflC_d9hvCWE9CnpyT:b-gs9178) ([raw file](https://github.com/servo/webrender/blob/8db149b2be80504ed78dce4153b60e8be2f63fc5/webrender/src/tiling.rs#L616)):*

This should be OPAQUE_TASK_INDEX I think.
</blockquote></details>

That was added when I needed it on multiple occasions, but it was removed later on. Anyway, I re-added it back now :)

---

*[webrender/src/tiling.rs, line 1750 at r1](https://reviewable.io:443/reviews/servo/webrender/595#-KXd92vnHfQiPgk8bCAI:-KXflKi5JoodEUSDd7tP:b32wtd) ([raw file](https://github.com/servo/webrender/blob/8db149b2be80504ed78dce4153b60e8be2f63fc5/webrender/src/tiling.rs#L1750)):*

Can be removed
</blockquote></details>

ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/595)*

kvark:
All concerns are addressed, and the PR is rebased now.

PR #597, opened 2016-11-25
Split channel.rs into mpsc and ipc files

### LINKED ISSUE #573: Split channel.rs into two files - mpsc / ipc versions

dignifiedquire:
Fixes #573

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/597)


webrender_traits/src/channel.rs
```diff
@@ -2,147 +2,14 @@
...
-
-#[cfg(not(feature = "ipc"))]
-impl<T> Deserialize for MsgSender<T> {
-    fn deserialize<D>(_: &mut D) -> Result<MsgSender<T>, D::Error>
-                      where D: Deserializer {
-        unreachable!();
-    }
-}
+#[cfg(feature = "ipc")]
+include!("channel_ipc.rs");
```
emilio: I'm definitely fine with this, though maybe this could be a bit more idiomatic doing something like:

```rust
#[cfg(feature = "ipc")]
mod channel_ipc;
#[cfg(feature = "ipc")]
pub use channel_ipc::*;
```

(And the same with the mpsc).

It duplicates the `cfg`, but avoids macros (which may be a bit nicer IMO). What do you think? Happy to concede if people have different opinions on it :)

Also, if you could add a line to `.travis.yml` to test both configs as part of this PR, that would be _awesome_.

Thanks!

dignifiedquire: Added tests to travis :)

I'm happy to change the imports. I saw the `include!` already being used in `lib.rs` so I thought it would be fine (and it was the first way I figured out to make the compiler happy :)
I don't have any strong opinions either way as I am just starting to figure out my way around Rust .

.travis.yml
```diff
@@ -12,6 +12,9 @@ addons:
       - libgles2-mesa-dev
 script:
   - (cd webrender_traits && cargo test --verbose)
+  - (cd webrender_traits && cargo test --verbose --features "ipc")
   - (cd webrender && cargo test --verbose)
+  - (cd webrender && cargo test --verbose --features "ipc")
```
emilio: `webrender` and `sample` don't have the `ipc` feature, this is only a `webrender_traits` thing, so this line and the one below you added should go away.

.travis.yml
```diff
@@ -12,6 +12,9 @@ addons:
       - libgles2-mesa-dev
 script:
   - (cd webrender_traits && cargo test --verbose)
+  - (cd webrender_traits && cargo test --verbose --features "ipc")
```
emilio: Please put this line the first one in the script to avoid unnecessary recompiles.

.travis.yml
dignifiedquire: done

.travis.yml
dignifiedquire: done

webrender_traits/src/channel.rs
dignifiedquire: I am getting `error: cannot declare a new module at this location` when trying to use the declarations suggested above in `channel.rs`. Not sure how to make this work, any tips?

emilio: Huh... Well, doesn't matter, as you said, we also use that for serde.

PR #599, opened 2016-11-27
Implement scroll handoff

gterzian:
"If a new scroll gesture starts while you're already scrolled all the way to the edge of the scrollable element under the mouse, then the scroll will be handed off to the next outer scrollable element which is actually scrollable in that direction" servo/servo#13249 (comment)

Page used for testing can be found online [here](https://samuknet.github.io/test_cases/nestedScroll/).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/599)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -352,9 +352,44 @@ impl Frame {
             (_, _, None) => return false,
         };
 
-        let scroll_root_id = match scroll_layer_id.info {
-            ScrollLayerInfo::Scrollable(_, scroll_root_id) => scroll_root_id,
-            ScrollLayerInfo::Fixed => unreachable!("Tried to scroll a fixed position layer."),
+        let non_root_overscroll = if scroll_layer_id != root_scroll_layer_id {
+            // true if the current layer is overscrolling,
+            // and it is not the root scroll layer.
+            let child_layer = self.layers.get(&scroll_layer_id).unwrap();
+            let overscroll_amount = child_layer.overscroll_amount();
+            overscroll_amount.width != 0.0 || overscroll_amount.height != 0.0
+        } else {
+            false
+        };
+
+        let switch_layer = match phase {
+            ScrollEventPhase::Start => {
+                // if this is a new gesture, we do not switch layer,
+                // however we do save the state of non_root_overscroll,
+                // for use in the subsequent Move phase.
+                let mut current_layer = self.layers.get_mut(&scroll_layer_id).unwrap();
+                current_layer.scrolling.should_handoff_scroll = non_root_overscroll;
+                false
+            },
+            ScrollEventPhase::Move(_) => {
+                // Switch layer if movement originated in a new gesture,
+                // from a non root layer in overscroll.
+                let current_layer = self.layers.get_mut(&scroll_layer_id).unwrap();
+                current_layer.scrolling.should_handoff_scroll && non_root_overscroll
+            },
+            ScrollEventPhase::End => {
+                // clean-up when gesture ends.
+                let mut current_layer = self.layers.get_mut(&scroll_layer_id).unwrap();
+                current_layer.scrolling.should_handoff_scroll = false;
+                false
+            },
+        };
+
+        let scroll_root_id = match (switch_layer, scroll_layer_id.info, root_scroll_layer_id.info) {
+            (true, _, ScrollLayerInfo::Scrollable(_, scroll_root_id)) |
+            (true, ScrollLayerInfo::Scrollable(_, scroll_root_id), ScrollLayerInfo::Fixed) |
+            (false, ScrollLayerInfo::Scrollable(_, scroll_root_id), _) => scroll_root_id,
+            (_, ScrollLayerInfo::Fixed, _) => unreachable!("Tried to scroll a fixed position layer."),
         };
 
         let mut scrolled_a_layer = false;
--- a/webrender/src/layer.rs
+++ b/webrender/src/layer.rs
@@ -128,6 +128,7 @@ pub struct ScrollingState {
     pub spring: Spring,
     pub started_bouncing_back: bool,
     pub bouncing_back: bool,
+    pub should_handoff_scroll: bool
 }
 
 impl ScrollingState {
@@ -137,6 +138,7 @@ impl ScrollingState {
             spring: Spring::at(LayerPoint::zero(), STIFFNESS, DAMPING),
             started_bouncing_back: false,
             bouncing_back: false,
+            should_handoff_scroll: false
         }
     }
 }


glennw:
@gterzian OK, I tested this on Linux. It doesn't seem to break anything, but it also doesn't seem to work - the old behaviour persists where if I keep the cursor in the nested scroll area, and scroll to the bottom, it doesn't scroll the outer scroll area. I tried this with both a mouse wheel and also scrolling on my touchpad. I might have some time during the week to debug this and see what's going on.

glennw:
@gterzian I did a little bit of investigation on Linux. It looks like on Linux we never receive TouchStart / TouchEnd events. From a very quick look in glutin, it seems that it does provide these events on Linux, but only for touch events (although I don't receive them even when using the touchpad).

I'm not actually sure whether we should expect to rely on always getting a start/end event, or if we should expect to receive Scroll / Move events without a corresponding start / end. Perhaps @mbrubeck  could comment on this?

I wonder if it's possible to update the logic in this PR to work without relying on the start / end events?

glennw:
@gterzian PS the above now explains why we see these issues on Linux and they are unrelated to the CAN_OVERSCROLL flag.

gterzian:
@glennw thanks a lot for the investigation! This makes a lot of sense and also explains why https://github.com/servo/webrender/pull/600 would break scrolling altogether. 

I will have to investigate more whether the desired behaviour can be implemented without relying on start/end scroll phases...

mstange:
Yes, macOS is the only platforms that groups wheel events into units, by giving you start+end events with phases like NSEventPhaseBegan and NSEventPhaseEnded. Unless you connect a traditional USB mouse to your Mac - in that case, even macOS can't help you, because the hardware has no way to tell whether your finger is still on the scroll wheel. (In terms of native wheel events, USB mouse wheel ticks result in wheel events with scroll phase NSEventPhaseNone.)

So on all non-Mac platforms, and on Mac for events with NSEventPhaseNone, you need to do your own grouping of events. In Firefox, we do this based on pause time and mouse movement. The code is around here: http://searchfox.org/mozilla-central/source/gfx/layers/apz/src/InputBlockState.cpp#361
New wheel events are treated as part of the current wheel event group if the wheel block state is "in transaction".

glennw:
@mstange Thanks for the info. @gterzian It seems this is turning into a larger issue than originally expected! I wonder if we should be doing that event grouping in WR or in Servo - any thoughts @mbrubeck  or others?

gterzian:
@glennw @mstange Thank you for your input. I would be happy to work on this 'event grouping', either in Servo or here.

gterzian:
Ok so looking at the Firefox code, it seems that they group the input events together into transactions, where a 'transaction' represents pretty much a continuous scroll gesture. The `scroll handoff` logic is handled together with the transaction logic, for example see http://searchfox.org/mozilla-central/source/gfx/layers/apz/src/InputBlockState.cpp#524

In Servo it's currently Webrender that does the scrolling, and Webrender receives those scroll events from the compositor (from looking at this it seems the compositor is already grouping events together in some way: https://github.com/servo/servo/blob/master/components/compositing/compositor.rs#L218).

I guess the choice we face is either to implement a 'scroll transaction' in compositor, and send events to webrender always with either `Start`(new transaction), `Move`(current transaction), and `End`(end transaction) phases, and then my current code will pretty much work, and it's webrender that would handle the `scroll handoff` logic, while remaining unaware of the transaction logic. 

We could also implement those transactions in webrender. By simply always starting/ending the transaction when we do receive `Start/End` phases, and otherwise rely on our own logic when we receive only `Move` events. A downside of doing those transactions in WR might be that we don't know what the original event was, so we can't change our 'transaction logic' based on the scrolling hardware used.

Please let me know what you think...

glennw:
I discussed with @mbrubeck - we think it makes sense to do the grouping the Servo compositor - that lets WR deal with the high level concept of scroll transactions, and leaves glutin to deal with raw events.

gterzian:
@glennw great! thanks for your input, I will start looking around compositor...

gterzian:
@glennw the Servo "scroll transaction" part has landed(https://github.com/servo/servo/pull/14470), and I have rebased to remove the conflicts. It's actually hard to test it once again, because I see the Servo side of https://github.com/servo/webrender/pull/540/files has not landed yet... 

I did test it extensively as part of the Servo PR. 

Once thing to note is that some details are still left to be implemented: "the scroll will be handed off to the next outer scrollable element which is actually scrollable in that direction".

Currently I just hand-off the scroll to the root scroll layer(even though I think `self.get_scroll_layer` probably grabs the next outer scollable element, if only on the next scroll event). 

I am certainly not checking for "which is actually scrollable in that direction". 

Do you think it would be best to further work on those features in this PR, or rather look at this one has a good first step, and leave those notes in a follow-up issue on Servo?

glennw:
I think it's fine to land this as a first step - it's a net improvement. I've landed the WR update in Servo, so it should be possible to rebase this, test it, and get it merged once the two formatting nits are fixed.

webrender/src/frame.rs
```diff
@@ -445,12 +480,12 @@ impl Frame {
             if CAN_OVERSCROLL {
                 layer.stretch_overscroll_spring();
             }
-
+
```
glennw: nit: extra space

webrender/src/frame.rs
```diff
@@ -445,12 +480,12 @@ impl Frame {
...
                 layer.stretch_overscroll_spring();
             }
-
+            
             scrolled_a_layer = scrolled_a_layer ||
                 layer.scrolling.offset != original_layer_scroll_offset ||
                 layer.scrolling.started_bouncing_back;
         }
-
+
```
glennw: nit: extra space

gterzian:
@glennw OK i've rebased off of https://github.com/servo/webrender/commit/4474785d4f3050bba4f1a44f65e68abcd8dedb2f (there were some deps issues with the current head), tested on a Mac in 'normal' mode, as well as with overscroll turned off and with start/end event handling in compositor turned off as well.

Seems to work fine, when you have a chance to test in linux proper it would be much appreciated! 😄

glennw:
@gterzian OK, I tested on Linux this morning - it doesn't seem to work for me on the test page referenced. I'll try to dig into why sometime this week.

gterzian:
@glennw thanks!

glennw:
@gterzian Apologies - I haven't tested this yet on Linux - will try to get to it during this week :)

glennw:
@gterzian I'm unsure what changed, but this is working perfectly for me on Linux now :)

gterzian:
@glennw thanks for all your help, it has been a long one! 😄

PR #600, opened 2016-11-27
Keeping the currently scrolling layer, unless new gesture starts

gterzian:
"If you scroll a long page, and somewhere on that page is a smaller scrollable element, you don't want that element to swallow the rest of your scroll gesture if it suddenly happens to move under your mouse. You want the current scroll gesture to keep scrolling the outer page." servo/servo#13249 (comment)

Page used for testing can be found online [here](https://samuknet.github.io/test_cases/nestedScroll/).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/600)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -53,6 +53,7 @@ pub struct Frame {
                                           BuildHasherDefault<FnvHasher>>,
     pub root_scroll_layer_id: Option<ScrollLayerId>,
     pending_scroll_offsets: HashMap<(PipelineId, ServoScrollRootId), LayerPoint>,
+    current_scroll_layer_id: Option<ScrollLayerId>,
     id: FrameId,
     debug: bool,
     frame_builder_config: FrameBuilderConfig,
@@ -212,6 +213,7 @@ impl Frame {
             layers: HashMap::with_hasher(Default::default()),
             root_scroll_layer_id: None,
             pending_scroll_offsets: HashMap::new(),
+            current_scroll_layer_id: None,
             id: FrameId(0),
             debug: debug,
             frame_builder: None,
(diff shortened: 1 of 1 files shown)


webrender/src/frame.rs
```diff
@@ -340,9 +342,15 @@ impl Frame {
             None => return false,
         };
 
-        let scroll_layer_id = match self.get_scroll_layer(&cursor, root_scroll_layer_id) {
-            Some(scroll_layer_id) => scroll_layer_id,
-            None => return false,
+        let scroll_layer_id = match (phase, self.get_scroll_layer(&cursor, root_scroll_layer_id),
+            self.current_scroll_layer_id) {
```
kvark: the indentation doesn't look right to me. Perhaps, align every tuple element after `match (`?

gterzian:
@kvark thanks for your review, I will address your points in time... By the way this PR, just like Handoff scroll when end of layer reached, and new gesture starts servo/webrender#599, will require additional work because it currently only works on Mac OS, and breaks scrolling on other platforms...

gterzian:
@glennw Ok I've just done some additional testing, and realized my previous implementations of 'scroll transactions' was missing one case see https://github.com/servo/servo/pull/14597 other than that this one shouldn't break scrolling on Linux anymore, and still implement the desired behavior. I've tested this on my mac by disabling overscrolling, as well as manually removing the start/end events from the compositor...

glennw:
@gterzian Great, thanks! Once that fix lands for Servo, I'll test this locally and get it merged :)

glennw:
Tested on Linux on a few sites, and this seems to work well. Thanks!

gterzian:
@glennw thanks for your help!

PR #602, opened 2016-11-28
Use strongly typed geometry - Part 2

nical:
Use strongly typed geometry in the public API. This changeset adds the LayoutPixel unit which is basically an alias to LayerPixel.
I think that it's best to keep the "Layer" terminology internal to webrender, and since the 1-1 correspondence between the api's layout pixels and internal layer pixels is somewhat coincidental, it will help to have separate names if things like async zoom introduce an actual difference between the two coordinate spaces (as it does in Gecko). Using an alias instead of a separate type comes from a mix of laziness (not having to cast from layout t layer pixels all over frame.rs) and the fact that currently layer and layout pixels are the same thing, but I'll add a separate unit if there is a preference for it.

I did not introduce ParentLayoutPixel. I don't know the API well enough yet to be sure whether some geometry is passed in a stacking context's parent coordinate space, but if so we should consider introducing a special unit for it, if only for the sake of proper documentation.

This PR is a lot easier to rebase than part 1 and is a breaking change to the public API, so it's fine to wait a bit if there are cross-crates changes that we want to coordinate before having to adapt servo to this (although it should be easy to do).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/602)

### DIFF
--- a/replay/src/main.rs
+++ b/replay/src/main.rs
@@ -8,7 +8,6 @@ extern crate webrender_traits;
 
 use bincode::serde::deserialize;
 use byteorder::{LittleEndian, ReadBytesExt};
-use euclid::Size2D;
 use gleam::gl;
 use std::any::TypeId;
 use std::mem;
@@ -16,7 +15,7 @@ use std::io::Read;
 use std::fs::File;
 use std::path::{Path, PathBuf};
 use std::env;
-use webrender_traits::{ApiMsg, RenderApi, PipelineId};
+use webrender_traits::{ApiMsg, RenderApi, PipelineId, LayerSize, DeviceUintSize};
 use webrender_traits::channel::PayloadHelperMethods;
 use glutin::{Event, ElementState, VirtualKeyCode as Key};
 
@@ -40,7 +39,7 @@ impl webrender_traits::RenderNotifier for Notifier {
 
     fn pipeline_size_changed(&mut self,
                              _pid: PipelineId,
-                             _size: Option<Size2D<f32>>) {
+                             _size: Option<LayerSize>) {
     }
 
     fn new_scroll_frame_ready(&mut self, _composite_needed: bool) {
@@ -145,7 +144,7 @@ fn main() {
                 println!("Rendering frame {}.", frame_num);
                 gl::clear(gl::COLOR_BUFFER_BIT);
                 renderer.update();
-                renderer.render(Size2D::new(width, height));
+                renderer.render(DeviceUintSize::new(width, height));
                 window.swap_buffers().unwrap();
             }
             Event::KeyboardInput(ElementState::Pressed, _, Some(Key::Right)) =>{
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use app_units::Au;
-use euclid::{Point2D, Point3D, Rect, Size2D};
+use euclid::Point3D;
 use fnv::FnvHasher;
 use geometry::ray_intersects_rect;
 use internal_types::{ANGLE_FLOAT_TO_FIXED, AxisDirection};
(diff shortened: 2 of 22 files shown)


replay/src/main.rs
```diff
@@ -8,13 +8,12 @@ extern crate webrender_traits;
...
 use bincode::serde::deserialize;
 use byteorder::{LittleEndian, ReadBytesExt};
-use euclid::Size2D;
 use gleam::gl;
 use std::io::Read;
 use std::fs::File;
 use std::path::{Path, PathBuf};
 use std::env;
-use webrender_traits::{RenderApi, PipelineId};
+use webrender_traits::{RenderApi, PipelineId, LayerSize, DeviceUintSize};
```
kvark: Having `DeviceIntRect`, `DeviceUintRect`, and just `DeviceRect` now, the first thing that comes to mind is that the latter is inconsistent. Would be more consistent if it was `DeviceFloatRect`. But really, it would be much simpler to just have `DeviceRect<i32>`, `DeviceRect<u32>`, and so forth - we'd still get the safe units but with somewhat clearer semantics.

webrender/src/frame.rs
```diff
@@ -331,8 +331,8 @@ impl Frame {
 
     /// Returns true if any layers actually changed position or false otherwise.
     pub fn scroll(&mut self,
-                  mut delta: Point2D<f32>,
-                  cursor: Point2D<f32>,
+                  mut delta: LayerPoint,
```
kvark: I believe this should really be `LayerSize`

webrender/src/frame.rs
```diff
@@ -472,9 +472,9 @@ impl Frame {
         // Insert global position: fixed elements layer
         debug_assert!(self.layers.is_empty());
         let root_fixed_layer_id = ScrollLayerId::create_fixed(root_pipeline_id);
-        let root_viewport = LayerRect::new(LayerPoint::zero(), LayerSize::from_untyped(&root_pipeline.viewport_size));
+        let root_viewport = LayerRect::new(LayerPoint::zero(), root_pipeline.viewport_size);
```
kvark: sweeet

webrender/src/internal_types.rs
```diff
@@ -146,18 +146,18 @@ impl GLContextWrapper {
...
                 let limits = ctx.borrow_limits().clone();
 
-                (real_size, texture_id, limits)
+                (DeviceIntSize::from_untyped(&real_size), texture_id, limits)
             }
         }
     }
 
-    pub fn resize(&mut self, size: &Size2D<i32>) -> Result<(), &'static str> {
+    pub fn resize(&mut self, size: &DeviceIntSize) -> Result<(), &'static str> {
```
kvark: it rubs me the wrong way to see `Int` there, but I can see that it was this way before you

webrender_traits/src/api.rs
```diff
@@ -207,8 +207,8 @@ impl RenderApi {
     }
 
     /// Translates a point from viewport coordinates to layer space
-    pub fn translate_point_to_layer_space(&self, point: &Point2D<f32>)
-                                          -> (Point2D<f32>, PipelineId) {
+    pub fn translate_point_to_layer_space(&self, point: &LayoutPoint)
```
kvark: hmm, this one makes me worried. The function is supposed to translate a point from one space to another, yet our type system does not reflect that here, using `LayoutPoint` on both ends

webrender/src/frame.rs
```diff
@@ -236,7 +236,7 @@ impl Frame {
         old_layer_scrolling_states
     }
 
-    pub fn get_scroll_layer(&self, cursor: &Point2D<f32>, scroll_layer_id: ScrollLayerId)
+    pub fn get_scroll_layer(&self, cursor: &LayerPoint, scroll_layer_id: ScrollLayerId)
```
glennw: cursor is in world space.

webrender/src/frame.rs
nical: I have proposed the addition of vector classes in euclid (separate from points which are currently used as vectors in a lot of places when size types are not used instead). The pull request is open and hopefully we can bikeshed and maybe merge it by the end of the workweek.
Let's wait a little bit before fixing the insistent point-vs-size-for-vectors situation.

webrender_traits/src/api.rs
nical: Good catch, I think that it should take a WorldPoint and return a LayoutPoint. I made the change locally, that said the implementation is ```panic!("unused api - remove from webrender_traits");``` so I guess it won't stick around for long.

replay/src/main.rs
nical: On the other hand this is the naming convention used for all geometry in Gecko which is a different kind of consistency but an appreciable one if gecko and servo are to share more code in the future.
The rationale is that we tend to manipulate float-based geometry more often so they get the short name. Typically there is no need for, say, a LayoutIntPoint (and we probably don't really need DeviceUintPoint as the choice between DeviceIntPoint and DeviceUintPoint is a bit inconsistent right now. In gecko we just used signed integers and floats for geometries).
I'm not found of typing "<>" brackets all over the place but I admit that's hardly an argument.

Let's open a separate issue about this and change it if enough people feel strongly one way or another.

glennw:
Although there's a couple of rough edges in places, this seems like a net improvement - let's get it rebased and then I'll merge it.

glennw:
@nical CI is failing - I guess wrench / sample / replay may need to be updated.

nical:
Ah, wrench landed \o/, I'm rebasing it now.

PR #603, opened 2016-11-28
Add VRCompositor commands for WebVR

MortimerGoro:
Some WebVR commands such as Vsync and SubmitFrame must be called in the WebGL render thread. This PR add traits required to run some render commands from the WebVR implementation. WebVR implementation and vendor VR SDK calls are decoupled using these traits and ipc-channel commands.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/603)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -20,6 +20,7 @@ use webrender_traits::{ApiMsg, AuxiliaryLists, BuiltDisplayList, IdNamespace, Im
 use webrender_traits::{FlushNotifier, RenderNotifier, RenderDispatcher, WebGLCommand, WebGLContextId};
 use webrender_traits::{DeviceIntSize};
 use webrender_traits::channel::{PayloadHelperMethods, PayloadReceiver, PayloadSender, MsgReceiver};
+use webrender_traits::{VRCompositorCommand, VRCompositorHandler};
 use tiling::FrameBuilderConfig;
 use offscreen_gl_context::GLContextDispatcher;
 
@@ -50,6 +51,8 @@ pub struct RenderBackend {
     main_thread_dispatcher: Arc<Mutex<Option<Box<RenderDispatcher>>>>,
 
     next_webgl_id: usize,
+
+    vr_compositor_handler: Arc<Mutex<Option<Box<VRCompositorHandler>>>>
 }
 
 impl RenderBackend {
@@ -66,7 +69,8 @@ impl RenderBackend {
                config: FrameBuilderConfig,
                debug: bool,
                enable_recording:bool,
-               main_thread_dispatcher:  Arc<Mutex<Option<Box<RenderDispatcher>>>>) -> RenderBackend {
+               main_thread_dispatcher: Arc<Mutex<Option<Box<RenderDispatcher>>>>,
+               vr_compositor_handler: Arc<Mutex<Option<Box<VRCompositorHandler>>>>) -> RenderBackend {
 
         let resource_cache = ResourceCache::new(texture_cache,
                                                 enable_aa);
@@ -89,6 +93,7 @@ impl RenderBackend {
             enable_recording:enable_recording,
             main_thread_dispatcher: main_thread_dispatcher,
             next_webgl_id: 0,
+            vr_compositor_handler: vr_compositor_handler
         }
     }
 
@@ -336,6 +341,10 @@ impl RenderBackend {
                             ctx.make_current();
                             ctx.apply_command(command);
                             self.current_bound_webgl_context_id = Some(context_id);
+                        },
+
+                        ApiMsg::VRCompositorCommand(context_id, command) => {
+                            self.handle_vr_compositor_command(context_id, command);
                         }
                         ApiMsg::GenerateFrame => {
                             let frame = profile_counters.total_time.profile(|| {
@@ -468,6 +477,20 @@ impl RenderBackend {
         //           cleaner way to do this, or use the OnceMutex on crates.io?
         let mut notifier = self.notifier.lock();
         notifier.as_mut().unwrap().as_mut().unwrap().new_scroll_frame_ready(composite_needed);
+    }
+
+    fn handle_vr_compositor_command(&mut self, ctx_id: WebGLContextId, cmd: VRCompositorCommand) {
+        let texture = match cmd {
+            VRCompositorCommand::SubmitFrame(..) => {
+                    match self.resource_cache.get_webgl_texture(&ctx_id).texture_id {
+                        SourceTexture::WebGL(texture_id) => Some(texture_id),
+                        _=> None
+                    }
+            },
+            _ => None
+        };
+        let mut handler = self.vr_compositor_handler.lock();
+        handler.as_mut().unwrap().as_mut().unwrap().handle(cmd, texture);
     }
 }
 
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -40,6 +40,7 @@ use webrender_traits::{ColorF, Epoch, FlushNotifier, PipelineId, RenderNotifier,
 use webrender_traits::{ExternalImageId, ImageFormat, RenderApiSender, RendererKind};
 use webrender_traits::{DeviceSize, DevicePoint, DeviceIntPoint, DeviceIntSize, DeviceUintSize};
 use webrender_traits::channel;
+use webrender_traits::VRCompositorHandler;
 
 pub const MAX_VERTEX_TEXTURE_WIDTH: usize = 1024;
 
@@ -419,6 +420,10 @@ pub struct Renderer {
 
     /// Map of external image IDs to native textures.
     external_images: HashMap<ExternalImageId, TextureId, BuildHasherDefault<FnvHasher>>,
+
+    // Optional trait object that handles WebVR commands.
+    // Some WebVR commands such as SubmitFrame must be synced with the WebGL render thread.
+    vr_compositor_handler: Arc<Mutex<Option<Box<VRCompositorHandler>>>>
 }
 
 impl Renderer {
@@ -680,6 +685,9 @@ impl Renderer {
         let backend_main_thread_dispatcher = main_thread_dispatcher.clone();
         let backend_flush_notifier = flush_notifier.clone();
 
+        let vr_compositor = Arc::new(Mutex::new(None));
+        let backend_vr_compositor = vr_compositor.clone();
+
         // We need a reference to the webrender context from the render backend in order to share
         // texture ids
         let context_handle = match options.renderer_kind {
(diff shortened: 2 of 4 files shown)


webrender/src/renderer.rs
```diff
@@ -414,6 +415,10 @@ pub struct Renderer {
 
     /// Map of external image IDs to native textures.
     external_images: HashMap<ExternalImageId, TextureId, BuildHasherDefault<FnvHasher>>,
+
+    // Optional trait object that allows to create a VRCompositor instance to handle WebVR commands.
+    // Some WebVR commands such as Vsync and SubmitFrame must be called in the WebGL render thread.
+    vr_compositor_creator: Arc<Mutex<Option<Box<VRCompositorCreator>>>>
```
kvark: This compositor creator is referenced from both the render backend and the renderer. And it looks like the only reason it's needed on the renderer is for `set_vr_compositor_creator`. Can we assume the compositor creator is known at the init time? In this case, we might be able to just send it over and store without all the `Arc`/`Mutex` wrappers.

webrender/src/render_backend.rs
```diff
@@ -470,6 +481,38 @@ impl RenderBackend {
...
+                if let Some(ref compositor) = self.vr_compositors.get(&compositor_id) {
+                    let pose = compositor.sync_poses(near, far);
+                    let _result = sender.send(Ok(pose));
+                } else {
+                    let _result = sender.send(Err(()));
+                }
+            }
+            VRCompositorCommand::SubmitFrame(compositor_id, left_bounds, right_bounds) => {
+                if let Some(ref compositor) = self.vr_compositors.get(&compositor_id) {
+                    let texture = self.resource_cache.get_webgl_texture(&ctx_id);
```
kvark: The whole VR logic seems to be mostly independent of the WR, except for this `resource_cache` access. Do the compositors and their compositor creator have to live within the WR backend/renderer at all? Perhaps, we could narrow down the interface between those and reduce the integration.

webrender/src/renderer.rs
MortimerGoro: I only need access to the compositor creator from the render_backend. I could add an Option<Box<VRCompositorCreator>> to the RendererOptions struct if that's not a problem. I followed other setter patterns like set_render_notifier, set_render_notifier  and kept RendererOptions intact. VRCompositor creator is know at init time and won't change, so we can avoid the Arc/Mutex wrappers.

Let me know if you like the RenderOptions change and I'll update this PR.

kvark: Could you explain a bit more on what option you want to add to `RenderOptions`?

webrender/src/render_backend.rs
MortimerGoro: Besides the resource_cache access to get the current texture_id (which might change after a webgl resize) another requirement is that specified VRCompositor calls must be run in the same thread as the WebGL commands. There could be many different VRCompositor implementations for different VR vendors, and some of them require to be created/initialized in the render thread too (because they use the shared OpenGL texture and the active OpenGL context).

These functions must run in response to some ipc-channel messages from JavaScript and we have to minimize the render path to achieve WebVR latency requirements. There is no way in webrender to run generic closures because of serde-serialization and sandboxed processes, so I implemented it using the traits. I'm about to publish a blogpost with more info about the overall WebVR implementation architecture, I'll send you the link. If you have some ideas to narrow down this interface while keeping the latency requirements that would be great.

webrender/src/renderer.rs
MortimerGoro: The compositor creator instance, something like this: 

```
pub struct RendererOptions {
    pub device_pixel_ratio: f32,
    pub resource_override_path: Option<PathBuf>,
    pub enable_aa: bool,
    pub enable_msaa: bool,
    pub enable_profiler: bool,
    pub debug: bool,
    pub enable_recording: bool,
    pub enable_scrollbars: bool,
    pub precache_shaders: bool,
    pub renderer_kind: RendererKind,
    pub enable_subpixel_aa: bool,
    pub vr_compositor_creator: Option<Box<VRCompositorCreator>>>
}

```

MortimerGoro: After that, In the Renderer constructor send the instance to RenderBackend::new() and remove Arc/Mutex wrappers.

kvark: I suppose that to be good, yeah. Interested in what @glennw thinks.

webrender/src/render_backend.rs
kvark: > specified VRCompositor calls must be run in the same thread as the WebGL commands

I understand this. But from WR's point of view, the user creates the `Renderer`, so the user is fully aware (and in control) of the thread that executes WebGL commands. Can't the user just manage the compositors (with their creator) then, only interacting with WR via the resource cache, if needed?

Disclaimer: I'm still learning the WR way of things, so I might just be missing the obvious answer.

MortimerGoro: > I understand this. But from WR's point of view, the user creates the Renderer, so the user is fully aware (and in control) of the thread that executes WebGL commands. 

The user creates the Renderer, but the Renderer internally creates the renderer_backend thread so the user has not really a direct access to that inner thread. At least I need a trait to run "hard code" because that thread can only can be accessed through predefined ipc-channel messages which have some limitations (can't send closures, complex objects, pointers, and more).

> Can't the user just manage the compositors (with their creator) then, only interacting with WR via the resource cache, if needed?
> 
> Disclaimer: I'm still learning the WR way of things, so I might just be missing the obvious answer.

Yes, I understand what you mean. I could reduce the current VRCompositor interfaces to a single vr_compositor_handler trait, that uses bytes (vec<u8> based messages) to forward-serialize all possible messages without defining them in WR. It has the pros of a less decoupled interface but it's a more obscure API. If that's the style you want in WR I can update the PR. I'm also interested in what @glennw thinks about both options.

The trait I'm talking about would look like this: 

```
// Single instance in render_backend
pub trait VRCompositorHandler {
    fn handle(&self, data: vec<u8>, sender:Option<IpcSender<vec<u8>>>);
}
```

```
// IPC-Channel message that uses the VRCompositorHandler instance
VRCompositorCommand(data: vec<u8>, sender:Option<IpcSender<vec<u8>>>)

```

kvark: > The user creates the Renderer, but the Renderer internally creates the renderer_backend thread so the user has not really a direct access to that inner thread.

Yes, but why do you need this thread? VRCompositor calls must be done on the same thread that `Renderer` is, not `RenderBackend`.

> I could reduce the current VRCompositor interfaces to a single VRCompositorHandler trait, that uses bytes (vec<u8> based messages) to forward-serialize all possible messages without defining them in WR

Oh, no, please no `Vec<u8>` messages if we have this option.

glennw:
@MortimerGoro I'm a little confused about this too, with the same question @kvark had.

The GL calls get issued on the Render thread, the RenderBackend thread never deals directly with a GL context or issues GL commands. Does this PR assume that there is GL work done on the backend thread, or am I misunderstanding the review comments so far?

MortimerGoro:
> Yes, but why do you need this thread? VRCompositor calls must be done on the same thread that Renderer is, not RenderBackend.

> The GL calls get issued on the Render thread, the RenderBackend thread never deals directly with a GL context or issues GL commands. Does this PR assume that there is GL work done on the backend thread, or am I misunderstanding the review comments so far?

Because of these reasons:

- VRCompositor calls must be totally synced with the WebGL calls. Renderbackend is the thread that runs WebGL commands now (https://github.com/servo/webrender/blob/master/webrender/src/render_backend.rs#L335). Running the functions in that thread provides the best latency to avoid motion sickness. Another requirement is that a JavaScript user can submit frames to the headset in the middle of a JavaScript animation call and then perform another WebGL commands that are not rendered to the headset on that frame. So synchronization with the WebGL calls is a must.

- VR headsets have it's own Vsync mechanism that might conflict with the main renderer Vsync. The spec allows to both render to a headset (90fps) and render to desktop window (60 fps) at the same time. We can have a requestAnimationFrame for a headset and a requestAnimationFrame for the window. Doing both on the main render thread would be a problem.

> Oh, no, please no Vec<u8> messages if we have this option.

glennw:
@MortimerGoro You're quite right, my apologies - I thought the webgl commands were proxied to the renderer thread to be applied there.

It's certainly not ideal to have the webgl commands replayed in the backend thread, but I'm not sure what the best alternative would be right now, or if it has to be this way.

@emilio Do you recall why they are replayed in the backend thread?

emilio:
@glennw I don't think there's a specific reason. Seemed easier to not pollute the main thread with WebGL state, plus the backend thread is the one that is directly aware of publishing a frame, so all the WebGL cleanup that needs to run before that was easier done from there.

We can move it to another thread if you wish, though it's going to be a fair amount of sync messaging for no good reason IMO.

MortimerGoro:
@glennw @emilio FYI when I saw the current implementation I thought It was done willingly that way for a better parallelism and as a sandboxed thread to avoid malicious input in the main thread (WebGL allows to execute arbitrary code on the GPU, forgetting a proper makeCurrent could be very dangerous).

IMO before doing any changes we should evaluate a lot of things to get the best out of WebGL:
- Performance
- Security
- Compatibility with WebGL on Webworkers
- Proper Synchronization (see https://github.com/servo/servo/issues/14235)
- WebVR latency and vsync loop.

I'll press this button (http://nooooooooooooooo.com/) if there are breaking changes before submitting the WebVR implementation I've been working on XD

glennw:
@MortimerGoro Sure, there's a lot of questions here, and it makes sense to consider all of these before making any changes :)

My main concern at the moment is that executing a vsync command in the backend thread sounds like it could cause performance issues (since the main job of the backend thread is to be preparing command buffers for the render thread to execute, and responding to queries from layout threads quickly).

Having said that, I'm happy to merge the WebVR stuff (after review of course) as is, so long as we have a plan to investigate / resolve issues like I mentioned above, and so long as it doesn't have any adverse effects when WebVR isn't active.

Apart from this PR, is there a lot more code to come that directly affects WR?

MortimerGoro:
> Having said that, I'm happy to merge the WebVR stuff (after review of course) as is, so long as we have a plan to investigate / resolve issues like I mentioned above, and so long as it doesn't have any adverse effects when WebVR isn't active.

@glennw Of course. I'd love to help investigating the best way of fixing that issue and improving WebGL performance.

> Apart from this PR, is there a lot more code to come that directly affects WR?

WR only requires this PR ;) Most of the code will be added in another PR to Servo and in a separate library (https://github.com/MortimerGoro/rust-webvr). I'm looking forward of sharing the blogpost with the description of the overall architecture.

MortimerGoro:
Updated PR. I've narrowed down the interfaces required in WR and moved more code to the WebVR component in Servo.

@kvark I didn't add the VRCompositorHandler instance to RenderOptions because it requires Clone and doesn't play well with the unsized trait.

glennw:
Opened https://github.com/servo/webrender/issues/607 as a follow up.

PR #604, opened 2016-11-28
GPU markers support

### LINKED ISSUE #557: Annotate passes/draw calls
OpenGL ES may support `KHR_Debug` and/or `EXT_debug_marker`, as mentioned on the apitrace usage page: https://github.com/apitrace/apitrace/blob/master/docs/USAGE.markdown#opengl-annotations
By adding the annotations we could greatly enhance the debugging experience.

kvark:
Fixes #557

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/604)


webrender/src/device.rs
```diff
@@ -639,6 +639,36 @@ impl<T> GpuProfiler<T> {
     }
 }
 
+pub struct GpuMarker(());
```
glennw: Can this just be pub struct GpuMarker; ?

webrender/src/renderer.rs
```diff
@@ -1076,6 +1077,7 @@ impl Renderer {
                    target_size: &DeviceSize,
                    cache_texture: Option<TextureId>,
                    should_clear: bool) {
+        let _ = GpuMarker::new("target draw");
```
glennw: Instead of these hard coded strings, could we use the label field that exists in the GpuProfileTag structure?

kvark: Oh, I missed the fact that those have names. Going to have the `add_marker` returning a `GpuMarker` object then, unless you think of a better way ;)

webrender/src/device.rs
kvark: No, since it's unsafe to create an instance of it without calling `GpuMarker::new()`

kvark:
Updated now. How does this look?

Issue #601, opened 2016-11-28
Allowing for transparent background color (to be allowed)


jaredpetker:
I am trying to use webrender itself outside of the context of servo, and found that there isn't currently a way to simply have a fully transparent root background color. There seems to be some [comments around such things here](https://github.com/servo/webrender/blob/master/webrender/src/frame.rs#L640).

It seems like handing in a fully transparent root background color is not allowable at the moment, basically my use case is drawing on top of something else with webrender, so I don't actually want to a) draw any background color or b) clear the background color at all, I just want to draw on top of some other stuff.

Any thoughts on adding some option which would allow for such behavior or how such behavior would like to be added? I've no issue adding the feature myself to help.

nical:
I think that it should be fine to remove the forced white background and make sure that servo specifies an opaque background color for its root stacking context to get the same effect. Gecko will eventually also need to be able to draw on top of a semi-transparent widget.

What do you thing @kvark?

kvark:
Yes, that should be fine.

glennw:
@jaredpetker Yup, adding optional functionality to skip any clear / background rect sounds good to me.

For now, it could probably just be a field in the RendererOptions struct that gets passed to WR during initialization. This would allow you to get the functionality working, and then we could consider a public API for it later if we ever want to change that option dynamically?

jaredpetker:
I can go with that. I've found everywhere that seems to do clearing (the initial stacking context bounds rect, tiling, etc), will do, thanks.

glennw:
I think this is fixed now - feel free to re-open if further work is required.

PR #605, opened 2016-11-29
Add more control over when and how to clear the framebuffer.

nical:
This pull request addresses issue #601. The RendererOptions now specify a default clear color for the main framebuffer (intermediate targets are cleared with transparent black in any case), an option to always clear using the device's clear function, and a separate option to use the clear shader for empty tiles (previously referred to as "clear tiles"). Note that the clear shader ignores the clear color and always uses opaque white.

I don't know if we want to keep the clear shader around, I am pretty sure it is currently never used because of the enforced background rectangle in frame.rs. If we want to get rid of it, I can add that to the PR (I don't know if we'd keep the information about empty tiles or not, probably should?), or we can leave it and decide later.

Edit: Removed confusion about non-root pipelines.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/605)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -651,41 +651,28 @@ impl Frame {
         };
 
         if level == 0 {
-            // Add a large white rectangle as the root display item if there is no root stacking
-            // context background color. This is removed by the occlusion culling for most tiles,
-            // and means that it's no longer necessary to clear the framebuffer.
-            //
-            // TODO(nical) Should painting a white background be optional if there is no stacking
-            // context background color? On deferred GPUs we probably still want to clear the
-            // framebuffer and Gecko currently supports semi-transparent windows.
-            //
-            // If we do need this, does it make sense to keep Frame::clear_tiles?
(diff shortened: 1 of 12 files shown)


webrender/src/frame.rs
```diff
@@ -655,42 +655,27 @@ impl Frame {
...
+                context.builder.push_layer(LayerRect::from_untyped(&clip_region.main),
+                                           &no_clip,
+                                           transform,
+                                           pipeline_id,
+                                           scroll_layer_id,
+                                           &composition_operations);
+
+                //Note: we don't use the original clip region here,
+                // it's already processed by the layer we just pushed.
+                context.builder.add_solid_rectangle(&LayerRect::from_untyped(&clip_region.main),
```
nical: Woops, this is not correct (passing the entire scene rect instead of the pipeline's rect), I'm looking into fixing it.

nical:
cc @kvark @glennw @jaredpetker

glennw:
@nical Could you rebase this, if we're planning to use this PR instead of the other one?

PR #606, opened 2016-11-29
Adding clear_method as a RendererOptions

jaredpetker:
This adds ClearMethod flags which can be used to select what type of method for clearing would like to be used during rendering, and adds an option to RendererOptions.

I went with flags vs just a boolean for clearing (yes / no) as I found there seemed to be different clears going on throughout the rendering (CLEAR_ALL, CLEAR_BACKGROUND, CLEAR_TILES, CLEAR_NONE).  However, I don't currently have a great reason on why you would choose to only clear the background and not do tile clearing. My need was for the CLEAR_NONE option, however I thought this might be more useful than just a boolean. Happy to reconsider this via someone that is more familiar with the current direction of the project.

[Issue Reference](https://github.com/servo/webrender/issues/601)

@nical  @kvark @glennw (tagging those that posted on the issue this far)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/606)

### DIFF
--- a/replay/src/main.rs
+++ b/replay/src/main.rs
@@ -108,6 +108,7 @@ fn main() {
         renderer_kind: webrender_traits::RendererKind::Native,
         debug: false,
         enable_subpixel_aa: false,
+        clear_method: webrender::CLEAR_ALL
     };
 
     let (mut renderer, sender) = webrender::renderer::Renderer::new(opts);
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -86,6 +86,7 @@ fn main() {
         precache_shaders: true,
         renderer_kind: RendererKind::Native,
         enable_subpixel_aa: false,
+        clear_method: webrender::CLEAR_ALL
     };
 
     let (mut renderer, sender) = webrender::renderer::Renderer::new(opts);
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -16,6 +16,7 @@ use scene::Scene;
 use std::collections::{HashMap, HashSet};
 use std::hash::BuildHasherDefault;
 use tiling::{AuxiliaryListsMap, FrameBuilder, FrameBuilderConfig, LayerMap, PrimitiveFlags};
+use renderer::{CLEAR_BACKGROUND};
 use util::MatrixHelpers;
 use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLayerId};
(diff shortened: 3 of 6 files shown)


nical:
See also PR #605. I don't mind going for this one over PR #605 but I would prefer that the decision to add or not an opaque rectangle behind each pipeline be taken by passing the pipeline's background color as an option on a per-pipeline basisis rather than the flags in the renderer options, otherwise it means that depending on the flags you pass, a parameter of set_root_displaylist may or may not be ignored which is error prone. The flag in the renderer options should still control whether the renderer calls glClear on the main framebuffer

jaredpetker:
Ah, yea, I see the overlap between the two PR's. I can look into taking your comment / PR into account and making the alterations here.

nical:
@jaredpetker or you can rebase on top of PR #605 , (although I think it already has everything you need). It incorporates controlling the backrounds at the pipeline level. It also lets you disable the clear shader which makes sense if the clear color is not opaque white (for example to draw on a semitransparent target) or when running on a tiled-deferred gpu where we are better off using glClear.

jaredpetker:
Can do makes sense. @nical is your thought basically then taking what you have but keeping the bitflags approach over the booleans used in that pr?

To be transparent (no pun intended) on my use case, I'm wanting to use webrender to essentially draw over existing graphics, which is why I'm looking for options to take away all clearing that is done.

nical:
As far as I'm concerned, feel free to replace the booleans for bitflags in the RendererOptions I don't mind either way (the reviewer here is @glennw).
On top of PR #605, for your use case, just set clear_empty_tiles and clear_framebuffer to false (or the equivalent bitflags), and pass None for the background_color of the pipelines that need to have transparent backgrounds.

jaredpetker:
Ah interesting got it, let me just try out the other PR myself then and confirm it works for my use case, if that's the case then perhaps there's no need for mine here.

glennw:
@jaredpetker OK, I've r+'ed the other similar PR to this. Once you confirm it contains the functionality you need and is working, I guess we can close this one? Thanks for working with @nical to get this functionality added! :)

glennw:
Closing since similar functionality has merged in another PR. Please re-open if this is still required.

jaredpetker:
Missed the comments, was away a bit, have just been using my fork for my use case, haven't tried out the (what is now) merged, but I think it will do what I need / can add what is needed if that's not the case, thanks.

PR #609, opened 2016-11-30
PrimitiveBatchData for all

### LINKED ISSUE #456: Optimize the PrimitiveBatchData enum
This enum has a Vec<PrimitiveInstance> that is the same for many batch types. Previously, they were separate types. We could unify this enum to use a single Vec<PrimitiveInstance> and then store offsets / lengths for each batch. This would allow us to change the renderer to create far fewer UBOs which would improve the speed of the compositor time.

kvark:
Fixes #456 

More stuff to come.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/609)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -1415,20 +1415,30 @@ pub struct PrimitiveInstance {
     user_data: [i32; 2],
 }
 
-#[derive(Debug, Clone)]
-pub struct PackedBlendPrimitive {
-    src_task_id: i32,
-    target_task_id: i32,
-    op: i32,
-    amount: i32,
-}
+impl PrimitiveInstance {
+    fn blend(src_task_id: i32, target_task_id: i32, op: i32, amount: i32) -> PrimitiveInstance {
+        PrimitiveInstance {
+            global_prim_id: -1,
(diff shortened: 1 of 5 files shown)


webrender/src/tiling.rs
```diff
@@ -1505,13 +1528,10 @@ impl PrimitiveBatch {
                       info: MixBlendMode) -> bool {
         match &mut self.data {
             &mut PrimitiveBatchData::Composite(ref mut ubo_data) => {
-                ubo_data.push(PackedCompositePrimitive {
```
glennw: Should be able to remove the PackedCompositePrimitive struct now.

webrender/src/tiling.rs
```diff
@@ -1485,13 +1511,10 @@ impl PrimitiveBatch {
                     LowLevelFilterOp::Opacity(amount) => (8, amount.to_f32_px()),
                 };
 
-                ubo_data.push(PackedBlendPrimitive {
```
glennw: Should be able to remove the PackedBlendPrimitive struct now.

kvark:
All done.

Issue #457, opened 2016-10-21
Remove UBO usage.


glennw:
Instead of using UBOs for submitting primitive instances and using gl_InstanceID, with the recent changes to prim_store, it would be better now to submit the PrimitiveInstance structures as instanced vertex attributes.

glennw:
This may not be worth doing - it seems that instanced vertex attributes are only available in GL3.3+, and we want to support GL 3.1 for some older Intel GPU systems. If that's correct, then we can close this and stick with using UBOs. @jrmuizel or @kvark does that sound correct to you?

kvark:
@glennw I disagree.

I don't think `ARB_instanced_arrays` support requirement limits us in any way. On Intel, 5th gen GPUs only support GL 2.1 (according to https://en.wikipedia.org/wiki/List_of_Intel_graphics_processing_units), while 6th gen (Sandy Bridge, starting 2011) already fits our needs:
>3.1 on Windows
3.3 on OS
3.3 on Linux

On Windows we are using Direct3D anyway, which has [InstanceDataStepRate field](https://msdn.microsoft.com/en-us/library/windows/desktop/bb205316(v=vs.85).aspx) as of version 10. And on Linux/OSX we got GL 3.3 core support.

In general, if we are talking about GL extensions and not core GL versions, I haven't seen a system that would support instanced draw calls but not instanced arrays.

As for GLES, version 3.0 has our stuff included.

PR #610, opened 2016-11-30
Vertex attributes cleanup

kvark:
This is the preparation work for #457

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/610)

### DIFF
--- a/webrender/res/debug_color.vs.glsl
+++ b/webrender/res/debug_color.vs.glsl
@@ -2,9 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+in vec4 aColor;
+varying vec4 vColor;
+
 void main(void)
 {
-    vColor = aColorRectTL;
+    vColor = aColor;
     vec4 pos = vec4(aPosition, 1.0);
     pos.xy = floor(pos.xy * uDevicePixelRatio + 0.5) / uDevicePixelRatio;
     gl_Position = uTransform * pos;
--- a/webrender/res/debug_font.vs.glsl
+++ b/webrender/res/debug_font.vs.glsl
@@ -2,10 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+in vec4 aColor;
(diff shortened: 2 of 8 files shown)


webrender/res/debug_color.vs.glsl
```diff
@@ -2,6 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+in vec4 aColorRectTL;
```
glennw: Let's make this just be aColor

webrender/res/debug_font.vs.glsl
```diff
@@ -2,6 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+in vec4 aColorRectTL;
+in vec4 aColorTexCoordRectTop;
```
glennw: Let's just have aColor and aUv0 or something similar.

webrender/src/internal_types.rs
```diff
@@ -225,21 +225,8 @@ pub const DEFAULT_TEXTURE: TextureSampler = TextureSampler::Color0;
 
 pub enum VertexAttribute {
     Position,
-    PositionRect,
     ColorRectTL,
```
glennw: Let's rename these to Color and Uv0 or something like that.

webrender/src/renderer.rs
```diff
@@ -308,7 +308,7 @@ fn create_clear_shader(name: &'static str,
                        max_ubo_vectors: usize) -> ProgramId {
     let prefix = format!("#define WR_MAX_UBO_VECTORS {}", max_ubo_vectors);
 
-    let includes = &["shared_other"];
```
glennw: Should we delete shared_other.glsl - is it ever used now?

kvark: That's what I did first, but then - `Device::create_program` requires a shared header, and so I decided to let it live for now :)

webrender/res/debug_color.vs.glsl
kvark: Yep, I have it on my list. Clearly, TL/RectTop suffices are misleading with this change

kvark:
Should be ready now.

PR #611, opened 2016-12-01
delete flush notification api

changm:
Since we use epochs for reftests like servo now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/611)


glennw:
Thanks!

PR #615, opened 2016-12-02
Instanced attributes

### LINKED ISSUE #457: Remove UBO usage.

kvark:
Closes #457 

Performance-wise, I wasn't able to register a noticeable difference. Tested on https://github.com/servo/servo, full screen after the second page down, on `Mesa DRI Intel(R) HD Graphics 5500 (Broadwell GT2)` with resolution 2560x1440.

With the change, I got 4.5 ms mean GPU time in the first test, and 4.3 ms time in the second.
Without the change, I got 4.4 ms mean GPU time in the only test.

I suppose the difference is not visible since we are far from being VS-bound.

Note: this does not replace the optimization of having one large buffer in #456. Instanced attributes will benefit from it in the same way as UBOs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/615)


glennw:
@kvark Is this still WIP or ready for review?

kvark:
Ready for review and landable, but I'm going to convert the rest of the shaders and eliminate the UBO completely.

glennw:
@kvark OK - would you prefer we land this after review, or wait until the UBO removal is done?

kvark:
Let me remove the UBOs completely in this PR. Meanwhile, @jamienicol is going to check for the HW extensions support on GLES w.r.t. the instanced attributes.

kvark:
The PR is ready for review now.
I've made more perf tests on the servo page and barely noticed any changes. The GPU time is about 3.5ms, the Compositor time jumps a bit, with the minimum bound about 0.80ms for the original version and 0.75ms for the new one. This is well within the measurement error threshold.

glennw:
Reviewed 23 of 24 files at r1.
Review status: 23 of 24 files reviewed at latest revision, 1 unresolved discussion.

---

*[webrender/src/device.rs, line 190 at r1](https://reviewable.io:443/reviews/servo/webrender/615#-KYp_h1OIOX9XMCMpT2B:-KYp_h1OIOX9XMCMpT2C:b-z6mmje) ([raw file](https://github.com/servo/webrender/blob/740b90d48eb6207d8b9fb464d82e24f8ddc64c3b/webrender/src/device.rs#L190)):*
> ```Rust
>                                             gl::INT,
>                                             instance_stride,
>                                             24);
> ```

Can we accumulate a variable here? Otherwise it'll be really easy to add an item to the array above and forget to update this value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/615)*

glennw:
@kvark Looks good, just one small issue to address.

kvark:
@glennw thanks for having a look. The offset is now calculated, and I agree this is better.

kvark:
Rebased now.

PR #625, opened 2016-12-12
Fix for some new intermittent ref test failures.

glennw:
This ensures that text run glyphs are added to the texture cache
in the same order between runs, to remove a source of floating
point accuracy issues in the rasterizer.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/625)


pcwalton:
I am so sorry.

@bors-servo: r+

PR #627, opened 2016-12-12
Fix default precision of `sampler2DArray`

mmatyas:
This fixes the `no default precision defined for variable 'sCache'` error produced by Servo on Android.

Related: servo/servo#13154

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/627)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -3,6 +3,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#ifdef GL_FRAGMENT_PRECISION_HIGH
+precision highp sampler2DArray;
+#else
+precision mediump sampler2DArray;
+#endif
+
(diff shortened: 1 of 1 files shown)


glennw:
Is GL_FRAGMENT_PRECISION_HIGH defined on non-ES (i.e. desktop implementations)? It might be best to wrap it with GL_ES defined too?

kvark:
FYI, @glennw I don't think it matters:
> Precision qualifiers in GLSL are supported for compatibility with OpenGL ES 2.0. They use the same syntax as ES 2.0's qualifiers, but they have no functional effects.

glennw:
@kvark OK, that should be fine then.

kvark:
It appears to be failing now on Windows/AMD:
> ERROR: 0:6: error(#228) Type should be float or int

It doesn't like precision qualifier for `sampler2DArray`

glennw:
@kvark OK, let's wrap it in GL_ES then?

kvark:
@glennw yes, your intuition was correct, let's wrap it up

PR #636, opened 2016-12-14
Handle external image life time in api.delete_image() call

JerryShih:
Currently, the delete_image() call just removes the image_key in the hash_map. If the key is associate to an external image, we should release that image using the ExternalImageHandler callbacks.

Since the ExternalImageHandler callbacks are only used in renderer, a pending releasing array is used. Then, pass this array to the renderer for ExternalImageHandler::release() call.

@nical @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/636)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1262,7 +1262,7 @@ impl Renderer {
                 let props = &deferred_resolve.image_properties;
                 let external_id = props.external_id
                                        .expect("BUG: Deferred resolves must be external images!");
-                let image = handler.get(external_id);
+                let image = handler.lock(external_id);
 
                 let texture_id = match image.source {
                     ExternalImageSource::NativeTexture(texture_id) => TextureId::new(texture_id),
@@ -1427,10 +1427,19 @@ pub struct ExternalImage {
     pub source: ExternalImageSource,
 }
 
-/// Interface that an application can implement
-/// to support providing external image buffers.
+/// The interfaces that an application can implement to support providing
+/// external image buffers.
+/// When the the application passes an external image to WR, it should kepp that
+/// external image life time untile the release() call.
 pub trait ExternalImageHandler {
-    fn get(&mut self, key: ExternalImageId) -> ExternalImage;
+    /// Lock the external image. Then, WR could start to read the image content.
+    /// The WR client should not change the image content until the unlock()
+    /// call.
+    fn lock(&mut self, key: ExternalImageId) -> ExternalImage;
+    /// Unlock the external image. The WR should not read the image content
+    /// after this call.
+    fn unlock(&mut self, key: ExternalImageId);
+    /// Tell the WR client that it could start to release this external image.
     fn release(&mut self, key: ExternalImageId);
 }
 
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -16,7 +16,7 @@ use device::{TextureFilter, VAOId, VertexUsageHint, FileWatcherHandler, TextureT
 use euclid::Matrix4D;
 use fnv::FnvHasher;
 use internal_types::{CacheTextureId, RendererFrame, ResultMsg, TextureUpdateOp};
-use internal_types::{TextureUpdateList, PackedVertex, RenderTargetMode};
+use internal_types::{ExternalImageUpdateList, TextureUpdateList, PackedVertex, RenderTargetMode};
 use internal_types::{ORTHO_NEAR_PLANE, ORTHO_FAR_PLANE, SourceTexture};
 use internal_types::{BatchTextures, TextureSampler, GLContextHandleWrapper};
 use profiler::{Profiler, BackendProfileCounters};
@@ -710,10 +710,12 @@ impl Renderer {
         // Pull any pending results and return the most recent.
         while let Ok(msg) = self.result_rx.try_recv() {
             match msg {
-                ResultMsg::UpdateTextureCache(update_list) => {
-                    self.pending_texture_updates.push(update_list);
-                }
-                ResultMsg::NewFrame(frame, profile_counters) => {
+                ResultMsg::NewFrame(frame, texture_update_list, external_image_update_list, profile_counters) => {
+                    self.pending_texture_updates.push(texture_update_list);
+
+                    // When a new frame is ready, we could start to update all pending external image requests here.
+                    self.release_external_images(external_image_update_list);
+
                     self.backend_profile_counters = profile_counters;
 
                     // Update the list of available epochs for use during reftests.
@@ -1277,13 +1279,25 @@ impl Renderer {
         }
     }
 
-    fn release_external_textures(&mut self) {
+    fn unlock_external_images(&mut self) {
         if !self.external_images.is_empty() {
             let handler = self.external_image_handler
                               .as_mut()
                               .expect("Found external image, but no handler set!");
 
             for (external_id, _) in self.external_images.drain() {
+                handler.unlock(external_id);
+            }
+        }
+    }
+
+    fn release_external_images(&mut self, mut pending_external_image_updates: ExternalImageUpdateList) {
+        if !pending_external_image_updates.is_empty() {
+            let handler = self.external_image_handler
+                              .as_mut()
+                              .expect("found external image updates, but no handler set!");
+
+            for external_id in pending_external_image_updates.drain(..) {
                 handler.release(external_id);
             }
         }
@@ -1388,7 +1402,7 @@ impl Renderer {
             }
         }
 
-        self.release_external_textures();
+        self.unlock_external_images();
     }
 
     pub fn debug_renderer<'a>(&'a mut self) -> &'a mut DebugRenderer {
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -275,6 +275,14 @@ impl ResourceCache {
                                  bytes: Vec<u8>) {
         let next_epoch = match self.image_templates.get(&image_key) {
             Some(image) => {
+                // This image should not be an external image.
+                match image.data {
+                    ImageData::External(id) => {
+                        panic!("Update an external image with buffer, id={} image_key={:?}", id.0, image_key);
+                    },
+                    _ => {},
+                }
+
                 let Epoch(current_epoch) = image.epoch;
                 Epoch(current_epoch + 1)
             }
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -6,7 +6,7 @@ use app_units::Au;
 use device::TextureFilter;
 use fnv::FnvHasher;
 use frame::FrameId;
-use internal_types::{FontTemplate, SourceTexture, TextureUpdateList};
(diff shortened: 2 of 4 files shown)


webrender/src/renderer.rs
```diff
@@ -1451,7 +1472,17 @@ pub struct ExternalImage {
 /// Interface that an application can implement
 /// to support providing external image buffers.
 pub trait ExternalImageHandler {
+    // Get the associate external image from the id. It also implies that the
+    // image is held by WR. The WR client will keep the life time for this
+    // image until the release() call.
```
emilio: nit: Please make these doc coments (`///` instead of `//`)

kvark:
I'm not convinced this handling needs to be integrated into WR. Can't we move this responsibility to WR user by doing the following scheme:
  - the user sets the root display list and pipeline, receiving a frame ID that starts going to the render backend for processing. Whatever external resources are involved, the user is able to associate with this frame ID outside of WR being aware.
  - the backend finishes the work, sends the frame to the Renderer
  - the Renderer draws the frame, issues a user callback and passes the frame ID in there.
  - the user can safely assume the external resources associated with this ID are done, and frees them

In this scheme, the only APIs that WR needs to provide are:
  - frame ID
  - user callback

@glennw @JerryShih does it make sense?

webrender/src/renderer.rs
```diff
@@ -740,10 +742,29 @@ impl Renderer {
...
                     self.pending_texture_updates.push(update_list);
+
+                    if self.external_image_handler.is_some() {
+                        self.pending_external_image_updates.push(external_image_update_list);
+                    }
                 }
                 ResultMsg::NewFrame(frame, profile_counters) => {
+                    // When a new frame is ready, we could start to update all pending external
+                    // image request here.
+                    if !self.pending_external_image_updates.is_empty() {
```
glennw: I think it's probably more "correct" to move this loop to the end of the render() call.

webrender/src/renderer.rs
```diff
@@ -740,10 +742,29 @@ impl Renderer {
...
+                        let handler = self.external_image_handler
+                                          .as_mut()
+                                          .expect("Found external image updates, but no handler set!");
+                        let mut pending_external_image_updates = mem::replace(&mut self.pending_external_image_updates, Vec::new());
```
glennw: This line shouldn't be necessary - the loop could just be ```self.pending_external_image_updates.drain(..)``` I think.

webrender/src/resource_cache.rs
```diff
@@ -294,7 +297,17 @@ impl ResourceCache {
     }
 
     pub fn delete_image_template(&mut self, image_key: ImageKey) {
-        self.image_templates.remove(&image_key);
+        let value = self.image_templates.remove(&image_key);
+
+        // If the key is associate to an external image, pass the external id to render for cleanup.
+        if let Some(image) = value {
```
glennw: Let's print a error if image is not Some(..) here as well, since that indicates a bug (such as double deletion of an image key).

JerryShih:
@kvark 
> the user sets the root display list and pipeline, receiving a frame ID that starts going to the render backend for processing. Whatever external resources are involved, the user is able to associate with this frame ID outside of WR being aware.
> the backend finishes the work, sends the frame to the Renderer
> the Renderer draws the frame, issues the user callback and passes the frame ID in there.
> the user can safely assume the external resources associated with this ID are done, and frees them

If we call Render::update() and Render::render() multiple times without update the display_list, when should we release the resources? The user will receive several callback with the same frame id.

webrender/src/renderer.rs
JerryShih: @glennw 
We could release the unused external image asap here. These external images will not be used because we set a new frame. If we move the deletion to the end of render() call, we just extend the life-time of these images.
What do you think for this reason?

glennw: I *think* what could happen is this sequence:

```
dl_builder.add_image(ext_image);
api.add_dl(dl_builder).
api.delete_image(blah)
```

In this case, it seems possible to me that you could have an image in the current frame's display list, which is also in the pending release list? That's why it seems safer to me to do it at the end of render(). The update() and render() are called directly after each other, so I don't think it will have any major impact on extra lifetimes. Does that seem possible to you? (I might be over-thinking it).

glennw:
@kvark I think it's probably fine to have this code in WR. Even though it's probably possible for the user to handle it externally, it seems like we could make it a safer interface by managing this internally. Thoughts?

webrender/src/renderer.rs
JerryShih: api.add_image(A)
dl.push_image() with A
set_root_dl(frame 1)
update()
render() <= the frame 1 is not ready, draw frame 0 instead
api.delete_image(A)
api.add_image(B)
dl.push_image with B
update()
render() <= the frame 1 is ready now, draw with A
-----
So, we the image A will still be used until frame 2 ready. That's why I put the release() call when we receives the new frame.

JerryShih: > dl_builder.add_image(ext_image);
> api.add_dl(dl_builder).
> api.delete_image(blah)

If we have this usage, it will have a panic in block_until_all_resources_added(). So, that problem is not related to this patch.

JerryShih:
@glennw 
I address the review comments in patch 3-6.

JerryShih:
@kvark , could you please review my updates? Glenn is on PTO.

kvark:
@JerryShih 
> If we call Render::update() and Render::render() multiple times without update the display_list, when should we release the resources? The user will receive several callback with the same frame id.

Right, but this is easily workable. You could free the resources associated with frame N when you get a callback of frame N+1.

> could you please review my updates? Glenn is on PTO.

Will make sure to review today.

@glennw 
> I think it's probably fine to have this code in WR. Even though it's probably possible for the user to handle it externally, it seems like we could make it a safer interface by managing this internally. 

It's not obvious to me. The external image stuff is not allocated/owned by WR, so we should not be responsible for freeing it either. The less API surface is the better for me. How do you see this being a part of WR to be potentially safer?

webrender/src/renderer.rs
```diff
@@ -1409,7 +1423,7 @@ impl Renderer {
                                       &projection);
         }
 
-        self.release_external_textures();
+        self.unlock_external_images();
```
kvark: I think you missed the corresponding `lock()` call somewhere

kvark:
Please squash it as well.

JerryShih:
@kvark 

> Right, but this is easily workable. You could free the resources associated with frame N when you get a callback of frame N+1.

The removing condition will become complicated. You should remember the frame id for each external image. Then, you also need to have a flag to check the external image is be called with api.delete_image(). We should not delete an external image that is still exist in resource cache.
Is it reasonable?

> I think you missed the corresponding lock() call somewhere
Oh, I should check all the flows for get(), release(), lock() and unlock().

I will update a squashed version later.

webrender/src/renderer.rs
JerryShih: done

webrender/src/renderer.rs
JerryShih: done

webrender/src/resource_cache.rs
JerryShih: done

webrender/src/renderer.rs
```diff
@@ -1263,7 +1265,7 @@ impl Renderer {
                 let props = &deferred_resolve.image_properties;
                 let external_id = props.external_id
                                        .expect("BUG: Deferred resolves must be external images!");
-                let image = handler.get(external_id);
+                let image = handler.lock(external_id);
```
JerryShih: turn to use lock() here.

webrender/src/renderer.rs
```diff
@@ -740,10 +740,12 @@ impl Renderer {
...
             match msg {
-                ResultMsg::UpdateTextureCache(update_list) => {
-                    self.pending_texture_updates.push(update_list);
-                }
-                ResultMsg::NewFrame(frame, profile_counters) => {
+                ResultMsg::NewFrame(frame, texture_update_list, external_image_update_list, profile_counters) => {
+                    self.pending_texture_updates.push(texture_update_list);
+
+                    // When a new frame is ready, we could start to update all pending external image requests here.
+                    self.release_external_images(external_image_update_list);
```
JerryShih: Only release the external images used in previous frames when a new frame is ready.

JerryShih:
@kvark , the updated patch is ready.

kvark:
@JerryShih 

> The removing condition will become complicated. You should remember the frame id for each external image.

Yes, that's what I [meant](https://github.com/servo/webrender/pull/636#issuecomment-267118267) originally:
> Whatever external resources are involved, the user is able to associate with this frame ID outside of WR being aware.

> Then, you also need to have a flag to check the external image is be called with api.delete_image(). We should not delete an external image that is still exist in resource cache.

I don't see which safety guarantees WR can impose on the external resources anyway, so I don't see a reason to complicate WR with an attempt to share any responsibility. The user should be fully responsible for the lifetime of those, and all information it needs to do so is just that frame ID. So, if the Renderer is done with frame N+1 (or later), the caller can free the resources associated with N. WR's delete_image() for the external resources would just clean up internal structures and not do anything dangerous by itself.

I realize that my suggestion is a little hand-wavy, but I believe we should consider this as a way of minimal API to support the externally owned objects. Looking forward to hear what @glennw thinks about the possible safety extras we can get from a fatter API.

glennw:
@kvark I think you're quite right - this could be handled on the client side with a smaller API.

I don't think having this API necessarily makes anything safer (that was bad wording on my part). But I do think if we can handle these details internally, it makes it easier for other clients to use the API correctly and safely (i.e. having the code to manage this in one place in WR is probably better than Servo, Gecko and other API clients having to correctly implement the semantics above).

Thoughts?

nical:
On the gecko side we're about to make some substantial changes how the different WebRender parts are organized and threads, so some of the assumptions that this PR has been designed against will probably change. While it's good to move forward with this and start iterating, I don't think we need to get all of details right just now. We can come back on how much is handled inside or outside of WebRender as the moving pieces come together.

kvark:
Given what @nical said, It looks like a classical textbook case for a fork, given that you need to play with API/internals and you don't know ~~what you are doing~~ what you'll need at the end, by which time you'd want to merge the polished changes upstream...
However, if @glennw is willing to tolerate the experiments happening here (upstream), and this will make GFX team life easier, then let's merge. I don't want to be stalling the team's progress.

JerryShih:
@kvark @glennw 
I think I could try to handle the external resource as @kvark 's suggestion(maybe just show up a simple pseudo code and api interface). And we can compare them later.

@nical 
Even though we will make render out of ipc-receiver thread, I think it's doesn't affect the design of these external image callbacks. We might just need to handle the thread problem for the TextureSource/TextureHost or compositables in WR's callback. What's your concern for this issue.

glennw:
I'm fine with API experiments happening here in upstream - so I'm happy to either merge this (once rebased and review is finished) or leave it until things settle down more - whatever will make it easiest for you.

JerryShih:
I'm trying to update the gecko's new WR model, so I would like to use these patches first. These patches are rebased.
@glennw @kvark

PR #638, opened 2016-12-14
Indicate macro exists as a bug workaround.

frewsxcv:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/638)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -74,6 +74,10 @@ in int aLayerIndex;
 in int aElementIndex;
 in ivec2 aUserData;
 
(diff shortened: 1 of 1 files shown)


webrender/res/prim_shared.glsl
```diff
@@ -74,6 +74,9 @@ in int aLayerIndex;
 in int aElementIndex;
 in ivec2 aUserData;
 
+// get_fetch_uv is a macro to workaround a macOS Intel driver parsing bug.
```
kvark: I'd add `//TODO: convert back to a function once the driver issues are resolved, if ever`
Also, "workaround" here should be "work around" for a verb is expected.

frewsxcv: Done.

PR #648, opened 2016-12-15
Use z-buffer for rejecting opaque fragments.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/648)

### DIFF
--- a/replay/src/main.rs
+++ b/replay/src/main.rs
@@ -120,7 +120,6 @@ fn main() {
         debug: false,
         enable_subpixel_aa: false,
         clear_framebuffer: true,
-        clear_empty_tiles: false,
         clear_color: ColorF::new(1.0, 1.0, 1.0, 1.0),
     };
 
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -356,7 +356,7 @@ Primitive load_primitive_custom(PrimitiveInstance pi) {
     prim.prim_index = pi.specific_prim_index;
     prim.sub_index = pi.sub_index;
     prim.user_data = pi.user_data;
-    prim.z = pi.z;
+    prim.z = float(pi.z);
 
     return prim;
 }
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -73,6 +73,7 @@ in int aClipTaskIndex;
 in int aLayerIndex;
 in int aElementIndex;
 in ivec2 aUserData;
+in int aZIndex;
 
 // get_fetch_uv is a macro to work around a macOS Intel driver parsing bug.
 // TODO: convert back to a function once the driver issues are resolved, if ever.
@@ -286,6 +287,7 @@ struct PrimitiveInstance {
     int clip_task_index;
     int layer_index;
     int sub_index;
+    int z;
     ivec2 user_data;
 };
 
@@ -299,6 +301,7 @@ PrimitiveInstance fetch_prim_instance() {
     pi.layer_index = aLayerIndex;
     pi.sub_index = aElementIndex;
     pi.user_data = aUserData;
+    pi.z = aZIndex;
 
     return pi;
 }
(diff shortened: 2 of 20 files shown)


glennw:
This still needs a little bit of cleanup - but it implements the functionality and passes all tests. So it should be fairly close to the final version.

webrender/src/renderer.rs
```diff
@@ -123,7 +122,6 @@ const CLIP_FEATURE: &'static str = "CLIP";
 
 enum ShaderKind {
     Primitive,
-    Clear,
```
kvark: sweeet!

webrender/src/renderer.rs
```diff
@@ -990,12 +961,87 @@ impl Renderer {
         self.profile_counters.draw_calls.inc();
     }
 
+    fn submit_batch(&mut self,
```
kvark: 👍

webrender/src/renderer.rs
```diff
@@ -1158,11 +1213,16 @@ impl Renderer {
         self.device.set_blend(false);
         let mut prev_blend_mode = BlendMode::None;
 
-        for batch in &target.alpha_batcher.batches {
-            let transform_kind = batch.key.flags.transform_kind();
-            let needs_clipping = batch.key.flags.needs_clipping();
-            debug_assert!(!needs_clipping || batch.key.blend_mode == BlendMode::Alpha);
+        self.device.enable_depth();
+        self.device.enable_depth_write();
```
kvark: do we set the depth function anywhere?

webrender/res/prim_shared.glsl
```diff
@@ -286,6 +287,7 @@ struct PrimitiveInstance {
     int clip_task_index;
     int layer_index;
     int sub_index;
+    int z;
```
glennw: Yep, I think it's fine to do this as a follow up though.

webrender/src/device.rs
```diff
@@ -1762,15 +1779,38 @@ impl Device {
...
+            gl::clear_depth(depth as f64);
+            clear_bits |= gl::DEPTH_BUFFER_BIT;
+        }
+
+        if clear_bits != 0 {
+            gl::clear(clear_bits);
+        }
+    }
+
+    pub fn enable_depth(&self) {
```
glennw: Yep, perhaps as a follow up we should tidy up the whole device interface?

webrender/res/ps_gradient.vs.glsl
```diff
@@ -39,6 +39,7 @@ void main(void) {
 #ifdef WR_FEATURE_TRANSFORM
     TransformVertexInfo vi = write_transform_vertex(segment_rect,
                                                     prim.local_clip_rect,
+                                                    prim.z,
```
glennw: Some of the shaders generate the local rect separately (as in the case above), which is why I wasn't passing in the prim directly - to avoid bugs where that code accesses the prim.local_rect accidentally. But we could probably tidy up those structure definitions in the shader code.

glennw:
@kvark @pcwalton I think this is ready for review now.

replay/src/main.rs
```diff
@@ -120,7 +120,6 @@ fn main() {
         debug: false,
         enable_subpixel_aa: false,
         clear_framebuffer: true,
-        clear_empty_tiles: false,
```
pcwalton: Why do we not clear empty tiles anymore?

glennw: Since we have to clear the z-buffer now, I figure we might as well just do a clear of the color buffer as well. This also makes it perform much better on mobile / tiled GPUs which rely on a clear due to the way tiling works. We could definitely re-instate this later as an optimization if it makes sense to.

kvark:
AMD GCN architecture, covering probably most of the AMD cards on the market, has meta-data for clears, fragment masks, and depth/stencil bounds - all per internal tile. So clearing on them is dirt cheap, you just pay a little before sampling from those targets for the first time, since the driver needs to fill up the untouched tiles. It basically does what is being removed from WR now, just more efficiently ;)

PR #649, opened 2016-12-15
Run wrench on travis CI with OSMesa

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/649)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
+dist: trusty
 language: rust
+rust: stable
 os:
   - linux
   - osx
@@ -8,8 +10,15 @@ notifications:
   webhooks: http://build.servo.org:54856/travis
 addons:
   apt:
+    sources:
+      - sourceline: 'deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.8 main'
+        keyurl: 'http://apt.llvm.org/llvm-snapshot.gpg.key'
     packages:
       - libgles2-mesa-dev
+      - llvm-3.8-dev
+      - libedit-dev
+before_install:
+  - pip install mako
 script:
   - (cd webrender_traits && cargo test --verbose --features "ipc")
   - (cd webrender_traits && cargo test --verbose)
@@ -17,3 +26,5 @@ script:
   - (cd sample && cargo test --verbose)
   - (cd replay && cargo test --verbose)
   - (cd wrench && cargo test --verbose)
+  - (cd wrench && cargo build --release --verbose --features headless)
+  - (cd wrench && python run_tests.py res/rect.yaml)
--- a/wrench/Cargo.toml
+++ b/wrench/Cargo.toml
@@ -20,6 +20,11 @@ serde = "0.8"
 serde_json = "0.8"
 time = "0.1"
 crossbeam = "0.2"
+osmesa-sys = { version = "0.1.2", optional = true }
+osmesa-src = { git = "https://github.com/servo/osmesa-src", optional = true }
+
+[features]
+headless = [ "osmesa-sys", "osmesa-src" ]
 
 [dependencies.webrender]
 path = "../webrender"
--- a/wrench/run_tests.py
+++ b/wrench/run_tests.py
@@ -1,0 +1,65 @@
+#!/usr/bin/python
+
+import contextlib
+import os
+import subprocess
+import sys
+import hashlib
+from os import path
+from glob import glob
+
+
+@contextlib.contextmanager
+def cd(new_path):
+    """Context manager for changing the current working directory"""
+    previous_path = os.getcwd()
+    try:
+        os.chdir(new_path)
+        yield
+    finally:
+        os.chdir(previous_path)
+
+
+def find_dep_path_newest(package, bin_path):
+    deps_path = path.join(path.split(bin_path)[0], "build")
+    with cd(deps_path):
+        candidates = glob(package + '-*')
+    candidates = (path.join(deps_path, c) for c in candidates)
+    candidate_times = sorted(((path.getmtime(c), c) for c in candidates), reverse=True)
+    if len(candidate_times) > 0:
+        return candidate_times[0][1]
+    return None
+
+
+def is_windows():
+    """ Detect windows, mingw, cygwin """
+    return sys.platform == 'win32' or sys.platform == 'msys' or sys.platform == 'cygwin'
+
+
+def is_macosx():
+    return sys.platform == 'darwin'
+
+
+def is_linux():
+    return sys.platform.startswith('linux')
+
+
+def set_osmesa_env(bin_path):
+    """Set proper LD_LIBRARY_PATH and DRIVE for software rendering on Linux and OSX"""
+    if is_linux():
+        osmesa_path = path.join(find_dep_path_newest('osmesa-src', bin_path), "out", "lib", "gallium")
+        print(osmesa_path)
+        os.environ["LD_LIBRARY_PATH"] = osmesa_path
+        os.environ["GALLIUM_DRIVER"] = "softpipe"
+    elif is_macosx():
+        osmesa_path = path.join(find_dep_path_newest('osmesa-src', bin_path),
+                                "out", "src", "gallium", "targets", "osmesa", ".libs")
+        glapi_path = path.join(find_dep_path_newest('osmesa-src', bin_path),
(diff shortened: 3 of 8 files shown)


wrench/Cargo.toml
```diff
@@ -33,3 +33,7 @@ dwrote = "0.1.2"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 font-loader = "0.1.0"
+
+[target.'cfg(any(target_os = "linux", target_os = "macos"))'.dependencies]
```
kvark: given that we use it for automated tests only, can we hide this dependency behind a feature gate?

wrench/run_tests.py
```diff
@@ -0,0 +1,63 @@
...
+    elif is_macosx():
+        osmesa_path = path.join(find_dep_path_newest('osmesa-src', bin_path),
+                                "out", "src", "gallium", "targets", "osmesa", ".libs")
+        glapi_path = path.join(find_dep_path_newest('osmesa-src', bin_path),
+                               "out", "src", "mapi", "shared-glapi", ".libs")
+        os.environ["DYLD_LIBRARY_PATH"] = osmesa_path + ":" + glapi_path
+        os.environ["GALLIUM_DRIVER"] = "softpipe"
+
+
+set_osmesa_env('../target/release/')
```
kvark: is our executable built by `cargo test` guaranteed to be in `release` folder?

wrench/src/main.rs
```diff
@@ -75,31 +86,161 @@ impl ThingKind {
...
+            width: width,
+            height: height,
+        }
+    }
+
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    fn get_proc_address(s: &str) -> *const c_void {
+        let c_str = CString::new(s).expect("Unable to create CString");
+        unsafe {
+            mem::transmute(osmesa_sys::OSMesaGetProcAddress(c_str.as_ptr()))
```
kvark: can we use a simple `X as *const _` here instead of `transmute`?

wrench/src/main.rs
```diff
@@ -175,97 +320,138 @@ fn main() {
...
+            if let Some(limit) = limit_seconds {
+                if (time::SteadyTime::now() - time_start) >= limit {
+                    break 'outer;
                 }
+            }
 
-                if show_help {
-                    wrench.show_onscreen_help();
-                }
+            let thing = thing.thing();
```
kvark: thing

wrench/src/main.rs
```diff
@@ -175,97 +320,138 @@ fn main() {
...
-                            VirtualKeyCode::Right => {
-                                thing.next_frame();
-                            },
-                            VirtualKeyCode::H => {
-                                show_help = !show_help;
-                            },
-                            _ => ()
+                glutin::Event::KeyboardInput(kstate, _scan_code, maybe_vk) => {
+                    if kstate == ElementState::Pressed {
+                        if let Some(vk) = maybe_vk {
```
kvark: nit: we could combine both of these `if`s

wrench/src/main.rs
```diff
@@ -175,97 +320,138 @@ fn main() {
...
             }
-            _ => ()
         }
     }
+
+    if is_headless {
+        let mut pixels = gl::read_pixels(0, 0,
+                                         size.width as gl::GLsizei,
+                                         size.height as gl::GLsizei,
+                                         gl::RGB, gl::UNSIGNED_BYTE);
```
kvark: would that incur a software format conversion from RGBA to RGB?

glennw:
This is a bit untidy, but it now runs wrench in headless / osmesa mode on travis (mac + linux), writes a screenshot, and the md5sum of that png matches on both CI machines, and also on my local machine when I check the image output visually.

@jrmuizel Hopefully this is enough for you to continue your work on adding basic test support. Let me know if anything isn't working or needs some additional functionality.

@kvark I've addressed most of the comments you left, but if you have time to take another look over it and see what else needs cleaning up, we could perhaps get this part merged soon-ish.

wrench/Cargo.toml
glennw: Moved this to the 'headless' feature.

wrench/run_tests.py
glennw: I explicitly build the release version in the travis script - perhaps there's a better way to do this though.

.travis.yml
```diff
@@ -8,8 +8,15 @@ notifications:
   webhooks: http://build.servo.org:54856/travis
 addons:
   apt:
+    sources:
+      - sourceline: 'deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.8 main'
+        keyurl: 'http://apt.llvm.org/llvm-snapshot.gpg.key'
     packages:
       - libgles2-mesa-dev
+      - llvm-3.8
+      - llvm-3.8-dev
```
glennw: Fixed

wrench/src/main.rs
```diff
@@ -21,18 +21,29 @@ extern crate serde;
 extern crate serde_json;
 extern crate crossbeam;
 
+#[cfg(any(target_os = "linux", target_os = "macos"))]
```
glennw: Added the headless feature.

wrench/src/main.rs
glennw: The OSMesa bindings return an Option<fn> here so I think a transmute is required here due to the interface mismatch. We could perhaps make these match in the future and remove the transmute?

wrench/src/main.rs
glennw: Unrelated to this PR, so leaving for now.

wrench/src/main.rs
glennw: Perhaps - I'm not entirely sure. But it's only done once when writing the screenshot in test mode, so probably not a big deal.

nox:
@glennw Seems that Travis is happy now.

wrench/src/main.rs
kvark: Sounds good! No need to have an extra `unsafe` where it's not required

wrench/src/main.rs
```diff
@@ -21,18 +21,29 @@ extern crate serde;
...
 use std::path::PathBuf;
 use std::cmp::{min, max};
-use std::ffi::CStr;
+use std::fs::File;
+use std::mem;
+use std::os::raw::c_void;
+use std::ptr;
 use webrender_traits::*;
 
+#[cfg(any(target_os = "linux", target_os = "macos"))]
```
kvark: should use `headless` feature

wrench/src/main.rs
```diff
@@ -75,31 +86,161 @@ impl ThingKind {
...
+}
+
+pub struct HeadlessEventIterater;
+
+impl WindowWrapper {
+    fn init(&self) {
+        match *self {
+            WindowWrapper::Window(ref window) => {
+                unsafe {
+                    window.make_current().ok();
```
kvark: I don't think it makes sense to ignore the returned `Err` here, we better just do `unwrap` or `expect` instead.

wrench/src/main.rs
```diff
@@ -175,97 +320,138 @@ fn main() {
...
+        let mut events = Vec::new();
+
+        match window {
+            WindowWrapper::Headless(..) => {
+                // Just keep rendering!
+                // TODO: Add an exit condition for headless mode!
+                events.push(glutin::Event::Awakened);
+            }
+            WindowWrapper::Window(ref window) => {
+                for event in window.poll_events() {
```
kvark: nit: could do shorter:
```rust
let events = match window {
  WindowWrapper::Headless(..) => {
    vec![glutin::Event::Awakened]
  }
  WindowWrapper::Window(ref window) => {
    window.poll_events().collect()
  }
};
```

glennw:
@kvark Thanks for the review, addressed those comments and squashed.

kvark:
Thanks!
I suppose this error at the end is expected?
> thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: RecvError', /Users/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-mac/build/src/libcore/result.rs:837
note: Run with `RUST_BACKTRACE=1` for a backtrace.
md5 = 31ca82c0aadec2c0ebfd0e3f95510c9c
/Users/travis/build.sh: line 57: shell_session_update: command not found

Either way, we can follow up with more work, at least Travis is green.
@bors-servo r+

PR #659, opened 2016-12-22
Fix DirectWrite text rendering

vvuk:
This uses the proper dwrite methods to pull out glyph images.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/659)


webrender/Cargo.toml
```diff
@@ -29,12 +29,13 @@ time = "0.1"
 threadpool = "1.3.2"
 webrender_traits = {path = "../webrender_traits", default-features = false}
 bitflags = "0.7"
+image = "0.10"
```
glennw: This doesn't appear to be used.

vvuk: quite, fixed.

PR #672, opened 2017-01-02
Add debug mode to visualize intermediate render targets.

glennw:
Optionally blit the contents of any intermediate render targets to the bottom of the frame buffer, to allow visual debugging and inspection of render target contents and usage.

rt-debug

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/672)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -18,7 +18,7 @@ use std::mem;
 use std::path::PathBuf;
 //use std::sync::mpsc::{channel, Sender};
 //use std::thread;
-use webrender_traits::{ColorF, ImageFormat};
+use webrender_traits::{ColorF, ImageFormat, DeviceIntRect};
 
 #[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
 const GL_FORMAT_A: gl::GLuint = gl::RED;
@@ -74,6 +74,11 @@ pub enum VertexFormat {
     Clear,
     Blur,
     Clip,
+}
+
+enum FBOTarget {
+    Read,
+    Draw,
 }
 
 fn get_optional_shader_source(shader_name: &str, base_path: &Option<PathBuf>) -> Option<String> {
@@ -325,8 +330,12 @@ impl UBOId {
 }
 
 impl FBOId {
-    fn bind(&self) {
-        gl::bind_framebuffer(gl::FRAMEBUFFER, self.0);
+    fn bind(&self, target: FBOTarget) {
+        let target = match target {
+            FBOTarget::Read => gl::READ_FRAMEBUFFER,
+            FBOTarget::Draw => gl::DRAW_FRAMEBUFFER,
+        };
+        gl::bind_framebuffer(target, self.0);
     }
 }
 
@@ -777,8 +786,10 @@ pub struct Device {
     bound_textures: [TextureId; 16],
     bound_program: ProgramId,
     bound_vao: VAOId,
-    bound_fbo: FBOId,
-    default_fbo: gl::GLuint,
+    bound_read_fbo: FBOId,
+    bound_draw_fbo: FBOId,
+    default_read_fbo: gl::GLuint,
+    default_draw_fbo: gl::GLuint,
     device_pixel_ratio: f32,
 
     // HW or API capabilties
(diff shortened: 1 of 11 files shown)


webrender/src/device.rs
```diff
@@ -1169,7 +1202,34 @@ impl Device {
             }
         }
 
-        gl::bind_framebuffer(gl::FRAMEBUFFER, self.default_fbo);
+        // TODO(gw): Hack! Modify the code above to use the normal binding interfaces the device exposes.
+        gl::bind_framebuffer(gl::READ_FRAMEBUFFER, self.bound_read_fbo.0);
```
kvark: While I do see this code being correct, I can't shake off the feeling that something is wrong, given that the old code was binding a different FBO. So the new code will behave differently. If it still works, perhaps the current FBO gets overwritten by something later?

webrender/src/device.rs
```diff
@@ -927,18 +943,31 @@ impl Device {
...
+            self.textures.get(&texture_id.0).unwrap().fbo_ids[texture_id.1 as usize]
+        });
+
+        if self.bound_read_fbo != fbo_id {
+            self.bound_read_fbo = fbo_id;
+            fbo_id.bind(FBOTarget::Read);
+        }
+    }
+
     pub fn bind_render_target(&mut self,
```
kvark: nit: we'd want `bind_draw_target` for consistency

webrender/src/device.rs
```diff
@@ -1169,7 +1202,34 @@ impl Device {
...
 
-        gl::bind_framebuffer(gl::FRAMEBUFFER, self.default_fbo);
+        // TODO(gw): Hack! Modify the code above to use the normal binding interfaces the device exposes.
+        gl::bind_framebuffer(gl::READ_FRAMEBUFFER, self.bound_read_fbo.0);
+        gl::bind_framebuffer(gl::DRAW_FRAMEBUFFER, self.bound_draw_fbo.0);
+    }
+
+    pub fn blit_render_target(&mut self,
+                              src_texture_id: TextureId,
+                              layer: i32,
```
kvark: the parameters here need more semantics in their name. While `layer` refers to the source texture layer, the `x, y, width, and height` are for the destination region. Perhaps, we should use something like `dst_rect: Rect<f32>`, where `Rect` is imported from `euclid`?

webrender/src/device.rs
```diff
@@ -927,18 +943,31 @@ impl Device {
         }
     }
 
+    pub fn bind_read_target(&mut self, texture_id: Option<(TextureId, i32)>) {
+        debug_assert!(self.inside_frame);
```
kvark: don't we also need to call `glReadBuffer` somewhere?

webrender/src/renderer.rs
```diff
@@ -1416,6 +1421,41 @@ impl Renderer {
...
+                                framebuffer_size: &DeviceUintSize) {
+        if self.render_target_debug {
+            // TODO(gw): Make the layout of the render targets a bit more sophisticated.
+            // Right now, it just draws them in one row at the bottom of the screen,
+            // with a fixed size.
+            let rt_debug_x0 = 16;
+            let rt_debug_y0 = 16;
+            let rt_debug_spacing = 16;
+            let rt_debug_size = 512;
+            let mut current_fbo = 0;
```
kvark: name is incorrect. Perhaps, `current_target`?

webrender/src/renderer.rs
```diff
@@ -1416,6 +1421,41 @@ impl Renderer {
...
+
+                    self.device.blit_render_target(*texture_id,
+                                                   layer_index as i32,
+                                                   x0,
+                                                   y0,
+                                                   rt_debug_size,
+                                                   rt_debug_size);
+
+                    // If we have more targets than fit on one row in screen, just early exit.
+                    if x0 > framebuffer_size.width as i32 {
```
kvark: shouldn't we do that before rendering outside of `x0`?

webrender/src/device.rs
glennw: I don't think it's necessary, although perhaps we could call it at the start of the frame once to be safe.

webrender/src/device.rs
glennw: I think the old code was just *wrong* but happened to not show up any bugs thus far. All the tests pass with this change, so it seems correct to me to make this behave the way I'd expect the semantics to work :)

glennw:
@kvark Thanks for the review - answered questions above and pushed a new commit to address all the other comments.

glennw:
@kvark Thanks, squashed.

PR #673, opened 2017-01-03
Remove clip_clear and rectangle_noblend from the clip batcher.

glennw:
Now that clip masks are stored in the red channel, we can just
clear the intermediate targets to (1,0,0,0). This satisfies the
requirement of clip masks being initialized to 1, while also
allowing the alpha component to be initialized to 0 for visual
render tasks that rely on it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/673)


kvark:
Makes sense, thanks!
@bors-servo r+

PR #678, opened 2017-01-04
Add RendererOptionsBuilder

vvuk:
Having to fix up dependent code whenever unrelated options are added/removed is annoying; so I made a builder.  Not required to use it so as to not break existing code (and I only updated the sample), but we can consider doing so in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/678)


kvark:
The change looks good, but I don't think `RenderOptions` really needs a builder. If a builder does anything non-trivial, then it makes sense to hide it. For simple assignments, it seems more straightforward/idiomatic to just use the `Default`:
```rust
impl Default for RenderOptions {...}
```
Then the sample could do:
```rust
let opts = webrender::RendererOptions {
  device_pixel_ratio: 1.0,
  resource_override_path: res_path,
  debug: true,
  clear_framebuffer: true,
  .. Default::default()  // sets the rest of the fields
};
```
The benefit is - less code, less API complexity.

Issue #654, opened 2016-12-16
Shader compile errors in `cs_clip_clear`


mmatyas:
I was trying to open http://en.wikipedia.org/wiki/Rust on a [OnePlus 2](http://www.gsmarena.com/oneplus_2-6902.php), but got the following shader compilation errors:

```
12-16 13:02:14.192 31641 31657 I Adreno  : ERROR: 0:18: '=' :  cannot convert from '2-component vector of int' to '2-component vector of float'
12-16 13:02:14.192 31641 31657 I Adreno  : ERROR: 0:6: 'CacheClipInstance' : undeclared identifier 
12-16 13:02:14.192 31641 31657 I Adreno  : ERROR: 0:6: 'cci' : Syntax error:  syntax error
12-16 13:02:14.192 31641 31657 I Adreno  : ERROR: 3 compilation errors.  No code generated.
12-16 13:02:14.192 31641 31657 I Adreno  : 
12-16 13:02:14.193 31641 31661 D RustAndroidGlueStdouterr: Failed to compile shader: "cs_clip_clear"
12-16 13:02:14.193 31641 31660 D RustAndroidGlueStdouterr: ERROR:servo: -- Shader compile failed - exiting --
12-16 13:02:14.194 31641 31661 D RustAndroidGlueStdouterr: ERROR: 0:18: '=' :  cannot convert from '2-component vector of int' to '2-component vector of float'
12-16 13:02:14.194 31641 31661 D RustAndroidGlueStdouterr: ERROR: 0:6: 'CacheClipInstance' : undeclared identifier 
12-16 13:02:14.194 31641 31661 D RustAndroidGlueStdouterr: ERROR: 0:6: 'cci' : Syntax error:  syntax error
12-16 13:02:14.194 31641 31661 D RustAndroidGlueStdouterr: ERROR: 3 compilation errors.  No code generated.
12-16 13:02:14.194 31641 31661 D RustAndroidGlueStdouterr: 
12-16 13:02:14.194 31641 31661 D RustAndroidGlueStdouterr: 
12-16 13:02:14.194 31641 31661 D RustAndroidGlueStdouterr: -- Shader compile failed - exiting -- (thread <unnamed>, at /home/mmatyas/Servo/servo-android/servo/.cargo/git/checkouts/webrender-c3596abe1cf4f320/c4a0c01/webrender/src/device.rs:834)
[...]

PR #681, opened 2017-01-04
Fixed shader compile error in cs_clip_image VS

kvark:
That's something I found while looking for the cause of #654
Doesn't look like it would fix it, but it would be great to have another try on Android after this PR is merged (cc @mmatyas ).

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/681)


mmatyas:
It works fine on Odroid, I'll test it on other devices as soon as I can.

Issue #676, opened 2017-01-04
Standalone texture allocation id's are freed in render_backend but not in renderer


rlhunt:
The CacheTextureId for standalone texture allocations is returned to the free list in the render_backend when it is freed [1], but no message is sent to sync this change to renderer. When that id is reused. this assert [2] is triggered.

I was able to trigger this by using ImageRendering::Pixelated which causes TextureFilter::Point which causes standalone texture allocations.

Either not returning the id to the free list or adding a new TextureUpdateOp to free from the cache_texture_id_map worked for me. I'm not sure if there are any side effects we'd need to be concerned about with freeing from cache_texture_id_map, but I think that'd be the way to go.

[1] https://github.com/servo/webrender/blob/master/webrender/src/texture_cache.rs#L880
[2] https://github.com/servo/webrender/blob/master/webrender/src/renderer.rs#L867

glennw:
@rlhunt This is fixed now, right?

kvark:
Yeah, closing.

PR #682, opened 2017-01-04
Free CacheTextureId's in Renderer but persist their TextureId.

rlhunt:
Currently the TextureCache will free CacheTextureId's for standalone texture allocations. This isn't synced to Renderer::cache_texture_id_map, which will then hit an assert when the CacheTextureId is reused. (#676)

When we get a TextureUpdateOp::Create and detect that they're reusing a CacheTextureId, we should just reuse the existing TextureId. The cache_texture_id_map can then drop Option and be Vec<TextureId>.

For the TextureId's that haven't been reused yet, we should send a TextureUpdateOp::Free that notifies Renderer to free the texture memory of the TextureId. This wouldn't gl::delete_textures the TextureId, but just deinit the TextureId until it is reused.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/682)


glennw:
Thanks!

PR #685, opened 2017-01-05
Removes the tiling and render task stack from clipping, draws clips in regions of interest.

glennw:
This also ensures that primitive clip masks never render
any larger than the primitive bounding rect, which improves
the timings on GitHub a lot.

This does cause a slight performance regression on some
sites (GitHub in particular) but is a first step to the
planned clipping and tiling improvements coming. After those
land, the performance will be better than originally.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/685)


glennw:
@kvark This decouples the clip mask generation from the tiling system, as we discussed.

It also removes the (currently unused) chained render task support. I removed it to simplify future refactoring, since it's always available in git history if we need it again, but I can restore it if you'd prefer.

I marked this as [WIP] since it doesn't have any of the improved clip mask ideas we talked about at the work week yet - but it does pass all tests and only cause a small (temporary) performance regression on some sites. So we could perhaps merge this as a standalone PR, to keep the patches for this work a bit more manageable and easy to review.

I'm happy to get this merged after review, or just leave it as a PR and continue the refactoring work on top of this branch. What do you think?

glennw:
The 2nd commit makes the clip shader only run in the regions of interest for a clip mask. This makes clip mask generation time negligible on any sites I've tried (e.g. GH.com drops from 9ms to ~3ms on my test setup).

We still have the memory wastage for very large clip rects, but we have a plan to solve that - so I think that can be done as a follow up, since this patch is already large enough.

The clip performance is now better on each site I tested than it was previously.

r? @kvark

webrender/src/mask_cache.rs
```diff
@@ -59,38 +71,41 @@ impl MaskCacheInfo {
...
+                    mask.image_components.push(ImageMaskComponent {
+                        gpu_address: clip_store.alloc(1),
+                        image_mask: image_mask,
+                    });
                 }
-            ),
+
+                for _ in 0..region.complex.length * CORNERS_PER_CLIP_REGION {
+                    mask.corner_components.push(CornerMaskComponent {
+                        gpu_address: clip_store.alloc(1),
```
kvark: dunno if it matters, but we could bulk-allocate here

webrender/src/mask_cache.rs
```diff
@@ -99,59 +114,38 @@ impl MaskCacheInfo {
...
             let mut local_rect;
-            let mut local_inner: Option<LayerRect>;
             match source {
                 &ClipSource::NoClip => unreachable!(),
                 &ClipSource::Complex(rect, radius) => {
-                    let slice = clip_store.get_slice_mut(self.clip_range.start, CLIP_DATA_GPU_SIZE);
                     let data = ClipData::uniform(rect, radius);
-                    PrimitiveStore::populate_clip_data(slice, data);
-                    debug_assert_eq!(self.clip_range.item_count, 1);
+                    for (corner, component) in data.corners.iter().zip(self.corner_components.iter()) {
```
kvark: should we require the length of these vectors to match?

webrender/src/mask_cache.rs
```diff
@@ -99,59 +114,38 @@ impl MaskCacheInfo {
...
+                    debug_assert_eq!(self.corner_components.len(), clips.len() * CORNERS_PER_CLIP_REGION);
+                    for (clip, chunk) in clips.iter().zip(self.corner_components.chunks_mut(CORNERS_PER_CLIP_REGION)) {
                         let data = ClipData::from_clip_region(clip);
-                        PrimitiveStore::populate_clip_data(chunk, data);
+                        for (corner, component) in data.corners.iter().zip(chunk) {
+                            let gpu_block = clip_store.get_mut(component.gpu_address);
+                            *gpu_block = GpuBlock32::from(*corner);
+                        }
                         local_rect = local_rect.and_then(|r| r.intersection(&clip.rect));
-                        local_inner = local_inner.and_then(|r| clip.get_inner_rect()
```
kvark: why are we removing the `local_inner`?

webrender/res/clip_shared.glsl
```diff
@@ -23,30 +21,65 @@ CacheClipInstance fetch_clip_item(int index) {
...
-TransformVertexInfo write_clip_tile_vertex(vec4 local_clip_rect,
+TransformVertexInfo write_clip_tile_vertex(vec4 local_rect,
                                            Layer layer,
                                            ClipArea area) {
-    vec2 lp0_base = local_clip_rect.xy;
-    vec2 lp1_base = local_clip_rect.xy + local_clip_rect.zw;
+    vec2 lp0_base = local_rect.xy;
+    vec2 lp1_base = local_rect.xy + local_rect.zw;
+
+    vec2 lp0 = clamp_rect(clamp_rect(lp0_base, local_rect),
```
glennw: Fixed

webrender/res/cs_clip_rectangle.fs.glsl
```diff
@@ -3,37 +3,15 @@
...
-
-    bvec4 is_out = bvec4(pos.x < ref_tl.x && pos.y < ref_tl.y,
-                         pos.x > ref_tr.x && pos.y < ref_tr.y,
-                         pos.x > ref_br.x && pos.y > ref_br.y,
-                         pos.x < ref_bl.x && pos.y > ref_bl.y);
 
-    float distance_from_border = dot(vec4(is_out),
-                                     max(vec4(0.0, 0.0, 0.0, 0.0), distances));
+    // TODO(gw): Support ellipse clip!
+    float d = (distance(pos, vClipRef) - vClipRadius.x + nudge) / pixels_per_fragment;
```
glennw: Fixed

webrender/res/cs_clip_rectangle.glsl
```diff
@@ -7,4 +7,5 @@
 varying vec3 vPos;
 flat varying vec4 vLocalRect;
 flat varying vec4 vClipRect;
-flat varying vec4 vClipRadius;
+flat varying vec2 vClipRadius;
+flat varying vec2 vClipRef;
```
glennw: Fixed

webrender/src/mask_cache.rs
glennw: Probably a good idea, I don't think it matters too much for now though.

webrender/src/mask_cache.rs
glennw: I think that's enforced by the debug assert above?

webrender/src/mask_cache.rs
glennw: I figured it wasn't necessary now that the tiles are decoupled from the clip mask - but perhaps there is still a good reason to have it?

glennw:
@kvark I had an idea in mind to handle the outside clip case that we were discussing last week.

But then I went to create a test case for that problem, and was unable to create a demonstration of the issue - I think I've convinced myself it's not possible with the way the current do_clip() works, but I'm probably just missing something.

Let's discuss the exact problem again this week and come up with an example that demonstrates the issue, and then I'll add the solution for it to this PR.

webrender/src/mask_cache.rs
kvark: which one?

webrender/src/mask_cache.rs
kvark: AFAIK, our earlier discussion was about having 4 clip rectangles generated instead of one so that the inner area gets excluded from the mask computation. Then we figured that these 4 clips can be just the corners of a rectangle *if* the rounded rectangle is all we got. So for that model, we'd still need the inner area calculated. However, I'm not sure you need it for your implementation. I'll do another review pass with that in mind.

kvark:
@glennw If you are talking about the test you added to the sample, then I have an explanation. Since the clip logic works on the AABB of intersection of the clips, and your 2 clip rectangles are axis aligned, their intersection AABB is completely within each clip, so the change of VS logic that I consider incorrect did not make any difference for that specific case.

glennw:
Closing in favour of https://github.com/servo/webrender/pull/696

PR #686, opened 2017-01-05
Handle images where width != stride

jrmuizel:
This 'unstrides' them into a temporary buffer

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/686)

### DIFF
--- a/wrench/src/yaml_frame_writer.rs
+++ b/wrench/src/yaml_frame_writer.rs
@@ -324,39 +324,49 @@ impl YamlFrameWriter {
         let bytes = data.bytes.take().unwrap();
         let (path_file, path) = Self::next_rsrc_paths(&self.rsrc_prefix, &mut self.next_rsrc_num, &self.rsrc_base, "img", "png");
 
-        let ok = match data.format {
+        // takes a buffer with a stride and copies it into a new buffer that has stride == width
+        fn unstride<T: Clone>(mut slice: &[T], width: usize, stride: usize) -> Vec<T> {
+            let mut result = Vec::new();
+            while slice.len() > width {
+                result.extend_from_slice(&slice[..width]);
+                slice = &slice[stride..];
+            }
+            result.extend_from_slice(slice);
+            result
+        }
+
+        assert!(data.stride > 0);
+        match data.format {
             ImageFormat::RGB8 => {
                 if data.stride == data.width * 3 {
                     save_buffer(&path_file, &bytes, data.width, data.height, ColorType::RGB(8)).unwrap();
-                    true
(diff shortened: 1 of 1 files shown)


wrench/src/yaml_frame_writer.rs
```diff
@@ -324,40 +324,50 @@ impl YamlFrameWriter {
         let bytes = data.bytes.take().unwrap();
         let (path_file, path) = Self::next_rsrc_paths(&self.rsrc_prefix, &mut self.next_rsrc_num, &self.rsrc_base, "img", "png");
 
-        let ok = match data.format {
+        // takes a buffer with a stride and copies it into a new buffer that has stride == width
+        fn unstride<T: Clone>(mut slice: &[T], width: usize, stride: usize) -> Vec<T> {
+            let mut result = Vec::new();
```
emilio: nit: You can probably use `Vec::with_capacity` here to avoid extra allocations.

emilio: Oh, thought this was on WR itself, if it's in the recording code probably doesn't matter that much.

wrench/src/yaml_frame_writer.rs
```diff
@@ -324,40 +324,50 @@ impl YamlFrameWriter {
         let bytes = data.bytes.take().unwrap();
         let (path_file, path) = Self::next_rsrc_paths(&self.rsrc_prefix, &mut self.next_rsrc_num, &self.rsrc_base, "img", "png");
 
-        let ok = match data.format {
+        // takes a buffer with a stride and copies it into a new buffer that has stride == width
+        fn unstride<T: Clone>(mut slice: &[T], width: usize, stride: usize) -> Vec<T> {
```
kvark: Since I like converting everything to iterators, as long as it makes sense, and this function does make some, I got this:
```rust
    slice.chunks(stride)
         .flat_map(|chunk| chunk[..width].iter().cloned())
         .collect()
```

wrench/src/yaml_frame_writer.rs
```diff
@@ -324,40 +324,50 @@ impl YamlFrameWriter {
...
+            while slice.len() > width {
+                result.extend_from_slice(&slice[..width]);
+                slice = &slice[stride..];
+            }
+            result.extend_from_slice(slice);
+            result
+        }
+
+        assert!(data.stride > 0);
+        match data.format {
```
kvark: the body of the arms can be merged outside of the `match`, all it needs to return is `(ColorType, BytesPerTexel)` for the subsequent saving/unstriding

PR #687, opened 2017-01-06
Support non-square texture page sizes.

glennw:
This will be taken advantage of in a follow up PR
that adjusts the intermediate render targets (which
use the texture page allocator) to be the same size
as the framebuffer.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/687)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -82,14 +82,14 @@ fn copy_pixels(src: &[u8],
 /// dynamic texture deallocation.
 pub struct TexturePage {
     texture_id: CacheTextureId,
-    texture_size: u32,
+    texture_size: DeviceUintSize,
     free_list: FreeRectList,
     allocations: u32,
     dirty: bool,
 }
 
 impl TexturePage {
-    pub fn new(texture_id: CacheTextureId, texture_size: u32) -> TexturePage {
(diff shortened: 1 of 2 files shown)


webrender/src/texture_cache.rs
```diff
@@ -327,18 +327,29 @@ impl TexturePage {
...
 
-    fn grow(&mut self, new_texture_size: u32) {
-        self.free_list.push(&DeviceUintRect::new(
-            DeviceUintPoint::new(self.texture_size, 0),
-            DeviceUintSize::new(new_texture_size - self.texture_size, new_texture_size)));
-        self.free_list.push(&DeviceUintRect::new(
-            DeviceUintPoint::new(0, self.texture_size),
-            DeviceUintSize::new(self.texture_size, new_texture_size - self.texture_size)));
+    fn grow(&mut self, new_texture_size: DeviceUintSize) {
+        let new_rects = [
```
kvark: we should probably assert on `new_texture_size >= texture_size` here, since the code expects it to be true and would do strange things in release (while failing on arithmetic overflow in debug)

glennw:
@kvark Thanks for the review - added assert to catch that case.

kvark:
Thanks @glennw !
@bors-servo r+

PR #688, opened 2017-01-06
Add an opaque external event message (for use in Gecko).

nical:
Gecko's WebRender integration needs to send a few gecko-specific messages to the render thread. These messages have to preserve their ordering with respect to the other WebRender messages such as set_root_display_list, so they have to go through the render backend thread along with the rest.

The simplest way I could find was to add a pointer-sized opaque handle that is to be passed to the RenderNotifier. Since gecko implements its own RenderNotifier we can let the meaning of the opaque handle be completely up to gecko and only have the render backend pass it along in order.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/688)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -349,8 +349,7 @@ impl RenderBackend {
                         }
                         ApiMsg::ExternalEvent(evt) => {
                             let mut notifier = self.notifier.lock();
-                            notifier.as_mut()
-                                    .unwrap()
+                            notifier.unwrap()
                                     .as_mut()
                                     .unwrap()
                                     .external_event(evt);
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -347,6 +347,14 @@ impl RenderBackend {
                                 frame_counter += 1;
                             }
                         }
+                        ApiMsg::ExternalEvent(evt) => {
+                            let mut notifier = self.notifier.lock();
+                            notifier.as_mut()
+                                    .unwrap()
+                                    .as_mut()
+                                    .unwrap()
+                                    .external_event(evt);
+                        }
                     }
                 }
                 Err(..) => {
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -56,23 +56,23 @@ pub enum ApiMsg {
     GenerateFrame,
     // WebVR commands that must be called in the WebGL render thread.
     VRCompositorCommand(WebGLContextId, VRCompositorCommand),
-    // An opaque handle that must be passed to the render notifier. It is used by Gecko
-    // to forward gecko-specific messages to the render thread preserving the ordering
-    // within the other messages.
+    /// An opaque handle that must be passed to the render notifier. It is used by Gecko
+    /// to forward gecko-specific messages to the render thread preserving the ordering
+    /// within the other messages.
     ExternalEvent(ExternalEvent),
 }
 
+/// An opaque pointer-sized value.
 #[derive(Clone, Deserialize, Serialize)]
 pub struct ExternalEvent {
-    // An opaque pointer-sized value.
     raw: usize,
 }
 
 unsafe impl Send for ExternalEvent {}
 
 impl ExternalEvent {
-    pub fn new(raw: usize) -> Self { ExternalEvent { raw: raw } }
-    // Consume self to make it obvious that the callback should be forwarded only once.
+    pub fn from_raw(raw: usize) -> Self { ExternalEvent { raw: raw } }
+    /// Consumes self to make it obvious that the event should be forwarded only once.
     pub fn unwrap(self) -> usize { self.raw }
 }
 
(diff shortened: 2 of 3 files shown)


webrender_traits/src/types.rs
```diff
@@ -55,7 +55,25 @@ pub enum ApiMsg {
...
+pub struct ExternalEvent {
+    // An opaque pointer-sized value.
+    raw: usize,
+}
+
+unsafe impl Send for ExternalEvent {}
+
+impl ExternalEvent {
+    pub fn new(raw: usize) -> Self { ExternalEvent { raw: raw } }
+    // Consume self to make it obvious that the callback should be forwarded only once.
```
emilio: In that case, maybe `Clone` shouldn't be derived?

nical: Clone is derived because ApiMsg derives Clone which apparently needs to implement it for RenderApiSender to implement Clone as well. I think that it boils down to MsgSender<T> using derive(Clone) instead of implementing it manually, forcing T to implement it as well even though cloning the sender does not clone any message. I guess I should entangle that.

nical: Hm, making ApiMsg not implement Clone breaks wrench in a not-so-trivial way. It's probably not too hard to fix but I don't think it's worth the effort at this point so I backed out the last part.

emilio: Ok, sounds good, I guess you could also implement a panicking `Clone` in `ExternalEvent`, and make sure `wrench` doesn't clone it, could that

webrender/src/render_backend.rs
```diff
@@ -347,6 +347,14 @@ impl RenderBackend {
                                 frame_counter += 1;
                             }
                         }
+                        ApiMsg::ExternalEvent(evt) => {
+                            let mut notifier = self.notifier.lock();
+                            notifier.as_mut()
```
kvark: this first `as_mut` is not needed either here or in any other mentions of `notifier`

webrender_traits/src/types.rs
```diff
@@ -55,7 +55,25 @@ pub enum ApiMsg {
...
+#[derive(Clone, Deserialize, Serialize)]
+pub struct ExternalEvent {
+    // An opaque pointer-sized value.
+    raw: usize,
+}
+
+unsafe impl Send for ExternalEvent {}
+
+impl ExternalEvent {
+    pub fn new(raw: usize) -> Self { ExternalEvent { raw: raw } }
```
kvark: nit: should idiomatically be called `from` or `from_raw`

webrender_traits/src/types.rs
```diff
@@ -55,7 +55,25 @@ pub enum ApiMsg {
     WebGLCommand(WebGLContextId, WebGLCommand),
     GenerateFrame,
     // WebVR commands that must be called in the WebGL render thread.
-    VRCompositorCommand(WebGLContextId, VRCompositorCommand)
+    VRCompositorCommand(WebGLContextId, VRCompositorCommand),
+    // An opaque handle that must be passed to the render notifier. It is used by Gecko
```
kvark: please use the `///` doc comments

webrender_traits/src/types.rs
```diff
@@ -55,7 +55,25 @@ pub enum ApiMsg {
...
+    // to forward gecko-specific messages to the render thread preserving the ordering
+    // within the other messages.
+    ExternalEvent(ExternalEvent),
+}
+
+#[derive(Clone, Deserialize, Serialize)]
+pub struct ExternalEvent {
+    // An opaque pointer-sized value.
```
kvark: `///` comments

webrender_traits/src/types.rs
kvark: `///` comments

kvark: And yes, it does look like clone-ability for messages is an intrinsic requirement for replay in general, so we can't get rid of that.

webrender_traits/src/types.rs
```diff
@@ -424,6 +442,7 @@ pub trait RenderNotifier: Send {
     fn new_frame_ready(&mut self);
     fn new_scroll_frame_ready(&mut self, composite_needed: bool);
     fn pipeline_size_changed(&mut self, pipeline_id: PipelineId, size: Option<LayoutSize>);
+    fn external_event(&mut self, _evt: ExternalEvent) { unimplemented!() }
```
kvark: why does it need a default implementation if it's not implemented?

nical: This is only implemented in Gecko. Adding a default implementation that panics is to make sure we don't accidentally start using it in servo and to avoid making this a breaking change for servo.

kvark: Thanks, makes sense. Worth a comment to describe, but it's not a blocker.

kvark:
Thanks @nical !
@bors-servo r+

glennw:
@nical I'm totally fine with this approach - it seems good to me. I would be interested to know what sort of things it's used for though, if that's easy to explain? Mostly just for my own interest, but also to find out if there's some useful API surface that we are going to need eventually for Servo anyway.

nical:
Pretty much everything that is sent to the render thread on the gecko side will pass through this so that there is no race between these messages and WebRender ones. Examples are the creation and destruction of Renderers (there is one per window but they all live on the render thread), shutting down the thread, and a bunch of texture sharing related things (the gecko side of the external image implementation will need some communication to happen between the compositor thread and the render thread, and it will go through this).

glennw:
@nical Makes sense, thanks for the explanation!

PR #693, opened 2017-01-08
Implement Default for RendererOptions

vvuk:
implement Default instead of the builder

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/693)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1495,3 +1495,23 @@ pub struct RendererOptions {
     pub clear_color: ColorF,
     pub render_target_debug: bool,
 }
+
+impl Default for RendererOptions {
+    fn default() -> RendererOptions {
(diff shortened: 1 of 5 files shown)


glennw:
@vvuk We should also use the default options in the replay/ dir as well, I guess? Apart from that, looks good to me.

webrender/src/renderer.rs
```diff
@@ -1495,3 +1495,23 @@ pub struct RendererOptions {
     pub clear_color: ColorF,
     pub render_target_debug: bool,
 }
+
+impl Default for RendererOptions {
+    fn default() -> RendererOptions {
+        RendererOptions {
+            device_pixel_ratio: 1.0,
+            resource_override_path: None,
+            enable_aa: false,
```
glennw: Let's make this default to true.

PR #695, opened 2017-01-10
Wrench Reftests

jrmuizel:
This is my work-in-progress that adds reftesting functionality to wrench

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/695)


glennw:
@jrmuizel Cool! Did you want this stuff reviewed / merged now, or wait until the [WIP] is removed?

jrmuizel:
I'll put a reviewable version up tomorrow.

kvark:
Can we have the `mask.yaml`, `mask.png`, and the reftest list in different folder than `src`?

wrench/src/reftest.rs
```diff
@@ -0,0 +1,78 @@
...
+        let s = &l[0..l.find("#").unwrap_or(l.len())];
+        let s = s.trim();
+        if l.len() == 0 {
+            continue;
+        }
+        let mut items = s.split_whitespace();
+        let kind = match items.next() {
+            Some("==") => ReftestOp::Equal,
+            Some("!=") => ReftestOp::NotEqual,
+            _ => panic!()
```
kvark: nit: maybe just `continue` with a warning instead

wrench/src/reftest.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+fn render_yaml(wrench: &mut Wrench, window: &mut WindowWrapper, filename: &str) -> Vec<u8>
+{
+    let mut reader = YamlFrameReader::new(Path::new(filename));
+    reader.do_frame(wrench);
+    wrench.render();
+    window.swap_buffers();
+    let size = window.get_inner_size();
+    gl::read_pixels(0, 0,
```
kvark: uh oh, we need to hide it behind WR's device abstraction

wrench/src/reftest.rs
```diff
@@ -0,0 +1,78 @@
...
+                    size.0 as gl::GLsizei,
+                    size.1 as gl::GLsizei,
+                    gl::RGB, gl::UNSIGNED_BYTE)
```
kvark: Can we use `RGBA` instead in order to avoid internal format conversion?

wrench/src/reftest.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+pub fn run_reftests(wrench: &mut Wrench, window: &mut WindowWrapper, filename: &str)
+{
+    parse_reftests(filename, |t: Reftest|
+                   {
```
kvark: formatting

jrmuizel:
This should be ready for review now.

wrench/src/reftest.rs
```diff
@@ -0,0 +1,108 @@
...
+pub fn parse_reftests<F>(filename: &str, mut runner: F)
+    where F: FnMut(Reftest)
+{
+    let manifest = Path::new(filename);
+    let dir = manifest.parent().unwrap();
+    let f = File::open(manifest).unwrap();
+    let file = BufReader::new(&f);
+    for line in file.lines() {
+        let l = line.unwrap();
+        if l.starts_with("#") {
```
kvark: not really needed here, the next `continue` would do

kvark:
Thanks @jrmuizel !
It appears that we are already using `gl::read_pixels` in `wrench`, so a fix to both this and the older case could be a follow-up.
@glennw leaving for you to either review or merge with r=me

wrench/src/main.rs
```diff
@@ -54,6 +54,9 @@ use yaml_frame_reader::YamlFrameReader;
 
 mod yaml_frame_writer;
 mod json_frame_writer;
+mod scene;
```
kvark: this looks like a leftover, and it fails to build

wrench/src/reftest.rs
```diff
@@ -0,0 +1,108 @@
...
+            continue;
+        }
+
+        // strip the comments
+        let s = &l[0..l.find("#").unwrap_or(l.len())];
+        let s = s.trim();
+        if l.len() == 0 {
```
kvark: that is comparing `l.len()` and not `s.len()`? I misread it at first.

wrench/src/reftest.rs
kvark: actually, the next `continue` is checking for `l.len()==0` so it's different

wrench/src/reftest.rs
```diff
@@ -0,0 +1,108 @@
...
+    let file = BufReader::new(&f);
+    for line in file.lines() {
+        let l = line.unwrap();
+        if l.starts_with("#") {
+            continue;
+        }
+
+        // strip the comments
+        let s = &l[0..l.find("#").unwrap_or(l.len())];
+        let s = s.trim();
```
kvark: a simpler way, possibly, is: `let s = l.split('#').next().unwrap().trim()`

glennw:
This is awesome, thanks!

PR #696, opened 2017-01-10
Use segments when drawing clip masks, decouple from tiles.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/696)


glennw:
@kvark This implements the clip mask changes we discussed earlier. Instead of applying the more general optimization, it just handles the common case of a single complex rect, and draws the corners only in this case. But this PR is easily extendable to the general case, either as a follow up or before merging.

This code passes the tests, and is fast, but I'm not super happy with it. *Subjectively*, it feels much more complicated than the other method to me. It is also less efficient in the common case, since the rectangle clip shader has to evaluate all four clip corners, whereas with the other method it only handles one clip corner at a time.

But anyway, I'm interested to hear your thoughts, and specifically (a) if you agree that this code solution is more complicated than the other option and (b) if you can think of a nicer / cleaner way to implement this functionality.

glennw:
@kvark Hmmm, it probably has to be this way actually, since I think the other method can suffer from double AA issues at the edges of clip corners. Need to think about that a bit more, but it may rule out the other method anyway.

kvark:
I'm not completely following you on AA issues, but you are making a good point for me on this part. If we are talking about MSAA, then in your approach we'd only need to super-sample the corner draws, while the main draw call of the primitive would use regular multi-sample. That is going to be much faster than my approach, which would need to super-sample the whole mask. This concerns only applies to MSAA though, which I'm not sure is what you were referring to.

webrender/src/tiling.rs
```diff
@@ -2663,10 +2600,58 @@ impl FrameBuilder {
...
+                                    profile_counters.visible_primitives.inc();
+                                }
+                            }
                         }
                     }
                 }
                 &PrimitiveRunCmd::PopStackingContext => {
+                    let sc_index = *layer_stack.last().unwrap();
+                    let layer = &mut self.layer_store[sc_index.0];
+                    if layer.is_visible() {
```
kvark: the push stuff is using a different check - `can_contribute_to_scene()`, so there might be a bug here

webrender_traits/src/display_item.rs
```diff
@@ -129,17 +129,16 @@ impl ComplexClipRegion {
     }
 
     //TODO: move to `util` module?
-    /// Return a maximum aligned rectangle that is fully inside the clip region.
+    /// Return an aligned rectangle that is fully inside the clip region.
```
kvark: why is `k` removed from here?

webrender/src/tiling.rs
```diff
@@ -2663,10 +2600,58 @@ impl FrameBuilder {
...
+                                // Try to create a mask if we may need to.
+                                if prim_clip_info.is_some() || !clip_info_stack.is_empty() {
+                                    let mask_opt = RenderTask::new_mask(prim_bounding_rect,
+                                                                        MaskCacheKey::Primitive(prim_index),
+                                                                        &clip_info_stack,
+                                                                        &self.layer_store);
+                                    match mask_opt {
+                                        MaskResult::Outside => {
+                                            // Primitive is completely clipped out.
+                                            prim_metadata.clip_task = None;
```
glennw: Fixed

webrender/src/tiling.rs
glennw: Fixed

webrender/src/tiling.rs
```diff
@@ -1149,46 +1144,60 @@ impl RenderTask {
...
+
+        // TODO(gw): This optimization is very conservative for now.
+        //           For now, only draw optimized geometry if it is
+        //           a single aligned rect mask with rounded corners.
+        //           In the future, we'll expand this to handle the
+        //           more complex types of clip mask geometry.
+        let mut geometry_kind = MaskGeometryKind::Default;
+
+        if inner_rect.is_some() && clips.len() == 1 {
+            let &(ref sc_index, ref clip_info) = clips.first().unwrap();
```
glennw: Fixed

webrender_traits/src/display_item.rs
glennw: I use the inner rect to generate the vertices for the corners in the vertex shader - so it needs to extend to the radius for this technique. I think that is fine now, for the way inner_rect is now used. But perhaps in the future we could add a mode to get_inner_rect to optionally get the maximal inner rect, if that's useful elsewhere?

glennw:
@kvark OK, comments addressed and [wip] removed. Thanks!

webrender_traits/src/display_item.rs
kvark: Makes sense, yeah. I was just shocked by the removal of my carefully computed mathematical constant that was supposed to optimize everything :)

kvark:
Thanks @glennw , :shipit: :)
@bors-servo r+

PR #699, opened 2017-01-10
Add a ShutDown API message and RenderNotifier callback.

nical:
The bulk of the shutdown implementation is up to the program embedding WebRender. This new API message makes it possible to stop a render backend explicitly instead of waiting for an error to happen on the receiver.
r? @kvark @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/699)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -348,15 +348,28 @@ impl RenderBackend {
                             }
                         }
                         ApiMsg::ExternalEvent(evt) => {
-                            let mut notifier = self.notifier.lock();
+                            let notifier = self.notifier.lock();
                             notifier.unwrap()
                                     .as_mut()
                                     .unwrap()
                                     .external_event(evt);
                         }
+                        ApiMsg::ShutDown => {
+                            let notifier = self.notifier.lock();
+                            notifier.unwrap()
+                                    .as_mut()
+                                    .unwrap()
+                                    .shut_down();
+                            break;
+                        }
                     }
                 }
                 Err(..) => {
+                    let notifier = self.notifier.lock();
+                    notifier.unwrap()
+                            .as_mut()
+                            .unwrap()
+                            .shut_down();
                     break;
                 }
             }
--- a/webrender_traits/src/api.rs
+++ b/webrender_traits/src/api.rs
@@ -248,6 +248,10 @@ impl RenderApi {
         self.api_sender.send(msg).unwrap();
     }
 
+    pub fn shut_down(&self) {
+        self.api_sender.send(ApiMsg::ShutDown).unwrap();
+    }
+
     #[inline]
     fn next_unique_id(&self) -> (u32, u32) {
         let IdNamespace(namespace) = self.id_namespace;
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -60,6 +60,7 @@ pub enum ApiMsg {
     /// to forward gecko-specific messages to the render thread preserving the ordering
     /// within the other messages.
     ExternalEvent(ExternalEvent),
+    ShutDown,
 }
 
 /// An opaque pointer-sized value.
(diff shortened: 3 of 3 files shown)


nical:
The second commit is begging for a followup. I don't know what the error type should look like. the ipc channel uses serde's DeserializeError while this implementation used to advertise std::io::Error but never actually returned an error. I suppose that in order to meaningfully handle the errors we'd want the same type for both scenarios. In the mean time returning ```Err(())``` seems to be what best reflects the state of things.

webrender_traits/src/channel_mpsc.rs
```diff
@@ -29,7 +30,7 @@ pub struct MsgReceiver<T> {
 
 impl<T> MsgReceiver<T> {
     pub fn recv(&self) -> Result<T, Error> {
-        Ok(self.rx.recv().unwrap())
+        if let Ok(msg) = self.rx.recv() { Ok(msg) } else { Err(()) }
```
kvark: Could be written as `self.rx.recv().map_err(|_e| ())`

webrender/src/render_backend.rs
```diff
@@ -348,15 +348,28 @@ impl RenderBackend {
...
                         ApiMsg::ExternalEvent(evt) => {
-                            let mut notifier = self.notifier.lock();
+                            let notifier = self.notifier.lock();
                             notifier.unwrap()
                                     .as_mut()
                                     .unwrap()
                                     .external_event(evt);
                         }
+                        ApiMsg::ShutDown => {
+                            let notifier = self.notifier.lock();
```
kvark: just wondering - why do we always have this `notifier` variable? can't we just go `self.notifier.lock().unwrap().as_mut()....` and so on?

webrender_traits/src/api.rs
```diff
@@ -248,6 +248,10 @@ impl RenderApi {
         self.api_sender.send(msg).unwrap();
     }
 
+    pub fn shut_down(&self) {
```
kvark: It would be nice to have this called automatically when all the senders are gone, although I don't see an easy way to do that.

webrender_traits/src/channel_mpsc.rs
```diff
@@ -3,7 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use serde::{Deserialize, Serialize};
-use std::io::Error;
```
kvark: what is the reason we don't want `io::Error` here?

webrender_traits/src/api.rs
nical: Yeah, I'm not found of being able to shutdown before all of the api senders are gone. Worth doing as a followup (this blocks some gecko work so I'd prefer to iterate towards a good solution rather than take the time to figure out the perfect one the first time).

webrender/src/render_backend.rs
nical: I have no idea. I shamelessly copy-pasted this part.

webrender_traits/src/channel_mpsc.rs
nical: We don't use io::Error at all here. we could return std::sync::mpsc::RecvError, but since the channel_ipc version returns another type of error, we won't be able to poke at the error value anyway (other than print it to stdout).

nical:
I'll update the patch tomorrow to return the mpsc::RecvError and do the more elegant map_err thing.

webrender_traits/src/api.rs
kvark: I was thinking along the following line:
  - add an `Arc<()>` field to always accompany `MsgSender`
  - hide `MsgSender` from public interfaces to enforce ^
  - when any owner of `MsgSender` gets destroyed, it can check if that dummy `Arc` reference is the last one
  - so that it can send the `Shutdown` signal before completely going down

This is a bit crappy/hacky, but it would work.

webrender_traits/src/channel_mpsc.rs
jrmuizel: This is my version of the same change: https://github.com/jrmuizel/webrender/commit/cdcf5af0a8d336563ca652b5e2aec5426001d2de

nical: Looks good, I'll remove this change from my PR and let you submit this instead.

kvark:
btw, you also seem to use the dark magic of changing the commit with force-pushing but preserving the date. This screws up Github a bit.
@bors-servo r+

PR #701, opened 2017-01-10
Don't generate a frame when calling set_root_pipeline

jrmuizel:
The current API causes us generate an extra frame when setting
the root pipeline. I can't think of reason why this would be desirable
behaviour. Getting rid of the extra frame also makes things easier
for the reftesting code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/701)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -214,6 +214,10 @@ impl RenderBackend {
                             }
                         }
                         ApiMsg::SetRootPipeline(pipeline_id) => {
+                            if scene.pipeline_map.get(&pipeline_id).is_none() {
+                                return;
+                            }
+
                             let frame = profile_counters.total_time.profile(|| {
                                 self.scene.set_root_pipeline_id(pipeline_id);
 


glennw:
@jrmuizel This is by design (although maybe the API design itself is wrong). For instance, when we navigate back/forward in Servo, we can set the root pipeline. Could you explain the flow that results in an extra frame?

jrmuizel:
Imagine you're drawing for the first time. If you call set_root_pipeline(), set_root_display_list() you get a black frame and then actual frame.

jrmuizel:
Or an alternative scenario of frame duplication during switching to another pipeline
- set_root_pipeline(Pipeline(1)) # paints black
- set_root_display_list(Pipeline(1)) # paints pipeline 1
- set_root_display_list(Pipeline(2)) # paints pipeline 1
- set_root_pipeline(Pipeline(2)) # paints pipeline 2

glennw:
What I'd expect to happen in each of those flows is:

1) Calling set_root_pipeline() when there is no root display list (for the first frame) is a no-op and nothing gets drawn.

2) First set_root_pipeline() is no-op. First set_root_display_list() results in drawing P1. 2nd set_root_display_list does nothing (since it's not in the root frame). 2nd set_root_pipeline() draws P2.

That would be my expectation - but perhaps it doesn't behave that way now due to bugs.

Perhaps it would just be simpler to make those API calls never generate a frame, and you have to explicitly call generate_frame() to draw?

jrmuizel:
Your expectations sounds reasonable, but I think there are bugs. That being said explicitly asking for a frame seems like it would keep things simpler.

glennw:
https://github.com/servo/webrender/issues/702

jrmuizel:
I've updated this pull request with a more conservative option. (If this gets merged will it still use the original description?)

glennw:
@jrmuizel You can click [Edit] beside the PR title in github to update it.

jrmuizel:
Let's just use a new one: #703

PR #703, opened 2017-01-10
Avoid frame notification when root pipeline doesn't exist

jrmuizel:
If we don't have a pipeline to draw don't try to draw it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/703)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -214,9 +214,13 @@ impl RenderBackend {
                             }
                         }
                         ApiMsg::SetRootPipeline(pipeline_id) => {
-                            let frame = profile_counters.total_time.profile(|| {
-                                self.scene.set_root_pipeline_id(pipeline_id);
+                            self.scene.set_root_pipeline_id(pipeline_id);
 
+                            if self.scene.display_lists.get(&pipeline_id).is_none() {
+                                continue;
+                            }
+
+                            let frame = profile_counters.total_time.profile(|| {
                                 self.build_scene();
                                 self.render()
                             });


webrender/src/render_backend.rs
```diff
@@ -214,6 +214,10 @@ impl RenderBackend {
                             }
                         }
                         ApiMsg::SetRootPipeline(pipeline_id) => {
+                            if scene.pipeline_map.get(&pipeline_id).is_none() {
```
kvark: This doesn't look right to me. This is a setter function that checks for the value to already be there? When would it ever set the pipeline then?

kvark:
I think it should be checking for the root display list in there, not the root pipeline.

webrender/src/render_backend.rs
kvark: Edit: I'm wrong, the pipeline is also set up in `set_root_display_list`. Perhaps, it would be cleaner (and less confusing) to check for the display list here?

jrmuizel: No this change is still wrong. We should be unconditionally calling set_root_pipeline_id. I originally did that but incorrectly changed the order.

jrmuizel:
This version should be saner.

webrender/src/render_backend.rs
```diff
@@ -214,9 +214,13 @@ impl RenderBackend {
                             }
                         }
                         ApiMsg::SetRootPipeline(pipeline_id) => {
-                            let frame = profile_counters.total_time.profile(|| {
-                                self.scene.set_root_pipeline_id(pipeline_id);
+                            self.scene.set_root_pipeline_id(pipeline_id);
+
+                            if self.scene.display_lists.get(&pipeline_id).is_none() {
```
kvark: uber-nit: could use `!self.scene.display_lists.contains_key(&pipeline_id)`
it's totally irrelevant here, I know :)

PR #704, opened 2017-01-11
Create subpixel text on mac to match gecko's rendering

changm:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/704)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -6,10 +6,10 @@ use app_units::Au;
 use core_graphics::base::{kCGImageAlphaNoneSkipFirst, kCGImageAlphaPremultipliedLast};
 use core_graphics::base::kCGBitmapByteOrder32Little;
 use core_graphics::color_space::CGColorSpace;
-use core_graphics::context::CGContext;
+use core_graphics::context::{CGContext, CGTextDrawingMode};
 use core_graphics::data_provider::CGDataProvider;
 use core_graphics::font::{CGFont, CGGlyph};
-use core_graphics::geometry::CGPoint;
+use core_graphics::geometry::{CGPoint, CGSize, CGRect};
 use core_text::font::CTFont;
 use core_text::font_descriptor::kCTFontDefaultOrientation;
 use core_text;
@@ -158,6 +158,23 @@ impl FontContext {
         })
     }
 
+    #[allow(dead_code)]
+    fn print_glyph_data(&mut self, data: &Vec<u8>, width: usize, height: usize) {
+        // Rust doesn't have step_by support on stable :(
+        for i in 0..height {
+            let current_height = i * width * 4;
+
+            for pixel in data[current_height .. current_height + (width * 4)].chunks(4) {
+                let b = pixel[0];
+                let g = pixel[1];
+                let r = pixel[2];
+                let a = pixel[3];
+                print!("({}, {}, {}, {}) ", r, g, b, a);
+            }
+            println!("");
+        }
+    }
+
     pub fn rasterize_glyph(&mut self,
                            font_key: FontKey,
                            size: Au,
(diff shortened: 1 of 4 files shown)


webrender_traits/Cargo.toml
```diff
@@ -25,7 +25,7 @@ serde_derive = {version = "0.8", optional = true}
 ipc-channel = { version = "0.5.0", optional = true }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-core-graphics = "0.4"
+core-graphics = "0.4.3"
```
kvark: this is not needed, given that the logic that depends on 0.4.3 is in `webrender` and it already forces the `0.4.3` version minimum.

webrender/src/platform/macos/font.rs
```diff
@@ -189,21 +209,53 @@ impl FontContext {
...
+                cg_context.set_rgb_fill_color(0.0, 0.0, 0.0, 1.0);
+
                 ct_font.draw_glyphs(&[glyph], &[rasterization_origin], cg_context.clone());
 
                 let rasterized_area = (metrics.rasterized_width * metrics.rasterized_height) as usize;
                 let mut rasterized_pixels = cg_context.data().to_vec();
 
+                // We need to invert the pixels back since right now
+                // transparent pixels are actually white.
+                for i in 0..rasterized_pixels.len() {
```
kvark: nit: use an iterator:
```rust
for p in rasterizer_pixels.iter_mut() {...}
```

webrender/src/platform/macos/font.rs
```diff
@@ -6,10 +6,10 @@ use app_units::Au;
 use core_graphics::base::{kCGImageAlphaNoneSkipFirst, kCGImageAlphaPremultipliedLast};
 use core_graphics::base::kCGBitmapByteOrder32Little;
 use core_graphics::color_space::CGColorSpace;
-use core_graphics::context::CGContext;
+use core_graphics::context::{CGContext,CGTextDrawingMode};
```
kvark: formatting: missing space after the comma

webrender/src/platform/macos/font.rs
```diff
@@ -6,10 +6,10 @@ use app_units::Au;
...
 use core_graphics::data_provider::CGDataProvider;
 use core_graphics::font::{CGFont, CGGlyph};
-use core_graphics::geometry::CGPoint;
+use core_graphics::geometry::{CGPoint,CGSize,CGRect};
```
kvark: formatting: formatting: missing spaces after the commas

webrender/src/platform/macos/font.rs
```diff
@@ -158,6 +158,26 @@ impl FontContext {
         })
     }
 
+    #[allow(dead_code)]
+    fn print_glyph_data(&mut self, cg_context: &mut CGContext, width: usize, height: usize)
+    {
```
kvark: formatting: put opening bracket on the same line

webrender/src/platform/macos/font.rs
```diff
@@ -158,6 +158,26 @@ impl FontContext {
...
+        let data = cg_context.data();
+        // Rust doesn't have step_by support on stable :(
+        for i in 0..height {
+            let current_height = i * width * 4;
+            let mut j = 0;
+
+            while j < width * 4 {
```
kvark: you can use `chunks`:
```rust
for d in data[current_height .. current_height+width*4].chunks(4) {
  // do something with d[0], d[1], d[2]
}
```

webrender/src/platform/macos/font.rs
```diff
@@ -189,21 +209,53 @@ impl FontContext {
                     FontRenderMode::Mono => (false, false),
                 };
 
+                cg_context.set_allows_font_subpixel_positioning(true);
```
kvark: shouldn't we be checking for `RendererOptions::enable_aa`?

webrender/src/platform/macos/font.rs
```diff
@@ -189,21 +209,53 @@ impl FontContext {
...
+                };
+
+                // Fill white
+                cg_context.fill_rect(rect);
+
+                // Now draw black
+                cg_context.set_text_drawing_mode(CGTextDrawingMode::CGTextFill);
+                cg_context.set_rgb_fill_color(0.0, 0.0, 0.0, 1.0);
+
                 ct_font.draw_glyphs(&[glyph], &[rasterization_origin], cg_context.clone());
```
kvark: Hmm, so previously the text was drawn with the white color, and now it's drawn with black. Is that intended?
We might need to run the servo test suite on Mac to see if it breaks.

changm: Yes, this is intentional. See the comment on line 221 or https://bugzilla.mozilla.org/show_bug.cgi?id=1230366#c35 for more information. Essentially, we have to draw black text on a white background since it produces a different glyph than white text on black.

changm: @kvark r?

kvark: Thanks @changm ! The code looks good now, but I need to make sure this will not break Servo. I'll run the test suite and get back to you.

glennw: @kvark I have access to a mac I can run the tests on today, if that will be quicker?

kvark: @glennw I'm still building servo on my Mac, should be able to run the tests very soon.

kvark:
I can confirm the change breaks quite a few tests in Servo. The discussion/investigation leaded us to a problem with the alpha.

webrender/src/platform/macos/font.rs
```diff
@@ -189,21 +206,52 @@ impl FontContext {
...
+                        height: metrics.rasterized_height as f64,
+                    }
+                };
+
+                // Fill white
+                cg_context.fill_rect(rect);
+
+                // Now draw black
+                cg_context.set_text_drawing_mode(CGTextDrawingMode::CGTextFill);
+                cg_context.set_rgb_fill_color(0.0, 0.0, 0.0, 1.0);
```
kvark: Given that both the rectangle and the text have alpha of `1.0`, it will be guaranteed to invert to `0.0` afterwards.

kvark: @changm Note: setting the alpha here to `0.0` does not fix the tests

kvark: Test results - [wpt.log.zip](https://github.com/servo/webrender/files/700362/wpt.log.zip)

webrender/src/platform/macos/font.rs
```diff
@@ -189,21 +206,62 @@ impl FontContext {
...
+                // transparent pixels are actually white.
+                for i in 0..metrics.rasterized_height {
+                    let current_height :usize = (i * metrics.rasterized_width * 4) as usize;
+                    let end_row :usize = current_height + (metrics.rasterized_width as usize * 4);
+
+                    for mut pixel in rasterized_pixels[current_height .. end_row].chunks_mut(4) {
+                        pixel[0] = 255 - pixel[0];
+                        pixel[1] = 255 - pixel[1];
+                        pixel[2] = 255 - pixel[2];
+                        pixel[3] = 255;
```
changm: I fixed the alpha here.

webrender/src/platform/macos/font.rs
changm: I checked what gecko does if both AA and subpixel AA are disabled. We still draw with black text on a white background.

kvark:
@changm it works much better now, failing only two tests:
  - absolute_hypothetical_with_intervening_inline_block_a
  - position_fixed_a

Both failures seem to be subtle differences in the bottom line of things. You can see for yourself by uploading this (unpacked) log to the [reftest analyser](http://hoppipolla.co.uk/410/reftest-analyser-structured.xhtml):
[wpt.log.zip](https://github.com/servo/webrender/files/700578/wpt.log.zip)

changm:
@kvark Let's try this again. I was able to reproduce the servo reftest failure. For mono + grayscale AA, WR only cares about the alpha value. When filling the background with an opaque value, the alpha values of the glyph are also set to opaque (255), so therefore all text was actually just blocks of black. There's a nice comment in here that explains what values we get out of CG based on different bg/fg fill values. This patch fixes it so that for mono / grayscale AA, we set the RGB + Alpha values to the alpha values WR was producing before this patch.

kvark:
Awesome, thanks @changm !
I'll make sure to test it tonight and merge if everything is good.

webrender/src/platform/macos/font.rs
```diff
@@ -183,38 +200,94 @@ impl FontContext {
...
-                match render_mode {
-                    FontRenderMode::Alpha | FontRenderMode::Mono => {
-                        for i in 0..rasterized_area {
-                            let alpha = (rasterized_pixels[i * 4 + 3] as f32) / 255.0;
-                            for channel in &mut rasterized_pixels[(i*4)..(i*4 + 3)] {
-                                *channel = ((*channel as f32) / alpha) as u8;
+                // We need to invert the pixels back since right now
+                // transparent pixels are actually opaque white.
+                for i in 0..metrics.rasterized_height {
+                    let current_height :usize = (i * metrics.rasterized_width * 4) as usize;
```
kvark: please drop the explicit type, since you cast it anyways

webrender/src/platform/macos/font.rs
```diff
@@ -183,38 +200,94 @@ impl FontContext {
...
+                    let end_row :usize = current_height + (metrics.rasterized_width as usize * 4);
```
kvark: same here - please drop explicit type

webrender/src/platform/macos/font.rs
```diff
@@ -183,38 +200,94 @@ impl FontContext {
...
+
+                    for mut pixel in rasterized_pixels[current_height .. end_row].chunks_mut(4) {
+                        pixel[0] = 255 - pixel[0];
+                        pixel[1] = 255 - pixel[1];
+                        pixel[2] = 255 - pixel[2];
+
+                        match render_mode {
```
kvark: could be better rewritten as
```rust
pixel[3] = match render_mode {
  FontRenderMode::Subpixel => 255,
  _ => {
    assert_eq!(pixel[0], pixel[1]);
    assert_eq!(pixel[0], pixel[2]);
    pixel[0]
  }
};
```

changm:
Thanks! Fixed the last review comments.

kvark:
Awesome!
@changm sorry about being so picky! I'm glad you made it to the end :)
@bors-servo r+

emilio:
@bors-servo retry

Travis seems to have stuck building osmesa

PR #705, opened 2017-01-11
Make the intermediate render targets select the correct size dynamically.

glennw:
The render targets need to be >= framebuffer size, to ensure they
can allocate worst case render tasks.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/705)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -37,7 +37,7 @@ use webrender_traits::{BorderDisplayItem, BorderSide, BorderStyle, YuvColorSpace
 use webrender_traits::{AuxiliaryLists, ItemRange, BoxShadowClipMode, ClipRegion};
 use webrender_traits::{PipelineId, ScrollLayerId, WebGLContextId, FontRenderMode};
 use webrender_traits::{DeviceIntRect, DeviceIntPoint, DeviceIntSize, DeviceIntLength, device_length};
-use webrender_traits::{DeviceUintSize, DeviceUintPoint, DeviceSize};
+use webrender_traits::{DeviceUintSize, DeviceUintPoint};
 use webrender_traits::{LayerRect, LayerPoint, LayerSize};
 use webrender_traits::{LayerToScrollTransform, LayerToWorldTransform, WorldToLayerTransform};
 use webrender_traits::{WorldPoint4D, ScrollLayerPixel, as_scroll_parent_rect};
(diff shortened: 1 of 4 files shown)


webrender/src/tiling.rs
```diff
@@ -960,11 +959,13 @@ pub struct RenderPass {
 }
 
 impl RenderPass {
-    fn new(pass_index: isize, is_framebuffer: bool) -> RenderPass {
+    fn new(pass_index: isize,
+           is_framebuffer: bool,
+           size: DeviceUintSize) -> RenderPass {
```
kvark: let's store the size inside a `RenderPass` in order to avoid asking for it in `allocate_target`?

webrender/src/tiling.rs
```diff
@@ -2872,6 +2875,20 @@ impl FrameBuilder {
...
+        // always be able to allocate the worst case render task (such as a clip mask that
+        // covers the entire screen).
+        // However, there are some extremely subtle rounding errors that occur in the
+        // reftests under OSMesa if the cache targets are exactly the size of the
+        // framebuffer. To work around this, we'll align the cache size to a multiple
+        // of the tile size. This can be removed once the tiling code is gone.
+        // TODO(gw): Remove this hack once the tiling code is sorted out!!
+        let max_dimension = cmp::max(screen_rect.size.width, screen_rect.size.height);
+        let aligned_max_dimension = (max_dimension + SCREEN_TILE_SIZE - 1) & !(SCREEN_TILE_SIZE-1);
+        let cache_size = DeviceUintSize::new(aligned_max_dimension as u32,
```
kvark: if this code wasn't temporary, I'd ask for each of width/height to be rounded up separately

webrender_traits/src/lib.rs
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #![cfg_attr(feature = "nightly", feature(nonzero))]
-#![cfg_attr(feature = "serde_derive", feature(proc_macro, rustc_attrs, structural_match))]
```
kvark: is this no longer needed for serde? I assume there is a compile warning at the moment then.

webrender/src/tiling.rs
glennw: Fixed - had to change the code structure slightly to make this work, but it's much cleaner now.

webrender/src/tiling.rs
glennw: It was originally separate - it actually needs to be square to avoid the rounding glitch in OSMesa :|

webrender_traits/src/lib.rs
glennw: Correct

glennw:
@kvark Thanks for the review - updated to address comments. I'll squash once you're happy with the changes.

kvark:
@glennw thanks!

PR #707, opened 2017-01-11
Add cargo:rerun-if-changed directive to wrench

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/707)


jrmuizel:
This is already there in a better form. How I managed to write it again without noticing is a mystery to me.

PR #710, opened 2017-01-11
Work around an older mac shader compiler bug.

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/710)


glennw:
Thanks!

PR #711, opened 2017-01-12
Select 24 bit z-buffer for wrench headless mode.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/711)


glennw:
r? @kvark 
cc @jrmuizel

PR #712, opened 2017-01-13
Update CoreGraphics glyph metrics to be like skia too

changm:
@kvark  Maybe see if this fixes it for you too? This updates the glyph metrics by rounding the metrics from CG to pixel boundaries and then expanding them a bit for LCD Smoothing, which is what skia does.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/712)


kvark:
@changm with this PR, 56 out of 533 tests are failing unexpectably.

kvark:
Closing in favour of #704

Issue #616, opened 2016-12-02
Validate shader code at build time


kvark:
Since we have `build.rs` now, we might as well hook up [glslang](https://github.com/KhronosGroup/glslang) and validate the shaders.

larsbergstrom:
You were way ahead of me :-)

Issue #665, opened 2016-12-22
Add CI for testing other shader compilers


larsbergstrom:
One issue that we've noticed (especially @dati91 and @mmatyas) is that WebRender shader changes often run afoul of the various "interesting" shader compilers on ODROID, Android, etc. e.g., #653 and #654 but also several Android fixes in the past.

Is it feasible to, say, create some VM images of various platforms that we think are likely to be representative of likely to have fragile shader code and run them as extra TravisCI builders?

Or is there a really solid shader verifier that we could be running as a build step?

cc @glennw

jrmuizel:
There are a couple of options here:

1. https://github.com/KhronosGroup/glslang is the reference compiler.
2. Once we get some basic testing up (#647) I'd like to get running on top of ANGLE. ANGLE should have a reasonably strict glsl compiler.

glennw:
@larsbergstrom Yup, I *think* that running one of the shader validator tools / ANGLE on CI will catch these errors, and be the simplest change to implement. But we could definitely consider having a VM to test android shader compiles on, if these errors are still not being caught.

kvark:
Speaking of glslang - #616

emilio:
Also, we have angle bindings for WebGL. They're far from complete, but we use them for compiling and validating shaders, so they should be usable for this purpose.

They live in https://github.com/servo/angle, and you can see how we use them in https://github.com/servo/servo/blob/master/components/script/dom/webglshader.rs

kvark:
Here is an example of integrating `glslang` into Travis CI: https://android.googlesource.com/platform/external/shaderc/shaderc/+/ndk-r11/.travis.yml

glennw:
We do shader validation on CI now.

PR #713, opened 2017-01-13
Add shader validation via ANGLE

mmatyas:
This adds shader validation to `cargo test` via ANGLE (use `-- --nocapture` to see the output). Tested on desktop (x86_64, GeForce 210) and Odroid XU3 (armv7, Mali T628), all shaders passed. Please review the shader concatenation if it's correct; there are some base GLSL files (eg. `cs_blur.glsl`), I wonder if they should be included too?

Requires https://github.com/servo/angle/pull/17 to land first.

Fixes https://github.com/servo/webrender/issues/616 and https://github.com/servo/webrender/issues/665 (we could still add `glslang` though).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/713)

### DIFF
--- a/webrender/tests/angle_shader_validation.rs
+++ b/webrender/tests/angle_shader_validation.rs
@@ -1,0 +1,85 @@
+extern crate angle;
+#[macro_use]
+extern crate lazy_static;
+extern crate webrender;
+
+use angle::hl::{BuiltInResources, Output, ShaderSpec, ShaderValidator};
+
+include!(concat!(env!("OUT_DIR"), "/shaders.rs"));
+
+
+// from glslang
+const FRAGMENT_SHADER: u32 = 0x8B30;
+const VERTEX_SHADER: u32 = 0x8B31;
+
+
+#[test]
+fn validate_shaders() {
+    angle::hl::initialize().unwrap();
+
+    let shared_src = SHADERS.get("shared").unwrap();
+    let prim_shared_src = SHADERS.get("prim_shared").unwrap();
+    let clip_shared_src = SHADERS.get("clip_shared").unwrap();
+
+    for (filename, file_source) in SHADERS.iter() {
+        let is_prim = filename.starts_with("ps_");
+        let is_clip = filename.starts_with("cs_");
+        let is_vert = filename.ends_with(".vs");
+        let is_frag = filename.ends_with(".fs");
+        if !(is_prim ^ is_clip) || !(is_vert ^ is_frag) {
+            continue;
+        }
+
+
+        let base_filename = filename.splitn(2, '.').next().unwrap();
+        let mut shader_prefix = format!("#version 300 es\n
+            // Base shader: {}\n
+            #define WR_MAX_VERTEX_TEXTURE_WIDTH {}\n",
+            base_filename, webrender::renderer::MAX_VERTEX_TEXTURE_WIDTH);
+
+        if is_vert {
+            shader_prefix.push_str("#define WR_VERTEX_SHADER\n");
+        } else {
+            shader_prefix.push_str("#define WR_FRAGMENT_SHADER\n");
+        }
+
+        let mut build_configs = vec!["#define WR_FEATURE_TRANSFORM\n"];
+        if is_prim {
+            // the transform feature may be disabled for the prim shaders
+            build_configs.push("// WR_FEATURE_TRANSFORM disabled\n");
+        }
+
+        for config_prefix in build_configs {
+            let mut shader_source = String::new();
+            shader_source.push_str(shader_prefix.as_str());
+            shader_source.push_str(config_prefix);
+            shader_source.push_str(shared_src);
+            shader_source.push_str(prim_shared_src);
+            if is_clip {
+                shader_source.push_str(clip_shared_src);
+            }
+            if let Some(optional_src) = SHADERS.get(base_filename) {
+                shader_source.push_str(optional_src);
+            }
+            shader_source.push_str(file_source);
+
+
+            let gl_type = if is_vert { VERTEX_SHADER } else { FRAGMENT_SHADER };
+            let resources = BuiltInResources::default();
+            let validator = ShaderValidator::new(gl_type,
+                                                 ShaderSpec::Gles3,
+                                                 Output::Essl,
+                                                 &resources).unwrap();
+
+            match validator.compile_and_translate(&[&shader_source]) {
+                Ok(_) => {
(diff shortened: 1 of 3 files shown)


webrender/tests/angle_shader_validation.rs
```diff
@@ -0,0 +1,77 @@
...
+                                             ShaderSpec::Gles3,
+                                             Output::Essl,
+                                             &resources).unwrap();
+
+        match validator.compile_and_translate(&[&shader_source]) {
+            Ok(_) => {
+                println!("Shader translated succesfully: {}", filename);
+            },
+            Err(_) => {
+                println!("Shader compilation failed: {}\n{}", filename, validator.info_log());
```
emilio: Should this panic?

webrender/tests/angle_shader_validation.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        let base_filename = filename.splitn(2, '.').next().unwrap();
+        let mut shader_source = format!("#version 300 es\n
+            // Base shader: {}\n
+            #define WR_MAX_VERTEX_TEXTURE_WIDTH {}\n",
+            base_filename, webrender::renderer::MAX_VERTEX_TEXTURE_WIDTH);
+
+        if is_vert {
+            shader_source.push_str("#define WR_VERTEX_SHADER\n");
+        }
```
emilio: nit: `} else {`

webrender/tests/angle_shader_validation.rs
mmatyas: If the validator was created successfully, it should not panic (there's an assert for that in `ShaderLang.cpp`). `ShaderValidator::new` fails if the requested feature wasn't built into ANGLE, but otherwise, it seems the ESSL translator's ctor that it calls doesn't do anything special. Any other shader error (eg. if it's not supported by the GPU) should be in the `info_log`.

webrender/tests/angle_shader_validation.rs
mmatyas: Fixed, thanks!

webrender/tests/angle_shader_validation.rs
emilio: Yes, my point is that if there's an error the fact that it's in the info log won't prevent the build from passing, so IMO this should panic so CI errors out when there are shader compilation failures.

webrender/tests/angle_shader_validation.rs
```diff
@@ -0,0 +1,76 @@
...
+            #define WR_MAX_VERTEX_TEXTURE_WIDTH {}\n",
+            base_filename, webrender::renderer::MAX_VERTEX_TEXTURE_WIDTH);
+
+        if is_vert {
+            shader_source.push_str("#define WR_VERTEX_SHADER\n");
+        } else {
+            shader_source.push_str("#define WR_FRAGMENT_SHADER\n");
+        }
+        if is_clip {
+            shader_source.push_str("#define WR_FEATURE_TRANSFORM\n");
```
kvark: All of the primitive shaders also need to be built with both on/off `WR_FEATURE_TRANSFORM`

mmatyas:
I've added a build matrix to test all the features (transform, subpixel aa, clip). What do you think?

webrender/tests/angle_shader_validation.rs
```diff
@@ -32,45 +32,80 @@ fn validate_shaders() {
...
-        if is_clip {
-            shader_source.push_str(clip_shared_src);
-        }
-        if let Some(optional_src) = SHADERS.get(base_filename) {
-            shader_source.push_str(optional_src);
+
+        let features = vec!["TRANSFORM", "SUBPIXEL_AA", "CLIP"];
+        let build_mx: Vec<_> = features.iter().map(|feature| vec![
+            format!("#define WR_FEATURE_{}\n", feature),
+            format!("// #define WR_FEATURE_{}\n", feature),
```
kvark: Is this for assisting the debugging?

webrender/tests/angle_shader_validation.rs
```diff
@@ -32,45 +32,80 @@ fn validate_shaders() {
...
+        ]).collect::<Vec<_>>();
```
kvark: no need for the explicit `<Vec<_>>` here

webrender/tests/angle_shader_validation.rs
```diff
@@ -32,45 +32,80 @@ fn validate_shaders() {
...
 
-        shader_source.push_str(shared_src);
-        shader_source.push_str(prim_shared_src);
-        if is_clip {
-            shader_source.push_str(clip_shared_src);
-        }
-        if let Some(optional_src) = SHADERS.get(base_filename) {
-            shader_source.push_str(optional_src);
+
+        let features = vec!["TRANSFORM", "SUBPIXEL_AA", "CLIP"];
```
kvark: nit: don't need to be an owned vector

mmatyas:
Ok, I've dropped the build matrix, now it only checks the disabled `WR_FEATURE_TRANSFORM`, if it's a a prim shader.
By the way, what kind of output would you like to see on shader error? Should I print out the full shader source?

webrender/tests/angle_shader_validation.rs
```diff
@@ -0,0 +1,85 @@
...
+            shader_prefix.push_str("#define WR_FRAGMENT_SHADER\n");
+        }
+
+        let mut build_configs = vec!["#define WR_FEATURE_TRANSFORM\n"];
+        if is_prim {
+            // the transform feature may be disabled for the prim shaders
+            build_configs.push("// WR_FEATURE_TRANSFORM disabled\n");
+        }
+
+        for config_prefix in build_configs {
```
kvark: I'm sorry, I might be missing something here. So there is a single element in `build_configs`, which mean that everything will be built with `WR_FEATURE_TRANSFORM` enabled. That doesn't sound right.

kvark:
Also, please merge when ready.

webrender/tests/angle_shader_validation.rs
mmatyas: The intention is that all shaders will be tested with `WR_FEATURE_TRANSFORM` enabled, but in addition, prim shaders will also be tested with `WR_FEATURE_TRANSFORM` disabled. Is this not how it should work? Sorry, I might have misunderstood your comment.

kvark: Ok, I see now. Looks correct, thank you!

PR #714, opened 2017-01-13
Require v0.4.2 of core-graphics for use of kCGBitmapByteOrder32Little

staktrace:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/714)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -37,5 +37,5 @@ freetype = {version = "0.1.2", default-features = false}
 dwrote = "0.1.4"
 
 [target.'cfg(target_os = "macos")'.dependencies]
-core-graphics = "0.4.1"
+core-graphics = "0.4.2"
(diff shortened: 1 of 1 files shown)


kvark:
You might want to update the `Cargo.lock` files for the sample/wrench as well.

staktrace:
I'm not following. The sample app and wrench share the top-level Cargo.lock file with webrender and webrender_traits. That Cargo.lock file already uses core-graphics 0.4.2 for webrender. That Cargo.lock file also pulls in core-graphics 0.3.2, but that's from a different dependency chain (glutin -> cocoa -> core-graphics) and to update that one we'd need to modify the cocoa package to use a newer core-graphics, which is outside the scope of this change. Is there something else I'm missing?

kvark:
@staktrace huh, I assumed `Cargo.lock` was using 0.4.1.
@bors-servo r+

PR #715, opened 2017-01-13
Make sure we track shader changes in build.rs

jrmuizel:
rerun-if-changed is not recursive so we need to make sure that
we output directive for all of the files as well as the directory

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/715)


kvark:
Thanks!
@bors-servo r+

PR #716, opened 2017-01-13
Update freetype with proper bindings.

emilio:
Needs https://github.com/servo/rust-freetype/pull/48, still need to port the servo code, but I'd rather leave this up for review.

r? @glennw or @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/716)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -31,7 +31,7 @@ webrender_traits = {path = "../webrender_traits", default-features = false}
 bitflags = "0.7"
 
 [target.'cfg(any(target_os = "android", all(unix, not(target_os = "macos"))))'.dependencies]
-freetype = {version = "0.1.2", default-features = false}
+freetype = { version = "0.2", default-features = false }
 
 [target.'cfg(target_os = "windows")'.dependencies]
 dwrote = "0.1.4"
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -5,9 +5,8 @@
 use app_units::Au;
 use webrender_traits::{FontKey, FontRenderMode, GlyphDimensions, NativeFontHandle};
 
-use freetype::freetype::{FTErrorMethods, FT_PIXEL_MODE_GRAY, FT_PIXEL_MODE_MONO, FT_PIXEL_MODE_LCD};
-use freetype::freetype::{FT_Done_FreeType, FT_RENDER_MODE_LCD, FT_Library_SetLcdFilter};
-use freetype::freetype::{FT_RENDER_MODE_NORMAL, FT_RENDER_MODE_MONO};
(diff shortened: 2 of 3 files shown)


webrender/Cargo.toml
```diff
@@ -31,7 +31,7 @@ webrender_traits = {path = "../webrender_traits", default-features = false}
 bitflags = "0.7"
 
 [target.'cfg(any(target_os = "android", all(unix, not(target_os = "macos"))))'.dependencies]
-freetype = {version = "0.1.2", default-features = false}
+freetype = { git = "https://github.com/servo/rust-freetype", branch = "bindgen", default-features = false }
```
kvark: I think we should be publishing 0.2 and using it here instead of the branch, which is about to be merged into master anyway.

webrender/src/platform/unix/font.rs
```diff
@@ -159,74 +159,71 @@ impl FontContext {
...
-                                        let byte_index = (y * bitmap.pitch) + (x >> 3);
-                                        let bit_index = x & 7;
-                                        let byte_ptr = bitmap.buffer.offset(byte_index as isize);
-                                        let bit = (*byte_ptr & (0x80 >> bit_index)) != 0;
-                                        if bit {
-                                            0xff
-                                        } else {
-                                            0
-                                        }
+                    if bitmap.pixel_mode == FT_Pixel_Mode::FT_PIXEL_MODE_MONO as u8 {
```
kvark: why is it not possible to `match` here any more?

webrender/src/render_backend.rs
```diff
@@ -426,12 +426,12 @@ impl RenderBackend {
...
                 None => {
                     // Was existing, not in current frame anymore
-                        let mut notifier = self.notifier.lock();
-                        notifier.as_mut()
-                                .unwrap()
-                                .as_mut()
-                                .unwrap()
-                                .pipeline_size_changed(pipeline_id, None);
+                    let mut notifier = self.notifier.lock();
+                    notifier.as_mut()
```
kvark: I don't think this first `as_mut` is needed, and neither is the `let mut` thing. It is not related to the PR, but neither is this change related to freetype :)

webrender/src/platform/unix/font.rs
emilio: Because now `FT_PIXEL_MODE_XXX` is a proper enum, and to match you'd need to transmute, or write a function to convert u8 -> FT_Pixel_Mode, but I thought it was not worth it.

emilio:
Updated to use the published version, backed out the alignment fix since it was completely unrelated.

emilio:
This is now the only patch blocking the freetype update in servo. @kvark or @glennw, is this fine to land?

glennw:
@emilio This looks good to me - feel free to merge when it suits.

PR #717, opened 2017-01-13
Add a nested clipping test case

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/717)


kvark:
Thanks!
@bors-servo r+

PR #718, opened 2017-01-13
wrench: Remove debugging println!

jrmuizel:
r? @vvuk

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/718)

### DIFF
--- a/wrench/src/yaml_frame_reader.rs
+++ b/wrench/src/yaml_frame_reader.rs
@@ -145,7 +145,6 @@ impl YamlFrameReader {
                 } else {
                     None
                 };
-                println!("R {:?} cx {:?}", bounds, complex);
                 Some(self.builder().new_clip_region(&bounds, complex, image_mask))
             }
             _ => {


kvark:
I suggest we use the standard `log` crate, where this kind of output would be marked with `info!`.

Manishearth:
@bors-servo r+

yeah,  but then you'd probably prettify the output. Also to avoid codesize bloat, we're probably not going to use the log crate at least for Gecko since it pulls in a dependency on regex which is large.

kvark:
@Manishearth 

> yeah, but then you'd probably prettify the output

Could you explain this bit?

> Also to avoid codesize bloat, we're probably not going to use the log crate at least for Gecko since it pulls in a dependency on regex which is large.

Looking at https://crates.io/crates/log, I don't see `regex` dependency.

Manishearth:
> Could you explain this bit?

Have a nicer log output message that makes sense

> I don't see regex dependency

oh yeah I was thinking of env_logger. This is fine.

kvark:
@Manishearth 

> yeah, but then you'd probably prettify the output
...
> Have a nicer log output message that makes sense

I think error prettiness is orthogonal to the use of either `info!` or `println!`. So I don't quite understand how this can apply as a counter-argument to the `log` crate, unless I misunderstand you (hence, my request to clarify).

Manishearth:
It's not a counterargument to the log crate (the regex thing was, but that's not correct anymore), it's just an argument for landing this as-is. We can add log statements if we want later.

PR #721, opened 2017-01-15
macOS support for font loading

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/721)


jrmuizel:
Looks like font_loader use fontconfig on macOS. That's no good.

MSleepyPanda:
That's a known [issue](https://github.com/MSleepyPanda/rust-font-loader/issues/1). I don't own a mac device and therefore can't provide a proper implementation.

PR #722, opened 2017-01-15
Publish a frame explicitly with RenderApi::generate_frame()

### LINKED ISSUE #702: Change API to only publish a frame explicitly (via generate_frame).

excaliburHisSheath:
Remove the frame publishing code from the message handling for `ApiMsg::SetRootDisplayList` and `ApiMsg::SetRootPipeline`. Instead, `ApiMsg::GenerateFrame` now builds the scene and `RenderApi::set_root_display_list()` and `set_root_pipeline()` now send a `GenerateFrame` message.

This change seems pretty straightforward, but I have a few things I'm not sure about:

- I moved the call to `build_scene()` from `ApiMsg::SetRootDisplayList` to `ApiMsg::GenerateFrame`, but `ApiMsg::SetRootPipeline` wasn't calling `build_scene()` so I'm not sure that was the right thing to do.
- I added a call to `generate_frame()` to `RenderApi::set_root_display_list()` and `set_root_pipeline()`, keeping the public API the same. I'm not sure if this was the desired change, I would understand if the goal was to have code using WebRender explicitly call `generate_frame()` after calling `set_root_display_list()` and/or `set_root_pipeline()`, but I'll wait for feedback before making that change.

I tested these changes with with the wr-sample project and everything seemed kosher, but let me know if there other regression tests I should run.

As an aside, is there any reading about WebRender's design/architecture that I could look at? I don't fully understand the rational behind this change or the context in which it was requested, but I figure it'd make more sense if I knew more about how WebRender works.

closes #702

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/722)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -202,9 +202,16 @@ impl RenderBackend {
                                                              background_color,
                                                              viewport_size,
                                                              auxiliary_lists);
+                            self.build_scene();
                         }
                         ApiMsg::SetRootPipeline(pipeline_id) => {
                             self.scene.set_root_pipeline_id(pipeline_id);
+
+                            if self.scene.display_lists.get(&pipeline_id).is_none() {
+                                continue;
+                            }
+
+                            self.build_scene();
                         }
                         ApiMsg::Scroll(delta, cursor, move_phase) => {
                             let frame = profile_counters.total_time.profile(|| {
@@ -321,7 +328,6 @@ impl RenderBackend {
                         }
                         ApiMsg::GenerateFrame => {
                             let frame = profile_counters.total_time.profile(|| {
-                                self.build_scene();
                                 self.render()
                             });
                             if self.scene.root_pipeline_id.is_some() {
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -205,10 +205,6 @@ impl RenderBackend {
                         }
                         ApiMsg::SetRootPipeline(pipeline_id) => {
                             self.scene.set_root_pipeline_id(pipeline_id);
-
-                            if self.scene.display_lists.get(&pipeline_id).is_none() {
-                                continue;
-                            }
                         }
                         ApiMsg::Scroll(delta, cursor, move_phase) => {
                             let frame = profile_counters.total_time.profile(|| {
@@ -497,4 +493,3 @@ impl GLContextDispatcher for WebRenderGLDispatcher {
         dispatcher.as_mut().unwrap().as_mut().unwrap().dispatch(f);
     }
 }
-
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -196,22 +196,12 @@ impl RenderBackend {
                                 AuxiliaryLists::from_data(auxiliary_lists_data,
                                                           auxiliary_lists_descriptor);
 
-                            let frame = profile_counters.total_time.profile(|| {
-                                self.scene.set_root_display_list(pipeline_id,
-                                                                 epoch,
-                                                                 built_display_list,
-                                                                 background_color,
-                                                                 viewport_size,
-                                                                 auxiliary_lists);
-
-                                self.build_scene();
-                                self.render()
-                            });
-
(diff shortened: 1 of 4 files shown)


excaliburHisSheath:
Looks like the reftest is failing. I'm looking at it now to try to figure out what changed, but I'd appreciate any advice or info anyone can offer since I'm not terribly familiar with webrender or wrench.

excaliburHisSheath:
The reftests were failing because the `set_root_pipeline()` call in `Wrench::new()` would generate a frame which would be caught by the reftest `Notifier`. This would cause the first render checked by the reftest to come out blank, causing the test to fail even though webrender was in fact producing the correct image.

My solution to this is to change `set_root_pipeline()` and `set_root_display_list()` to not automatically generate a frame. Instead, the client must call `generate_frame()` after setting the display list and root pipeline. I've updated wrench and wr-sample, now the reftests pass and the sample still runs.

Unfortunately, this is a breaking change. Client code will have to add a call to `generate_frame()` after calling `set_root_pipeline()` and `set_root_display_list()`. While this seems reasonable to me, I don't know if it matches the intended design of webrender.

jrmuizel:
> While this seems reasonable to me, I don't know if it matches the intended design of webrender.

Yes, this is the intent for this issue.

excaliburHisSheath:
@jrmuizel so it's fine that this is a breaking change?

jrmuizel:
Yes.

webrender/src/render_backend.rs
```diff
@@ -196,38 +196,15 @@ impl RenderBackend {
...
 
-                            let frame = profile_counters.total_time.profile(|| {
-                                self.scene.set_root_display_list(pipeline_id,
-                                                                 epoch,
-                                                                 built_display_list,
-                                                                 background_color,
-                                                                 viewport_size,
-                                                                 auxiliary_lists);
-
-                                self.build_scene();
```
glennw: We should still do the build_scene() here, but not the render().

webrender/src/render_backend.rs
```diff
@@ -196,38 +196,15 @@ impl RenderBackend {
...
                         }
                         ApiMsg::SetRootPipeline(pipeline_id) => {
                             self.scene.set_root_pipeline_id(pipeline_id);
-
-                            if self.scene.display_lists.get(&pipeline_id).is_none() {
-                                continue;
-                            }
-
-                            let frame = profile_counters.total_time.profile(|| {
-                                self.build_scene();
```
glennw: Same here - let's keep the build_scene() but not the render().

webrender/src/render_backend.rs
```diff
@@ -344,6 +321,7 @@ impl RenderBackend {
                         }
                         ApiMsg::GenerateFrame => {
                             let frame = profile_counters.total_time.profile(|| {
+                                self.build_scene();
```
glennw: We should be able to remove the build_scene() from here, once it's re-added to the other API messages.

excaliburHisSheath:
@glennw I've made the changes you requested, looks like all tests still pass.

glennw:
@excaliburHisSheath Thanks! Looking over the code, this change does change / break the semantics of the backend profile counters (through no fault of yours!). Specifically, the time spent in build_scene() will no longer be accounted for since the profile counters cannot be updated currently unless publishing a frame. I'm fine with merging like this though, and I'll modify how the profile counters work as a follow up.

@vvuk @jrmuizel Do these API changes look correct to you now? Does the replay application also need to be updated?

Once the above questions are answered, this should be good to merge.

excaliburHisSheath:
@vvuk @jrmuizel Have you had a chance to look this over? I don't want this to start rotting if left for too long.

vvuk:
Looks entirely reasonable to me, and more like I would expect things to work - frames are only explicitly generated, which gives the consumer of the API more control.

We can probably nuke replay; wrench subsumes it, though perhaps someone wants a simpler replay still around?

glennw:
Opened #745 and #746 to track the follow up work. Thanks!

excaliburHisSheath:
:D Glad I could be of help.

PR #724, opened 2017-01-16
Add support for raw fonts on Windows

vvuk:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/724)


glennw:
@vvuk Huh, that's a strange travis failure :/ It looks like a genuine bug in wrench, but probably unrelated to this PR?

PR #726, opened 2017-01-16
Add a reader for LayoutTransform.

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/726)

### DIFF
--- a/wrench/src/yaml_helper.rs
+++ b/wrench/src/yaml_helper.rs
@@ -145,6 +145,15 @@ impl YamlHelper for Yaml {
     }
 
     fn as_matrix4d(&self) -> Option<LayoutTransform> {
+        if let Some(mut nums) = self.as_vec_f32() {
+            if nums.len() != 16 {
+                panic!("expected 16 floats, got '{:?}'", self);
+            }
+            return Some(LayoutTransform::row_major(nums[0], nums[1], nums[2], nums[3],
+                                                   nums[4], nums[5], nums[6], nums[7],
+                                                   nums[8], nums[9], nums[10], nums[11],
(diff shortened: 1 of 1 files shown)


wrench/src/yaml_helper.rs
```diff
@@ -145,6 +145,15 @@ impl YamlHelper for Yaml {
     }
 
     fn as_matrix4d(&self) -> Option<LayoutTransform> {
+        if let Some(mut nums) = self.as_vec_f32() {
```
glennw: Could be ```self.as_vec_f32().map()```

wrench/src/yaml_helper.rs
```diff
@@ -145,6 +145,15 @@ impl YamlHelper for Yaml {
...
+            if nums.len() != 16 {
```
glennw: This could be assert!()

jrmuizel: The existing style is sort of if condition panic! so I kept it that way.

wrench/src/yaml_helper.rs
jrmuizel: I'd like to keep this with the if because I want to add support for short hand like "rotate(45deg)" or "translate(1, 2)"

glennw: It's probably still possible to use map() to do this, but we can re-visit when the code gets updated.

wrench/src/yaml_helper.rs
glennw: Added https://github.com/servo/webrender/issues/727

PR #729, opened 2017-01-17
Allow nested clips to be shared between primitives where possible.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/729)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -192,12 +192,13 @@ impl AlphaBatchHelpers for PrimitiveStore {
         let layer_index = layer_index.0 as i32;
         let global_prim_id = prim_index.0 as i32;
         let prim_address = metadata.gpu_prim_index;
-        let clip_task_key = RenderTaskKey::CacheMask(MaskCacheKey::Primitive(prim_index));
-        let clip_task_index = if metadata.clip_task.is_some() {
-            let cache_task_id = RenderTaskId::Dynamic(clip_task_key);
-            render_tasks.get_task_index(&cache_task_id, child_pass_index)
-        } else {
-            OPAQUE_TASK_INDEX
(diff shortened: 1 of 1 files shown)


webrender/src/tiling.rs
```diff
@@ -2618,9 +2619,18 @@ impl FrameBuilder {
                                 }
 
                                 // Try to create a mask if we may need to.
-                                if prim_clip_info.is_some() || !clip_info_stack.is_empty() {
+                                if !clip_info_stack.is_empty() {
```
kvark: Why are we no longer checking for `prim_clip_info` here?

webrender/src/tiling.rs
```diff
@@ -192,12 +192,13 @@ impl AlphaBatchHelpers for PrimitiveStore {
         let layer_index = layer_index.0 as i32;
         let global_prim_id = prim_index.0 as i32;
         let prim_address = metadata.gpu_prim_index;
-        let clip_task_key = RenderTaskKey::CacheMask(MaskCacheKey::Primitive(prim_index));
-        let clip_task_index = if metadata.clip_task.is_some() {
-            let cache_task_id = RenderTaskId::Dynamic(clip_task_key);
-            render_tasks.get_task_index(&cache_task_id, child_pass_index)
-        } else {
-            OPAQUE_TASK_INDEX
+        let clip_task_index = match metadata.clip_task {
```
kvark: nice refactor!

webrender/src/tiling.rs
glennw: We don't need to, as the line above pushes the primitive clip item on to the clip stack, if applicable.

kvark: Uh oh, you are right.

Issue #728, opened 2017-01-17
Fix texture update stalls in WR


glennw:
When running with INTEL_DEBUG=perf there are a number of performance warnings about CPU being blocked the driver due to updating resources. We should identify the cause of these and fix them. These will explain at least some of the spikes we sometimes see in the GPU profiler.

kvark:
Is this still a problem?

glennw:
There are other stalls present, yes. I'm not sure if they are textures or something else.

glennw:
I don't see these problems any more.

PR #731, opened 2017-01-17
Add a (temporary) workaround for driver stalls when updating vertex textures.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/731)


glennw:
Related to #728.

PR #733, opened 2017-01-17
Add a helper for parsing functions and use it for transform shortcuts

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/733)

### DIFF
--- a/wrench/src/parse_function.rs
+++ b/wrench/src/parse_function.rs
@@ -1,0 +1,86 @@
+use std::str::CharIndices;
+// A crapy parser for parsing strings like "translate(1, 3)"
+
+pub fn parse_function(s: &str) -> (&str, Vec<&str>) {
+    // XXX: This it not particular easy to read. Sorry.
+    struct Parser<'a> {
+        itr: CharIndices<'a>,
+        start: usize,
+        o: Option<(usize, char)>,
+    }
+    impl<'a> Parser<'a> {
+        fn skip_whitespace(&mut self) {
+            while let Some(k) = self.o {
+                if !k.1.is_whitespace() {
+                    break;
+                }
+                self.start = k.0 + 1;
+                self.o = self.itr.next();
+            }
+
+        }
+    }
+    let mut c = s.char_indices();
+    let o = c.next();
+    let mut p = Parser{itr: c, start: 0, o: o};
+
+    p.skip_whitespace();
+
+    let mut end = p.start;
+    while let Some(k) = p.o {
+        if !k.1.is_alphabetic() {
+            break;
+        }
+        end = k.0 + 1;
+        p.o = p.itr.next();
+    }
+
+    let name = &s[p.start..end];
+    let mut args = Vec::new();
+
+    p.skip_whitespace();
+
+    if let Some(k) = p.o {
+        if !(k.1 == '(') {
+            return (name, args);
+        }
+        p.start = k.0 + 1;
+        p.o = p.itr.next();
+    }
+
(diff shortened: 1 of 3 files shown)


wrench/src/yaml_helper.rs
```diff
@@ -168,6 +169,17 @@ impl YamlHelper for Yaml {
                                                    nums[8], nums[9], nums[10], nums[11],
                                                    nums[12], nums[13], nums[14], nums[15]))
         }
+        if let Some(s) = self.as_str() {
+            let (name, args) = parse_function(s);
+            match name {
```
kvark: nit: would be simpler to just do `match parse_function(s)`

wrench/src/parse_function.rs
```diff
@@ -0,0 +1,90 @@
+use std::str::CharIndices;
+// A crapy parser for parsing strings like "translate(1, 3)"
+
+pub fn parse_function(s: &str) -> (&str, Vec<&str>) {
+    // XXX: This it not particular easy to read. Sorry.
+    struct Parser<'a> {
+        itr: CharIndices<'a>,
+        start: usize,
+        o: Option<<CharIndices<'a> as Iterator>::Item>,
```
kvark: perhaps, `Option<(usize, char)>` would be simpler?

wrench/src/parse_function.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+    impl<'a> Parser<'a> {
+        fn skip_whitespace(&mut self) {
+            while let Some(k) = self.o {
```
kvark: This pattern seems to be used about 5 times in the code, so it's begging for a helper function receiving some sort of a predicate. I could also see `self.o` removed entirely if you used a peekable iterator. Actually, the `self.start` would go away as well ;) You'd only need to implement something like a `peek_while` for as a helper.

wrench/src/parse_function.rs
```diff
@@ -0,0 +1,90 @@
...
+        args.push(&s[p.start..end]);
+
+        p.skip_whitespace();
+
+        if let Some(k) = p.o {
+            if k.1 == ')' {
+                break;
+            }
+        }
+        if let Some(k) = p.o {
```
kvark: merge this `if` with the one above?

wrench/src/yaml_helper.rs
jrmuizel: What would the match cases look like if doing that?

wrench/src/parse_function.rs
jrmuizel: I switched to this.

wrench/src/parse_function.rs
jrmuizel: I just eliminated one of the if statements completely.

wrench/src/parse_function.rs
jrmuizel: I'll still look into this but it would be nice to get merge this functionality now as I have reftests that depend on it and I've already spent more time on this function than I want to.

I didn't know about peekable iterators so I'll give them a look.

wrench/src/yaml_helper.rs
kvark:
```rust
match parse_function(s) {
  ("translate", args) if args.len() == 2 => {...},
  (name, _) => panic!("unknown function {}", name),
}
```

kvark:
Refactor to come as a follow up.
Thanks @jrmuizel !
@bors-servo r+

PR #734, opened 2017-01-17
add support for radial gradients

lsalzman:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/734)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -31,6 +31,7 @@
 #define UV_PIXEL         uint(1)
 
 #define MAX_STOPS_PER_ANGLE_GRADIENT 8
+#define MAX_STOPS_PER_RADIAL_GRADIENT 8
 
 uniform sampler2DArray sCache;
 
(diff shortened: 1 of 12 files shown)


webrender/res/prim_shared.glsl
```diff
@@ -31,6 +31,7 @@
 #define UV_PIXEL         uint(1)
 
 #define MAX_STOPS_PER_ANGLE_GRADIENT 8
+#define MAX_STOPS_PER_RADIAL_GRADIANT 8
```
glennw: nit: gradiant -> gradient

webrender/src/renderer.rs
```diff
@@ -473,6 +475,10 @@ impl Renderer {
                                                      &mut device,
                                                      &[],
                                                      options.precache_shaders);
+        let ps_radial_gradient = PrimitiveShader::new("ps_radial_gradient",
+                                                     &mut device,
```
glennw: nit: indentation not quite right

webrender/res/ps_radial_gradient.fs.glsl
```diff
@@ -0,0 +1,67 @@
...
+        if (vStartRadius + rd * t0 >= 0.0) {
+            x = t0;
+        } else if (vStartRadius + rd * t1 >= 0.0) {
+            x = t1;
+        } else {
+            discard;
+        }
+    }
+
+    oFragColor = mix(vColors[0],
```
kvark: Can the radial gradient be masked? I think we are missing a mask fetch here.

lsalzman:
Based the shader off the angle gradient shader, which didn't do it either.
So presumably if radial gradients are missing that, then so are angle
gradients...

PR #736, opened 2017-01-18
Fix missing Multiply case in ps_composite.fs.glsl

rlhunt:
It looks like the Multiply case for a composite operation was missing. This would cause mix-blend-mode: multiply to yield yellow pixels. I've included a reftest for this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/736)

### DIFF
--- a/wrench/src/yaml_helper.rs
+++ b/wrench/src/yaml_helper.rs
@@ -25,6 +25,7 @@ pub trait YamlHelper {
     fn as_pt_to_au(&self) -> Option<Au>;
     fn as_vec_string(&self) -> Option<Vec<String>>;
     fn as_border_radius(&self) -> Option<BorderRadius>;
+    fn as_mix_blend_mode(&self) -> Option<MixBlendMode>;
 }
 
 fn string_to_color(color: &str) -> Option<ColorF> {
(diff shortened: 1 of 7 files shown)


wrench/src/reftest.rs
```diff
@@ -22,10 +22,28 @@ pub struct Reftest<'a> {
     reference: &'a Path,
 }
 
-pub fn parse_reftests<F>(filename: &str, mut runner: F)
+pub fn parse_reftests_recursive<F>(directory: &Path, runner: &mut F)
+    where F: FnMut(Reftest)
+{
```
glennw: nit: brace should be on previous line

wrench/src/reftest.rs
```diff
@@ -22,10 +22,28 @@ pub struct Reftest<'a> {
...
+    let manifest = directory.join("reftest.list");
+
+    if manifest.is_file()
+    {
```
glennw: nit: brace should be on previous line

wrench/src/reftest.rs
```diff
@@ -22,10 +22,28 @@ pub struct Reftest<'a> {
...
+        parse_reftest_manifest(manifest.as_path(), runner);
+    }
+
+    for entry in directory.read_dir().unwrap().map(|x| x.unwrap())
+    {
```
glennw: nit: brace should be on previous line

wrench/src/reftest.rs
```diff
@@ -47,6 +65,18 @@ pub fn parse_reftests<F>(filename: &str, mut runner: F)
         };
         let test = dir.join(items.next().unwrap());
         let reference = dir.join(items.next().unwrap());
+
+        if !test.is_file()
```
glennw: Perhaps these should return failure / panic, rather than continue?

wrench/src/reftest.rs
emilio: I think the common rust style is brace on a new line for where clauses.

glennw: Huh, didn't know that. Ok!

wrench/src/reftest.rs
```diff
@@ -47,6 +62,9 @@ pub fn parse_reftests<F>(filename: &str, mut runner: F)
         };
         let test = dir.join(items.next().unwrap());
         let reference = dir.join(items.next().unwrap());
+
+        assert!(test.is_file() && reference.is_file());
```
rlhunt: Reftest parsing is very panicky right now. This at the least creates a better panic for a missing file than before. Eventually it would be nice to have error messages.

rlhunt:
I don't have a strong opinion on this. I'll remove the reftest parsing commits and add the reftest to the common directory for now. I don't want to expand this pull request anymore.

wrench/src/yaml_helper.rs
```diff
@@ -239,4 +262,11 @@ impl YamlHelper for Yaml {
             }
         }
     }
+
+    fn as_mix_blend_mode(&self) -> Option<MixBlendMode> {
+        return match self.as_str() {
```
glennw: This could be self.as_str().map(...) rather than a match {}

rlhunt: Huh, didn't know that Option had helper methods. That's nice!

Map wasn't working for me because string_to_mix_blend_mode returns an Option also. But it looks like and_then will do the job.

wrench/src/yaml_helper.rs
```diff
@@ -47,6 +48,28 @@ fn string_to_color(color: &str) -> Option<ColorF> {
     }
 }
 
+fn string_to_mix_blend_mode(mix_blend_mode: &str) -> Option<MixBlendMode> {
```
kvark: I suggest we make a constant list of pairs here, and then just walk it up back and forth for conversion between strings and blend modes. That would avoid repetition (DRY FTW).

rlhunt:
I rewrote the mix-blend-mode conversion functions to use a macro. The advantages of this are that it is reusable and it will catch if a new MixBlendMode is added but not handled. At the expense of potentially being harder to understand. If that's a problem it can be rewritten using constants.

glennw:
Looks good, thanks!

PR #748, opened 2017-01-19
Pipe text color information to rasterize_glyphs

changm:
@kvark r? Required to do the gamma look up table stuff. The unfortunate side effect of this is that the Glyph keys require ColorU as u8 rgb channels instead of floats since f32 doesn't implement std::eq :/. Also, all the other API calls expects color information as ColorF. Instead of making users of the api give text as ColorU, we just translate it into/out of ColorU when we create the glyph key. I'd be happy if you become picky again and find a better way :)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/748)

### DIFF
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -191,6 +191,55 @@ pub struct ColorF {
 }
 known_heap_size!(0, ColorF);
 
+#[derive(Clone, Copy, Hash, Eq, Debug, Deserialize, PartialEq, PartialOrd, Ord, Serialize)]
+pub struct ColorU {
+    pub r: u8,
+    pub g: u8,
+    pub b: u8,
+    pub a: u8,
+}
+
+impl From<ColorF> for ColorU {
+    fn from(color: ColorF) -> ColorU {
+        ColorU {
+            r: ColorU::round_to_int(color.r),
+            g: ColorU::round_to_int(color.g),
+            b: ColorU::round_to_int(color.b),
+            a: ColorU::round_to_int(color.a),
+        }
+    }
+}
+
+impl Into<ColorF> for ColorU {
+    fn into(self) -> ColorF {
+        ColorF {
+            r: self.r as f32 / 255.0,
+            g: self.g as f32 / 255.0,
+            b: self.b as f32 / 255.0,
+            a: self.a as f32 / 255.0,
+        }
+    }
+}
+
+impl ColorU {
+    fn round_to_int(x: f32) -> u8 {
+        debug_assert!((0.0 <= x) && (x <= 1.0));
+        let f = (255.0 * x) + 0.5;
+        let val = f.floor();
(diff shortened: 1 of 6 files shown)


glennw:
@changm Looks like there are compile errors on travis / appveyor.

emilio:
I think you should be able to implement `Eq` in `ColorF` if you don't care about `Inf`/`Nan`

glennw:
@changm @emilio You could implement Eq, but quantizing is probably a good idea. That way, we won't end up with extra keys / glyphs being rasterized into the cache due to very slight floating point differences.

changm:
@glennw @kvark hoorah! Finally can merge r?

webrender_traits/src/types.rs
```diff
@@ -191,6 +191,53 @@ pub struct ColorF {
...
+    fn round_to_int(x: f32) -> u8 {
+        assert!((0.0 <= x) && (x <= 1.0));
+        let f = (255.0 * x) + 0.5;
+        let val = f.floor();
+        assert!(val <= 255.0);
+        val as u8
+    }
+
+    #[allow(non_snake_case)]
+    pub fn from_colorF(color: ColorF) -> ColorU {
```
kvark: Typically, in Rust this is done by implementing the `From` and `Into` traits:
```rust
impl From<ColorF> for ColorU {}
impl Into<ColorF> for ColorU {}
```

webrender_traits/src/types.rs
```diff
@@ -191,6 +191,53 @@ pub struct ColorF {
 }
 known_heap_size!(0, ColorF);
 
+#[derive(Clone, Copy, Hash, Eq, Debug, Deserialize, PartialEq, PartialOrd, Ord, Serialize)]
+pub struct ColorU {
```
kvark: It feels slightly wrong that both `ColorU` and `ColorF` are exposed in webrender_traits. Shouldn't one of them be just an implementation detail (hidden inside `webrender`)?

changm: We need it for GlyphKey :(. Otherwise yes, I would've just passed ColorF around until we were inside webrender.

webrender_traits/src/types.rs
```diff
@@ -191,6 +191,55 @@ pub struct ColorF {
...
+            g: self.g as f32 / 255.0,
+            b: self.b as f32 / 255.0,
+            a: self.a as f32 / 255.0,
+        }
+    }
+}
+
+impl ColorU {
+    fn round_to_int(x: f32) -> u8 {
+        assert!((0.0 <= x) && (x <= 1.0));
```
emilio: You may want to make these `debug_assert!` instead (assertions run in release by default in rust).

kvark:
Thanks @changm !
Looks like everything is addressed, merging.
@bors-servo r+

PR #754, opened 2017-01-19
Move all layer tree operations to a ScrollTree struct

mrobinson:
This will make it easier to introduce the idea of reference frames
into WebRender. This should not change behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/754)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -3,32 +3,23 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use app_units::Au;
-use euclid::Point3D;
 use fnv::FnvHasher;
-use geometry::ray_intersects_rect;
 use internal_types::{ANGLE_FLOAT_TO_FIXED, AxisDirection};
 use internal_types::{CompositionOp};
 use internal_types::{LowLevelFilterOp};
 use internal_types::{RendererFrame};
-use layer::{Layer, ScrollingState};
+use layer::Layer;
 use resource_cache::ResourceCache;
 use scene::Scene;
-use std::collections::{HashMap, HashSet};
+use scroll_tree::{ScrollTree, ScrollStates};
+use std::collections::HashMap;
 use std::hash::BuildHasherDefault;
-use tiling::{AuxiliaryListsMap, FrameBuilder, FrameBuilderConfig, LayerMap, PrimitiveFlags};
-use webrender_traits::{AuxiliaryLists, PipelineId, Epoch, ScrollPolicy, ScrollLayerId};
-use webrender_traits::{ClipRegion, ColorF, DisplayItem, StackingContext, FilterOp, MixBlendMode};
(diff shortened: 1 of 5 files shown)


glennw:
This looks good, and I've tested scrolling on a number of different sites.

mrobinson:
Thanks!

webrender/src/frame.rs
```diff
@@ -334,172 +231,11 @@ impl Frame {
...
-                layer.stretch_overscroll_spring();
-            }
-
-            scrolled_a_layer = scrolled_a_layer ||
-                layer.scrolling.offset != original_layer_scroll_offset ||
-                layer.scrolling.started_bouncing_back;
-        }
-
-        scrolled_a_layer
+        self.scroll_tree.scroll(scroll_location, cursor, phase,)
```
kvark: extra comma?

webrender/src/frame.rs
```diff
@@ -746,11 +460,11 @@ impl Frame {
 
         if level == 0 && self.frame_builder_config.enable_scrollbars {
             let scrollbar_rect = LayerRect::new(LayerPoint::zero(), LayerSize::new(10.0, 70.0));
-            context.builder.add_solid_rectangle(&scrollbar_rect,
-                                                &ClipRegion::simple(&scrollbar_rect),
-                                                &DEFAULT_SCROLLBAR_COLOR,
-                                                PrimitiveFlags::Scrollbar(self.root_scroll_layer_id.unwrap(),
-                                                                          4.0));
+            context.builder.add_solid_rectangle(
```
kvark: isn't the new formatting (see below) against the style guidelines?

webrender/src/scroll_tree.rs
```diff
@@ -0,0 +1,424 @@
...
+            pending_scroll_offsets: HashMap::new(),
+            current_scroll_layer_id: None,
+            root_scroll_layer_id: None,
+        }
+    }
+
+    pub fn collect_layers_bouncing_back(&self)
+                                        -> HashSet<ScrollLayerId, BuildHasherDefault<FnvHasher>> {
+        let mut layers_bouncing_back = HashSet::with_hasher(Default::default());
+        for (scroll_layer_id, layer) in self.layers.iter() {
```
kvark: nit (unrelated to the PR): we could do something fancy like
```rust
self.layers.iter()
           .filter(|(_, layer)| layer.scrolling.bouncing_back)
           .map(|(id, _)| *id)
           .collect()
```

webrender/src/frame.rs
mrobinson: I'm not sure, but the PrimitiveFlags::Scrollbar(self.root_scroll_layer_id.unwrap(), line wouldn't fit within the line length limit any other way.

webrender/src/frame.rs
mrobinson: Yes! I will fix this in a followup.

webrender/src/frame.rs
kvark: ok, I don't know what the guidelines are for this case then

PR #755, opened 2017-01-19
Ensure that shared clip masks don't intersect with the primitive bounding rect.

### LINKED ISSUE #752: Text with multiple runs triggers assert
I'm running into an issue where trying rendering text with more than 8 glyphs causes [this debug assert][debug assert] to trigger. I've traced the underlying issue to the changes introduced in #729 by @glennw, though I can't tell if those changes are actually the source of the bug. What seems to be happening is that text primitives with more than 8 glyphs get broken up into multiple runs. These runs don't have their own clip info so their render tasks all get keyed with the stacking context, but they all have different sizes which causes them to trigger the assert.

If I'm understanding the situation correctly, the bug goes as follows:

- The user pushes some text with more than 8 glyphs.
- Later, when the frame is created, `FrameBuilder::add_text()` is called for that text. `add_text()` is hardcoded to break text up into runs of at most 8 glyphs, so the user's text is broken into 2 or more runs, each of which is passed to `FrameBuilder::add_primitive()`.
- `add_primitive()` creates a `PrimitiveRunCmd::PrimitiveRun` for each glyph run.
- Later, `FrameBuilder::cull_layers()` processes the primitive run commands. If the run command's primitive metadata (created when the primitive was added to the frame builder) has no clip info (which the glyph runs don't seem to) their render task is keyed with the stacking context index. Each run in the original text is therefore given the same key because all runs in the same text will have the same stacking context.
- Later, `RenderPass::build()` will try to reuse a previous dynamic allocation, but will fail the assert. This is because all the 8 glyph runs in the user's text retrieve the same cached dynamic allocation, but each render pass has a different size.

I'd like to submit a PR to fix this, but my inexperience with webrender makes it hard for me to tell what the correct solution is. Does anyone know more about what's going on or what the cause of the bug might be?

[debug assert]: https://github.com/servo/webrender/blob/69f5804414865a0143def46564b2aadee2ad08e8/webrender/src/tiling.rs#L1036

glennw:
Fixes #752.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/755)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -2670,11 +2670,16 @@ impl FrameBuilder {
                                     // that two primitives which are only clipped by the
                                     // stacking context stack can share clip masks during
                                     // render task assignment to targets.
-                                    let mask_key = match prim_clip_info {
-                                        Some(..) => MaskCacheKey::Primitive(prim_index),
-                                        None => MaskCacheKey::StackingContext(*sc_index),
+                                    let (mask_key, mask_rect) = match prim_clip_info {
+                                        Some(..) => {
+                                            (MaskCacheKey::Primitive(prim_index), prim_bounding_rect)
+                                        }
+                                        None => {
+                                            let layer_rect = layer.xf_rect.as_ref().unwrap().bounding_rect;
+                                            (MaskCacheKey::StackingContext(*sc_index), layer_rect)
+                                        }
                                     };
-                                    let mask_opt = RenderTask::new_mask(prim_bounding_rect,
+                                    let mask_opt = RenderTask::new_mask(mask_rect,
                                                                         mask_key,
                                                                         &clip_info_stack,
                                                                         &self.layer_store);


webrender/src/tiling.rs
```diff
@@ -2670,11 +2670,16 @@ impl FrameBuilder {
...
                                     // render task assignment to targets.
-                                    let mask_key = match prim_clip_info {
-                                        Some(..) => MaskCacheKey::Primitive(prim_index),
-                                        None => MaskCacheKey::StackingContext(*sc_index),
+                                    let (mask_key, mask_rect) = match prim_clip_info {
+                                        Some(..) => {
+                                            (MaskCacheKey::Primitive(prim_index), prim_bounding_rect)
+                                        }
+                                        None => {
+                                            let layer_rect = layer.xf_rect.as_ref().unwrap().bounding_rect;
```
kvark: I suppose the `unwrap` here is fine since we already check for `layer.is_visible` at the top

PR #756, opened 2017-01-19
Add ImageDescriptor to describe the layout and format of images in memory.

nical:
The most notable change is that this moves the responsibility of specifying whether images are opaque to the consumer of the webrender api rather than webrender itself, which should be more efficient, and lets us handle opaque RGBA surfaces efficiently.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/756)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -39,6 +39,7 @@ use util::TransformedRectKind;
 use webrender_traits::{ColorF, Epoch, PipelineId, RenderNotifier, RenderDispatcher};
 use webrender_traits::{ExternalImageId, ImageFormat, RenderApiSender, RendererKind};
 use webrender_traits::{DeviceIntRect, DevicePoint, DeviceIntPoint, DeviceIntSize, DeviceUintSize};
+use webrender_traits::ImageDescriptor;
 use webrender_traits::channel;
 use webrender_traits::VRCompositorHandler;
 
@@ -551,19 +552,25 @@ impl Renderer {
         // TODO: Ensure that the white texture can never get evicted when the cache supports LRU eviction!
         let white_image_id = texture_cache.new_item_id();
         texture_cache.insert(white_image_id,
-                             2,
-                             2,
-                             None,
-                             ImageFormat::RGBA8,
+                             ImageDescriptor {
+                                width: 2,
+                                height: 2,
+                                stride: None,
+                                format: ImageFormat::RGBA8,
+                                is_opaque: false,
+                             },
                              TextureFilter::Linear,
                              Arc::new(white_pixels));
 
         let dummy_mask_image_id = texture_cache.new_item_id();
         texture_cache.insert(dummy_mask_image_id,
-                             2,
-                             2,
-                             None,
-                             ImageFormat::A8,
+                             ImageDescriptor {
+                                width: 2,
+                                height: 2,
+                                stride: None,
+                                format: ImageFormat::A8,
+                                is_opaque: true,
+                             },
                              TextureFilter::Linear,
                              Arc::new(mask_pixels));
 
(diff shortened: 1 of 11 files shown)


webrender/src/renderer.rs
```diff
@@ -551,19 +552,25 @@ impl Renderer {
...
-                             2,
-                             2,
-                             None,
-                             ImageFormat::A8,
+                             ImageDescriptor {
+                                width: 2,
+                                height: 2,
+                                stride: None,
+                                format: ImageFormat::A8,
+                                is_opaque: true,
```
kvark: I don't think this one is opaque

wrench/src/json_frame_writer.rs
```diff
@@ -203,15 +203,14 @@ impl webrender::ApiRecordingReceiver for JsonFrameWriter {
                 self.fonts.insert(*key, CachedFont::Native(native_font_handle.clone()));
             }
 
-            &ApiMsg::AddImage(ref key, width, height, stride,
-                              format, ref data) => {
-                let stride = if let Some(stride) = stride {
+            &ApiMsg::AddImage(ref key, ref descriptor, ref data) => {
+                let stride = if let Some(stride) = descriptor.stride {
```
kvark: nit: could be rewritten via `unwrap_or`

wrench/src/json_frame_writer.rs
```diff
@@ -203,15 +203,14 @@ impl webrender::ApiRecordingReceiver for JsonFrameWriter {
...
                     stride
                 } else {
-                    match format {
-                        ImageFormat::A8 => width,
-                        ImageFormat::RGBA8 | ImageFormat::RGB8 => width*4,
-                        ImageFormat::RGBAF32 => width*16,
+                    match descriptor.format {
+                        ImageFormat::A8 => descriptor.width,
```
kvark: don't we have just a `ImageFormat::bytes_per_pixel()`? we should

wrench/src/json_frame_writer.rs
```diff
@@ -220,19 +219,20 @@ impl webrender::ApiRecordingReceiver for JsonFrameWriter {
                     &ImageData::External(_) => { return; }
                 };
                 self.images.insert(*key, CachedImage {
-                    width: width, height: height, stride: stride,
-                    format: format,
+                    width: descriptor.width,
```
kvark: Any reason we don't want to store the descriptor directly in `CachedImage`?

wrench/src/yaml_frame_writer.rs
```diff
@@ -647,15 +647,14 @@ impl webrender::ApiRecordingReceiver for YamlFrameWriterReceiver {
...
+            &ApiMsg::AddImage(ref key, ref descriptor, ref data) => {
+                let stride = if let Some(stride) = descriptor.stride {
                     stride
                 } else {
-                    match format {
-                        ImageFormat::A8 => width,
-                        ImageFormat::RGBA8 | ImageFormat::RGB8 => width*4,
-                        ImageFormat::RGBAF32 => width*16,
+                    match descriptor.format {
+                        ImageFormat::A8 => descriptor.width,
```
kvark: duplicated code, one more reason to add `ImageFormat::bytes_per_pixel()`

webrender_traits/src/types.rs
```diff
@@ -191,6 +191,39 @@ pub struct ColorF {
...
+    pub is_opaque: bool,
+}
+
+// TODO(gw): If this ever shows up in profiles, consider calculating
+// this lazily on demand, possibly via the resource cache thread.
+// It can probably be made a lot faster with SIMD too!
+// This assumes that A8 textures are never opaque, since they are
+// typically used for alpha masks. We could revisit that if it
+// ever becomes an issue in real world usage.
+pub fn is_image_opaque(format: ImageFormat, bytes: &[u8]) -> bool {
```
kvark: Let's move it to `wrench`, it's not idiomatic to WR any more.

PR #762, opened 2017-01-20
Followups to the ImageDescriptor pull request.

nical:
r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/762)

### DIFF
--- a/wrench/src/json_frame_writer.rs
+++ b/wrench/src/json_frame_writer.rs
@@ -204,16 +204,9 @@ impl webrender::ApiRecordingReceiver for JsonFrameWriter {
             }
 
             &ApiMsg::AddImage(ref key, ref descriptor, ref data) => {
-                let stride = if let Some(stride) = descriptor.stride {
-                    stride
-                } else {
-                    match descriptor.format {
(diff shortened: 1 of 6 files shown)


wrench/src/json_frame_writer.rs
```diff
@@ -204,16 +204,9 @@ impl webrender::ApiRecordingReceiver for JsonFrameWriter {
...
-                    stride
-                } else {
-                    match descriptor.format {
-                        ImageFormat::A8 => descriptor.width,
-                        ImageFormat::RGBA8 | ImageFormat::RGB8 => descriptor.width*4,
-                        ImageFormat::RGBAF32 => descriptor.width*16,
-                        _ => panic!("Invalid image format"),
-                    }
-                };
+                let stride = descriptor.stride.unwrap_or_else(||{
```
kvark: These could just be `unwrap_or`, but I will not lose my sleep over it ;)

kvark:
Thanks!
@bors-servo r+

wrench/src/json_frame_writer.rs
nical: ah yes indeed

PR #763, opened 2017-01-20
Include the clipping rectangle of a stacking context into the clip stack

### LINKED ISSUE #735: rectangle clips don't apply to nested stacking contexts
The following yaml file draws incorrectly. We should not see a 100x100 box.

```
---
root:
  items:
    - 
      type: stacking_context
      bounds: [0, 0, 95, 88]
      clip:
        rect: [9, 9, 10, 10]
      items:
        - 
          type: stacking_context
          bounds: [0, 0, 95, 88]
          clip:
              rect: [0, 0, 100, 100]
          items:
            - 
              bounds: [0, 0, 100, 100]
              type: rect
              color: blue
```

kvark:
Fixes #735 
r? @glennw 
(please don't merge until I confirm that servo tests are good)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/763)


kvark:
@glennw this should be ready for your look now. I wasn't able to confirm that the tests pass on my machine because they time out (with and without my PR) for some reason.

glennw:
@kvark I'll pull this change and run the tests locally here.

glennw:
OK, I see a number of test failures at mask_cache.rs:118 in all the border radius tests.

kvark:
Ready, this time for real!

glennw:
This seems to only handle the clip rect for the most recent stacking context, which doesn't seem quite right? It also introduces (a little) more work during adding of primitives and layers, which *might* not be ideal in the case of a very long page (since the work has to be done for each primitive regardless of visibility).

Would it make sense to introduce a (combined) world space clip rect during cull_layers(), which is handled in the vertex shader in a similar way as the local clip rect? In that way, as soon as we encounter a world space clip rect that is not axis-aligned, we'd add that to the mask for this stacking context / primitive. Thoughts?

kvark:
> This seems to only handle the clip rect for the most recent stacking context, which doesn't seem quite right?

It handles the top clip rectangle on the stack if everything below is axis-aligned. Also, it makes sure to intersect it with the previous ones, so that only the top needs to be checked (and it's considered "combined"):
```rust
Some(&Some(ref rect)) if transform.can_losslessly_transform_a_2d_rect() => {
                let clip_rect = clip_region.main.intersection(rect)
                                           .unwrap_or(LayerRect::zero());
                Some(clip_rect)
            }
```

> Would it make sense to introduce a (combined) world space clip rect during cull_layers(), which is handled in the vertex shader in a similar way as the local clip rect?

Currently with this PR, the transformed layers have an extra clip instance for their clip rectangle:
```rust
        let clip_source = if is_transformed || clip_region.is_complex() {
            ClipSource::Region(clip_region.clone())
```

And since we compute the intersection of all the prior layers in world space in order to figure out the task area of the mask (see below), it's already handled:
```rust
        let result = clips.iter()
                          .fold(Some(actual_rect), |current, clip| {
            current.and_then(|rect| rect.intersection(&clip.1.outer_rect))
        });
```
So basically it seems to me that adding a world clip rect to the vertex shaders would do nothing, since we already make sure that we only work inside that area.

glennw:
Ah, great! I missed that part where it intersects with what's currently on the stack.

My only remaining concern is that we calculate this stuff during push_layer() and add_primitive(). I think this will cause issues once we introduce the OMTA support (which will allow animating the transform matrices without re-creating the display list). Is it possible to make this work during cull_layers() instead? (I guess that might complicate the allocation of clip items?).

kvark:
> OMTA support (which will allow animating the transform matrices without re-creating the display list)

Ah, if that's the case, then I definitely need to change the logic and move it to `cull_layers`. I recall from our scrolling-related discussions that if any transformation changed we expect a new frame to be built. Was this assumption challenged afterwards? I got an impression that our brainstorming relied on it heavily.

glennw:
I *think* when we discussed this in regards to scrolling I was referring to a new frame generated via cull_layers(), rather then a new display list altogether. But I can see how that would be ambiguous.

@mrobinson Is that going to work OK with the scrolling changes you're working on? In the OMTA case, we will know during the initial display list build whether a transform *may* be animated at a later point, which would allow us to know whether to build a new reference frame. Do you foresee any other issues with OMTA + scrolling changes?

mrobinson:
@kvark I think maybe we can eliminate clipping rectangles from stacking contexts entirely, but I'd like to get your opinion. In Servo, the StackingContext clip is always defined like this:

```
                let clip = builder.new_clip_region(&stacking_context.overflow.to_rectf(),                                                                                                                          
                                                   vec![],                                                                                                                                                         
                                                   None);    
```

Scrolling overflow regions always create ScrollLayers in the current display list, so we can add them to the clip stack that way. ```overflow: hidden`` `in Servo doesn't affect the stacking context at all. Instead these clips are applied directly to children. In the future this could be optimized by handling them like ```overflow: scroll``` and ```overflow: auto```. As it seems that StackingContext clips are completely unused, I think we can just ignore them for purposes of clipping. What do you think?

kvark:
@mrobinson thanks for taking a look!

> I think maybe we can eliminate clipping rectangles from stacking contexts entirely
> Scrolling overflow regions always create ScrollLayers in the current display list, so we can add them to the clip stack that way.

Yes, that's the end-game plan. The whole point of this PR is to patch the current behavior so that we can fix a few Gecko/WR ref tests as well as open the door for some cleanup of the old code.

kvark:
I moved the transformed clip rect handling to `MaskCacheInfo::update()` now, so you should be able to change the layer transformations without rebuilding the DLs.
Getting 5 white-space tests failing, unrelated to the PR (happens both with and without my changes).

glennw:
Reviewed 7 of 7 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/763)*

glennw:
The build seems to keep timing out on the mac machine.

You could try extending the allowable time without a log message as mentioned here - https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

It does seem odd for that command to take > 10 mins though - I wonder if it's hanging rustc perhaps?

kvark:
Ugh, it's a pity that the OSX bots get stuck. I wonder if we can make Rustc more verbose to address that.

glennw:
@kvark It's perhaps worth bumping the rustc version in .travis.yml to see if that helps. I wonder if we're hitting a rustc compiler bug with a hang / resource exhaustion?

glennw:
:lgtm:

---

Reviewed 1 of 1 files at r5.
Review status: 5 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/763#-:-KbmNzyks_JnjFuKZq4G:bnfp4nl)*

PR #765, opened 2017-01-23
Move binary recordings to a single file instead of a directory

vvuk:
No reason to use separate files.  In the future, we could even gzip as we stream out (on a different thread maybe). This also cleans up the recording code to just have a single unified path instead of separate detour stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/765)


kvark:
I read through it, looks great!
Just need to fix the CI errors.

glennw:
Looks like there is still a compile error on CI?

vvuk:
Yes - working on it, I think I pushed the wrong branch. Checking on laptop
shortly

glennw:
@kvark OK, The CI errors are resolved, if you have time to re-look at this.

kvark:
Thanks @vvuk, the code is great, and the changes are much welcome!
Please r=me after rebasing.

PR #767, opened 2017-01-23
Move more code to be Layer struct methods

mrobinson:
This change continues the cleanup of ScrollTree/Layer code in
preparation for adding the concept of reference frames to WebRender.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/767)

### DIFF
--- a/webrender/src/scroll_tree.rs
+++ b/webrender/src/scroll_tree.rs
@@ -2,22 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use euclid::Point3D;
 use fnv::FnvHasher;
-use geometry::ray_intersects_rect;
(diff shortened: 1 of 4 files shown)


mrobinson:
@glennw r?

Tested this with servo master with a small patch to properly call ```generate_frame()``` and all tests passed.

webrender/src/scroll_tree.rs
```diff
@@ -61,36 +53,14 @@ impl ScrollTree {
             }
 
             match scroll_layer_id.info {
```
glennw: Perhaps:
```
if scroll_layer_id.info.is_none() {
  return None;
}
```

mrobinson: Thanks. This is quite a bit cleaner. I'll update the PR.

mrobinson:
Thank you both very kindly for the review.

PR #768, opened 2017-01-23
Propagate errors when initializing WebRender instead of panicking

nical:
First step towards proper error handing in WebRender. This patch focuses on making the initialization of the Renderer fallible. This is important in order to let Gecko be able to catch potential errors and fallback to a different rendering backend.

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/768)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -375,8 +375,7 @@ impl Program {
     fn attach_and_bind_shaders(&mut self,
                                vs_id: gl::GLuint,
                                fs_id: gl::GLuint,
-                               vertex_format: VertexFormat,
-                               panic_on_fail: bool) -> bool {
+                               vertex_format: VertexFormat) -> Result<(), ShaderError> {
         gl::attach_shader(self.id, vs_id);
         gl::attach_shader(self.id, fs_id);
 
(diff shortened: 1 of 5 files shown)


webrender/src/device.rs
```diff
@@ -406,17 +405,14 @@ impl Program {
...
-                panic!("-- Program link failed - exiting --");
-            }
-            false
-        } else {
-            //println!("{}", gl::get_program_info_log(self.id));
-            true
+            return Err(ShaderError::Link(error_log));
         }
+
+        Ok(())
```
kvark: similarly with shader compiling, we should still print out warnings on the successful builds too

webrender/src/device.rs
```diff
@@ -1391,27 +1387,28 @@ impl Device {
...
                                            &program.fs_source,
                                            gl::FRAGMENT_SHADER,
-                                           &fs_preamble,
-                                           panic_on_fail);
+                                           &fs_preamble);
 
         match (vs_id, fs_id) {
-            (Some(vs_id), None) => {
+            (Ok(vs_id), Err(e)) => {
                 println!("FAILED to load fs - falling back to previous!");
```
kvark: I don't think failing back to a previous shader is a good idea, especially if we get in a partial state where, for example, the FS is new and the VS is old.
It would be much nicer to do this idiomatically, as in - `let vs_id = try!(Device::compile_shader(..));` without the need to do this match afterwards.

webrender/src/renderer.rs
```diff
@@ -412,6 +413,20 @@ pub struct Renderer {
     vr_compositor_handler: Arc<Mutex<Option<Box<VRCompositorHandler>>>>
 }
 
+#[derive(Debug)]
+pub enum InitError {
+    Shader(ShaderError),
+    Thread(std::io::Error),
+}
+
+impl std::convert::From<ShaderError> for InitError {
```
kvark: I don't think you need the full path here: `std::convert::From` should be in prelude already

webrender/src/renderer.rs
```diff
@@ -1006,59 +1058,59 @@ impl Renderer {
 
         let (data, marker, shader) = match &batch.data {
             &PrimitiveBatchData::CacheImage(ref data) => {
-                let shader = self.ps_cache_image.get(&mut self.device, transform_kind);
+                let shader = self.ps_cache_image.get(&mut self.device, transform_kind).unwrap();
```
kvark: we could get away without `uwnrap`-ping in each match arm by doing it after the match

glennw:
Needs rebasing.

glennw:
Merging manually, as bors seems stuck, but travis / appveyor have passed.

PR #769, opened 2017-01-23
Add a way to be notified when new epochs have been rendered.

nical:
This makes it possible for Gecko to send notifications back to the content process for each rendered transaction.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/769)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -741,6 +741,12 @@ impl Renderer {
         self.pipeline_epoch_map.get(&pipeline_id).cloned()
     }
 
+    /// Returns a HashMap containing the pipeline ids that have been received by the renderer and
+    /// their respective epochs since the last time the method was called.
+    pub fn flush_rendered_epochs(&mut self) -> HashMap<PipelineId, Epoch, BuildHasherDefault<FnvHasher>> {
+        return mem::replace(&mut self.pipeline_epoch_map, HashMap::with_hasher(Default::default()));
+    }
+
     /// Processes the result queue.
     ///
     /// Should be called before `render()`, as texture cache updates are done here.


glennw:
@nical Just to clarify - the intent here that it behaves the same as pipeline_epoch_map but that you only get notification that the epoch changed the first time it was rendered?

nical:
Yes, the intent is to notify the content process that an update to a pipeline has been rendered. This is used to throttle the refresh driver if the compositor/renderer can't keep up with the content process producing frames (we throttle content if it gets more than 2 or 3 frames ahead of the renderer).

@jrmuizel just suggested adding another notifier trait which which would be used by the renderer instead of returning the iterator. This notifier api would make things a lot nicer to implement on the gecko side than the iterator, so I'll update the PR.

nical:
@glennw I updated the PR so that a RenderedPipelinesNotifier can be registered to the Renderer, and the latter will use it to signal every pipeline that is rendered at the end of every render() call (with its most recent epoch).
In the process of doing that I have the renderer clear its pipeline_epoch_map every time it renders, which means that servo has to use this new notifier API instead of the thing it is currently using to for reftests.
It's for the best because the current way pipeline_epoch_map works is a slow memory leak which would have to be changed one way or another.

webrender/src/renderer.rs
```diff
@@ -856,6 +861,11 @@ impl Renderer {
             // Restore frame - avoid borrow checker!
             self.current_frame = Some(frame);
         }
+        if let Some(ref mut notifier) = self.rendered_pipeline_notifier {
```
glennw: Can we make the notifier callback just take &PipelineEpochMap? That way, the mem::replace() can be removed and the clear() below can be executed unconditionally, which seems tidier.

nical:
How about [this](https://github.com/servo/webrender/pull/769/files#diff-f5062b694b9fe53fc1757ed483d577d9R865)? In gecko this map needs to be sent to another thread so moving it out avoids the boilerplate of allocating another container and copying into it while we know this one is going to be cleared.

If you prefer I can change this to be a function on the renderer that would look like ```pub fn get_frame_pipelines(&mut self) -> PipelineEpochMap { mem::replace(&mut self.pipeline_epoch_map, PipelineEpochMap::new()) }```. Gecko and Servo would need to call this every frame for that information to make sense and so that the map doesn't slowly accumulate.

As long as we can get the list of pipelines rendered each frame with their current epochs, and send that back to the content process through the ipc thread. Our render thread does not know which pipelines to query and pulling the list of all pipeline epochs in existence would be a bit much when hundreds of tabs are open. Ideally we'd only get notified for new epochs but having the few ones that are rendered each frame is enough and we can filter out notifications for already-rendered epochs when that info makes it to the ipc thread.

nical:
Nevermind my previous comment. In Gecko we want to be able to send a bit of extra information along with the notifications so the notifier API ends up getting in the way.

I replaced the commit by a simpler one which just adds a function to flush and return the epoch map. This will be enough for Gecko to do its thing and we can come back and improve the API later as the moving parts settle.

PR #770, opened 2017-01-23
Pipe dwrite rendering options to rasterize_glyphs

changm:
Essentially replicating this from gecko - http://searchfox.org/mozilla-central/source/gfx/2d/DrawTargetSkia.cpp#1390

I contemplated various ways to pass glyph rendering options to rasterize_glyphs either through traits or some struct, but we don't do anything special for linux or mac, so using a option seemed like the best thing, even though it exposes "dwrite" and forces the other platforms to use it. Nor could I find a clean way to pass derived types with data only as traits seemed to only pass down method interfaces.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/770)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -227,7 +227,8 @@ fn main() {
                           font_key,
                           ColorF::new(1.0, 1.0, 0.0, 1.0),
                           Au::from_px(32),
-                          Au::from_px(0));
+                          Au::from_px(0),
+                          None);
     }
 
     builder.pop_stacking_context();
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -554,7 +554,8 @@ impl Frame {
                                              text_info.size,
                                              text_info.blur_radius,
                                              &text_info.color,
-                                             text_info.glyphs);
+                                             text_info.glyphs,
+                                             text_info.glyph_options);
                 }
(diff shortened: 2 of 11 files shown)


glennw:
Do these options need to be stored per text run dynamically? For instance, can they be set once per frame / font or at init time perhaps? How often do they change?

changm:
They can't be stored once per frame, it depends on the text run :(. Layout determines whether or not to enable these options for each text run. Even for the same font, depending on the font size, the options can change.

glennw:
OK, in that case, how about we rename the struct to just be GlyphOptions, and then we can specify which platforms each field applies to. This will allow us to use the same struct for other platform-specific options in the future if we need to.

changm:
How do you specify how each field applies to which platform? I couldn't find something like that. I thought about going the GlyphOptions route but we don't use that for CG/Unix and I'd rather keep it very explicit what it's for. That's also why I passed it in as an Option rather than the full struct.

glennw:
You can use cfg() on each field to make it only available on that platform - but I was actually just suggesting a comment above each field specifying that this applies to Windows only. So I think if we just rename DWriteGlyphOptions -> GlyphOptions and add a comment for each of the two fields saying they only apply to Windows, that would be fine for now.

changm:
Thanks for the review! Renamed and a comment added saying that GlyphOptions are only used on windows for now.

PR #772, opened 2017-01-24
Remove pipeline_size_changed from RenderNotifier and update toml

hgallagher1993:
r? @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/772)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -361,8 +361,6 @@ impl RenderBackend {
 
     fn build_scene(&mut self) {
         // Flatten the stacking context hierarchy
-        let mut new_pipeline_sizes = HashMap::new();
-
         if let Some(id) = self.current_bound_webgl_context_id {
             self.webgl_contexts[&id].unbind();
             self.current_bound_webgl_context_id = None;
@@ -381,52 +379,7 @@ impl RenderBackend {
             webgl_context.unbind();
         }
 
-        self.frame.create(&self.scene, &mut new_pipeline_sizes);
-
-        let mut updated_pipeline_sizes = HashMap::new();
-
-        for (pipeline_id, old_size) in self.scene.pipeline_sizes.drain() {
-            let new_size = new_pipeline_sizes.remove(&pipeline_id);
-
-            match new_size {
-                Some(new_size) => {
-                    // Exists in both old and new -> check if size changed
-                    if new_size != old_size {
-                        let mut notifier = self.notifier.lock();
-                        notifier.as_mut()
-                                .unwrap()
-                                .as_mut()
-                                .unwrap()
-                                .pipeline_size_changed(pipeline_id, Some(new_size));
-                    }
-
-                    // Re-insert
-                    updated_pipeline_sizes.insert(pipeline_id, new_size);
-                }
-                None => {
-                    // Was existing, not in current frame anymore
-                        let mut notifier = self.notifier.lock();
-                        notifier.as_mut()
-                                .unwrap()
-                                .as_mut()
-                                .unwrap()
-                                .pipeline_size_changed(pipeline_id, None);
-                }
-            }
-        }
-
-        // Any remaining items are new pipelines
-        for (pipeline_id, new_size) in new_pipeline_sizes.drain() {
-            let mut notifier = self.notifier.lock();
-            notifier.as_mut()
-                    .unwrap()
-                    .as_mut()
-                    .unwrap()
-                    .pipeline_size_changed(pipeline_id, Some(new_size));
-            updated_pipeline_sizes.insert(pipeline_id, new_size);
-        }
-
-        self.scene.pipeline_sizes = updated_pipeline_sizes;
+        self.frame.create(&self.scene);
     }
 
     fn render(&mut self) -> RendererFrame {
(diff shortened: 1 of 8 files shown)


jdm:
Did you build after making this change? There are places in the code that call the method that was removed.

hgallagher1993:
I actually forgot to build it :frowning:  . . .there was 3 calls to it anyway in render_backend.rs, I'm just waiting for a file lock to lift before it'll build

hgallagher1993:
Ya after just removing those 3 calls to `pipeline_size_changed` webrender compiles but not Servo because 

`fn pipeline_size_changed(&mut self,_: webrender_traits::PipelineId,_: Option<webrender_traits::LayoutSize>) { }`

is still in the `impl` in compositor.rs, everything compiled fine when I removed that. So that should be removed fairly soon after this pull request is finished should it?

jdm:
Yep.

hgallagher1993:
If this is ok do you want me to squash the commit?

webrender/src/render_backend.rs
```diff
@@ -401,8 +401,7 @@ impl RenderBackend {
                         notifier.as_mut()
                                 .unwrap()
                                 .as_mut()
-                                .unwrap()
-                                .pipeline_size_changed(pipeline_id, Some(new_size));
+                                .unwrap();
```
jdm: This code doesn't do anything now, since it isn't used to notify anything. We can remove all of the code that compares the sizes of pipelines and stores them for later.

jdm:
Squashing commits is fine.

hgallagher1993:
I can make the change to Servo as well if you want

webrender/src/render_backend.rs
hgallagher1993: Sorry I'm just struggling to get this, it's the first time I've ever seen this code :-/, Is it basically everything to do with `updated_pipeline_sizes` that can be removed? So pretty much everything bellow `self.frame.create(&self.scene, &mut new_pipeline_sizes);`

jdm: As far as I know, the only use of storing the known pipeline sizes is to compare them in this code. Since we don't need to notify any other code about changes to pipeline sizes, that means we can get rid of all of the code that compares them, as well as any code that stores them for later use. Basically, any code that now does not do anything with the result of comparisons can be removed.

glennw:
@jdm The Servo code that uses this feature is being reverted due to causing test failures, I believe (see https://github.com/servo/servo/pull/15164). If that's right, we should hold off on this until the Servo code changes land without causing regressions?

jdm:
I'm inclined to hold off on merging it, yes. But I think working on it before then is fine.

glennw:
Yes, that's fine - it may just involve a lot of rebasing the patch, depending on how long it takes to get the Servo changes to land.

emilio:
I think the servo changes were legitimately green on try, so should be
easy to reland them I hope.

glennw:
I think the Servo changes have landed now? However, there are CI compile failures for this PR that need to be resolved first.

hgallagher1993:
I think it's just the Cargo.lock I have to commit to get it compiling, it's compiling for me locally now like, just have to makes the changes jdm requested first

jdm:
I will defer to the proper reviewers for this code, but that looks more like what I expected to see :)

glennw:
This looks good - there's a couple CI errors to fix though. If you ```cd wrench && cargo build``` and then ```cd sample && cargo build``` you should be able to see the errors to fix locally.

hgallagher1993:
Wrench and the sample are building now

emilio:
This looks good, but could you remove the merge commits, please?

hgallagher1993:
To remove the 2 commits is it just `git revert -m 1 [sha]` then commit and push the reverts? I've been reading up on reverting merges and I've found a few different answers. . . . .[This is the answer I'm going off atm](https://stackoverflow.com/questions/7099833/how-to-revert-a-merge-commit-thats-already-pushed-to-remote-branch/31223198#31223198)

emilio:
You can do `git rebase -i HEAD~5`, then remove the lines corresponding to merge commits.

glennw:
Waiting on squash and rebase, then this should be good to go.

hgallagher1993:
Just on the merges I've to remove, I ran `git rebase -i HEAD~5` but the 2 merge commits don't appear anywhere in the list, so is there any other way they could be removed?

emilio:
If you check the "Allow edits from collaborators" I can do it for you, exporting the patches with `git format-patch`, then resetting to master, then applying them with `git am --3way` works, though I'm pretty sure there should be an easier way...

hgallagher1993:
That's checked now

hgallagher1993:
The Allow edits from maintainers

emilio:
Just `git fetch upstream` and `git rebase upstream/master` works, if you want to give it a try. Otherwise I can just push the squashed version :)

hgallagher1993:
Is there any chance you could do it? because I'm actually just about to go back to a class in college that I'll be in for a few hours, I was only on a break just now :laughing:

emilio:
Sure :)

emilio:
Thanks for working on this @hgallagher1993!

PR #773, opened 2017-01-24
Allow to not depend on offscreen_gl_context.

emilio:
This PR allows (via feature-gate) to compile out enough WebGL-related code in
order to not depend in offscreen_gl_context.

Fixes bug 1331546
Fixes https://github.com/emilio/rust-offscreen-rendering-context/issues/77

r? @glennw or @kvark 
cc @sotaroikeda

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/773)


webrender/src/resource_cache.rs
```diff
@@ -186,6 +186,7 @@ pub struct ResourceCache {
     cached_images: ResourceClassCache<ImageRequest, CachedImageInfo>,
 
     // TODO(pcwalton): Figure out the lifecycle of these.
+    #[cfg_attr(not(feature = "webgl"), allow(dead_code))]
```
kvark: why do we need to allow all these `dead_code` pieces instead of completely removing them?

emilio: We can remove them indeed, but overall in big structs, I didn't think it is worth to duplicate all the constructors just to avoid that field. I can do it if you prefer it that way, I just thought adding fields would become a bit more annoying.

emilio:
Let me know if you want me to do a followup removing those fields.

@bors-servo r=kvark

emilio:
Had to work around https://github.com/serde-rs/serde/issues/720. Isn't ultra-pretty, but didn't think of a better way.

Mind reviewing the last commit @kvark?

glennw:
@emilio This appears to have compile errors on CI.

emilio:
Gah, of course, this builds fine with derive + WebGL (Servo's configuration), and codegen - WebGL (Gecko's), but none of those are the default.

@sotaroikeda, if this is not hi-priority, we could wait a few weeks until rust stable supports custom derive (it's on beta right now), are you fine with that?

webrender_traits/src/types.rs
```diff
@@ -16,6 +16,15 @@ use std::sync::Arc;
 #[cfg(target_os = "macos")] use core_graphics::font::CGFont;
 #[cfg(target_os = "windows")] use dwrote::FontDescriptor;
 
+// Some stubs to work around https://github.com/serde-rs/serde/issues/720
+#[cfg(not(feature = "webgl"))]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct GLContextAttributes([u8; 0]);
```
kvark: just curious, what is the point of having `[u8; 0]` internally?

emilio: Apparently it's the sound way of making opaque enums that can't be instantiated, see https://github.com/rust-lang/rfcs/pull/1861#issuecomment-274613536

glennw:
@emilio @sotaroikeda Have we decided to hold off on this until the rustc support lands?

emilio:
Yes, I believe so, unless it's super-hi-pri for Gecko, it's going to be cleaner than adding a lots of workarounds and then removing them shortly after.

sotaroikeda:
@emilio, can we do it cleaner with #941?

emilio:
Yes, I'll rebase that as soon as I have the time.

PR #774, opened 2017-01-24
Apply gamma correction for coregraphics fonts

changm:
I finished Jeff's port of Skia's gamma mask code here - https://github.com/changm/gamma-lut-rs, which is the packaged added in this patch. 

The original source is here - http://searchfox.org/mozilla-central/source/gfx/skia/skia/src/core/SkMaskGamma.cpp

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/774)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -16,12 +16,14 @@ use core_text;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
 use webrender_traits::{ColorU, FontKey, FontRenderMode, GlyphDimensions, GlyphOptions};
+use gamma_lut::{GammaLut, Color as ColorLut};
 
 pub type NativeFontHandle = CGFont;
 
 pub struct FontContext {
     cg_fonts: HashMap<FontKey, CGFont>,
     ct_fonts: HashMap<(FontKey, Au), CTFont>,
+    gamma_lut: GammaLut,
 }
 
 pub struct RasterizedGlyph {
@@ -92,9 +94,14 @@ impl FontContext {
     pub fn new() -> FontContext {
         debug!("Test for subpixel AA support: {}", supports_subpixel_aa());
 
+        // Force CG to use sRGB color space to gamma correct.
+        let contrast = 0.0;
+        let gamma = 0.0;
+
         FontContext {
             cg_fonts: HashMap::new(),
             ct_fonts: HashMap::new(),
+            gamma_lut: GammaLut::new(contrast, gamma, gamma),
         }
     }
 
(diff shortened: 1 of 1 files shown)


webrender/src/platform/macos/font.rs
```diff
@@ -158,9 +165,36 @@ impl FontContext {
         })
     }
 
+    // Assumes the pixels here are linear values from CG
+    fn gamma_correct_pixels(&self, pixels: &mut Vec<u8>, width: usize,
```
emilio: nit: As far as I understand it, `pixels`'s length cannot change, right? If so, you could probably make it a `&mut [u8]` so you save an indirection.

changm: Ahh good point. Thanks! Since I already have the dwrite gamma correction also using vector, can I fix this + dwrite + the gamma-lut library in a follow up patch? Too many patches in mid flight right now.

emilio: Sounds fine to me :)

webrender/src/platform/macos/font.rs
```diff
@@ -16,12 +16,14 @@ use core_text;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
 use webrender_traits::{ColorU, FontKey, FontRenderMode, GlyphDimensions};
+use gamma_lut;
```
kvark: would be cleaner to `use gamma_lut::{GammaLut, Color as ColorLut};`

webrender/src/platform/macos/font.rs
```diff
@@ -158,9 +165,36 @@ impl FontContext {
...
 
+    // Assumes the pixels here are linear values from CG
+    fn gamma_correct_pixels(&self, pixels: &mut Vec<u8>, width: usize,
+                            height: usize, render_mode: FontRenderMode,
+                            color: ColorU) {
+        // Then convert back to gamma corrected values.
+        match render_mode {
+            FontRenderMode::Alpha => {
+                self.gamma_lut.preblend_grayscale_bgra(pixels, width, height,
+                                                       gamma_lut::Color::new(color.r,
```
kvark: could move the `ColorLut` initialization out of the `match`

webrender/src/platform/macos/font.rs
```diff
@@ -92,9 +94,14 @@ impl FontContext {
...
         debug!("Test for subpixel AA support: {}", supports_subpixel_aa());
 
+        // Force CG to use sRGB color space to gamma correct.
+        let contrast = 0.0;
+        let gamma = 0.0;
+
         FontContext {
             cg_fonts: HashMap::new(),
             ct_fonts: HashMap::new(),
+            gamma_lut: gamma_lut::GammaLut::new(contrast, gamma, gamma),
```
kvark: given that you pass the same gamma as both paint gamma and device gamma, the whole LUT'ing thing becomes just a contrast application, according to the source:
```rust
if (src - dst).abs() < (1.0 / 256.0) {
        let mut ii : f32 = 0.0;
        for i in 0..256 {
            let raw_srca = ii / 255.0;
            let srca = apply_contrast(raw_srca, adjusted_contrast);

            table[i] = round_to_u8(255.0 * srca);
            ii += 1.0;
        }
}
```
Is this by design? Are we going to use different gamma parameters here in the long term?

changm: It is by design at the moment, it's just a port from Skia - http://searchfox.org/mozilla-central/source/gfx/skia/skia/src/core/SkPaint.cpp#1498

Also, that bit of code isn't the src/dst gamma. That's after the luminance conversion - http://searchfox.org/mozilla-central/source/gfx/skia/skia/src/core/SkMaskGamma.cpp#79. We don't actually hit that if block.

changm:
@kvark Did I answer your question? If so, please merge. Thanks!

kvark:
Thanks!
@bors-servo r+

Issue #661, opened 2016-12-22
init_texture() does not care format change on arm


sotaroikeda:
On arm, ImageFormat::A8 uses GL_FORMAT_BGRA. Then init_texture() need to expand data like update_texture(). By it, debug_render's rendering seems to be broken on my nexus5X with quantum graphic branch.
  https://bugzilla.mozilla.org/show_bug.cgi?id=1323612

device.rs defines the following, but it seems not used.
  >#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
  > const GL_FORMAT_A: gl::GLuint = gl::ALPHA;

glennw:
This was fixed in https://github.com/servo/webrender/pull/779.

PR #779, opened 2017-01-24
Expand data in init_texture() on arm like update_texture()

sotaroikeda:
Fixes issue #661.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/779)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -14,6 +14,7 @@ use std::collections::HashMap;
 use std::fs::File;
 use std::hash::BuildHasherDefault;
 use std::io::Read;
+use std::iter::repeat;
 use std::mem;
 use std::path::PathBuf;
 //use std::sync::mpsc::{channel, Sender};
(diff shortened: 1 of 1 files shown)


webrender/src/device.rs
```diff
@@ -1111,13 +1111,33 @@ impl Device {
...
                 self.set_texture_parameters(texture_id.target, filter);
-                self.upload_texture_image(texture_id.target,
-                                          width,
-                                          height,
-                                          internal_format as u32,
-                                          gl_format,
-                                          type_,
-                                          pixels);
+                let mut expanded_data = Vec::new();
+                if pixels.is_some() &&
```
kvark: we can rewrite this to be nicer, like
```rust
let actual_pixels = if <same condition> {
    // need to `use std::iter::repeat` at the top
    pixels.unwrap().iter().flat_map(|&byte| repeat(byte).take(4)).collect();
} else {
    pixels
};
self.upload_texture_image(..., actual_pixels);
```

sotaroikeda:
@kvark, I updated a patch. Can you review again?

kvark:
Thanks!
@bors-servo r+

PR #780, opened 2017-01-25
Quantize glyph positions for subpixel text positioning.

changm:
Changes the GlyphKey to take into account subpixel text positioning. After this patch, I'll pipe in the subpixel text position into rasterize_glyphs. (I'm trying to make smaller, more easily reviewable patches. Please let me know if you prefer bigger ones that are all "done").

The big issue I had with this patch was that GlyphInstances are attached to auxiliary display lists which are only available with a stacking context. To get around it, I added the stacking context id to TextRunPrimitiveCpu and assign it when we're iterating over the primitive commands in tiling.rs::cull_layers. That's where we were originally getting access to the GlyphInstances. I'm not sure this is the best way to do this, but the logic seems to be cull_layers, rasterize the glyphs, then fetch the glyphs based on the GlyphKeys. However, the GlyphKeys now need the GlyphInstances to calculate the subpixel text position key to fetch later, but we don't get access to those without the auxiliary display lists.

I'd be happy if you have a better idea (or we rewrite how GlyphInstances are handled in general). Was passing around GlyphInstances that expensive vs indices (1 u32 vs 1 u32 + 2 f32)? I can understand the raw font data, but the indices? Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/780)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -18,8 +18,8 @@ use webrender_traits::{FontKey, FontRenderMode, WebGLContextId};
 use webrender_traits::{device_length, DeviceIntRect, DeviceIntSize};
 use webrender_traits::{DeviceRect, DevicePoint, DeviceSize};
 use webrender_traits::{LayerRect, LayerSize, LayerPoint};
-use webrender_traits::LayerToWorldTransform;
-use webrender_traits::{GlyphOptions};
+use webrender_traits::{LayerToWorldTransform, GlyphOptions};
+use tiling::{AuxiliaryListsMap, StackingContext, StackingContextIndex};
 
 pub const CLIP_DATA_GPU_SIZE: usize = 5;
 pub const MASK_DATA_GPU_SIZE: usize = 1;
@@ -457,7 +457,7 @@ pub struct PrimitiveStore {
     pub gpu_resource_rects: GpuStore<TexelRect>,
 
     // General
-    prims_to_resolve: Vec<PrimitiveIndex>,
+    prims_to_resolve: Vec<(StackingContextIndex, PrimitiveIndex)>,
 }
 
 impl PrimitiveStore {
@@ -718,10 +718,12 @@ impl PrimitiveStore {
 
     pub fn resolve_primitives(&mut self,
                               resource_cache: &ResourceCache,
-                              device_pixel_ratio: f32) -> Vec<DeferredResolve> {
+                              device_pixel_ratio: f32,
+                              layer_store: &Vec<StackingContext>,
+                              auxiliary_lists_map: &AuxiliaryListsMap) -> Vec<DeferredResolve> {
         let mut deferred_resolves = Vec::new();
 
-        for prim_index in self.prims_to_resolve.drain(..) {
+        for (sc_index, prim_index) in self.prims_to_resolve.drain(..) {
             let metadata = &mut self.cpu_metadata[prim_index.0];
             if let Some(ref clip_info) = metadata.clip_cache_info {
                 Self::resolve_clip_cache_internal(&mut self.gpu_data32, clip_info, resource_cache);
@@ -735,14 +737,21 @@ impl PrimitiveStore {
                 PrimitiveKind::RadialGradient=> {}
                 PrimitiveKind::TextRun => {
                     let text = &mut self.cpu_text_runs[metadata.cpu_prim_index.0];
+
                     let font_size_dp = text.logical_font_size.scale_by(device_pixel_ratio);
 
                     let dest_rects = self.gpu_resource_rects.get_slice_mut(text.resource_address,
                                                                            text.glyph_range.length);
+
+                    let layer = &layer_store[sc_index.0];
+                    let auxiliary_lists = auxiliary_lists_map.get(&layer.pipeline_id)
+                                                             .expect("No auxiliary lists?");
+                    let src_glyphs = auxiliary_lists.glyph_instances(&text.glyph_range);
+
                     let texture_id = resource_cache.get_glyphs(text.font_key,
                                                                font_size_dp,
                                                                text.color,
-                                                               &text.glyph_indices,
+                                                               src_glyphs,
                                                                text.render_mode,
                                                                text.glyph_options, |index, uv0, uv1| {
                         let dest_rect = &mut dest_rects[index];
@@ -878,7 +887,8 @@ impl PrimitiveStore {
                                    resource_cache: &mut ResourceCache,
                                    layer_transform: &LayerToWorldTransform,
                                    device_pixel_ratio: f32,
-                                   auxiliary_lists: &AuxiliaryLists) -> bool {
+                                   auxiliary_lists: &AuxiliaryLists,
+                                   stackingcontext_index: StackingContextIndex) -> bool {
 
         let metadata = &mut self.cpu_metadata[prim_index.0];
         let mut prim_needs_resolve = false;
@@ -918,7 +928,9 @@ impl PrimitiveStore {
             }
             PrimitiveKind::TextRun => {
                 let text = &mut self.cpu_text_runs[metadata.cpu_prim_index.0];
+
                 let font_size_dp = text.logical_font_size.scale_by(device_pixel_ratio);
+                let src_glyphs = auxiliary_lists.glyph_instances(&text.glyph_range);
                 prim_needs_resolve = true;
 
                 if text.cache_dirty {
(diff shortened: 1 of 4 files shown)


webrender/src/prim_store.rs
```diff
@@ -290,6 +291,7 @@ pub struct TextRunPrimitiveCpu {
     pub color: ColorF,
     pub render_mode: FontRenderMode,
     pub resource_address: GpuStoreAddress,
+    pub stackingcontext_index: usize,
```
glennw: I think there's probably a cleaner way to achieve this without requiring storing the stacking context index here. I'll take a closer look at this tomorrow.

webrender/src/prim_store.rs
```diff
@@ -292,6 +292,7 @@ pub struct TextRunPrimitiveCpu {
     pub render_mode: FontRenderMode,
     pub resource_address: GpuStoreAddress,
     pub glyph_options: Option<GlyphOptions>,
+    pub stackingcontext_index: usize,
```
glennw: Let's remove this.

webrender/src/prim_store.rs
```diff
@@ -735,14 +738,22 @@ impl PrimitiveStore {
                 PrimitiveKind::RadialGradient=> {}
                 PrimitiveKind::TextRun => {
                     let text = &mut self.cpu_text_runs[metadata.cpu_prim_index.0];
+                    let sc_index = text.stackingcontext_index;
```
glennw: Let's make the prims_to_resolve field in PrimitiveStore a ```Vec<(StackingContextIndex, PrimitiveIndex)>```. Then the stacking context index is available for each primitive, and we don't need to store it in the primitive data itself.

This is still not ideal - since it means a hash lookup in the aux lists for each text run, rather than once per stacking context. But we can fix that later, as a follow up.

webrender/src/prim_store.rs
```diff
@@ -927,18 +942,24 @@ impl PrimitiveStore {
...
+                                                      src_glyphs[0].x,
+                                                      src_glyphs[0].y,
+                                                      text.render_mode,
+                                                      text.glyph_options);
                     let mut local_rect = LayerRect::zero();
                     let mut actual_glyph_count = 0;
 
                     for src in src_glyphs {
                         glyph_key.index = src.index;
+                        glyph_key.set_x_offset(src.x);
```
glennw: This could probably just be set_subpixel_offset(x, y) instead of two calls.

webrender/src/resource_cache.rs
```diff
@@ -62,28 +62,6 @@ pub struct CacheItem {
     pub uv1: DevicePoint,
 }
 
-#[derive(Clone, Hash, PartialEq, Eq, Debug, Ord, PartialOrd)]
```
glennw: Does this need to be removed? The idea here is that when passing around the glyph keys, they can be smaller by not including duplicated fields that are the same for an entire text run (the glyph options, and render mode).

webrender/src/resource_cache.rs
```diff
@@ -359,22 +337,26 @@ impl ResourceCache {
...
                          color: ColorF,
-                         glyph_indices: &[u32],
+                         glyph_instances: &[GlyphInstance],
                          render_mode: FontRenderMode,
                          glyph_options: Option<GlyphOptions>,
                          mut f: F) -> SourceTexture where F: FnMut(usize, DevicePoint, DevicePoint) {
         debug_assert!(self.state == State::QueryResources);
         let cache = self.cached_glyphs.as_ref().unwrap();
         let render_mode = self.get_glyph_render_mode(render_mode);
-        let mut glyph_key = RenderedGlyphKey::new(font_key,
```
glennw: I'd prefer to keep the RenderedGlyphKey type, if possible.

webrender/src/tiling.rs
```diff
@@ -1681,8 +1681,16 @@ impl PrimitiveBatch {
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct StackingContextIndex(usize);
 
-struct StackingContext {
-    pipeline_id: PipelineId,
+#[derive(Debug)]
+struct TileRange {
```
glennw: This is an unused struct - probably a rebase error?

webrender/src/tiling.rs
```diff
@@ -2160,6 +2168,8 @@ impl FrameBuilder {
                 render_mode: render_mode,
                 glyph_options: glyph_options,
                 resource_address: GpuStoreAddress(0),
+                // Note this is filled later when we assign this to astacking context
+                stackingcontext_index: 0,
```
glennw: I think we can remove this, once the other changes are made.

webrender_traits/src/types.rs
```diff
@@ -322,18 +330,62 @@ pub struct GlyphKey {
...
+            glyph_options: glyph_options,
+        };
+
+        key.set_x_offset(x);
+        key.set_y_offset(y);
+
+        key
+    }
+
+    pub fn set_x_offset(&mut self, x: f32) {
```
glennw: Let's combine this with the set_y_offset method below.

webrender_traits/src/types.rs
```diff
@@ -322,18 +330,62 @@ pub struct GlyphKey {
...
+        self.x_suboffset = self.subpixel_quantize_offset(x);
+    }
+
+    pub fn set_y_offset(&mut self, y: f32) {
+        self.y_suboffset = self.subpixel_quantize_offset(y);
+    }
+
+    // Skia quantizes subpixel offets into 1/4 increments.
+    // Given the absolute position, return the quantized increment
+    fn subpixel_quantize_offset(&self, pos: f32) -> SubpixelOffset {
```
glennw: Let's make this a free function that takes a render mode as a parameter. Then the constructor for GlyphKey can be simplified to call this directly while creating the struct.

changm:
@glennw r? I'm not sure how much we're saving from having RenderedGlyphKey be a separate thing than GlyphKey, since almost all accesses require all the same data. I guess the only difference will be the Option<GlyphOptions>, but I don't feel strong enough either way. 

Otherwise though, much cleaner! Thanks!

webrender/src/prim_store.rs
```diff
@@ -718,10 +718,15 @@ impl PrimitiveStore {
...
     pub fn resolve_primitives(&mut self,
                               resource_cache: &ResourceCache,
-                              device_pixel_ratio: f32) -> Vec<DeferredResolve> {
+                              device_pixel_ratio: f32,
+                              layer_store: &Vec<StackingContext>,
+                              auxiliary_lists_map: &AuxiliaryListsMap) -> Vec<DeferredResolve> {
         let mut deferred_resolves = Vec::new();
 
-        for prim_index in self.prims_to_resolve.drain(..) {
+        for resolved_prim in self.prims_to_resolve.drain(..) {
```
glennw: This can be written as:

```for (sc_index, prim_index) in self.prims_to_resolve.drain(..) { ```

**Comment by changm** [comment]
Thanks, fixed.

**Comment by glennw** [comment]
Looks good, thanks!

File: webrender/src/prim_store.rs [hunk 12]
```diff
@@ -718,10 +718,12 @@ impl PrimitiveStore {
 
     pub fn resolve_primitives(&mut self,
                               resource_cache: &ResourceCache,
-                              device_pixel_ratio: f32) -> Vec<DeferredResolve> {
+                              device_pixel_ratio: f32,
+                              layer_store: &Vec<StackingContext>,
```
kvark: nit: should use `&[StackingContext]` instead of `Vec`

webrender/src/prim_store.rs
```diff
@@ -718,10 +718,12 @@ impl PrimitiveStore {
...
+                              auxiliary_lists_map: &AuxiliaryListsMap) -> Vec<DeferredResolve> {
```
kvark: Ideally, I'd like to see the auxiliary lists accessed from a single place, like `cull_layers`. Although, it doesn't seem to have any other downsides here than the code complexity, so we can have it for now and then tentatively look into streamlining the use of the aux lists.

webrender/src/resource_cache.rs
```diff
@@ -74,10 +74,12 @@ impl RenderedGlyphKey {
                size: Au,
                color: ColorF,
                index: u32,
+               x: f32,
```
kvark: nit: we should be using one of the euclid-defined types for (x, y)

webrender_traits/src/types.rs
```diff
@@ -311,6 +311,33 @@ pub enum FontRenderMode {
...
+pub enum SubpixelOffset {
+    Zero,
+    Quarter,
+    Half,
+    ThreeQuarters,
+}
+
+// Skia quantizes subpixel offets into 1/4 increments.
+// Given the absolute position, return the quantized increment
+fn subpixel_quantize_offset(pos: f32, render_mode: FontRenderMode) -> SubpixelOffset {
```
kvark: nit: could be a function of `FontRenderMode`

webrender_traits/src/types.rs
```diff
@@ -322,20 +349,32 @@ pub struct GlyphKey {
     pub size: Au,
     pub index: u32,
     pub color: ColorU,
+    pub x_suboffset: SubpixelOffset,
+    pub y_suboffset: SubpixelOffset,
```
kvark: ideally, we'd want to have more variation of the X sub-pixel offsets then Y, right?

webrender/src/prim_store.rs
changm: Yeah, I didn't really like the aux list thing either. But we did a bunch of stuff after cull_layers and cull_layers was already decently big. We can refactor later?

webrender/src/resource_cache.rs
changm: Should we change GlyphInstance to use those as well? Since we're basically passing data from there and GlyphInstance is just f32.

webrender_traits/src/types.rs
changm: How come? Skia uses 4 for each. Also, there are vertical text languages that have subpixel text as well.

PR #781, opened 2017-01-25
Replace fixed position layer with reference frame

mrobinson:
Instead of creating fixed position layers which hang onto the outside of the
ScrollTree, create reference frames, which exist within the tree of
scroll layers. Eventually this tree of reference frames and scroll
layers will be combined with stacking context offsets to position and
clip all display items.

This change fixes an issue where fixed position items wouldn't scroll
even when their transformed ancestors which defined their coordinate
system did.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/781)


mrobinson:
@glennw r?

glennw:
Nice! This looks good to me. We don't have great test coverage of scrolling / fixed elements, but so long as this passes the servo tests and you've tested on a few of the more common sites, r=me.

glennw:
@mrobinson Related - we should perhaps consider adding some support to the wrench reftest code for setting scroll positions, and start making use of that for WR-specific scrolling reftests?

mrobinson:
@glennw Thanks! I'll work on adding scrolling support to wrench.

PR #783, opened 2017-01-26
Make ImageKey and FontKey members public.

nical:
We need some ffi types to represent them and to that end we have to poke at their members. Some other types like `PipelineId` have their members public so I suppose this should not be an issue.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/783)

### DIFF
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -302,7 +302,7 @@ pub enum FilterOp {
 }
 
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize, Ord, PartialOrd)]
-pub struct FontKey(u32, u32);
+pub struct FontKey(pub u32, pub u32);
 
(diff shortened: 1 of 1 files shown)


kvark:
I'd prefer having a getter in them instead of exposing the members. This would technically protect these structs from modification in safe code, at least, and I assume Gecko doesn't need to modify them (only read and/or create).

nical:
We need to be able to read the members and create new ids. so it's very simple to go ```foo.id = PipelineId::new(food.id.namespace(), foo.id.handle());```. If someone wants to modify the value them. I think that a better way to prevent that is to store the value immutably.

I don't mind having getters and constructors but it means we have to be disciplined about having them for all id types and make them have the same form (so that we can generate the ffi types using a macro. just adding pub in a few places seems simpler for more or less the same result.

kvark:
Good point. I think it's not critical, and should block you for now. We can revisit this decision later (as in - add a compile flag that exposes the fields, for example, if this is a concern).
@bors-servo r+

kvark:
Uh, we are running out of time on OSX Travis :(
@bors-servo retry

PR #790, opened 2017-01-26
Use dwrite recommended rendering mode and force gdi modes

changm:
Replicates the measuring and rendering modes gecko uses with skia. Depends on https://github.com/vvuk/dwrote-rs/pull/5 so it won't build right now along with the lack of bump for the dwrote-rs version. But everything else should be good to go. I'll add those to the commit once the dwrote-rs is done. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/790)

### DIFF
--- a/webrender/src/platform/windows/font.rs
+++ b/webrender/src/platform/windows/font.rs
@@ -18,7 +18,6 @@ lazy_static! {
     };
 }
 
-#[allow(dead_code)]
 pub struct FontContext {
     fonts: HashMap<FontKey, dwrote::FontFace>,
     gamma_lut: GammaLut,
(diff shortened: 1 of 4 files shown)


webrender/src/platform/windows/font.rs
```diff
@@ -40,22 +39,49 @@ fn dwrite_texture_type(render_mode: FontRenderMode) ->
     }
 }
 
-fn dwrite_measure_mode(render_mode: FontRenderMode) ->
+fn dwrite_measure_mode(render_mode: FontRenderMode, options: Option<GlyphOptions>) ->
                        dwrote::DWRITE_MEASURING_MODE {
+    if let Some(option) = options {
```
kvark: there is quite a few places where you check the options to be `Some()` and the contents to have `force_gdi_rendering == true`. FYI, you could combine the pattern match like this to simplify the code:
```rust
if let Some(GlyphOptions{ force_gdi_rendering: true, .. }) = option {...}
```

webrender/src/platform/windows/font.rs
```diff
@@ -65,6 +91,8 @@ fn get_glyph_dimensions_with_analysis(analysis: dwrote::GlyphRunAnalysis,
 
     let width = (bounds.right - bounds.left) as u32;
     let height = (bounds.bottom - bounds.top) as u32;
+    assert!(width > 0);
+    assert!(height > 0);
```
kvark: no need to have 2 different asserts here

vvuk:
@bors-servo delegate=changm

Looks fine on my end, I like @kvark's suggestions as well, especially the options match (TIL!). (Just put a space between `GlyphOptions` and `{` ;)

PR #792, opened 2017-01-27
revise gradients to use a lookup texture (supporting large amounts of…

lsalzman:
… stops cheaply) and allow repeating

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/792)

### DIFF
--- a/webrender/res/ps_angle_gradient.fs.glsl
+++ b/webrender/res/ps_angle_gradient.fs.glsl
@@ -2,38 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-float offset(int index) {
-    return vOffsets[index / 4][index % 4];
-}
-
-float linearStep(float lo, float hi, float x) {
-    float d = hi - lo;
-    float v = x - lo;
-    if (d != 0.0) {
-        v /= d;
-    }
-    return clamp(v, 0.0, 1.0);
-}
+uniform sampler2D sGradients;
 
 void main(void) {
-    float angle = atan(-vEndPoint.y + vStartPoint.y,
-                        vEndPoint.x - vStartPoint.x);
-    float sa = sin(angle);
-    float ca = cos(angle);
+    vec2 texture_size = vec2(textureSize(sGradients, 0));
 
-    float sx = vStartPoint.x * ca - vStartPoint.y * sa;
-    float ex = vEndPoint.x * ca - vEndPoint.y * sa;
-    float d = ex - sx;
+    // Either saturate or modulo the offset depending on repeat mode, then scale to number of
+    // gradient color entries (texture width / 2).
+    float x = mix(clamp(vOffset, 0.0, 1.0), fract(vOffset), vGradientRepeat) * 0.5 * texture_size.x;
 
-    float x = vPos.x * ca - vPos.y * sa;
+    // Start at the center of first color in the nearest 2-color entry, then offset with the
+    // fractional remainder to interpolate between the colors. Rely on texture clamping when
+    // outside of valid range.
+    x = 2.0 * floor(x) + 0.5 + fract(x);
 
-    oFragColor = mix(vColors[0],
-                     vColors[1],
-                     linearStep(sx + d * offset(0), sx + d * offset(1), x));
-
-    for (int i=1 ; i < vStopCount-1 ; ++i) {
-        oFragColor = mix(oFragColor,
-                         vColors[i+1],
-                         linearStep(sx + d * offset(i), sx + d * offset(i+1), x));
-    }
+    // Normalize the texture coordates so we can use texture() for bilinear filtering.
+    oFragColor = texture(sGradients, vec2(x, vGradientIndex) / texture_size);
 }
--- a/webrender/res/ps_radial_gradient.fs.glsl
+++ b/webrender/res/ps_radial_gradient.fs.glsl
@@ -2,18 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-float offset(int index) {
-    return vOffsets[index / 4][index % 4];
-}
-
-float linearStep(float lo, float hi, float x) {
-    float d = hi - lo;
-    float v = x - lo;
-    if (d != 0.0) {
-        v /= d;
-    }
-    return clamp(v, 0.0, 1.0);
-}
+uniform sampler2D sGradients;
 
 void main(void) {
     vec2 cd = vEndCenter - vStartCenter;
@@ -55,13 +44,17 @@ void main(void) {
         }
     }
 
-    oFragColor = mix(vColors[0],
-                     vColors[1],
-                     linearStep(offset(0), offset(1), x));
+    vec2 texture_size = vec2(textureSize(sGradients, 0));
 
-    for (int i=1 ; i < vStopCount-1 ; ++i) {
-        oFragColor = mix(oFragColor,
-                         vColors[i+1],
-                         linearStep(offset(i), offset(i+1), x));
-    }
+    // Either saturate or modulo the offset depending on repeat mode, then scale to number of
+    // gradient color entries (texture width / 2).
+    x = mix(clamp(x, 0.0, 1.0), fract(x), vGradientRepeat) * 0.5 * texture_size.x;
+
+    // Start at the center of first color in the nearest 2-color entry, then offset with the
+    // fractional remainder to interpolate between the colors. Rely on texture clamping when
+    // outside of valid range.
+    x = 2.0 * floor(x) + 0.5 + fract(x);
+
+    // Normalize the texture coordates so we can use texture() for bilinear filtering.
+    oFragColor = texture(sGradients, vec2(x, vGradientIndex) / texture_size);
 }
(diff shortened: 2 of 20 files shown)


glennw:

---

*[webrender/src/internal_types.rs, line 292 at r1](https://reviewable.io:443/reviews/servo/webrender/792#-Kbhlxwi8HY2wUnz73hG:-Kbhlxwi8HY2wUnz73hH:b7atcg0) ([raw file](https://github.com/servo/webrender/blob/d9606184d5bbf0d6307f88360ef21dee26dbc580/webrender/src/internal_types.rs#L292)):*
> ```Rust
>     }
> 
>     pub fn to_bgra(&self) -> PackedColor {
> ```

If we *do* have a BGRA color (see comments below), this should probably be a different type, to make use of the type system.

---

*[webrender/src/prim_store.rs, line 275 at r1](https://reviewable.io:443/reviews/servo/webrender/792#-KbhmXlem4PuwkVaK4xQ:-KbhmXlem4PuwkVaK4xR:bjwvhm2) ([raw file](https://github.com/servo/webrender/blob/d9606184d5bbf0d6307f88360ef21dee26dbc580/webrender/src/prim_store.rs#L275)):*
> ```Rust
> // the offset within that entry bucket is used to interpolate between the two colors in that entry.
> // This layout preserves hard stops, as the end color for a given entry can differ from the start
> // color for the following entry, despite them being adjacent. Colors are stored within in BGRA8
> ```

Is there a specific reason to introduce the BGRA code? It seems to introduce some additional complexity (esp. if PackedColor can be either format).

---

*[webrender/src/prim_store.rs, line 289 at r1](https://reviewable.io:443/reviews/servo/webrender/792#-KbhlMxfMzlVQ1P5kPLP:-KbhlMxfMzlVQ1P5kPLQ:b70ds6i) ([raw file](https://github.com/servo/webrender/blob/d9606184d5bbf0d6307f88360ef21dee26dbc580/webrender/src/prim_store.rs#L289)):*
> ```Rust
> }
> 
> impl Clone for GradientData {
> ```

I don't think there's any need to implement this - it can be derived if required with #[derive(Clone)]

---

*[webrender/src/prim_store.rs, line 728 at r1](https://reviewable.io:443/reviews/servo/webrender/792#-Kbhlf9nVLT_9WkyCpAo:-Kbhlf9oGm9YR_CLAdtd:b-gbhifu) ([raw file](https://github.com/servo/webrender/blob/d9606184d5bbf0d6307f88360ef21dee26dbc580/webrender/src/prim_store.rs#L728)):*
> ```Rust
> 
>                 let metadata = PrimitiveMetadata {
>                     is_opaque: false,
> ```

Let's add a TODO comment here to calculate if the primitive is actually opaque (so that it can be renderer in the opaque pass).

---

*[webrender/src/renderer.rs, line 149 at r1](https://reviewable.io:443/reviews/servo/webrender/792#-KbhjfDgcdJrISry_U4O:-KbhjfDhXYAndYbFNPEX:b-w1pkvv) ([raw file](https://github.com/servo/webrender/blob/d9606184d5bbf0d6307f88360ef21dee26dbc580/webrender/src/renderer.rs#L149)):*
> ```Rust
> 
>     fn texture_width() -> usize {
>         2 * GRADIENT_DATA_RESOLUTION
> ```

We should add a comment here explaining why it's 2*

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/792)*

glennw:
This looks great! I left a few comments and questions via reviewable. This will also need a rebase, and it looks like the shader validation is currently failing on CI.

webrender/res/ps_angle_gradient.fs.glsl
```diff
@@ -2,38 +2,17 @@
...
+    // Either saturate or modulo the offset depending on repeat mode, then scale to number of
+    // gradient color entries.
+    float x = mix(clamp(vOffset, 0.0, 1.0), fract(vOffset), vGradientRepeat) * vGradientScale;
 
-    oFragColor = mix(vColors[0],
-                     vColors[1],
-                     linearStep(sx + d * offset(0), sx + d * offset(1), x));
+    // Start at the center of first color in the nearest 2-color entry, then offset with the
+    // fractional remainder to interpolate between the colors.
+    x = 2.0 * floor(x) + 0.5 + fract(x);
```
kvark: This equation assumes that `0.0 <= x < 1.0`, so it will misbehave if `vOffset >= 1.0 && vGradientRepeat==false`

webrender/res/ps_radial_gradient.fs.glsl
```diff
@@ -55,13 +44,14 @@ void main(void) {
...
-        oFragColor = mix(oFragColor,
-                         vColors[i+1],
-                         linearStep(offset(i), offset(i+1), x));
-    }
+    // Start at the center of first color in the nearest 2-color entry, then offset with the
+    // fractional remainder to interpolate between the colors.
+    x = 2.0 * floor(x) + 0.5 + fract(x);
+
+    // Normalize the texture coordates so we can use texture() for bilinear filtering.
+    oFragColor = texture(sGradients, vec2(x, vGradientIndex) / vec2(textureSize(sGradients, 0)));
```
kvark: optional: since we never want to sample across gradients, it would make sense to have this texture as 1DArray

webrender/src/internal_types.rs
```diff
@@ -287,6 +288,15 @@ impl PackedColor {
             a: (0.5 + color.a * COLOR_FLOAT_TO_FIXED).floor() as u8,
         }
     }
+
+    pub fn to_bgra(&self) -> PackedColor {
```
kvark: why do we need to convert it to bgra?

webrender/src/tiling.rs
```diff
@@ -2038,21 +2031,20 @@ impl FrameBuilder {
...
-            GradientType::Rotated
-        };
+                        stops: ItemRange,
+                        extend_mode: ExtendMode) {
+        // Fast path for clamped, axis-aligned gradients:
+        let aligned = extend_mode == ExtendMode::Clamp &&
+                      (start_point.x == end_point.x ||
+                       start_point.y == end_point.y);
+        let reverse_stops = !aligned &&
+                            (start_point.x > end_point.x ||
```
lsalzman: The important thing here is that if the data is supplied in reverse, that we can produce an equivalent result. What the reference direction is becomes arbitrary, so long as we choose some reference direction. The aligned case didn't have to worry about that because its results rendered actually the same reversed or not, but the new fallback path does need to worry about that in case we have a repeating gradient that would have normally been aligned.

webrender/src/prim_store.rs
```diff
@@ -257,15 +252,119 @@ pub struct RadialGradientPrimitiveGpu {
...
+        } else {
+            for src in src_stops {
+                cur_idx = self.fill_colors(cur_idx, Self::get_index(src.offset),
+                                           &cur_color, &src.color);
+                cur_color = src.color;
+            }
+        }
+
+        // Fill out any remaining entries in the gradient.
+        self.fill_colors(cur_idx, 2 * GRADIENT_DATA_RESOLUTION, &cur_color, &cur_color);
```
lsalzman: Nope, the inputs are already reversed before they get to fill_colors, so that everything from 0..N has already been filled. It's only potentially from N..end that we need to fill.

webrender/res/ps_angle_gradient.fs.glsl
lsalzman: This is intentional, as the texture is clamped, so this will just stop at the last color in the gradient.

webrender/res/ps_radial_gradient.fs.glsl
lsalzman: I prefer to keep this the lowest spec feature that will implement it, as it is the most likely to perform well and be bug-free.

webrender/src/internal_types.rs
lsalzman: That is the format that is used for ImageFormat::RGBA8 texture uploads. Glenn and I agreed to encapsulate this in a PackedTexel type that will at least abstract this detail a bit better rather than forcing the caller to know too much about it.

webrender/src/tiling.rs
lsalzman: I will add a comment to this effect to make it clearer.

webrender/res/ps_angle_gradient.fs.glsl
kvark: Ok, but in this case you assume that `vGradientScale == textureSize(sGradients, 0).x / 2.0`. Why do we need the `vGradientScale` then?

webrender/src/prim_store.rs
kvark: ok

webrender/res/ps_angle_gradient.fs.glsl
lsalzman: I could just remove that and base it on the textureSize since that will address your concern about gpu_data_count as well.

kvark: Sounds good!

webrender/res/ps_radial_gradient.fs.glsl
kvark: Sure, but the thing is - we are already relying on the texture arrays to work good. So it's not like this code would impose any further limitations.

kvark:
ugh, github review comments are aweful
/me makes a note to switch to using Reviewable

glennw:
This looks good to me. I've run all the CSS tests on Servo. Needs a rebase (and the shader validation fix on CI if that remains), then this should be ready to merge.

PR #793, opened 2017-01-27
Add an optimization for drawing simple borders.

glennw:
This is primarily to speed up the time it takes to draw
large, simple borders on OSMesa. This makes some tests
run a lot faster when using software rendering, which
reduces the chances of timeouts occurring due to the
large number of tests that the Servo CI machines run
in parallel.

It's also a reasonable optimization for simple borders
when running a hardware renderer.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/793)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -2000,6 +2000,51 @@ impl FrameBuilder {
         let br_inner = br_outer - LayerPoint::new(radius.bottom_right.width.max(right.width),
                                                   radius.bottom_right.height.max(bottom.width));
 
+        // The border shader is quite expensive. For simple borders, we can just draw
+        // the border with a few rectangles. This generally gives better batching, and
+        // a GPU win in fragment shader time.
+        // More importantly, the software (OSMesa) implementation we run tests on is
+        // particularly slow at running our complex border shader, compared to the
+        // rectangle shader. This has the effect of making some of our tests time
+        // out more often on CI (the actual cause is simply too many Servo processes and
+        // threads being run on CI at once).
+        // TODO(gw): Detect some more simple cases and handle those with simpler shaders too.
+        // TODO(gw): Consider whether it's only worth doing this for large rectangles (since
+        //           it takes a little more CPU time to handle multiple rectangles compared
+        //           to a single border primitive).
+        if left.style == BorderStyle::Solid {
+            let same_color = left_color == top_color &&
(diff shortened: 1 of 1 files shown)


glennw:
r? @kvark 

This should allow us to finally land the pending WR update!

glennw:
Before this change:

```
time ./mach test-css --release --processes=8

real	21m18.107s
user	132m27.628s
sys	20m24.080s
```

After this change:

```
time ./mach test-css --release --processes=8

real	18m35.054s
user	113m37.369s
sys	20m32.638s
```

So this improves the time to run the CSS WPT tests by around 12.5%.

webrender/src/tiling.rs
```diff
@@ -2000,6 +2000,66 @@ impl FrameBuilder {
...
+        //           it takes a little more CPU time to handle multiple rectangles compared
+        //           to a single border primitive).
+        if left.style == BorderStyle::Solid {
+            let same_color = left_color == top_color &&
+                             left_color == right_color &&
+                             left_color == bottom_color;
+            let same_style = left.style == top.style &&
+                             left.style == right.style &&
+                             left.style == bottom.style;
+            let zero_radius = radius.top_left == LayerSize::zero() &&
```
kvark: could do `radius.is_zero()`

webrender/src/tiling.rs
```diff
@@ -2000,6 +2000,66 @@ impl FrameBuilder {
...
+                let left_rect = LayerRect::new(LayerPoint::new(tl_outer.x, tl_inner.y),
+                                               LayerSize::new(left.width,
+                                                              rect.size.height - top.width - bottom.width));
+                let right_rect = LayerRect::new(tr_inner,
+                                                LayerSize::new(right.width,
+                                                               rect.size.height - top.width - bottom.width));
+                let bottom_rect = LayerRect::new(LayerPoint::new(bl_outer.x, bl_inner.y),
+                                                 LayerSize::new(rect.size.width, bottom.width));
+
+                self.add_solid_rectangle(&top_rect,
```
kvark: let's iterate here over `[top_rect, left_rect, right_rect, bottom_rect]` given that the only other changing part here is the color, and we have already ensured that it matches for all 4.

glennw:
@kvark Thanks for the review - fixed those items.

kvark:
Thanks!
@bors-servo r+

PR #794, opened 2017-01-27
Add initial tidy support

mrobinson:
Currently many directories are ignored, but we can add them as we tidy
things up

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/794)

### DIFF
--- a/servo-tidy.toml
+++ b/servo-tidy.toml
@@ -1,0 +1,23 @@
+[configs]
+skip-check-length = false
+skip-check-licenses = false
+
+[ignore]
+
+# Files that are ignored for all tidy and lint checks.
+files = [
+    # servo-tidy does not properly support workspaces yet.
+    "./Cargo.toml",
+    # There are many duplicated packages, which we probably need to review
+    # one by one.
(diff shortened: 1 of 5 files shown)


servo-tidy.toml
```diff
@@ -0,0 +1,24 @@
...
+    # There are many duplicated packages, which we probably need to review
+    # one by one.
+    "./Cargo.lock"
+]
+
+# Many directories are currently ignored while we tidy things up
+# gradually.
+directories = [
+   # Generated and upstream code combined with our own. Could use cleanup
+  "./target",
```
kvark: why do we need to tidy up the generated code?

mrobinson: The comment is indeed a little unclear, but this is a list of directories excluded from the tidy. We will gradually remove entries.

glennw:
@mrobinson Needs a rebase. Also, the removal of replay/ dir has landed now - so we can remove that from the config file.

mrobinson:
@glennw Thanks for the review.

PR #795, opened 2017-01-27
Rasterize CG glyphs with subpixel text positioning

changm:
Also updates to generate glyph metrics like Skia does.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/795)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -15,7 +15,8 @@ use core_text::font_descriptor::kCTFontDefaultOrientation;
 use core_text;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
-use webrender_traits::{ColorU, FontKey, FontRenderMode, GlyphDimensions, GlyphOptions};
+use webrender_traits::{ColorU, FontKey, FontRenderMode, GlyphDimensions};
+use webrender_traits::{GlyphOptions, SubpixelOffset};
 use gamma_lut::{GammaLut, Color as ColorLut};
 
 pub struct FontContext {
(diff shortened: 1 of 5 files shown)


webrender/src/platform/macos/font.rs
```diff
@@ -69,23 +70,47 @@ fn supports_subpixel_aa() -> bool {
...
-    let rasterized_height = (rasterized_descent + rasterized_ascent) as u32;
-
-    GlyphMetrics {
-        rasterized_ascent: rasterized_ascent,
-        rasterized_descent: rasterized_descent,
-        rasterized_left: rasterized_left,
-        rasterized_width: rasterized_width,
-        rasterized_height: rasterized_height,
-    }
+    let x_offset :f64 = SubpixelOffset::into(x_subpixel);
```
kvark: I think you should be able to just do `let x_offset = f64::from(x_subpixel);`, or at least `x_subpixel.into()`

webrender/src/platform/macos/font.rs
```diff
@@ -69,23 +70,47 @@ fn supports_subpixel_aa() -> bool {
     data[0] != data[1] || data[1] != data[2]
 }
 
-fn get_glyph_metrics(ct_font: &CTFont, glyph: CGGlyph) -> GlyphMetrics {
+fn get_glyph_metrics(ct_font: &CTFont,
+                     glyph: CGGlyph,
+                     x_subpixel: SubpixelOffset,
```
kvark: it looks like the pair of x/y sub-pixel offsets is repeated often, so it's begging to be wrapped into a struct. I don't see an appropriate `euclid` type, so just declaring a new struct should be fine, since you aren't going to do any operations with the struct itself anyway.

changm: I was actually thinking of rewriting this and passing in GlyphKeys instead to both get_glyph_metrics and rasterize_glyphs. What do you think about that?

webrender/src/platform/macos/font.rs
changm: TIL!

changm:
Fixed the other things. Thanks!

webrender/src/platform/macos/font.rs
kvark: I think, if you already have `GlyphKey` in those spots (as opposed to need to create them on the stack), it should be fine to pass them directly.

kvark:
If you can address the repeating x/y of the subpixel offset passed around, that would be great!
Otherwise, let me know, and we'll proceed with what we have now.

changm:
I'd like to land it for now since some follow up patches rely on it. Then I'll clean it up in one patch. Will that work @kvark ?

webrender/src/platform/macos/font.rs
changm: I'll do that after https://github.com/servo/webrender/pull/798#issuecomment-275822832 lands. I got a couple of things to rewrite but I'd like to just land some PRs first.

glennw:
Needs rebasing.

changm:
Rebased.

glennw:
@bors-servo r-

Sorry, just want to confirm with @changm / @kvark  that this passes the Servo reftests before merging.

changm:
This seems to be passing the servo reftests for me locally.

Issue #788, opened 2017-01-26
device.rs code conditional on ARM should probably not be


jrmuizel:
There are a bunch of places in device.rs where we have things like
cfg!(any(target_arch="arm", target_arch="aarch64"))

arm seems like the wrong thing for these to be testing for. I suspect that these should actually be tests for opengles.

jrmuizel:
e.g. in gecko we use GL_LUMINANCE on all platforms instead of GL_RED or GL_ALPHA

Issue #771, opened 2017-01-24
Support run-time selection between GL and GLES


sotaroikeda:
Linked issue - servo/gleam#99

glennw:
This happened!

PR #796, opened 2017-01-27
Replaces some code conditionals on ARM by gl::GlType::default()

sotaroikeda:
preliminary patch for #771 and related to #788.
The patch replaces some code conditionals on ARM by gl::GlType::default().

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/796)


kvark:
This is on hold until we get the struct generator with different `gl` and `gles` structs.

kvark:
The code needs to be rewritten mostly. @sotaroikeda please open another PR with the new code.

PR #797, opened 2017-01-27
Move subpixel_quantize_offset into render mode

changm:
@kvark Implementing some of the review feedback from https://github.com/servo/webrender/pull/780#pullrequestreview-18877347

Not sure how you feel about the f32 instead of euclid units for GlyphInstances though.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/797)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -719,7 +719,7 @@ impl PrimitiveStore {
     pub fn resolve_primitives(&mut self,
                               resource_cache: &ResourceCache,
                               device_pixel_ratio: f32,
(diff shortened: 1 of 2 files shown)


kvark:
Thanks @changm ! I think we can just introduce a new `GlyphSpace` to parametrize `euclid`'s types with it. Let me know if you want to do this within this PR or the follow-ups.

changm:
@kvark Can we do it in another PR? I got a bunch of others that depend on this code. I'd rather do it in one swoop along dwrite/cg fonts

PR #798, opened 2017-01-27
Enable subpixel text for dwrite fonts

changm:
Depends on https://github.com/servo/webrender/pull/795#issuecomment-275777205 

The only real changes here are in webrender/src/platform/windows/font.rs

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/798)

### DIFF
--- a/webrender/src/platform/windows/font.rs
+++ b/webrender/src/platform/windows/font.rs
@@ -4,8 +4,8 @@
 
 use app_units::Au;
 use std::collections::HashMap;
-use webrender_traits::{FontKey, ColorU, FontRenderMode, GlyphDimensions, GlyphOptions};
-use webrender_traits::{SubpixelOffset};
+use webrender_traits::{FontKey, ColorU, FontRenderMode, GlyphDimensions};
+use webrender_traits::{GlyphOptions, SubpixelOffset};
 use gamma_lut::{GammaLut, Color as ColorLut};
 
 use dwrote;
@@ -159,6 +159,8 @@ impl FontContext {
     fn create_glyph_analysis(&self, font_key: FontKey,
                             size: Au, glyph: u32,
                             render_mode: FontRenderMode,
+                            x_suboffset: SubpixelOffset,
+                            y_suboffset: SubpixelOffset,
                             options: Option<GlyphOptions>) ->
                             dwrote::GlyphRunAnalysis {
         let face = self.fonts.get(&font_key).unwrap();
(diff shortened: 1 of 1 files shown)


vvuk:
Why is the subpixel offset not part of GlyphOptions? That's already part of the hash key, no?

changm:
It's part of the GlyphKey but not the GlyphOptions. GlyphOptions are force gdi rendering and embedded bitmaps.

vvuk:
But that stuff changes the rendering, so it should affect how the glyph is
cached no? Or are those constant per cache?

changm:
Ahh yeah you're right. It should be part of the GlyphKey. I'll fix that up in a follow up?

glennw:
Needs rebasing.

changm:
Rebased.

changm:
@bors-servo retry 

I think this is actually passing. I see this same travis-ci error in a couple of PRs and this PR didn't touch wrench at all. It doesn't even seem to run on windows?

glennw:
Yup, that CI failure is an intermittent which should be fixed by #807

glennw:
@vvuk Could you take a look over this?

webrender/src/platform/windows/font.rs
```diff
@@ -186,9 +188,14 @@ impl FontContext {
                                                     dwrite_measure_mode,
                                                     options);
 
-        // XX use the xform to handle subpixel positioning (what skia does), I believe that keeps
-        //let xform = dwrote::DWRITE_MATRIX { m11: 1.0, m12: 0.0, m21: 0.0, m22: 1.0, dx: 0.0, dy: 0.0 };
-        dwrote::GlyphRunAnalysis::create(&glyph_run, 1.0, None,
+        let x_offset: f32 = Into::<f64>::into(x_suboffset) as f32;
```
glennw: This seems a bit over complicated - perhaps convert it to a local f64 and then cast to f32 when creating the matrix or something like that. Can be done as a follow up anyway, it's not major.

glennw:
@bors-servo r-
 * travis issues

glennw:
I have no idea what's going on with bors / travis etc, but maybe this PR is what's messing it up. I finally convinced travis to try and build this PR, but it fails saying that the git SHA is not a valid reference. Want to close this, and re-open as a new PR and I'll merge it manually first thing in the morning?

glennw:
Merging manually due to bors problems, travis passed.

PR #799, opened 2017-01-27
Make some of the public types more ffi-friendly.

nical:
It would make gecko integration a lot easier if some of the types were ```#[repr(C)]```. This patch takes an opinionated stance of making everything that can be exposed to the ffi boundary repr C with an explicit size and explicit variant values for C-style enums.

I know that not all of this is strictly necessary, but when dealing with weird bugs close to the ffi boundary it is very reassuring to be able to look at the type definitions and be certain of their memory layouts and values without having to double-check the rules for both languages.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/799)

### DIFF
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -64,6 +64,7 @@ pub enum ApiMsg {
 }
 
 /// An opaque pointer-sized value.
+#[repr(C)]
 #[derive(Clone, Deserialize, Serialize)]
 pub struct ExternalEvent {
     raw: usize,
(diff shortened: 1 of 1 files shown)


webrender_traits/src/types.rs
```diff
@@ -332,12 +345,14 @@ impl FontRenderMode {
     }
 }
 
+#[repr(C)]
```
emilio: you only need `repr(u32)` here and in all the other enums.

glennw:
Makes perfect sense to me - I'm fine with merging this once emilio's comment is addressed.

glennw:
Let's get this merged and tidy up the repr(C) on the enums as a follow up.

PR #800, opened 2017-01-28
Fix currently scrolling layer id  caching

gterzian:
Fix for https://github.com/servo/servo/issues/15251

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/800)

### DIFF
--- a/webrender/src/scroll_tree.rs
+++ b/webrender/src/scroll_tree.rs
@@ -194,7 +194,16 @@ impl ScrollTree {
                 self.current_scroll_layer_id = Some(scroll_layer_id);
                 scroll_layer_id
             },
-            (_, _, Some(scroll_layer_id)) => scroll_layer_id,
+            (_, scroll_layer_id, Some(cached_scroll_layer_id)) => {
+                let scroll_layer_id = match self.layers.get(&cached_scroll_layer_id) {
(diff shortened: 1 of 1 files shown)


gterzian:
@jdm So I am still unable to reproduce the panic. If you can do so consistently, could I please ask you to test if this fixes it? This one requires building with https://github.com/servo/servo/pull/15111

glennw:
This seems OK to me - but let's double check with @mrobinson first.

gterzian:
@glennw Thank you. I changed the variables names a bit for clarity...

mrobinson:
I think that we will need to change the API in order to let WebRender know when a batch of scroll layers doesn't have any correspondence with the previous one. I need that for a PR that I'm working on, but a decent fallback here makes sense as well. Thanks!

mrobinson:
The new API is here: https://github.com/servo/webrender/pull/806

This patch should work fine in tandem with that.

gterzian:
@glennw @mrobinson Thanks for your help, I will also take a look at the new API, and the scroll testing capabilities of Wrench. Sorry I didn't squash that one extra commit, I guess it's not the end of the world...

glennw:
No worries, thanks for the fix!

PR #802, opened 2017-01-29
Support multiple composite operations in a StackingContext

rlhunt:
A StackingContext holds a Vec<FilterOp> and a MixBlendMode, but currently only one operation is applied. This change modifies the render task building code to respect multiple composite operations. It's also careful to apply filters first and then the mix-blend-mode according to the spec.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/802)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -7,7 +7,7 @@ use batch_builder::BorderSideHelpers;
 use fnv::FnvHasher;
 use frame::FrameId;
 use gpu_store::GpuStoreAddress;
-use internal_types::{ANGLE_FLOAT_TO_FIXED, LowLevelFilterOp, CompositionOp};
+use internal_types::{ANGLE_FLOAT_TO_FIXED, LowLevelFilterOp};
 use internal_types::{BatchTextures, CacheTextureId, SourceTexture};
 use mask_cache::{ClipSource, MaskCacheInfo};
 use prim_store::{PrimitiveGeometry, RectanglePrimitive, PrimitiveContainer};
@@ -1679,7 +1679,7 @@ pub struct StackingContext {
     local_rect: LayerRect,
     scroll_layer_id: ScrollLayerId,
     xf_rect: Option<TransformedRect>,
-    composite_kind: CompositeKind,
+    composite_ops: CompositeOps,
     clip_source: ClipSource,
     clip_cache_info: Option<MaskCacheInfo>,
 }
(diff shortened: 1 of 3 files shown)


rlhunt:
I've tested this on some basic filter and mix-blend-mode combinations and it seems to work. I used wrench plus the changes in #801.

glennw:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[webrender/src/tiling.rs, line 1880 at r1](https://reviewable.io:443/reviews/servo/webrender/802#-KbhyiPl6AA2lCub4g6B:-KbhyiPmhn3Z8chiPf0i:bukmih6) ([raw file](https://github.com/servo/webrender/blob/d8671166133014dfa57e7350294d24ca51f56b00/webrender/src/tiling.rs#L1880)):*
> ```Rust
>                       pipeline_id: PipelineId,
>                       scroll_layer_id: ScrollLayerId,
>                       composite_ops: &CompositeOps) {
> ```

Let's remove the & and move the CompositeOps here - that will avoid the clone() below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/802)*

glennw:
This looks great! Just one minor issue above.

rlhunt:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[webrender/src/tiling.rs, line 1880 at r1](https://reviewable.io:443/reviews/servo/webrender/802#-KbhyiPl6AA2lCub4g6B:-Kbl8P8s8a0dp6m2OMR3:bzf5p81) ([raw file](https://github.com/servo/webrender/blob/d8671166133014dfa57e7350294d24ca51f56b00/webrender/src/tiling.rs#L1880)):*

Let's remove the & and move the CompositeOps here - that will avoid the clone() below.
</blockquote></details>

The &[CompositionOp] was passed into push_layer twice if the stacking context was root with a background color. But I don't think that extra layer needs CompositeOps, so I've removed the & and give that layer CompositeOps::empty().

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/802)*

webrender/src/tiling.rs
```diff
@@ -2594,18 +2592,16 @@ impl FrameBuilder {
...
-                            let layer_rect = layer.xf_rect.as_ref().unwrap().bounding_rect;
-                            let location = RenderTaskLocation::Dynamic(None, layer_rect.size);
-                            let new_task = RenderTask::new_alpha_batch(next_task_index,
-                                                                       layer_rect.origin,
-                                                                       location);
-                            next_task_index.0 += 1;
-                            let prev_task = mem::replace(&mut current_task, new_task);
-                            alpha_task_stack.push(prev_task);
-                        }
+                    let composite_count = layer.composite_ops.count();
```
kvark: why do you need to separate task creation from task initialization (which is happening in `PopStackingContext`)? Could we have all the logic in one place (in the pop code) instead?

webrender/src/tiling.rs
```diff
@@ -2616,28 +2612,25 @@ impl FrameBuilder {
...
-                            next_z += 1;
-                            prev_task.as_alpha_batch().alpha_items.push(item);
-                            prev_task.children.push(current_task);
-                            prev_task.children.push(readback_task);
-                            current_task = prev_task;
-                        }
+                    for filter in &layer.composite_ops.filters {
+                        let mut prev_task = alpha_task_stack.pop().unwrap();
+                        let item = AlphaRenderItem::Blend(sc_index, current_task.id, *filter, next_z);
+                        next_z += 1;
```
kvark: I don't think changing the Z makes sense here, given that we are rendering into the task cache and that's not supposed to use the depth test anyway.

glennw: The Z is used here when drawing the blend item into the main scene.

glennw:
@rlhunt Yup, you're right about the extra layer - your change seems correct to me. r=me once the question from @kvark is resolved.

glennw:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/src/tiling.rs, line 2595 at r2](https://reviewable.io:443/reviews/servo/webrender/802#-KbdTAGSpN1_SH_Ilgl4-r2-2595:-KbmzlHnI3L0hsy50O8v:b-9s7sek) ([raw file](https://github.com/servo/webrender/blob/425bfecbace8c57bdce3c81fd3ab651f69bd0f0a/webrender/src/tiling.rs#L2595)):*

why do you need to separate task creation from task initialization (which is happening in `PopStackingContext`)? Could we have all the logic in one place (in the pop code) instead?
</blockquote></details>

The tasks are created here and pushed onto a stack, so that the items in the primitive runs get pushed on to the right task. Or am I completely misunderstanding your comment? :) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/802)*

rlhunt:
So yes, basically what Glenn said. Tasks need to be created before the following primitive run so the primitives are added to them. And I'm not familiar with how the Z order code is used, but Glenn's comment makes sense.

I tried to preserve the original task creation logic. I only extended it so multiple filters on a single stacking context would act like multiple identical stacking contexts with a single filter.

rlhunt:
Hmm, so I rebased my branch on the latest master. Not sure why that isn't showing up here.

glennw:
I'm not sure why appveyor is confused either. Could you try rebasing again, or if it's already up to date, try do just run ```git commit --amend``` which will modify the timestamp and commit SHA, and then force push?

rlhunt:
Okay, that did the trick!

glennw:
Merging manually - homu is not listening, but travis has passed on both platforms.

PR #803, opened 2017-01-30
Fix a regression in text rendering where non-renderable glyphs are present.

glennw:
The text run code in prepare_prim_for_render() takes account of
non-renderable glyphs and pushes only the renderable glyphs into
the list of glyphs for the text run.

The recent subpixel positioning changes stopped using this and were
using the source glyph indices. This can result in incorrect indexing
of glyph information in the text run shader.

This patch modifies the text run request and render code to use
the glyph indices array. It extends it to contain the x/y coords
of the glyph, allowing the subpixel positioning to work. It also
tidies up some of the code so that the resolve step no longer
requires access to the stacking context and aux maps.

Fixes several of the WPT ref tests that use whitespace: pre.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/803)

### DIFF
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -133,10 +133,10 @@ impl FontContext {
     pub fn rasterize_glyph(&mut self,
                            font_key: FontKey,
                            size: Au,
-                           color: ColorU,
+                           _color: ColorU,
                            character: u32,
                            render_mode: FontRenderMode,
-                           glyph_options: Option<GlyphOptions>) -> Option<RasterizedGlyph> {
+                           _glyph_options: Option<GlyphOptions>) -> Option<RasterizedGlyph> {
         let mut glyph = None;
 
         if let Some(slot) = self.load_glyph(font_key,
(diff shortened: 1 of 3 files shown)


glennw:
r? @pcwalton  or @kvark 

cc @changm

emilio:
Stealing review as requested by @glennw. I went through this fix on IRC with him. He actually has addressed them before I managed to comment :)

@bors-servo r+

```
2:21 <@gw> emilio: yep, the glyph indices one
02:24 <emilio> gw: so the key change apart of including the x and y pixel values is operating directly on glyph_indices instead of src_glyphs, that could contain non-renderable glyphs, is that right?
02:25 <@gw> emilio: yep, that's right
02:26 <emilio> gw: one nit: since the glyph indices are no longer glyph indices, perhaps the field should be renamed to renderable_glyphs, to make the difference clear?
02:28 <emilio> gw: (sorry, still looking through all the context to be able to sign this off)
02:28 <@gw> emilio: yup, all good - pushed an update renaming them to glyph_instances
02:29 <emilio> gw: the other change is that we no longer have to do the double lookup in the auxiliary lists since now we propagate all the info down to that field, right?
02:29 <@gw> emilio: yep - this is how it used to be before the subpixel positioning PR landed
02:31 <emilio> gw: and the rest is just a bit of unused args cleanup and similar no?
02:31 <@gw> emilio: yep
```

PR #804, opened 2017-01-30
Fix is_uniform() method on border radius.

glennw:
This was breaking the CSS border tests in one case, as a result
of this method being used by the simple border optimization
that recently landed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/804)

### DIFF
--- a/webrender_traits/src/display_item.rs
+++ b/webrender_traits/src/display_item.rs
@@ -49,12 +49,12 @@ impl BorderRadius {
         }
     }
 
-    pub fn is_uniform(&self) -> Option<f32> {
-        let uniform_radius = LayoutSize::new(self.top_left.width, self.top_left.width);
+    pub fn is_uniform(&self) -> Option<LayoutSize> {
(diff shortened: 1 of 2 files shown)


webrender_traits/src/display_item.rs
```diff
@@ -49,20 +49,20 @@ impl BorderRadius {
         }
     }
 
-    pub fn is_uniform(&self) -> Option<f32> {
-        let uniform_radius = LayoutSize::new(self.top_left.width, self.top_left.width);
+    pub fn is_uniform(&self) -> Option<LayoutSize> {
+        let uniform_radius = LayoutSize::new(self.top_left.width, self.top_left.height);
```
cbrewster: nit: Could this just be `let uniform_radius = self.top_left;`? Or just compare with `self.top_left` in the if statement. If you prefer this way, thats fine.

glennw: Yep, fixed.

PR #806, opened 2017-01-30
Add scrolling support to wrench

mrobinson:
This PR includes two commits:
- The first modifies the API of WebRender to instruct it to throw away scrolling offsets during changes to unrelated display lists. In the future this will need to be more complicated to handle back and forward navigation.
- The second adds scroll layers and scroll-offset to wrench itself in order to allow tests of scrolling behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/806)


mrobinson:
cc @jrmuizel @glennw @gterzian

mrobinson:
Looks like I inadvertently broke the sample app. I've fixed that now. @jrmuizel, is it okay to send this to bors again?

mrobinson:
One more quick update to the PR that properly supports writing scroll-policy in the YAML writer.

mrobinson:
The changes from the originally reviewed patch were very minor, so I feel pretty confident landing this to avoid timezone induced +1 day lag.

PR #812, opened 2017-01-30
Use a newer rustc on CI, to possibly workaround a rustc compiler bug.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/812)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
   - linux
   - osx
 rust:
-  - nightly-2016-12-19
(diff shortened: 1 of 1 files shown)


jrmuizel:
FWIW, it would probably be good to be using same version of rustc that we use in the gecko CI

glennw:
@jrmuizel Agreed - where can I find out that information?

glennw:
We'll push this as a commit in the PR that is causing issues, so can close this.

PR #814, opened 2017-01-31
Remove redundant #[repr(C)] when #[repr(u32)] is present.

nical:
Followup to PR #799.

r? @emilio

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/814)


emilio:
@bors-servo r+

Thanks for doing this :)

PR #817, opened 2017-02-01
Upgrade image and byteorder; update pinned dependencies

mbrubeck:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/817)


glennw:
Merging manually due to bors weirdness.

PR #818, opened 2017-02-01
Remove image borders

### LINKED ISSUE #789: Avoid bordering images in the cache
WIP branch - https://github.com/nical/webrender/commit/1b905fefc3d1df8c2124a8b3d133a248603ec748
TODO:
  - rebase
  - test the performance on some case that is full of images

kvark:
Fixes #789 

Clamping the texture coordinates in pixel shaders (as well as using textureLod) to ensure sampling within original borders now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/818)

### DIFF
--- a/webrender/res/ps_image.fs.glsl
+++ b/webrender/res/ps_image.fs.glsl
@@ -24,8 +24,13 @@ void main(void) {
     // account the spacing in between tiles. We only paint if our fragment does
     // not fall into that spacing.
     vec2 position_in_tile = mod(relative_pos_in_rect, vStretchSize + vTileSpacing);
-    vec2 st = vTextureOffset + ((position_in_tile / vStretchSize) * vTextureSize);
+    // We clamp the texture coordinates to the half-pixel offset from the borders
+    // in order to avoid sampling outside of the texture area.
+    vec2 st = vTextureOffset + clamp(
+        (position_in_tile / vStretchSize) * vTextureSize,
+        vHalfTexel, vTextureSize - vHalfTexel);
+
     alpha = alpha * float(all(bvec2(step(position_in_tile, vStretchSize))));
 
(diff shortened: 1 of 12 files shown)


kvark:
r? @glennw 

BTW, I removed the distinction between `requested_size` and `allocated_size` now, following the general idea of relying on git history to restore the obsolete stuff if needed.

glennw:
This generally looks good to me (I'll run the CSS tests locally to confirm). We'll need to apply the same clamp in the text run shader though.

kvark:
Good point @glennw , the text is now also converted.

glennw:
This looks good, but will need to be rebased after the external texture images changes landed.

kvark:
@glennw rebased now

glennw:
Nice one!

JerryShih:
@kvark @nical 
Are we use mipmap now? I don't see any mipmap generating in WR. So, I'm curious about the textureLod usage here.

kvark:
@JerryShih there are 2 reasons for using textureLod in this case:
  1. It clearly states the pre-condition. Since the image border code uses a half-texel offset, it's not going to work properly for any Lod > 0.0. So now if you add any mipmap levels, the code doesn't care and will still work properly.
  2. It may be slightly faster. The shader compiler doesn't know if a texture has mipmap levels or not (unless it re-compiles the shader on the fly), so it will generate the gradient instructions and a regular sample, only to be dismissed by the texture sampler that discovers no mipmap levels. Using `textureLod` directly produces the efficient HW shader assembly right away.

leeoniya:
@kvark maybe worth adding these architectural decisions to the wiki? probably makes sense for any questions that can answer "why not use <x>?" type questions for common graphics tech.

kvark:
@leeoniya that should be a good start:
https://github.com/servo/webrender/wiki/Graphics-Features

leeoniya:
excellent, thanks 👍

PR #821, opened 2017-02-01
Cache cargo

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/821)


jrmuizel:
This is garbage

PR #823, opened 2017-02-02
Fix the panic when transform is non-invertible

### LINKED ISSUE #418: Panic when a stacking context has an invalid transform.
If a transform matrix cannot be inverted, the webrender code will currently panic. Instead, it should detect this and not try to draw the stacking context.

canaltinova:
Fix the panic when transform is non-invertible. It also marks the `combined_local_viewport_rect` as not visible.
Fixes #418.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/823)


glennw:
Merging manually since bors won't.

Issue #742, opened 2017-01-18
Separate stacking context and scroll root concepts.


glennw:
The eventual solution may be quite different to the title of this bug - but we can use this as a tracking bug anyway.

glennw:
cc @mrobinson @kvark @nical @jrmuizel

glennw:
See https://github.com/servo/servo/issues/13530

mrobinson:
I think this is complete now.

glennw:
\o/

PR #824, opened 2017-02-02
Clarify the difference between packed layers and stacking contexts

mrobinson:
Eventually scroll layers will create packed layers as well, so this will
help to keep things clear. This shouldn't change behavior, because
currently packed layer ids will always be equal to the stacking context
ids of their owners.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/824)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -321,12 +321,12 @@ impl Frame {
         let layer_rect = LayerRect::new(LayerPoint::zero(),
                                         LayerSize::new(content_size.width + clip.origin.x,
                                                        content_size.height + clip.origin.y));
-        context.builder.push_layer(layer_rect,
-                                   &ClipRegion::simple(&layer_rect),
-                                   LayerToScrollTransform::identity(),
-                                   pipeline_id,
-                                   current_scroll_layer_id,
-                                   CompositeOps::empty());
+        context.builder.push_stacking_context(layer_rect,
+                                              &ClipRegion::simple(&layer_rect),
+                                              LayerToScrollTransform::identity(),
+                                              pipeline_id,
+                                              current_scroll_layer_id,
+                                              CompositeOps::empty());
 
         self.flatten_items(traversal,
                            pipeline_id,
@@ -336,7 +336,7 @@ impl Frame {
                            LayerToScrollTransform::identity(),
                            level);
 
-        context.builder.pop_layer();
+        context.builder.pop_stacking_context();
     }
 
     fn flatten_stacking_context<'a>(&mut self,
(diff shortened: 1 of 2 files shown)


mrobinson:
@glennw r?

kvark:
> currently packed layer ids will always be equal to the stacking context
ids of their owners

Let's hope all of the are called `sc_index` so that we don't miss any ;)

webrender/src/frame.rs
```diff
@@ -340,12 +340,12 @@ impl Frame {
...
                                                        content_size.height + clip.origin.y));
-        context.builder.push_layer(layer_rect,
-                                   &ClipRegion::simple(&layer_rect),
-                                   LayerToScrollTransform::identity(),
-                                   pipeline_id,
-                                   current_scroll_layer_id,
-                                   &[]);
+        context.builder.push_stacking_context(layer_rect,
+                                              &ClipRegion::simple(&layer_rect),
+                                              LayerToScrollTransform::identity(),
```
kvark: Not directly related to this PR, but is it correct to still call it `LayerToScrollTransform`?
AFAIK, the SC's/items are attached to their reference frames, but the transformation we pass here is relative to the parent stacking context (not necessarily the scroll layer).

mrobinson: @kvark, I think you are right. The naming of this is off now. There is a chance that in the future StackingContexts won't have transforms at all. It depends if we decide to make all coordinates relative to reference frames or not.

mrobinson:
@kvark, thanks for looking at this. I've pushed a new rebased version with a few other minor naming changes and a new use of Default::default().

webrender/src/frame.rs
kvark: Yep, I'd say let's not have the transformations in there. Since they only affect the *initial* placement of items, we should use it and forget about them, only keeping the reference frames in our sight.

webrender/src/tiling.rs
```diff
@@ -1682,20 +1702,21 @@ pub struct StackingContext {
...
     transform: LayerToWorldTransform,
     inv_transform: WorldToLayerTransform,
     local_clip_rect: LayerRect,
     screen_vertices: [WorldPoint4D; 4],
 }
 
-impl Default for PackedStackingContext {
-    fn default() -> PackedStackingContext {
-        PackedStackingContext {
+impl Default for PackedLayer {
```
kvark: I believe the semantics of `Default` is misused here. It's not a "default" layer, it's an empty one. `Default` is applicable where you could drop it if you don't care, and it's going to somewhat work. This is not the case here.

mrobinson: Hrm. Okay. I'm happy to change it, but does that mean that the implementation of Default::default should change.

kvark: Since there was already `impl Default for PackedStackingContext`, this is not something that has to be done in your PR. But for the future I think we should remove the `impl Default` and just have the same thing as an `fn empty() -> PackedLayer` constructor. No implementation change needed.

mrobinson:
The build failure seems to be

```
Could not open dynamic library `OsMesa`
```

glennw:
I *think* the build failure should be resolved now (I reverted the commit that seems to cause that) - but travis might need a rebuild?

mrobinson:
This is related to #742, by the way.

PR #825, opened 2017-02-02
Make more types repr C.

nical:
r? @emilio

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/825)


glennw:
Merging manually since bors is apparently still not merging things?

PR #826, opened 2017-02-02
Update to core-graphics 0.6 and core-text 3.0

mbrubeck:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/826)


mbrubeck:
Added version bumps.  This is a breaking change because CTFont is used in public APIs.

glennw:
Merging manually - homu not attached to this PR at all?

PR #827, opened 2017-02-03
Revert "Cache .cargo on travis"

glennw:
This reverts commit e1a33cd3758aa17e160b48f44bb737d4f8b76d75.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/827)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ os:
   - osx
 rust:
   - nightly-2017-01-26
-cache: cargo
 notifications:
   webhooks: http://build.servo.org:54856/travis
 addons:


glennw:
This appears to break the test code that searches for OSMesa. Reverting until we can find a better solution.

jrmuizel:
Ah right. That makes some sense. I'll try to come up with a better solution.

jrmuizel:
I think the best thing might be to statically link with osmesa.

jrmuizel:
Though that might conflict with the system library so maybe that's not a great approach either.

PR #828, opened 2017-02-03
Have scroll layers clip instead of stacking contexts

mrobinson:
This is another step on the path to make clipping totally independent
of stacking contexts in order to have contents of the same stacking
contents clipped by different scrolling roots. As part of this change,
scrolling roots are integrated into the frame builder's primitive run
commands.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/828)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -282,6 +282,19 @@ impl Frame {
             let mut traversal = DisplayListTraversal::new_skipping_first(display_list);
             let reference_frame_id = self.scroll_tree.root_reference_frame_id();
             let topmost_scroll_layer_id = self.scroll_tree.topmost_scroll_layer_id();
+            debug_assert!(reference_frame_id != topmost_scroll_layer_id);
+
+            let viewport_rect = LayerRect::new(LayerPoint::zero(), root_pipeline.viewport_size);
+            let clip = ClipRegion::simple(&viewport_rect);
+            context.builder.push_scroll_layer(reference_frame_id,
+                                              &clip,
+                                              &LayerPoint::zero(),
+                                              &root_pipeline.viewport_size);
+            context.builder.push_scroll_layer(topmost_scroll_layer_id,
+                                              &clip,
+                                              &LayerPoint::zero(),
+                                              &root_clip.main.size);
+
             self.flatten_stacking_context(&mut traversal,
                                           root_pipeline_id,
                                           &mut context,
@@ -291,6 +304,9 @@ impl Frame {
                                           0,
                                           &root_stacking_context,
                                           root_clip);
+
+            context.builder.pop_scroll_layer();
+            context.builder.pop_scroll_layer();
         }
 
         self.frame_builder = Some(frame_builder);
@@ -302,10 +318,10 @@ impl Frame {
                                 pipeline_id: PipelineId,
                                 context: &mut FlattenContext,
                                 current_reference_frame_id: ScrollLayerId,
-                                mut current_scroll_layer_id: ScrollLayerId,
+                                parent_scroll_layer_id: ScrollLayerId,
                                 layer_relative_transform: LayerToScrollTransform,
                                 level: i32,
-                                clip: &LayerRect,
+                                clip: &ClipRegion,
                                 content_size: &LayerSize,
                                 new_scroll_layer_id: ScrollLayerId) {
         // Avoid doing unnecessary work for empty stacking contexts.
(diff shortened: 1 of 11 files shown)


mrobinson:
@kvark or @glennw r?

webrender/src/frame.rs
```diff
@@ -282,6 +282,19 @@ impl Frame {
...
             let topmost_scroll_layer_id = self.scroll_tree.topmost_scroll_layer_id();
+
+
+            let viewport_rect = LayerRect::new(LayerPoint::zero(), root_pipeline.viewport_size);
+            let clip = ClipRegion::simple(&viewport_rect);
+            context.builder.push_scroll_layer(reference_frame_id,
+                                              &clip,
+                                              &LayerPoint::zero(),
+                                              &root_pipeline.viewport_size);
+            context.builder.push_scroll_layer(topmost_scroll_layer_id,
```
kvark: Does it handle the case of `reference_frame_id == topmost_scroll_layer_id`?

webrender/src/frame.rs
```diff
@@ -494,6 +502,15 @@ impl Frame {
                                    iframe_scroll_layer_id,
                                    iframe_reference_frame_id);
 
+        context.builder.push_scroll_layer(iframe_reference_frame_id,
```
kvark: why are we pushing a scroll layer for a reference frame?

webrender/src/tiling.rs
```diff
@@ -1265,13 +1265,11 @@ impl RenderTask {
...
         if inner_rect.is_some() && clips.len() == 1 {
-            let (stacking_context_index, ref clip_info) = clips[0];
-            let stacking_context = &stacking_context_store[stacking_context_index.0];
-
-            if clip_info.image.is_none() && clip_info.effective_clip_count == 1 &&
-               stacking_context.xf_rect.as_ref().unwrap().kind == TransformedRectKind::AxisAligned {
+            let (_, ref clip_info) = clips[0];
+            if clip_info.image.is_none() &&
+               clip_info.effective_clip_count == 1 &&
+               clip_info.is_aligned {
```
kvark: sweet 👍

webrender/src/tiling.rs
```diff
@@ -2431,25 +2476,51 @@ impl FrameBuilder {
...
-                    if let Some(ref mut clip_info) = stacking_context.clip_cache_info {
-                        let auxiliary_lists = auxiliary_lists_map.get(&stacking_context.pipeline_id)
+                    if let Some(ref mut clip_info) = scroll_layer.clip_cache_info {
+                        let pipeline_id = scroll_layer.scroll_layer_id.pipeline_id;
+                        let auxiliary_lists = auxiliary_lists_map.get(&pipeline_id)
                                                                  .expect("No auxiliary lists?");
-                        clip_info.update(&stacking_context.clip_source,
-                                         &packed_layer.transform,
+                        clip_info.update(&scroll_layer.clip_source,
+                                         &packed_layer.transform, /// XXX Get tform here
```
kvark: what's wrong with `packed_layer.transform` here?

webrender/src/frame.rs
mrobinson: ScrollTree::establish_root should always create two separate layers for these, so this shouldn't happen. I can add an assertion here though.

webrender/src/frame.rs
mrobinson: I think this is one place where the current terminology doesn't quite work. Reference frames have a layer in the ScrollTree, even though they themselves don't scroll. This will allow primitives to be positioned relative to the reference frame though. In this case, I don't think it is absolutely necessary, but I think it's useful to avoid having consistency for the moment.

webrender/src/tiling.rs
mrobinson: Ack! I meant to remove this comment, but forgot. I'll update the PR.

webrender/src/frame.rs
kvark: would be great, yeah

webrender/src/frame.rs
kvark: > This will allow primitives to be positioned relative to the reference frame though

Please tell me that's on your list to address, since it doesn't sound like a good solution.

mrobinson: I think this is necessary to do, because for primitives that are direct children of a stacking context that establishes a reference frame, they should not get the scroll offset of the containing scroll root (which is already folded into the transformation of the reference frame). Basically a reference frame needs a packed layer in order to properly position its children. Glenn and are discussing conceptually collapsing reference frames and scroll layers into one concept, but we aren't entirely clear on the details yet.

mrobinson:
The latest version of this PR updates the versions and fixes the issues mentioned.

PR #829, opened 2017-02-03
Convert GlyphInstance to use euclid Point2D

changm:
Follow up from  https://github.com/servo/webrender/pull/780#discussion_r98240732

In one more follow up, I'll clean up the subpixel offset.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/829)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use app_units::Au;
-use euclid::Size2D;
+use euclid::{Point2D, Size2D};
 use gpu_store::GpuStoreAddress;
 use internal_types::{SourceTexture, PackedTexel};
 use mask_cache::{ClipSource, MaskCacheInfo};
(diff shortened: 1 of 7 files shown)


webrender/src/prim_store.rs
```diff
@@ -1068,15 +1068,14 @@ impl PrimitiveStore {
...
-                                                      src_glyphs[0].y,
+                                                      src_glyphs[0].point,
                                                       text.render_mode);
                     let mut local_rect = LayerRect::zero();
                     let mut actual_glyph_count = 0;
 
                     for src in src_glyphs {
                         glyph_key.index = src.index;
-                        glyph_key.set_subpixel_offset(src.x, src.y, text.render_mode);
+                        glyph_key.set_subpixel_offset(src.point.x, src.point.y, text.render_mode);
```
glennw: set_subpixel_offset() should probably just take a Point2D as parameter now?

changm: Yes. From the first comment, I was going to do that in a follow up to make each patch smaller.

wrench/src/yaml_frame_reader.rs
```diff
@@ -360,8 +360,8 @@ impl YamlFrameReader {
             let glyphs = glyph_indices.iter().enumerate().map(|k| {
                 GlyphInstance {
                     index: *k.1,
-                    x: origin.x + glyph_offsets[k.0*2],
-                    y: origin.y + glyph_offsets[k.0*2+1],
+                    point: Point2D::new(origin.x + glyph_offsets[k.0*2],
```
kvark: nit: could use euclid's point addition or `add_size`. See https://docs.rs/euclid/0.10.5/euclid/point/struct.TypedPoint2D.html

glennw:
@bors-servo r+

Sounds good, thanks!

PR #831, opened 2017-02-06
Fix webgl test regression in image shader clamping.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/831)

### DIFF
--- a/webrender/res/ps_image.vs.glsl
+++ b/webrender/res/ps_image.vs.glsl
@@ -36,5 +36,7 @@ void main(void) {
     vTextureOffset = st0;
     vTileSpacing = image.stretch_size_and_tile_spacing.zw;
     vStretchSize = image.stretch_size_and_tile_spacing.xy;
-    vHalfTexel = vec2(0.5) / texture_size;
+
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

Perhaps there's a simpler / cleaner way to achieve this in the shader you can think of? This fixes the regressions though - I've done a full WPT / CSS test run with these changes.

webrender/res/ps_image.vs.glsl
```diff
@@ -36,5 +36,7 @@ void main(void) {
     vTextureOffset = st0;
     vTileSpacing = image.stretch_size_and_tile_spacing.zw;
     vStretchSize = image.stretch_size_and_tile_spacing.xy;
-    vHalfTexel = vec2(0.5) / texture_size;
+
+    vec2 half_texel = vec2(0.5) / texture_size;
+    vStRect = vec4(min(st0, st1) + half_texel, max(st0, st1) - half_texel);
```
kvark: uh oh, so the input UV rectangle can be inversed?

PR #832, opened 2017-02-06
Implement an API that allows animating certain properties of an existing display list.

glennw:
Currently, this allows animating the transform and opacity of
a stacking context, but it can easily be extended in the future
to handle other property types (e.g. other filters etc).

The current list of property values are passed via the
GenerateFrame message, which ensures that all animated properties
are updated atomically.

The implementation right now is far from optimal. The idea here
is to get the public API correct and implemented, and then we can
focus on optimizing the implementation, once the in-progress work
on scroll layers and reference frames lands.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/832)

### DIFF
--- a/webrender/src/scene.rs
+++ b/webrender/src/scene.rs
@@ -7,7 +7,71 @@ use std::collections::HashMap;
 use std::hash::BuildHasherDefault;
 use tiling::AuxiliaryListsMap;
 use webrender_traits::{AuxiliaryLists, BuiltDisplayList, PipelineId, Epoch, ColorF};
-use webrender_traits::{DisplayItem, LayerSize};
+use webrender_traits::{DisplayItem, DynamicProperties, LayerSize, LayoutTransform};
+use webrender_traits::{PropertyBinding, PropertyBindingId};
+
+/// Stores a map of the animated property bindings for the current display list. These
+/// can be used to animate the transform and/or opacity of a display list without
+/// re-submitting the display list itself.
+pub struct SceneProperties {
+    transform_properties: HashMap<PropertyBindingId, LayoutTransform>,
+    float_properties: HashMap<PropertyBindingId, f32>,
+}
+
+impl SceneProperties {
+    pub fn new() -> SceneProperties {
+        SceneProperties {
+            transform_properties: HashMap::with_hasher(Default::default()),
+            float_properties: HashMap::with_hasher(Default::default()),
+        }
+    }
+
+    /// Set the current property list for this display list.
+    pub fn set_properties(&mut self, properties: DynamicProperties) {
+        self.transform_properties.clear();
+        self.float_properties.clear();
+
+        for property in properties.transforms {
+            self.transform_properties.insert(property.key.id, property.value);
+        }
+
+        for property in properties.floats {
+            self.float_properties.insert(property.key.id, property.value);
+        }
+    }
+
+    /// Get the current value for a transform property.
+    pub fn resolve_layout_transform(&self, property: &PropertyBinding<LayoutTransform>) -> LayoutTransform {
+        match *property {
+            PropertyBinding::Value(matrix) => matrix,
+            PropertyBinding::Binding(ref key) => {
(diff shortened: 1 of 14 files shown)


glennw:
Marked as [WIP] though for feedback on the API itself (the functionality itself is complete).

glennw:
cc @nical @kvark @jrmuizel

glennw:
This commit contains a sample of how the API is used by calling code - https://github.com/glennw/webrender/commit/988acbd3c7bf7a5c901e83ddaf567b6fdda8a839

kvark:
@glennw I think, while we could have more assumptions baked into compile-time (types) instead of run-time, the current approach is sensible.

kvark:
Btw, I don't think we need to pass `Option<>` in here:
```rust
api.generate_frame(Some(properties));
```
An empty vector doesn't allocate any memory, and iterating over it is cheap, so we might as well just receive a vector all the time (as opposed to `Option<Vec<_>>`).

glennw:
@kvark The idea of the option is to explicitly mark that there are no animated properties, rather than an empty vec, which could be a scene with animated properties where none of the properties are set.

The distinction is used at the moment to decide whether to rebuild the scene. I do agree, it's not entirely clear to me if it makes sense to have this in the API though.

kvark:
If we wanted better types, we could do something like this:
```rust
let animated_transform_key = api.generate_property_binding_key::<Matrix4>(); //type could probably be inferred from the usage.
let animated_opacity_key = api.generate_property_binding_key::<f32>();

...
// notice that we don't need `webrender_traits::*Property` wrappers
// and we can't mess up by passing a wrongly typed property
vec![ webrender_traits::FilterOp::Opacity(animated_opacity_key) ]);
 
...
// notice that types are not erased here
let properties = DynamicProperties {
    transform: vec![ animated_transform_key.with(transform) ],
    f32: vec![ animated_opacity_key.with(current_opacity) ],
];

**Comment by glennw** [comment]
@kvark Yea, I had pondered doing something like that. I think it probably makes sense, I'll update the PR today and see how it looks.

**Comment by howareyou322** [comment]
About the 'generate_frame' API, I prefer 'Option<Vec<_>>' because we don't need to change original codes that use this API and it also tells we are going to generate frame because these properties if animations exist.

**Comment by glennw** [comment]
r? @kvark 

I've pushed a commit which switches to use strongly typed units. If you can think of a cleaner way to implement it, let me know.

Also pushed an updated sample to https://github.com/glennw/webrender/commit/9142b695f79e948779628f87bda84878c694ff7b.

File: webrender/src/scene.rs [hunk 1]
```diff
@@ -7,7 +7,71 @@ use std::collections::HashMap;
...
+    }
+
+    /// Get the current value for a transform property.
+    pub fn resolve_layout_transform(&self, property: &PropertyBinding<LayoutTransform>) -> LayoutTransform {
+        match *property {
+            PropertyBinding::Value(matrix) => matrix,
+            PropertyBinding::Binding(ref key) => {
+                self.transform_properties
+                    .get(&key.id)
+                    .map(|m| *m)
```
kvark: nit: can use `cloned()`

webrender_traits/src/types.rs
```diff
@@ -81,6 +82,57 @@ impl ExternalEvent {
...
+            value: value,
+        }
+    }
+}
+
+/// A binding property can either be a specific value
+/// (the normal, non-animated case) or point to a binding location
+/// to fetch the current value from.
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize)]
+pub enum PropertyBinding<T> {
```
kvark: It would make sense to `impl From<T> for PropertyBinding<T>` as well as `impl From<PropertyBindingKey<T>> for PropertyBinding<T>`

glennw:
@kvark Thanks for the review - pushed a commit for those issues. I'll squash before merge, once r=you.

Issue #587, opened 2016-11-22
Split tiling.rs into sub-modules


kvark:
`tiling.rs` is the biggest WR source file with almost 3K lines. It's asking to be split into more independent sub-modules.

For example, the following classes can be moved out:
  - `FrameBuilder`
  - `ScreenTile`
  - `AlphaBatcher`
  - `ClipBatcher`
  - `RenderTarget`

glennw:
Most of this has been done. We can continue moving bits out incrementally as we refactor the remaining parts.

PR #834, opened 2017-02-06
Separate FrameBuilder into it's own file

mrobinson:
This is the first step toward splitting up tiling.rs into logical
units and will help to reduce merge conflicts in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/834)


mrobinson:
@glennw or @kvark r?

kvark:
I think the PR closes #587

mrobinson:
@kvark It does help, but I still think there is more work to do. tiling.rs still contains many things that can be moved out. :smile:

webrender/src/tiling.rs
```diff
@@ -860,10 +845,10 @@ impl ClipBatcher {
     }
 }
 
-struct RenderTargetContext<'a> {
-    stacking_context_store: &'a [StackingContext],
-    prim_store: &'a PrimitiveStore,
-    resource_cache: &'a ResourceCache,
+pub struct RenderTargetContext<'a> {
```
kvark: Can we have the fields hidden and only exposed to the module that uses them?
I imagine that both `tiling` and `frame_builder` need to create a `RenderTargetContext`, but do they both need to access the fields?

webrender/src/tiling.rs
```diff
@@ -1693,32 +1670,32 @@ pub struct StackingContextIndex(pub usize);
...
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct ScrollLayerIndex(pub usize);
 
 pub struct ScrollLayer {
-    scroll_layer_id: ScrollLayerId,
-    clip_source: ClipSource,
-    clip_cache_info: Option<MaskCacheInfo>,
-    packed_layer_index: PackedLayerIndex,
-    xf_rect: Option<TransformedRect>,
+    pub scroll_layer_id: ScrollLayerId,
```
kvark: ok, so there is a lot of stuff that is now public
I suppose we can address the privacy concerns in follow-ups

mrobinson: I agree. I think once everything has been moved it makes sense to go back and reduce the visibility of everything by adding methods to the structs.

mrobinson:
Thanks for the reviews.

PR #836, opened 2017-02-07
Refactor subpixel offsets to be it's own struct.

changm:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/836)

### DIFF
--- a/webrender_traits/src/types.rs
+++ b/webrender_traits/src/types.rs
@@ -351,7 +351,7 @@ impl FontRenderMode {
     }
 }
 
-#[repr(u32)]
+#[repr(u8)]
 #[derive(Hash, Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub enum SubpixelOffset {
     Zero            = 0,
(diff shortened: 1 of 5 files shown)


webrender_traits/src/types.rs
```diff
@@ -372,6 +372,35 @@ impl Into<f64> for SubpixelOffset {
 }
 
 #[derive(Clone, Hash, PartialEq, Eq, Debug, Deserialize, Serialize, Ord, PartialOrd)]
+pub struct SubpixelPoint {
```
kvark: Any reason we can just do `type SubpixelPoint = euclid::Point2D<SubpixelOffset>` ?

webrender_traits/src/types.rs
```diff
@@ -372,6 +372,35 @@ impl Into<f64> for SubpixelOffset {
...
+impl SubpixelPoint {
+    pub fn new(point: Point2D<f32>,
+               render_mode: FontRenderMode) -> SubpixelPoint {
+        SubpixelPoint {
+            x: render_mode.subpixel_quantize_offset(point.x),
+            y: render_mode.subpixel_quantize_offset(point.y),
+        }
+    }
+
+    pub fn x_offset(&self) -> f64 {
```
kvark: Having `fn to_f64(&self) -> (f64, f64)` would be simpler both here and at the caller sites

changm: Fixed!

webrender_traits/src/types.rs
changm: I tried to fix this, but had to implement PartialOrd for Point2D<SubpixelOffset>. When I tried to do that, I can't because you aren't allowed to implement traits for types outside your crate.

kvark:
Thanks @changm !
@bors-servo r+

PR #837, opened 2017-02-07
Do some more refactoring before adding support for stacking context independent clipping

mrobinson:
This refactoring should help to keep changes more compartmentalized when I add support for stacking context independent clipping.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/837)


mrobinson:
@kvark or @glennw r?

mrobinson:
> When the stuff is moved, and you don't separate the move from the cleanup into different commits, it's impossible to see what may be broken.

Sorry, I should have specified in the commit, but the cleanups are entirely comment reformatting. I got a little carried away and it seemed silly to undo it.

PR #839, opened 2017-02-07
Wrench cleanup and perspective support

kvark:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/839)

### DIFF
--- a/wrench/src/yaml_frame_reader.rs
+++ b/wrench/src/yaml_frame_reader.rs
@@ -142,25 +142,23 @@ impl YamlFrameReader {
             }
             &Yaml::Hash(_) => {
                 let bounds = item["rect"].as_rect().unwrap_or(*item_bounds);
-                let complex = if let Some(complex) = item["complex"].as_vec() {
-                    complex.iter().map(|item|
-                        match item {
-                            &Yaml::String(_) | &Yaml::Array(_) => {
-                                ComplexClipRegion::new(item.as_rect().expect("not a rect"), BorderRadius::zero())
-                            }
-                            &Yaml::Hash(_) => {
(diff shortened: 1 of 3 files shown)


wrench/src/yaml_frame_reader.rs
```diff
@@ -492,6 +493,10 @@ impl YamlFrameReader {
         let mix_blend_mode = yaml["mix-blend-mode"].as_mix_blend_mode().unwrap_or(MixBlendMode::Normal);
         let sc_full_rect = LayoutRect::new(LayoutPoint::new(0.0, 0.0), bounds.size);
         let clip = self.to_clip_region(&yaml["clip"], &sc_full_rect, wrench).unwrap_or(ClipRegion::simple(&sc_full_rect));
+        //let clip = ClipRegion::simple(&sc_full_rect);
```
glennw: Can we remove these comments?

wrench/src/yaml_helper.rs
```diff
@@ -237,12 +237,22 @@ impl YamlHelper for Yaml {
                                                                             1.0,
                                                                             Radians::new(theta));
 
-                    return Some(pre_transform.pre_mul(&transform).pre_mul(&post_transform))
+                    Some(pre_transform.pre_mul(&transform).pre_mul(&post_transform))
+                }
+                ("project", ref args) if args.len() == 1 => {
```
glennw: I think this should be ```"perspective"``` which is what CSS uses.

kvark: > I think this should be "perspective" which is what CSS uses.

Note that yaml would look like `perspective: perspective(100)`

kvark:
Fixes applied, waiting for CI.

glennw:
Looks good, thanks!

PR #843, opened 2017-02-07
Add a mix-blend-mode: difference reftest

rlhunt:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/843)


glennw:
@bors-servo r+

Thanks!

PR #845, opened 2017-02-08
Allow specifying a reftest to run in wrench

rlhunt:
This allows `wrench reftest` to run a specific reftest by filename. The first reftest that uses the filename for either the test or reference scene is used.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/845)

### DIFF
--- a/wrench/src/reftest.rs
+++ b/wrench/src/reftest.rs
@@ -10,7 +10,7 @@ use image::png::PNGEncoder;
 use std::cmp;
 use std::fs::File;
 use std::io::{BufRead, BufReader};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::mpsc::{channel, Receiver, Sender};
 use webrender_traits::*;
 use wrench::{Wrench, WrenchThing};
@@ -20,11 +20,10 @@ pub enum ReftestOp {
     Equal,
     NotEqual,
 }
-
-pub struct Reftest<'a> {
+pub struct Reftest {
     op: ReftestOp,
-    test: &'a Path,
-    reference: &'a Path,
+    test: PathBuf,
+    reference: PathBuf,
 }
 
 struct ReftestImage {
@@ -35,6 +34,7 @@ enum ReftestImageComparison {
     Equal,
     NotEqual { max_difference: usize, count_different: usize },
 }
+
 impl ReftestImage {
     fn compare(&self, other: &ReftestImage) -> ReftestImageComparison {
         assert!(self.size == other.size);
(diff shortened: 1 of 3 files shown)


wrench/src/reftest.rs
```diff
@@ -93,108 +93,144 @@ impl ReftestImage {
...
-    wrench.renderer.set_render_notifier(Box::new(Notifier { tx: tx }));
 
-    let mut total_passing = 0;
-    let mut total_failing = 0;
+    pub fn run(mut self, base_manifest: &Path, specific_reftest: Option<&Path>) {
+        let manifest = ReftestManifest::new(base_manifest);
+        let mut total_passing = 0;
+        let mut total_failing = 0;
+
+        match specific_reftest {
```
kvark: we can simplify this:
```rust
let reftests = match specific_reftest {
  Some(path) => &[manifest.find(path).expect(..)],
  None => &manifest.reftests,
};
for t in reftests.iter() { ... }
```

wrench/src/reftest.rs
```diff
@@ -93,108 +93,144 @@ impl ReftestImage {
...
-        let test = render_yaml(wrench, window, t.test, &rx);
-        let reference = render_yaml(wrench, window, t.reference, &rx);
+        let test = self.render_yaml(t.test.as_path());
+        let reference = self.render_yaml(t.reference.as_path());
         let comparison = test.compare(&reference);
 
-        let success = match (t.op, comparison) {
-            (ReftestOp::Equal, ReftestImageComparison::Equal) => true,
-            (ReftestOp::Equal, ReftestImageComparison::NotEqual { max_difference, count_different }) => {
+        let success = match (&t.op, comparison) {
```
kvark: what's the reason to switch it to `&t.op`?

wrench/src/reftest.rs
```diff
@@ -205,25 +241,38 @@ pub fn run_reftests(wrench: &mut Wrench, window: &mut WindowWrapper, filename: &
...
 
-        if success {
-            total_passing += 1;
-        } else {
-            total_failing += 1;
-        }
-    });
-
-    println!("REFTEST INFO | {} passing, {} failing", total_passing, total_failing);
+        success
```
kvark: no need to have the `success` variable now at all, we can just end the function with that `match` statement

wrench/src/reftest.rs
```diff
@@ -93,108 +93,144 @@ impl ReftestImage {
...
+                    let kind = match x {
+                        "==" => ReftestOp::Equal,
+                        "!=" => ReftestOp::NotEqual,
+                        _ => panic!("unexpected match operator"),
+                    };
+                    let test = dir.join(items.next().unwrap());
+                    let reference = dir.join(items.next().unwrap());
+                    reftests.push(Reftest {
+                        op: kind,
+                        test: test.clone(),
```
kvark: these `clone()`s seem unnecessary, we can just move

wrench/src/reftest.rs
rlhunt: t is now &Reftest and so match (t.op, comparison) is complaining on "moving out of borrowed content"

wrench/src/reftest.rs
rlhunt: I had a hard time fighting with the type checker on this with the incompatability of [&Reftest] and [Reftest]. I restructed the code to instead always search for a prefix match of the input. So you can do `wrench reftest reftests/blend` for all the reftests in that folder. This makes both cases the same.

kvark:
Thanks!
@bors-servo r+

glennw:
@rlhunt Looks like this needs a rebase, then is ready to merge.

PR #846, opened 2017-02-08
Fix up the pixel snapping code.

glennw:
This changes the pixel snapping code to snap to a device pixel
after transformation, rather than snapping in local space.

Previously, fixing this code was difficult to get exactly right
due to the tiling clamp and UV calculation, but now it's very simple.

This adds 31 new passes to the WPT/CSS tests, and 4 new failures. One
of the new failures also fails on Gecko / Blink. The other three new
failures appear to be a bug in osmesa - they don't occur on llvmpipe
or any real GPU hardware I've tested on (on osmesa, a very small number
of pixels in the 3 tests are off by 1/255 in one of the color channels).

Overall, this is a big improvement on the current situation, and means
that we'll be able to start the work to make subpixel positioning correct
in more situations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/846)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -422,13 +422,13 @@ VertexInfo write_vertex(vec4 instance_rect,
                         float z,
                         Layer layer,
                         AlphaBatchTask task) {
-    vec2 p0 = floor(0.5 + instance_rect.xy * uDevicePixelRatio) / uDevicePixelRatio;
-    vec2 p1 = floor(0.5 + (instance_rect.xy + instance_rect.zw) * uDevicePixelRatio) / uDevicePixelRatio;
+    vec2 p0 = instance_rect.xy;
+    vec2 p1 = instance_rect.xy + instance_rect.zw;
 
     vec2 local_pos = mix(p0, p1, aPosition.xy);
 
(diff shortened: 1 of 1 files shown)


kvark:
Note that this will only work when:
  - the local rectangle size is pixel-round
  - the transform doesn't have any rotation or scale

In order for the snapping to work for a general transform case, we need to back-propagate the `device_pos` changes to the `local_pos` and `local_rect`.

glennw:
Yup, that's correct. We specifically don't apply any snapping in the ```write_transform_vertex``` case as all bets are off when you have a rotation in regards to snapping.

kvark:
@glennw 

> We specifically don't apply any snapping in the write_transform_vertex case as all bets are off when you have a rotation in regards to snapping.

That's not enough. We use the transformed vertex code only for layers that don't fulfil `can_losslessly_transform_and_perspective_project_a_2d_rect` requirement, which is not the same as "has rotation or scale". For example, a layer rotated around Y axis would still be considered axis-aligned, and so does a scaled layer. These cases are not handled properly by the new pixel snapping code.

glennw:
@kvark True - we may need to handle those cases separately.

PR #847, opened 2017-02-08
Tidy up wrench and have tidy process it

mrobinson:
Unfortunately this includes a lot of changes, but I think it's better to get it out of the way sooner than later.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/847)

### DIFF
--- a/wrench/src/json_frame_writer.rs
+++ b/wrench/src/json_frame_writer.rs
@@ -1,21 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 // the json code is largely unfinished; allow these to silence a bunch of warnings
 #![allow(unused_variables)]
 #![allow(dead_code)]
 
 use image::{ColorType, save_buffer};
+use serde_json;
 use std::borrow::BorrowMut;
 use std::collections::HashMap;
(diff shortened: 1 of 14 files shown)


wrench/src/main.rs
```diff
@@ -276,10 +267,15 @@ fn main() {
     let dp_ratio = args.value_of("dp_ratio").map(|v| v.parse::<f32>().unwrap());
     let limit_seconds = args.value_of("time").map(|s| time::Duration::seconds(s.parse::<i64>().unwrap()));
     let save_type = args.value_of("save").map(|s| {
-        if s == "yaml" { wrench::SaveType::Yaml }
-        else if s == "json" { wrench::SaveType::Json }
-        else if s == "binary" { wrench::SaveType::Binary }
-        else { panic!("Save type must be json, yaml, or binary"); }
+        if s == "yaml" {
```
kvark: dat match

wrench/src/json_frame_writer.rs
```diff
@@ -1,21 +1,23 @@
...
+use serde_json;
 use std::borrow::BorrowMut;
 use std::collections::HashMap;
-use std::fmt;
-use std::fs;
 use std::fs::File;
 use std::io::Write;
 use std::path::{Path, PathBuf};
+use std::{fmt, fs};
+use super::CURRENT_FRAME_NUMBER;
```
kvark: it seems slightly wrong to me that `super::*` includes are mixed with the others

PR #850, opened 2017-02-08
Recalculate clip stack when it changes in FrameBuilder

mrobinson:
Instead of maintaining a stack of MaskCacheInfo recalculate the
MaskCacheInfo stack every time we see a new primitive and the scroll
layer has changed. This will allow us to switch arbitrarily between
scroll layers in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/850)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -31,7 +31,7 @@ use webrender_traits::{BoxShadowClipMode, ClipRegion, ColorF, device_length, Dev
 use webrender_traits::{DeviceIntRect, DeviceIntSize, DeviceUintSize, ExtendMode, FontKey};
 use webrender_traits::{FontRenderMode, GlyphOptions, ImageKey, ImageRendering, ItemRange};
 use webrender_traits::{LayerPoint, LayerRect, LayerSize, LayerToScrollTransform, PipelineId};
-use webrender_traits::{ScrollLayerId, ScrollLayerPixel, WebGLContextId, WorldRect, YuvColorSpace};
+use webrender_traits::{ScrollLayerId, ScrollLayerPixel, WebGLContextId, YuvColorSpace};
 
 #[derive(Clone, Copy)]
 pub struct FrameBuilderConfig {
@@ -65,6 +65,9 @@ pub struct FrameBuilder {
     packed_layers: Vec<PackedLayer>,
 
     scrollbar_prims: Vec<ScrollbarPrimitive>,
+
+    /// A stack of scroll layers used during building to properly parent new scroll layers.
+    scroll_layer_stack: Vec<ScrollLayerIndex>,
 }
 
 impl FrameBuilder {
@@ -81,6 +84,7 @@ impl FrameBuilder {
             packed_layers: Vec::new(),
             scrollbar_prims: Vec::new(),
             config: config,
+            scroll_layer_stack: Vec::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


mrobinson:
@kvark Do you mind looking in particular at the changes dealing with empty clips? I didn't entirely understand the situation where this happens. I believe the code will still deal properly with empty clips now, but I didn't have a test case to use.

webrender/src/frame_builder.rs
```diff
@@ -1098,16 +1080,45 @@ impl<'a> LayerRectCalculationAndCullingPass<'a> {
...
+        if let Some(previous_scroll_layer) = self.current_clip_stack_scroll_layer {
+            if previous_scroll_layer == scroll_layer_index {
+                return;
+            }
+        }
+
+        // TODO(mrobinson): If we notice that this process is expensive, we can special-case
+        // more common situations, such as moving from a child or a parent.
+        self.current_clip_stack_scroll_layer = Some(scroll_layer_index);
+        let mut new_clip_info_stack = Vec::new();
```
kvark: why are we working with `self.current_clip_stack` directly here?

kvark:
Let me try to explain how it works today. The actual clip stack that we consider for the mask task has all the clips on the way to the item plus all the non-aligned clip rectangles of the parent scroll layers.
The layers with axis-aligned clip rectangles are handled differently:
  - we keep the `world_clip_rect_stack` of these aligned clip rects so that we can go back on "pop_scroll_layer".
  - each next rectangle in this `world_clip_rect_stack` is inside the parent. We intersect the current layer's rectangle with the top of the stack before pushing.
  - we also transform the combined clip rectangle back into the layer space and *patch* the `local_clip_rect` of the layer, making all the contained items to automatically be clipped by this new rectangle

webrender/src/frame_builder.rs
mrobinson: Hrm. Do you mean why aren't we working with ```self.current_clip_stack```? If so, it's to avoid a double mutable borrow. This probably deserves a comment!

mrobinson:
@kvark Do you have a test case which shows the problem with the empty clips? I can try to reintegrate your approach while keeping the ability to move between clips without relying on a nested traversal.

kvark:
@mrobinson `wrench/reftests/mask/aligned-layer-rect.yaml` should test for it
I'll start looking into why it passes with your change.

webrender/src/frame_builder.rs
kvark: There is no double mutable borrow. This code runs:
```rust
        self.current_clip_stack_scroll_layer = Some(scroll_layer_index);
        self.current_clip_stack.clear();
        loop {
            let scroll_layer = &self.frame_builder.scroll_layer_store[scroll_layer_index.0];
            match scroll_layer.clip_cache_info {
                Some(ref clip_info) if clip_info.is_masking() =>
                    self.current_clip_stack.push((scroll_layer.packed_layer_index, clip_info.clone())),
                _ => {},
            };

            if scroll_layer.parent_index == scroll_layer_index {
                break;
            }
            scroll_layer_index = scroll_layer.parent_index;
        }

        self.current_clip_stack.reverse();
```

mrobinson: Thanks! I could have sworn that didn't compile before. Likely this was a holdover from an older version of the code. I've fixed this in the latest version of the branch.

PR #852, opened 2017-02-08
Support using the same vector cache thread and thread pool with several render backends

nical:
This PR renames the glyph cache thread into vector cache thread in preparation for the upcoming vector image renderer work, and makes it possible for several render backends to use the same vector cache thread.
This makes it possible to prevent creating 5 threads for each window (Gecko tends to have a bunch of them since each tooltip and popup is its own window), although by default (as in RendererOption::default()), each Renderer creates its own vector cache thread with 4 worker threads.

The main difference is that the vector cache thread doesn't keep a Sender<VectorCacheResultMsg> anymore, the latter being passed in the EndFrame message every time.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/852)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -28,25 +28,7 @@ use euclid::Point2D;
 
 thread_local!(pub static FONT_CONTEXT: RefCell<FontContext> = RefCell::new(FontContext::new()));
 
-type GlyphCache = ResourceClassCache<RenderedGlyphKey, Option<TextureCacheItemId>>;
-
-/// Message sent from the resource cache to the glyph cache thread.
-enum GlyphCacheMsg {
-    /// Begin the frame - pass ownership of the glyph cache to the thread.
-    BeginFrame(FrameId, GlyphCache),
-    /// Add a new font.
-    AddFont(FontKey, FontTemplate),
-    /// Request glyphs for a text run.
-    RequestGlyphs(FontKey, Au, ColorF, Vec<GlyphInstance>, FontRenderMode, Option<GlyphOptions>),
-    /// Finished requesting glyphs. Reply with new glyphs.
-    EndFrame,
-}
-
-/// Results send from glyph cache thread back to main resource cache.
-enum GlyphCacheResultMsg {
-    /// Return the glyph cache, and a list of newly rasterized glyphs.
-    EndFrame(GlyphCache, Vec<GlyphRasterJob>),
-}
+pub type GlyphCache = ResourceClassCache<RenderedGlyphKey, Option<TextureCacheItemId>>;
 
 // These coordinates are always in texels.
 // They are converted to normalized ST
@@ -177,11 +159,6 @@ struct ImageRequest {
     rendering: ImageRendering,
 }
 
-struct GlyphRasterJob {
-    key: RenderedGlyphKey,
-    result: Option<RasterizedGlyph>,
-}
-
 struct WebGLTexture {
     id: SourceTexture,
     size: DeviceIntSize,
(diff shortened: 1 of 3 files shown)


webrender/src/resource_cache.rs
```diff
@@ -621,18 +601,40 @@ impl Resource for CachedImageInfo {
...
+    EndFrame(GlyphCache, Vec<GlyphRasterJob>),
+}
+
+pub struct GlyphRasterJob {
+    pub key: RenderedGlyphKey,
+    pub result: Option<RasterizedGlyph>,
+}
+
+pub fn spawn_vector_cache_thread(worker_count: usize) -> Sender<VectorCacheMsg> {
+    let worker_count = if worker_count > 0 { worker_count } else { 1 };
```
kvark: I assume we'll have something like `const DEFAULT_WORKER_COUNT: usize = 1` eventually

glennw:
I haven't read this is detail yet - but it sounds like the GlyphCache thread is shared as well as the worker thread pool?

I was thinking of sharing the thread pool only, and still having one GlyphCache thread per instance. This would be a lot simpler and fix any issues with BeginFrame/EndFrame namespacing etc.

How does that sound?

kvark:
@glennw 

> I was thinking of sharing the thread pool only, and still having one GlyphCache thread per instance. This would be a lot simpler and fix any issues with BeginFrame/EndFrame namespacing etc.

Citing a person all of us have met: "That would be glorious!"

nical:
Sharing only the thread pool was what I was going for initially, but I ended up going for sharing the vector cache as well. The reason is that it itches me that when a user hovers his cursor over something and a little tooltip appears, two threads have to be created (one for the render backend and one for the vector cache (and previously 4 more threads due to the thread pool). There is also the vague hope that several windows could share some resources like font data and rasterized glyphs.
In gecko windows add up a lot quicker than in servo (which I think has only one).

I only superficially looked at the actual logic in the thread and indeed after a closer look it wouldn't work out of the box when sharing the vector cache thread.

So I'm a bit undecided about this. I guess we have more important things to do so I'll stick with sharing only the thread pool for now since it is dead simple. Some time later it would be worth coming back to this and think about how we can make this all play optimally with many windows (including not creating a render backend thread for every tooltip). In the mean time we can rely on gecko only using WebRender for the main windows and using the old BasicLayerManager for tooltips and popups.

glennw:
@nical Sounds good, thanks! I agree that we should revisit this later with a view to making multiple instances share resources / threads better. In the meantime, sharing the thread pool is a nice easy win :)

kvark:
@nical 
> The reason is that it itches me that when a user hovers his cursor over something and a little tooltip appears, two threads have to be created

What's going on on the Gecko side in this case? Is the new window created for a tooltip? Does it have its own GL context, and if so - can it share resources with the main one?
Asking to probe if Gecko can re-cycle WR instances instead of re-creating them.

nical:
In gecko, things like tooltips, drop-down menus, the awesome-bar's autocomplete thing, popups, and the likes all have their own windows. I wouldn't be surprised that we recycle tooltip windows because we typically only ever have one or none present at a time. For the rest I don't think we do. Each window has its own GL context which it is tightly coupled to, so I don't think we can recycle a renderer and put it on a different window (unless we can replace the gl context of an existing renderer, but then we add complexity in a place where it is less useful).

For now it's fair to assume that WebRender is only for big/complex windows and not use it with smaller popup type things. In a distant and bright future, it would make sense to be able to turn some knobs and make it lighter-weight for small/simple windows (and not have to use the legacy rendering code for these).
That and also leverage the probability that several windows are using the same resources like fonts.

glennw:
Closing in favour of https://github.com/servo/webrender/pull/855

PR #854, opened 2017-02-09
Add hooks for thread_profiler.

glennw:
This produces trace files that can be viewed in the chromium thread profiler.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/854)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ script:
   - (cd webrender_traits && cargo test --verbose)
   - (cd webrender && cargo test --verbose)
   - (cd sample && cargo test --verbose)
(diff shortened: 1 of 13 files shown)


glennw:
@kvark I'm interested to hear your thoughts on this approach, and if it seems reasonable to you. If so, I'll clean up the PR, publish the profiler crate etc.

You end up with a CPU profile that shows what each thread is doing, e.g:

wr

glennw:
@kvark As you can see from the picture above, this already identifies several issues we have with our thread overlapping that need to be fixed :)

glennw:
r? @kvark 

I've tidied up the PR and switched to using a published crate, to avoid vendoring issues.

PR #855, opened 2017-02-09
Enable several WebRender instances to share a thread pool.

nical:
This is a simplified version of PR #852 which only enables sharing the thread pool.

It would be worth looking into why ThreadPool isn't Sync and see if we can avoid the mutex.

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/855)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -17,6 +17,7 @@ use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Sender;
 use texture_cache::TextureCache;
 use thread_profiler::register_thread_with_profiler;
+use threadpool::ThreadPool;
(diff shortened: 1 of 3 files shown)


kvark:
Thanks!
@bors-servo r+

kvark:
@bors-servo retry
  * mac bots are grumpy again

kvark:
@bors-servo retry
  * mac bots are still sleepy

kvark:
Ok, let's try again. It's Friday, what can possibly go wrong?
@bors-servo r+

kvark:
Uh, that's new:
> The command "curl -sL https://static.rust-lang.org/rustup.sh -o ~/rust-installer/rustup.sh" failed and exited with 35 during .

@bors-servo retry

kvark:
I wish bors didn't have to rebuid the platforms that already succeeded for the same changelist.

PR #856, opened 2017-02-09
Remove unnecessary workspace pointers.

staktrace:
Cargo will automatically walk up the file system hierarchy to find a containing
workspace, so the explicitly specifying the workspace path in the workspace
members is not necessary. For standalone webrender it is harmless, but it is
causing a problem when trying to wrap all the rust projects in mozilla-central
in another workspace (see https://bugzilla.mozilla.org/show_bug.cgi?id=1302704#c25).
Removing this should fix the problem.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/856)


emilio:
Sounds ok.

@bors-servo r+

PR #857, opened 2017-02-09
Update serde to 0.9

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/857)


kvark:
Looks good!
What's the deal with "WIP REMOVEME" commit?

nox:
@kvark Doing a try build on servo/servo first before merging anything, and for various reasons we can't have the bumps in the code to use `[replace]` to test before merging and publishing.

PR #858, opened 2017-02-10
Blob image renderer

nical:
Here is a simple integration of a VectorImageRenderer abstraction that can be implemented to render vector graphics to an image.

The main ideas are that:
 - The interface should be implementable outside of WebRender in order for gecko to put in there things that WebRender cannot handle yet.
 - The result should be transparently usable as a regular image.
 - The intent is for the renderer to use the same thread pool as the glyph cache (an implementation detail of the vector renderer thanks to PR #855)
 - The interface should be extended to support some sort of sparse/tiled output for very large scrollable vector images.

The wr_sample implements a dummy vector renderer that always outputs a checkerboard pattern. You can have a look and see how it works.

Implementation-wise, vector images are treated like raw images, except that when an image is requested, if it is a vector image the resource cache asks the renderer to paint it asynchronously, and waits for the result at the end of the frame, just before inserting it into the texture cache. I am pretty happy with how simple it turns out (but I guess it will become a lot trickier with zoom support), it took me a few attempts, the first few implementations were a lot more complicated.

The VectorImageRenderer interface needs a bit of work. The request expects a certain size, which will probably not play well with the scaling factor that we should apply when zooming. It also passes an ImageDescriptor but it's not obvious to me whether we want to support anything other than RGBA8 (and maybe A8 for masks) and the stride member is also awkward.
I'm not sure what's best between VectorImageData being a ```Vec<u8>``` or a pointer-sized opaque handle like ExternalEvent (in which case we'd need to figure out who gets to delete the thing).

Feedback very welcome @kvark @glennw @jrmuizel @JerryShih

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/858)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -24,6 +24,7 @@ use webrender_traits::{Epoch, FontKey, GlyphKey, ImageKey, ImageFormat, ImageRen
 use webrender_traits::{FontRenderMode, ImageData, GlyphDimensions, WebGLContextId};
 use webrender_traits::{DevicePoint, DeviceIntSize, ImageDescriptor, ColorF};
 use webrender_traits::{ExternalImageId, GlyphOptions, GlyphInstance};
+use webrender_traits::{BlobImageRenderer, BlobImageDescriptor, BlobImageError};
 use threadpool::ThreadPool;
 use euclid::Point2D;
 
@@ -209,11 +210,15 @@ pub struct ResourceCache {
     glyph_cache_tx: Sender<GlyphCacheMsg>,
     glyph_cache_result_queue: Receiver<GlyphCacheResultMsg>,
     pending_external_image_update_list: ExternalImageUpdateList,
+
+    blob_image_renderer: Option<Box<BlobImageRenderer>>,
+    blob_image_requests: HashSet<ImageRequest>,
 }
 
 impl ResourceCache {
     pub fn new(texture_cache: TextureCache,
                workers: Arc<Mutex<ThreadPool>>,
+               blob_image_renderer: Option<Box<BlobImageRenderer>>,
                enable_aa: bool) -> ResourceCache {
         let (glyph_cache_tx, glyph_cache_result_queue) = spawn_glyph_cache_thread(workers);
 
@@ -232,6 +237,9 @@ impl ResourceCache {
             glyph_cache_tx: glyph_cache_tx,
             glyph_cache_result_queue: glyph_cache_result_queue,
             pending_external_image_update_list: ExternalImageUpdateList::new(),
+
+            blob_image_renderer: blob_image_renderer,
+            blob_image_requests: HashSet::new(),
         }
     }
 
@@ -321,14 +329,38 @@ impl ResourceCache {
         webgl_texture.size = size;
     }
 
-    pub fn request_image(&mut self,
-                         key: ImageKey,
-                         rendering: ImageRendering) {
+    pub fn request_image(&mut self, key: ImageKey, rendering: ImageRendering) {
         debug_assert!(self.state == State::AddResources);
-        self.pending_image_requests.push(ImageRequest {
+        let request = ImageRequest {
             key: key,
             rendering: rendering,
-        });
+        };
+
+        let template = self.image_templates.get(&key).unwrap();
+        if let ImageData::Vector(ref data) = template.data {
+            if let Some(ref mut renderer) = self.blob_image_renderer {
+                let same_epoch = match self.cached_images.resources.get(&request) {
+                    Some(entry) => entry.epoch == template.epoch,
+                    None => false,
+                };
+
+                if !same_epoch && self.blob_image_requests.insert(request) {
+                    renderer.request_blob_image(
+                        key,
+                        data.clone(),
+                        &BlobImageDescriptor {
+                            width: template.descriptor.width,
+                            height: template.descriptor.height,
+                            format: template.descriptor.format,
+                            // TODO(nical): figure out the scale factor (should change with zoom).
+                            scale_factor: 1.0,
+                        },
+                    );
+                }
+            }
+        } else {
+            self.pending_image_requests.push(request);
+        }
     }
 
     pub fn request_glyphs(&mut self,
@@ -453,7 +485,7 @@ impl ResourceCache {
         let external_id = match image_template.data {
             ImageData::ExternalHandle(id) => Some(id),
             // raw and externalBuffer are all use resource_cache.
-            ImageData::Raw(..) | ImageData::ExternalBuffer(..) => None,
+            ImageData::Raw(..) | ImageData::ExternalBuffer(..) | ImageData::Vector(..) => None,
         };
 
         ImageProperties {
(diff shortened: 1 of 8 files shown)


glennw:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[webrender/src/internal_types.rs, line 24 at r1](https://reviewable.io:443/reviews/servo/webrender/858#-KcpUyYU7H4JquMGNOnR:-KcpUyYVmqB7ylIGWrRC:bu36yez) ([raw file](https://github.com/servo/webrender/blob/ee4783ddb39119b84243ff06f4c18f4095c84333/webrender/src/internal_types.rs#L24)):*
> ```Rust
> use webrender_traits::{ImageFormat, NativeFontHandle};
> use webrender_traits::{ExternalImageId, ScrollLayerId, WebGLCommand};
> use webrender_traits::{ImageKey, ImageDescriptor, ImageData, VectorImageData};
> ```

These shouldn't be needed - since nothing else in this file has changed?

---

*[webrender/src/resource_cache.rs, line 350 at r1](https://reviewable.io:443/reviews/servo/webrender/858#-KcpX69awmzE15ClXsyz:-KcpX69bHfZD36AUXIXR:b-qet23q) ([raw file](https://github.com/servo/webrender/blob/ee4783ddb39119b84243ff06f4c18f4095c84333/webrender/src/resource_cache.rs#L350)):*
> ```Rust
>             }
>         }
>         self.pending_image_requests.push(request);
> ```

I think it would make more sense to not push on to the pending_image_requests list at all in the case of a vector image (explained further below).

---

*[webrender/src/resource_cache.rs, line 566 at r1](https://reviewable.io:443/reviews/servo/webrender/858#-KcpXFriS2ihgd3z0NYq:-KcpXFrj3D9VpEdtgkoY:b-8b2fm6) ([raw file](https://github.com/servo/webrender/blob/ee4783ddb39119b84243ff06f4c18f4095c84333/webrender/src/resource_cache.rs#L566)):*
> ```Rust
>             let mut image_data = image_template.data.clone();
> 
>             if let Some(ref mut renderer) = self.vector_image_renderer {
> ```

Instead of having to check each pending image to see if it is a vector image, how about we just iterate the vector_image_requests list in a separate loop? (moving the image_template.data code below into a separate function).

---

*[webrender/src/texture_cache.rs, line 746 at r1](https://reviewable.io:443/reviews/servo/webrender/858#-KcpVQ7XeBtw0VDPIhcz:-KcpVQ7XeBtw0VDPIhd-:b6zzob8) ([raw file](https://github.com/servo/webrender/blob/ee4783ddb39119b84243ff06f4c18f4095c84333/webrender/src/texture_cache.rs#L746)):*
> ```Rust
>                   data: ImageData) {
>         if let ImageData::Vector(..) = data {
>             panic!("must rsterize the vector image before adding to the cache");
> ```

nit: spelling

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/858)*

glennw:
@nical In general, this looks pretty good to me - and I think it's fine to merge an initial interface and iterate on it in-tree. I have two comments apart from the reviewable notes above:

1) Instead of calling everything VectorImage and related, should it instead be called something more generic (since presumably you would use this same interface for any procedurally generated images, not just vectors)?

2) Are we planning to have the tiling logic handled in WR, or by the application? If in WR (which makes sense to me, since WR has more relevant information available), I think it would probably make sense for the rasterize callback to work more like the glyph cache rasterizer. That is, the glyph cache thread would be executing jobs on the thread pool for each tile, rather than issuing one large callback. That's not important for an initial PR, but something we should consider in the future.

nical:
I am totally open to suggestions for the naming scheme, naming thing isn't my strong suit (especially in english).
I think that we should have tiling integrated in WebRender, if only because other features like support for very large images need something like that. However I think that the threading model should be up to the vector image renderer, since depending on what you render and what third party library you use, you get different constraints with respect to threading.
With the thread pool PR that recently landed we can just go:
```rust
let vector_renderer = SkiaVectorRenderer::new(thread_pool.clone());
let options = RendererOptions {
    vector_image_renderer: vector_renderer,
    workers: thread_pool.clone(),
    // ...
};
```
And implement the threading details in the vector renderer implementation.

This is easily composable so for the simplest case where we can just go run one renderer/tile per worker thread we could have a TiledVectorRenderer that has an array of vector renderers and dispatches them with work on the thread pool.

nical:
> Instead of having to check each pending image to see if it is a vector image, how about we just iterate the vector_image_requests list in a separate loop?

This isn't necessarily simpler because ```vector_image_requests``` refers to the list of vector images that need to be rasterized this frame whereas ```pending_image_requests``` refers to images that are used in this frame. If a vector image is rasterized during frame N, it goes through ```vector_image_requests``` during frame N but not frame N+1. In both frames it should go through the same logic as images handled by ```pending_image_requests``` so that it gets sent to the texture cache if needed and have its epoch updated.

I'll look into to simplifying this.

nical:
I just realized this PR doesn't in its current state support updating vector images, I'll fix that soon. The PR is also still pretty rough when it comes to the API for ```VectorImageRenderer::request_vector_image``` and how to make it work with varying zoom levels, although I suppose that part can be polished later as we experiment with an actual implementation of the trait.

glennw:
@nical Do you want review comments on this now, or is it still WIP?

nical:
@glennw I'd love to get a review. Structurally I think that the patch is in a good state. If you have ideas for more adequate names don't hesitate to ask me to change them!

glennw:
@nical Sorry, I didn't get to this today - I'll take a proper look at it on Monday, but it looks quite good from a quick skim!

webrender/src/resource_cache.rs
```diff
@@ -321,14 +329,37 @@ impl ResourceCache {
...
             key: key,
             rendering: rendering,
-        });
+        };
+
+        let template = self.image_templates.get(&key).unwrap();
+        if let ImageData::Vector(ref data) = template.data {
+            if let Some(ref mut renderer) = self.vector_image_renderer {
+                let same_epoch = self.cached_images.resources.get(&request)
+                    .map(|entry|{ entry.epoch == template.epoch})
```
glennw: nit: the ```{``` and ```}``` are not required here.

glennw: nit: I think this would probably be tidier written as a match { Some(..) => {} None => {} } statement.

webrender_traits/src/types.rs
```diff
@@ -605,9 +605,44 @@ pub enum YuvColorSpace {
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub struct ExternalImageId(pub u64);
 
+pub trait VectorImageRenderer: Send {
```
glennw: How about we change this to BlobImageRenderer (and associated types). Blob seems like a reasonable description (Binary Large Object) and makes it clear that this can be used for more than just vector image data.

glennw:
Thanks! We should also rename ```ImageData::Vector``` to ```ImageData::Blob``` - but let's get this merged and do that as a follow up.

kvark:
Reviewed 1 of 10 files at r2, 2 of 8 files at r3, 6 of 9 files at r4.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[webrender/src/renderer.rs, line 1759 at r4](https://reviewable.io:443/reviews/servo/webrender/858#-KdWxFqJ6q7Jw1cmBZMO:-KdWxFqJ6q7Jw1cmBZMP:btr9nnr) ([raw file](https://github.com/servo/webrender/blob/bc59999b2cfaf5d019a41854b0a0f7219e45c628/webrender/src/renderer.rs#L1759)):*
> ```Rust
> }
> 
> #[derive(Debug)]
> ```

that's unfortunate. I guess we'll need to implement `Debug` manually in some of the follow-ups

---

*[webrender/src/resource_cache.rs, line 339 at r4](https://reviewable.io:443/reviews/servo/webrender/858#-KdWyCPYBazxKxwNc6qX:-KdWyCPZEJ4AwshQO13H:b-55xm9a) ([raw file](https://github.com/servo/webrender/blob/457a25a7a08524bd618f463762f57cd43a90646f/webrender/src/resource_cache.rs#L339)):*
> ```Rust
>         };
> 
>         let template = self.image_templates.get(&key).unwrap();
> ```

hmm, I wonder why it has to go through the image templates at all. The templates imply pixel data, while our blobs are just blobs. Besides, keeping the template around forces the clone of the data vector.

---

*[webrender/src/resource_cache.rs, line 574 at r4](https://reviewable.io:443/reviews/servo/webrender/858#-KdWyeDWYbNblLyOkPKY:-KdWyeDWYbNblLyOkPKZ:bpk9hlv) ([raw file](https://github.com/servo/webrender/blob/457a25a7a08524bd618f463762f57cd43a90646f/webrender/src/resource_cache.rs#L574)):*
> ```Rust
>         }
> 
>         let mut image_requests = mem::replace(&mut self.pending_image_requests, Vec::new());
> ```

I think we should replace the old vector back after draining it. This is non-critical, but essentially allows us to retain capacity and thus avoid re-allocating data on the heap each frame.

---

*[webrender/src/resource_cache.rs, line 580 at r4](https://reviewable.io:443/reviews/servo/webrender/858#-KdWytRba_3CNwAl1clG:-KdWytRcmo2nDhEwkdQe:bd9dzn1) ([raw file](https://github.com/servo/webrender/blob/457a25a7a08524bd618f463762f57cd43a90646f/webrender/src/resource_cache.rs#L580)):*
> ```Rust
> 
>         let mut blob_image_requests = mem::replace(&mut self.blob_image_requests, HashSet::new());
>         if self.blob_image_renderer.is_some() {
> ```

any reason this is not `if let Some(ref renderer) = ` ?

---

*[webrender/src/resource_cache.rs, line 611 at r4](https://reviewable.io:443/reviews/servo/webrender/858#-KdWzOXtWdKhEsUANERp:-KdWzOXuNp_dWBLDeqRD:b-psj2e5) ([raw file](https://github.com/servo/webrender/blob/457a25a7a08524bd618f463762f57cd43a90646f/webrender/src/resource_cache.rs#L611)):*
> ```Rust
>         let image_template = &self.image_templates[&request.key];
>         let image_data = image_data.unwrap_or_else(||{
>             image_template.data.clone()
> ```

another data clone that would be nice to avoid

---

*[webrender/src/resource_cache.rs, line 614 at r4](https://reviewable.io:443/reviews/servo/webrender/858#-KdWzQpKf4C7eKwGl5As:-KdWzQpKf4C7eKwGl5At:b-fpt3f3) ([raw file](https://github.com/servo/webrender/blob/457a25a7a08524bd618f463762f57cd43a90646f/webrender/src/resource_cache.rs#L614)):*
> ```Rust
>         });
> 
>         match image_template.data {
> ```

hmm, so the logic inside this `match` statement used to be done inside `block_until_all_resources_added` for all pending images. With your PR it's only done for the image blobs now. This seems suspicious. Is it by design?

---

*[webrender_traits/src/types.rs, line 610 at r4](https://reviewable.io:443/reviews/servo/webrender/858#-KdWzxfuvVxF8d9Gz3fx:-KdWzxfuvVxF8d9Gz3fy:b-e6d4hj) ([raw file](https://github.com/servo/webrender/blob/457a25a7a08524bd618f463762f57cd43a90646f/webrender_traits/src/types.rs#L610)):*
> ```Rust
> pub trait BlobImageRenderer: Send {
>     fn request_blob_image(&mut self,
>                             key: ImageKey,
> ```

nit: alignment

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/858)*

kvark:
@nical I'm a bit late to the party, got a few questions/concerns.

nical:
> any reason this is not if let Some(ref renderer) = ?

Yeah, unfortunately this block into to make some ```&mut self``` method calls but self would be borrowed already by the ```if let Some(ref renderer)```. That was my first attempt.

>  Besides, keeping the template around forces the clone of the data vector
> (...)
> another data clone that would be nice to avoid

Note that we are cloning ```Arc<Vec<T>>``` in these places, so it's actually cheap. I very much dislike how ```clone``` (which is one of the most vaguely defined concept in rust) is the only method name for incrementing a simple reference count.

> hmm, I wonder why it has to go through the image templates at all. The templates imply pixel data, while our blobs are just blobs.

My mental model is that ImageTemplate implies "something that describes the image". We need to store the descriptor and blob data somewhere anyway, my first prototype had a separate data structure for that but it ended up being a lot more complicated because the use of blob image is transparent (image key and all) so you'd have to know somehow whether an image is a regular or blob when referring to it with an image key in all of the resource_cache code before looking up the proper map.

> hmm, so the logic inside this match statement used to be done inside block_until_all_resources_added for all pending images. With your PR it's only done for the image blobs now. This seems suspicious. Is it by design?

I will double check tomorrow but I think that this logic is also run for non-blob pending images (finalize_image_request being called in the loop that processes pending regular image requests in addition to the loop that goes through pending blob image requests).

kvark:
> Note that we are cloning Arc<Vec<T>> in these places, so it's actually cheap

Oh, right. Not a problem then.

> We need to store the descriptor and blob data somewhere anyway

Yeah, if the image templates fit this purpose, then we are good. I'm more confused about the naming here, since for your case it's no longer a template for an image, as it is a template for a blob.

PR #860, opened 2017-02-10
Use color_type instead of wrongly assuming RGB8

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/860)


kvark:
Thanks!
@bors-servo r+

PR #867, opened 2017-02-11
Use glBlend for certain mix-blend-mode types

rlhunt:
Certain mix-blend-mode types can be implemented with glBlend. This adds an AlphaRenderItem for a composite operation that doesn't need readback or processing in the shader, but instead uses glBlend. And then certain mix-blend-modes are expressed in this AlphaRenderItem.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/867)

### DIFF
--- a/webrender/res/ps_hardware_composite.fs.glsl
+++ b/webrender/res/ps_hardware_composite.fs.glsl
@@ -1,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+void main(void) {
+    oFragColor = texture(sCache, vUv);
+}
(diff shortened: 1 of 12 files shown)


rlhunt:
Naming things is hard, I'm not attached to the name HardwareComposite, so I'd be happy to change it. Also there are tests for the mix-blend-mode's affected by this PR.

glennw:
@rlhunt This generally looks good to me. There is one manual test in webrender/tests/mix-blend-mode.html that seems to be broken with this change however. If I run it in Servo with this patch, it no longer matches what Gecko renders. I'm not sure if it would be easier for you to build Servo to test this, or just port that html file to a wrench test.

rlhunt:
Okay, I'll take a look at that test.

rlhunt:
Oops, I didn't pay close enough attention to the spec.

mix-blend-mode: difference is |Cb - Cs|, not Cb - Cs. I can't really think of a way using glBlend to do this at the moment. But I also figured out we could do mix-blend-mode: multiply, and then noticed it was already hooked up partially. So I hooked that up and removed the difference code.

glennw:
@bors-servo r+

Thanks!

PR #868, opened 2017-02-11
Properly isolate stacking contexts with mix-blend-mode children

rlhunt:
Anything in HTML that creates a stacking context must create an isolated group for mix-blend-mode. [1] This means that the initial backdrop for any children with mix-blend-mode must be transparent black. Currently stacking context's don't create isolated groups, and instead the backdrop can be from multiple parents up the stacking context tree.

This is a WIP that tries to implement the right behavior. When a stacking context has mix-blend-mode children, it's contents need to be drawn independently with the correct backdrop (transparent black), then composited.

It looks like the code works, but I'd like some feedback to see if this is an okay approach.

[1] https://www.w3.org/TR/compositing-1/#csscompositingrules_CSS

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/868)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -118,6 +118,7 @@ pub struct FrameBuilder {
     /// A stack of scroll nodes used during display list processing to properly
     /// parent new scroll nodes.
     clip_scroll_node_stack: Vec<ScrollLayerIndex>,
+    stacking_context_stack: Vec<StackingContextIndex>,
 }
 
 impl FrameBuilder {
@@ -136,6 +137,7 @@ impl FrameBuilder {
             scrollbar_prims: Vec::new(),
             config: config,
             clip_scroll_node_stack: Vec::new(),
+            stacking_context_stack: Vec::new(),
         }
     }
 
@@ -203,14 +205,30 @@ impl FrameBuilder {
                                  rect: LayerRect,
                                  pipeline_id: PipelineId,
                                  scroll_layer_id: ScrollLayerId,
+                                 is_page_root: bool,
                                  composite_ops: CompositeOps) {
         let stacking_context_index = StackingContextIndex(self.stacking_context_store.len());
+        let parent_index = self.stacking_context_stack.last().map(|x| x.0);
+        self.stacking_context_stack.push(stacking_context_index);
+
+        if let Some(parent_index) = parent_index {
+            let parent_is_root = self.stacking_context_store[parent_index].is_page_root;
+
+            if composite_ops.mix_blend_mode.is_some() && !parent_is_root {
+                // the parent stacking context of a stacking context with mix-blend-mode
+                // must be drawn with a transparent background, unless the parent stacking context
+                // is the root of the page
+                self.stacking_context_store[parent_index].should_isolate = true;
+            }
+        }
+
         let group_index = self.create_clip_scroll_group(stacking_context_index,
                                                         scroll_layer_id,
                                                         pipeline_id);
         self.stacking_context_store.push(StackingContext::new(pipeline_id,
                                                               reference_frame_offset,
                                                               rect,
+                                                              is_page_root,
                                                               composite_ops,
                                                               group_index));
         self.cmds.push(PrimitiveRunCmd::PushStackingContext(stacking_context_index));
@@ -218,6 +236,7 @@ impl FrameBuilder {
 
     pub fn pop_stacking_context(&mut self) {
         self.cmds.push(PrimitiveRunCmd::PopStackingContext);
+        self.stacking_context_stack.pop();
     }
 
     pub fn push_clip_scroll_node(&mut self,
@@ -941,6 +960,7 @@ impl FrameBuilder {
         let mut sc_stack = Vec::new();
         let mut current_task = RenderTask::new_alpha_batch(next_task_index,
                                                            DeviceIntPoint::zero(),
+                                                           false,
                                                            RenderTaskLocation::Fixed);
         next_task_index.0 += 1;
         let mut alpha_task_stack = Vec::new();
@@ -957,10 +977,23 @@ impl FrameBuilder {
 
                     let stacking_context_rect = &stacking_context.bounding_rect;
                     let composite_count = stacking_context.composite_ops.count();
+
+                    if composite_count == 0 && stacking_context.should_isolate {
+                        let location = RenderTaskLocation::Dynamic(None, stacking_context_rect.size);
+                        let new_task = RenderTask::new_alpha_batch(next_task_index,
+                                                                   stacking_context_rect.origin,
+                                                                   stacking_context.should_isolate,
+                                                                   location);
+                        next_task_index.0 += 1;
+                        let prev_task = mem::replace(&mut current_task, new_task);
+                        alpha_task_stack.push(prev_task);
+                    }
+
                     for _ in 0..composite_count {
                         let location = RenderTaskLocation::Dynamic(None, stacking_context_rect.size);
                         let new_task = RenderTask::new_alpha_batch(next_task_index,
                                                                    stacking_context_rect.origin,
+                                                                   stacking_context.should_isolate,
                                                                    location);
                         next_task_index.0 += 1;
                         let prev_task = mem::replace(&mut current_task, new_task);
@@ -975,6 +1008,20 @@ impl FrameBuilder {
                         continue;
                     }
 
+                    let composite_count = stacking_context.composite_ops.count();
+
+                    if composite_count == 0 && stacking_context.should_isolate {
+                        let mut prev_task = alpha_task_stack.pop().unwrap();
+                        let item = AlphaRenderItem::HardwareComposite(stacking_context_index,
+                                                                      current_task.id,
+                                                                      HardwareCompositeOp::Alpha,
+                                                                      next_z);
+                        next_z += 1;
+                        prev_task.as_alpha_batch().alpha_items.push(item);
+                        prev_task.children.push(current_task);
+                        current_task = prev_task;
+                    }
+
                     for filter in &stacking_context.composite_ops.filters {
                         let mut prev_task = alpha_task_stack.pop().unwrap();
                         let item = AlphaRenderItem::Blend(stacking_context_index,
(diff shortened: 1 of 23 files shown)


glennw:
@rlhunt I'm not quite sure I understand this - each stacking context in a mix-blend-mode case should already be isolated due to it being in a render target, shouldn't it?

I'm probably just missing something obvious here - but are you able to construct a HTML file that demonstrates this with a different rendering between Servo and Gecko perhaps?

rlhunt:
Yeah I can! Not sure the best way to link to a html file, so I'll just embed it. I adapted this from a gecko reftest.

``` html
<!DOCTYPE html>
<html>

<style type="text/css">
	.yellow {
		background-color: rgb(255, 255, 0);
	}
	.green {
		background-color: rgb(0, 255, 0);
	}
	.box {
		width: 100px;
		height: 100px;
	}
	.mbm {
		mix-blend-mode: difference;
	}
	.stacking-context {
		will-change: opacity;
	}
</style>

<!-- should see just the green box hiding the yellow box -->
<!-- the stacking context should create a new transparent backdrop for its children with mix-blend-mode -->
<!-- if this doesn't happen the green box will do difference with yellow creating red -->
<div class="yellow box">
	<div class="stacking-context">
		<div class="green box mbm">
		</div>
	</div>
</div>

</html>
```

emilio:
Note that this may also be a servo bug, since the `.stacking-context { will-change: opacity }` in Servo does _not_ force the creation of a stacking context. Display list dump:

```
┌ Display List
│  ├─ Items
│  │  ├─ PushStackingContext(StackingContext at TypedRect(1024px×740px at (0px,0px)) with overflow TypedRect(1024px×740px at (0px,0px)): StackingContextId(0)) StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 0, 0, 0) @ TypedRect(1024px×740px at (0px,0px)) ClippingRegion::Max StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 0, 0, 0) @ TypedRect(1008px×100px at (8px,8px)) ClippingRegion::Max StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(1, 1, 0, 1) @ TypedRect(100px×100px at (8px,8px)) ClippingRegion::Max StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 0, 0, 0) @ TypedRect(100px×100px at (8px,8px)) ClippingRegion::Max StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ PushStackingContext(StackingContext at TypedRect(100px×100px at (8px,8px)) with overflow TypedRect(100px×100px at (0px,0px)): StackingContextId(140394916598976)) StackingContext: StackingContextId(140394916598976) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 1, 0, 1) @ TypedRect(100px×100px at (0px,0px)) ClippingRegion::Max StackingContext: StackingContextId(140394916598976) ScrollRoot: ScrollRootId(0)
│  │  ├─ PopStackingContext(StackingContextId(140394916598976) StackingContext: StackingContextId(140394916598976) ScrollRoot: ScrollRootId(0)
│  │  └─ PopStackingContext(StackingContextId(0) StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
```

rlhunt:
Okay, I'm able to observe the same differences between servo and gecko by changing to `.stacking-context { position: absolute; z-order: 1 }` 

```
┌ Display List
│  ├─ Items
│  │  ├─ PushStackingContext(StackingContext at TypedRect(1024px×740px at (0px,0px)) with overflow TypedRect(1024px×740px at (0px,0px)): StackingContextId(0)) StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 0, 0, 0) @ TypedRect(1024px×740px at (0px,0px)) ClippingRegion::Max StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 0, 0, 0) @ TypedRect(1008px×100px at (8px,8px)) ClippingRegion::Max StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(1, 1, 0, 1) @ TypedRect(100px×100px at (8px,8px)) ClippingRegion::Max StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
│  │  ├─ PushStackingContext(StackingContext at TypedRect(100px×100px at (8px,8px)) with overflow TypedRect(100px×100px at (0px,0px)): StackingContextId(140413524628528)) StackingContext: StackingContextId(140413524628528) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 0, 0, 0) @ TypedRect(100px×100px at (0px,0px)) ClippingRegion::Max StackingContext: StackingContextId(140413524628528) ScrollRoot: ScrollRootId(0)
│  │  ├─ PushStackingContext(StackingContext at TypedRect(100px×100px at (0px,0px)) with overflow TypedRect(100px×100px at (0px,0px)): StackingContextId(140413524628672)) StackingContext: StackingContextId(140413524628672) ScrollRoot: ScrollRootId(0)
│  │  ├─ SolidColor rgba(0, 1, 0, 1) @ TypedRect(100px×100px at (0px,0px)) ClippingRegion::Max StackingContext: StackingContextId(140413524628672) ScrollRoot: ScrollRootId(0)
│  │  ├─ PopStackingContext(StackingContextId(140413524628672) StackingContext: StackingContextId(140413524628672) ScrollRoot: ScrollRootId(0)
│  │  ├─ PopStackingContext(StackingContextId(140413524628528) StackingContext: StackingContextId(140413524628528) ScrollRoot: ScrollRootId(0)
│  │  └─ PopStackingContext(StackingContextId(0) StackingContext: StackingContextId(0) ScrollRoot: ScrollRootId(0)
```

glennw:
Ah, I see what you mean now, thanks for the example! I'll take a look over the PR today.

glennw:
The general idea here seems quite good, I think. It seems a little unfortunate to have to modify frame.rs and the conditions used are also a bit confusing (I guess the next_z == 0 is a way of checking if at the root?).

I wonder if we could remove the frame.rs changes, and handle this all internally within FrameBuilder? I'd be completely fine with FrameBuilder internally keeping a stack of the current stacking context indices. I think that would make it feasible to implement completely inside FrameBuilder?

It also might make sense to make the concept of whether a stacking context needs to be isolated a little bit more generic, rather than naming things specifically related to having mix blend mode children?

Sorry that's a bit vague - just some initial thoughts. Thanks for looking into this!

glennw:
@rlhunt Do the comments / questions above make sense? We could discuss on IRC / Vidyo if that would be easier?

rlhunt:
@glennw Yes they do make sense! I agree that the conditions could be more clear and cleaned up. I like the idea of moving this to FrameBuilder. It can probably be done completely in there.

I've been side tracked with other tasks, but I'll try and take a look at this sometime soon!

glennw:
No worries, thanks for looking into this stuff! :)

rlhunt:
@glennw Okay, I've rebased this and cleaned it up. Now all the logic for when to isolate a stacking context happens in frame_builder.rs, and I've tried to add a few comments to explain what 'should_isolate' means and the conditions for it.

webrender/src/frame_builder.rs
```diff
@@ -157,6 +159,17 @@ impl FrameBuilder {
                                  scroll_layer_id: ScrollLayerId,
                                  composite_ops: CompositeOps) {
         let stacking_context_index = StackingContextIndex(self.stacking_context_store.len());
+        let parent_index = self.stacking_context_stack.last().map(|x| x.0);
+        self.stacking_context_stack.push(stacking_context_index);
+
+        if let Some(parent_index) = parent_index {
+          if composite_ops.mix_blend_mode.is_some() {
```
glennw: nit: should be 4 space indent

glennw:
@rlhunt OK, seeing some test failures that appear related:

```
  ▶ FAIL [expected PASS] /compositing-1_dev/html/compositing_simple_div.htm
  └   → /compositing-1_dev/html/compositing_simple_div.htm 13ee95e760e850374940f5c9633148e0574e0287
/compositing-1_dev/html/reference/compositing_simple_div-ref.htm 66bf04c1597f2bdb7a831172129d16d8e21bea17
Testing 13ee95e760e850374940f5c9633148e0574e0287 == 66bf04c1597f2bdb7a831172129d16d8e21bea17

  ▶ FAIL [expected PASS] /compositing-1_dev/html/mix-blend-mode-stacking-context-creates-isolation.htm
  └   → /compositing-1_dev/html/mix-blend-mode-stacking-context-creates-isolation.htm 33fee4db992ec9f203e9c4b45a7e83ebea8fbcfe
/compositing-1_dev/html/reference/mix-blend-mode-stacking-context-creates-isolation-ref.htm e712a73bb530ce05bad070100c1831f6795f6bcd
Testing 33fee4db992ec9f203e9c4b45a7e83ebea8fbcfe == e712a73bb530ce05bad070100c1831f6795f6bcd
```

You can investigate these locally by building Servo with a local WR override, and then running:

```
./mach test-css --release --processes=6
```

glennw:
(I should note - it's possible that these were accidental passes previously - I haven't compared the reftest output before / after this PR on those tests).

rlhunt:
Thanks for running the tests! I'll take a look at those failures.

rlhunt:
Okay, so I think I've figured out the test failures. I'll update with a patch sometime soon.

rlhunt:
So by soon I guess I meant almost a week :)

There are two major problems
1. We aren't taking the backdrop's alpha into account ([spec](https://drafts.fxtf.org/compositing-1/#generalformula))
2. The backdrop readback is upside down when we aren't doing readback from the framebuffer

I've updated and rebased to take both into account. I've removed the hardware blend mode path because I couldn't think of how to incorporate the backdrop's alpha. I ran it locally on a build of servo and it didn't seem to have any related failures.

rlhunt:
I should also note that '/compositing-1_dev/html/compositing_simple_div.htm' displays differently in Chrome vs Firefox. Chrome has the div blend with the canvas background while Firefox does not.

Chrome seems to have the expected behavior for the test so that is the what I went with.

glennw:
@bors-servo r+

Thanks for following up on this!

PR #870, opened 2017-02-13
Further improvements to pixel snapping.

glennw:
The previous snapping update didn't correctly handle the case
where the size in local space had a fractional component (it
could result in a slight stretch of the UV surface).

Now, determine the snap distance based on the position of the
top left corner of the quad. This snap distance gets applied to
all vertices.

In the common case (size has no fractional component) this gives
the same result as the previous pixel snapping PR (since the
fractional distance from each vertex is the same). In the less common
case, this snaps the quad to a device pixel, but leaves the size of
the quad with a fractional component, which avoids any UV stretching.

This makes several new tests pass in the WPT/CSS test suite, and does
not cause any additional failures.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/870)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -422,42 +422,49 @@ VertexInfo write_vertex(vec4 instance_rect,
                         float z,
                         Layer layer,
                         AlphaBatchTask task) {
-    vec2 p0 = instance_rect.xy;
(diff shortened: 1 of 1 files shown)


kvark:
It makes sense and looks good!
@bors-servo r+

PR #871, opened 2017-02-13
Fix the image tex-coord clamping

kvark:
The invalid argument order we use now would still clip the lower bound but ignore the higher bound.

r? @glennw 
cc @jrmuizel

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/871)


glennw:
Oops!

PR #872, opened 2017-02-13
Avoid the fractional GPU data in texture rows

### LINKED ISSUE #865: Some images are not rendered
In http://people.mozilla.org/~jmuizelaar/tmp/buggy.zip there should be colored images in all of the boxes. The last two images are not rendered. Commenting out some of the earlier images makes the the last two ones appear.

This is causing reftest regressions in Gecko.

kvark:
Fixes #865

r? @glennw 
cc @jrmuizel

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/872)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -40,8 +40,9 @@ varying vec3 vClipMaskUv;
 
 #ifdef WR_VERTEX_SHADER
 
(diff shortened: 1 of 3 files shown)


glennw:
This used to be explicitly handled elsewhere, so this shouldn't have been necessary. But since it's blocking an update, let's merge it and investigate further in a followup.

PR #877, opened 2017-02-14
Make it possible to tweak the maximum texture size.

nical:
As a prelude to the work on supporting very large images, this pull request makes it possible to lower the maximum texture size. By default the maximum size is the one reported by GL and we can lower it with a setting in the RendererOptions (but not increase it).

This should make it a lot easier to test support for large images, will give us the possibility to cap the maximum size for drivers that are too optimistic for their own sake, and also make it easier to reproduce bugs happening on hardware with lower capabilities.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/877)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -42,12 +42,6 @@ const SHADER_VERSION: &'static str = "#version 300 es\n";
 
 static SHADER_PREAMBLE: &'static str = "shared";
 
-lazy_static! {
-    pub static ref MAX_TEXTURE_SIZE: gl::GLint = {
-        gl::get_integer_v(gl::MAX_TEXTURE_SIZE)
-    };
-}
-
 #[repr(u32)]
 pub enum DepthFunction {
     Less = gl::LESS,
@@ -825,6 +819,8 @@ pub struct Device {
     // Used on android only
     #[allow(dead_code)]
     next_vao_id: gl::GLuint,
+
+    max_texture_size: u32,
 }
 
 impl Device {
@@ -863,7 +859,13 @@ impl Device {
 
             next_vao_id: 1,
             //file_watcher: file_watcher,
+
+            max_texture_size: gl::get_integer_v(gl::MAX_TEXTURE_SIZE) as u32
         }
+    }
+
+    pub fn max_texture_size(&self) -> u32 {
+        self.max_texture_size
     }
 
     pub fn get_capabilities(&self) -> &Capabilities {
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -675,7 +675,10 @@ impl Renderer {
                                      options.precache_shaders)
         };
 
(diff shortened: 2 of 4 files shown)


webrender/src/device.rs
```diff
@@ -863,9 +859,13 @@ impl Device {
 
             next_vao_id: 1,
             //file_watcher: file_watcher,
+
+            max_texture_size: gl::get_integer_v(gl::MAX_TEXTURE_SIZE) as u32
         }
     }
 
+    pub fn max_texture_size(&self) -> u32 { self.max_texture_size }
```
glennw: nit: There should be a new line here (unless there is something in the rust style guide specifically about this?).

webrender/src/resource_cache.rs
```diff
@@ -235,6 +235,8 @@ impl ResourceCache {
         }
     }
 
+    pub fn max_texture_size(&self) -> u32 { self.texture_cache.max_texture_size() }
```
glennw: nit: newline

webrender/src/renderer.rs
```diff
@@ -675,7 +675,12 @@ impl Renderer {
                                      options.precache_shaders)
         };
 
-        let mut texture_cache = TextureCache::new();
+        let device_max_size = device.max_texture_size();
+        let max_texture_size = options.max_texture_size
```
glennw: nit: This can avoid the map with something like:

```
cmp::min(device_max_size, options.max_texture_size.unwrap_or(device_max_size))
```

webrender/src/texture_cache.rs
```diff
@@ -556,16 +556,23 @@ pub struct AllocationResult {
 }
 
 impl TextureCache {
-    pub fn new() -> TextureCache {
+    pub fn new(mut max_texture_size: u32) -> TextureCache {
+        if max_texture_size * max_texture_size > MAX_RGBA_PIXELS_PER_TEXTURE {
```
glennw: nit: could probably just be a cmp::min()

webrender/src/texture_cache.rs
```diff
@@ -556,16 +556,23 @@ pub struct AllocationResult {
...
             cache_id_list: CacheTextureIdList::new(),
             free_texture_levels: HashMap::with_hasher(Default::default()),
             items: FreeList::new(),
             pending_updates: TextureUpdateList::new(),
             arena: TextureCacheArena::new(),
+            max_texture_size: max_texture_size,
         }
     }
 
+    pub fn max_texture_size(&self) -> u32 { self.max_texture_size }
```
glennw: nit: newline

glennw:
Looks good, let's squash these and then r=me

glennw:
This seems to have an extra commit in it now? (The separate out debug/release builds commit).

glennw:
r=me once that stray commit is removed.

PR #882, opened 2017-02-14
Separate out debug and release builds.

jrmuizel:
This should let us run them in parallel.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/882)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,15 +17,18 @@ addons:
       - libgles2-mesa-dev
       - llvm-3.8-dev
       - libedit-dev
+env:
+  - BUILD_KIND=DEBUG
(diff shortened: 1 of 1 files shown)


glennw:
Looks good to me - let's see what travis makes of it and check the reported build times.

jrmuizel:
Looks like this needs some modification to not fail

glennw:
@jrmuizel I guess the potential problem with this is that we now need 2 mac builds per run, which might be a problem since there is often a mac builder backlog? Happy to merge this anyway and see how it goes, if you like?

jrmuizel:
Yeah, let's merge it and see how things go.

Issue #880, opened 2017-02-14
Add Taskcluster CI for webrender


staktrace:
@jrmuizel wanted to try running webrender CI on taskcluster instead of travis to maybe speed things up. Doing this is fairly straightforward, see http://code.v.igoro.us/posts/2017/02/taskcluster-github-improvements.html and https://tools.taskcluster.net/quickstart/.

I built a docker image and taskcluster script to try this out, and have a sample CI run currently in progress at https://tools.taskcluster.net/task-inspector/#Xq6XN5JFS5qX2gV6BeFp5A/0 - if that goes well we should turn on the taskcluster-github integration and land the .taskcluster.yml file into the repo. That will get us CI running on taskcluster, and we can then work on optimizing it further by splitting up the tests to run in different tasks, trying to cache more build artifacts from dependencies, etc.

staktrace:
The CI run took 34 minutes and completed successfully.

staktrace:
I also tried splitting the debug vs release tests. You can see it in the task group at https://tools.taskcluster.net/task-group-inspector/#/MVjO8EHCQUS1FVfM8Bb6uA - the debug stuff took ~17 minutes and the release-mode build and reftests took ~35 minutes. So splitting it doesn't seem to really help much, at least on taskcluster.

kvark:
And since this is about using docker, we might want to CI build for Android as well:
https://github.com/tomaka/android-rs-glue#with-docker

glennw:
Are we still wanting to explore this?

staktrace:
Let's close it for now. If somebody cares enough they can reopen it.

PR #883, opened 2017-02-14
Add a .taskcluster.yml file to run CI using taskcluster

staktrace:
Part of the fix for #880 (the other part is enabling the integration service, which an owner of the repo has to do). We might want to hold off on actually doing this since I think @jrmuizel is looking into speeding up the travis run by running things in parallel.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/883)


glennw:
@staktrace Do you want this merged now?

staktrace:
@glennw No, let's wait and see how much #882 helps. If we can get it fast enough using travis we might not need to use taskcluster at all.

glennw:
I think even if #882 helps, we still want to consider taskcluster. Although travis runs are quite slow, the major issue is the mac builder backlog on travis. This means that it often takes 1-2 hours before jobs are even scheduled to run on a mac builder.

metajack:
Is there a patch to homu that goes with this? Because otherwise this is going to break our bots.

In general, I think we may have the ability to guarantee mac availability on Travis CI. @larsbergstrom?

staktrace:
No patch to homu yet. As a first step I just wanted to stand up Taskcluster CI alongside Travis.

I made a docker image with a precompiled webrender (both debug and release) in it already. That speeds up the test run so that it completes in about 10 minutes, see https://tools.taskcluster.net/task-group-inspector/#/IWmDQd0VSLi_YjBCNBhA2w for example. However if there is rust version bump or if one of the webrender dependencies changes then it'll fall back to rebuilding everything which is slow again. Also the 10 minutes can probably be shaved down further if we prebuild different `--feature` options. I just did debug/release with no `--feature` options.

larsbergstrom:
It's possible to guarantee mac availablity on Travis CI, but we've avoided it on Servo to date due to some pretty heavy budget impact (at least, in terms the Research team thinks of). Travis is also scheduled to increase their Mac capacity significantly in the near future, but it was delayed by the recent macstadium meltdown.

All that said, I'm not against using TC for some of the builds, too, so long as they have capacity and it doesn't "fight" with homu's github build status checking. Hopefully the initial version would just add some flags that homu would ignore while details are worked out.

jrmuizel:
Now that we don't depend on syntex, builds should be faster too. Should we just turn this on?

glennw:
It sounds from the comment by @metajack above that enabling this would break homu?

metajack:
If we want to replace travis, we'll need homu to start gating on taskcluster instead of travis. I have no idea if it is capable of that right now, or if it would be difficult to add.

staktrace:
I think we should be ok to merge this PR and turn on the taskcluster-github integration on servo/webrender. It will just add another status report which should basically be ignored by homu (I say this without knowing anything about how homu works really, but I would be very surprised if it barfed on unrecognized status annotations).

That being said, merging the PR is only really useful if we plan to go beyond what it's doing now, and run tests on Mac as well. Right now it just does linux, and I'm not sure what steps are needed to make it also run on Mac. I presume we need to add another task in the .taskcluster.yml file that uses a different provisioner to get a mac machine, and then has a different mac-specific image and steps. I don't recall who @jrmuizel talked to on releng to discover that mac hardware is available on TC, but we should also find out how to hook it up.

jrmuizel:
The next thing we should hook up after this is doing mac builds on linux. That's what we do for Firefox and will be much cheaper than doing the mac builds on mac hardware.

metajack:
To be clear, you mean building on Linux for Mac and then testing the result on the Mac?

jrmuizel:
That would be the goal yes.

glennw:
As of last week, Travis appears to have (nearly) doubled their mac builder capacity, which has made the mac wait time come down a lot. Given that, and the way that we now split the CI jobs across multiple machines, I'm tempted to close this for now. Thoughts?

jrmuizel:
I'd like to keep pushing this forward even if it's only at low priority. I think it will give us the potential to tighten up the turn around times a lot more and helps avoid getting caught if the load on Travis increases to match the new capacity.

glennw:
@staktrace @jrmuizel @metajack Do we want to merge this PR as is now? If not, I'll probably close it for now until we have something ready to be reviewed and merged.

staktrace:
I'm ok with closing it. I don't see the point in merging it as-is because it will just waste TC resources.

PR #885, opened 2017-02-15
Fix regression in vertex texture size calculation.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/885)


glennw:
r? @kvark 

This is the fix we discussed on IRC - and includes a revert for the temporary workaround.

kvark:
I actually expected a different kind of fix. Given a fixed texture width, I thought we'd check during allocation if the chunk is linear and otherwise pad it to the next texture row.

This fix is good though. Thanks!
@bors-servo r+

PR #886, opened 2017-02-15
Fix the UV rect clamping for image masks.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/886)


glennw:
r? @kvark 

cc @jrmuizel

glennw:
As a follow up: https://github.com/servo/webrender/issues/887

kvark:
Uh oh, I knew those rectangles would bite us but didn't expect to fall into that trap myself.
Thanks for catching this!
@bors-servo r+

PR #890, opened 2017-02-16
Add ClipScrollGroup

mrobinson:
ClipScrollGroups represent a portion of a stacking context that is
clipped and scrolled. Before we thought of the entire contents of a
StackingContext as clipped and scrolled in the same way. This isn't the
case on the web, so we add preliminary support for the concept, while
maintaining the 1-to-1 relationship between them for the moment.

Parts of this PR are based on a proof of concept by Glenn Watson.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/890)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -20,10 +20,10 @@ use render_task::RenderTaskLocation;
 use resource_cache::ResourceCache;
 use scroll_tree::ScrollTree;
 use std::{cmp, f32, i32, mem, usize};
-use tiling::{AuxiliaryListsMap, CompositeOps, Frame, PackedLayer, PackedLayerIndex};
-use tiling::{PrimitiveFlags, PrimitiveRunCmd, RenderPass, RenderTargetContext};
-use tiling::{RenderTaskCollection, ScrollbarPrimitive, ScrollLayer, ScrollLayerIndex};
-use tiling::{StackingContext, StackingContextIndex};
+use tiling::{AuxiliaryListsMap, ClipScrollGroup, ClipScrollGroupIndex, CompositeOps, Frame};
+use tiling::{PackedLayer, PackedLayerIndex, PrimitiveFlags, PrimitiveRunCmd, RenderPass};
+use tiling::{RenderTargetContext, RenderTaskCollection, ScrollbarPrimitive, ScrollLayer};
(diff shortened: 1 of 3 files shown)


mrobinson:
@glennw r?

glennw:
I tested this locally on several sites and also ran the WPT/CSS tests locally - these all seem fine.

I didn't have time to take a detailed review over the PR yet, I'll do that on Monday. From a quick look it seems quite good. I think it never clears the visibility of the clip/scroll groups though, so once it gets noted visible it will always be drawn? I could just be missing something though - you can check this by having a test with a div at the top of the page and make sure it stops drawing it once you scroll that off the page.

I'm not a huge fan of ClipScrollGroup as a name, but I don't have any better suggestions right now!

mrobinson:
@glennw Thanks for the review! You're right that visibility wasn't taking into account the clipped scroll group's xf_rect property. I've fixed that. Additionally, it should be possible to start skipping stacking contexts even if their contents don't intersect with the screen rect. For example, to make your test case work, I needed to ensure that the div had it's own stacking context in order to avoid having it a child of the full page stacking context. The painting of the div wouldn't be avoided in the latter case, because the stacking context is the size of the entire page.

I also dislike the name I chose (ClipScrollGroup). I do like the sound of ReferenceFrame, but I think it conflicts with the Gecko GFX naming and I'm afraid that will cause confusion. Here are some other names that I'm batting around. If any of them sound good or even better than ClipScrollGroup, I'll happily change it.

- Frame (as opposed to a ReferenceFrame, which is used as a reference for other frames)
- ItemFrame
- FramedItems
- RenderSetup
- ClippedScrolledItems
- ClippedScrolledGroups
- ClippedItems
- ScrolledItems
- ItemGroup
- Layer (I plan to change the current Layer to be ClipScrollNode and ScrollTree to ClipScrollTree)
- StackingContextSlice
- StackingContextPiece
- ScrolledStackingContext
- ClippedStackingContext

mrobinson:
I plan to try out the improved stacking context skipping in another patch, because it changes behavior a bit and I want to keep things simple.

glennw:
@bors-servo r+

Nice! Let's get this merged now. We should revisit the naming of ClipScrollGroup - but that can be done in a follow up since this is all internal API only.

PR #891, opened 2017-02-16
Add some simple support for premultiply and unpremultiply

jrmuizel:
This will be useful for producers and consumers of image data

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/891)


jrmuizel:
This is not going to work that easily with u8's so let's just do a dumber thing for now.

PR #897, opened 2017-02-17
Tiled image support

nical:
Here is a pretty straightforward implementation of image tiling which aims at being able to render images (and vector images) that are larger than the maximum texture size.

The change mostly boils down to:
 - making it possible to upload textures with an offset in the source buffer
 - identifying individual tiles in the resource cache (See ```ImageRequest::tile```)
 - break up image display items into several primitives when the image is tiled

The current limitation is that repeating images isn't supported with tiling yet. That will need to be fixed. It should only be a matter of extending the code in frame.rs so that it generate a set of image primitive for each repetition.

I have been changing my mind regularly on the topic of whether tiling should be requested explicitly in the API or if it should be done somewhat automatically inside of WebRender. Currently I think that making this explicit in the API is simpler to implement and also simpler to test, so that's what I ended up implementing. It would probably make sense to also automatically tile images that are larger than the maximum texture size even if tiling isn't requested, because there isn't any way to render them otherwise.
I don't have a particular attachment to the way I exposed it in the api (added ```RenderApi::add_image_with_tiling``` rather than passing an extra parameter in add_image to avoid changing the api), let me know if you have a preference.

I am quite happy with how this works compared to what we have in gecko's layers, because tiles are uploaded on-demand which means we don't need to pay unnecessary cost for large images that are partially visible.

The separation between the commits isn't too bad so I kept it this way in case it helps with the review, but don't hesitate to ask me to squash it into one commit before the review if you prefer. Also the PR is rebased on top of PR #877 which makes more code appear here than it should, but this will clear out when the latter PR lands.

r? @glennw @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/897)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -105,7 +105,7 @@ fn main() {
     let vector_img = api.generate_image_key();
     api.add_image(
         vector_img,
-        ImageDescriptor::new(100, 100, ImageFormat::RGBA8).with_opaque_flag(true),
+        ImageDescriptor::new(100, 100, ImageFormat::RGBA8, true),
         ImageData::new_blob_image(Vec::new()),
         None,
     );
@@ -143,7 +143,7 @@ fn main() {
         let mask_image = api.generate_image_key();
         api.add_image(
             mask_image,
-            ImageDescriptor::new(2, 2, ImageFormat::A8).with_opaque_flag(true),
+            ImageDescriptor::new(2, 2, ImageFormat::A8, true),
             ImageData::new(vec![0, 80, 180, 255]),
             None,
         );
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -107,6 +107,7 @@ fn main() {
         vector_img,
         ImageDescriptor::new(100, 100, ImageFormat::RGBA8).with_opaque_flag(true),
         ImageData::new_blob_image(Vec::new()),
+        None,
     );
 
     let pipeline_id = PipelineId(0, 0);
@@ -143,7 +144,8 @@ fn main() {
         api.add_image(
             mask_image,
             ImageDescriptor::new(2, 2, ImageFormat::A8).with_opaque_flag(true),
-            ImageData::new(vec![0, 80, 180, 255])
+            ImageData::new(vec![0, 80, 180, 255]),
+            None,
         );
         let mask = webrender_traits::ImageMask {
             image: mask_image,
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -111,6 +111,7 @@ fn main() {
             height: 100,
             stride: None,
             is_opaque: true,
+            offset: 0,
         },
         ImageData::new_blob_image(Vec::new()),
     );
@@ -154,6 +155,7 @@ fn main() {
                 stride: None,
                 format: ImageFormat::A8,
                 is_opaque: true,
+                offset: 0,
             },
             ImageData::new(vec![0, 80, 180, 255])
         );
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -105,14 +105,7 @@ fn main() {
     let vector_img = api.generate_image_key();
     api.add_image(
         vector_img,
-        ImageDescriptor {
-            format: ImageFormat::RGBA8,
-            width: 100,
-            height: 100,
-            stride: None,
-            is_opaque: true,
-            offset: 0,
-        },
+        ImageDescriptor::new(100, 100, ImageFormat::RGBA8).with_opaque_flag(true),
         ImageData::new_blob_image(Vec::new()),
     );
 
@@ -149,14 +142,7 @@ fn main() {
         let mask_image = api.generate_image_key();
         api.add_image(
             mask_image,
-            ImageDescriptor {
-                width: 2,
-                height: 2,
-                stride: None,
-                format: ImageFormat::A8,
-                is_opaque: true,
-                offset: 0,
-            },
+            ImageDescriptor::new(2, 2, ImageFormat::A8).with_opaque_flag(true),
             ImageData::new(vec![0, 80, 180, 255])
         );
         let mask = webrender_traits::ImageMask {
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -1631,7 +1631,12 @@ impl Device {
             None => width,
         };
 
-        assert!(data.len() as u32 == bpp * row_length * height);
+        // Take the stride into account for all rows, except the last one.
+        let len = bpp * row_length * (height - 1)
+                + width * bpp;
+
+        assert!(data.len() as u32 >= len);
+        let data = &data[0..len as usize];
 
         if let Some(..) = stride {
             gl::pixel_store_i(gl::UNPACK_ROW_LENGTH, row_length as gl::GLint);
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -29,6 +29,7 @@ static DEFAULT_SCROLLBAR_COLOR: ColorF = ColorF { r: 0.3, g: 0.3, b: 0.3, a: 0.6
 struct FlattenContext<'a> {
     scene: &'a Scene,
     builder: &'a mut FrameBuilder,
+    resource_cache: &'a mut ResourceCache,
 }
 
 // TODO: doc
@@ -276,6 +277,7 @@ impl Frame {
             let mut context = FlattenContext {
                 scene: scene,
                 builder: &mut frame_builder,
+                resource_cache: resource_cache
             };
 
             let mut traversal = DisplayListTraversal::new_skipping_first(display_list);
@@ -295,7 +297,6 @@ impl Frame {
                                                   &root_clip.main.size);
 
             self.flatten_stacking_context(&mut traversal,
-                                          resource_cache,
                                           root_pipeline_id,
                                           &mut context,
                                           reference_frame_id,
@@ -315,7 +316,6 @@ impl Frame {
 
     fn flatten_scroll_layer<'a>(&mut self,
                                 traversal: &mut DisplayListTraversal<'a>,
-                                resource_cache: &mut ResourceCache,
                                 pipeline_id: PipelineId,
                                 context: &mut FlattenContext,
                                 current_reference_frame_id: ScrollLayerId,
@@ -343,7 +343,6 @@ impl Frame {
                                               &content_size);
 
         self.flatten_items(traversal,
-                           resource_cache,
                            pipeline_id,
                            context,
                            current_reference_frame_id,
@@ -356,7 +355,6 @@ impl Frame {
 
     fn flatten_stacking_context<'a>(&mut self,
                                     traversal: &mut DisplayListTraversal<'a>,
-                                    resource_cache: &mut ResourceCache,
                                     pipeline_id: PipelineId,
                                     context: &mut FlattenContext,
                                     current_reference_frame_id: ScrollLayerId,
@@ -446,7 +444,6 @@ impl Frame {
                                               composition_operations);
 
         self.flatten_items(traversal,
-                           resource_cache,
                            pipeline_id,
                            context,
                            reference_frame_id,
@@ -467,7 +464,6 @@ impl Frame {
     }
 
     fn flatten_iframe<'a>(&mut self,
-                          resource_cache: &mut ResourceCache,
                           pipeline_id: PipelineId,
                           bounds: &LayerRect,
                           context: &mut FlattenContext,
@@ -525,7 +521,6 @@ impl Frame {
         let mut traversal = DisplayListTraversal::new_skipping_first(display_list);
 
         self.flatten_stacking_context(&mut traversal,
-                                      resource_cache,
                                       pipeline_id,
                                       context,
                                       iframe_reference_frame_id,
@@ -541,7 +536,6 @@ impl Frame {
 
     fn flatten_items<'a>(&mut self,
                          traversal: &mut DisplayListTraversal<'a>,
-                         resource_cache: &mut ResourceCache,
                          pipeline_id: PipelineId,
                          context: &mut FlattenContext,
                          current_reference_frame_id: ScrollLayerId,
@@ -555,7 +549,7 @@ impl Frame {
                                                         &item.clip, info.context_id);
                 }
                 SpecificDisplayItem::Image(ref info) => {
-                    let image = resource_cache.get_image_properties(info.image_key);
+                    let image = context.resource_cache.get_image_properties(info.image_key);
                     if let Some(tile_size) = image.tiling {
                         // The image resource is tiled. We have to generate an image primitive
                         // for each tile.
@@ -629,7 +623,6 @@ impl Frame {
                 }
                 SpecificDisplayItem::PushStackingContext(ref info) => {
                     self.flatten_stacking_context(traversal,
-                                                  resource_cache,
                                                   pipeline_id,
                                                   context,
                                                   current_reference_frame_id,
@@ -641,7 +634,6 @@ impl Frame {
                 }
                 SpecificDisplayItem::PushScrollLayer(ref info) => {
                     self.flatten_scroll_layer(traversal,
-                                              resource_cache,
                                               pipeline_id,
                                               context,
                                               current_reference_frame_id,
@@ -653,8 +645,7 @@ impl Frame {
                                               info.id);
                 }
                 SpecificDisplayItem::Iframe(ref info) => {
-                    self.flatten_iframe(resource_cache,
-                                        info.pipeline_id,
+                    self.flatten_iframe(info.pipeline_id,
                                         &item.rect,
                                         context,
                                         current_scroll_layer_id,
@@ -674,7 +665,7 @@ impl Frame {
                              tile_size: u32) {
         // The image resource is tiled. We have to generate an image primitive
         // for each tile.
-        // We need to do this because the image is borken up into smaller tiles in the texture
+        // We need to do this because the image is broken up into smaller tiles in the texture
         // cache and the image shader is not able to work with this type of sparse representation.
 
         // The tiling logic works as follows:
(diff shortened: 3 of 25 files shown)


nical:
I added support for stretching tiled images, as well as a special case to support repeating an image along an axis if the image is not tiled along that axis (or in other words if that image fits in a single tile for this particular axis).
Although the latter sounds like a very specific use case, we talked about it in the graphics meeting and it turns out web authors will some times create very long and thin images (say, 20000 by 5 px) as a repeating background. In this scenario, generating a set of image primitives for each repetition along the y axis would add up to a *lot* of primitives so I added support for repeating along the non-tiled axis to handle this.

Remains the general case case where we want to repeat a tiled image along an axis where it is tiled, which I have not implemented yet. However I think that I have already put way too much in one pull request so I won't add any more features in this one. Let's deal with the remaining cases later.

nical:
I just cleaned up and rebased this patch queue.
Most of the important work is in the commit "Implement image tiling". I also implemented support for tiling in wrench and added some very basic reftests (in the process of which I failed at not adding more features to this PR, sorry!).

The PR is in a pretty good state for review now.

The current limitations that come to mind are:
 - Repeating a tiled image is not yet supported if the image is tiled along that axis (for example if an image is very tall but its width fits in a tile, we can repeat it along the x axis but not the y axis).
 - Pixel snapping being applied per primitive, each tile gets snapped which can produce slightly different results compared to a non-tiled image. The difference is hardly noticeable but makes it hard to write proper reftests if the image and tile sizes don't line up in certain ways. @kvark suggested a solution for this which I haven't tried out yet,
 - With some tile sizes I sometimes get a 1px seam between tiles, which I think is related to rounding differences + pixel snapping.

I'll look into these as followup items.

kvark:
I finished the first pass, got some notes. Definitely wish to do a second pass after your responses.

> I have been changing my mind regularly on the topic of whether tiling should be requested explicitly in the API or if it should be done somewhat automatically inside of WebRender.

I wish there was a clean choice between being explicit/implicit. If we go the explicit route, maybe we can declare a maximum image size and requires the client to do tiling of large images too? What do you think are the biggest pros/cons of the automatic tiling? E.g. is there a case where the client knows more about the best tiling than WR does?

---

Review status: 0 of 25 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[sample/src/main.rs, line 106 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXFeHzEiHXZji7ADrU:-KdXFeI-4M3VAMoRrS9F:bllu970) ([raw file](https://github.com/servo/webrender/blob/4221987984718bfc6312f92df9501d8fd7a88ea8/sample/src/main.rs#L106)):*
> ```Rust
> 
>     let vector_img = api.add_image(
>         ImageDescriptor {
> ```

What's the reasoning for changing the `ImageDescriptor` and implementing the builder?
I'd prefer us to go the same route as with `RenderOptions`, which is simpler.

---

*[webrender/src/device.rs, line 1638 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXGA-yckgCPKYgKKV2:-KdXGA-yckgCPKYgKKV3:b-wvbt9q) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/webrender/src/device.rs#L1638)):*
> ```Rust
>                 + width * bpp;
> 
>         assert!(data.len() as u32 >= len);
> ```

This seems to be a weaker test than we had. Perhaps, we can also make sure it's less or equal than `bpp * row_length * height`?

---

*[webrender/src/frame.rs, line 298 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXGaYfuHgkLi7AGkFD:-KdXGaYfuHgkLi7AGkFE:bsip83s) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/webrender/src/frame.rs#L298)):*
> ```Rust
> 
>             self.flatten_stacking_context(&mut traversal,
>                                           resource_cache,
> ```

Can we move it to `FlattenContext`?

---

*[webrender/src/frame.rs, line 676 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXH0JoMBfZ8BsSs2Z6:-KdXH0JoMBfZ8BsSs2Z7:bssnjuz) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/webrender/src/frame.rs#L676)):*
> ```Rust
>         // The image resource is tiled. We have to generate an image primitive
>         // for each tile.
>         // We need to do this because the image is borken up into smaller tiles in the texture
> ```

nit: "borken"

---

*[webrender/src/frame.rs, line 776 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXIO9Q4X5dg4iOKDio:-KdXIO9RF5UVIW9Nk6pA:btdis9w) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/webrender/src/frame.rs#L776)):*
> ```Rust
>         }
> 
>         if leftover.height != 0 {
> ```

let's combine this into the loop to avoid code repetition

---

*[webrender/src/frame.rs, line 848 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXJP0JL4tGYw345Tt6:-KdXJP0JL4tGYw345Tt7:b51xb0j) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/webrender/src/frame.rs#L848)):*
> ```Rust
> 
>         // Fix up the primitive's rect if it overflows the original item rect.
>         let x_overflow = prim_rect.max_x() - item.rect.max_x();
> ```

shouldn't this basically be the intersection of `prim_rect` and `item.rect`?

---

*[webrender/src/resource_cache.rs, line 95 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXKriKDaI-6UhPtOpz:-KdXKriLU56Xk-1geuyB:ba9w962) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/webrender/src/resource_cache.rs#L95)):*
> ```Rust
>     pub descriptor: ImageDescriptor,
>     pub external_id: Option<ExternalImageId>,
>     pub tiling: Option<u16>,
> ```

need some proper typing for this `u16`. I suppose `TileOffset` is the same thing, just the name doesn't match.

---

*[wrench/src/wrench.rs, line 424 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXMTDHMFJN_VFT67hD:-KdXMTDHMFJN_VFT67hE:b-k8pm8t) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/wrench/src/wrench.rs#L424)):*
> ```Rust
> }
> 
> fn generate_xy_gradient_image(w: u32, h: u32) -> (ImageDescriptor, ImageData) {
> ```

this sort of asks for a special cargo feature to be gated by

---

*[wrench/src/wrench.rs, line 445 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXN23BUhYcSX8hdmct:-KdXMpN5L2EMFLLrnndO:b-iepdo3) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/wrench/src/wrench.rs#L445)):*
> ```Rust
>     let buf_size = (w*h*4) as usize;
>     let mut pixels = Vec::with_capacity(buf_size);
>     // Unsafely filling the buffer is horrible. Unfortunately doing this idiomatically
> ```

this is certainly disturbing, something to address in follow-ups

---

*[wrench/src/wrench.rs, line 459 at r2](https://reviewable.io:443/reviews/servo/webrender/897#-KdXNNWD1bhAfyg33B2K:-KdXNNWElafOI4S69qfA:bn1uk22) ([raw file](https://github.com/servo/webrender/blob/46d99ef114612bb626a4d7b74871babcf0eec5b6/wrench/src/wrench.rs#L459)):*
> ```Rust
>     }
> 
>     return (
> ```

`return`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/897#-:-KdXDITBcqw2QF7WfYqH:b-rlvrv4)*

nical:
> I wish there was a clean choice between being explicit/implicit. If we go the explicit route, maybe we can declare a maximum image size and requires the client to do tiling of large images too? What do you think are the biggest pros/cons of the automatic tiling? E.g. is there a case where the client knows more about the best tiling than WR does?

Asking the tiling explicitly in the API makes it very easy to test the feature and that alone is more than enough for me to want it exposed. Automatically tiling large images is trivial so there isn't a compelling reason to not do it since we'd otherwise crash, especially as we want to be able to survive a compromised child process that sends valid but unreasonable data.

I don't know if we'll have situations where gecko/servo knows better than WR about when/how to tile.

> What's the reasoning for changing the ImageDescriptor and implementing the builder?

ImageDescriptor doesn't have a sensible default like RendererOption (only a subset of its parameters like stride and offset which are most of the time their default). You'll always want to pass at least the size and format, and sometimes the opaque flag so I felt like this would be less error prone than having ImageDescriptor::default() potentially set the format for you (to what?). Unless you mean having a default method that takes the size and format parameters like this?

```rust
let descriptor = ImageDescriptor {
    is_opaque: true,
    ..ImageDescriptor::default(64, 64, ImageFormat::A8)
};
```
I'm not convinced that this is simpler. Having to set the offset and stride each time although they are not useful most in most cases is too cumbersome so I think we should have some shorthand in whichever form.

> This seems to be a weaker test than we had. Perhaps, we can also make sure it's less or equal than bpp * row_length * height?

It's the strongest guarantee you have if you are uploading tiles, because the offset of the tile in the buffer has already been applied to the input slice. If you are allocating a buffer that fits the image exactly, at the time of uploading the last tile you only have ```(height - 1) * stride + width``` worth of data instead of ```height * stride``` since the beginning of the slice is not aligned with the stride.

One way to rearrange this could be to add the offset as a parameter to this function and not apply it to the input slice.

> need some proper typing for this u16. I suppose TileOffset is the same thing, just the name doesn't match.

This is an ```Option<TileSize>``` rather than tile offset (which is the x and y coordinate of the tile in number of tiles), but yeah I agree at least a type alias would make this easier to understand.

> (about generate_xy_gradient_image)
> this sort of asks for a special cargo feature to be gated by

What would it buy us? Adding compilation configurations is something i'd rather not do if I can avoid it, and wrench is mostly a tool for people to test out WebRender so having a shorthand to generate images with various parameters is handy to debugging and trying things out in addition to writing tests.

> (about the unsafe implementation generate_solid_color_image) 
> this is certainly disturbing, something to address in follow-ups

Yeah, I tried a whole bunch of things, I don't know what to do other than wait for rustc to generate better debug code, though. 24 seconds on my brand new xps 15 for the safe implementation to generate the image in the test case images/very-big.yaml. It's terrible.

> (about the code in frame.rs)
> let's combine this into the loop to avoid code repetition

Not sure what you mean there (but maybe I'll see clearer after a good night of sleep)?

kvark:
> Asking the tiling explicitly in the API makes it very easy to test the feature and that alone is more than enough for me to want it exposed. Automatically tiling large images is trivial so there isn't a compelling reason to not do it since we'd otherwise crash, especially as we want to be able to survive a compromised child process that sends valid but unreasonable data. I don't know if we'll have situations where gecko/servo knows better than WR about when/how to tile.

So, to recap, WR can tile efficiently automatically, no need to pass more info from Gecko, and it already does tiling for large images. Let's not have tiling exposed then! For the matter of testing, you can have a test-only API extension, but I don't see why we'd need in public.

> ImageDescriptor doesn't have a sensible default like RendererOption

It's not about the default, it can be `new` or any other constructor. If size and format are required, then let's just have this constructor, and let the rest of the fields be manageable by the user.

> Having to set the offset and stride each time although they are not useful most in most cases is too cumbersome

You seem to contradict yourself here. If they are not useful, then don't set it (each time). Let the constructor take care of them.

> It's the strongest guarantee you have if you are uploading tiles, because the offset of the tile in the buffer has already been applied to the input slice

Are you saying that `assert(data.len() <= bpp * row_length * height)` would be an incorrect check if we add it? I don't propose to remove your assert condition, I propose to add more.

> This is an Option<TileSize> rather than tile offset (which is the x and y coordinate of the tile in number of tiles), but yeah I agree at least a type alias would make this easier to understand.

Yes, I understand. Sorry, I think I jumped over too many steps :) First, it appears to me that the tile size should be configurable in both X and Y. When you do it, you'll end up with a type signature matching `TileOffset` exactly, so that's what I meant.

> What would it buy us? Adding compilation configurations is something i'd rather not do if I can avoid it, and wrench is mostly a tool for people to test out WebRender

I think it comes down to us formally defining `wrench` purpose and supposed usage. I see it as a general purpose tool that just *happens* to be convenient for testing. From that point of view, adding arbitrary constants or functions into it doesn't seem reasonable. @glennw might have a different POV here.

> let's combine this into the loop to avoid code repetition

You have a loop (actually, with a nested one), and then an extra iteration outside of it. I understand why you did this, but there seem to be quite a bit of repeated code in that extra iteration, so maybe we can move it into the loop for DRY?

webrender/res/prim_shared.glsl
```diff
@@ -485,7 +485,7 @@ VertexInfo write_vertex(RectWithSize instance_rect,
 
     // Apply offsets for the render task to get correct screen location.
     vec2 final_pos = device_p0_pos.zw -
-                     snap_delta -
+                     //snap_delta -
```
kvark: is this intentional?

nical: woops! not at all.

nical:
> You have a loop (actually, with a nested one), and then an extra iteration outside of it. I understand why you did this, but there seem to be quite a bit of repeated code in that extra iteration, so maybe we can move it into the loop for DRY?

My first iteration on this code was doing this, but the tiles on the right and bottom edges have a slightly different logic which made handling it along with the general case a bit confusing. Although the code looks structurally similar, I found that separating these special tiles from the general case simplified the code.

> It's not about the default, it can be new or any other constructor. If size and format are required, then let's just have this constructor, and let the rest of the fields be manageable by the user.

I guess the compromise we are looking for is a constructor with which incorporates the opaque flag rather than adding a builder-like method for this flag?

> So, to recap, WR can tile efficiently automatically, no need to pass more info from Gecko, and it already does tiling for large images. Let's not have tiling exposed then! For the matter of testing, you can have a test-only API extension, but I don't see why we'd need in public.

I don't know whether WR is well suited for deciding whether or not to tile and with which size, or if gecko/servo have are in a better spot. I meant that we'll probably figure this out along the way. If the heuristic is solely based on the image size (and not based on how the image will be used), letting WR choose should be fine. If the primitives using the image have an effect on the decision, it's a bit hard to figure it out in the render backend because as we generate the primitives we need to know whether the image is tiled.

An earlier version of this patch had a separate add_image_with_tiling function which sounds like the extension you are suggesting. It still has to be public because I want to be able to call it from the sample and from wrench. If you prefer this style I can add the function back and remove the tiling parameter in add_image.

Anyway, it is very easy to change the way the tiling is exposed in the RenderApi, so as long as it is exposed we can make it a second class citizen and let gecko and servo use an add_image methid that doesn't have this parameter.

> Are you saying that assert(data.len() <= bpp * row_length * height) would be an incorrect check if we add it? I don't propose to remove your assert condition, I propose to add more.

Yes, lemme summon my ascii-art-jutsu and consider the image below:
```
|-----------| stride
|---| width (= tile size)

+---+---+---+
|t1 |t2 |t3 |
+--- ---a---+    ---
|t4 |t5 |t6 |     |  height (= tile size)
+---+---+---b    ---
 ```

When uploading the last tile t6, the slice passed to this function goes from a (the first pixel of this tile) to b (the end of the buffer). Since the slide does not start at the beginning of a line (from the point of view of the image) but instead at the beginning of tile, the data you have left in the slice is at least as big as the stride times the number of rows below a + the tile width. In order for the assertion you suggest to be possible, we would need the slice to start at the beginning of the row from the point of view of the entire image rather than the first pixel of the tile.

kvark:
> I guess the compromise we are looking for is a constructor with which incorporates the opaque flag rather than adding a builder-like method for this flag?

Now that I think of it, these approaches are not conflicting, they are compatible, even if slightly redundant. So either the user wants to say `opaque: false`, or call `.with_opaque(false)` is not important - both API variants are clear and non-exclusive to each other. So I'm OK with the one you have so far.

> I don't know whether WR is well suited for deciding whether or not to tile and with which size, or if gecko/servo have are in a better spot. I meant that we'll probably figure this out along the way.

Sure thing, we can postpone this discussion up until a moment where we could see the big picture better.

> the data you have left in the slice is at least as big as the stride times the number of rows below a + the tile width

So you'll get `data.len() == (tile_height - 1) * stride + tile_width`, right? My assertion would gladly accept it, given that `stride >= tile_width`.

nical:
> So you'll get data.len() == (tile_height - 1) * stride + tile_width, right? My assertion would gladly accept it, given that stride >= tile_width.

Ah! I read your assertion the other way around. So the way it works, your assertion would also blow up here because the slice we pass to this function is the entire image buffer minus the part before the offset, so, for example the first tile gets a slice of the size of the entire image. I'm not sure that it is worth rejecting slices that are bigger than what is required for the upload. to do that we'd need the code in renderer.rs that calls this function to compute the minimum required size and take it off the slice before passing it to the function. I don't think the added logic would buy us much.

nical:
This PR tends to bitrot quickly and has been quite tedious to rebase. I won't be around next week to rebase it so I'd appreciate it if we could merge it before a week of merge conflicts have accumulated.

kvark:
@nical ok, let's do it. I think you addressed all my concerns. I'm going to run the Servo test suite and proceed.

kvark:
Tests appear to be fine, "FAIL [expected PASS] /_mozilla/css/iframe/hide_and_show.html" is intermittent as it used to be.
@bors-servo r+

nical:
Thanks a lot!

PR #902, opened 2017-02-20
Fix inset box shadows with blur radius.

### LINKED ISSUE #893: Inset box shadows don't work
E.g., nothing gets rendered.

glennw:
The geometry rects being sent for inset box shadows on the
blur path were incorrect.

The test coverage we have enabled in WPT/CSS only exercises the
inset fast path (where the blur radius is zero). Added a reftest
that just ensures inset box shadows at least draw something - this
is not ideal, but we don't have any proper infrastructure for
reftesting blurs right now. Also confirmed that the reftest fails
without this commit.

Fixes #893.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/902)


glennw:
r? @kvark 

cc @changm

kvark:
@glennw it appears to be asking for the wrench VS image comparison tests, while we only have wrench VS wrench.
@bors-servo r+

PR #904, opened 2017-02-20
Rename ScrollTree and Layer to ClipScrollTree and ClipScrollNode

mrobinson:
This should reduce the amount of collision between names and make it
bit easier to understand the code. ScrollLayerId doesn't change in
order to avoid an API change before it is necessary. We will be
modifying this when we add support for arbitrary clipping and scrolling
regions.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/904)


mrobinson:
@glennw r?

I'm not fixed on the name, so we can always rename this later. For the moment, this is just to ensure some consistency.

PR #905, opened 2017-02-20
Properly handle content transformed to empty rect

### LINKED ISSUE #894: panic in handle_primitive_run
http://people.mozilla.com/~jmuizelaar/tmp/scroll-frame-crash.zip panics at
https://github.com/servo/webrender/blob/master/webrender/src/frame_builder.rs#L1194

cc @mrobinson

mrobinson:
When content was transformed to an empty rectangle, we didn't always
make its ClipScrollGroup xf_rect member None. This caused an issue when
we tried to use the xf_rect member of the ScrollLayer for masking,
which was properly set to None. We fix this issue by making the
behavior consistent between the two structs.

Fixes #894.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/905)


mrobinson:
@glennw or @jrmuizel r?

mrobinson:
The build failure looks unrelated to my PR:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

webrender/src/frame_builder.rs
```diff
@@ -1066,16 +1066,20 @@ impl<'a> LayerRectCalculationAndCullingPass<'a> {
...
-                                                         self.device_pixel_ratio);
-
-                if layer_xf_rect.bounding_rect.intersects(&self.screen_rect) {
-                    packed_layer.screen_vertices = layer_xf_rect.vertices.clone();
-                    packed_layer.local_clip_rect = layer_local_rect;
-                    group.xf_rect = Some(layer_xf_rect);
-                }
+
+            let layer_local_rect = match layer_local_rect {
+                Some(layer_local_rect) if !layer_local_rect.is_empty() => layer_local_rect,
```
kvark: that seems a bit hacky, deserves at least a comment

mrobinson: Hrm. This was the old behavior as well. Would it be better to move it out of the helper method entirely. I can definitely add a comment in the meantime though.

mrobinson: Essentially if there is no intersection between the stacking context and the scroll layer, we let xf_rect be None. I can add a comment explaining this behavior in the next patch that I'm working on for this.

(Unless you meant the particular syntax here...)

kvark: but we aren't touching `xf_rect` here, can it be left over from a previous frame?

mrobinson: The diff makes it a little hard to see, but before all this it is initialized to None.

PR #906, opened 2017-02-20
Simplify ExternalImageId type.

JerryShih:
The ExternalImageId only contains a u64 field. Use u64 directly instead of a struct.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/906)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -266,7 +266,7 @@ impl ResourceCache {
                 // This image should not be an external image.
                 match image.data {
                     ImageData::ExternalHandle(id) => {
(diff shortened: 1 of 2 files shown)


glennw:
Changing it to a u64 removes a level of type safety from the interface. Is there a specific reason it makes sense to do this?

JerryShih:
If the ExternalImageId is a structure, we should have an additional structure type in ffi for the u64 passing.
I can catch your idea. If we use u64 directly, we can't make sure the incoming u64 is from ExternalImageId or other u64 value.

glennw:
Yup, let's make the FFI bindings work that way. Thanks!

PR #908, opened 2017-02-21
Rename ImageData::Vector into ImageData::Blob.

nical:
Followup form PR #858. 

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/908)


kvark:
Looks clean and simple, thanks!
@bors-servo r+

PR #912, opened 2017-02-21
separate generate_font_key from add_raw_font/add_native_font to allow…

lsalzman:
… for custom font key generation

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/912)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,8 +19,8 @@ dependencies = [
  "serde_json 0.9.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-dwrote 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.18.0",
- "webrender_traits 0.18.0",
+ "webrender 0.19.0",
+ "webrender_traits 0.19.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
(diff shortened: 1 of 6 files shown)


jrmuizel:
I implemented something similar to this but with a different approach. I kept the existing api but added a add_raw_font_with_key function that takes a key as a parameter instead of returning it. I don't feel strongly about which approach is better.

glennw:
This looks good to me, but there's a bit of CI fiddling that you'll need to do:

* Something else is in the process of landing that updates the version (the border-image PR). If you can modify the version of both wr and wr_traits to 0.19 that would be ideal.
* Then you'll need to run ```cargo update``` to update the cargo.lock file. This will allow wrench and the sample to build correctly with these changes.

If you want to give me access to push to your branch, I can make those changes. Or if you can make them and amend the commit, that'd be great.

jrmuizel:
It would be good to also do the same for image ids at the same time so we only need to break the api once.

kvark:
How is the "custom font key generation" supposed to work in the future?

kvark:
The PR seems a bit unsound to me. Now we have `generate_image_key`, `add_image`, and `update_image`, which has multiple issues:
  - implicit requirement that `add_image` is done once?
  - distinction between `add_image` and `update_image` is not obvious
  - `add_image` name becomes misleading, should probably be `set_image` or something

Thus, I'd like to see the reasoning behind this generate/add split.

nox:
This was also a breaking change, it would have been nice to accompany it with the corresponding servo/servo PR.

PR #913, opened 2017-02-21
Dummy sCache texture input

### LINKED ISSUE #907: Clipping issues on Android: textureSize(sCache, 0) is 0
Hello,

I'm working on WebVR Servo for Android and couldn't get some basic WebGL canvas to show on a simple html. It also happens with other non WebGL simple elements as images. After some debugging found that the textures from images and WebGL canvases are ok but the problem is related to the clipping implementation.

In the ps_image.fs.gl shader, the  `textureLod(sColor0, st, 0.0)` gives the correct color but it's multiplied by a 0 alpha. The alpha 0 is set by the call to do_clip() function.

```
 **alpha = min(alpha, do_clip());**

    // We calculate the particular tile this fragment belongs to, taking into
    // account the spacing in between tiles. We only paint if our fragment does
    // not fall into that spacing.
    vec2 position_in_tile = mod(relative_pos_in_rect, vStretchSize + vTileSpacing);
    // We clamp the texture coordinates to the half-pixel offset from the borders
    // in order to avoid sampling outside of the texture area.
    st = clamp(st, vStRect.xy, vStRect.zw);

oFragColor = vec4(1.0, 1.0, 1.0, alpha) * textureLod(sColor0, st, 0.0);
```
`write_clip()` function is correctly called in the vertex shader. The problem seems to be that `textureSize(sCache, 0).xy` returns a `vec(0, 0)` value, which leads to an invalid alpha value. In the same demo run on my desktop machine, `textureSize(sCache, 0).xy` returns a `vec(1, 1)` value, which leads to a correct alpha value.

- Where is sCache texture bound? I checked some OpenGL calls and it seemed that both Desktop & Android have a 0 texture_id bound to sCache (in the GL_TEXTURE4 unit). Should sCache be bound to an active texture in a simple demo or that only happens with multiple render passes in more complex demos?
- By the way I have also checked to call textureSize() of a new sampler2DArray uniform which is initialized to 0 and not used anywhere else. The Desktop OpenGL textureSize call still returns a vec(1,1) size vs Android textureSize call a vec(0,0) size.

kvark:
Fixes #907

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/913)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1736,6 +1736,12 @@ impl Renderer {
     }
 }
 
+impl Drop for Renderer {
+    fn drop(&mut self) {
+        self.device.deinit_texture(self.dummy_cache_texture_id);
+    }
+}
+
 pub enum ExternalImageSource<'a> {
     RawData(&'a [u8]),      // raw buffers.
     NativeTexture(u32),     // Is a gl::GLuint texture handle
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -463,6 +463,11 @@ pub struct Renderer {
     /// use a hashmap, and allows a flat vector for performance.
     cache_texture_id_map: Vec<TextureId>,
 
+    /// A special 1x1 dummy cache texture used for shaders that expect to work
+    /// with the cache but are actually running in the first pass
+    /// when no target is yet provided as a cache texture input.
+    dummy_cache_texture_id: TextureId,
+
     /// Optional trait object that allows the client
     /// application to provide external buffers for image data.
     external_image_handler: Option<Box<ExternalImageHandler>>,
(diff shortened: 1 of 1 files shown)


webrender/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "webrender"
-version = "0.17.0"
```
glennw: There's a few other PRs in the queue that also update the version, so we can probably leave the version number change out of this (this drops the version number below what the queued PRs do).

webrender/src/renderer.rs
```diff
@@ -715,6 +720,15 @@ impl Renderer {
                              TextureFilter::Linear,
                              ImageData::Raw(Arc::new(mask_pixels)));
 
+        let dummy_cache_texture_id = device.create_texture_ids(1, TextureTarget::Array)[0];
+        device.init_texture(dummy_cache_texture_id,
+                            1,
```
glennw: The comment above says the texture is 2x2 but the size here is 1x1.

kvark:
Thanks!
@bors-servo r=glennw

webrender/src/renderer.rs
```diff
@@ -715,6 +720,15 @@ impl Renderer {
                              TextureFilter::Linear,
                              ImageData::Raw(Arc::new(mask_pixels)));
 
+        let dummy_cache_texture_id = device.create_texture_ids(1, TextureTarget::Array)[0];
+        device.init_texture(dummy_cache_texture_id,
```
emilio: This is minor, but am I right assuming we're leaking this texture, or is there something preventing that? If not, would it make sense to destroy it on drop?

kvark:
@emilio good point!
@bors-servo r-

PR #919, opened 2017-02-22
Texture allocation refactoring

### LINKED ISSUE #914: Texture pages are not properly reused or deallocated
When running the gecko webgl mochitest suite, WR memory usage climbs into multiple gigabytes (see https://bugzilla.mozilla.org/show_bug.cgi?id=1341745). I traced this down to the fact that WR periodically allocates a new TexturePage but never goes back and reuses old texture pages or deallocates them if they are empty. I checked with @kvark and he agreed that this appears to be a bug. Quoting him from IRC:

> I don't see TexturePage objects being de-allocated. Also, they are not re-used properly. As in - 
> allocation only checks for the last page, while other pages may have more free space if they got 
> some content removed from them
> so I think 2 modifications need to be done: 1) removal of the last page in a vector if it becomes 
> empty, and 2) check for all the pages when allocating

kvark:
Fixes #914

Note: the [comment](https://github.com/servo/webrender/issues/914#issuecomment-281816803) appears to be irrelevant, since the actual texture freeing happens in `texture_cache`, and the free list just contains texture items, which are plain old data (PODs).

@staktrace I suggest we proceed with this PR and check if it resolves your issue, then potentially come with a follow-up.
@glennw r?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/919)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -9,9 +9,10 @@ use internal_types::{LowLevelFilterOp};
 use internal_types::{RendererFrame};
 use frame_builder::{FrameBuilder, FrameBuilderConfig};
 use clip_scroll_node::ClipScrollNode;
+use clip_scroll_tree::{ClipScrollTree, ScrollStates};
+use profiler::TextureCacheProfileCounters;
 use resource_cache::ResourceCache;
 use scene::{Scene, SceneProperties};
-use clip_scroll_tree::{ClipScrollTree, ScrollStates};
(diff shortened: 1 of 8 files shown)


glennw:
Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/919)*

glennw:
@bors-servo r+

Nice one, this is much cleaner!

PR #920, opened 2017-02-22
Remove unused DisplayListMode

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/920)


glennw:
@mrobinson Do you see any need to keep this around?

jrmuizel:
I've added another field removal

mrobinson:
@glennw Nope. I think that these can be safely removed. In fact, I think I had meant to do it myself. ☺

PR #922, opened 2017-02-24
Have ClipScrollTree use reference frame relative coordinates

mrobinson:
Instead of assuming that ClipScrollNodes and stacking contexts exist in
a contiguous tree, make the coordinate system of the ClipScrollTree
independent of the one used for stacking contexts. Now all coordinates
are calculated relative to the parent reference frame of every node.

Additionally, stacking contexts no longer have a transformation,
since they already rely on reference frames to be positioned. Stacking
contexts still need a rectangle and an origin though, because display
items may not be positioned relative to the top left corner due to the
way that overflow is calculated in Servo.

This change is necessary in order to clip and transform arbitrary
stacking context items with arbitrary scrolling nodes (within the same
reference frame).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/922)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -5,10 +5,9 @@
 use euclid::Point3D;
 use geometry::ray_intersects_rect;
 use spring::{DAMPING, STIFFNESS, Spring};
-use webrender_traits::{LayerPoint, LayerRect, LayerSize, LayerToScrollTransform};
+use webrender_traits::{LayerPixel, LayerPoint, LayerRect, LayerSize, LayerToScrollTransform};
 use webrender_traits::{LayerToWorldTransform, PipelineId, ScrollEventPhase, ScrollLayerId};
-use webrender_traits::{ScrollLayerRect, ScrollLocation, ScrollToWorldTransform, WorldPoint};
-use webrender_traits::{WorldPoint4D};
+use webrender_traits::{ScrollLayerRect, ScrollLocation, WorldPoint, WorldPoint4D};
 
 #[cfg(target_os = "macos")]
 const CAN_OVERSCROLL: bool = true;
@@ -16,6 +15,15 @@ const CAN_OVERSCROLL: bool = true;
 #[cfg(not(target_os = "macos"))]
 const CAN_OVERSCROLL: bool = false;
 
+#[derive(Clone)]
+pub enum NodeType {
+    /// Transform for this layer, relative to parent reference frame. A reference
+    /// frame establishes a new coordinate space in the tree.
+    ReferenceFrame(LayerToScrollTransform),
+
+    /// Other nodes just do clipping, but no transformation.
+    ClipRect,
+}
 
 /// Contains scrolling and transform information stacking contexts.
 #[derive(Clone)]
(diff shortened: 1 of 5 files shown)


mrobinson:
@glennw r?

Sorry, this patch is quite large, but I've tried to add thorough comments in the complicated parts.

glennw:
Reviewed 5 of 5 files at r1.

---

*[webrender/src/clip_scroll_node.rs, line 55 at r1](https://reviewable.io:443/reviews/servo/webrender/922#-Kdx19Hld80af9j_vnlB:-Kdx19Hld80af9j_vnlC:bqc2mve) ([raw file](https://github.com/servo/webrender/blob/26bc94144d5531d07771c76f109f9ce77153696f/webrender/src/clip_scroll_node.rs#L55)):*
> ```Rust
> 
>     /// Whether or not this node is a reference frame.
>     pub is_reference_frame: bool,
> ```

It seems like this should instead be a rust enum, with the local_transform a field of the enum for a reference frame?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/922)*

glennw:
Looks good! Just the one suggested change above.

mrobinson:

---

*[webrender/src/clip_scroll_node.rs, line 55 at r1](https://reviewable.io:443/reviews/servo/webrender/922#-Kdx19Hld80af9j_vnlB:-KdyecJSKWYQj6KyYqXm:blmzyfc) ([raw file](https://github.com/servo/webrender/blob/26bc94144d5531d07771c76f109f9ce77153696f/webrender/src/clip_scroll_node.rs#L55)):*

It seems like this should instead be a rust enum, with the local_transform a field of the enum for a reference frame?
</blockquote></details>

Yeah, I think this is a great idea. I'll update the branch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/922)*

staktrace:
This change causes gecko reftests to fail, if gecko is built with a webrender including the cset. See https://bugzilla.mozilla.org/show_bug.cgi?id=1343019#c1 onwards

@mrobinson is the change intended to change the semantics of the external API? I'm trying to figure out if this is a bug in the PR, or if we need to change some of our webrender_bindings glue code as the semantics of the API have changed. I noticed that the PR didn't touch the sample or wrench folders which usually happens when there's a change in the API semantics.

mrobinson:
@staktrace No, this shouldn't change the semantics of the API, so the failure is likely a bug.

mrobinson:
The issue here is that my change assumes that the coordinates provided by a scroll layers clip property are relative to the parent reference frame. Since Gecko is the only thing using this feature and it assumes they are relative to the scroll layer origin, I think we can go with this interpretation for the moment.

I posted a fix here: https://github.com/servo/webrender/pull/944

staktrace:
@mrobinson Thanks, I tested the patch and it fixes the failures I was seeing. 👍

webrender/src/clip_scroll_node.rs
```diff
@@ -117,9 +145,16 @@ impl ClipScrollNode {
...
-                            parent_world_transform: &ScrollToWorldTransform,
-                            parent_viewport_rect: &ScrollLayerRect) {
-        let inv_transform = match self.local_transform.inverse() {
+                            parent_reference_frame_transform: &LayerToWorldTransform,
+                            parent_combined_viewport_rect: &ScrollLayerRect,
+                            parent_accumulated_scroll_offset: LayerPoint) {
+
+        let local_transform = match self.node_type {
+            NodeType::ReferenceFrame(transform) => transform,
+            NodeType::ClipRect => LayerToScrollTransform::identity(),
```
kvark: shouldn't our offset from the parent reference frame be the transform?

webrender/src/clip_scroll_tree.rs
```diff
@@ -296,23 +299,41 @@ impl ClipScrollTree {
...
+                                          parent_viewport_rect,
+                                          parent_accumulated_scroll_offset);
+
+                    // The transformation we are passing is the transformation of the parent
+                    // reference frame and the offset is the accumulated offset of all the nodes
+                    // between us and the parent reference frame. If we are a reference frame,
+                    // we need to reset both these values.
+                    let (transform, offset) = match node.node_type {
+                        NodeType::ReferenceFrame(..) =>
+                            (node.world_viewport_transform, LayerPoint::zero()),
```
kvark: why is the offset here not `parent_accumulated_scroll_offset`?

webrender/src/tiling.rs
```diff
@@ -1393,6 +1399,31 @@ impl PackedLayer {
...
+        };
+
+        let xf_rect = TransformedRect::new(&local_rect, &self.transform, device_pixel_ratio);
+        if !xf_rect.bounding_rect.intersects(screen_rect) {
+            return None;
+        }
+
+        self.screen_vertices = xf_rect.vertices.clone();
+        self.local_clip_rect = local_rect;
+        Some(xf_rect)
```
kvark: shouldn't we return the intersection with `screen_rect`?

webrender/src/clip_scroll_tree.rs
mrobinson: The parent _accumulated_scroll_offset is rolled into the transformation of reference frames. Reference frames "reset" the offset because they establish a new coordinate system and origin in the tree.

webrender/src/tiling.rs
mrobinson: You might be right about this. I'm going to double-check tomorrow that a screen rect intersection wasn't lost in an earlier PR. It looks like the code immediately previous to this wasn't doing the intersection.

webrender/src/clip_scroll_node.rs
mrobinson: The offset and transform are combined into the self.world_viewport_transform below. Perhaps I'm not answering the right question though?

webrender/src/tiling.rs
mrobinson: ```screen_rect``` intersection is done in ```handle_pop_stacking_context```. Perhaps visibility should also be computed on a per-group basis as well. I believe the current design is holdover from when stacking contexts visibility was computed with a 1-to-1 stacking context and scroll layer pair.

PR #924, opened 2017-02-24
Rustify MaskSegment enum names

mrobinson:
The current names cause compilation warnings because they are not
sufficiently rusty.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/924)


glennw:
@bors-servo r+

Thanks! Been meaning to do this myself.

PR #925, opened 2017-02-24
Support resizing an image in the texture cache.

nical:
I was looking into something that I could do in an hour before taking a week off, and gave resizing images in update_image a shot.

It seems to work, except that I see some flickering when trying to continuously resize the image. I am not sure if it is because of logic itself or because of the hacky way I tested it. 
I accidentally committed my test code in the sample, I'll remove it if we want to land this.

There's no urgency to support this feature as far as I know so it's fine if we don't want to merge this. It's not complicated either so if this patch is correct, we might as well take it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/925)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -110,6 +110,13 @@ fn main() {
         None,
     );
 
+    let (desc, data) = generate_xy_gradient_image(100, 100);
+    let resized_img = api.generate_image_key();
+    api.add_image(
+        resized_img, desc, ImageData::new(data),
+        None,
+    );
+
     let pipeline_id = PipelineId(0, 0);
     let mut builder = webrender_traits::DisplayListBuilder::new(pipeline_id);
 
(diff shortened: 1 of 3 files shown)


kvark:
> I was looking into something that I could do in an hour before taking a week off, and gave resizing images in update_image a shot.

What issue is this PR addressing? Why do you need to resize in `update_image`?
I don't think it's a good idea. WR should treat image allocation permanently. The allocation is done in `add_image`, and the update shouldn't be allowed to change either the format or the size of that allocation.

nical:
It is addressing the TODO comment that was in the code. If we don't want to support this we should change the API.

kvark:
The comment is [from](https://github.com/servo/webrender/commit/52db2faad72433a220459433bfe972e7d9055b7f) Nov 2015, and it has a question mark at the end.
I say we treat resource allocation as immutable, similarly to how real (as in - non GL) 3D APIs treat it.
Since I recall @glennw being in agreement with this statement, let's just remove this comment line instead of actually providing the resize ability.

nical:
Sounds good to me. The only cases in gecko where we could use this that I can think of would already be changing the display list itself so we don't loose much from re-creating the image instead.
Let's also modify the API so that we don't pass parameters that only accept one value, then.

kvark:
Yep, agreed.

glennw:
Given the discussion above, I'm assuming we should close this. Please re-open if that's not the case!

Issue #921, opened 2017-02-23
Inset box shadows with a spread radius don't work


changm:
Nothing renders

kvark:
If only I new a person working on box shadows full time... oh wait!

PR #926, opened 2017-02-24
properly account for spread radius with inset box shadows

changm:
Inset box shadow spread radius needs to deflate, not inflate the rect. Fixes https://github.com/servo/webrender/issues/921

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/926)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -716,8 +716,13 @@ impl FrameBuilder {
         // The local space box shadow rect. It is the element rect
         // translated by the box shadow offset and inflated by the
         // box shadow spread.
+        let inflate_amount = match clip_mode {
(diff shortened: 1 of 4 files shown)


webrender/src/frame_builder.rs
```diff
@@ -716,8 +716,13 @@ impl FrameBuilder {
         // The local space box shadow rect. It is the element rect
         // translated by the box shadow offset and inflated by the
         // box shadow spread.
-        let bs_rect = box_bounds.translate(box_offset)
-                                .inflate(spread_radius, spread_radius);
+        let mut bs_rect = box_bounds.translate(box_offset);
```
glennw: This could probably be written something like:

```
let inflate_amount = match clip_mode { ... };
```

Which just negates the amount, and then keep the original code to calculate the bs_rect.

PR #927, opened 2017-02-24
Change set_root_display_list to accept built display lists

jrmuizel:
This will let us build our display lists in the child process in Gecko,
transport them over Gecko IPC and send them into webrender in the
compositor process.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/927)


kvark:
We discussed this with @jrmuizel and came up with a decent API.
@bors-servo delegate=jrmuizel

PR #928, opened 2017-02-25
Only gamma correct for non-mono windows fonts

changm:
We had some crashes as we'd expect a 3x1 cleartype texture, but it was really 1x1.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/928)

### DIFF
--- a/webrender/src/platform/windows/font.rs
+++ b/webrender/src/platform/windows/font.rs
@@ -230,13 +230,8 @@ impl FontContext {
                 let length = pixels.len() / 3;
                 let mut rgba_pixels: Vec<u8> = vec![0; length * 4];
                 for i in 0..length {
-                    // TODO(vlad): we likely need to do something smarter
-                    // This is what skia does
-                    let alpha = ((pixels[i*3+0] as u32 +
-                                pixels[i*3+1] as u32 +
-                                pixels[i*3+2] as u32)
-                                / 3) as u8;
-
+                    // Only take the G channel, as its closest to D2D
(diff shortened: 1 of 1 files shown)


webrender/src/platform/windows/font.rs
```diff
@@ -278,22 +278,24 @@ impl FontContext {
...
-            },
-            None => &self.gamma_lut
-        };
-
-        lut_correction.preblend_rgb(&mut pixels, width, height,
-                                    ColorLut::new(key.color.r,
-                                                  key.color.g,
-                                                  key.color.b,
-                                                  key.color.a));
+        if render_mode != FontRenderMode::Mono {
```
glennw: This will run the gamma correct on alpha and subpixel modes (just not AA disabled). Is that intentional or is it meant to only be on subpixel mode?

webrender/src/platform/windows/font.rs
```diff
@@ -278,22 +278,24 @@ impl FontContext {
...
+                    } else {
+                        &self.gamma_lut
+                    }
+                },
+                None => &self.gamma_lut
+            };
+
+            lut_correction.preblend_rgb(&mut pixels, width, height,
+                                        ColorLut::new(key.color.r,
+                                                    key.color.g,
```
glennw: nit: indentation

webrender/src/platform/windows/font.rs
changm: It is intentional. We still gamma correct and also fixed to only take the G channel, which is what we do in gecko.

PR #931, opened 2017-02-27
Implement support for page zoom.

glennw:
This allows setting a page (desktop style) zoom factor. It is multiplied
by the hidpi-factor to give a css pixel -> device pixel scale factor.
This value is then used to select rasterization size for items such
as glyphs, box shadow caches etc to provide high resolution zoom support.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/931)


glennw:
r? @kvark 

I've tested this in Servo and it appears to work well - the zoom in/out matches FF.

Next step after this is to support pinch-zoom, pan, and overscroll, which should be fairly simple extensions to this.

PR #934, opened 2017-02-27
[RFC] Add support for pushing already built display lists.

jrmuizel:
Gecko has the concept of 'empty transactions' where we need to get a new
frame composited but aren't rebuilding the Gecko display lists. To
support this we'd like to keep around a retained representation of the
partial WebRender display list.

One easy way to implement this is to just use a BuiltDisplayList
as the retained representation and support appending them
to display lists. Unfortunately, the implementation is a bit fragile
because of needing to deal with ItemRanges.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/934)

### DIFF
--- a/webrender_traits/src/display_list.rs
+++ b/webrender_traits/src/display_list.rs
@@ -355,6 +355,40 @@ impl DisplayListBuilder {
         self.list.push(item);
     }
 
+    // Don't use this function. It will go away.
+    // We're using it as a hack in Gecko to retain parts sub-parts of display lists so that
+    // we can regenerate them without building Gecko display items. 
+    pub fn push_built_display_list(&mut self, dl: BuiltDisplayList, aux: AuxiliaryLists) {
+        use SpecificDisplayItem::*;
+        // It's important for us to make sure that all of ItemRange structures are relocated
+        // when copying from one list to another. To avoid this problem we could use a custom
+        // derive implementation that would let ItemRanges relocate themselves.
+        for i in dl.all_display_items() {
+            let mut i = *i;
+            match i.item {
+                Text(ref mut item) => {
+                    item.glyphs = self.auxiliary_lists_builder.add_glyph_instances(aux.glyph_instances(&item.glyphs));
+                }
+                Gradient(ref mut item) => {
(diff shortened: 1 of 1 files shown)


glennw:
@jrmuizel If you call api.generate_frame(None) it will redraw the current display list. Does this achieve what you want or is there a difference between that and what gecko requires above?

jrmuizel:
My understanding is that the display list that's sent to webrender is not identical, but changed in a way that doesn't require us to rebuild the display list.

It may be that this empty transaction functionality is not needed in the long term, but I'll need to investigate that further.

jrmuizel:
I have some more details on the scenarios in which this happens. Imagine you have a page that's using js to continuously set the transform or opacity on some piece of content. In Gecko we'll avoid building the Gecko display list and just send an updated layer tree to the compositor. In this case we need some kind of retained version of the display list that we keep around on the gecko side so that we can send it over as needed.

I think we'll be able to come a better solution to this in the long term but for now this will allow Gecko to avoid having to have it's own duplicate way of retaining parts of a display list.

glennw:
For that use case, the preferred API is here - https://github.com/servo/webrender/pull/832.

The way it works, roughly, is:

* During DL construction, you set properties you want to animate to be a "binding" (opacity and transform are supported now, but others are easy to add).
* Then, when you call api.generate_frame() you can pass an optional list of property values (these are the current value of the animated properties).

In this way, there is no need to maintain a client-side copy of the display list. You can just re-draw the same display list by calling generate_frame(), but specifying different values for the animated properties.

Does this sound like it would cover all the use cases you know of?

glennw:
I guess the problem is that you don't know which properties JS might change in this case, so you end up setting every stacking context as a binding? Which isn't necessarily a problem - I don't think that will have any particularly bad side effects.

jrmuizel:
Yes, #832 is the longer term api that I had in mind. And yes, we'd need to set every stacking context as a binding.

How would you feel about adding api like this temporarily until we can transition to something saner?

glennw:
I guess that would be OK - so long as we clearly mark that it's a temporary API and will be removed.

jrmuizel:
I have this being used in https://bugzilla.mozilla.org/show_bug.cgi?id=1344396 so it would be good to land in webrender.

r? @glennw

Issue #725, opened 2017-01-16
Sample not displaying correctly on Hi DPI display


randomPoison:
When I run wr-sample on my MacBook Pro this is how it looks:

<img width="1061" alt="screen shot 2017-01-16 at 3 50 35 pm" src="https://cloud.githubusercontent.com/assets/1900829/22000253/c13f6c16-dc03-11e6-8959-8bf035ac6cd2.png">

Everything but the background seems to be limited to a quarter of the screen, and the hidpi factor for my display is 2, so it looks like webrender isn't taking the hidpi scaling factor into account.

## System Info

- MackBook Pro (Retina, 15-inch, Mid 2015)
- OS X EL Capitan (Version 10.11.6)
- OpenGL version 4.1 INTEL-10.14.73, Intel Iris Pro OpenGL Engine
- hidpi factor: 2 (native 2)
- webrender version 0.11.1

jrmuizel:
I see this as well. I think this might be fixed by calling glViewport with actual size in pixels of the window.

waywardmonkeys:
@glennw thought this might be fixed by his change in #935, but I'm still seeing this problem today after checking out `master` and running on a Retina MBP.

glennw:
@waywardmonkeys Hmm, I thought that would fix it too. Unfortunately I don't have a retina MBP here to test with or debug.

glennw:
This is resolved by #1081.

PR #935, opened 2017-02-28
Pass the hidpi factor through to WR init in sample.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/935)


glennw:
I *think* this will fix #725 - but I don't have a retina mac to check on.

PR #939, opened 2017-02-28
Let primitive runs specify their own ClipScrollNode

mrobinson:
This will allow us to modify the API to expose this ability.
Additionally, we now only create ClipScrollGroups when a primitive
is actually used for a particular stacking context / ClipScrollNode
combination. This allows us to remove a few temporary hacks as well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/939)


glennw:
@mrobinson This will need a rebase. I think the main conflict will be that the mix-blend-mode PR previous to this also introduces a stacking context stack in the frame builder.

PR #940, opened 2017-02-28
Modify run_tests.py for running any wrench command headless

rlhunt:
It would be nice to specify the command to run under OSMesa and not have to type `cargo build --release --features headless` so often.

@glennw does this seem like an ok change?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/940)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,5 +30,4 @@ script:
   - if [ $BUILD_KIND = DEBUG ]; then (cd sample && cargo test --verbose); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd sample && cargo build --verbose --features=profiler); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd wrench && cargo test --verbose); fi
-  - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && cargo build --release --verbose --features headless); fi
-  - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && python run_tests.py res/rect.yaml); fi
+  - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && python headless.py reftest); fi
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -31,4 +31,5 @@ script:
   - if [ $BUILD_KIND = DEBUG ]; then (cd sample && cargo build --verbose --features=profiler); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd wrench && cargo test --verbose); fi
   - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && cargo build --release --verbose --features headless); fi
-  - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && python run_tests.py res/rect.yaml); fi
+  - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && python headless.py -t 1 show res/rect.yaml); fi
```
glennw: I don't think this line needs to be here at all - I had that there when I first wrote the CI + wrench + headless code, to test that it was working on travis.

rlhunt:
Hm. I'm not sure what's going on with the test failure.

I've been running into an issue locally where ./run_tests.py only works after a clean rebuild. If I modify a file and build (--release --features headless), ./run_tests.py selects the wrong os_mesa folder. I'm not sure if that could be happening here.

glennw:
I've occasionally seen that issue too - it seems that sometimes cargo ends up creating two osmesa-src dirs, one which has the build artifacts and one without.

The python code that searches for the lib fails in this case as it just looks for the newest dir.

We can work around the issue by modifying our python code to only select the dir that contains the libs - I just haven't gotten around to doing this yet.

rlhunt:
Ah that makes sense. I'll see if I can incorporate that into this PR.

rlhunt:
The folder without libs doesn't seem to have an 'out' dir. I updated the PR with a hack to filter those out.

glennw:
@bors-servo r+

I'll r+ now, let's see if that makes travis happy.

PR #941, opened 2017-02-28
Remove serde_codegen support

nox:
We require Rust 1.15 and use serde_derive now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/941)


jrmuizel:
\o/

jrmuizel:
Sorry about closing this.

webrender_traits/src/lib.rs
```diff
@@ -27,10 +25,6 @@ extern crate core_graphics;
 #[cfg(target_os = "windows")]
 extern crate dwrote;
 
-#[cfg(feature = "serde_codegen")]
-include!(concat!(env!("OUT_DIR"), "/types.rs"));
-
-#[cfg(feature = "serde_derive")]
 include!("types.rs");
```
emilio: Could we just inline this file here or use `mod`? This will allow better error messages.

nox: @emilio I would rather have a WR contributor do that after this PR lands, because each type in `types.rs` should actually live in one of the existing modules.

PR #942, opened 2017-02-28
Add `png' command to save the result of a yaml file

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/942)

### DIFF
--- a/wrench/src/png.rs
+++ b/wrench/src/png.rs
@@ -1,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use WindowWrapper;
+use gleam::gl;
+use image::ColorType;
+use image::png::PNGEncoder;
+use std::fs::File;
+use std::sync::mpsc::{channel, Sender};
+use webrender_traits::*;
+use wrench::{Wrench, WrenchThing};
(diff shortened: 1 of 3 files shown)


wrench/src/png.rs
```diff
@@ -0,0 +1,62 @@
...
+    let mut data = gl::read_pixels(0,
+                                   0,
+                                   size.0 as gl::GLsizei,
+                                   size.1 as gl::GLsizei,
+                                   gl::RGBA,
+                                   gl::UNSIGNED_BYTE);
+    let width = size.0;
+    let height = size.1;
+
+    // flip image vertically (texture is upside down)
```
glennw: We should probably put this into a function (since it's used elsewhere).

wrench/src/png.rs
```diff
@@ -0,0 +1,62 @@
...
+    let orig_pixels = data.clone();
+    let stride = width as usize * 4;
+    for y in 0..height as usize {
+        let dst_start = y * stride;
+        let src_start = (height as usize - y - 1) * stride;
+        let src_slice = &orig_pixels[src_start .. src_start + stride];
+        (&mut data[dst_start .. dst_start + stride]).clone_from_slice(&src_slice[..stride]);
+    }
+
+    let encoder = PNGEncoder::new(File::create("out.png").unwrap());
```
glennw: We should probably allow specifying a filename on the command line.

glennw:
Ah, the servo-tidy script failed - you can run the script yourself or see the error in the travis logs.

PR #945, opened 2017-03-01
Make perspective and transform Option types

mrobinson:
CSS makes a distinction between an identity transform and an empty
transform. For instance, stacking contexts with identity
transformations still create what we call reference frames, which means
they serve as coordinates spaces for fixed position elements. This will
also help us to properly assign ScrollLayerIds during display list
construction if we want.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/945)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -372,9 +372,6 @@ impl Frame {
             return;
         }
 
-        let stacking_context_transform =
-            context.scene.properties.resolve_layout_transform(&stacking_context.transform);
-
         let mut reference_frame_id = current_reference_frame_id;
         let mut scroll_layer_id = match stacking_context.scroll_policy {
             ScrollPolicy::Fixed => current_reference_frame_id,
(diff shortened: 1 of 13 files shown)


webrender/src/frame.rs
```diff
@@ -383,17 +380,20 @@ impl Frame {
 
         // If we have a transformation, we establish a new reference frame. This means
         // that fixed position stacking contexts are positioned relative to us.
-        if stacking_context_transform != LayoutTransform::identity() ||
-           stacking_context.perspective != LayoutTransform::identity() {
+        if stacking_context.transform.is_some() || stacking_context.perspective.is_some() {
+            let transform =
+                context.scene.properties.resolve_layout_transform(&stacking_context.transform);
```
emilio: nit: I think usually `Option<&T>` is preferred, given it saves an indirection and gets optimized as a null pointer for the `None` case (you'd need to do `stacking_context.transform.as_ref()` here). That being said, I don't think it matters a lot here.

mrobinson: @emilio Thanks for the tip. I think your suggestion makes a lot of sense.

glennw:
@bors-servo r+

Nice! That was a much simpler change than I expected :)

PR #946, opened 2017-03-01
Use correct yaml keys for image stretch_size and tile_spacing

jamienicol:
We weren't being consistent between writing and reading.

Keep the read keys the same so as to not break existing yaml files, plus I think they're the better keys.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/946)

### DIFF
--- a/wrench/src/yaml_frame_writer.rs
+++ b/wrench/src/yaml_frame_writer.rs
@@ -549,8 +549,8 @@ impl YamlFrameWriter {
                     if let Some(&CachedImage { tiling: Some(tile_size), .. }) = self.images.get(&item.image_key) {
                         u32_node(&mut v, "tile-size", tile_size as u32);
                     }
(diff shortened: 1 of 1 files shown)


wrench/src/yaml_frame_writer.rs
```diff
@@ -549,8 +549,8 @@ impl YamlFrameWriter {
                     if let Some(&CachedImage { tiling: Some(tile_size), .. }) = self.images.get(&item.image_key) {
                         u32_node(&mut v, "tile-size", tile_size as u32);
                     }
-                    size_node(&mut v, "strech", &item.stretch_size);
-                    size_node(&mut v, "spacing", &item.tile_spacing);
+                    size_node(&mut v, "strech_size", &item.stretch_size);
```
glennw: Spelling mistake: this should be stretch_size

jamienicol:
Whoops. Fixed now!

PR #947, opened 2017-03-01
add delete_font API for deleting fonts

lsalzman:
This allows us to better synchronize font lifetimes with gecko, so that when we delete fonts in gecko we can ensure WebRender is no longer keeping them around.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/947)


glennw:
This generally looks good to me - compile error on mac CI though.

Issue #189, opened 2016-02-19
box-shadow + border-radius doesn't work


glennw:
It currently sets a clip out rect around the visual box. Could we instead modify the box-shadow raster op shader to detect if inside the visual box and output 0?

cc @pcwalton

jaredpetker:
I don't quite grok what you mean @glennw, I was looking for a ticket with regards to the fact that corner radii clipping doesn't seem to be supported for shadows, and only clips out a square, is this what you are referring to here as well?

changm:
Here's a test case:

`<html>
<head>
<style>
div {
  box-shadow: 0px 0px 5px 0px #000000;
  width: 50px;
  height: 50px;
  border-radius: 20px;
}
</style>
</head>
<body>
<div></div>
</body>
</html>`

What it should look like:
<img width="93" alt="correct" src="https://cloud.githubusercontent.com/assets/1015726/23190232/8f5c8346-f84b-11e6-8c91-5f29c2bd0fb6.png">

What it actually looks like:
<img width="104" alt="incorrect" src="https://cloud.githubusercontent.com/assets/1015726/23190245/a323b106-f84b-11e6-9235-f2168e344fe6.png">

PR #950, opened 2017-03-02
Box shadow with border radius properly clips

changm:
Fixes issue https://github.com/servo/webrender/issues/189

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/950)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -22,6 +22,7 @@ use webrender_traits::{BlobImageData, BlobImageDescriptor, BlobImageError, BlobI
 use webrender_traits::{BlobImageResult, ClipRegion, ColorF, DeviceUintSize, Epoch, GlyphInstance};
 use webrender_traits::{ImageData, ImageDescriptor, ImageFormat, ImageKey, ImageRendering};
 use webrender_traits::{LayoutPoint, LayoutRect, LayoutSize, PipelineId, RasterizedBlobImage};
+use webrender_traits::{BoxShadowClipMode};
 
 
 fn load_file(name: &str) -> Vec<u8> {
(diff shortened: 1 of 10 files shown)


changm:
I'm not sure how you feel about having the display list builder auto generate the clip versus having users explicitly provide a clip for the border-radius. I can move the clip building generation into the display list builder if you'd prefer.

glennw:
I'm a bit unsure about merging this.

It fixes the common case for box shadows, where there is only a single (complex) clip, which is an improvement on what we have now.

On the other hand, inverting the clip mask is a hack that won't work correctly if the box shadow happens to intersect with any other complex clips. Making this work correctly will require proper support for clip in / clip out, or achieving it in a different way.

What do you think @kvark ?

glennw:
@kvark @changm I wonder if we could add support to the clip mask generation to handle clip in / out fairly simply. I haven't thought about the details much yet, but we have multiple color channels available in the render task for clip masks. Perhaps we could accumulate clip in separately from clip out in two channels, and then the do_clip() shader code can sample both the clip in/out channels and combine them?

kvark:
> On the other hand, inverting the clip mask is a hack that won't work correctly if the box shadow happens to intersect with any other complex clips. Making this work correctly will require proper support for clip in / clip out, or achieving it in a different way.

The clip mask output by `do_clip` is final, it hides all combining and inversion that happens during the mask generation. So I consider it safe to invert it for the box shadow case, given that it needs to clip "out" as opposed to clip "in".

> but we have multiple color channels available in the render task for clip masks.

I'd rather not use the other channels. Instead, I'd want the clip stuff to go into a single-channel render target, which would reduce the bandwidth cost and improve the texture cache performance upon sampling from it.

My only small-ish concern about the PR that it introduces another shader. If the box shadows are not omnipresent on the pages, we could just enable the clip logic for all of them.

kvark:
I regress, @glennw concern is valid. We should add a mode to clip mask generation that would invert it. We should not invert it in the client (ps_box_shadow) shader.

glennw:
Closing this (for now). We'll add support to clip mask generation for in/out, and then the box shadow change should be straightforward.

PR #951, opened 2017-03-02
Support overscroll for browser.html prototyping.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/951)

### DIFF
--- a/sample/src/main.rs
+++ b/sample/src/main.rs
@@ -93,7 +93,8 @@ fn main() {
         .. Default::default()
     };
 
-    let (mut renderer, sender) = webrender::renderer::Renderer::new(opts).unwrap();
+    let size = DeviceUintSize::new(width, height);
+    let (mut renderer, sender) = webrender::renderer::Renderer::new(opts, size).unwrap();
     let api = sender.create_api();
 
     let notifier = Box::new(Notifier::new(window.create_window_proxy()));
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -37,6 +37,10 @@ pub struct ClipScrollNode {
     /// Viewing rectangle in the coordinate system of the parent reference frame.
     pub local_viewport_rect: LayerRect,
 
+    /// Clip rect of this node - typically the same as viewport rect, except
+    /// in overscroll cases.
+    pub local_clip_rect: LayerRect,
+
     /// Viewport rectangle clipped against parent layer(s) viewport rectangles.
     /// This is in the coordinate system of the parent reference frame.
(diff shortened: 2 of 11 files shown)


glennw:
@paulrouget 

OK, here is the proof of concept of the overscroll support.

I've done some quick testing and it seems to work (and it also
makes the css viewport tests in our test suite pass!).

Right now, it's not exposed as an API - you can tweak the setting
in send_window_size() in compositor.rs where it sets the inner rect.
If you want to prototype, feel free to hack an API to that so you
can set it from your embedding API.

The corresponding Servo WIP patch is here - https://github.com/glennw/servo/tree/overscroll-poc.
You'll need both this and this WR branch.

The main thing right now is checking that it works as expected. If
so, I'll get the PR cleaned up and reviewed.

@mrobinson - I ended up not being able to remove the rect from the
reference frame (yet). It ended up getting complicated by the mask
coordinates fix - we can discuss on IRC the best way to
resolve that. It would be great if you have any feedback or
comments on this general approach though :)

paulrouget:
This appears to work well, except for position:fixed elements not being fixed anymore.

paulrouget:
Code I'm using:

```html
<style>
  body {
    margin: 0;
    border: 10px solid red;
    height: 2000px;
  }
  div {
    position: fixed;
    top: 0;
    width: 100px;
    height: 100px;
    background: blue;
    top: 10px;
    left: 10px;
  }
</style>
<div></div>
```

paulrouget:
It's looking pretty good :)

In this demo, the actual size of the frame is always the same. The DOM window is resized, while the extra margins (visible under the toolbar and tabbar) are still being painted:

legitimatelightheartedbudgie

glennw:
@paulrouget Hmm, the fixed position element scrolling appears to be unrelated to this PR (it occurs on current WR master builds). It's probably a regression in one of the recent PRs that hasn't made it into Servo yet (thankfully!).

I'll bisect WR today and get that sorted out separately.

glennw:
@paulrouget Ah, I know the cause. There's an incompatibility between WR master and current Servo. I'll do a Servo update that fixes up this incompatibility, and that will fix the position: fixed issue.

glennw:
@paulrouget OK, I landed a WR update in Servo that resolves the incompatibility with fixed elements. I've rebased this and the Servo branch on that and confirmed that your test case is working with the overscroll support.

So if you update those branches, you should be able to continue your prototyping.

r? @mrobinson

mrobinson:
Looks like the sample app needs to be updated, but other than that, these changes look good.

glennw:
@mrobinson Thanks! Updated the sample and wrench code.

Issue #936, opened 2017-02-28
Support external image for yuv format


JerryShih:
@glennw @kvark @nical

a)
Right now, the yuv format image only get the data from resource_cache. We should update it to support external image. 
What we do for external image:
https://github.com/servo/webrender/blob/5c1774462ee57b209ab81295ab944a176f2913f5/webrender/src/prim_store.rs#L908
Current yuv implementation:
https://github.com/servo/webrender/blob/5c1774462ee57b209ab81295ab944a176f2913f5/webrender/src/prim_store.rs#L955

b)
The current yuv image usage are:
```
1. add 3 image in api
y_key = api.add_image(...., y_buffer)
u_key = api.add_image(...., u_buffer)
v_key = api.add_image(...., v_buffer)
2. add yuv_image in frame_builder
builder.add_yuv_image(...., y_key, u_key, v_key, ...)
```
If we try to use external yuv_image, we need to use 3 external id with current api interface. Is it too complicated? 
How about add more interface like:
```
1. add new interface yuv_image() in api
yuv_key = api.add_yuv_image(...., y_buffer, ...., u_buffer, ...., v_buffer)
2. update yuv_image() interface in frame_builder
builder.add_yuv_image(...., yuv_key, ...)
```
Then, external_image_handler could to return 3 external image at the same time.

PR #952, opened 2017-03-02
Support external image for yuv format.

JerryShih:
#936 
@nical @kvark @glennw 
This is the straightforward implementation for external yuv format image. When WR has the external images for some specific yuv channels, it will put that images into a deferred_resolves list. Then, resolve the image content and uv coordinate at renderer.
The yuv_image() interface is not changed here.

But I would like to have a single external image id for a yuv image:
```
1. add new interface yuv_image() in api
yuv_key = api.add_yuv_image(...., y_buffer, ...., u_buffer, ...., v_buffer)
2. update yuv_image() interface in frame_builder
builder.add_yuv_image(...., yuv_key, ...)
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/952)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -9,7 +9,7 @@ use internal_types::{SourceTexture, PackedTexel};
 use mask_cache::{ClipSource, MaskCacheInfo};
 use renderer::{VertexDataStore, GradientDataStore};
 use render_task::{RenderTask, RenderTaskLocation};
-use resource_cache::{ImageProperties, ResourceCache};
+use resource_cache::{CacheItem, ImageProperties, ResourceCache};
 use std::mem;
 use std::usize;
 use util::TransformedRect;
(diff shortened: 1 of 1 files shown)


glennw:
This approach works, but I don't think it's ideal to be patching the general gpu_data64 array from the renderer. This could complicate things in the future once we're doing incremental updates.

As an alternative approach, could we port the external image code to store the UVs for the images in the resource_rects structure (like normal images do)? This could even be done as an initial step and landed independently of this work.

I *think* that will simplify this patch quite a lot - by making the deferred resolve in the renderer consistent (so it only ever needs to operate on the resource_rects patching).

Does that sound reasonable?

JerryShih:
Thanks, I will check how to use resource_rects for yuv_image first.

JerryShih:
@glennw @kvark 
This update is base on #970(use gpu_resource_rects for yuv image).

JerryShih:
rebase for #970

glennw:
@bors-servo r+

@JerryShih Nice one, this worked out really well!

PR #954, opened 2017-03-03
Fix typo transforming clip masks incorrectly

mrobinson:
I recently made a typo when fixing a bug, meaning that clip masks where
transformed as if they were content of their scroll nodes. Instead we
should transform using the viewport transform. Also add test coverage
for this bug.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/954)


mrobinson:
@glennw r?

PR #955, opened 2017-03-03
Remove duplicate contexts on stacking context stack

mrobinson:
During display list flattening stacking contexts were being push onto
the stacking context stack twice, due to a merge issue. It seems that
two PRs added this functionality, so it was duplicated. This PR removes
the duplicate entries (which were mainly harmless) and cleans up the
code slightly.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/955)


mrobinson:
@glennw r?

mrobinson:
Looks like the Mac failure is unrelated to this PR.

PR #959, opened 2017-03-05
Assign ScrollLayerId per primitive

mrobinson:
Allow assigning a ScrollLayerId directly to a primitive. In the future,
this will allow API users to assign primitive to arbitrary scroll
layers. As part of this work we collapse the scroll_layer_store in
FrameBuilder into the ClipScrollTree itself, making the architecture
a bit less complicated. The biggest potential change here is that
changing between clipping layers requires HashMap lookups instead of
array indexing. If this proves to be a performance issue, it is
possible to optimize this code at the cost of simplicity.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/959)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -4,10 +4,16 @@
 
 use euclid::Point3D;
 use geometry::ray_intersects_rect;
+use mask_cache::{ClipSource, MaskCacheInfo};
+use prim_store::GpuBlock32;
+use renderer::VertexDataStore;
 use spring::{DAMPING, STIFFNESS, Spring};
-use webrender_traits::{LayerPixel, LayerPoint, LayerRect, LayerSize, LayerToScrollTransform};
-use webrender_traits::{LayerToWorldTransform, PipelineId, ScrollEventPhase, ScrollLayerId};
-use webrender_traits::{ScrollLayerRect, ScrollLocation, WorldPoint, WorldPoint4D};
+use tiling::PackedLayerIndex;
+use util::TransformedRect;
+use webrender_traits::{ClipRegion, LayerPixel, LayerPoint, LayerRect, LayerSize};
+use webrender_traits::{LayerToScrollTransform, LayerToWorldTransform, PipelineId};
+use webrender_traits::{ScrollEventPhase, ScrollLayerId, ScrollLayerRect, ScrollLocation};
+use webrender_traits::{WorldPoint, WorldPoint4D};
 
 #[cfg(target_os = "macos")]
 const CAN_OVERSCROLL: bool = true;
@@ -15,18 +21,61 @@ const CAN_OVERSCROLL: bool = true;
 #[cfg(not(target_os = "macos"))]
 const CAN_OVERSCROLL: bool = false;
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
+pub struct ClipInfo {
+    /// The ClipSource for this node, which is used to generate mask_cache_info.
+    pub clip_source: ClipSource,
+
+    /// The MaskCacheInfo for this node, which is produced by processing the
+    /// provided ClipSource.
+    pub mask_cache_info: Option<MaskCacheInfo>,
+
+    /// The packed layer index for this node, which is used to render a clip mask
+    /// for it, if necessary.
+    pub packed_layer_index: PackedLayerIndex,
+
+    /// The final transformed rectangle of this clipping region for this node,
+    /// which depends on the screen rectangle and the transformation of all of
+    /// the parents.
+    pub xf_rect: Option<TransformedRect>,
+}
+
+impl ClipInfo {
+    pub fn new(clip_region: &ClipRegion,
+               clip_store: &mut VertexDataStore<GpuBlock32>,
+               packed_layer_index: PackedLayerIndex)
+               -> ClipInfo {
+        // We pass true here for the MaskCacheInfo because this type of
+        // mask needs an extra clip for the clip rectangle.
+        let clip_source = ClipSource::Region(clip_region.clone());
+        ClipInfo {
+            mask_cache_info: MaskCacheInfo::new(&clip_source, true, clip_store),
+            clip_source: clip_source,
+            packed_layer_index: packed_layer_index,
+            xf_rect: None,
+        }
+    }
+
+    pub fn is_masking(&self) -> bool {
+        match self.mask_cache_info {
+            Some(ref info) => info.is_masking(),
+            _ => false,
+        }
+    }
+
+}
+#[derive(Clone, Debug)]
 pub enum NodeType {
     /// Transform for this layer, relative to parent reference frame. A reference
     /// frame establishes a new coordinate space in the tree.
     ReferenceFrame(LayerToScrollTransform),
 
     /// Other nodes just do clipping, but no transformation.
-    ClipRect,
+    Clip(ClipInfo),
 }
 
 /// Contains scrolling and transform information stacking contexts.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct ClipScrollNode {
     /// Manages scrolling offset, overscroll state etc.
     pub scrolling: ScrollingState,
@@ -56,6 +105,9 @@ pub struct ClipScrollNode {
     /// Pipeline that this layer belongs to
     pub pipeline_id: PipelineId,
 
+    /// Parent layer. If this is None, we are the root node.
+    pub parent: Option<ScrollLayerId>,
+
     /// Child layers
     pub children: Vec<ScrollLayerId>,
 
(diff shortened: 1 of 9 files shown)


mrobinson:
@glennw r?

Sorry for the size of this patch!

glennw:
This will need a rebase after the mask typo fix landed.

mrobinson:
@glennw Okay. I've pushed the rebased branch.

glennw:
Reviewed 9 of 9 files at r1.

---

*[webrender/src/clip_scroll_node.rs, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/959#-KeWego5SgyTBV46gVXC:-KeWego6H7E6P8o1M7k_:b3knlg8) ([raw file](https://github.com/servo/webrender/blob/5b9fbf3e9558bf442630fecae1e566cd48d68131/webrender/src/clip_scroll_node.rs#L26)):*
> ```Rust
> #[derive(Clone, Debug)]
> pub struct ClipInfo {
>     /// The ClipSource for this node, assuming it is not a reference frame.
> ```

This comment seems partially redundant.

---

*[webrender/src/clip_scroll_node.rs, line 108 at r1](https://reviewable.io:443/reviews/servo/webrender/959#-KeWeZN-kw4sdHj22lgm:-KeWeZN0IkSKgvdq4IpY:b-j9b16u) ([raw file](https://github.com/servo/webrender/blob/5b9fbf3e9558bf442630fecae1e566cd48d68131/webrender/src/clip_scroll_node.rs#L108)):*
> ```Rust
>     pub pipeline_id: PipelineId,
> 
>     /// Parent layer. If this is the same as the node id, this is the root node.
> ```

It seems like this would be clearer as an Option where None signifies the root?

---

*[webrender/src/frame_builder.rs, line 186 at r1](https://reviewable.io:443/reviews/servo/webrender/959#-KeWfilEfYk3v_gG7r3p:-KeWfilEfYk3v_gG7r3q:besu84d) ([raw file](https://github.com/servo/webrender/blob/5b9fbf3e9558bf442630fecae1e566cd48d68131/webrender/src/frame_builder.rs#L186)):*
> ```Rust
>                     return prim_index;
>             }
>             _ => {}
> ```

I'd prefer we list all the enum types here - then we get a compile error when a new type is introduced.

---

*[webrender_traits/src/display_list.rs, line 86 at r1](https://reviewable.io:443/reviews/servo/webrender/959#-KeWbjLyIniUx1cRo-em:-KeWbjLyIniUx1cRo-en:bbljbbl) ([raw file](https://github.com/servo/webrender/blob/5b9fbf3e9558bf442630fecae1e566cd48d68131/webrender_traits/src/display_list.rs#L86)):*
> ```Rust
>     }
> 
>     fn push_new_empty_item(&mut self, item: SpecificDisplayItem) {
> ```

In the future, we should probably move rect/clip from the base display item into the specific item enum, and then this won't be necessary. Doesn't need to be done for this PR though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/959)*

glennw:
This looks mostly good to me - just a few minor style things. I've run the Servo WPT suite locally and done a bit of manual testing, and it seems fine. r=me once those style nits are sorted!

mrobinson:

---

*[webrender/src/clip_scroll_node.rs, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/959#-KeWego5SgyTBV46gVXC:-KeXiiiL9eRe5dzW9ic1:b8torlr) ([raw file](https://github.com/servo/webrender/blob/5b9fbf3e9558bf442630fecae1e566cd48d68131/webrender/src/clip_scroll_node.rs#L26)):*

This comment seems partially redundant.
</blockquote></details>

Okay. I'll clean this up.

---

*[webrender/src/clip_scroll_node.rs, line 108 at r1](https://reviewable.io:443/reviews/servo/webrender/959#-KeWeZN-kw4sdHj22lgm:-KeXilSC6T_G6MH0l-Mf:b-k4utm9) ([raw file](https://github.com/servo/webrender/blob/5b9fbf3e9558bf442630fecae1e566cd48d68131/webrender/src/clip_scroll_node.rs#L108)):*

It seems like this would be clearer as an Option where None signifies the root?
</blockquote></details>

You're right. That is much better and makes the code simpler in other places.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/959)*

mrobinson:
Review status: 6 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[webrender/src/frame_builder.rs, line 186 at r1](https://reviewable.io:443/reviews/servo/webrender/959#-KeWfilEfYk3v_gG7r3p:-KeXmvgllKsLJO9hQTCg:b-896fix) ([raw file](https://github.com/servo/webrender/blob/5b9fbf3e9558bf442630fecae1e566cd48d68131/webrender/src/frame_builder.rs#L186)):*

I'd prefer we list all the enum types here - then we get a compile error when a new type is introduced.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/959)*

staktrace:
I'm seeing an intermittent crash during a particular gecko reftest with this PR. I'm not 100% sure it's introduced by this PR - it might have been there before but we just never tripped over it. Still, the stack looks related enough to this PR that I thought it might be worth mentioning. The error is:

```
thread 'RenderBackend' panicked at 'called `Option::unwrap()` on a `None` value'
stack backtrace:
   1:     0x7fd39688187a - std::sys::imp::backtrace::tracing::imp::write::h3188f035833a2635
   2:     0x7fd396880ccf - std::panicking::default_hook::{{closure}}::h6385b6959a2dd25b
   3:     0x7fd39688031c - std::panicking::rust_panic_with_hook::hf00b8130f73095ec
   4:     0x7fd39687fe9f - std::panicking::begin_panic::h6227f62cb2cdaeb4
   5:     0x7fd39687fe49 - std::panicking::begin_panic_fmt::h173eadd80ae64bec
   6:     0x7fd39688abf6 - core::panicking::panic_fmt::h3b2d1e30090844ff
   7:     0x7fd39688c725 - core::panicking::panic::h2596388ccef1871c
   8:     0x7fd396824fb4 - webrender::clip_scroll_tree::ClipScrollTree::add_node::hae2a81be0ade8a2c
   9:     0x7fd396819dba - webrender::frame_builder::FrameBuilder::push_reference_frame::he9a6ad0bb2f1a254
  10:     0x7fd39681b2bb - webrender::frame::Frame::flatten_stacking_context::hbbf410b4036c9a6d
  11:     0x7fd39681fbd5 - webrender::frame::Frame::flatten_items::hdd0bcbfe02558789
  12:     0x7fd396821855 - webrender::frame::Frame::flatten_items::hdd0bcbfe02558789
  13:     0x7fd39681b619 - webrender::frame::Frame::flatten_stacking_context::hbbf410b4036c9a6d
  14:     0x7fd39681fbd5 - webrender::frame::Frame::flatten_items::hdd0bcbfe02558789
  15:     0x7fd39681b77b - webrender::frame::Frame::flatten_stacking_context::hbbf410b4036c9a6d
  16:     0x7fd396822018 - webrender::frame::Frame::flatten_items::hdd0bcbfe02558789
  17:     0x7fd396821855 - webrender::frame::Frame::flatten_items::hdd0bcbfe02558789
  18:     0x7fd39681b619 - webrender::frame::Frame::flatten_stacking_context::hbbf410b4036c9a6d
  19:     0x7fd39681fbd5 - webrender::frame::Frame::flatten_items::hdd0bcbfe02558789
  20:     0x7fd39681b77b - webrender::frame::Frame::flatten_stacking_context::hbbf410b4036c9a6d
  21:     0x7fd396817b8b - webrender::render_backend::RenderBackend::build_scene::hd6964feae6b3ce39
  22:     0x7fd39680100c - webrender::render_backend::RenderBackend::run::h4b2d24e7ab78f243
  23:     0x7fd3967f8dea - <F as alloc::boxed::FnBox<A>>::call_box::hc39188c73de76516
  24:     0x7fd3968883e3 - std::sys::imp::thread::Thread::new::thread_start::he018521f53b24939
  25:     0x7fd3a37fd6b9 - start_thread
  26:     0x7fd3a288682c - clone
  27:                0x0 - <unknown>
```

staktrace:
Also at the moment this isn't blocking us, as I plan to update WR to the cset immediately preceding your PR. However if it turns out this crash is caused by your PR then it will block the next WR update we try to do.

mrobinson:
@staktrace Do you have any more information on this failure? What test was failing intermittently? I'm going to try to debug this today.

mrobinson:
@staktrace, @nical pointed me in the right direction to find the failing test, so I was able to debug this. The failure was caused by a pre-existing problem that this PR uncovered. I have posted a PR to fix it at https://github.com/servo/webrender/pull/967.

Additionally, I detected an issue with the way that Gecko is building the WebRender display list. It seems to be giving every stacking context a Some(identity) transformation, which will not result in proper behavior for fixed position content as well as leading to a lot of extra work in WebRender. Maybe we can chat discuss later about how to fix this. Thanks for reporting this!

staktrace:
Your PR seems to be working to fix the crash, as far as I can tell. Also I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1345577 about the Some(identity) thing, so we don't forget.

PR #960, opened 2017-03-06
Expose (a subset of) the profile counters via an API.

glennw:
This will allow us to add external reporting of some of the profile
stats for each test run in wrench.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/960)


glennw:
r? @pcwalton or @kvark 

This also provides a FrameId for each profile, to allow matching up the CPU profiles to the GPU profiles (which arrive some number of frames after the corresponding CPU profile).

glennw:
Rebased after the types.rs changes.

PR #961, opened 2017-03-06
Support border-radius with different x/y value for solid border.

demo99:
Handle different x/y border radius. I haven't handled the smoothstep for it since the calculation is a little bit complicated.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/961)

### DIFF
--- a/webrender/res/ps_border.fs.glsl
+++ b/webrender/res/ps_border.fs.glsl
@@ -56,12 +56,57 @@ float alpha_for_solid_border(float distance_from_ref,
   return 1.0 - smoothstep(0.0, 1.0, distance_from_border);
 }
 
+float alpha_for_solid_ellipse_border(vec2 local_pos,
+                                     vec2 inner_radius,
+                                     vec2 outer_radius,
+                                     float pixels_per_fragment) {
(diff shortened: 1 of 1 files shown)


glennw:
Thanks for the PR! I'm unsure about merging this, since I don't think it's going to be feasible to correctly calculate the AA correctly with this technique. If that's the case, I think we might be better off leaving it until we have a solution that correctly handles AA. What do you think?

demo99:
Okay, I'll think how to handle AA for this.

demo99:
@glennw , I add the AA calculation for the ellipse type. We can get the approximate distance to border by calculating the distance between the point to the two focus points. Can you have a review again? Thanks.

glennw:
Thanks for the update! Do you have any information / references on the error metrics using a method like this?

demo99:
I don't have reference about the error metrics. But this method is simple and the error of the 4 points (top, bottom, left, right) is zero. The method can also apply to circle and there will be no error.

glennw:
I think we'll need a better solution long term, but this is a net improvement for now. Thanks!

demo99:
@glennw Can you retry again?  The test failure seems to have nothing to do with the patch.

Issue #958, opened 2017-03-05
Remove types.rs


nox:
This was needed only for serde_codegen support. Now that this is gone, we can remove types.rs and move back each type from it to the module where each of them truly belongs.

nical:
This can be closed.

PR #963, opened 2017-03-06
Remove types.rs and move its types into the relevant files.

nical:
Addresses issue #958.

The rules followed to decide in where to move each type were roughly:
- Above all, types and their impls should be in the same file.
- If a "well defined category" of types and their impls represents more than a 100 lines of code, move it into its own file (example: font.rs and image.rs).
- If something is only used in one file it goes there (for example a lot of the code that moved into api.rs).
- On the other hand if it is not immediately obvious where a type should be (and it's more than a few lines), move it in its own file (like color.rs). 
- display_item.rs contains display item types while display_list.rs contains code to build and serialize display lists. both files being rather big, keep them separate.
- Since all types were exported in webrender_traits' global namespace, all relevant modules ```pub use foo::*;``` there types in as well. If we want to change that, let's do it as a followup since it would add a lot code churn outside of webrender_traits.

The resulting organization is, I think, a lot less surprising. Let me know what you think.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/963)


glennw:
Those guidelines seem reasonable to me - will leave this here for @kvark 's thoughts too.

kvark:
@glennw I haven't gone through the actual code but the guidelines look reasonable to me.

glennw:
Reviewed 10 of 10 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/963)*

PR #965, opened 2017-03-07
Use gl::Gl trait object in webrender

sotaroikeda:
The pull request is preparation of servo/webrender#771. It just replace gleam::gl function calls with function calls to gleam::gl::Gl trait object.

API between winit and glutin is broken now. Then the pull request does not update glutin. The glutin still refers to old cgl and winit. It causes old gleam usage within them. Fixing the break is handled by tomaka/glutin#864.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/965)


sotaroikeda:
@kvark, can you review the patch?

webrender/src/device.rs
```diff
@@ -98,56 +99,56 @@ pub trait FileWatcherHandler : Send {
 }
 
 impl VertexFormat {
-    fn bind(&self, main: VBOId, instance: VBOId, offset: gl::GLuint, instance_stride: gl::GLint) {
-        main.bind();
+    fn bind(&self, gl_: &Rc<gl::Gl>, main: VBOId, instance: VBOId, offset: gl::GLuint, instance_stride: gl::GLint) {
```
kvark: any reason not to simply use `gl: &gl::Gl`?

webrender/src/device.rs
```diff
@@ -652,34 +658,40 @@ impl<T> GpuProfiler<T> {
...
-pub struct GpuMarker(());
+pub struct GpuMarker{
+    gl_: Rc<gl::Gl>,
+}
 
 #[cfg(any(target_arch="arm", target_arch="aarch64"))]
 impl GpuMarker {
-    pub fn new(_: &str) -> GpuMarker {
-        GpuMarker(())
+    pub fn new(gl_: Rc<gl::Gl>, _: &str) -> GpuMarker {
```
kvark: could receive `&Rc<gl::Gl>` and clone only when necessary

kvark:
>API between winit and glutin is broken now.

Why do you need a new glutin?

webrender/src/device.rs
sotaroikeda: When I created the patch, I used self.gl() to get &gl::Gl. In this case, it caused the following error.
> error[E0502]: cannot borrow `*self` as immutable because `self.programs` is also borrowed as mutable

I tried again to remove &Rc<gl::Gl> and succeeded to do it. If &*self.gl is used instead of self.gl(), the compile error was addressed. In a next patch, I will use &*self.gl in places that caused the compile error.

webrender/src/device.rs
sotaroikeda: I will update it in a next patch.

sotaroikeda:
> Why do you need a new glutin?

glutin depends on cgl and winit. Winit depends on cgl. And cgl depends on gleam. Then if glutin is not updated, webrender build continue to depends on old gleam. Cargo.lock is going to have 2 gleam versions(old and new).

sotaroikeda:
@kvark, I updated a patch, can you review again?

kvark:
@sotaroikeda 
Servo is using [servo-glutin](https://crates.io/crates/servo-glutin), which doesn't depend on `winit` (older version). Perhaps, we should work with it instead of the upstream `gluti`n, plus it will be easier to update WR for Servo.

kvark:
@sotaroikeda looks good now, thanks. We just need to resolve the glutin problem before proceeding.

sotaroikeda:
> @sotaroikeda looks good now, thanks. We just need to resolve the glutin problem before proceeding.

I created servo/glutin#120, but it might take time until the request is merged. Until then, we might use fork of servo-glutin.

sotaroikeda:
@kvark, I updated a patch as to use a fork of servo-glutin. Can we use it until servo/glutin#120 is addressed?

sotaroikeda:
@kvark, can you review the pull request again?

kvark:
It's finally happening!
@bors-servo r+

jrmuizel:
This seems to have broken wrench on macOS

kvark:
@jrmuizel fixed it in https://github.com/servo/glutin/pull/123

PR #969, opened 2017-03-08
Add a benchmarking module to wrench.

glennw:
This adds two new commands to wrench:

perf - This runs all the benchmarks in the benchmarks manifest, and records
the timing of various profile counters into a JSON file. This is
very similar to how the reftest code works.

compare_perf - This allows comparing two files output from the perf command above.

There are two new workflows enabled by this change:

(1) Working locally:
 - Run `perf` to get a baseline benchmark.
 - Develop new feature or optimization.
 - Run `perf` to get new results, and use `compare_perf` to compare to baseline.

(2) Automated performance measurement:
 - This is not available yet, but the idea is to upload the results
   of the perf command to a web server, which graphs the results
   over time.

New benchmarks can be added to the benchmarks/ directory, in a similar
way to adding reftests.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/969)

### DIFF
--- a/wrench/src/perf.rs
+++ b/wrench/src/perf.rs
@@ -1,0 +1,281 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use WindowWrapper;
+use serde_json;
+use std::collections::{HashMap, HashSet};
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+use std::sync::mpsc::{channel, Receiver, Sender};
+use webrender_traits::*;
+use wrench::{Wrench, WrenchThing};
+use yaml_frame_reader::YamlFrameReader;
+
+const COLOR_DEFAULT: &'static str = "\x1b[0m";
+const COLOR_RED: &'static str = "\x1b[31m";
+const COLOR_GREEN: &'static str = "\x1b[32m";
+const COLOR_MAGENTA: &'static str = "\x1b[95m";
+
+pub struct Benchmark {
+    pub test: PathBuf,
+}
+
+pub struct BenchmarkManifest {
+    pub benchmarks: Vec<Benchmark>,
+}
+
+impl BenchmarkManifest {
+    pub fn new(manifest: &Path) -> BenchmarkManifest {
+        let dir = manifest.parent().unwrap();
+        let f = File::open(manifest).expect(&format!("couldn't open manifest: {}", manifest.display()));
+        let file = BufReader::new(&f);
+
+        let mut benchmarks = Vec::new();
+
+        for line in file.lines() {
+            let l = line.unwrap();
+
+            // strip the comments
+            let s = &l[0..l.find("#").unwrap_or(l.len())];
+            let s = s.trim();
+            if s.len() == 0 {
+                continue;
+            }
+
+            let mut items = s.split_whitespace();
+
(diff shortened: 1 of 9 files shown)


glennw:
@kvark Of particular note in this PR - making sure you agree with the general approach I've taken for recording statistics. The current strategy is:

* Keep rendering until we have at least 10 recorded frame profiles (for each of CPU and GPU).
* Sort the profiles array by timings.
* Remove the two fastest frames and the two slowest frames.
* Calculate the mean time from the remaining samples.

I'm completely open to changing that method if think something else would be more suitable.

kvark:
@glennw 
> Calculate the mean time from the remaining samples.

This should be the arithmetic average at this step (instead of mean). It's not robust to extreme samples, but that's why we filter out the 2 highest and 2 lowest in the earlier step.

glennw:
@kvark Ummm... isn't the arithmetic average the same as the mean?

kvark:
Oops, you are correct. I was thought mean=middle. Lost in translation.

wrench/src/perf.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    fn render_yaml(&mut self, filename: &Path) -> TestProfile {
+        let mut reader = YamlFrameReader::new(filename);
+        reader.do_frame(self.wrench);
+
+        // wait for the frame
+        self.rx.recv().unwrap();
+
+        // Loop until we get a reasonable number of CPU and GPU
+        // frame profiles. Then take the median.
```
kvark: huh, maybe I just read your mind? :) Comment needs to  be updated

wrench/src/perf.rs
```diff
@@ -0,0 +1,280 @@
...
+              gpu_frame_profiles.len() < 10 {
+            self.wrench.render();
+            self.window.swap_buffers();
+            let (cpu_profiles, gpu_profiles) = self.wrench.get_frame_profiles();
+            cpu_frame_profiles.extend(cpu_profiles);
+            gpu_frame_profiles.extend(gpu_profiles);
+        }
+
+        cpu_frame_profiles.sort_by(|a, b| {
+            assert!(a.draw_calls == b.draw_calls);
```
kvark: nit: there is `assert_eq`

wrench/src/perf.rs
```diff
@@ -0,0 +1,280 @@
...
+        while cpu_frame_profiles.len() < 10 ||
+              gpu_frame_profiles.len() < 10 {
+            self.wrench.render();
+            self.window.swap_buffers();
+            let (cpu_profiles, gpu_profiles) = self.wrench.get_frame_profiles();
+            cpu_frame_profiles.extend(cpu_profiles);
+            gpu_frame_profiles.extend(gpu_profiles);
+        }
+
+        cpu_frame_profiles.sort_by(|a, b| {
```
kvark: nit: I think `sort_by_key` is simpler

wrench/src/perf.rs
```diff
@@ -0,0 +1,280 @@
...
+        gpu_frame_profiles.sort_by(|a, b| {
+            a.paint_time_ns.cmp(&b.paint_time_ns)
+        });
+
+        // Remove the two slowest and fastest frames.
+        let cpu_samples = &cpu_frame_profiles[2..cpu_frame_profiles.len() - 2];
+        let gpu_samples = &gpu_frame_profiles[2..gpu_frame_profiles.len() - 2];
+
+        // Use the mean value from the remaining frames.
+        let composite_time = cpu_samples.iter()
```
kvark: nit: could do `map` followed by `sum`

glennw:
Thanks for the review - comments addressed.

kvark:
Thanks!
@bors-servo r+

PR #972, opened 2017-03-08
Implement repetition and spacing for tiled images.

nical:
Followup of PR #897. This should cover the all of the remaining cases when rendering tiled images (image display item specifically).
If the image is tiled and repeated along the same dimension it is decomposed into a primitive for each repetition, taking spacing into account. These primitives are then split in a per tile basis.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/972)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use app_units::Au;
+use euclid::rect::rect;
 use fnv::FnvHasher;
 use internal_types::{ANGLE_FLOAT_TO_FIXED, AxisDirection};
 use internal_types::{LowLevelFilterOp};
@@ -573,12 +574,13 @@ impl Frame {
                         // The image resource is tiled. We have to generate an image primitive
                         // for each tile.
                         let image_size = DeviceUintSize::new(image.descriptor.width, image.descriptor.height);
-                        self.decompose_tiled_image(scroll_layer_id,
-                                                   context,
-                                                   &item,
-                                                   info,
-                                                   image_size,
-                                                   tile_size as u32);
+                        self.decompose_image(scroll_layer_id,
+                                             context,
+                                             &item.rect,
+                                             &item.clip,
+                                             info,
+                                             image_size,
+                                             tile_size as u32);
                     } else {
                         context.builder.add_image(scroll_layer_id,
                                                   item.rect,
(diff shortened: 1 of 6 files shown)


nical:
Sadly the test tile-repeat-prim-or-decompose.yaml which I just added fails on travis (but not on my computer) due to rounding differences. It would be great to be able to fuzz tests (my attempt at using gecko's fuzzing syntax didn't work).

nical:
I just added some basic support for fuzzy reftests in wrench using the same syntax as gecko. I'll move the commit into a separate PR but just pushing it here temporarily to check that it fixes the test.

kvark:
> due to rounding differences

Did you find out what exactly those differences are?

webrender/src/frame.rs
```diff
@@ -686,10 +688,83 @@ impl Frame {
...
+        let no_vertical_spacing = info.tile_spacing.height == 0.0;
+        if no_vertical_tiling && no_vertical_spacing {
+            self.decompose_image_row(scroll_layer_id, context, item_rect, item_clip, info, image_size, tile_size);
+            return;
+        }
+
+
+        // Decompose each vertical repetition into rows.
+        let layout_stride = info.stretch_size.height + info.tile_spacing.height;
+        let num_repetitions = (item_rect.size.width / layout_stride).ceil() as u32;
```
kvark: shouldn't this be `item_rect.size.height  / layout_stride`?

webrender/src/frame.rs
```diff
@@ -686,10 +688,83 @@ impl Frame {
...
+        for i in 0..num_repetitions {
+            if let Some(row_rect) = rect(
+                item_rect.origin.x,
+                item_rect.origin.y + (i as f32) * layout_stride,
+                item_rect.size.width,
+                info.stretch_size.height
+            ).intersection(&item_rect) {
+                self.decompose_image_row(scroll_layer_id, context, &row_rect, item_clip, info, image_size, tile_size);
```
kvark: it's a bit weird to see `decompose_image_row` called from inside of this loop, since it's supposed to handle image repetitions (which is not the case for this loop, is it?)

webrender/src/frame.rs
```diff
@@ -686,10 +688,83 @@ impl Frame {
...
+        if no_horizontal_tiling && no_horizontal_spacing {
+            self.decompose_tiled_image(scroll_layer_id, context, item_rect, item_clip, info, image_size, tile_size);
+            return;
+        }
+
+        // Decompose each horizontal repetition.
+        let layout_stride = info.stretch_size.width + info.tile_spacing.width;
+        let num_repetitions = (item_rect.size.width / layout_stride).ceil() as u32;
+        for i in 0..num_repetitions {
+            if let Some(img_rect) = rect(
```
kvark: rename to `column_rect` for consistency?

webrender/src/frame.rs
nical: Oops!

webrender/src/frame.rs
nical: It is (or maybe I misunderstand your question). ```decompose_image_row``` handles horizontal repetition (of a given row) and this loop takes care of the vertical repetition by splitting the primitive into rows.

webrender/src/frame.rs
nical: I find ```column_rect``` a bit misleading in the sense that the rect doesn't describe an entire column of the image but just a single repetition within the current row. Granted ```img_rect``` isn't great either since it is not the rect for the entire image.

nical:
For the fuzzy test, if you want to have a look the easiest is to copy paste the entire [log](https://travis-ci.org/servo/webrender/jobs/209337755) in [reftest analyzer](https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml). The max difference is 1 for a lot of pixels and you can see bands of differing pixels in lines perpendicular to the stretching direction.
It and looks very much like the sampling interpolation rounding differently. The "correct" value for these pixels isn't strictly defined, it's really up to the precision/rounding behavior of the sampler.
I'm currently battling with the headless.py locally (for some reason I'm getting a max texture size of zero which causes a panic as soon as we try to add something to the texture caches).

nical:
Yep, OSMesa gives me the same result on my desktop as travis (unsurprisingly).

kvark:
Great, thanks!
@bors-servo r+

PR #975, opened 2017-03-08
Add basic support for pinch zoom and pan.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/975)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -18,7 +18,7 @@ use std::sync::mpsc::Sender;
 use texture_cache::TextureCache;
 use thread_profiler::register_thread_with_profiler;
 use threadpool::ThreadPool;
-use webrender_traits::{DeviceUintPoint, DeviceUintRect, DeviceUintSize};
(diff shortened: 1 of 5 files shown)


glennw:
@mrobinson The review comments don't seem to have been saved?

webrender/src/render_backend.rs
```diff
@@ -416,6 +427,10 @@ impl RenderBackend {
         self.frame.discard_frame_state_for_pipeline(pipeline_id);
     }
 
+    fn device_pixel_ratio(&self) -> f32 {
```
mrobinson: You might consider renaming this to something like 'accumulated_scale_factor' or something like that. In my mind device_pixel_ratio is pegged to be the HiDPI scaling factor, so I think this could be potentially confusing to others as well.

mrobinson: Sorry, for the disappearing comment. I'm not sure what happened, but I've tried to reproduce it above.

glennw:
@bors-servo r=mrobinson

Thanks - fixed those up, and we will rename other instances of that variable in a follow up.

PR #976, opened 2017-03-09
Add basic support for fuzzy reftests in wrench.

nical:
This adds basic support for fuzzy-equal reftests using gecko's reftest syntax. This is important in order to be able to reftest things that involve stretching or anti-aliasing or anything where drivers may decide to round differently.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/976)

### DIFF
--- a/wrench/src/reftest.rs
+++ b/wrench/src/reftest.rs
@@ -7,6 +7,7 @@ use base64;
 use gleam::gl;
 use image::ColorType;
 use image::png::PNGEncoder;
+use parse_function::parse_function;
 use std::cmp;
 use std::fs::File;
 use std::io::{BufRead, BufReader};
@@ -24,6 +25,8 @@ pub struct Reftest {
     op: ReftestOp,
     test: PathBuf,
     reference: PathBuf,
+    max_difference: usize,
+    num_differences: usize,
 }
 
 struct ReftestImage {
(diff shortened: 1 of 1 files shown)


wrench/src/reftest.rs
```diff
@@ -114,29 +117,39 @@ impl ReftestManifest {
...
-                    let reference = dir.join(items.next().unwrap());
+                x => {
+                    let mut max = 0;
+                    let mut count = 0;
+                    // If the first item is "fuzzy(<val>,<count>)" the positions of the operator
+                    // and paths in the array are offset.
+                    // TODO: This is simple but not great because it does not support having spaces
+                    // in the fuzzy syntax, like between the arguments.
+                    let mut offset = 0;
+                    if x.starts_with("fuzzy(") {
```
mrobinson: Here it would be a bit cleaner (and avoid mutability) to write:
```
let (max, count, offset) = if item.starts_with("fuzzy(") {
    ...
    (args[0].parse().unwrap(), args[1].parse().unwrap(), 1)
} else {
    (0, 0, 0)
}
```

wrench/src/reftest.rs
```diff
@@ -114,29 +117,39 @@ impl ReftestManifest {
...
                 }
-                Some(x) => {
-                    let kind = match x {
-                        "==" => ReftestOp::Equal,
-                        "!=" => ReftestOp::NotEqual,
-                        _ => panic!("unexpected match operator"),
-                    };
-                    let test = dir.join(items.next().unwrap());
-                    let reference = dir.join(items.next().unwrap());
+                x => {
```
mrobinson: I think it's better to give this a full name like 'item' or 'item_string'.

wrench/src/reftest.rs
```diff
@@ -114,29 +117,39 @@ impl ReftestManifest {
...
+                    // TODO: This is simple but not great because it does not support having spaces
+                    // in the fuzzy syntax, like between the arguments.
+                    let mut offset = 0;
+                    if x.starts_with("fuzzy(") {
+                        let (_, args) = parse_function(x);
+                        max = args[0].parse().unwrap();
+                        count = args[1].parse().unwrap();
+                        offset = 1;
+                    }
+                    let kind = parse_operator(items[offset]).expect("unexpected match operator");
```
mrobinson: It might be slightly more readable to avoid the variables here and just assign them directly in the struct initialization, but it's up to you.

wrench/src/reftest.rs
```diff
@@ -152,6 +165,14 @@ impl ReftestManifest {
     }
 }
 
+fn parse_operator(s: &str) -> Option<ReftestOp> {
```
mrobinson: Maybe 'string' or 'operator' string instead of 's'?

PR #978, opened 2017-03-10
Allow API clients to specify arbitrary clip primitives

mrobinson:
Modify the API to a transitionally state allowing clients to select
their containing clip/scroll. This will allow Servo and Gecko to
transition to the new style API while maintaining backward compatibility
for a small performance cost. Eventually we will remove the old API.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/978)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -15,10 +15,10 @@ use scene::{Scene, SceneProperties};
 use std::collections::HashMap;
 use std::hash::BuildHasherDefault;
 use tiling::{AuxiliaryListsMap, CompositeOps, PrimitiveFlags};
-use webrender_traits::{AuxiliaryLists, ClipRegion, ColorF, DeviceUintRect, DeviceUintSize};
-use webrender_traits::{DisplayItem, Epoch, FilterOp, ImageDisplayItem, LayerPoint, LayerRect};
-use webrender_traits::{LayerSize, LayerToScrollTransform, LayoutTransform, MixBlendMode};
-use webrender_traits::{PipelineId, PushScrollLayerItem, ScrollEventPhase, ScrollLayerId};
+use webrender_traits::{AuxiliaryLists, ClipDisplayItem, ClipRegion, ColorF, DeviceUintRect};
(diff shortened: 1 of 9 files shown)


mrobinson:
@glennw r?

webrender/src/frame.rs
```diff
@@ -548,8 +535,7 @@ impl Frame {
                                       &iframe_stacking_context,
                                       iframe_clip);
 
-        context.builder.pop_clip_scroll_node();
-        context.builder.pop_clip_scroll_node();
```
kvark: we still push the clip scroll nodes (https://github.com/servo/webrender/pull/978/files#diff-054157e919dfed0477381da46476b556L531). Where do we pop them now?

webrender_traits/src/display_list.rs
```diff
@@ -340,28 +338,45 @@ impl DisplayListBuilder {
...
+            parent_id: *self.clip_stack.last().unwrap(),
             scroll_root_id: scroll_root_id,
         });
 
         self.push_item(item, clip.main, clip);
-        self.scroll_layer_stack.push(scroll_layer_id);
+        id
+    }
+
+    pub fn push_scroll_layer(&mut self,
```
kvark: it looks like this method is here just for the backwards compatibility?

webrender/src/frame.rs
mrobinson: We no longer push the node to a stack (https://github.com/servo/webrender/pull/978/files#diff-d368cc4d16f067bc1480e7806819a74eL339), so perhaps it would be more accurate to rename that method to add_clip_scroll_node.

webrender_traits/src/display_list.rs
mrobinson: That's right. The idea is that this is a transitional PR. We will land this, update all clients to use the new API and then remove the old API.

webrender/src/frame.rs
kvark: Oh, I see. Yeah, renaming it would avoid confusion.

mrobinson: Okay. I've renamed the method. Thanks for the suggestion!

PR #981, opened 2017-03-13
Fix double border's border width.

demo99:
The double border seems to be broke. Maybe I should add a wrench test?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/981)

### DIFF
--- a/webrender/res/ps_border.fs.glsl
+++ b/webrender/res/ps_border.fs.glsl
@@ -244,16 +244,16 @@ vec4 draw_double_edge_vertical(vec2 local_pos,
                                float distance_from_mix_line,
                                float pixels_per_fragment) {
   // Get our position within this specific segment
-  float position = local_pos.x - vLocalRect.x;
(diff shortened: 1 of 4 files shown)


jrmuizel:
> Maybe I should add a wrench test?

Yes please.

demo99:
@glennw r?

glennw:
I'm not sure why these need to be a fuzzy reftest. Shouldn't they be exactly the same pixels? It seems like there is an issue in the double border drawing causing the fuzziness?

demo99:
The fuzzy is due to the AA processing. Solid border calls 'draw_solid_border' and there is no AA processing for top/bottom/right/left. And double border uses 'alpha_for_solid_border' which has AA processing since we need to do some calculations to make it have two lines.

glennw:
OK, so there's two problems here.

The initial bug is solved by this patch, but that then shows a bug in how the AA is working. If I reproduce this same test in Gecko, there's no AA on the double with border.

If we fix the AA up, then we won't need it to be a fuzzy reftest. But that is really a separate issue, so let's merge this and fix that as a follow up.

PR #983, opened 2017-03-14
Remove ServoScrollRootId

mrobinson:
Remove the concept of ServoScrollRootId and just wrap up external ids
into ScrollLayerId. This will allow API clients to specify ids without
having to do complicated bookkeeping when building display lists. In
addition, this makes the code much simpler. This is a major API break
because clips cannot share the same id. This is only possible because
now Servo does not need to split scroll layers / clips.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/983)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,8 +21,8 @@ dependencies = [
  "serde_json 0.9.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-glutin 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.24.0",
- "webrender_traits 0.25.0",
+ "webrender 0.25.0",
+ "webrender_traits 0.26.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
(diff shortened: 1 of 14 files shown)


mrobinson:
@glennw r?

glennw:
@mrobinson Is there a corresponding Servo patch for this that I can do some testing with / use for the next WR update?

mrobinson:
@glennw I've cleaned up my Servo work here: https://github.com/mrobinson/servo/tree/scrolling. I'm getting a couple timeouts locally, but I think they are independent of my change. If you had a moment to run the tests, I would be very grateful. In WebRender I had to revert a commit to run it with the most recent Servo (https://github.com/servo/webrender/commit/59c6fe706cd21728e1853cd5307fb55fe313cb5d).

glennw:

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/983)*

glennw:
@mrobinson OK, this looks good to me. I'd like to hold off on merging it until we have the corresponding Servo PR ready to go for the WR update though. I'm working through the remaining changes in Servo and dependencies to deal with the GL function pointer change, so that shouldn't be too far away.

glennw:
@mrobinson I realized my comment above might be ambiguous - I'm fine with merging this so long as we have a corresponding commit for Servo that I can cherry-pick when I do a WR update. This doesn't need to wait on the GL function pointer changes.

glennw:
r=me once there is a servo patch available that I can use in the WR update.

mrobinson:
Okay. Thanks! The patch is in my scrolling branch here: https://github.com/mrobinson/servo/tree/scrolling. Just let me know if you need my help in any way. I'm happy to do merges or even work on the function pointer change, if you like.

PR #984, opened 2017-03-15
Use Gl::get_type() to select Gl/Gles in WebRender

sotaroikeda:
The pull request makes webrender dynamically switchable between GL and GLES.

It is related to servo/webrender#771

cc @kvark @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/984)


kvark:
I read your branch name as "giggles" 😆

PR #988, opened 2017-03-16
Reduce transforms fillrate, improve AA quality, fix some transform tests.

glennw:
This commit adds a few related but separate changes to the transform
shader:

* Reduce the fillrate required. Instead of calculating a worst case
  bounding rect, it now calculates a minimal set of pixels to run
  the fragment shader on for each quad. The basic idea is to push
  the vertices of the quad along the screen space normals of the
  quad edges by a small number of device pixels. This gives enough
  extra pixels for high quality AA, but minimal wasted FS invocations.

* Add (device) pixel snapping to the transform shader using the same
  method that the non-transform shader uses. This adds some new test
  passes in the WPT suite.

* Use analytic signed distance to apply AA. This gives a higher quality
  AA result than the previous code, and fixes some edge cases where
  the previous AA code was producing incorrect results.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/988)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -503,52 +503,104 @@ struct TransformVertexInfo {
     vec4 clipped_local_rect;
 };
 
+float cross2(vec2 v0, vec2 v1) {
+    return v0.x * v1.y - v0.y * v1.x;
+}
+
+// Return intersection of line (p0,p1) and line (p2,p3)
+vec2 intersect_lines(vec2 p0, vec2 p1, vec2 p2, vec2 p3) {
+    vec2 d0 = p0 - p1;
+    vec2 d1 = p2 - p3;
+
+    float s0 = cross2(p0, p1);
+    float s1 = cross2(p2, p3);
+
+    float d = cross2(d0, d1);
+    float nx = s0 * d1.x - d0.x * s1;
+    float ny = s0 * d1.y - d0.y * s1;
+
+    return vec2(nx / d, ny / d);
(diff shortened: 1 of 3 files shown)


glennw:
The first two images show an outline of the FS invocations. The red outline shows
the extended quad area where AA may be applied.

outline1

outline2

The next two images show the AA being applied to a rotated rect and also
a rect with a perspective transform.

aa1

aa2

mrobinson:
This looks a lot cleaner than my code!

webrender/res/prim_shared.glsl
```diff
@@ -663,20 +715,35 @@ void write_clip(vec2 global_pos, ClipArea area) {
...
+    // anti-aliasing when the fragment has partial coverage.
+    float d = signed_distance_rect(pos,
+                                   local_rect.xy + dxdy,
+                                   local_rect.xy + local_rect.zw - dxdy);
+
+    // Find the appropriate distance to apply the AA smoothstep over.
+    float afwidth = 0.5 / length(fw);
+
+    // Only apply AA to fragments outside the signed distance field.
+    fragment_alpha = smoothstep(afwidth, 0.0, d);
```
kvark: you keep calling `smoothstep` incorrectly:

> genType smoothstep( 	genType edge0, genType edge1, genType x);
> smoothstep performs smooth Hermite interpolation between 0 and 1 when edge0 < x < edge1.
> Results are undefined if edge0 ≥ edge1.

PR #989, opened 2017-03-16
Fix undefined behaviour in prim_shared.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/989)


kvark:
I think we should also clamp `d` to be in 0-1 range

glennw:
I'm not sure I follow why we'd want to clamp ```d``` - could you explain?

kvark:
Nvm, shipit ;)
@bors-servo r+

PR #992, opened 2017-03-19
Ensure that inset box shadows don't fetch texels outside cache rect.

### LINKED ISSUE #990: Drawing two box shadows results in a wierd second drawing
Test branch here - https://github.com/changm/webrender/tree/test

Just run the WR sample app

Some observations:
1) Rendering each of the box shadows independently works fine, its only when they are both drawn. Whichever box shadow is drawn second is the messed up one.
2) The cached box shadows look alright according to apitrace
3) The offsets / rects into the cached box shadow FBO look right
4) The segment rects here are correct and drawing into the right location

It almost seems like the top of the second box shadow is drawing the first cached box shadow, but at 1/4 scale which doesn't really make sense to me.

I'm kind of lost at this point, so any guidance would be helpful!

glennw:
Fixes #990.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/992)

### DIFF
--- a/webrender/res/ps_box_shadow.fs.glsl
+++ b/webrender/res/ps_box_shadow.fs.glsl
@@ -3,7 +3,18 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 void main(void) {
-    vec2 uv = min(vec2(1.0), vMirrorPoint - abs(vUv.xy - vMirrorPoint));
+    // Mirror and stretch the box shadow corner over the entire
+    // primitives.
+    vec2 uv = vMirrorPoint - abs(vUv.xy - vMirrorPoint);
+
+    // Ensure that we don't fetch texels outside the box
+    // shadow corner. This can happen, for example, when
+    // drawing the outer parts of an inset box shadow.
+    uv = clamp(uv, vec2(0.0), vec2(1.0));
+
+    // Map the unit UV to the actual UV rect in the cache.
     uv = mix(vCacheUvRectCoords.xy, vCacheUvRectCoords.zw, uv);
+
+    // Modulate the box shadow by the color.
     oFragColor = vColor * texture(sCache, vec3(uv, vUv.z));
 }


glennw:
r? @kvark 

cc @changm

webrender/res/ps_box_shadow.fs.glsl
```diff
@@ -3,7 +3,18 @@
...
+    // primitives.
+    vec2 uv = vMirrorPoint - abs(vUv.xy - vMirrorPoint);
+
+    // Ensure that we don't fetch texels outside the box
+    // shadow corner. This can happen, for example, when
+    // drawing the outer parts of an inset box shadow.
+    uv = clamp(uv, vec2(0.0), vec2(1.0));
+
+    // Map the unit UV to the actual UV rect in the cache.
     uv = mix(vCacheUvRectCoords.xy, vCacheUvRectCoords.zw, uv);
```
kvark: I don't think that's going to be enough. Given that we are doing a regular `texture` call for sampling, and the sampler is bilinear, and `vCacheUvRectCoords` denotes the bounds of the UV space, sampling on that bound (when uv==0 or uv==1) would touch the neighboring texels. If that sounds reasonable to you, we should subtract half a texel from the bounds.

kvark: Also, it wouldn't harm to use `textureLod` instead of `texel`

glennw:
I *think* in this case we don't need to inset the UVs - the box shadow shader already draws an extra 1-pixel border around the cache element - https://github.com/servo/webrender/blob/9dcc5590421029be3b1bd8a790c684390e084ed3/webrender/src/prim_store.rs#L1062 and https://github.com/servo/webrender/blob/4221987984718bfc6312f92df9501d8fd7a88ea8/webrender/res/ps_box_shadow.vs.glsl#L21

kvark:
Oh, ok, makes sense.
@bors-servo r+

PR #996, opened 2017-03-20
Create a simpler RadialGradient DisplayItem

rlhunt:
RadialGradient has more properties in common with SVG radialGradients
than with CSS radial-gradients().

* CSS gradients need to support ellipse ending shapes, while SVG does not.
* CSS gradients are used as images and so they need tiling, while SVG does not.
* SVG gradients need different start and end center's and radii, while CSS does not.

We can simplify the shaders by changing the DisplayItem to be more in line
with CSS radial-gradients(). The existing RadialGradient can then be
renamed to ComplexRadialGradient

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/996)


rlhunt:
r? @glennw 
cc @lsalzman

mstange:
Note that with `-moz-element` you can use SVG gradients as a CSS background image which can be tiled. So you might want to have a single display item type that just supports everything.

rlhunt:
So another way of doing this could keep one display item with all the properties for both, and then optimize to a better shader for things like the gradient having the same start and end center.

The reason I didn't go that way was mostly because the display item would need to support either
- startRadiusXY and endRadiusXY, which makes the gpu data another vec4 larger and the math in the general case more difficult but also never needed because no radial gradients need that much information
- startRadiusX and endRadiusX and ratioXY (for an ellipse transform), which doesn't seem elegant, but is okay

I am open to ways of keeping one DisplayItem because internally we can optimize for the specific type of gradient.

SVG and `-moz-element` support also seems to be a long time away.

nical:
I think that moz-element should be treated generically (render in an intermediate surface and composite it using the image primitive), since it needs to support any kind of content. I don't think we should worry about it for now.

glennw:
Apologies for not getting to this one yet - it looks like the plan is to simplify the items (based on reading the #gfx logs)? If so, should I wait for those changes before looking at this now?

rlhunt:
@glennw Yes, the new plan is to not split the RadialGradient display item. That should be simpler and easier to maintain. I'll close this PR and open a new one for the new branch of work.

PR #997, opened 2017-03-21
GL_TEXTURE_RECTANGLE texture target support

JerryShih:
@glennw @kvark @nical @sotaroikeda 
The hardware video decoder in mac platform is usually decoding the video content into MacIOSurface. The MacIOSurface could just map to  an GL_TEXTURE_RECTANGLE gl texture without any uploading. That's good for video playing.
These patches try to add a new type of external gl rect image and its corresponding shader.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/997)

### DIFF
--- a/webrender/res/ps_image.fs.glsl
+++ b/webrender/res/ps_image.fs.glsl
@@ -24,12 +24,15 @@ void main(void) {
     // account the spacing in between tiles. We only paint if our fragment does
     // not fall into that spacing.
     vec2 position_in_tile = mod(relative_pos_in_rect, vStretchSize + vTileSpacing);
-    // We clamp the texture coordinates to the half-pixel offset from the borders
-    // in order to avoid sampling outside of the texture area.
     vec2 st = vTextureOffset + ((position_in_tile / vStretchSize) * vTextureSize);
     st = clamp(st, vStRect.xy, vStRect.zw);
 
     alpha = alpha * float(all(bvec2(step(position_in_tile, vStretchSize))));
 
+#ifdef WR_FEATURE_TEXTURE_RECT
+    // textureLod doesn't support sampler2DRect. Use texture() instead.
+    oFragColor = vec4(alpha) * texture(sColor0, st);
+#else
     oFragColor = vec4(alpha) * textureLod(sColor0, st, 0.0);
+#endif
 }
--- a/webrender/res/ps_image.glsl
+++ b/webrender/res/ps_image.glsl
@@ -2,10 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// If this is in WR_FEATURE_TEXTURE_RECT mode, the rect and size use non-normalized
+// texture coordinates. Otherwise, it uses normalized texture coordinates. Please
+// check GL_TEXTURE_RECTANGLE.
 flat varying vec2 vTextureOffset; // Offset of this image into the texture atlas.
 flat varying vec2 vTextureSize;   // Size of the image in the texture atlas.
 flat varying vec2 vTileSpacing;   // Amount of space between tiled instances of this image.
-flat varying vec4 vStRect;     	  // Rectangle of valid texture rect, in st-space.
+flat varying vec4 vStRect;        // Rectangle of valid texture rect.
 
 #ifdef WR_FEATURE_TRANSFORM
 varying vec3 vLocalPos;
--- a/webrender/res/ps_image.vs.glsl
+++ b/webrender/res/ps_image.vs.glsl
@@ -27,16 +27,28 @@ void main(void) {
 
     write_clip(vi.screen_pos, prim.clip_area);
 
+    vTileSpacing = image.stretch_size_and_tile_spacing.zw;
+    vStretchSize = image.stretch_size_and_tile_spacing.xy;
+
+    // If this is in WR_FEATURE_TEXTURE_RECT mode, the rect and size use
+    // non-normalized texture coordinates.
+#ifdef WR_FEATURE_TEXTURE_RECT
+    vec2 texture_size_normalization_factor = vec2(1, 1);
+#else
+    vec2 texture_size_normalization_factor = vec2(textureSize(sColor0, 0));
+#endif
+
     // vUv will contain how many times this image has wrapped around the image size.
-    vec2 texture_size = vec2(textureSize(sColor0, 0));
-    vec2 st0 = res.uv_rect.xy / texture_size;
-    vec2 st1 = res.uv_rect.zw / texture_size;
+    vec2 st0 = res.uv_rect.xy / texture_size_normalization_factor;
+    vec2 st1 = res.uv_rect.zw / texture_size_normalization_factor;
 
     vTextureSize = st1 - st0;
     vTextureOffset = st0;
     vTileSpacing = image.stretch_size_and_tile_spacing.zw;
     vStretchSize = image.stretch_size_and_tile_spacing.xy;
 
-    vec2 half_texel = vec2(0.5) / texture_size;
+    // We clamp the texture coordinates to the half-pixel offset from the borders
+    // in order to avoid sampling outside of the texture area.
+    vec2 half_texel = vec2(0.5) / texture_size_normalization_factor;
     vStRect = vec4(min(st0, st1) + half_texel, max(st0, st1) - half_texel);
 }
--- a/webrender/res/shared.glsl
+++ b/webrender/res/shared.glsl
@@ -33,9 +33,15 @@
 //======================================================================================
 // Shared shader uniforms
 //======================================================================================
+#ifndef WR_FEATURE_TEXTURE_RECT
 uniform sampler2D sColor0;
 uniform sampler2D sColor1;
 uniform sampler2D sColor2;
+#else
+uniform sampler2DRect sColor0;
+uniform sampler2DRect sColor1;
+uniform sampler2DRect sColor2;
+#endif
 uniform sampler2D sDither;
 uniform sampler2D sMask;
 
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -49,7 +49,7 @@ use thread_profiler::{register_thread_with_profiler, write_profile};
 use util::TransformedRectKind;
 use webgl_types::GLContextHandleWrapper;
 use webrender_traits::{ColorF, Epoch, PipelineId, RenderNotifier, RenderDispatcher};
-use webrender_traits::{ExternalImageId, ImageData, ImageFormat, RenderApiSender};
+use webrender_traits::{ExternalImageId, ExternalImageType, ImageData, ImageFormat, RenderApiSender};
 use webrender_traits::{DeviceIntRect, DevicePoint, DeviceIntPoint, DeviceIntSize, DeviceUintSize};
 use webrender_traits::{ImageDescriptor, BlobImageRenderer};
 use webrender_traits::channel;
@@ -1039,10 +1039,10 @@ impl Renderer {
     fn resolve_source_texture(&mut self, texture_id: &SourceTexture) -> TextureId {
         match *texture_id {
             SourceTexture::Invalid => TextureId::invalid(),
-            SourceTexture::WebGL(id) => TextureId::new(id),
-            SourceTexture::External(ref key) => {
+            SourceTexture::WebGL(id) => TextureId::new(id, TextureTarget::Default),
+            SourceTexture::External(external_image) => {
                 *self.external_images
-                     .get(key)
+                     .get(&external_image.id)
                      .expect("BUG: External image should be resolved by now!")
             }
             SourceTexture::TextureCache(index) => {
@@ -1193,24 +1193,31 @@ impl Renderer {
                                                              mode,
                                                              Some(raw.as_slice()));
                                 }
-                                ImageData::ExternalBuffer(id) => {
-                                    let handler = self.external_image_handler
-                                                      .as_mut()
-                                                      .expect("Found external image, but no handler set!");
+                                ImageData::External(ext_image) => {
+                                    match ext_image.image_type {
+                                        ExternalImageType::ExternalBuffer => {
+                                            let handler = self.external_image_handler
+                                                              .as_mut()
+                                                              .expect("Found external image, but no handler set!");
 
-                                    match handler.lock(id).source {
-                                        ExternalImageSource::RawData(raw) => {
-                                            self.device.init_texture(texture_id,
-                                                                     width,
-                                                                     height,
-                                                                     format,
-                                                                     filter,
-                                                                     mode,
-                                                                     Some(raw));
+                                            match handler.lock(ext_image.id).source {
+                                                ExternalImageSource::RawData(raw) => {
+                                                    self.device.init_texture(texture_id,
+                                                                             width,
+                                                                             height,
+                                                                             format,
+                                                                             filter,
+                                                                             mode,
+                                                                             Some(raw));
+                                                }
+                                                _ => panic!("No external buffer found"),
+                                            };
+                                            handler.unlock(ext_image.id);
                                         }
-                                        _ => panic!("No external buffer found"),
-                                    };
-                                    handler.unlock(id);
+                                        _ => {
+                                            panic!("External texture handle should not use TextureUpdateOp::Create.");
+                                        }
+                                    }
                                 }
                                 _ => {
                                     panic!("No suitable image buffer for TextureUpdateOp::Create.");
@@ -1665,16 +1672,16 @@ impl Renderer {
             for deferred_resolve in &frame.deferred_resolves {
                 GpuMarker::fire(self.device.gl(), "deferred resolve");
                 let props = &deferred_resolve.image_properties;
-                let external_id = props.external_id
-                                       .expect("BUG: Deferred resolves must be external images!");
-                let image = handler.lock(external_id);
+                let ext_image = props.external_image
+                                     .expect("BUG: Deferred resolves must be external images!");
+                let image = handler.lock(ext_image.id);
 
                 let texture_id = match image.source {
                     ExternalImageSource::NativeTexture(texture_id) => TextureId::new(texture_id),
                     _ => panic!("No native texture found."),
                 };
 
-                self.external_images.insert(external_id, texture_id);
+                self.external_images.insert(ext_image.id, texture_id);
                 let resource_rect_index = deferred_resolve.resource_address.0 as usize;
                 let resource_rect = &mut frame.gpu_resource_rects[resource_rect_index];
                 resource_rect.uv0 = DevicePoint::new(image.u0, image.v0);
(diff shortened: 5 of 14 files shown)


JerryShih:
There is still a the black screen when I pass an GL_TEXTURE_RECTANGLE gl texture to WR. I have no idea why I can't see the video content. Maybe @nical @kvark @glennw could point out the error for my patch.

webrender/res/shared.glsl
```diff
@@ -33,9 +33,15 @@
 //======================================================================================
 // Shared shader uniforms
 //======================================================================================
+#ifndef WR_FEATURE_TEXTURE_RECT
 uniform sampler2D sColor0;
 uniform sampler2D sColor1;
 uniform sampler2D sColor2;
+#else
+uniform sampler2DRect sColor0;
```
JerryShih: macIOSurface should use sampler2DRect

webrender/res/ps_image_rect.fs.glsl
```diff
@@ -0,0 +1,35 @@
...
+    // not fall into that spacing.
+    vec2 position_in_tile = mod(relative_pos_in_rect, vStretchSize + vTileSpacing);
+    // We clamp the texture coordinates to the half-pixel offset from the borders
+    // in order to avoid sampling outside of the texture area.
+    vec2 st = vTextureOffset + ((position_in_tile / vStretchSize) * vTextureSize);
+    st = clamp(st, vStRect.xy, vStRect.zw);
+
+    alpha = alpha * float(all(bvec2(step(position_in_tile, vStretchSize))));
+
+    oFragColor = vec4(alpha) * texture(sColor0, st);
```
JerryShih: textureLod doesn't support Sample2DRect. Use texture() instead.

webrender/src/renderer.rs
```diff
@@ -659,6 +662,13 @@ impl Renderer {
                                  options.precache_shaders)
         };
 
+        let ps_image_rect = try!{
```
JerryShih: I add a new shader for textureRect. We could also merge this back to the original image shader. I don't like too much "#ifdef" in the code.

webrender/src/tiling.rs
```diff
@@ -68,7 +69,24 @@ impl AlphaBatchHelpers for PrimitiveStore {
...
             PrimitiveKind::BoxShadow => AlphaBatchKind::BoxShadow,
-            PrimitiveKind::Image => AlphaBatchKind::Image,
+            PrimitiveKind::Image => {
+                let image_cpu = &self.cpu_images[metadata.cpu_prim_index.0];
+
+                match image_cpu.color_texture_id {
+                    SourceTexture::External(ext_image) => {
+                        match ext_image.image_type {
+                            ExternalImageType::Texture2DHandle => AlphaBatchKind::Image,
+                            ExternalImageType::TextureRectHandle => AlphaBatchKind::ImageRect,
```
JerryShih: When we have "ExternalImageType::TextureRectHandle", use ImageRect shader for this type.

nical:
For reference, Gecko appears to be using texture2DRect and sampler2DRect:
https://searchfox.org/mozilla-central/rev/ee7cfd05d7d9f83b017dd54c2052a2ec19cbd48b/gfx/layers/opengl/OGLShaderProgram.cpp#368

I would prefer to not add extra shader source files for this (and just ```#define``` our way into implementing this)

kvark:
I'd like to avoid first-class support for texture rectangles in WR. This concept is not supported outside of GL land, so it would make it more difficult to introduce more graphics backends. Instead, I'd like to see as minimal hack as possible to support `MacIOSurface`'s limitations.

nical:
We don't have a choice on OSX as it's the only way to render hardware accelerated video frames as far as I know.

kvark:
@nical yes, that's why I said let's hack this in. Doing a compile switch for the shaders is certainly a smaller hack then exposing the rectangles as a valid texture type.

nical:
Ah yes, I agree. And not such a bad hack since the only things that change are the uniform sampler definition and sampling function name. We already have compile time switches for bigger things like transforms.
Even for the rest of the code I'm sure we can simplify things a fair amount (just need to pass enough information to select the proper texture target).

JerryShih:
@kvark @nical 
> I'd like to avoid first-class support for texture rectangles in WR. This concept is not supported outside of GL land, so it would make it more difficult to introduce more graphics backends.

Yes, this concept is not supported outside of GL land. But why this is the problem if we try to have more graphics backends? We could just have a panic hint for the new backend if we try to use the non-supported TextureRectHandle feature.

And how could we recognize the current usage for texture2d or textureRect without adding the new type of externalImage? That information is required to switch the different shaders for texture2d and textureRect. We might have both usage of texture2d or textureRect at mac platform. 

So, I extend the external image type into:
```
+pub enum ExternalImageType {
+    Texture2DHandle,    // gl TEXTURE_2D handle
+    TextureRectHandle,  // gl TEXTURE_RECT handle
+    ExternalBuffer,
+}
```
https://github.com/servo/webrender/pull/997/commits/0ce9f9741b0c17d482b0f1577f46c635f30d0cc4#diff-9aae73cdde1c142339d62fa040c56ec1

Most of changes are for these new types updating.

kvark:
@JerryShih yeah, having `ExternalImage` case for this seems to be unavoidable.

JerryShih:
@kvark @nical 

> @JerryShih yeah, having ExternalImageType::TextureRectHandle case for this seems to be unavoidable.

Thanks, I will keep working base on the new type "ExternalImageType::TextureRectHandle".

The ps_image_rect and ps_image shader will be merged together. And I will use a feature "#ifdef" to switch the different functionality(texture2D and textureRect).

webrender/src/texture_cache.rs
```diff
@@ -803,24 +820,32 @@ impl TextureCache {
...
-                    ImageData::ExternalHandle(..) => {
-                        panic!("External handle should not go through texture_cache.");
+                    ImageData::External(ext_image) => {
+                        match ext_image.image_type {
+                            ExternalImageType::Texture2DHandle |
+                            ExternalImageType::TextureRectHandle => {
+                                panic!("External texture handle should not go through texture_cache.");
+                            }
+                            ExternalImageType::ExternalBuffer => {
+                                let update_op = TextureUpdate {
```
JerryShih: This block comes from:
https://github.com/servo/webrender/pull/997/commits/b4e91e1b7832ef9a983a5243042d8dd0dff9c91c#diff-d52ac938b0c681e0d7bb3f946e47489fL825

webrender/src/texture_cache.rs
```diff
@@ -781,8 +781,25 @@ impl TextureCache {
...
```
JerryShih: This block comes from:
https://github.com/servo/webrender/pull/997/commits/b4e91e1b7832ef9a983a5243042d8dd0dff9c91c#diff-d52ac938b0c681e0d7bb3f946e47489fL806

webrender/src/resource_cache.rs
```diff
@@ -651,6 +674,87 @@ impl ResourceCache {
             }
         }
     }
+    fn update_texture_cache(cached_images: &mut ResourceClassCache<ImageRequest, CachedImageInfo>,
```
JerryShih: This function is extracted from:
https://github.com/servo/webrender/pull/997/commits/b4e91e1b7832ef9a983a5243042d8dd0dff9c91c#diff-77cbdf7ba9ebae81feb38a64c21b8454L669

webrender/src/resource_cache.rs
```diff
@@ -662,84 +766,32 @@ impl ResourceCache {
...
-                                                  descriptor,
-                                                  filter,
-                                                  image_data,
-                                                  texture_cache_profile);
-
-                        entry.insert(CachedImageInfo {
-                            texture_cache_id: image_id,
-                            epoch: image_template.epoch,
-                        });
+                    ExternalImageType::ExternalBuffer => {
```
JerryShih: We only call update_texture_cache for ExternalBuffer, Raw and Blob image.

webrender/src/resource_cache.rs
```diff
@@ -523,15 +537,24 @@ impl ResourceCache {
     pub fn get_image_properties(&self, image_key: ImageKey) -> ImageProperties {
         let image_template = &self.image_templates[&image_key];
 
-        let external_id = match image_template.data {
-            ImageData::ExternalHandle(id) => Some(id),
-            // raw and externalBuffer are all use resource_cache.
-            ImageData::Raw(..) | ImageData::ExternalBuffer(..) | ImageData::Blob(..) => None,
+        let external_image = match image_template.data {
```
JerryShih: Texture2DHandle, TextureRectHandle => some(ext_image)
Raw, Blob, ExternalBuffer => None

JerryShih:
@kvark @nical @glennw 
The image and image_shader are merged together now. Then, we could set the TEXTURE_RECT_FEATURE flag to work for TEXTURE_RECT type.
```
let ps_image_rect = try!{
    PrimitiveShader::new("ps_image",
                         &mut device,
                         &[ TEXTURE_RECT_FEATURE ],
                         options.precache_shaders)
};
```

Could you please take a look at this?

webrender/res/ps_image.fs.glsl
```diff
@@ -24,12 +24,15 @@ void main(void) {
...
     // not fall into that spacing.
     vec2 position_in_tile = mod(relative_pos_in_rect, vStretchSize + vTileSpacing);
-    // We clamp the texture coordinates to the half-pixel offset from the borders
-    // in order to avoid sampling outside of the texture area.
     vec2 st = vTextureOffset + ((position_in_tile / vStretchSize) * vTextureSize);
     st = clamp(st, vStRect.xy, vStRect.zw);
 
     alpha = alpha * float(all(bvec2(step(position_in_tile, vStretchSize))));
 
+#ifndef WR_FEATURE_TEXTURE_RECT
```
kvark: I'd prefer having `ifdef` path instead of `ifndef`

webrender/res/ps_image.glsl
```diff
@@ -2,10 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// If this is in WR_FEATURE_TEXTURE_RECT mode, the rect and size use non-normalized
+// texture coordinates. Otherwise, it's use normalized texture coordinates. Please
```
kvark: fix `it's use` -> `it uses`

webrender/res/ps_image.vs.glsl
```diff
@@ -27,16 +27,28 @@ void main(void) {
...
     write_clip(vi.screen_pos, prim.clip_area);
 
+    vTileSpacing = image.stretch_size_and_tile_spacing.zw;
+    vStretchSize = image.stretch_size_and_tile_spacing.xy;
+
+    // We clamp the texture coordinates to the half-pixel offset from the borders
+    // in order to avoid sampling outside of the texture area.
     // vUv will contain how many times this image has wrapped around the image size.
+#ifndef WR_FEATURE_TEXTURE_RECT
     vec2 texture_size = vec2(textureSize(sColor0, 0));
```
kvark: could be less of a code change if just `texture_size` is branched out, and the rest of the code is uniform

webrender/src/renderer.rs
```diff
@@ -1641,16 +1663,23 @@ impl Renderer {
...
-                                       .expect("BUG: Deferred resolves must be external images!");
-                let image = handler.lock(external_id);
+                let ext_image = props.external_image
+                                     .expect("BUG: Deferred resolves must be external images!");
+                let image = handler.lock(ext_image.id);
+                let texture_target = match ext_image.image_type {
+                    ExternalImageType::Texture2DHandle => TextureTarget::Default,
+                    ExternalImageType::TextureRectHandle => TextureTarget::Rect,
+                    _ => {
+                        panic!("Not a suitable image type.");
```
kvark: nit: actually include the type in the message

webrender/src/resource_cache.rs
kvark: nit: add new line before the function

kvark: nit: may look better if it's not static
What borrow problems did you encounter? Maybe we can work around them in a cleaner way.

JerryShih:
Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions.

---

*[webrender/res/ps_image.fs.glsl, line 32 at r3](https://reviewable.io:443/reviews/servo/webrender/997#-KgCxhrTuNyJn_jswl6u-r3-32:-KgHuK-FOxZcgm8X_QcJ:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6b5794978164a9f0a00aa4d2f89a087ab6e6420f/webrender/res/ps_image.fs.glsl#L32)):*

I'd prefer having `ifdef` path instead of `ifndef`
</blockquote></details>

Done.

---

*[webrender/res/ps_image.glsl, line 6 at r3](https://reviewable.io:443/reviews/servo/webrender/997#-KgCxhrTuNyJn_jswl6v-r3-6:-KgHuTiJvL_4vWUth6Ec:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6b5794978164a9f0a00aa4d2f89a087ab6e6420f/webrender/res/ps_image.glsl#L6)):*

fix `it's use` -> `it uses`
</blockquote></details>

Done.

---

*[webrender/res/ps_image.vs.glsl, line 37 at r3](https://reviewable.io:443/reviews/servo/webrender/997#-KgCxhrTuNyJn_jswl6w-r3-37:-KgHxe5U4JDxkp2A_mWj:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6b5794978164a9f0a00aa4d2f89a087ab6e6420f/webrender/res/ps_image.vs.glsl#L37)):*

could be less of a code change if just `texture_size` is branched out, and the rest of the code is uniform
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 1673 at r3](https://reviewable.io:443/reviews/servo/webrender/997#-Kfj1kBn6UJI4dNSgKn7-r3-1673:-KgI1BsTinq9NVrRE1bo:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6b5794978164a9f0a00aa4d2f89a087ab6e6420f/webrender/src/renderer.rs#L1673)):*

nit: actually include the type in the message
</blockquote></details>

Done.

---

*[webrender/src/resource_cache.rs, line 677 at r3](https://reviewable.io:443/reviews/servo/webrender/997#-Kfj1kBn6UJI4dNSgKn8-r3-677:-KgI1J_MzGj_mjl3PPHK:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6b5794978164a9f0a00aa4d2f89a087ab6e6420f/webrender/src/resource_cache.rs#L677)):*

nit: add new line before the function
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/997)*

JerryShih:
fn update_texture_cache2(&mut self) {
        println!("test");
    }

    fn finalize_image_request(&mut self,
                              request: ImageRequest,
                              image_data: Option<ImageData>,
                              texture_cache_profile: &mut TextureCacheProfileCounters) {
        let image_template = self.image_templates.get_mut(&request.key).unwrap();
        let image_data = image_data.unwrap_or_else(||{
            image_template.data.clone()
        });

        match image_template.data {
            ImageData::External(ext_image) => {
                match ext_image.image_type {
                    ExternalImageType::Texture2DHandle |
                    ExternalImageType::TextureRectHandle => {
                        // external handle doesn't need to update the texture_cache.
                    }
                    ExternalImageType::ExternalBuffer => {
                        self.update_texture_cache2();   // use member function

                        ResourceCache::update_texture_cache(&mut self.cached_images,
                                                            &mut self.texture_cache,
                                                            self.current_frame_id,
                                                            &request,
                                                            image_template,
                                                            image_data,
                                                            texture_cache_profile);
                    }
                }
            }
            ImageData::Raw(..) | ImageData::Blob(..) => {
                ResourceCache::update_texture_cache(&mut self.cached_images,
                                                    &mut self.texture_cache,
                                                    self.current_frame_id,
                                                    &request,
                                                    image_template,
                                                    image_data,
                                                    texture_cache_profile);
            }
        }
    }

The error message:

    error[E0499]: cannot borrow `*self` as mutable more than once at a time
       --> src/resource_cache.rs:781:25
        |
    768 |         let image_template = self.image_templates.get_mut(&request.key).unwrap();
        |                              -------------------- first mutable borrow occurs here
    ...
    781 |                         self.update_texture_cache2();
        |                         ^^^^ second mutable borrow occurs here
    ...
    803 |     }
        |     - first borrow ends here

---

Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/997#-:-KgI2NZLN309PVOdYox0:b-zbdfks)*

kvark:
@JerryShih 
We can work around it by doing:
```rust
let (old_data, new_data) = {
    let image_template = self.image_templates.get_mut(&request.key).unwrap();
    let new_data = image_data.unwrap_or_else(||{
        image_template.data.clone()
    });
    (image_template.data.clone(), new_data)
};

match old_data {
...
}
```

This would contain the borrow of `self.image_templates` within a small scope and allow borrowing `self` for method calls inside the match arms.

JerryShih:
@kvark 

```
let (old_data, new_data) = {
    let image_template = self.image_templates.get_mut(&request.key).unwrap();
    let new_data = image_data.unwrap_or_else(||{
        image_template.data.clone()
    });
    (image_template.data.clone(), new_data)
};

match old_data {
...
}
```
Sorry, my test code doesn't show all detail. We should not use the cloned image_template here. It will be updated in update_texture_cache().

webrender/src/resource_cache.rs
```diff
@@ -675,13 +675,15 @@ impl ResourceCache {
         }
     }
 
-    fn update_texture_cache(cached_images: &mut ResourceClassCache<ImageRequest, CachedImageInfo>,
-                            texture_cache: &mut TextureCache,
-                            current_frame_id: FrameId,
+    fn update_texture_cache(&mut self,
```
JerryShih: Here is another version of update_texture_cache(). This version spends one more "image_templates" hash table lookup, but it turns to a member function.
I still prefer the static function one.

JerryShih:
r? @kvark 
https://github.com/servo/webrender/pull/997/commits/5da747cfb5912f32285f296783fc6ae4f5b00d0b

JerryShih:
Sorry, rebase again.

kvark:
Thanks @JerryShih !
Do you mind squashing the commits a little before merging?

JerryShih:
@kvark 
I don't know why the comments in reviewable are not shown in github. The concern of my last version is the one additional hash table lookup.

> 
> Here is another version of update_texture_cache(). This version spends one more "image_templates" hash table lookup, but it turns to a member function.
> I still prefer the static function one.

JerryShih:
@kvark 
If ths is not a problem, here are the squashed patches.

> 
> I don't know why the comments in reviewable are not shown in github. The concern of my last version is the one additional hash table lookup.
> 
>     Here is another version of update_texture_cache(). This version spends one more "image_templates" hash table lookup, but it turns to a member function.
>     I still prefer the static function one.
>

kvark:
@JerryShih I see, thanks. The double lookup should be fine (until we get deeper into the optimization stage), given that the relevant hashmap cell is already in the cache by that time.
@bors-servo r+

PR #998, opened 2017-03-21
Fix groove and ridge while there is no border radii

demo99:
Currently draw_mixed_border seems to only consider radii case. I add some code to handle non-radii case. I still need fuzzy for the reftests since in the test I set different border color to solid border and we'll  mix the color for the corner. I haven't figured out an easy way to handle it for groove/ridge.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/998)


glennw:
Let's merge this, but I think we need to reconsider the border shader altogether - it's too complex for a single fragment shader. I'm thinking something along the lines of a cache render task for the complex parts / segments of borders, that are then drawn together as a combination of primitives / subrects or something like that. Will try to write up something more detailed in an issue.

PR #999, opened 2017-03-21
Use hyphens instead of underscores for wrench keys

rlhunt:
Right now it's a bit of a mess. I propose we standardize on hypens.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/999)

### DIFF
--- a/wrench/benchmarks/transforms-simple.yaml
+++ b/wrench/benchmarks/transforms-simple.yaml
@@ -1,10 +1,10 @@
 ---
 root:
   items:
-    - type: stacking_context
+    - type: stacking-context
       bounds: [0, 0, 1024, 1024]
       items:
-      - type: stacking_context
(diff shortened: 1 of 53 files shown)


wrench/res/gradient.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
mrobinson: Did you include this file by accident?

rlhunt: Whoops, yeah that's for a different patch. I'll update.

rlhunt:
Okay, I've removed the file.

kvark:
I don't see any consensus in the world about this naming, and hyphens are easier to type, so 👍

glennw:
@bors-servo r+

Much better!

PR #1000, opened 2017-03-22
Add support for both RGBA8 and A8 cache render targets.

glennw:
Switch clip masks to use the A8 cache target. This improves
memory usage and performance of clip mask rendering and fetch.

In the future, we can move other effects to the A8 texture,
such as the box shadow corner shader.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1000)


glennw:
r? @kvark 

I've changed around the ordering of the main render loop a bit - it should be equivalent as far as I can tell. I've run all the WPT/CSS tests and also done some manual Servo testing of various websites with this patch, and all appears to be working fine.

webrender/src/renderer.rs
```diff
@@ -1694,25 +1668,52 @@ impl Renderer {
...
+                                             RenderTargetMode::LayerRenderTarget(target_count as i32),
+                                             None);
+                }
+                if let Some(texture_id) = pass.alpha_texture_id {
+                    let target_count = pass.required_target_count(RenderTargetKind::Alpha);
+                    self.device.init_texture(texture_id,
+                                             frame.cache_size.width as u32,
+                                             frame.cache_size.height as u32,
+                                             ImageFormat::A8,
+                                             TextureFilter::Nearest,
```
kvark: Is this deliberate?

webrender/src/renderer.rs
```diff
@@ -1722,34 +1723,78 @@ impl Renderer {
...
+                                                 size.width as f32,
+                                                 0.0,
+                                                 size.height as f32,
+                                                 ORTHO_NEAR_PLANE,
+                                                 ORTHO_FAR_PLANE);
+                }
 
-                for (target_index, target) in pass.targets.iter().enumerate() {
-                    let render_target = target_id.map(|texture_id| {
+                self.device.bind_texture(TextureSampler::CacheA8, src_alpha_id);
```
kvark: There is an option to call them `CacheAlpha`/`CacheColor`, but I'm not seeing it particularly superior

kvark:
Woot, #1000 !
@bors-servo r+

PR #1002, opened 2017-03-22
Stop clipping on stacking context boundaries

mrobinson:
Now that clipping regions are first-class elements in the display list,
it doesn't make as much sense to clip along stacking context
boundaries. Additionally, this removes a lot of complexity around the
difference between stacking context bounds and overflow.

Using the transparent rectangle test, I couldn't detect a difference in
performance, and that creates a lot of stacking contexts.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1002)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,8 +21,8 @@ dependencies = [
  "serde_json 0.9.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-glutin 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.25.0",
- "webrender_traits 0.26.0",
+ "webrender 0.26.0",
+ "webrender_traits 0.27.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
@@ -886,7 +886,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "webrender"
-version = "0.25.0"
+version = "0.26.0"
 dependencies = [
  "angle 0.1.2 (git+https://github.com/servo/angle?branch=servo)",
  "app_units 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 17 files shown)


mrobinson:
@kvark or @glennw r?

glennw:
This generally looks good to me, but causes the following WPT failures:

```
  ▶ FAIL [expected PASS] /_mozilla/css/viewport_meta.html
  └   → /_mozilla/css/viewport_meta.html d58106cabc91de4e78009ef8df86a350e85fd75e
/_mozilla/css/viewport_rule_ref.html 7ecbe5bddfcf0b411f3e39faedb840a139035a34
Testing d58106cabc91de4e78009ef8df86a350e85fd75e == 7ecbe5bddfcf0b411f3e39faedb840a139035a34

  ▶ FAIL [expected PASS] /_mozilla/css/viewport_rule.html
  └   → /_mozilla/css/viewport_rule.html d58106cabc91de4e78009ef8df86a350e85fd75e
/_mozilla/css/viewport_rule_ref.html 7ecbe5bddfcf0b411f3e39faedb840a139035a34
Testing d58106cabc91de4e78009ef8df86a350e85fd75e == 7ecbe5bddfcf0b411f3e39faedb840a139035a34
```

mrobinson:
@glennw I could be wrong, but I think my change actually fixes the behavior of these tests and the reference reflects the incorrect behavior. With my change the width of the outer div is 240px, which seems like the correct width when it's specified as 100vw and the viewport width is 240px. A vw should be 1% of the viewport width, so 100% should be 240px. I've confirmed this is the correct width of div in both Firefox and Chromium using the web inspector to modify the viewport width. It seems like these references captured the broken behavior of the previous code.

emilio:
@mrobinson is right, the way the viewport rule works right now in Servo is broken, i was actually surprised those tests passed :)

mrobinson:
@emilio It seems like they were written taking the bug into account.

glennw:
OK, we can disable those tests when this lands then!

jdm:
We should either remove the tests or update them to pass; disabling tests that do not pass in any browser seems silly to me.

glennw:
@mrobinson is going to update the tests to work correctly as part of landing the next WR update.

PR #1003, opened 2017-03-22
Implement sub-image updates.

nical:
Add an optional dirty rect to the update_image command (None means invalidate all), and only upload the required part of the image if the image was already in the texture cache. This will become more important when used with blob images since the blob image renderer will use this information to avoid repainting the entire image when possible.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1003)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -147,8 +147,8 @@ impl RenderBackend {
                             }
                             self.resource_cache.add_image_template(id, descriptor, data, tiling);
                         }
-                        ApiMsg::UpdateImage(id, descriptor, bytes) => {
-                            self.resource_cache.update_image_template(id, descriptor, bytes);
+                        ApiMsg::UpdateImage(id, descriptor, bytes, dirty_rect) => {
+                            self.resource_cache.update_image_template(id, descriptor, bytes, dirty_rect);
                         }
                         ApiMsg::DeleteImage(id) => {
                             self.resource_cache.delete_image_template(id);
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -23,7 +23,7 @@ use texture_cache::{TextureCache, TextureCacheItemId};
 use thread_profiler::register_thread_with_profiler;
 use webrender_traits::{Epoch, FontKey, GlyphKey, ImageKey, ImageFormat, ImageRendering};
 use webrender_traits::{FontRenderMode, ImageData, GlyphDimensions, WebGLContextId};
-use webrender_traits::{DevicePoint, DeviceIntSize, ImageDescriptor, ColorF};
+use webrender_traits::{DevicePoint, DeviceIntSize, DeviceUintRect, ImageDescriptor, ColorF};
 use webrender_traits::{ExternalImageId, GlyphOptions, GlyphInstance, TileOffset, TileSize};
 use webrender_traits::{BlobImageRenderer, BlobImageDescriptor, BlobImageError};
 use threadpool::ThreadPool;
@@ -110,6 +110,7 @@ struct ImageResource {
     descriptor: ImageDescriptor,
     epoch: Epoch,
     tiling: Option<TileSize>,
+    dirty_rect: Option<DeviceUintRect>
 }
 
 struct CachedImageInfo {
@@ -285,6 +286,7 @@ impl ResourceCache {
             data: data,
             epoch: Epoch(0),
             tiling: tiling,
+            dirty_rect: None,
         };
 
         self.image_templates.insert(image_key, resource);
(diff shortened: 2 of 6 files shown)


webrender/src/resource_cache.rs
```diff
@@ -317,6 +320,11 @@ impl ResourceCache {
             data: ImageData::new(bytes),
             epoch: next_epoch,
             tiling: None,
+            dirty_rect: match (dirty_rect, prev_dirty_rect) {
```
glennw: I don't really understand this part. I'm assuming the union() here is for the case where you update different parts of the same image multiple times in one frame? But then, the second update loses the bytes from the original update doesn't it?

webrender/src/texture_cache.rs
```diff
@@ -721,14 +722,28 @@ impl TextureCache {
...
             ImageData::Raw(bytes) => {
-                TextureUpdateOp::Update {
-                    page_pos_x: existing_item.allocated_rect.origin.x,
-                    page_pos_y: existing_item.allocated_rect.origin.y,
-                    width: descriptor.width,
-                    height: descriptor.height,
-                    data: bytes,
-                    stride: descriptor.stride,
-                    offset: descriptor.offset,
+                if let Some(dirty) = dirty_rect {
```
glennw: This ```if let``` is better as a ```match``` since there is an ```else``` clause anyway.

webrender/src/render_backend.rs
```diff
@@ -147,8 +147,8 @@ impl RenderBackend {
                             }
                             self.resource_cache.add_image_template(id, descriptor, data, tiling);
                         }
-                        ApiMsg::UpdateImage(id, descriptor, bytes) => {
-                            self.resource_cache.update_image_template(id, descriptor, bytes);
+                        ApiMsg::UpdateImage(id, descriptor, bytes, dirty_rect) => {
```
kvark: naming nit: this is not a "dirty" rectangle, just the one to update. "dirty" brings on more semantics than the interface really provides.

webrender/src/texture_cache.rs
```diff
@@ -721,14 +722,28 @@ impl TextureCache {
...
+                    let stride = descriptor.compute_stride();
```
kvark: that's a bit confusing that we have both `descriptor.stride` and `descriptor.compute_stride()`

webrender/src/resource_cache.rs
```diff
@@ -317,6 +320,11 @@ impl ResourceCache {
             data: ImageData::new(bytes),
```
kvark: where does the old image data go? Looks like it gets lost

webrender/src/texture_cache.rs
```diff
@@ -721,14 +722,28 @@ impl TextureCache {
...
+                    let offset = descriptor.offset + dirty.origin.y * stride + dirty.origin.x;
```
kvark: gotta double-check this line, but it depends on the answers to previous questions

webrender/src/resource_cache.rs
nical: The case I'm trying to support is updating an image several times (either several times in the same frame or across several frames if the image isn't requested (not in the viewport?)). In this situation the idea is to keep a dirty rectangle containing all dirty rectangles since the last time the texture_cache was updated.

I'm not sure what you mean by losing the bytes of the original update.

webrender/src/texture_cache.rs
nical: What you mean is that ImageDescriptor::stride being an option is confusing, and that's true (but it's convenient because in a lot of cases we don't have a stride larger than ```width * bpp```).
You can fix that by making the stride not optional, and computing it in a construction function if you want.

kvark: right. It's not related to this PR though

webrender/src/resource_cache.rs
nical: The old image data is replaced by the new one, yes. The underlying buffer is reference counted, not sure what you mean by lost.

nical:
Yes, I considered only passing the updated rect's worth of pixel data in the update message, but that makes the implementation a lot more complicated, and isn't necessarily more efficient because we currently save copies thanks to out image data being immutable and atomically ref counted.

I'm not against doing it this way but I would prefer to have a compelling use case to motivate the extra complexity.

webrender/src/render_backend.rs
nical: I can rename that updated_rect if you prefer.

glennw:
@nical OK, that makes sense. Let's get this merged - but it would be good to change that ```if let``` to a ```map``` as a follow up, which is a bit tidier.

PR #1004, opened 2017-03-22
Allow updating images that are not raw buffers

nical:
```update_image``` currently takes a vector of bytes and the implementation expects this vector to be regular image data. We need to be able to update blob images as well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1004)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -29,6 +29,8 @@ use webrender_traits::{BlobImageRenderer, BlobImageDescriptor, BlobImageError};
 use threadpool::ThreadPool;
 use euclid::Point2D;
 
+const DEFAULT_TILE_SIZE: TileSize = 512;
+
 thread_local!(pub static FONT_CONTEXT: RefCell<FontContext> = RefCell::new(FontContext::new()));
 
 type GlyphCache = ResourceClassCache<RenderedGlyphKey, Option<TextureCacheItemId>>;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by nical** [comment]
This PR is written on top of PR #1003, so the latter will show up in the diff until it lands.
With update_image taking ImageData instead of a bag of bytes, I think that it makes sense to allow updating external images (by removing the assertion), or at least explain why we can't update external image handles (if we can't).

r? @glennw

**Comment by glennw** [comment]
@nical Could you rebase this on master to remove that extra commit?

**Comment by nical** [comment]
The extra commit is the followup nit fix from PR #1003, it is very small and somewhat close to what's happening in this PR so I put it here. Do you want me to extract it in a separate pull request ?

How do you feel about removing the assertion in resource_cache that prevents from updating an external image? I'd be happy to do it in this PR.
@JerryShih do you know if it makes sense to be able to update an external image handle (now that update_image can take any image type and not just an image buffer)?

**Comment by JerryShih** [comment]
> How do you feel about removing the assertion in resource_cache that prevents from updating an external image? I'd be happy to do it in this PR.
> @JerryShih do you know if it makes sense to be able to update an external image handle (now that update_image can take any image type and not just an image buffer)?

@nical 
Yes, it could be. But, the size and the format of new image should be the same as the original one. We haven't deal with the different size image in resource cache. And the update_image also needs to trigger the render_backend for the new display item. Do we have benefit with that?

**Comment by nical** [comment]
I clarified things a bit by asserting that the format and size don't change, and I fixed the way tiling is passed: if the size was larger than the maximum texture size we used to force the tile size at 512 instead of using the tiling parameter, we also used to loose the tiling information in update_image.
I also made updating an image that doesn't exist a panic. Long term I think that we should gracefully recover from these with an error instead of panicking but just transparently creating the image seems like papering over bugs. I hope that servo does not rely on this behavior.

r? @glennw

File: webrender/src/resource_cache.rs [hunk 1]
```diff
@@ -254,6 +256,16 @@ impl ResourceCache {
         self.texture_cache.max_texture_size()
     }
 
+    fn should_tile(&self, descriptor: &ImageDescriptor, data: &ImageData) -> bool {
+        let limit = self.max_texture_size();
+        return match data {
+            // Tiled external images are not implemented.
+            &ImageData::ExternalHandle(_) => false,
+            &ImageData::ExternalBuffer(_) => false,
+            _ => { descriptor.width > limit || descriptor.height > limit }
```
glennw: As a follow up, could we explicitly list all the variations here rather than _ ? I like the safety of a compiler bug when new enum variants are introduced.

PR #1005, opened 2017-03-23
Splitting the clipped rectangles into opaque/transparent sub-parts

kvark:
This is a simple optimization targeted at the case of large rectangles with rounded cornered clips. Instead of moving it into the transparent bin and rendering late, we are computing the inner intersection of the clip items and splitting the rectangle into opaque/transparent parts.

On my machine, this reduces the GPU time of https://github.com/servo/servo by approximately 10%: from 4.6ms to 3.9ms. More profiling might be needed, especially on windows.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1005)


glennw:
Nice simple change!

@bors-servo r+

PR #1007, opened 2017-03-23
Interpolate gradient stops with premultiplied alpha

rlhunt:
Gradient stop colors are supposed to be interpreted as premultiplied[1].

[1] https://www.w3.org/TR/2012/CR-css3-images-20120417/#color-stop-syntax

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1007)


rlhunt:
Not sure if this conflicts with #966.

Test yaml to show the difference:
```yaml
---
root:
  items:
    - type: stacking-context
      bounds: 50 50 200 200
      items:
        - type: gradient
          bounds: 0 0 200 200
          start: 0 100
          end: 200 100
          stops: [0.0, blue, 0.5, [0.0, 0.0, 0.0, 0.0], 1.0, red]
```
With premultiplied alpha,
pre
And without,
non-pre
Which matches the example the spec gives.

glennw:
Can we add a wrench reftest for this?

rlhunt:
We should be able to add a test for this, I'll add one.

I just noticed there is some trailing whitespace in the commit, and more importantly I tested this based on branch with changes to gradients. I rebased on master and just noticed it's not drawing the same anymore. I'll need to investigate more.

rlhunt:
I tracked down the issue that I was noticing and it's unrelated to premultiplied alpha. So I've updated this PR with wrench reftests for each of the gradient types. With premultiplied alpha the test should just be white, but without it will draw a solid gray box.

I also added some syntax to wrench for specifying a blank page. It's not necessary but I'd rather not have a bunch of `blank.yaml` files lying around.

wrench/src/reftest.rs
```diff
@@ -39,6 +40,18 @@ enum ReftestImageComparison {
 }
 
 impl ReftestImage {
+    fn blank(size: DeviceUintSize) -> ReftestImage {
+        ReftestImage {
+            data: [255, 255, 255, 255]
```
kvark: could just do `vec![0xFF, (size.width * size.height * 4) as usize]`

wrench/src/reftest.rs
```diff
@@ -96,6 +109,65 @@ impl ReftestImage {
     }
 }
 
+enum ReftestScene {
+    Blank,
```
kvark: I don't think `Blank` logic here is worth it, given that it could be replaced by a single `wrench/reftests/gradient/blank.yaml` file

webrender/src/tiling.rs
```diff
@@ -129,7 +129,10 @@ impl AlphaBatchHelpers for PrimitiveStore {
                     BlendMode::Alpha
                 }
             }
-            PrimitiveKind::Image => {
+            PrimitiveKind::Image |
+            PrimitiveKind::AlignedGradient |
+            PrimitiveKind::AngleGradient |
+            PrimitiveKind::RadialGradient => {
```
kvark: given that you added all the primitive kinds, let's remove the `_` case at the end?

rlhunt: We are still missing the cases for Rectangle, YuvImage, Border.

wrench/src/reftest.rs
rlhunt: Yeah, I'm inclined to agree now that I think about it.

kvark:
Thanks @rlhunt !
@bors-servo r+

PR #1008, opened 2017-03-24
Fix incorrect yaml type in wrench reftests

rlhunt:
Looks like a couple of tests were merged with underscores. I think they are passing because wrench will skip over the incorrect node, leaving just a blank page for the ref and test.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1008)

### DIFF
--- a/wrench/reftests/border/border-groove-simple-ref.yaml
+++ b/wrench/reftests/border/border-groove-simple-ref.yaml
@@ -1,7 +1,7 @@
 ---
 root:
   items:
-    - type: stacking_context
+    - type: stacking-context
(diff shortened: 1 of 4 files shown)


glennw:
@bors-servo r+

We should perhaps add a way to verify that tests aren't empty or something like that?

rlhunt:
@glennw That could be useful. Some gecko reftests do a `!= about:blank` to verify that something draws, which would make that more complicated. No wrench tests do that yet, but it might happen.

I think just failing or panicking when we read an unknown node might be a good solution.

glennw:
That sounds like a good idea.

PR #1009, opened 2017-03-24
Allow to not depend on offscreen_gl_context

sotaroikeda:
It is created based on #773.

cc @emilio, @kvark, @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1009)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ script:
   - servo-tidy
   - if [ $BUILD_KIND = DEBUG ]; then (cd webrender_traits && cargo test --verbose --features "ipc"); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd webrender_traits && cargo test --verbose); fi
+  - if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo build --verbose --no-default-features); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo test --verbose); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd sample && cargo test --verbose); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd sample && cargo build --verbose --features=profiler); fi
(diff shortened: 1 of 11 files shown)


sotaroikeda:
A priority became higher because of Bug 1349967[1]

[1]
  https://bugzilla.mozilla.org/show_bug.cgi?id=1349967

nical:
I would prefer it if instead of adding conditional compilation all over the code we would just contain that inside of WebGLContextWrapper (and have a WebGLContextWrapper with mostly empty functions if the webrender feature isn't enabled.

webrender/src/record.rs
```diff
@@ -77,6 +78,19 @@ pub fn should_record_msg(msg: &ApiMsg) -> bool {
         &ApiMsg::TickScrollingBounce |
         &ApiMsg::WebGLCommand(..) =>
```
kvark: should just apply the `cfg` to this item only

sotaroikeda:
Updated a patch as to reduce cfg.

sotaroikeda:
@kvark, can you review the pull request again?

emilio:
I guess another alternative for this would be to depend on `offscreen_gl_context`, but without the `x11` feature (which is enabled by default).

This would make it just use the `not_implemented` platform by default, which would mean that you wouldn't need any `cfg` at all I think.

kvark:
Oh interesting, @emilio !
@sotaroikeda would you want to try that idea out? Should be simpler.

sotaroikeda:
It was my first proposal, but now I prefer current approach since it could make crate dependency of gecko simpler.

sotaroikeda:
Current approach removes several crates from /third_party/rust of gecko.

nical:
I like the idea of avoiding unnecessary dependencies in gecko's repository.

emilio:
Sounds good, then this looks good to me, does anybody else have any concern with the current approach? If not, feel free to r=me.

kvark:
Sounds like a consensus to me :)
@bors-servo r=emilio

Issue #911, opened 2017-02-21
Add wrench vs image comparison test mode.


glennw:
Initially this could be exact pixel comparison only, but eventually we'll need to work out some support for "fuzzy" tests that differ slightly depending on the rasterizer.

nical:
Should this be closed now that we have initial support for fuzzy reftests? Or did you also mean some syntax to compare directly against an image directly without creating a yaml file?

kvark:
Yes, we still need the test against image logic. Fuzzy tests are pre-requisite.

PR #1010, opened 2017-03-24
Adjust stops colors for aligned gradients relative to segment_rect

rlhunt:
When there are multiple gradient stops for an aligned gradient,
the gradient stop colors are incorrectly adjusted. The weights
should be relative to the current segment rect not the entire
gradient rect.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1010)

### DIFF
--- a/webrender/res/ps_gradient.vs.glsl
+++ b/webrender/res/ps_gradient.vs.glsl
@@ -15,31 +15,43 @@ void main(void) {
     vec4 adjusted_color_g0 = g0.color;
     vec4 adjusted_color_g1 = g1.color;
     if (gradient.start_end_point.y == gradient.start_end_point.w) {
-        vec2 x = mix(gradient.start_end_point.xx, gradient.start_end_point.zz,
-                     vec2(g0.offset.x, g1.offset.x));
+        // Calculate the x coord of the gradient stops
+        vec2 g01_x = mix(gradient.start_end_point.xx, gradient.start_end_point.zz,
+                         vec2(g0.offset.x, g1.offset.x));
+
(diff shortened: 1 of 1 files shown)


rlhunt:
Test yaml:
```yaml
---
root:
  items:
        - type: gradient
          bounds: [0, 0, 200, 200]
          start: 0 100
          end: 200 100
          stops: [0.0, red, 0.5, green, 1.0, blue]
```
Before:
before

After:
after

The gradient in this example is split in half. The problem is the adjusted_color used for the middle vertices. Because they are 0.5 relative to gradient_start, gradient_end, they are blended to be 0.5 the first stop and 0.5 the second stop. They should be measured by the segment rect they are on.

kvark:
Looks great!
I suppose we'll need #911 done before adding a test for this.
@bors-servo r+

kvark:
Added a reftest in #1012

glennw:
@rlhunt @kvark This PR broke one of the reftests in the Servo test suite. Would you be able to investigate the cause?

```
  ▶ FAIL [expected PASS] /_mozilla/css/border-image-linear-gradient.html
  └   → /_mozilla/css/border-image-linear-gradient.html 0504f258c4556f71424911c4acbeb7a18943a9ef
/_mozilla/css/border-image-linear-gradient-ref.html e68bebc2dbcc53b852fd3b6e77485a9031fb5f35
Testing 0504f258c4556f71424911c4acbeb7a18943a9ef == e68bebc2dbcc53b852fd3b6e77485a9031fb5f35
```

glennw:
@rlhunt Did you have a chance to look into this?

PR #1011, opened 2017-03-27
Optimize GPU time spent in clears.

glennw:
For alpha and color render targets, track the used rect in
the texture allocator. Use this to apply a scissor rect when
clearing the targets.

On my GPU (HD530) this brings the GPU time on https://github.com/servo/servo
down from 2.7ms to 2.2ms. Other sites that use 1+ render targets also
see significant wins for GPU time.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1011)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -42,7 +42,7 @@ use std::sync::mpsc::{channel, Receiver, Sender};
 use std::thread;
 use texture_cache::TextureCache;
 use threadpool::ThreadPool;
-use tiling::{AlphaBatchKind, BlurCommand, Frame, PrimitiveBatch, PrimitiveBatchData};
+use tiling::{AlphaBatchKind, BlurCommand, Frame, PrimitiveBatch, PrimitiveBatchData, RenderTarget};
 use tiling::{AlphaRenderTarget, CacheClipInstance, PrimitiveInstance, ColorRenderTarget, RenderTargetKind};
 use time::precise_time_ns;
 use thread_profiler::{register_thread_with_profiler, write_profile};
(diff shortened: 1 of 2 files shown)


jrmuizel:
It seems like this could negatively interact with some of the fast-clear optimizations that gpus/drivers can do.

glennw:
It's possible on mobile / tiled GPUs perhaps. It seems like a large optimization on everything I've tested so far though, so I think it's fine to apply this win now, and add a different code path if/when we see some profiles that don't get a win from this.

nical:
It would be nice to clearly document in the code which kind of GPUs benefits from this and that it is an optimization that may not make sense on tiled GPUs. I can see myself forgetting and in a year test it on a tiled gpu and rip it all out without knowing that it might be a win on intel chips (All GPUs in the current generation except intel integrated ones are tiled now, even on desktop).

kvark:
@nical 
>All GPUs in the current generation except intel integrated ones are tiled now, even on desktop

There is different ways of tiling, or rather - using tiles for rendering.

I agree that this is a non-optimization on hardware that supports clear meta-data, which is at least all of AMD starting with GCN architecture. This meta-data is typically cleared by a compute shader, but setting any scissor state would cause the driver to go the slower path of just render a rectangle instead.

I propose to proceed as follows:
  1. benchmark on NV and AMD, bonus points for some Android device
  2. have a switch for this optimization that is only enabled on Intel

glennw:
I profiled this on my nVidia GPU, which is ```GeForce GTX 960M NVIDIA 375.39```. I also saw very significant performance gains there (perhaps more than the Intel GPU, on average).

I've added comments mentioning that this is a large win on Intel/nVidia hardware, but may not be on all GPUs. I don't have any AMD hardware to benchmark.

I think it makes sense to merge this - it's a very significant performance win, and (I assume) the majority of our target market will be running Intel GPUs. We can always benchmark and measure on other GPUs in the future.

kvark:
Thanks @glennw !
@bors-servo r+

PR #1012, opened 2017-03-27
Image reftests

### LINKED ISSUE #911: Add wrench vs image comparison test mode.

kvark:
Fixes  #911
Includes a reftest for gradient stops. ~~Might need some fuzzying.~~
r? @glennw 

P.S. I figured we better get the testing infrastructure early (as opposed to waiting for external contributors and making it an introductory task), so that more features that are getting in are able to supply the tests right away (in particular, gradient stuff).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1012)


wrench/src/reftest.rs
```diff
@@ -244,8 +249,8 @@ impl<'a> ReftestHarness<'a> {
                              "REFTEST TEST-UNEXPECTED-FAIL", name,
                              "image comparison, max difference", max_difference,
                              "number of differing pixels", count_different);
-                    println!("REFTEST   IMAGE 1 (TEST): {}", test.create_data_uri());
-                    println!("REFTEST   IMAGE 2 (REFERENCE): {}", reference.create_data_uri());
+                    println!("REFTEST   IMAGE 1 (TEST):\n{}", test.create_data_uri());
```
rlhunt: Adding a newline here breaks image comparisons using the gecko [reftest analyzer](https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml).

kvark: good point! fixed now

glennw:
Looks good, thanks!

@bors-servo r+

PR #1013, opened 2017-03-27
Fix alpha blend mode and use premultiplied alpha for compositing

rlhunt:
Cache render targets and the framebuffer are stored using premultiplied alpha and so we should composite assuming premultiplied alpha too. It also looks like Device::set_blend_mode_alpha is not using the correct glBlend mode so I added a fix for that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1013)


webrender/src/tiling.rs
```diff
@@ -1431,7 +1431,7 @@ impl PrimitiveBatch {
         });
         let key = AlphaBatchKey::new(AlphaBatchKind::Composite,
                                      AlphaBatchKeyFlags::empty(),
-                                     BlendMode::Alpha,
+                                     BlendMode::PremultipliedAlpha,
```
kvark: why is this using `PremultipliedAlpha`?

wrench/reftests/blend/reftest.list
```diff
@@ -8,3 +8,4 @@
 == isolated.yaml isolated-ref.yaml
 == isolated-2.yaml isolated-2-ref.yaml
 == isolated-with-filter.yaml isolated-ref.yaml
+== isolated-premultiplied.yaml blank.yaml
```
kvark: what's the reason for including `transparent-white.png` if we are not using it?

webrender/src/device.rs
```diff
@@ -2049,9 +2049,8 @@ impl Device {
     }
 
     pub fn set_blend_mode_alpha(&self) {
-        //self.gl.blend_func(gl::SRC_ALPHA, gl::ONE_MINUS_SRC_ALPHA);
         self.gl.blend_func_separate(gl::SRC_ALPHA, gl::ONE_MINUS_SRC_ALPHA,
-                                     gl::ONE, gl::ONE);
+                                    gl::ONE, gl::ONE_MINUS_SRC_ALPHA);
```
kvark: Will it not break the existing blended logic? Perhaps, we should have separate functions for the blend mode alpha and pre-multiplied alpha.

webrender/src/tiling.rs
rlhunt: You're right, we should be using BlendMode::Alpha for the results of the MixBlendMode shader. I misunderstood how Gecko was handling this case.

wrench/reftests/blend/reftest.list
rlhunt: `transparent-white.png` is referenced in a test as a src image.

rlhunt:
There are a three kind of different issues here.

1. The blend mode used by most primitives for non-premultiplied alpha combines alpha with (1, 1) when the correct factors should be (1, 1-srcAlpha). Changing this should not break any blending logic to my knowledge. It is significant if a semitransparent primitive is drawn on a semitransparent background which can happen with mix-blend-mode. This was the change made in `device.rs`.

2. When we composite an isolated group (from mix-blend-mode) we use HardwareComposite::Alpha. But the result we are compositing is premultiplied so this is incorrect. This was the change made to HardwareComposite in `frame_builder.rs`.

3. The mix-blend-mode functions assume straight alpha, but the source and backdrop that we sample are premultiplied (except for the initial color for some render targets which is rgba(1, 1, 1, 0), but I don't think that is an issue). I just realized this in a second pass through. Gecko has the same problem and does a conversion in the shader. I've updated the PR with this change.

@kvark does this make sense?

Also, I wasn't very organized when I submitted this PR, so 1 is kind of different from 2, 3. So this could be split up into different issues. There should also be more tests and examples for this, which I will look into tomorrow morning.

kvark:
Sounds reasonable, thanks @rlhunt !
@bors-servo r+

PR #1015, opened 2017-03-28
Add support for clip in/out.

### LINKED ISSUE #189: box-shadow + border-radius doesn't work

glennw:
This is not a super-efficient implementation, but it adds the
functionality required for box shadows that have border radii.

Once this is in place, we can do some follow up optimization passes.

Fixes #189.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1015)

### DIFF
--- a/webrender/res/cs_clip_rectangle.vs.glsl
+++ b/webrender/res/cs_clip_rectangle.vs.glsl
@@ -5,7 +5,7 @@
 
 struct ClipRect {
     vec4 rect;
-    vec4 dummy;
+    vec4 mode;
 };
 
 ClipRect fetch_clip_rect(int index) {
@@ -14,8 +14,7 @@ ClipRect fetch_clip_rect(int index) {
     ivec2 uv = get_fetch_uv_2(index);
 
     rect.rect = texelFetchOffset(sData32, uv, 0, ivec2(0, 0));
-    //rect.dummy = texelFetchOffset(sData32, uv, 0, ivec2(1, 0));
-    rect.dummy = vec4(0.0, 0.0, 0.0, 0.0);
+    rect.mode = texelFetchOffset(sData32, uv, 0, ivec2(1, 0));
 
     return rect;
 }
@@ -70,6 +69,7 @@ void main(void) {
     vLocalRect = vi.clipped_local_rect;
     vPos = vi.local_pos;
 
+    vClipMode = clip.rect.mode.x;
     vClipRect = vec4(local_rect.xy, local_rect.xy + local_rect.zw);
     vClipRadius = vec4(clip.top_left.outer_inner_radius.x,
                        clip.top_right.outer_inner_radius.x,
--- a/webrender/src/mask_cache.rs
+++ b/webrender/src/mask_cache.rs
@@ -20,14 +20,21 @@ pub enum ClipMode {
     ClipOut,        // Pixels outside the region are visible.
 }
 
+#[repr(C)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum RegionMode {
+    IncludeRect,
+    ExcludeRect,
+}
+
 #[derive(Clone, Debug)]
 pub enum ClipSource {
     Complex(LayerRect, f32, ClipMode),
-    // The bool here specifies whether to consider the rect
+    // The RegionMode here specifies whether to consider the rect
     // from the clip region as part of the mask. This is true
     // for clip/scroll nodes, but false for primitives, where
     // the clip rect is handled in local space.
-    Region(ClipRegion, bool),
+    Region(ClipRegion, RegionMode),
 }
 
 impl ClipSource {
@@ -80,8 +87,6 @@ impl Geometry {
 /// more clever with some proper region handling.
 #[derive(Clone, Debug, PartialEq)]
 pub enum MaskBounds {
-    /// We haven't calculated the bounds yet.
-    Unknown,
     /// We know both the outer and inner rect. This is the
     /// fast path for, e.g. a simple rounded rect.
     OuterInner(Geometry, Geometry),
@@ -93,18 +98,12 @@ pub enum MaskBounds {
     None,
 }
 
-impl MaskBounds {
-    fn new() -> MaskBounds {
-        MaskBounds::Unknown
-    }
-}
-
 #[derive(Clone, Debug)]
 pub struct MaskCacheInfo {
     pub clip_range: ClipAddressRange,
     pub effective_clip_count: usize,
     pub image: Option<(ImageMask, GpuStoreAddress)>,
-    pub bounds: MaskBounds,
+    pub bounds: Option<MaskBounds>,
     pub is_aligned: bool,
 }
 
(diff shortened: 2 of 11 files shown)


glennw:
r? @kvark 

cc @changm 

I've checked against Mason's test case, and run the CSS/WPT tests successfully.

webrender/src/mask_cache.rs
```diff
@@ -8,158 +8,273 @@ use prim_store::{CLIP_DATA_GPU_SIZE, MASK_DATA_GPU_SIZE};
...
+/// Depending on the complexity of the clip, we may either
+/// know the outer and/or inner rect, or neither or these.
+/// In the case of a clip-out, we currently set the mask
+/// bounds to be unknown. This is conservative, but ensures
+/// correctness. In the future we can make this a lot
+/// more clever with some proper region handling.
+#[derive(Clone, Debug, PartialEq)]
+pub enum MaskBounds {
+    /// We haven't calculated the bounds yet.
+    Unknown,
```
kvark: I think this should be idiomatically used as `Option<MaskBounds>` (instead of `MaskBounds::Unknown`)

webrender/src/mask_cache.rs
```diff
@@ -8,158 +8,273 @@ use prim_store::{CLIP_DATA_GPU_SIZE, MASK_DATA_GPU_SIZE};
...
+            self.effective_clip_count = 0;
+            self.is_aligned = is_aligned;
+
+            for source in sources {
+                match source {
+                    &ClipSource::Complex(rect, radius, mode) => {
+                        // Once we encounter a clip-out, we just assume the worst
+                        // case clip mask size, for now.
+                        if mode == ClipMode::ClipOut {
+                            clip_out = true;
```
kvark: nit: it's not actually `clip_out` for the whole thing as it is `has_clip_out`

webrender/src/mask_cache.rs
```diff
@@ -8,158 +8,273 @@ use prim_store::{CLIP_DATA_GPU_SIZE, MASK_DATA_GPU_SIZE};
...
+
+            // Work out the type of mask geometry we have, based on the
+            // list of clip sources above.
+            if clip_out {
+                // For clip-out, the mask rect is not known.
+                self.bounds = MaskBounds::None;
+            } else {
+                // local inner is only valid if there's a single clip (for now).
+                // This can be improved in the future, with some proper
+                // rectangle region handling.
```
kvark: nit: make it a TODO item

webrender/src/mask_cache.rs
```diff
@@ -8,158 +8,273 @@ use prim_store::{CLIP_DATA_GPU_SIZE, MASK_DATA_GPU_SIZE};
...
+        // geometry.
+        match self.bounds {
+            MaskBounds::Unknown => unreachable!(),
+            MaskBounds::None => {}
+            MaskBounds::Outer(ref mut outer) => {
+                outer.update(transform, device_pixel_ratio);
+            }
+            MaskBounds::OuterInner(ref mut outer, ref mut inner) => {
+                outer.update(transform, device_pixel_ratio);
+                inner.update(transform, device_pixel_ratio);
```
kvark: I don't think this is correct. The `update` here would compute the bounding box of the inner area, which is useless to us (and confusing). Not sure if we actually use it anywhere, but it doesn't appear that `inner` fits the whole `Geometry` struct concept.

webrender/src/prim_store.rs
```diff
@@ -109,7 +109,11 @@ pub enum PrimitiveCacheKey {
 #[derive(Debug)]
 pub struct PrimitiveMetadata {
     pub is_opaque: bool,
-    pub clip_source: Box<ClipSource>,
+    // TODO(gw): It's only a very small subset of primitives
+    // that end up with > 1 clip region. Consider an optimization
+    // for this case - perhaps an enum that either holds one
+    // clip or a Vec of clips?
```
kvark: I don't think the optimization is necessary, given that we are using `Box<ClipSource>` already.

webrender/src/prim_store.rs
```diff
@@ -994,21 +1002,25 @@ impl PrimitiveStore {
...
+                let (rect, is_complex) = match source {
+                    ClipSource::Complex(rect, radius, _) => (rect, radius > 0.0),
+                    ClipSource::Region(ref region, _) => (region.main, region.is_complex()),
+                };
+                self.gpu_geometry.get_mut(GpuStoreAddress(index.0 as i32))
+                    .local_clip_rect = rect;
+                if is_complex {
+                    metadata.clip_cache_info = None; //CLIP TODO: re-use the existing GPU allocation
+                }
+                metadata.clips = vec![source];
```
kvark: nit: could move it up to `metadata.clips = match source {...};`

webrender/src/render_task.rs
```diff
@@ -192,19 +192,46 @@ impl RenderTask {
...
+
+        // If we encounter a clip with unknown bounds, we'll just use
+        // the original rect. This is overly conservative, but can
+        // be optimized later.
+        let mut result = Some(actual_rect);
+        for &(_, ref clip) in clips {
+            match clip.bounds {
+                MaskBounds::Unknown => unreachable!(),
+                MaskBounds::OuterInner(ref outer, _) |
+                MaskBounds::Outer(ref outer) => {
```
kvark: oh man, these match arms are awesome 😍

webrender/src/render_task.rs
```diff
@@ -192,19 +192,46 @@ impl RenderTask {
         // We scan through the clip stack and detect if our actual rectangle
         // is in the intersection of all of all the outer bounds,
         // and if it's completely inside the intersection of all of the inner bounds.
-        let result = clips.iter()
-                          .fold(Some(actual_rect), |current, clip| {
-            current.and_then(|rect| rect.intersection(&clip.1.outer_rect))
-        });
+
+        // If we encounter a clip with unknown bounds, we'll just use
+        // the original rect. This is overly conservative, but can
```
kvark: agreed. Let's make it a real `//TODO` item

webrender/src/mask_cache.rs
```diff
@@ -8,158 +8,273 @@ use prim_store::{CLIP_DATA_GPU_SIZE, MASK_DATA_GPU_SIZE};
...
 pub enum ClipSource {
-    NoClip,
-    Complex(LayerRect, f32),
-    Region(ClipRegion),
+    Complex(LayerRect, f32, ClipMode),
+    // The bool here specifies whether to consider the rect
+    // from the clip region as part of the mask. This is true
+    // for clip/scroll nodes, but false for primitives, where
+    // the clip rect is handled in local space.
+    Region(ClipRegion, bool),
```
kvark: it would be great to turn the `bool` into an `enum` type, similar to `ClipMode`

glennw:
@kvark Thanks for the review - pushed a commit which fixes those issues. The only thing I left unchanged is the inner rect calculating the bounding rect - we use it for that special / hacky case of only drawing the corners of simple clip masks. We can probably get rid of that in the future though.

kvark:
Looks great, thanks!
@bors-servo r+

webrender/res/cs_clip_rectangle.vs.glsl
```diff
@@ -14,8 +14,7 @@ ClipRect fetch_clip_rect(int index) {
     ivec2 uv = get_fetch_uv_2(index);
 
     rect.rect = texelFetchOffset(sData32, uv, 0, ivec2(0, 0));
-    //rect.dummy = texelFetchOffset(sData32, uv, 0, ivec2(1, 0));
```
Hywan: Why keeping this comment?

Issue #1017, opened 2017-03-28
Improve BlobImageRenderer serialization


nical:
We have a naive BlobImageRenderer implementation in Gecko in which _everything_ is serialized (including images, font data, paths, etc.) each time a blob image is created or updated. This is far from ideal.
I would like to extend the BlobImageRenderer interface so that it can get access to some of the data stored in the resource cache, in particular font data and images.

As part of that I would also like to add path objects to WebRender's API which would be stored in the resource cache (but wouldn't have a display item that can use them yet until someone has a crack at implementing SVG features in WebRender) to avoid re-serializing paths every time a blob image is updated. The API for paths would be similar to images (retained objects with add_path/update_path/delete_path methods on the RenderApi).

Thoughts?

kvark:
Isn't this a dupe of #402?

nical:
The overlap is only the addition of retained path objects. This issue focuses on making BlobImageRenderer not prohibitively expensive to use (I expect most of the wins to be in reusing the images and font data) while #402 pursues the more long term goal of supporting SVG primitives in WebRender proper.

glennw:
@nical Is this still relevant?

nical:
It'd still be nice to have path objects in the resource cache but I don't think it's critical. We can revisit whenever the time's right.

PR #1018, opened 2017-03-28
Expose images to the BlobImageRenderer.

nical:
See issue #1017. The idea would be to do the same for fonts and paths.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1018)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -25,7 +25,7 @@ use webrender_traits::{Epoch, FontKey, GlyphKey, ImageKey, ImageFormat, ImageRen
 use webrender_traits::{FontRenderMode, ImageData, GlyphDimensions, WebGLContextId};
 use webrender_traits::{DevicePoint, DeviceIntSize, DeviceUintRect, ImageDescriptor, ColorF};
 use webrender_traits::{GlyphOptions, GlyphInstance, TileOffset, TileSize};
-use webrender_traits::{BlobImageRenderer, BlobImageDescriptor, BlobImageError};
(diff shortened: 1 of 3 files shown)


glennw:
This seems like a minimal change that doesn't affect other areas anyway. Just need to fix the compile errors on CI, then r=me.

glennw:
@nical Needs another rebase.

nical:
Can we land this? I have work in progress that I need to rebase on top of it.

kvark:
@nical sure thing, but it's not mergable atm.

nical:
Woops I missed that. Rebased.

webrender/src/resource_cache.rs
```diff
@@ -116,6 +116,40 @@ struct ImageResource {
     dirty_rect: Option<DeviceUintRect>
 }
 
+struct ImageTemplates {
```
kvark: any reason for this to be a separate struct?

nical: The reason is the borrow checker. Without the separate struct the resource cache ends up being "entirely" borrowed when passed to the blob renderer which can't happen because we are already borrowing it partially (the blob renderer itself).

kvark: I don't understand. How does this change from the borrowck point of view if you just do `type ImageTemplates = HashMap<...>`?

kvark: as per IRC discussion, having a separate type allows us to implement `ImageStore`, which is needed to abstract away from `BlobImageRenderer`

PR #1022, opened 2017-03-29
Have ScrollLayerId use u64

### LINKED ISSUE #489: Use a u64 for the scrollid instead of usize
Right now webrender uses a usize for the scroll id (the thing inside `ScrollLayerInfo::Scrollable`) I would prefer it use a u64 as that would match what Gecko does, and make it easier to map from the C++ value directly to the rust value. It also seems kind of odd to have the size of this field be platform-dependent, but I guess that doesn't really matter too much.

/cc @mrobinson

mrobinson:
This makes the size of the id platform independent and matches what
Gecko uses.

Fixes #489.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1022)


mrobinson:
@glennw r?

kvark:
looks clean
@bors-servo r+

Issue #727, opened 2017-01-17
Update wrench code to use assert!() instead of if ... { panic!(..) }


glennw:

hgallagher1993:
Ya I could pick this up. . .does it need a custom error message or just `assert!(nums.len() != 16)`?

mrmiywj:
I can pick this up.

igilham:
Looks like this can be closed.

kvark:
@igilham right, thanks for triaging!
Fixed by #1023

PR #1023, opened 2017-03-29
use assert instead of if-panic

mrmiywj:
#727

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1023)

### DIFF
--- a/wrench/src/yaml_frame_reader.rs
+++ b/wrench/src/yaml_frame_reader.rs
@@ -20,9 +20,8 @@ fn broadcast<T: Clone>(base_vals: &[T], num_items: usize) -> Vec<T> {
         return base_vals.to_vec();
     }
 
-    if num_items % base_vals.len() != 0 {
-        panic!("Cannot broadcast {} elements into {}", base_vals.len(), num_items);
-    }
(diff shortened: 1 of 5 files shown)


wrench/src/yaml_frame_reader.rs
```diff
@@ -491,9 +487,7 @@ impl YamlFrameReader {
                                      .expect("Text items with glyphs require bounds [for now]");
             (glyphs, rect)
         } else {
-            if native_key.is_none() {
-                panic!("Can't layout simple ascii text with raw font [for now]");
-            }
+            assert!(!native_key.is_none(), "Can't layout simple ascii text with raw font [for now]");
```
kvark: nit: could use `is_some()`

mrmiywj:
PTAL @kvark

kvark:
@mrmiywj please use `assert_eq!`

glennw:
@mrmiywj Could you squash those commits? Then this should be good to go :)

mrmiywj:
@glennw PTAL

kvark:
Thanks, looks great!
@bors-servo r+

PR #1026, opened 2017-03-29
Add a scaleY property to RadialGradient

rlhunt:
This will allow RadialGradient's to represent gradient with ellipse ending shapes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1026)

### DIFF
--- a/webrender/res/ps_radial_gradient.vs.glsl
+++ b/webrender/res/ps_radial_gradient.vs.glsl
@@ -23,12 +23,19 @@ void main(void) {
     // and not snap here?
     vStartCenter = floor(0.5 + gradient.start_end_center.xy * uDevicePixelRatio) / uDevicePixelRatio;
     vEndCenter = floor(0.5 + gradient.start_end_center.zw * uDevicePixelRatio) / uDevicePixelRatio;
-    vStartRadius = gradient.start_end_radius_extend_mode.x;
-    vEndRadius = gradient.start_end_radius_extend_mode.y;
+    vStartRadius = gradient.start_end_radius_ratio_xy_extend_mode.x;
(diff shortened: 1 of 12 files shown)


rlhunt:
r? @glennw @kvark 

I believe this needs a version bump because it changes the interface in `webrender_traits`. This is a part of a larger [branch](https://github.com/rlhunt/webrender/commits/gradients-2) on gradients with a bunch of reftests added at the end so there aren't any here.

rlhunt:
Oh, I should note that the whole branch is ready for review. I'm just splitting it up to make it easier to review.

webrender/res/ps_radial_gradient.vs.glsl
```diff
@@ -23,12 +23,19 @@ void main(void) {
     // and not snap here?
     vStartCenter = floor(0.5 + gradient.start_end_center.xy * uDevicePixelRatio) / uDevicePixelRatio;
     vEndCenter = floor(0.5 + gradient.start_end_center.zw * uDevicePixelRatio) / uDevicePixelRatio;
-    vStartRadius = gradient.start_end_radius_extend_mode.x;
-    vEndRadius = gradient.start_end_radius_extend_mode.y;
+    vStartRadius = gradient.start_end_radius_ratio_xy_extend_mode.x;
```
kvark: don't you need to squeeze the start/end radius as well?

rlhunt: The start and end radius are given in terms of x and we're only scaling y so they're unaffected. I could rename the field to `start_end_radius_x_ratio_xy_extend_mode` or add a comment if that would make it more clear.

PR #1028, opened 2017-03-29
Update bincode and ipc-channel

cbrewster:
This was causing a compilation issue in one of my projects. When cargo gets webrender it doesn't have any .lock file, so it picks up the newer version of bincode (which appears to have broken semver rules) causing a compilation failure. The workspace cargo.lock kept travis from discovering this sooner.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1028)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = "0.2.2"
 dependencies = [
  "app_units 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "bincode 1.0.0-alpha2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


cbrewster:
it appears `ipc-channel` has the same issue :(

glennw:
@cbrewster Should we keep this open for now?

cbrewster:
I can open a PR on ipc-channel to do the update there first

cbrewster:
Should versions on webrender and wrench be bumped with these changes?

staktrace:
I think this PR should have updated the Cargo.toml file as well to point to the new version, since now webrender doesn't build with the "old" bincode (1.0.0-alpha2). webrender built into gecko uses alpha2 and our update process won't upgrade to alpha6 unless it's explicitly specified. I'll make a new PR for it.

PR #1032, opened 2017-03-29
Include build_scene time in render_backend time

jrmuizel:
Not sure if this is what we want but we should probably be accounting for this time somehow.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1032)


glennw:
Yep, it's not ideal because the build_scene doesn't (necessarily) result in a frame publish, which could make the results of this counter seem quite unstable. Ideally we should track them separately. But I agree that it's better to account for this somewhere now, and improve that later.

@bors-servo r+

PR #1035, opened 2017-03-30
Add support to create animated properties with binding key

howareyou322:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1035)


howareyou322:
@glennw I added these functions to construct animated properties with binding key.

Issue #1027, opened 2017-03-29
Assert in `max_texture_size`


kvark:
See logs:
https://treeherder.mozilla.org/logviewer.html#?job_id=87300231&repo=try&lineNumber=3307
Related PR #1025 didn't fix it. Need to investigate.

cc @nical @staktrace

staktrace:
@nical It was definitely something in #1004 that caused this, I bisected the WR changes to isolate the regression. You can find a full try push showing the failures at https://treeherder.mozilla.org/#/jobs?repo=try&revision=07cca001d72d6fc9d90618016124203c1d4f5317

nical:
I am pretty sure this is happening because the image is an external buffer and I assumed we did not support tiling those (yet), so I disabled automatically tiling this type of images in that PR. Thankfully it's trivial to re-enable. I still think we don't properly tile external image buffers (as in the interface when later uploading the external buffer does not take tiling into account), we'll have to fix that as a followup.

glennw:
This has merged.

staktrace:
Fix verified in try push https://treeherder.mozilla.org/#/jobs?repo=try&revision=b96726c4cf97169672cc938067d1f0c97a09e40c

Thanks!

PR #1036, opened 2017-03-30
Tile external buffer images when they exceed the maximum texture size.

nical:
Fixes issue #1027, and addresses a review comment from PR #1004 about explicitly enumerating all variants instead of using ```_ => { ... }``` in this match expression.

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1036)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -259,10 +259,15 @@ impl ResourceCache {
 
     fn should_tile(&self, descriptor: &ImageDescriptor, data: &ImageData) -> bool {
         let limit = self.max_texture_size();
+        let size_check = descriptor.width > limit || descriptor.height > limit;
         return match data {
-            // Tiled external images are not implemented.
-            &ImageData::External(_) => false,
-            _ => { descriptor.width > limit || descriptor.height > limit }
+            &ImageData::Raw(_) => { size_check }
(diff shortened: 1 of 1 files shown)


kvark:
Much better, thanks!
@bors-servo r+

nical:
Rebased. It doesn't read as nicely anymore due to the verbosity of ExternalImageData. I am tempted to change ImageData::External to be ```External(ExternalImageId, ExternalImageType)``` to reduce the noise.

webrender/src/resource_cache.rs
```diff
@@ -259,10 +259,16 @@ impl ResourceCache {
 
     fn should_tile(&self, descriptor: &ImageDescriptor, data: &ImageData) -> bool {
         let limit = self.max_texture_size();
+        let size_check = descriptor.width > limit || descriptor.height > limit;
         return match data {
-            // Tiled external images are not implemented.
+            &ImageData::Raw(_) => { size_check }
+            &ImageData::Blob(_) => { size_check }
+            &ImageData::External(ExternalImageData { image_type: ExternalImageType::ExternalBuffer, .. }) => {
+                size_check
```
kvark: nit could do `data.image_type == ExternalImageType::ExternalBuffer && size_check` instead of pattern matching

PR #1037, opened 2017-03-30
Favor the function over the method syntax when cloning a reference counted pointer.

nical:
```data.clone()``` reads like performance hazard, due to the vague definition of Clone, which can be either very expensive if data is an uniquely owned large resource or very cheap if it is a reference counted pointer to such a resource. This has already confused people reading the code and even code reviews several times in webrender, so I would like to work around this unfortunate ambiguity.

I first tried to address this at the [standard library level](https://github.com/rust-lang/rfcs/pull/1954) but the decision of the rust team appears to be that we should use the function syntax instead of the method syntax to address the problem.

See also the [clippy issue](https://github.com/Manishearth/rust-clippy/issues/1645) about providing a lint for this.

I decided to take a systematic approach and use the function call syntax for all of the outstanding reference counted pointer clones that I came across (I would not be surprised that I missed some of them). Rather than try to separate the obvious from the ambiguous ones. I would rather make this a default policy than have to debate about where the line is in each code review.

The three commits are independent and can be reviewed separately (although they are small and simple enough to look at as one diff if you prefer).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1037)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -406,7 +406,7 @@ impl ResourceCache {
                 if !same_epoch && self.blob_image_requests.insert(request) {
                     renderer.request_blob_image(
                         key,
-                        data.clone(),
+                        Arc::clone(&data),
                         &BlobImageDescriptor {
                             width: template.descriptor.width,
                             height: template.descriptor.height,
@@ -846,7 +846,7 @@ fn spawn_glyph_cache_thread(workers: Arc<Mutex<ThreadPool>>) -> (Sender<GlyphCac
 
         let barrier = Arc::new(Barrier::new(worker_count));
         for i in 0..worker_count {
-            let barrier = barrier.clone();
+            let barrier = Arc::clone(&barrier);
             workers.lock().unwrap().execute(move || {
                 register_thread_with_profiler(format!("Glyph Worker {}", i));
                 barrier.wait();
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: webrender/src/resource_cache.rs [hunk 1]
```diff
@@ -709,7 +709,7 @@ impl ResourceCache {
                         is_opaque: image_descriptor.is_opaque,
                     }
                 } else {
-                    image_template.descriptor.clone()
+                    image_template.descriptor
```
nox: This change is not related to what the PR describes, please remove of put in its own commit.

kvark: @nox I think you are overly sensitive here. It doesn't appear to me that the change bears any danger, and it's a bit related to the `clone()` refactor.

kvark:
@nical I see where you are coming from, and the change looks clean.
However, I'm not yet convinced that this is the way to go for all Rust libraries. The RFC discussion basically concluded that if you want it, you are free to use it, and the clippy issue discussion is empty so far (I know, the issue is recent). I'd like to get some opinion from @glennw  and Rust Core team.

glennw:
I really like this idea.

nical:
I did a last rebase before I disappear for 2 weeks. Don't hesitate to re-submit this in my place if a consensus is approached while I'm away.
I really think we would benefit from favoring this syntax. The rust core team decided stand on the side of letting people use the function syntax if they want to. Whether we want to hand pick places where we use it or apply it consistently is really up to us. Clippy support would be a plus but doesn't affect whether this style is to be considered good or bad for us.

kvark:
I'm not going to block this. @glennw please proceed if you are happy about it.

glennw:
OK, let's merge it. We can always change back if we decide it's not worthwhile. Thanks!

PR #1039, opened 2017-03-30
Don't panic when dealing with empty masks

mrobinson:
An empty mask shouldn't cause a panic. Instead we should use it as an
opportunity to optimize away the contents of that ClipScrollGroup. In
the future, perhaps this optimization can be extended to other kind of
empty ClipScrollGroups.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1039)


mrobinson:
@glennw or @kvark r?

kvark:
LGTM, thanks!
@bors-servo r+

PR #1041, opened 2017-03-30
Avoid unnecessary copies of Payload data when not using ipc channel

jrmuizel:
This improves the performance of
http://people.mozilla.com/~jmuizelaar/implementation-tests/dl-test.yaml
in wrench from 17fps to 19.5fps on my machine

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1041)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use byteorder::{LittleEndian, ReadBytesExt};
 use frame::Frame;
 use frame_builder::FrameBuilderConfig;
 use internal_types::{FontTemplate, SourceTexture, ResultMsg, RendererFrame};
@@ -11,7 +10,6 @@ use record::ApiRecordingReceiver;
 use resource_cache::ResourceCache;
 use scene::Scene;
 use std::collections::HashMap;
-use std::io::{Cursor, Read};
 use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Sender;
 use texture_cache::TextureCache;
@@ -21,7 +19,7 @@ use webgl_types::{GLContextHandleWrapper, GLContextWrapper};
 use webrender_traits::{DeviceIntPoint, DeviceUintPoint, DeviceUintRect, DeviceUintSize, LayerPoint};
 use webrender_traits::{ApiMsg, AuxiliaryLists, BuiltDisplayList, IdNamespace, ImageData};
 use webrender_traits::{PipelineId, RenderNotifier, RenderDispatcher, WebGLCommand, WebGLContextId};
-use webrender_traits::channel::{PayloadHelperMethods, PayloadReceiver, PayloadSender, MsgReceiver};
+use webrender_traits::channel::{PayloadSenderHelperMethods, PayloadReceiverHelperMethods, PayloadReceiver, PayloadSender, MsgReceiver};
 use webrender_traits::{BlobImageRenderer, VRCompositorCommand, VRCompositorHandler};
 #[cfg(feature = "webgl")]
 use offscreen_gl_context::GLContextDispatcher;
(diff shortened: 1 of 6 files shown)


jrmuizel:
This is still a bit messy, but I wanted to get feedback early.

jrmuizel:
This changes the serialization of Payload to include the length of the two parts. I'll bump the binary recording version.

webrender_traits/src/channel.rs
```diff
@@ -2,10 +2,52 @@
...
+#[derive(Clone)]
+pub struct Payload {
+    pub epoch: Epoch,
+    pub display_list_data: Vec<u8>,
+    pub auxiliary_lists_data: Vec<u8>
+}
+
+impl Payload {
+    pub fn to_data(&self) -> Vec<u8> {
+        let mut data = vec![];
```
kvark: we could use `with_capacity`

webrender_traits/src/channel.rs
```diff
@@ -2,10 +2,52 @@
...
+
+#[derive(Clone)]
+pub struct Payload {
+    pub epoch: Epoch,
+    pub display_list_data: Vec<u8>,
+    pub auxiliary_lists_data: Vec<u8>
+}
+
+impl Payload {
+    pub fn to_data(&self) -> Vec<u8> {
```
kvark: nit: consider using `From`/`Into` traits. Not really helping anything, just seems to be more idiomatic

webrender_traits/src/channel_ipc.rs
```diff
@@ -19,9 +21,16 @@ pub type PayloadSender = IpcBytesSender;
 
 pub type PayloadReceiver = IpcBytesReceiver;
 
-impl PayloadHelperMethods for PayloadSender {
-    fn send_vec(&self, data: Vec<u8>) -> Result<(), Error> {
-        self.send(&data)
+impl PayloadSenderHelperMethods for PayloadSender {
+    fn send(&self, data: Payload) -> Result<(), Error> {
+        self.send(&data.to_data())
```
kvark: is there any benefit of heap-allocating the vector, serializing the data to it, and then sending, versus just sending the data right away during serialization?

webrender_traits/src/api.rs
```diff
@@ -305,20 +304,18 @@ impl RenderApi {
...
+                                             aux_desc,
                                              preserve_frame_state);
         self.api_sender.send(msg).unwrap();
 
-        let mut payload = vec![];
-        payload.write_u32::<LittleEndian>(epoch.0).unwrap();
-        payload.extend_from_slice(display_list.data());
-        payload.extend_from_slice(auxiliary_lists.data());
-        self.payload_sender.send_vec(payload).unwrap();
+        PayloadSenderHelperMethods::send(&self.payload_sender, Payload{epoch: epoch, display_list_data: dl_data, auxiliary_lists_data: aux_data}).unwrap();
```
kvark: what's the reason of not calling it like `self.payload_sender.send(...)`?

webrender/src/render_backend.rs
```diff
@@ -188,36 +186,26 @@ impl RenderBackend {
                             let mut leftover_auxiliary_data = vec![];
                             let mut auxiliary_data;
                             loop {
-                                auxiliary_data = self.payload_rx.recv().unwrap();
+                                auxiliary_data = PayloadReceiverHelperMethods::recv(&self.payload_rx).unwrap();
```
kvark: similarly, what's the reason of not doing `self.payload_rx.recv()`?

webrender_traits/src/channel_ipc.rs
jrmuizel: Yes. send(Vec) is still message based. So the receiver expects to get everything in recv call. In the long term we'd probably want to fix this to be more efficient, but it looks like ipc-channel has lower hanging efficiency problems to deal with first.

kvark:
Thanks!
@bors-servo r+

kvark:
Classy timout on OSX
@bors-servo retry

PR #1053, opened 2017-04-03
Rename set_root_display_list to set_display_list.

waywardmonkeys:
The naming of this was confusing as it wasn't just for root display
lists, but all display lists. (And then you set the one that is the
root via `set_root_pipeline_id`.)

This was a artifact of an older design according to @glennw.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1053)


glennw:
This looks good, but we should bump the version numbers of WR and WR_traits, since it's an API change.

waywardmonkeys:
@glennw Addressed and force-pushed.

glennw:
Looks like a CI failure - you might need to ```cargo update``` to update the lock files for the sample / wrench, I think.

waywardmonkeys:
Sorry about that. I fixed that and have repushed again.

glennw:
Thanks!

@bors-servo r+

PR #1054, opened 2017-04-03
Make non-empty case statements have break or return

### LINKED ISSUE #1044: glsl non-empty case statements must have break or return for ANGLE
ps_border.fs.glsl and ps_border.vs.glsl has non-empty case statements must have break or return, it causes failures of compiling the shaders. I faced the problem on gecko webrender on graphics branch on Windows. When I tested on windows, "discard;"  also seemed to request break/return

The following are actual error log.
>C:\fakepath(4586,1): error X3533: non-empty case statements must have break or return
>C:\fakepath(4589,1): error X3533: non-empty case statements must have break or return

sotaroikeda:
Fixes  #1044

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1054)

### DIFF
--- a/webrender/res/ps_border.fs.glsl
+++ b/webrender/res/ps_border.fs.glsl
@@ -348,6 +348,7 @@ void draw_solid_border(float distanceFromMixLine, vec2 localPos) {
     default:
       oFragColor = vHorizontalColor;
       discard_pixels_in_rounded_borders(localPos);
+      break;
   }
 }
 
@@ -427,6 +428,7 @@ void main(void) {
 #endif
 
     vec2 brightness_mod = vec2(0.7, 1.3);
+    bool needs_discard = false;
 
     // Note: we can't pass-through in the following cases,
     // because Angle doesn't support it and fails to compile the shaders.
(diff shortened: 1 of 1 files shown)


emilio:
Ugh, this looks pretty awful IMO, there's a lot of duplicated code. Are you sure just `break`ing under the `default` cases doesn't fix it?

glennw:
Perhaps we can at least put the duplicated code sections into functions, so that only the function call is duplicated?

sotaroikeda:
I created the pull request be referencing #749. But I misunderstood the error code, I tried to minimize the change, and duplicated code parts were not necessary.

sotaroikeda:
> Are you sure just breaking under the default cases doesn't fix it?

It seems to work, I am going to update the pull request.

webrender/res/ps_border.fs.glsl
```diff
@@ -458,10 +459,8 @@ void main(void) {
         case BORDER_STYLE_RIDGE:
           draw_mixed_border(distance_from_mix_line, distance_from_middle, local_pos, brightness_mod.xy);
           break;
-        case BORDER_STYLE_HIDDEN:
-          discard;
         default:
-          discard;
+          break;
```
kvark: removing `discard` here will change the behavior

kvark:
@bors-servo r-
sorry for interrupting - I'd like the `discard` changes to be clarified

glennw:
@kvark Ah, you're right - I misread that change, thinking that the discard was added for the hidden case.

sotaroikeda:
I removed discard since the following did not work on windows
>        case BORDER_STYLE_HIDDEN:
>          discard;

The following works on windows, but it did not work on linux
>        case BORDER_STYLE_HIDDEN:
>          discard;
>          break;

sotaroikeda:
> removing discard here will change the behavior

Hmm

kvark:
Try working around it by moving the discard to the end of an execution path (e.g. an "if" branch).

PR #1055, opened 2017-04-03
Fix the way that clip positioning is specified

mrobinson:
Make the way that clip positions are specified a bit more predictable.
Now the display item bounds are the "content bounds," which specify the
location and size of the content that goes inside this clip. The clip
coordinates are specified relative to this location, while the content
size is only used for scrolling.

Currently we only support clips rectangles that start at the origin of
the content, which is equivalent to the capabilities we had before.
Later we can add this functionality, if it is necessary.

Also, add a test for this as well as updating all reftest files to
reflect the new arguments.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1055)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -118,15 +118,18 @@ pub struct ClipScrollNode {
 impl ClipScrollNode {
     pub fn new(pipeline_id: PipelineId,
                parent_id: ScrollLayerId,
-               local_viewport_rect: &LayerRect,
-               content_size: LayerSize,
+               content_rect: &LayerRect,
+               clip_rect: &LayerRect,
                clip_info: ClipInfo)
                -> ClipScrollNode {
+        // FIXME(mrobinson): We don't yet handle clipping rectangles that don't start at the origin
+        // of the node.
+        let local_viewport_rect = LayerRect::new(content_rect.origin, clip_rect.size);
         ClipScrollNode {
             scrolling: ScrollingState::new(),
-            content_size: content_size,
-            local_viewport_rect: *local_viewport_rect,
-            local_clip_rect: *local_viewport_rect,
+            content_size: content_rect.size,
+            local_viewport_rect: local_viewport_rect,
+            local_clip_rect: local_viewport_rect,
             combined_local_viewport_rect: LayerRect::zero(),
             world_viewport_transform: LayerToWorldTransform::identity(),
             world_content_transform: LayerToWorldTransform::identity(),
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -354,14 +354,12 @@ impl Frame {
                         pipeline_id: PipelineId,
                         parent_id: ScrollLayerId,
                         item: &ClipDisplayItem,
-                        reference_frame_relative_offset: LayerPoint,
+                        content_rect: &LayerRect,
                         clip: &ClipRegion) {
-        let clip_rect = clip.main.translate(&reference_frame_relative_offset);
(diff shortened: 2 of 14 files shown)


mrobinson:
@glennw or @kvark r?

glennw:

---

*[webrender/src/clip_scroll_node.rs, line 125 at r1](https://reviewable.io:443/reviews/servo/webrender/1055#-KgqysyV10h6JghyDqRO:-KgqysyWYAIxfiJB0zEf:b29vah3) ([raw file](https://github.com/servo/webrender/blob/2774a20a0185a1c6347078413e145f8753317d03/webrender/src/clip_scroll_node.rs#L125)):*
> ```Rust
>                clip_info: ClipInfo)
>                -> ClipScrollNode {
>         // FIXME(mrobinson): We don't yet handle clipping rectangles that don't start at the origin
> ```

Should we assert / log a warning if this case occurs?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1055)*

glennw:
@mrobinson r=me once that comment above is addressed, if you think it's worthwhile.

glennw:
@mrobinson Actually, when I run the WPT tests I get the following failures:

```
  ▶ FAIL [expected PASS] /_mozilla/css/overflow_auto.html
  └   → /_mozilla/css/overflow_auto.html 5fc5daa41d2f7c5a6cadeb9512b4ee0bb01b0b85
/_mozilla/css/overflow_simple_b.html 7949e80732fa0f4afdbae3b1cad96ccd75bb04f6
Testing 5fc5daa41d2f7c5a6cadeb9512b4ee0bb01b0b85 == 7949e80732fa0f4afdbae3b1cad96ccd75bb04f6

  ▶ FAIL [expected PASS] /_mozilla/css/overflow_scroll.html
  └   → /_mozilla/css/overflow_scroll.html 5fc5daa41d2f7c5a6cadeb9512b4ee0bb01b0b85
/_mozilla/css/overflow_simple_b.html 7949e80732fa0f4afdbae3b1cad96ccd75bb04f6
Testing 5fc5daa41d2f7c5a6cadeb9512b4ee0bb01b0b85 == 7949e80732fa0f4afdbae3b1cad96ccd75bb04f6
```

It could just be that I made a mistake with the Servo API changes while testing this?

mrobinson:
@glennw Oh right. This patch breaks API, so I need to update Cargo.toml for the projects. I'll update the branch for that. I'll also post a Servo branch that should pass all the tests.

mrobinson:
I have updated the branch to increment the webrender and webrender_traits versions. I have also uploaded a Servo branch here: https://github.com/mrobinson/servo/tree/clipping-coordinates. The tests you mentioned are now passing, but some flexbox tests are failing, which seems unrelated to my PR.

staktrace:
This PR is causing gecko reftests to fail. See https://bugzilla.mozilla.org/show_bug.cgi?id=1353484#c8. Since this PR includes an API change, I had to update our webrender_bindings code to make it compile - it was a straightforward change but maybe that's the problem? My fixup is at https://hg.mozilla.org/try/rev/cdad4c05189e69c98161c72b8eb5762d7fdebf27

mrobinson:
@staktrace Chances are you will also need to change the coordinates of the clip itself. Now the clip should be relative to the content_rect of the scroll layer that you push. I'll try to take a look at Gecko today and come up with a potential fix.

staktrace:
Just to close the loop - you were right, I had made an error with my fixup and the coordinates I was passing as the scroll layer's content rect were incorrect. I have it working fine now. Thanks!

PR #1056, opened 2017-04-03
Bump bincode dependency.

staktrace:
Webrender doesn't actually compile with 1.0.0-alpha2 any more, so we
should update the Cargo.toml to set a new minimum bound on the bincode
version.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1056)


staktrace:
Note that the Cargo.lock file was already updated in #1028

PR #1058, opened 2017-04-03
Update servo-glutin

jrmuizel:
This fixes running on macOS by picking up
https://github.com/servo/glutin/pull/123

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1058)


glennw:
The Cargo.toml files should probably be updated too?

jrmuizel:
Why not.

glennw:
This should also update the sample Cargo.toml :)

jrmuizel:
Why not.

PR #1061, opened 2017-04-04
Support more cases of simple border fast path.

glennw:
This is the first step in refactoring how the border
shader and code path works.

Specifically, this change detects many more cases
where borders can be drawn by a series of simple
rectangles rather than the existing border shader. It now
correctly uses the fast path for any border which is a
combination of solid / none / hidden styles without
a border radius.

Follow up work will make use of the new border corner
and edge enum types to select different code paths
for specific border optimizations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1061)

### DIFF
--- a/webrender/src/border.rs
+++ b/webrender/src/border.rs
@@ -1,0 +1,202 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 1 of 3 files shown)


glennw:
CSS test time before and after this patch:

```
./mach test-css --release --processes=8  8407.78s user 1448.45s system 715% cpu 22:57.38 total
./mach test-css --release --processes=8  5800.28s user 1405.50s system 707% cpu 16:58.84 total
```

So this saves ~25% off the time to run the complete CSS tests.

PR #1066, opened 2017-04-05
Prefer &[..] over &Vec<..>.

waywardmonkeys:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1066)


glennw:
@bors-servo r+

Thanks!

PR #1067, opened 2017-04-05
Remove unused inner_rect from BorderPrimitiveCpu struct.

glennw:
This means the BorderPrimitiveCpu struct is currently empty.
I have not removed that, however, as it will be used in some of
the upcoming border refactoring work.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1067)


kvark:
Oh wow!
@bors-servo r+

PR #1068, opened 2017-04-05
webrender_traits: Fix or allow all clippy warnings.

waywardmonkeys:
We now silently allow (with this commit) clippy warnings
about comparing floating point values as well as for functions
with too many arguments.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1068)

### DIFF
--- a/webrender_traits/src/api.rs
+++ b/webrender_traits/src/api.rs
@@ -67,33 +67,33 @@ pub enum ApiMsg {
 
 impl fmt::Debug for ApiMsg {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            &ApiMsg::AddRawFont(..) => { write!(f, "ApiMsg::AddRawFont") }
-            &ApiMsg::AddNativeFont(..) => { write!(f, "ApiMsg::AddNativeFont") }
-            &ApiMsg::DeleteFont(..) => { write!(f, "ApiMsg::DeleteFont") }
-            &ApiMsg::GetGlyphDimensions(..) => { write!(f, "ApiMsg::GetGlyphDimensions") }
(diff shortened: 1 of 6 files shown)


webrender_traits/src/api.rs
```diff
@@ -67,33 +67,33 @@ pub enum ApiMsg {
...
-            &ApiMsg::GenerateFrame(..) => { write!(f, "ApiMsg::GenerateFrame") }
-            &ApiMsg::VRCompositorCommand(..) => { write!(f, "ApiMsg::VRCompositorCommand") }
-            &ApiMsg::ExternalEvent(..) => { write!(f, "ApiMsg::ExternalEvent") }
-            &ApiMsg::ShutDown => { write!(f, "ApiMsg::ShutDown") }
-            &ApiMsg::SetPageZoom(..) => { write!(f, "ApiMsg::SetPageZoom") }
-            &ApiMsg::SetPinchZoom(..) => { write!(f, "ApiMsg::SetPinchZoom") }
-            &ApiMsg::SetPan(..) => { write!(f, "ApiMsg::SetPan") }
-            &ApiMsg::SetWindowParameters(..) => { write!(f, "ApiMsg::SetWindowParameters") }
+        match *self {
+            ApiMsg::AddRawFont(..) => { write!(f, "ApiMsg::AddRawFont") }
```
emilio: this could be:

```
write!(f, match *self {
   // ...
})
```

While you're at it

waywardmonkeys: @emilio That fails to compile with:

```
error: format argument must be a string literal.
  --> webrender_traits/src/api.rs:70:19
   |
70 |         write!(f, match *self {
   |                   ^

```

waywardmonkeys: Changed it to be a `f.write_str` call and added an additional commit with that.

Issue #899, opened 2017-02-17
Deduplicate dependencies


nox:
There are three versions of `core-graphics-rs` referenced in `Cargo.lock`.

kvark:
Another offender:
> 1519: "checksum yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)" = "<none>"
 1520: "checksum yaml-rust 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)" = "e66366e18dc58b46801afbf2ca7661a9f59cc8c5962c29892b6039b4f86fa992"

waywardmonkeys:
I've submitted https://github.com/MSleepyPanda/rust-font-loader/pull/7 to help move this along.

PR #1069, opened 2017-04-05
Update font-loader.

waywardmonkeys:
This removes older versions of core-text, core-graphics and serde
from our transitive dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1069)


waywardmonkeys:
This is further progress towards the resolution of #899.

glennw:
I think the CI failure is a travis issue.

@bors-servo r+

PR #1070, opened 2017-04-05
Add a basic text reftest

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1070)


wrench/reftests/text/blank.yaml
```diff
@@ -0,0 +1,5 @@
+---
+root:
+  items:
+        - type: rect
+          bounds: [0, 0, 95, 88]
```
rlhunt: Does this need to have a blank rect?

jrmuizel: I believe we panic if items is empty. I didn't investigate what was happening.

rlhunt: Oh yes, that does happen. `reftest/gradients/blank.yaml` gets around that by not specifying a `items:`
```yaml
---
root:
```

PR #1072, opened 2017-04-05
Correctly set clip in for inset box shadows with border radii

changm:
Inset box shadows weren't working properly because we were performing an out clip. But inset box shadows draw inwards, so we need a clip in.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1072)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -882,12 +882,17 @@ impl FrameBuilder {
                     BoxShadowClipMode::Inset => 1.0,
                 };
 
-                // If we have a border radius, we'll need to apply
-                // a clip-out mask.
+                // Outset box shadows with border radius
+                // need a clip out of the center box.
+                let extra_clip_mode = match clip_mode {
+                    BoxShadowClipMode::Outset | BoxShadowClipMode::None => ClipMode::ClipOut,
+                    BoxShadowClipMode::Inset => ClipMode::Clip,
+                };
+
                 let extra_clip = if border_radius > 0.0 {
                     Some(ClipSource::Complex(*box_bounds,
                                              border_radius,
-                                             ClipMode::ClipOut))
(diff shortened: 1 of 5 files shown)


jrmuizel:
Can you add a reftest?

webrender/src/frame_builder.rs
```diff
@@ -882,12 +882,17 @@ impl FrameBuilder {
                     BoxShadowClipMode::Inset => 1.0,
                 };
 
-                // If we have a border radius, we'll need to apply
-                // a clip-out mask.
+                // Outset box shadows with border radius
+                // need a clip out of the center box.
+                let extra_clip_mode = match clip_mode {
```
glennw: I *think* ```extra_clip_mode``` can just be ```None``` rather than ```Some(ClipMode::Clip)``` in this case?

changm: Not really unless you want to make ClipSource::Complex take an Option instead?

glennw: What I mean is, instead of selecting a different clip mode, I think it would work (and be more GPU efficient) if you change the ```if border_radius > 0.0``` condition to ```if border_radius > 0.0 && clip_mode == BoxShadowClipMode::Inset``` ?

changm: I suspect you mean clip_mode != BoxShadowClipMode::Inset? This didn't work, the border-radius clip was ignored completely and we'd just get a square inset box shadow. I suspect because the clip_mode we're passing in when we create the display item doesn't include the proper rounded rect clip. We could alter the passed in clip region to translate it to a complex one since it's passed directly by users of the API, but that is confusing to API users.

glennw: Ah, you're right - and we only apply this is radius > 0.0, so it should be fine. Thanks!

PR #1073, opened 2017-04-06
Tidy up several parts of the batching code.

glennw:
This is prep work for some of the upcoming border refactoring.

None of the diffs in this file should result in any behaviour
changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1073)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -58,6 +58,14 @@ trait AlphaBatchHelpers {
                           src_task_index: RenderTaskIndex,
                           filter: LowLevelFilterOp,
                           z_sort_index: i32);
+    fn add_composite_to_batch(&self,
+                              batch: &mut PrimitiveBatch,
+                              stacking_context_index: StackingContextIndex,
+                              task_index: RenderTaskIndex,
+                              backdrop_task: RenderTaskIndex,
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

Sorry the patch looks quite large - the majority of it is just shuffling code around, so it should hopefully be easy to review.

glennw:
I've run against the WPT/CSS tests successfully.

webrender/src/tiling.rs
```diff
@@ -714,28 +730,17 @@ impl AlphaBatcher {
...
-                        let needs_clipping_flag = if needs_clipping {
-                            NEEDS_CLIPPING
-                        } else {
-                            AlphaBatchKeyFlags::empty()
-                        };
-                        let flags = match transform_kind {
-                            TransformedRectKind::AxisAligned => AXIS_ALIGNED | needs_clipping_flag,
-                            _ => needs_clipping_flag,
-                        };
+                        debug_assert!(!prim_metadata.needs_clipping());
```
kvark: I'd rather not tie up clipping, axis-alignment, and opaque-ness properties together just yet.

kvark:
The code is much prettier now!
@bors-servo r+

PR #1074, opened 2017-04-06
Interpret gradient start and end positions as relative to bounds

rlhunt:
Gradient start and end positions are currently relative to their
parent stacking context. This isn't necessary and slightly complicates
the representation of tiled gradients.

The angle gradient shader is modified so the offset calculation
is done in the fragment shader. This is needed for tiling gradients,
which will happen in a follow up. If this is an issue, the change can
be done later. The aligned gradient shader is relatively unchanged
as it won't be used for tiling.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1074)

### DIFF
--- a/webrender/res/ps_angle_gradient.fs.glsl
+++ b/webrender/res/ps_angle_gradient.fs.glsl
@@ -5,11 +5,14 @@
 uniform sampler2D sGradients;
 
 void main(void) {
+    // Normalized offset of this vertex within the gradient, before clamp/repeat.
+    float offset = dot(vPos - vStartPoint, vScaledDir);
+
     vec2 texture_size = vec2(textureSize(sGradients, 0));
 
     // Either saturate or modulo the offset depending on repeat mode, then scale to number of
     // gradient color entries (texture width / 2).
-    float x = mix(clamp(vOffset, 0.0, 1.0), fract(vOffset), vGradientRepeat) * 0.5 * texture_size.x;
+    float x = mix(clamp(offset, 0.0, 1.0), fract(offset), vGradientRepeat) * 0.5 * texture_size.x;
 
     x = 2.0 * floor(x) + 0.5 + fract(x);
 
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -577,23 +577,27 @@ impl FrameBuilder {
             }
             BorderDetails::Gradient(ref border) => {
                 for segment in create_segments(border.outset) {
+                    let segment_rel = segment.origin - rect.origin;
+
                     self.add_gradient(scroll_layer_id,
                                       segment,
                                       clip_region,
-                                      border.gradient.start_point,
-                                      border.gradient.end_point,
+                                      border.gradient.start_point - segment_rel,
+                                      border.gradient.end_point - segment_rel,
                                       border.gradient.stops,
                                       border.gradient.extend_mode);
                 }
             }
             BorderDetails::RadialGradient(ref border) => {
                 for segment in create_segments(border.outset) {
+                    let segment_rel = segment.origin - rect.origin;
+
                     self.add_radial_gradient(scroll_layer_id,
                                              segment,
                                              clip_region,
-                                             border.gradient.start_center,
+                                             border.gradient.start_center - segment_rel,
                                              border.gradient.start_radius,
-                                             border.gradient.end_center,
+                                             border.gradient.end_center - segment_rel,
                                              border.gradient.end_radius,
                                              border.gradient.ratio_xy,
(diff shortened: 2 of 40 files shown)


rlhunt:
r? @glennw @kvark 

I've ran the servo wpt tests with this change. With a patch to `display_list_builder` it seems to work. There is one test that needs very slight fuzzing for some reason.
```diff
diff --git a/components/layout/display_list_builder.rs b/components/layout/display_list_builder.rs
index 8c390487f5..091df7e9c8 100644
--- a/components/layout/display_list_builder.rs
+++ b/components/layout/display_list_builder.rs
@@ -952,8 +952,8 @@ impl FragmentDisplayListBuilding for Fragment {
             })
         }
 
-        let center = Point2D::new(absolute_bounds.origin.x + absolute_bounds.size.width / 2,
-                                  absolute_bounds.origin.y + absolute_bounds.size.height / 2);
+        let center = Point2D::new(absolute_bounds.size.width / 2,
+                                  absolute_bounds.size.height / 2);
```
On the gecko side, this should be an easy change.

kvark:
> The aligned gradient shader is relatively unchanged as it won't be used for tiling.

Why?

webrender/res/ps_angle_gradient.fs.glsl
```diff
@@ -5,11 +5,14 @@
 uniform sampler2D sGradients;
 
 void main(void) {
+    // Normalized offset of this vertex within the gradient, before clamp/repeat.
+    float offset = dot(vPos - vStartPoint, vDir) / dot(vDir, vDir);
```
kvark: Let's have `vDir` adjusted in the vertex shader before interpolation in order to avoid the computation here:
```glsl
vec2 dir = end_point - start_point;
vDir = dir / dot(dir, dir);
```

webrender/src/frame_builder.rs
```diff
@@ -577,23 +577,27 @@ impl FrameBuilder {
             }
             BorderDetails::Gradient(ref border) => {
                 for segment in create_segments(border.outset) {
+                    let segment_rel_rect = segment.origin - rect.origin;
```
kvark: nit: name should not have `_rect` at the end, since it's not a rectangle

wrench/reftests/gradient/norm-linear-degenerate-ref.yaml
```diff
@@ -1,11 +1,8 @@
 ---
 root:
   items:
-    - type: stacking-context
+    - type: gradient
```
kvark: nice to see the tests simplified!

kvark:
Side note:
I consider the move of the offset calculation to the pixel shader to be a good thing. When I was looking at the gradient imperfections, I found out that we can't reach the desired precision of color change if we interpolate the gradient from VS. This PR still does (interpolates `vPos`), but it will make the transition easier if we want it. The transition could be - taking the `gl_FragCoord.xy` and back-projecting it onto the primitive space in order to compute the gradient offset.

rlhunt:
> > The aligned gradient shader is relatively unchanged as it won't be used for tiling.
>
> Why?

The aligned gradient path works by generating a primitive for each contiguous stop pair, and the first approach I took for tiling was to do it all with one primitive in the shaders. I see benefits to generating multiple primitives, but I think that can be done in a follow up once the API is introduced and it works.

rlhunt:
> Side note:
I consider the move of the offset calculation to the pixel shader to be a good thing. When I was looking at the gradient imperfections, I found out that we can't reach the desired precision of color change if we interpolate the gradient from VS. This PR still does (interpolates vPos), but it will make the transition easier if we want it. The transition could be - taking the gl_FragCoord.xy and back-projecting it onto the primitive space in order to compute the gradient offset.

Just curious, what's the difference between interpolating vPos vs, using gl_FragCoord?

rlhunt:
I rebased to fix the merge conflict and addressed the review comments.

kvark:
> Just curious, what's the difference between interpolating vPos vs, using gl_FragCoord?

The key thing to understand is that the fragment coordinate is not interpolated, it's what the rasterizer uses as input to produce those barycentric coordinates for hardware interpolators. So if there is an option (given our rendering model for gradients) to produce the gradient coordinates directly from the `gl_FragCoord`, it will be more precise than interpolating anything from VS.

I did a few tests with a full-screen linear gradient that @squarewave provided (he raised the gradient imperfection issue in the first place, although we don't have it formally registered (yet)), and going for `gl_FragCoord` was the only solution I found (it wasn't a complete research, since I realised this has nothing to do with the ordered dithering).

kvark:
Thanks! :shipit: 
@bors-servo r+

PR #1075, opened 2017-04-06
Do not snap gradient start and end positions to device pixels

rlhunt:
I don't believe this code is necessary anymore. I was able to run the Servo CSS and WPT test suites succesfully without them.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1075)

### DIFF
--- a/webrender/res/ps_angle_gradient.vs.glsl
+++ b/webrender/res/ps_angle_gradient.vs.glsl
@@ -13,16 +13,11 @@ void main(void) {
                                  prim.layer,
                                  prim.task);
 
-    // Snap the start/end points to device pixel units.
-    // I'm not sure this is entirely correct, but the
-    // old render path does this, and it is needed to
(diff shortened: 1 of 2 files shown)


rlhunt:
r? @glennw @kvark 

If you know why this code was needed and it turns out it still is, it should be able to be moved to `PrimStore::prepare_prims_for_render`.

kvark:
The gradients are computed from the local position, and we are only snapping the final position, so it doesn't make sense to snap the gradient start/stop any more:
```glsl
    // Calculate the distance to snap the vertex by (snap top left corner).
    vec2 snap_delta = device_p0_pos.xy - floor(device_p0_pos.xy + 0.5);

    // Apply offsets for the render task to get correct screen location.
    vec2 final_pos = device_p0_pos.zw -
                     snap_delta -
                     task.screen_space_origin +
                     task.render_target_origin;
```

@bors-servo r+

PR #1076, opened 2017-04-06
Inner rectangle compute refactor

kvark:
This should be cleaner and a bit faster w.r.t the transparency optimization of #1005
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1076)


glennw:
Nice!

@bors-servo r+

PR #1077, opened 2017-04-06
Fix subpixel position rounding to be consistent

### LINKED ISSUE #1047: panic in subpixel_quantize_offset
webrender_traits::font::{{impl}}::subpixel_quantize_offset(self=0x000000013b1656bd, pos=-3) + 183 at font.rs:66
   63  	            0.25...0.5 => SubpixelOffset::Quarter,
   64  	            0.5...0.75 => SubpixelOffset::Half,
   65  	            0.75...1.0 => SubpixelOffset::ThreeQuarters,
-> 66  	            _ => panic!("Should only be given the fractional part"),

jrmuizel:
Previously we could end up with negative offsets if the input
to frac was negative. Using floor makes sure this works properly.

This change also elaborates the 0.125 constant.

Fixes #1047

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1077)

### DIFF
--- a/webrender_traits/src/font.rs
+++ b/webrender_traits/src/font.rs
@@ -47,6 +47,14 @@ pub enum FontRenderMode {
     Subpixel,
 }
 
+const FIXED16_SHIFT: i32 = 16;
+
+// This matches the behaviour of SkScalarToFixed
+fn f32_truncate_to_fixed16(x: f32) -> i32 {
+    let fixed1 = (1 << FIXED16_SHIFT) as f32;
+    (x * fixed1) as i32
+}
+
 impl FontRenderMode {
     // Skia quantizes subpixel offets into 1/4 increments.
     // Given the absolute position, return the quantized increment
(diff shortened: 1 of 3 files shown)


kvark:
Interestingly, there was a discussion on rust-gamedev recently about code like `pos - pos.floor()` sometimes returning `1.0`. There appear to be a few possible causes to this (I can leave it as an exercise for you ;)). The only robust method of getting the true unsigned fractional part is calling `fract()` and then adjusting for the sign of the original.

webrender_traits/src/font.rs
```diff
@@ -47,6 +47,15 @@ pub enum FontRenderMode {
     Subpixel,
 }
 
+const FIXED16_SHIFT: u32 = 16;
+
+// This matches the behaviour of SkScalarToFixed
+fn f32_truncate_to_fixed16(x: f32) -> u32
+{
```
kvark: nit: opening bracket on the new line

webrender_traits/src/font.rs
```diff
@@ -55,14 +64,19 @@ impl FontRenderMode {
             return SubpixelOffset::Zero;
         }
 
-        const SUBPIXEL_ROUNDING :f32 = 0.125; // Skia chosen value.
-        let fraction = (pos + SUBPIXEL_ROUNDING).fract();
+
+        const SUBPIXEL_BITS: u32 = 2;
+        const SUBPIXEL_FIXED16_MASK: u32 = 0b11 << (FIXED16_SHIFT - SUBPIXEL_BITS);
```
kvark: `0b11` is a magic constant?

webrender_traits/src/font.rs
```diff
@@ -47,6 +47,15 @@ pub enum FontRenderMode {
     Subpixel,
 }
 
+const FIXED16_SHIFT: u32 = 16;
+
+// This matches the behaviour of SkScalarToFixed
+fn f32_truncate_to_fixed16(x: f32) -> u32
+{
+    let fixed1 = (1 << FIXED16_SHIFT) as f32;
+    (x * fixed1) as u32
```
kvark: I'm a bit [worried](http://www.embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int/) about this cast

kvark:
Terrific, thanks!
@bors-servo r+

PR #1078, opened 2017-04-06
Add tiling for gradients

rlhunt:
This adds `tile_size` and `tile_spacing` fields to linear and radial gradient
display items to control tiling. A gradient is tiled starting at `bounds.origin`,
with tile size of `tile_size` and spacing between tiles of `tile_spacing`. Tiles
will fill and be clipped to `bounds`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1078)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,8 +21,8 @@ dependencies = [
  "serde_json 0.9.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-glutin 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.30.0",
- "webrender_traits 0.31.0",
+ "webrender 0.31.0",
+ "webrender_traits 0.32.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
@@ -861,7 +861,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "webrender"
-version = "0.30.0"
+version = "0.31.0"
 dependencies = [
  "angle 0.1.2 (git+https://github.com/servo/angle?branch=servo)",
  "app_units 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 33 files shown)


rlhunt:
r? @kvark @glennw 

I've successfully ran the Servo WPT, CSS tests with this change. On the gecko side this should simplify some code for us and let us create less display items.

kvark:
The test coverage is really impressive, thank you!
@bors-servo r+

kvark:
When testing #1087 on the latest, I tried hooking up to servo and made the following change:
```diff
-                builder.push_gradient(rect, clip, gradient);
+                builder.push_gradient(rect, clip, gradient, rect.size, webrender_traits::LayoutSize::zero());
```
And got 2 WPT failures:
  - `/_mozilla/css/border-image-linear-gradient.html`
  - `/_mozilla/css/linear_gradients_parsing_a.html`

Could this PR be causing them?
[wpt.log.zip](https://github.com/servo/webrender/files/906925/wpt.log.zip)

rlhunt:
The log looks like the failures I was seeing when testing #1074.

Can you give it a try with the following diff? It makes gradient start stop relative like the #1074 expects.

```diff
diff --git a/components/layout/display_list_builder.rs b/components/layout/display_list_builder.rs
index 8c390487f5..091df7e9c8 100644
--- a/components/layout/display_list_builder.rs
+++ b/components/layout/display_list_builder.rs
@@ -952,8 +952,8 @@ impl FragmentDisplayListBuilding for Fragment {
             })
         }
 
-        let center = Point2D::new(absolute_bounds.origin.x + absolute_bounds.size.width / 2,
-                                  absolute_bounds.origin.y + absolute_bounds.size.height / 2);
+        let center = Point2D::new(absolute_bounds.size.width / 2,
+                                  absolute_bounds.size.height / 2);
```

kvark:
Ah, right, the change makes sense. Testing now.

kvark:
Hmm, still failing in `/_mozilla/css/border-image-linear-gradient.html` with that change.

rlhunt:
I got the same thing with the patch, but looking at the reftest comparison they are a fuzzy match with 16 or so pixels (IIRC).

Is that what you're getting too?

kvark:
@rlhunt indeed, there is about 40 pixels along a horisontal line that show a 1-value difference (33/34) with the reference. Interesting to see why, but not a showstopper.

glennw:
@kvark @rlhunt I'm not sure if this is relevant - but the WPT/CSS test runner doesn't (currently) support fuzzy reftests.

glennw:
So anything we land in WR has to result in exact matches for the WPT/CSS tests (which we've managed to do OK so far).

rlhunt:
@glennw Okay, I'd be happy to investigate the fuzziness further. I'll take a look at it next chance I get.

I'm not sure the best way to debug something like this, so I'm open to suggestions. I might take a look at renderdoc.

PR #1080, opened 2017-04-07
Add support for a vector of clip sources and correctly clip box shadow cases where there is no blur radius but there is a border radius

changm:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1080)


changm:
A sample test case:
<div style="
  width: 240px; height: 240px;
  border-radius: 9px;
  box-shadow: 20px 20px grey;
"></div>
~

webrender/src/frame_builder.rs
```diff
@@ -793,6 +792,59 @@ impl FrameBuilder {
...
+
+        // We can draw a rectangle instead with the proper border radius clipping.
+        let mut extra_clips = Vec::new();
+
+        let bs_clip_mode = match clip_mode {
+            BoxShadowClipMode::Outset | BoxShadowClipMode::None => ClipMode::Clip,
+            BoxShadowClipMode::Inset => ClipMode::ClipOut,
+        };
+
+        let box_clip_mode = match clip_mode {
```
kvark: was it intentional to keep this `match` separate from the previous?

webrender/src/frame_builder.rs
```diff
@@ -793,6 +792,59 @@ impl FrameBuilder {
...
+        extra_clips.push(ClipSource::Complex(*box_bounds,
+                                             border_radius,
+                                             box_clip_mode));
+
+
+        let prim = RectanglePrimitive {
+            color: *color,
+        };
+
+        let rect_to_draw = match clip_mode {
```
kvark: could also combine with the previous one

webrender/src/frame_builder.rs
```diff
@@ -793,6 +792,59 @@ impl FrameBuilder {
...
+            BoxShadowClipMode::Inset => ClipMode::ClipOut,
+        };
+
+        let box_clip_mode = match clip_mode {
+            BoxShadowClipMode::Outset | BoxShadowClipMode::None => ClipMode::ClipOut,
+            BoxShadowClipMode::Inset => ClipMode::Clip,
+        };
+
+        // Clip the inside
+        extra_clips.push(ClipSource::Complex(bs_rect,
```
kvark: if `add_primitive` accepts `&[ClipSource]`, you could just have a fixed-size array here:
`let extra_clips = [ClipSource::Complex(...), ClipSource::Complex(...)];`

webrender/src/frame_builder.rs
```diff
@@ -793,6 +792,59 @@ impl FrameBuilder {
...
+        let mut extra_clips = Vec::new();
+
+        let bs_clip_mode = match clip_mode {
+            BoxShadowClipMode::Outset | BoxShadowClipMode::None => ClipMode::Clip,
+            BoxShadowClipMode::Inset => ClipMode::ClipOut,
+        };
+
+        let box_clip_mode = match clip_mode {
+            BoxShadowClipMode::Outset | BoxShadowClipMode::None => ClipMode::ClipOut,
+            BoxShadowClipMode::Inset => ClipMode::Clip,
```
kvark: It appears to me that the `box_clip_mode` is always the opposite of the `bs_clip_mode`. You can elegantly encode this by implementing the [Not](https://doc.rust-lang.org/stable/std/ops/trait.Not.html) operator on `ClipMode` and using it.

webrender/src/frame_builder.rs
```diff
@@ -169,12 +169,9 @@ impl FrameBuilder {
...
             clip_sources.push(ClipSource::Region(clip_region.clone(), RegionMode::ExcludeRect));
         }
-        // TODO(gw): Perhaps in the future it's worth passing in an array
-        //           so that callers can provide an arbitrary number
-        //           of clips?
-        if let Some(extra_clip) = extra_clip {
-            clip_sources.push(extra_clip);
-        }
+
+        clip_sources.extend(extra_clips.iter().cloned());
```
kvark: nit: can use [extend_from_slice](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.extend_from_slice)

webrender/src/frame_builder.rs
```diff
@@ -788,11 +785,53 @@ impl FrameBuilder {
...
+        let prim = RectanglePrimitive {
+            color: *color,
+        };
+
+        self.add_primitive(scroll_layer_id,
+                           &rect_to_draw,
+                           clip_region,
+                           &extra_clips,
+                           PrimitiveContainer::Rectangle(prim));
+        return;
```
kvark: no need for this return

glennw:
I haven't looked at this in detail - but it's not clear to me why this case needs >1 complex clip? It seems that there should only be one complex clip required to handle this?

changm:
@glennw Let's say you have an inset box shadow with a border radius but no blur radius with an offset. We need to apply the border-radius clip for both the outer border radius and inside cut out border radius.

Here's a test case:
`<html>
<head>
<style>
div {
  width: 200px;
  height: 200px;
  box-shadow: 10px 10px black inset;
  border-radius: 10px;
}
</style>
</head>
<body>
<div id="test">
</div>
</body>
</html>`

And a screenshot:

boxshadow

Notice the border radius both inside and outside the box shadow.

glennw:
@changm Ah, that makes sense! There's one remaining nit I can see, but let's get this merged and fix that in a follow up. Thanks!

@bors-servo r+

PR #1081, opened 2017-04-07
Sample now uses full window on Retina display on macOS.

waywardmonkeys:
Using the number of pixels gives us the correct device size.

Fixes issue #725.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1081)


kvark:
Nice, I think `get_inner_size` is deprecated anyway.
Thanks!
@bors-servo r+

PR #1082, opened 2017-04-07
Don't spin the event loop

jrmuizel:
Currently we're polling the cocoa event loop.
This causes us to use unnecessary amount of cpu when we could just be blocked.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1082)

### DIFF
--- a/wrench/src/main.rs
+++ b/wrench/src/main.rs
@@ -417,116 +417,114 @@ fn main() {
             }
         }
 
-        let events = match window {
+        let event = match window {
             WindowWrapper::Headless(..) => {
-                vec![glutin::Event::Awakened]
+                glutin::Event::Awakened
             }
             WindowWrapper::Window(ref window, _) => {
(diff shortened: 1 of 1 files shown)


wrench/src/main.rs
```diff
@@ -417,116 +417,114 @@ fn main() {
...
 
-                    wrench.render();
-                    window.swap_buffers();
+                let now = time::SteadyTime::now();
+                let dur = now - last;
 
-                    let now = time::SteadyTime::now();
-                    let dur = now - last;
+                min_time = min(min_time, dur);
+                min_min_time = min(min_min_time, dur);
```
kvark: these (`min_min` and `max_max`) are not needed here since they are in the `if warmed_up` path below

jrmuizel: My change here was just adjusting the white space. I'd rather make changes like this separately.

PR #1083, opened 2017-04-07
Run reftests with subpixel-aa turned on.

jrmuizel:
This is the configuration that Gecko is currently using and the
configuration that we'd like to end up in so it makes sense to switch
now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1083)


kvark:
AA everything!
@bors-servo r+

PR #1084, opened 2017-04-07
Add testing of text blur

jrmuizel:
For now, just test that blurred text is not equal to unblurred text.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1084)


kvark:
thanks!
@bors-servo r+

PR #1092, opened 2017-04-10
More refactoring of how the batching code works.

glennw:
The primary goal of this change is to allow each primitive type
to easily specify >1 batch. Specifically, we will now be able to
render border corners with one shader and edges with a different
shader, where it makes sense to.

It also cleans up some of the related code, and makes the batching
for the opaque pass work similarly to the alpha pass batching.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1092)


glennw:
r? @kvark 

Sorry for another large patch - at least this one has a net negative line count! :)

PR #1094, opened 2017-04-10
Fix a panic that can occur when the window size is zero.

glennw:
Early exiting here can result in the aux. lists getting out of
sync with the display list.

In the future, we can revisit this and early out. However, this
is an immediate fix for a minimize crash on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1094)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -292,8 +292,7 @@ impl Frame {
         };
 
         if window_size.width == 0 || window_size.height == 0 {
-            println!("ERROR: Invalid window dimensions! Please call api.set_window_size()");
-            return;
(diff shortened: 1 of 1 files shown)


glennw:
Fixes https://github.com/servo/servo/issues/15942

webrender/src/frame.rs
```diff
@@ -293,7 +293,6 @@ impl Frame {
 
         if window_size.width == 0 || window_size.height == 0 {
             println!("ERROR: Invalid window dimensions! Please call api.set_window_size()");
```
mrobinson: Would it make sense to use the error! macro here?

mrobinson:
r=me with one small comment that can be addressed or not.

Issue #1091, opened 2017-04-09
WPT gradient regressions


glennw:
A couple of the gradient reftests are broken with current WR:

```
  ▶ FAIL [expected PASS] /_mozilla/css/border-image-linear-gradient.html
  └   → /_mozilla/css/border-image-linear-gradient.html 9374f4b215c8e79652fcacf5d111ba0a0307ef96
/_mozilla/css/border-image-linear-gradient-ref.html 905bde431c319c7a86a4df9b7543b2a39bd9a22b
Testing 9374f4b215c8e79652fcacf5d111ba0a0307ef96 == 905bde431c319c7a86a4df9b7543b2a39bd9a22b

  ▶ FAIL [expected PASS] /_mozilla/css/linear_gradients_parsing_a.html
  └   → /_mozilla/css/linear_gradients_parsing_a.html d24436d78365996861161375cdad7dcff2c2d5ca
/_mozilla/css/linear_gradients_parsing_ref.html fa3dc596ca9e1f14471de9cb56893e94bd9861b2
Testing d24436d78365996861161375cdad7dcff2c2d5ca == fa3dc596ca9e1f14471de9cb56893e94bd9861b2
```

![rt1](https://cloud.githubusercontent.com/assets/2350882/24841289/9489bbf6-1dc4-11e7-86de-6457182871b1.png)
![rt2](https://cloud.githubusercontent.com/assets/2350882/24841288/948936c2-1dc4-11e7-805d-7300ae5c6297.png)
![rt3](https://cloud.githubusercontent.com/assets/2350882/24841286/94736c84-1dc4-11e7-8661-3d93d0dacbed.png)
![rt4](https://cloud.githubusercontent.com/assets/2350882/24841287/9479c6ec-1dc4-11e7-8736-ec9a6de12cbf.png)

We'll need to get these resolved before we can land a WR update.

cc @kvark @rlhunt

rlhunt:
So one of these failures is resolved with the following patch (from #1078), and the remaining one is a fuzzy match with about 40 pixels off. I will look into the fuzziness problem as soon as I can.
```diff
diff --git a/components/layout/display_list_builder.rs b/components/layout/display_list_builder.rs
index 8c390487f5..091df7e9c8 100644
--- a/components/layout/display_list_builder.rs
+++ b/components/layout/display_list_builder.rs
@@ -952,8 +952,8 @@ impl FragmentDisplayListBuilding for Fragment {
             })
         }
 
[...]

PR #1097, opened 2017-04-10
Fix the condition for using aligned gradients

rlhunt:
The condition for using aligned gradients was not updated when the start
and end positions were made relative to `rect`. This was causing test
fuzzy failures from a mismatched aligned and angle gradient shaders.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1097)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -626,11 +626,11 @@ impl FrameBuilder {
                            tile_repeat.height >= rect.size.height;
 
         let aligned_and_fills_rect = (start_point.x == end_point.x &&
-                                      start_point.y.min(end_point.y) <= rect.min_y() &&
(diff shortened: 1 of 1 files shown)


rlhunt:
r? @glennw @kvark 

This and the patch in #1091 should resolve the WPT regressions. At least until I mess up this condition again in a couple weeks :)

kvark:
The patch looks great! But I feel like this change should be a part of WR update for Servo. Merging it separately may make those tests break until WR is updated, will it not?

rlhunt:
This change shouldn't make any wrench tests break, and it should prevent a regression when the next servo webrender update lands.

Unless I am misunderstanding?

kvark:
D'oh, I through this is going into Servo, not WR, for some reason.
@bors-servo r+

rlhunt:
No problem, thanks!

Issue #1102, opened 2017-04-11
Rounded borders are not clipped properly


mrobinson:
The follow wrench YAML does not show the bottom corners of the border clipped properly:

```yaml
---
root:
  items:
    - type: scroll-layer
      bounds: [0, 0, 100, 100]
      clip:
        rect: [0, 0, 100, 100]
        complex:
          - rect: [0, 0, 100, 100]
            radius:
              top-left: [0, 0]
              top-right: [0, 0]
              bottom-right: [30, 30]
              bottom-left: [30, 30]
      items:
        - type: border
          bounds: [0, 0, 100, 100]
          width: [5, 5, 5, 5]
          border-type: normal
          style: [solid, solid, solid, solid]
          color: [blue, blue, blue, blue]
          radius:
            top-left: [30, 30]
            top-right: [30, 30]
            bottom-right: [0, 0]
            bottom-left: [0, 0]
```

This seems to be an issue with the radius itself. Removing the radius on the border causes the corners to be clipped properly.

PR #1103, opened 2017-04-11
Make it easier to specify border radius in wrench

mrobinson:
This allows for a "radius: [0, 10, 0, 10]" type syntax as well as
allowing any radius component to either be an integer or an array.
A radius component where both the horizontal and vertical component are
the same is the more common case.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1103)


mrobinson:
There are currently no YAML files checked in that use this, but it will be useful when testing a fix for #1102.

PR #1105, opened 2017-04-12
Revert "Inner rectangle compute refactor"

glennw:
This reverts commit 39478f7edb90899f9c8ea7fe4e03efd1b78109d7.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1105)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -16,7 +16,7 @@ use scene::{Scene, SceneProperties};
 use std::collections::HashMap;
 use std::hash::BuildHasherDefault;
 use tiling::{AuxiliaryListsMap, CompositeOps, PrimitiveFlags};
-use util::{ComplexClipRegionHelpers, subtract_rect};
+use util::subtract_rect;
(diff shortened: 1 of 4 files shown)


glennw:
This unfortunately causes a couple of reftest failures in Servo. I think it's probably a fairly simple fix, but in the interests of being able to land a WR update, I'm reverting this temporarily.

glennw:
@kvark There's two tests that fail with this patch - ```overflow_border_radius``` and ```calc-border-radius-1```. There is a crack that appears. I believe what's happening is that the inner rect gets a fractional coordinate, and this interacts badly with the device pixel snapping between the inner (opaque) rect and the outer (clipped) rect.

glennw:
@kvark I think we could probably do something like ensure the fractional coord of the inner rect matches the fractional coord of the outer rect?

glennw:
r? anyone

kvark:
@glennw I understand why it doesn't work correctly with pixel snapping. Will fix tomorrow, hopefully. Sorry about the trouble!

glennw:
@kvark No worries, thanks!

PR #1106, opened 2017-04-12
Introduce border corner and edge shader path.

glennw:
This introduces the new border shader path, and uses it for all
solid border segments. It fixes a lot of cases that weren't previously
handled correctly. Specifically,

* AA on edge color transitions in corners.
* Correct elliptical border radii and AA.
* Improved AA for circular radii.
* Handle AA correctly for pinch zoom / perspective transforms.
* Correct angle for color transitions in corners.

The next step is to port the other border styles over to this new
code path.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1106)


glennw:
Examples of the handled border cases and AA when zooming:

aa2
borders1
borders2
borders3
borders4
borders5
borders6
borders7

kvark:
Merging now, can be continued in follow-ups.
@bors-servo r+

staktrace:
@glennw I believe this patch caused a reftest failure in gecko - see https://bugzilla.mozilla.org/show_bug.cgi?id=1355475#c4

glennw:
@staktrace  This is fixed by https://github.com/servo/webrender/pull/1120.

staktrace:
Verified that #1120 fixed the regression, thanks!

PR #1107, opened 2017-04-12
Revive "Inner rectangle compute refactor"

kvark:
Reverts #1105
Fixes the inner rectangle to always use rounded offsets
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1107)


glennw:
Makes sense.

@bors-servo r+

PR #1108, opened 2017-04-12
Use save_flipped to save the png

jrmuizel:
THis avoids duplicating the flip code and makes the read_pixels
call match the other read_pixels calls

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1108)


kvark:
Code looks great, thank you!

Notice that this is a slight behavior change - the code used to ignore alpha of the framebuffer, and now it's actually read and saved. I see this as an improvement.

@bors-servo r+

PR #1109, opened 2017-04-12
Convenient data fetching in the shaders

kvark:
Something I meant to do since the inception...
> 62 additions and 130 deletions

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1109)


glennw:
Nice! Hopefully this doesn't trigger that weird mac-specific shader compiler bug we hit a while ago :)

@bors-servo r+

PR #1111, opened 2017-04-13
Fix cracks that can appear at high zoom factors.

glennw:
Some primitives (such as borders and gradients) can be drawn
as multiple sub-primitives.

In these cases, use a common reference point for the device
pixel snap amount, so that no cracks appear at high zoom factors.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1111)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -524,12 +524,13 @@ VertexInfo write_vertex(RectWithSize instance_rect,
                         RectWithSize local_clip_rect,
                         float z,
                         Layer layer,
-                        AlphaBatchTask task) {
+                        AlphaBatchTask task,
+                        vec2 snap_ref) {
     // Select the corner of the local rect that we are processing.
     vec2 local_pos = instance_rect.p0 + instance_rect.size * aPosition.xy;
 
     // xy = top left corner of the local rect, zw = position of current vertex.
-    vec4 local_p0_pos = vec4(instance_rect.p0, local_pos);
+    vec4 local_p0_pos = vec4(snap_ref, local_pos);
 
     // Clamp to the two local clip rects.
     local_p0_pos = clamp_rect(local_p0_pos, local_clip_rect);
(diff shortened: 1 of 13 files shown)


kvark:
looks good!
@bors-servo r+

staktrace:
@glennw This patch seems to have caused a reftest failure as well. https://bugzilla.mozilla.org/show_bug.cgi?id=1355475#c6

glennw:
@staktrace I'll wait until https://github.com/servo/webrender/pull/1120 lands and has a try run to see if that fixes this bug too.

staktrace:
It looks like #1120 did not fix the regression here, unfortunately. [Latest try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=d43afededdec762b9adde7b32efa7cb9955ca803)

staktrace:
I just realized that the reftest in question (layout/reftests/columns/column-balancing-overflow-005.html) was already marked as fuzzy-if(webrender,255,6). The failure with this PR is that instead of pixels being off, 12 pixels are off. It's quite possible that this is "not a bug" and we just need to update the test expectations to by fuzzy-if(webrender,255,12) instead.

glennw:
@staktrace OK, interesting. Unless I made an obvious typo / bug in the PR itself, I think this PR should strictly be an improvement - it just ensures that every sub-rect of a single primitive is snapped consistently. Do you want me to look into it further today, or should I leave it to you to investigate further first?

glennw:
@staktrace It is always possible that even though this is an improvement, it exposes a different / unrelated WR issue in a different way, which I guess is what you're saying above...

staktrace:
@glennw So I did a some comparisons using:
- OS X Firefox build from the [graphics branch](https://queue.taskcluster.net/v1/task/OaaoTt4pTfGuBxelSZtuRA/runs/0/artifacts/public/build/target.dmg) (this uses the older WR, without your patch).
- OS X Firefox build from the [try push](https://queue.taskcluster.net/v1/task/MsWYCT1YREO8orUnse7RIg/runs/0/artifacts/public/build/target.dmg) (this uses the newer WR, with patch)

On both builds, I loaded the layout/reftests/columns/column-balancing-overflow-005.html file and zoomed in to various zoom levels. Visually they look pretty much the same, so I think I'll just update the fuzziness on the test and call it a day.

I did notice however that the WR-drawn outline looks a lot worse than the Gecko-drawn outline (e.g. compare [gecko](https://people-mozilla.org/~kgupta/tmp/nightly-240.png) with [wr](https://people-mozilla.org/~kgupta/tmp/newwr-240.png) - that's the test page at 240% zoom). But that seems to be the case with and without your patch.

glennw:
@staktrace Sure, sounds good - thanks for investigating.

In terms of the quality - I've landed the initial border refactoring, but this is only currently used for solid / inset / outset borders and outlines. I'm working this week on porting the rest of the border styles to the new code path, which should drastically improve the rendering quality of those other styles.

PR #1114, opened 2017-04-13
Fix combined_local_viewport_rect for nested nodes

mrobinson:
The calculation of the combined_local_viewport_rect was being done
using the accumulated_scroll_offset, but this was not correct, because
the parent combined_local_viewport_rect was already calculated to be
independent of those offsets. This means that the only scrolling offset
that matters is one between the parent and the child.

Add a test for this as well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1114)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -235,6 +235,7 @@ impl ClipScrollTree {
         self.update_node_transform(root_reference_frame_id,
                                    &LayerToWorldTransform::create_translation(pan.x, pan.y, 0.0),
                                    &as_scroll_parent_rect(&root_viewport),
+                                   LayerPoint::zero(),
                                    LayerPoint::zero());
     }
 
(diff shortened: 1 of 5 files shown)


webrender/src/clip_scroll_tree.rs
```diff
@@ -235,39 +235,48 @@ impl ClipScrollTree {
...
-                    let (transform, offset) = match node.node_type {
+                    let (transform, offset, accumulated_scroll_offset) = match node.node_type {
                         NodeType::ReferenceFrame(..) =>
-                            (node.world_viewport_transform, LayerPoint::zero()),
+                            (node.world_viewport_transform, LayerPoint::zero(), LayerPoint::zero()),
                         NodeType::Clip(_) => {
                             (*parent_reference_frame_transform,
-                             parent_accumulated_scroll_offset + node.scrolling.offset)
+                             node.scrolling.offset,
+                             node.scrolling.offset + parent_accumulated_scroll_offset)
```
kvark: why does this still need to use `parent_accumulated_scroll_offset`?

mrobinson: The parent_accumulated_scroll_offset is used to setup the content and viewport transformation for each node. In those cases we do need to take into account the combined scroll offsets.

PR #1115, opened 2017-04-13
Remove last duplicated cargo dependency

### LINKED ISSUE #899: Deduplicate dependencies

mrobinson:
Also have servo-tidy start looking at Cargo.lock so we can more quickly
detect duplicated dependencies in the future.

Fixes #899.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1115)

### DIFF
--- a/wrench/Cargo.toml
+++ b/wrench/Cargo.toml
@@ -16,7 +16,7 @@ app_units = "0.4"
 image = "0.12"
 clap = { version = "2", features = ["yaml"] }
 lazy_static = "0.2"
-yaml-rust = { git = "https://github.com/vvuk/yaml-rust", features = ["preserve_order"] }
+yaml-rust = "0.3.5"
 serde_json = "0.9"
 time = "0.1"
 crossbeam = "0.2"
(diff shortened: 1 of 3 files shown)


mrobinson:
Going to close this for the moment, because I think the preserve_order feature that I accidentally removed is causing some issues.

wrench/Cargo.toml
```diff
@@ -16,7 +16,7 @@ app_units = "0.4"
 image = "0.12"
 clap = { version = "2", features = ["yaml"] }
 lazy_static = "0.2"
-yaml-rust = { git = "https://github.com/vvuk/yaml-rust", features = ["preserve_order"] }
+yaml-rust = "0.3.5"
```
kvark: I think we still need to use the `preserve_order` feature

mrobinson:
The issue here is that clap has an implicit API depndency on yaml-rust 0.3.4, so we can't remove this duplicate with removing the yaml feature of clap.

wrench/Cargo.toml
jrmuizel: Yep, preserve_order is needed to make the yaml writer's output saner.

PR #1117, opened 2017-04-13
correct for empty box shadow rects

changm:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1117)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -786,9 +786,13 @@ impl FrameBuilder {
 
         let bs_rect = box_bounds.translate(box_offset)
                                 .inflate(inflate_amount, inflate_amount);
+        // If we have negative inflate amounts.
+        // Have to explicitly check this since euclid::TypedRect relies on negative rects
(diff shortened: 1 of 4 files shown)


changm:
The explicit empty check is to work around empty rects until we resolve https://github.com/servo/euclid/pull/183#issuecomment-293935215

webrender/src/frame_builder.rs
```diff
@@ -786,9 +786,13 @@ impl FrameBuilder {
 
         let bs_rect = box_bounds.translate(box_offset)
                                 .inflate(inflate_amount, inflate_amount);
+        // If we have negative inflate amounts.
+        // Have to explicitly check this since euclid::TypedRect relies on negative rects
+        let bs_rect_empty = bs_rect.size.width <= 0.0 || bs_rect.size.height <= 0.0;
```
kvark: should the conditions be `<` instead of `<=`, given the comment says "negative"?

changm: No, but I put "negative" there since Euclid correctly handles 0 sized rects. I can update the comment to reflect that if you want. I was hoping to just do bs_rect.is_empty().

PR #1118, opened 2017-04-13
Update WebRender version

mrobinson:
This is necessary so that we can pull in recent fixes to Servo.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1118)


kvark:
@bors-servo r+
you might also want to wait for #1120 and #1117 before updating Servo

glennw:
@mrobinson There is a pending update PR here https://github.com/servo/servo/pull/16398 that you may want to either land first, or base your update PR on.

mrobinson:
@glennw Thanks. I'll prepare a new Servo patch. I don't want to hold up your patch while I run tests and such.

PR #1119, opened 2017-04-13
two minor fixups

Gankro:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1119)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -654,6 +654,13 @@ impl Profiler {
         self.x_right = 400.0;
         self.y_right = 40.0;
 
+        let mut gpu_time = 0;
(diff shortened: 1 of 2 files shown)


webrender/src/profiler.rs
```diff
@@ -654,6 +654,13 @@ impl Profiler {
         self.x_right = 400.0;
         self.y_right = 40.0;
 
+        let mut gpu_time = 0;
+        let gpu_samples = mem::replace(&mut renderer_timers.gpu_samples, Vec::new());
```
kvark: alternatively,
```rust
let gpu_samples = renderer_timers.gpu_samples.drain(..).fold(0, |u, s| u + s.time_ns);
```

Gankro: I'll go with `renderer_timers.gpu_samples.drain(..).map(|x| x.time_ns).sum();`

PR #1120, opened 2017-04-13
Fix incorrect border radii for top right corner.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1120)


glennw:
r? @kvark 

This fixes a regression in the gecko reftests.

jrmuizel:
Can you add a test?

kvark:
Nice!
I agree that more reftests are needed, but there is no need to hold this patch.

@bors-servo r+

glennw:
Added test :)

PR #1121, opened 2017-04-14
Tidy: Ignore selected packages instead of entire Cargo.lock file

UK992:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1121)


kvark:
Looks reasonable, and Travis agrees :)
Thanks!
@bors-servo r+

PR #1123, opened 2017-04-16
Add profiling for IPC overhead

Gankro:
Note: this adds a dependency on `time` to `webrender_traits`.

The purpose of this is to evaluate the overhead of different IPC implementations and serialization strategies. In particular, we need to replace the current implementation, as it is unsound; it just blindly transmutes, leaving room for a malicious process to create Undefined representations of e.g. enums.

We take 4 time stamps:

* before serialization to `Vec<u8>` happens
* after serialization
* before deserialization
* after deserialization

Because some of these time stamps happen in webrender_traits and cross-process, we need to take them manually and send the first two over IPC. Currently I'm just stuffing them in one of the descriptors, since that's easy and minimally invasive (but Firefox's headers will need to be updated, I think).

We consider the time to perform the message send cross-process to be the time between serialization ending and deserialization starting. Overall IPC overhead is the whole process. We also record how large the two sent buffers are, as this may change significantly with a Proper Serialization Mechanism.

In my preliminary testing basically all overhead is currently in the actual message send process (about 0.5ms on Mac in Servo). There's sometimes some overhead from deserialization, which is probably stuffing the auxlist in a map. About 200k is being sent for the display list, and 100k for the aux list.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1123)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -287,19 +287,58 @@ impl TextureCacheProfileCounters {
 
 #[derive(Clone)]
 pub struct BackendProfileCounters {
+    pub total_time: TimeProfileCounter,
+    pub resources: ResourceProfileCounters,
+    pub ipc: IpcProfileCounters,
+}
+
+#[derive(Clone)]
+pub struct ResourceProfileCounters {
     pub font_templates: ResourceProfileCounter,
     pub image_templates: ResourceProfileCounter,
-    pub total_time: TimeProfileCounter,
     pub texture_cache: TextureCacheProfileCounters,
+}
+
+#[derive(Clone)]
+pub struct IpcProfileCounters {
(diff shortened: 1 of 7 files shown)


Gankro:
Where would you suggest putting them? I didn't want to mess up the code threading multiple profiling structs through.

kvark:
@Gankro I was thinking - a separate struct `IpcProfileCounters` alongside the backend ones. Alternatively, refactor this to be something like:
```rust
struct BackendProfileCounters {
  total_time: TimeProfileCounter,
  resources: ResourceProfileCounters, //includes templates and texture cache
  ipc: IpcProfileCounters,
}
```

Gankro:
@kvark done

kvark:
Thanks @Gankro !
@bors-servo r+

PR #1126, opened 2017-04-17
Add a new image format RG8 in ImageFormat.

JerryShih:
@glennw @kvark @nical 
Some video decoders are decoding to NV12 format. WR already have A8 for y channel. We also need RG8 for uv interleaving channel.

https://wiki.videolan.org/YUV/#NV12

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1126)


kvark:
Ok, it appears that GL 3.0 and GLES 3.0 support RG textures in the core.
@bors-servo r+

PR #1128, opened 2017-04-17
Clipping the clip group rectangle to the screen bounds

### LINKED ISSUE #1122: thread 'RenderBackend' panicked at 'Each render task must allocate <= size of one target! (1601×1485)
While running a gecko reftest with some local changes to the gecko-side code, I am able to trigger an assertion failure in webrender. It might very well be that I'm doing something wrong on the gecko side, but given the assertion that's failing I think this is likely a bug on the webrender side. Note that I'm using webrender at cset 3dd68f54e12bd5abf8ef41de4d4ec851620f7e4e. I was able to save a binary recording which reproduces the assertion if I run `wrench replay --play wr-record-3.bin`. I don't know if that's useful but you can get the wr-record-3.bin [here](https://people-mozilla.org/~kgupta/tmp/wr-record-3.bin).

kvark:
Fixes #1122 

r? @mrobinson or @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1128)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -9,11 +9,11 @@ use prim_store::GpuBlock32;
 use renderer::VertexDataStore;
 use spring::{DAMPING, STIFFNESS, Spring};
 use tiling::PackedLayerIndex;
-use util::TransformedRect;
+use util::TransformedRectKind;
 use webrender_traits::{ClipId, ClipRegion, LayerPixel, LayerPoint, LayerRect, LayerSize};
 use webrender_traits::{LayerToScrollTransform, LayerToWorldTransform, PipelineId};
 use webrender_traits::{ScrollEventPhase, ScrollLayerRect, ScrollLocation, WorldPoint};
-use webrender_traits::WorldPoint4D;
+use webrender_traits::{DeviceIntRect, WorldPoint4D};
 
 #[cfg(target_os = "macos")]
 const CAN_OVERSCROLL: bool = true;
@@ -37,7 +37,7 @@ pub struct ClipInfo {
     /// The final transformed rectangle of this clipping region for this node,
     /// which depends on the screen rectangle and the transformation of all of
     /// the parents.
-    pub xf_rect: Option<TransformedRect>,
+    pub screen_bounding_rect: Option<(TransformedRectKind, DeviceIntRect)>,
 }
 
 impl ClipInfo {
@@ -52,7 +52,7 @@ impl ClipInfo {
             mask_cache_info: MaskCacheInfo::new(&clip_sources, clip_store),
             clip_sources: clip_sources,
             packed_layer_index: packed_layer_index,
-            xf_rect: None,
+            screen_bounding_rect: None,
         }
     }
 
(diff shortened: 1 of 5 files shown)


kvark:
Not sure if the purpose of the old code was compatibility with scrolling, or just an oversight. Please review carefully.

glennw:
This looks fine to me, but I'll leave it for @mrobinson to take a look over too. It'd be good to add a comment around the screen rect intersection explaining why it's necessary.

jrmuizel:
Add a test?

webrender/src/frame_builder.rs
```diff
@@ -1535,9 +1535,10 @@ impl<'a> LayerRectCalculationAndCullingPass<'a> {
             };
 
             if bounding_rect.is_none() {
-                bounding_rect =
-                    Some(clip_info.xf_rect.as_ref().map_or_else(DeviceIntRect::zero,
-                                                                |x| x.bounding_rect))
+                bounding_rect = Some(match clip_info.xf_rect {
```
mrobinson: This all seems okay to me. A slightly cleaner change might be just delay the intersection until this point and write:
```rust
            if bounding_rect.is_none() {
                let rect = clip_info.xf_rect.as_ref().map_or_else(DeviceIntRect::zero,
                                                                  |x| x.bounding_rect);
                let rect = rect.intersection(self.screen_rect).unwrap_or_else(DeviceIntRect::zero);
                bounding_rect = Some(rect);
            }
```
I think that would be the only change necessary. This will also be useful if we ever need to use the real translated rectangle in the future again.

kvark: Right. My concern is - if we only patch with `screen_rect` here, then there might be other places where `xf_rect` going out of bounds will become a problem. So I figured fixing it upstream would be safer than patching downstream.

kvark:
@jrmuizel a basic test is added now

webrender/src/frame_builder.rs
mrobinson: If you want to calculate the intersection sooner then I recommend that you rename the `xf_rect` members to something like `screen_bounding_rect` and update the comments to reflect that they are the intersected onscreen bounding rects for their respective data structures.

kvark: Interestingly, the old comments seem to match the new code better:
>     /// The final transformed rectangle of this clipping region for this node,
>     /// which depends on the screen rectangle and the transformation of all of
>     /// the parents.

kvark:
@mrobinson `xf_rect` is renamed now

webrender/src/clip_scroll_node.rs
```diff
@@ -37,7 +37,7 @@ pub struct ClipInfo {
     /// The final transformed rectangle of this clipping region for this node,
```
mrobinson: I think it's useful to say "the bounding rectangle of the transformed rectangle" in the comment just to clarify. Might also be worth mentioning the other part of the tuple as well.

mrobinson:
Oops. This is about to land, so I've deleted my comment nit comment.

PR #1129, opened 2017-04-18
Fix broken payload deserialization.

jrmuizel:
Payload deserialization was broken when we changed the format
a little while ago. Using the from_data() will help us avoid
the code from diverging in the futre.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1129)

### DIFF
--- a/webrender_traits/src/channel.rs
+++ b/webrender_traits/src/channel.rs
@@ -45,8 +45,8 @@ impl Payload {
     }
 
     /// Deserializes the given payload from a raw byte vector.
-    pub fn from_data(data: Vec<u8>) -> Payload {
-        let mut payload_reader = Cursor::new(&data[..]);
+    pub fn from_data(data: &[u8]) -> Payload {
+        let mut payload_reader = Cursor::new(data);
         let epoch = Epoch(payload_reader.read_u32::<LittleEndian>().unwrap());
         let pipeline_id = PipelineId(payload_reader.read_u32::<LittleEndian>().unwrap(),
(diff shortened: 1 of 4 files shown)


wrench/src/yaml_frame_writer.rs
```diff
@@ -308,14 +309,10 @@ impl YamlFrameWriter {
         let dl_desc = self.dl_descriptor.take().unwrap();
         let aux_desc = self.aux_descriptor.take().unwrap();
 
-        assert_eq!(data.len(), dl_desc.size() + aux_desc.size() + 4);
+        let payload = Payload::from_data(data);
```
kvark: is there way for us to still have an assert? It appears that the new code no longer tries to check for the data length

webrender_traits/src/channel.rs
```diff
@@ -45,7 +45,7 @@ impl Payload {
     }
 
     /// Deserializes the given payload from a raw byte vector.
-    pub fn from_data(data: Vec<u8>) -> Payload {
+    pub fn from_data(data: &[u8]) -> Payload {
         let mut payload_reader = Cursor::new(&data[..]);
```
kvark: `&data[..]` is no longer needed here

kvark:
Wonderful, thanks @jrmuizel !
Please tidy up before merging:
> ./wrench/src/yaml_frame_writer.rs:22: use statement is not in alphabetical order
	expected: webrender_traits::channel::Payload
	found: webrender_traits::SpecificDisplayItem::*

PR #1135, opened 2017-04-18
Add support for font index to add_raw_font

lsalzman:
In Gecko, we need to be able to support font indexes in case the raw font is actually a font collection.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1135)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -126,11 +126,12 @@ impl FontContext {
         }
     }
 
-    pub fn add_raw_font(&mut self, font_key: &FontKey, bytes: &[u8]) {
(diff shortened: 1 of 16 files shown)


webrender/src/platform/macos/font.rs
```diff
@@ -126,7 +126,7 @@ impl FontContext {
         }
     }
 
-    pub fn add_raw_font(&mut self, font_key: &FontKey, bytes: &[u8]) {
+    pub fn add_raw_font(&mut self, font_key: &FontKey, bytes: &[u8], _index: u32) {
```
kvark: should we `assert_eq!(index, 0)` here?

kvark:
Thank you! Please fix CI to proceed.

kvark:
:shipit: 
@bors-servo r+

PR #1137, opened 2017-04-19
Tidy up and optimize a few cases in the border corner shader.

glennw:
This is mostly tidy up to make it simpler to add the other border styles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1137)

### DIFF
--- a/webrender/res/ps_border_corner.fs.glsl
+++ b/webrender/res/ps_border_corner.fs.glsl
@@ -59,6 +59,17 @@ float distance_to_line(vec2 p0, vec2 perp_dir, vec2 p) {
     return dot(normalize(perp_dir), dir_to_p0);
 }
 
+float distance_to_ellipse(vec2 p, vec2 radii) {
+    // sdEllipse fails on exact circles, so handle equal
+    // radii here. The branch coherency should make this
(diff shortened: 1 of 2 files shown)


webrender/res/ps_border_corner.fs.glsl
```diff
@@ -78,32 +89,11 @@ void main(void) {
...
-            vec2 p = local_pos - vClipCenter;
-            d0 = sdEllipse(p, vOuterRadii);
-        }
-
-        if (vInnerRadii.x == vInnerRadii.y) {
-            d1 = distance(vClipCenter, local_pos) - vInnerRadii.x;
-        } else {
-            // Get distance to inner ellipse. Skip if the inner
-            // radius is <= 0.0 to avoid FP errors.
-            if (vInnerRadii.x <= 0.0 || vInnerRadii.y <= 0.0) {
```
kvark: are we missing this check with the new code?
Edit: I don't think we ever hit it. `vInnerRadii` is either both zeroes, or both positive. In the former case it will go the `vInnerRadii.x == vInnerRadii.y` path, so this branch will never hit (with either new or old code).

PR #1141, opened 2017-04-19
Text shadows should reference the proper render task data index

changm:
This for some reason only appeared if we had a single glyph with a blur. In these cases, no text would render. Text shadows render the text shadow as two blurs, then renders the cached image. The cached image would come from a render task. However, we would read the incorrect render task data [1], resulting in rects that are incorrect. This patch correctly identifies the render task id.

What's super confusing for me is that multiple glyphs would blur (But somewhat incorrectly). It seems like if the render_task_id it outside the length of the render_task_data, we should be getting junk data. But somehow, something was still rendering, which makes no sense to me.

Also, there is a red outline around the blurred text now, which I will fix in a follow up. 

[1] https://github.com/servo/webrender/blob/master/webrender/res/ps_cache_image.vs.glsl#L19

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1141)


changm:
Pre-patch screenshot of the sample/main.rs with a 10 blur radius:

orig

changm:
With this patch:

fix

webrender/src/tiling.rs
```diff
@@ -489,9 +489,22 @@ impl AlphaRenderItem {
                         let key = AlphaBatchKey::new(batch_kind, flags, blend_mode, textures);
                         let batch = batch_list.get_suitable_batch(&key, item_bounding_rect);
 
+                        let mut cache_task_index = 0;
```
kvark: this would look cleaner in functional style, `let cache_task_index = match prim_metadata.render_task {...};`

webrender/src/tiling.rs
```diff
@@ -489,9 +489,22 @@ impl AlphaRenderItem {
...
+                            let cache_task_id = &prim_metadata.render_task.as_ref().unwrap().id;
+                            cache_task_index = render_tasks.get_task_index(cache_task_id,
+                                                                           child_pass_index).0 as i32;
+                        }
+
                         for glyph_index in 0..prim_metadata.gpu_data_count {
+                            let user_data0 = match batch_kind {
+                                AlphaBatchKind::TextRun => text_cpu.resource_address.0 + glyph_index,
+                                AlphaBatchKind::CacheImage => cache_task_index,
+                                _ => panic!("Invalid batch kind for text"),
```
kvark: nit: this is actually `unreachable!`

changm:
I still have no clue why this worked before. Can anyone care to explain?

kvark:
@changm I looked at it and couldn't find a reasonable explanation ¯\_(ツ)_/¯
@bors-servo r+

PR #1142, opened 2017-04-20
Add support for double style borders to new border path.

glennw:
Also add a border suite to test various parts of the new
border shader code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1142)


glennw:
In some cases the AA is not quite strong enough - but I figure it's worth getting this merged now and iterating on it. In the current state it passes all the Servo tests and also draws borders of 1, 2, 3 px widths (mostly) correctly.

jrmuizel:
In gecko we'll tend to use a mask on the antialiasing of the edges of borders for reftests so they don't need updating when if the antialiasing is different for some reason.

glennw:
@jrmuizel That sounds a good idea in general - although in this case I'm specifically wanting to test the AA algorithm for quality purposes. We could definitely use that for reftests that don't care about the AA result though.

glennw:
@kvark Thanks for the review - updated and pushed fixes for those.

kvark:
Thanks!
@bors-servo r+

PR #1143, opened 2017-04-20
Use LayoutPoints for glyph offsets instead of untyped units

jrmuizel:
This will help with the binding generator

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1143)


glennw:
Seems sensible.

@bors-servo r+

jrmuizel:
I forgot to change the sample. The commit has been updated.

Issue #1060, opened 2017-04-03
Panic in render_backend when saving a wrench recording to yaml


rlhunt:
```
[rhunt@rhunt-thinkpad wrench]$ RUST_BACKTRACE=1 cargo run -- --save yaml replay ../../firefox/wr-record-1.bin 
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `/home/rhunt/src/webrender/target/debug/wrench --save yaml replay ../../firefox/wr-record-1.bin`
OpenGL version 4.5 (Core Profile) Mesa 17.0.2, Mesa DRI Intel(R) HD Graphics P530 (Skylake GT2) 
hidpi factor: 1 (native 1)
Shader override path: None
Warning: binary file ApiMsg type mismatch: expected 0x3c2747c08416cc16, found 0xbd69ed4130599056
Frame 0 offset: 16
Frame 1 offset: 41
thread 'RenderBackend' panicked at 'called `Option::unwrap()` on a `None` value', /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libcore/option.rs:323
stack backtrace:
   1:     0x55aff9cffd1c - std::sys::imp::backtrace::tracing::imp::write::hf33ae72d0baa11ed
                        at /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:42
   2:     0x55aff9d037ae - std::panicking::default_hook::{{closure}}::h59672b733cc6a455
                        at /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/panicking.rs:351
   3:     0x55aff9d033b4 - std::panicking::default_hook::h1670459d2f3f8843
                        at /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/panicking.rs:367
   4:     0x55aff9d03c4b - std::panicking::rust_panic_with_hook::hcf0ddb069e7beee7
                        at /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/panicking.rs:555
   5:     0x55aff9d03ae4 - std::panicking::begin_panic::hd6eb68e27bdf6140
                        at /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/panicking.rs:517
   6:     0x55aff9d03a09 - std::panicking::begin_panic_fmt::hfea5965948b877f8
[...]

PR #1144, opened 2017-04-20
Record root_scroll_node and use ClipId::Clip in show

rlhunt:
I ran into #1060 again. The problem is `replay --save yaml` followed by a `show yaml`. I put together a fix which makes sense to me, but I don't know the logic behind this code well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1144)

### DIFF
--- a/wrench/src/yaml_frame_reader.rs
+++ b/wrench/src/yaml_frame_reader.rs
@@ -576,7 +576,7 @@ impl YamlFrameReader {
 
             let yaml_clip_id = item["clip-id"].as_i64();
             if let Some(yaml_id) = yaml_clip_id {
-                let id = ClipId::new(yaml_id as u64, self.builder().pipeline_id);
+                let id = ClipId::Clip(yaml_id as u64, self.builder().pipeline_id);
                 self.builder().push_clip_id(id);
             }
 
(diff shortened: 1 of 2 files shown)


rlhunt:
@mrobinson thoughts?

Bin file with the [issue](https://drive.google.com/file/d/0B84Jf1P5fuSpREF3U1BlRFRodlk/view?usp=sharing)
cset it was generated with, `07b6c6a1f93b5d8af1dd9ae825906dcf5c310810`

mrobinson:
Ah, I think I understand the issue and by chance it's related to a change I was working on yesterday. Can you tell me if this patch fixes the issue?

https://github.com/mrobinson/webrender/commit/7ab70ec8dc92b2e4a1c698e09acce9421bebcdc1

mrobinson:
@rlhunt, I think I have a fix for this, based on yours, but do you think you could upload the YAML output somewhere? I cannot reproduce the problem you are seeing using the bin file. I'm guessing this is due to incompatibilities in the raw output.

glennw:
@mrobinson @rlhunt What's the status of this PR?

mrobinson:
I'm hoping that my commit in https://github.com/mrobinson/webrender/commit/03bd8d430a49d223c7a0f4a2ca18ebffc25c47ce fixes this issue. Perhaps @rlhunt can confirm since I'm having trouble reproducing it.

rlhunt:
mrobinson@03bd8d4 fixes that crashes I was getting. There is now a problem when I show the yaml frames, something weird with transforms. But that was there with this PR too so it's something else.

Thanks!

PR #1146, opened 2017-04-20
Avoid unneeded WebGL ctx.make_current() calls.

MortimerGoro:
Avoid unneeded WebGL ctx.make_current() calls

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1146)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -310,6 +310,9 @@ impl RenderBackend {
                                 };
 
                                 let result = wrapper.new_context(size, attributes, dispatcher);
+                                // Creating a new GLContext may make the current bound context_id dirty.
+                                // Clear it to ensure that  make_current() is called in subsequent commands.
+                                self.current_bound_webgl_context_id = None;
 
                                 match result {
                                     Ok(ctx) => {
@@ -336,7 +339,10 @@ impl RenderBackend {
                         }
                         ApiMsg::ResizeWebGLContext(context_id, size) => {
                             let ctx = self.webgl_contexts.get_mut(&context_id).unwrap();
-                            ctx.make_current();
+                            if Some(context_id) != self.current_bound_webgl_context_id {
+                                ctx.make_current();
+                                self.current_bound_webgl_context_id = Some(context_id);
+                            }
                             match ctx.resize(&size) {
                                 Ok(_) => {
                                     // Update webgl texture size. Texture id may change too.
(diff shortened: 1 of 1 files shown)


MortimerGoro:
@kvark could that happen? render_backend thread can only be accessed via predefined channel messages. If someone decides to call ctx.make_current() it must be in another thread which seems a lot of trouble, optimized current_bound_webgl_context_id checks or not.

emilio:
Is it that expensive to do? We already check whether the context is not current. That calls into the driver, but I wouldn't expect that to be a perf bottleneck. If it is, then this looks a reasonable solution.

emilio:
(To clarify, I mean that all the `make_current` impls are guarded with a `!self.is_current()` clause: https://github.com/emilio/rust-offscreen-rendering-context/blob/master/src/platform/with_glx/native_gl_context.rs#L122)

kvark:
Right, it doesn't make sense to cache it on WR level if it's already done at the lower one.

MortimerGoro:
I was worried about perf because of the driver call, the difference might be small but with e.g. 200 WebGL calls per frame may start having some impact. 

I'll do some tests to measure performance. I checked some modifications at the same time (e.g. disabling glError) and it made a difference. But didn't check this one alone. I'll let you know.

It should be a fast call, but I'm a bit paranoid with some GPUs, specially on Android.

kvark:
I don't think you even need to benchmark it. You can take an apitrace capture of webgl running, and then see if it's riddled with "MakeCurrent" calls.

MortimerGoro:
I did a quick test using offscreen-context-library:
```
#[test]
fn test_time() {
    let ctx = (...)
    ctx.make_current();
    for _ in 0..10 {
        let prev = timestamp();
        for i in 0..400 {
             let _ = ctx.make_current();
        }
        let t = timestamp() - prev;
        println!("Time: {}", t);
    }
}
```
It's negligible on desktop but perceptible on mobile. I think that it's worth it to optimize for mobile.

Times in milliseconds:

Google Pixel
```
Time: 0.249267578125
Time: 0.247314453125
Time: 0.314453125
Time: 0.24560546875
Time: 0.2470703125
Time: 0.2470703125
Time: 0.24560546875
Time: 0.2470703125
Time: 0.246826171875
```
 Nexus 4
```
Time: 0.3662109375
Time: 0.3662109375
Time: 0.366455078125
Time: 0.396728515625
Time: 0.3662109375
Time: 0.3662109375
Time: 0.3662109375
Time: 0.3662109375
Time: 0.396728515625
```

 Samsung Galaxy Tab
```
Time: 0.54931640625
Time: 0.54931640625
Time: 0.54931640625
Time: 0.549560546875
Time: 0.54931640625
Time: 0.54931640625
Time: 0.54931640625
Time: 0.54931640625
Time: 0.549560546875
```

Ubuntu Desktop 1060 GTX
```
Time: 0.05126953125
Time: 0.051025390625
Time: 0.051025390625
Time: 0.051025390625
Time: 0.051025390625
Time: 0.051025390625
Time: 0.05126953125
Time: 0.051025390625
Time: 0.05126953125
```

 iMac 2010
```
Time: 0.001953125
Time: 0.003173828125
Time: 0.001953125
Time: 0.001953125
Time: 0.001953125
Time: 0.001953125
Time: 0.001953125
Time: 0.001953125
Time: 0.003173828125
```

kvark:
@emilio ah, `offscreen_rendering_context` is calling into the driver for those checks. In this case, I consider the PR quite helpful, and @MortimerGoro timings confirm that (thanks for benchmarking!).

kvark:
@emilio 
> That calls into the driver, but I wouldn't expect that to be a perf bottleneck. If it is, then this looks a reasonable solution.

Not a bottleneck, but still visible. I suppose you are not against merging this.

@bors-servo r+

PR #1148, opened 2017-04-21
Clear render targets with rgba(0,0,0,0) and remove isolate_clear

rlhunt:
The composite shader relies on its render target being premultiplied.
This causes issues with transparent surfaces behind a mix-blend-mode.
The clipping code no longer relies on the previous behavior, so this
can be changed.

The isolate_clear code then no longer has an effect, so that is removed
too. Notably, the code for isolating stacking context's remains as only
the code that clears the rects is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1148)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1076,7 +1076,6 @@ impl FrameBuilder {
         let mut sc_stack = Vec::new();
         let mut current_task = RenderTask::new_alpha_batch(next_task_index,
                                                            DeviceIntPoint::zero(),
-                                                           false,
                                                            RenderTaskLocation::Fixed);
         next_task_index.0 += 1;
         let mut alpha_task_stack = Vec::new();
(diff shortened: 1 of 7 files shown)


rlhunt:
r? @glennw 

I've successfully ran servo `./mach test-wpt` with this change.

glennw:
This generally looks good - I just have a couple of questions.

1) The commit message mentions the composite shader but there's no changes related to that from what I can see?
2) Unrelated to this commit - do you know what makes the gradient dithering different when on different render targets?

rlhunt:
1. The important change for the composite shader is the clear color. The clear color being (1,1,1,0) screws up the composite shader because it assumes that the colors it reads are premultiplied, which isn't the case with (1,1,1,0).
2. My guess is that the `gl_FragCoord` used for dithering is different when a gradient is in the framebuffer vs. in a render target (render tasks are allocated positions), but I didn't look too closely.

glennw:
Huh, I definitely changed the color to 0,0,0,0 when I tried to make this optimization, and was still seeing test failures. Perhaps I messed up my build or something.

@bors-servo r+

rlhunt:
That is odd. Hopefully I didn't mess something up either. Let me know if there are any problems updating Servo.

staktrace:
@rlhunt this patch appears to be causing a reftest failure in Gecko for layout/reftests/bugs/1156129-1.html. See https://bugzilla.mozilla.org/show_bug.cgi?id=1358156#c1 and https://bugzilla.mozilla.org/show_bug.cgi?id=1358156#c4

rlhunt:
I'm looking into this now.

PR #1149, opened 2017-04-21
Make image tiling work with external buffer images.

nical:
Tiling is already handled in the resource cache by specifying the proper size, stride and offset in order to upload the correct part of the image (the tile) to the texture atlas. Getting this to work with external buffer images is only a matter of passing the correct offset which was ignored before this commit.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1149)


nical:
cc @JerryShih r? @glennw

kvark:
Looks reasonable. Merging, given that Glenn is on weekend already.
@bors-servo r+

PR #1151, opened 2017-04-21
Update the sample app to demonstrate clipping and scrolling

staktrace:
This adds some stuff to the sample app that demos clipping and scrolling, including nested scrolling. It also adds a paragraph to the CLIPPING.md detailing the behaviour when multiple define_clip calls are made with the same clip id.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1151)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -47,3 +47,6 @@ core-text = "4.0"
 
 [[example]]
 name = "basic"
+
+[[example]]
+name = "scrolling"
--- a/webrender/examples/scrolling.rs
+++ b/webrender/examples/scrolling.rs
@@ -1,0 +1,275 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+extern crate gleam;
+extern crate glutin;
+extern crate webrender;
+extern crate webrender_traits;
+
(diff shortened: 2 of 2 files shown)


staktrace:
FWIW I would really like it if this landed and then was kept up to date by subsequent changes to the clipping API.

kvark:
Apologies for an unpopular opinion. This is great thing to have in a sample, but I feel that it's big enough to deserve its own executable.
@glennw maybe we should migrate the new code to `examples/scrolling` and old code to `examples/basic`, with `webrender` project actually having them specified as examples?

jrmuizel:
> @glennw maybe we should migrate the new code to examples/scrolling and old code to examples/basic, with webrender project actually having them specified as examples?

That sounds like a good idea to me.

staktrace:
That sounds reasonable to me as well. I'll update the PR next week.

glennw:
That sounds like a great idea.

staktrace:
Updated PR to be rebased on top of #1160. This now adds an examples/scrolling.rs example app that demonstrates scrolling. The app is mostly what I had before (the old sample app with my scrolling changes) but now I ripped out all the stuff not related to scrolling as well (like the blob image stuff and touch-gestures stuff). The interesting part of the app is around line 145 in scrolling.rs, inside the if true { ... } block.

glennw:
It probably makes sense to put all the common init code in a shared file for all examples to use, but that can be done once we have a few examples and a better idea of which code can be shared.

@bors-servo r+

PR #1152, opened 2017-04-21
Widen IPC profiling

Gankro:
NOTE: this changes the public API; firefox will need a minor patch
to its IPC code, in addition to a regeneration of its bindings.

We now track the time spent building and consuming display lists,
as future work will shift costs into these areas.

Stats are also better formatted, and handle multiple display lists.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1152)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -158,6 +158,10 @@ impl TimeProfileCounter {
         val
     }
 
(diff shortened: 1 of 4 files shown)


kvark:
Looks very clean to me. Thank you!
@bors-servo r+

PR #1154, opened 2017-04-21
fill box shadow path for inset box shadows

changm:
We have to fill the area between the box bounds and the bs_rect for inset box shadows for cases where we have a large offset.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1154)


kvark:
> ... for cases where we have a large offset

Where do you check for the offset?

Looks fine to me otherwise. The fuzzy test settings look rather worrysome now, and I think it fails on Travis.
FYI, you can check it locally with "./headless.py reftest"

PR #1155, opened 2017-04-21
Remove unused z-index parameter

jrmuizel:
This isn't used anywhere inside WebRender and who knows what the
semantics would be. Servo and Gecko both z-sort before creating
WebRender display lists.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1155)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -258,7 +258,6 @@ fn main() {
     let bounds = LayoutRect::new(LayoutPoint::zero(), LayoutSize::new(width as f32, height as f32));
     builder.push_stacking_context(webrender_traits::ScrollPolicy::Scrollable,
                                   bounds,
-                                  0,
                                   None,
                                   TransformStyle::Flat,
                                   None,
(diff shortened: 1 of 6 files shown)


kvark:
I've also not found any use of it in WR so far.
Leaving for @glennw to have a look.

glennw:
We seem to be hitting some kind of timeout on the mac builder permanently with all the pending PRs. Any suggestions @metajack @larsbergstrom ? Perhaps we could try updating the rustc version?

jrmuizel:
It looks like this pull request succeeded: https://github.com/servo/webrender/pull/1158

larsbergstrom:
Yeah, it looks like there's something going on:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

A 10 minute pause without any output from rustc or it completing seems not unusal, though, considering that even though the build is fast on debug:
```
Finished dev [unoptimized + debuginfo] target(s) in 49.80 secs
```
Looking at a successful build on release, it did take 21.95 minutes:
```
Finished release [optimized + debuginfo] target(s) in 1317.76 secs
```

I suspect that you should either send mail to support@travis-ci.org to ask them to increase the timeout slightly or use one of the wrappers:
https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

That said, ~22 minutes seems like a REALLY long time to compile just webrender. Is there something going on here, maybe? Or a compiler regression?

glennw:
Other PRs have been getting through today, so maybe it was a travis issue? Let's just retry this once it's rebased.

kvark:
Retrying...
@bors-servo r+

PR #1156, opened 2017-04-23
Return straight alpha in ps_composite_fs early returns

rlhunt:
The composite primitive shader can early return in cases where the source
alpha or backdrop alpha are zero. In these cases we need to unpremultiply
the source and backdrop colors because the composite shader uses a
straight alpha blend mode.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1156)


rlhunt:
r? @glennw @kvark 

I used the wrong fuzzy value in the first commit, the PR has been updated.

Issue #1031, opened 2017-03-29
Lots of cpu (~100ms) used in the render backend on this test case


jrmuizel:
https://people-mozilla.org/~jmuizelaar/implementation-tests/dl-test.yaml

This is a test case generated from Gecko running https://people-mozilla.org/~jmuizelaar/implementation-tests/dl-test.html.

The test case is not representative and Gecko still uses stacking contexts a lot more than necessary but it would be good if webrender could handle this better.

jrmuizel:
Make sure to run with the --rebuild option. This matches the original html because the scene is actually changing.

jrmuizel:
https://github.com/servo/webrender/pull/1041 helps with this test case.

jrmuizel:
https://github.com/servo/webrender/pull/1043 will help too.

glennw:
I did some profiling on this today. First, the overall profile counters:

* ~8300 primitives, of which ~5000 are visible.
* ~8300 stacking contexts.

![profile2](https://cloud.githubusercontent.com/assets/2350882/24538491/cd15c78e-162c-11e7-97ef-156053c965aa.png)

Things to note:
* CPU backend time is clearly the problem.
* Compositor time and GPU time are fine.
* Total vertices is ~31k - should be no problem for any GPU these days.
* Total draw calls is 1 - batching works! :smile: 

Looking at a CPU profile of the backend thread:

![profile1](https://cloud.githubusercontent.com/assets/2350882/24538544/2c11e3ee-162d-11e7-80f1-ba27ac85e442.png)

* Both scene and frame build are slow, due to the sheer number of stacking contexts.
* The majority of the time is dealing with the scene - traversing the stacking context tree and extracting primitives.
* ```flatten_stacking_context``` itself is ~18ms in this test case to traverse the ~8300 stacking contexts.

There's a few obvious things we can do to improve the speed of this scene:

1. Don't create as many stacking contexts :grinning: 
1. Improve memory allocation patterns and structure sizes.
 * I would like to experiment with a custom allocator. Many of the structures we use to build a frame follow a very simple allocation and access pattern. I believe this could give us quite large performance wins.
 * Retain more of the allocated structures between frames. This is probably not required if a custom allocator works.
1. Run some of this workload across CPU threads.
[...]

Issue #1029, opened 2017-03-29
Optimize array allocation


kvark:
Profiling on OSX shows that we spend a large chunk of time (54% IIRC) in the routine that doubles the allocated storage for `Vec`. We should make a pass over the code and try to re-use the original storage in as many spaces as possible. Another way to address it is to use `Vec::with_capacity` constructors.

The issue can't be fixed once and for all, so a completion criterion would be:
  - making a pass over all the code, especially the `FrameBuilder::add_primitive` path
  - profiling CPU again and checking how much the time dropped

cc @jrmuizel

glennw:
We have ```Vec``` recycling now, and I haven't seen resizing in the profiles for a long time.

PR #1158, opened 2017-04-24
Recycle some of the vector allocations when re-building the frame.

nical:
Currently, most of the data in FrameBuilder is regenerated from scratch when RenderBackend::build_scene is called. When this happens all of the FrameBuilder data from the previous frame still exists but is thrown away as we re-allocate a new FrameBuilder. In some test cases a lot of time is spent growing vectors when building frames. We can take advantage of the likelihood that the next frame will require the same amount of allocations as the previous frame and reduce the cose of reallocating by recycling vectors from the previous frame. This commit does that for some of the vectors in FrameBuilder and PrimitiveStore which show up high in some profiles.

This commit does not reuse an allocation if it is more than twice the size of the actually used data, so that if a frame allocates an outstanding amount of memory, the peak does not stay allocated forever.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1158)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -316,7 +316,8 @@ impl Frame {
             }
         });
 
-        let mut frame_builder = FrameBuilder::new(window_size,
+        let mut frame_builder = FrameBuilder::new(self.frame_builder.take(),
+                                                  window_size,
                                                   background_color,
                                                   self.frame_builder_config);
 
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -24,7 +24,7 @@ use tiling::StackingContextIndex;
 use tiling::{AuxiliaryListsMap, ClipScrollGroup, ClipScrollGroupIndex, CompositeOps, Frame};
 use tiling::{PackedLayer, PackedLayerIndex, PrimitiveFlags, PrimitiveRunCmd, RenderPass};
 use tiling::{RenderTargetContext, RenderTaskCollection, ScrollbarPrimitive, StackingContext};
-use util::{self, pack_as_float, subtract_rect};
+use util::{self, pack_as_float, subtract_rect, recycle_vec};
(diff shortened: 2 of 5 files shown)


nical:
This helps a bit (~12% improvement) with issue #1031 and is relevant to issue #1029, although the latter advertises using sensible defaults for vector allocations which is a different/complementary approach that could typically help with the first frame.

After this there are still outstanding reallocations when profiling issue #1031. I'll follow up with those.

r? @kvark

webrender/src/frame.rs
```diff
@@ -320,6 +320,10 @@ impl Frame {
                                                   background_color,
                                                   self.frame_builder_config);
 
+        if let Some(ref mut old_builder) = self.frame_builder {
```
kvark: I think it would be cleaner to do something like
```rust
let mut frame_builder = self.frame_builder.take().unwrap_or(FrameBuilder::new(...));
frame_builder.clear(); // resets all the contents, keeps the capacity/allocation
```

nical: I started with something like this and the reason I ended up changing is that if you add members to FrameBuilder it is easy to forget to handle them in clear() which can lead to some subtle misbehavior. With this approach, if you forget to handle the member in recycle you won't have a bug (just potentially miss an optim, which will show up in profiles if it turns out to have any impact).

If you have a strong preference for doing it the other way I can rearrange the patch.

kvark: I see. You have a point. I wish we could do `#[derive(Clear)]` to enforce that without manually tracking the new items.

kvark: Another approach that could be cleaner is to pass `self.frame_builder` as a parameter into `FrameBuilder::new`, which would just take & clear the vectors it needs from the old one. That would make sure no items are left behind (without `recycle_vec`, which feels a bit hacky to me).

PR #1160, opened 2017-04-24
Move the sample app from sample/src/main.rs to webrender/examples/basic.rs

staktrace:
As suggested in #1151 this moves the existing sample app to be an "example" of the webrender library. As it gets built when running `cargo test` on the webrender library, it doesn't need separate build steps in the travis/appveyor CI configs.

I'm not sure why the sample app pulled in webrender_traits `default_features = false` but I dropped that when migrating the Cargo.toml spec, since there doesn't seem to be a way of specifying it just for a particular example.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1160)


kvark:
Yeah, I don't see any default features on `webrender_traits`, so it should be safe to drop.
Looks great to me!
@bors-servo r+

PR #1164, opened 2017-04-25
Add mousewheel support to the scrolling example

mrobinson:
This makes it a bit easier to quickly scrolling the different layers.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1164)

### DIFF
--- a/webrender/examples/scrolling.rs
+++ b/webrender/examples/scrolling.rs
@@ -10,39 +10,9 @@ extern crate webrender_traits;
 use gleam::gl;
 use std::env;
 use std::path::PathBuf;
-use webrender_traits::{ClipId, ClipRegion, ColorF, Epoch};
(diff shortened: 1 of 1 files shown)


staktrace:
Is it possible to read out the scroll offset using the WR API? Because it would be nice to update the `root_scroller_pos` and `nested_scroller_pos` values in the ScrollState when you wheel-scroll, otherwise the keyboard scrolling state gets out of sync.

mrobinson:
@staktrace Okay, I've eliminated the need to sync the scrolling states with the keyboard scrolling in the latest patch. I think this also simplifies this in the way that we discussed via IRC.

staktrace:
LGTM, thanks!

kvark:
Looks clean.
@bors-servo r+

kvark:
Eh, there we go again, Mac bots (also see #1155)

PR #1166, opened 2017-04-25
Update clipping documentation

staktrace:
When I updated the commit in #1151 I forgot to include the CLIPPING.md changes. I still think they're worth keeping though, so here they are.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1166)

### DIFF
--- a/webrender/doc/CLIPPING.md
+++ b/webrender/doc/CLIPPING.md
@@ -82,6 +82,10 @@ let generated_id = define_clip(content_rect, clip, id);
 assert!(id == generated_id);
 ```
 
+Note that calling `define_clip` multiple times with the same `clip_id` value
+results in undefined behaviour, and should be avoided. There is a debug mode
+assertion to catch this.
+
 ## Pending changes
 1. Normalize the way that clipping coordinates are defined. Having them
    specified in two different ways makes for a confusing API. This should be


mrobinson:
I think that clip coordinates should be relative to the origin of the boundaries of the clip. Have you seen it acting differently at some point?

staktrace:
What do you mean by the "boundaries of the clip"?

mrobinson:
@staktrace Sorry, I mean the content_rect of the clip. The clip is defined by the content rectangle and a clip region. The clip region should be relative to the origin of the content rectangle. So for the case of the ClipDisplayItem:
```rust
    pub fn define_clip(&mut self,
                       content_rect: LayoutRect,
                       clip: ClipRegion,
                       id: Option<ClipId>)
```
Here `content_rect` is relative to the parent stacking context, but the coordinates of the `ClipRegion` should be relative to `content_rect`.

staktrace:
Ah, I don't think I tried calling `define_clip` with a nonzero origin in `content_rect`. So it's quite possible that I meant to say the clip would be relative to the `content_rect` which in turn would be relative to the parent stacking context.

The point I wanted to make in the documentation was merely that the last call to `define_clip` for a particular `clip_id` is the one that takes effect, so whatever the parent stacking context (and I guess `content_rect`) is in scope at that point is the one that gets used.

That being said I'm now trying to reproduce my earlier results and am just getting garbage drawn on the screen. So maybe I was doing something erroneously before? Not sure, I'll keep digging. If you agree with the point about the last call to `define_clip` being the one that takes effect, I can update the PR to reflect the relevance of the `content_rect`.

staktrace:
Turns out my results were bogus because I had been running in release mode. In debug mode attempting to call `define_clip` twice with the same `clip_id` trips an assertion. I've updated the documentation to reflect that it is undefined behaviour.

mrobinson:
@staktrace Thanks for the investigation!

PR #1167, opened 2017-04-26
Revert "Use debug assertions to check WebGL errors"

glennw:
This reverts commit 57c5cd430e2360ef4e3fc4473be25bc313979d35.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1167)

### DIFF
--- a/webrender_traits/src/webgl.rs
+++ b/webrender_traits/src/webgl.rs
@@ -612,10 +612,9 @@ impl WebGLCommand {
                 ctx.gl().generate_mipmap(target),
         }
 
-        if cfg!(debug_assertions) {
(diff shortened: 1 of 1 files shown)


glennw:
Unfortunately this breaks a number of tests in the WPT suite - reverting for now so that we can land the WR update.

kvark:
I'm curious, how does checking/non-checking for GL errors ended up affecting the reftests?

glennw:
@kvark I think what probably happens is that the GL error check results in a glFlush call internally, which is affecting the reftests on OSMesa. That's just a guess though - I haven't confirmed.

PR #1168, opened 2017-04-26
Only send scroll state for clips that actually scroll

mrobinson:
This should reduce the amount of traffic across the API boundary as
well as prevent clients from having to track which nodes actually scroll
and which don't.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1168)


webrender/src/clip_scroll_tree.rs
```diff
@@ -109,10 +109,9 @@ impl ClipScrollTree {
     pub fn get_scroll_node_state(&self) -> Vec<ScrollLayerState> {
         let mut result = vec![];
         for (id, node) in self.nodes.iter() {
```
kvark: offtopic: I think this could be neatly rewritten as `iter().flat_map(...).collect()`

Issue #739, opened 2017-01-18
Implement preserve-3d


jrmuizel:
The basic plan is to use a bsp tree to do plane-splitting like Gecko does.

It probably makes the most sense for this work to wait for the work in #742

kvark:
Me and @mrobinson are interested to implement this.

kvark:
Got https://github.com/kvark/plane-split, going to try hooking it up to WR next.

jrmuizel:
The license should probably at least be dual with MPL2

kvark:
Some observations from looking into this:
  1. when "preserve3d" is active for a sub-tree of the document, the elements are supposed to naturally be ordered by depth on a per-pixel level. It can have children that are "flat", not sure if we can support it from the start or implement as a follow-up.
  2. plane splitting is one of the ways to draw intersecting semi-transparent objects, it is viable on the way because most of the things are planes and thus can be trivially split
  3. we could have used the depth buffer for opaque objects, but that conflicts with the current approach. Thus, we need to split all the planes (including opaque ones) and assign the produced ordering to Z indices that are currently obtained from the original primitive order.
  4. when 2 planes intersect, we should only need to split one of them to avoid intersection (not both), and then return the draw order
  5. splitting can be done incrementally: we can have a splitter object that gets mutated each time we add a plane to it, this function returns the split plane and ordering info

glennw:
preserve-3d support has merged. Let's close this and open new issues for any specific bugs / missing features.

PR #1169, opened 2017-04-26
Basic preserve-3d support

kvark:
Goes towards #739 

This is a basic implementation that features a green test case. It can be merged independently, but more work is expected to solidify it.
There is a lot of interaction of "preserve3d" stacking contexts with filters, composite ops, and such, that is not thought-through or tested at the moment. The new code should be totally safe as long as the client continues to use `TransformStyle::Flat` (as both Gecko/Servo still do).

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1169)


glennw:
Reviewed 17 of 22 files at r1.
Review status: 17 of 22 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[wrench/reftests/split/simple.yaml, line 7 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigM62JJP8ExxwLjpx8:-KigM62KdGaLlYmwND4j:bkxdrgh) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/wrench/reftests/split/simple.yaml#L7)):*
> ```YAML
>     - type: stacking-context
>       bounds: [0, 0, 1024, 1024]
>       transform-style: preserve3d
> ```

CSS uses `preserve-3d` - let's use the CSS keyword in YAML files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1169)*

glennw:
Partial review only so far - I'll be working through the remaining files today.

glennw:
Reviewed 5 of 22 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[webrender/res/ps_split_composite.vs.glsl, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigRsmK4jr3pBFH71Ym:-KigRsmL1ukvbE9oE5eo:bke2gps) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/res/ps_split_composite.vs.glsl#L26)):*
> ```GLSL
> 
>     vec2 normalized_pos = mix(
>         mix(geometry.points[0], geometry.points[1], aPosition.x),
> ```

Extracting this into a `bilerp()` function might make the intent clearer.

---

*[webrender/src/frame.rs, line 440 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigSSw7_t2SSkmfQqza:-KigSSw89TerClrH1euv:b-5bh3rm) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame.rs#L440)):*
> ```Rust
>                                               level == 0,
>                                               composition_operations,
>                                               bounds.clone(),
> ```

I think this type might be `Copy`?

---

*[webrender/src/frame_builder.rs, line 85 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigV-jF_OjKBr-8NAth:-KigV-jF_OjKBr-8NAti:buwvmfm) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L85)):*
> ```Rust
> }
> 
> fn make_polygon(sc: &StackingContext, node: &ClipScrollNode, anchor: usize)
> ```

Could you explain how the anchor parameter works?

---

*[webrender/src/frame_builder.rs, line 254 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigSm8uK87lUs5Rr-kZ:-KigSm8uK87lUs5Rr-k_:bz1w48u) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L254)):*
> ```Rust
>                 // is the root of the page
>                 let isolation = &mut self.stacking_context_store[parent_index.0].isolation;
>                 if *isolation != ContextIsolation::None {
> ```

What could cause this - I guess if we have mix-blend-mode on a preserve-3d?

---

*[webrender/src/frame_builder.rs, line 1145 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigTLSB0GZfH_qV1lhP:-KigTLSCKI6afrqXnaZc:b-kcp50y) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L1145)):*
> ```Rust
> 
>                     if stacking_context.isolation == ContextIsolation::Items {
>                         let task_size = DeviceIntSize::from_untyped(&stacking_context.local_bounds.size.ceil().to_i32().to_untyped());
> ```

This doesn't seem right - the local bounds won't take the zoom / transform etc into effect. Shouldn't this be based on the screen bounding rect?

---

*[webrender/src/frame_builder.rs, line 1155 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigU4F8a33vIlMSgBX7:-KigU4F8a33vIlMSgBX8:b-meav3c) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L1155)):*
> ```Rust
>                     }
> 
>                     if composite_count == 0 && stacking_context.isolation == ContextIsolation::Full {
> ```

It might be clearer to combine this if and the one above into a match statement?

---

*[webrender/src/frame_builder.rs, line 1284 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigWEvTH_Cbb5sDDDU6:-KigWEvTH_Cbb5sDDDU7:bv314gx) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L1284)):*
> ```Rust
>                     }
> 
>                     let group_index_opt = match stacking_context.isolation {
> ```

I'm not sure I fully understand this part - does this imply that any primitives in a preserve-3d context get rendered without any transform to the intermediate surface?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1169)*

glennw:
@kvark In general this looks great! It's far fewer lines of code than I expected (although I guess it helps having the plane splitting crate separate). I have a few questions about the PR above that I don't fully understand - it might be easier to discuss on IRC / Vidyo?

glennw:
@kvark Ah, I think I understand a bit better now. So the current implementation renders those intermediate sources untransformed into a target. Then it splits them and composites them, applying the transform when compositing the split planes - is that right?

I think that if we do it this way, we may have quality issues when the transformed planes have an interesting transform on them - since, I think, we'll end up drawing things like text and borders at the un-zoomed quality level into the intermediate target.

If that's right - I wonder if we could do it the inverse way - draw the intermediate surfaces transformed at full zoom / scale, and then composite them in screen space?

I'm not sure how easy / feasible that is, or if I'm completely mis-understanding something - does that make any sense? :)

glennw:
@kvark Also, even if the above is correct - it might still make sense to merge this now, as a step towards that?

kvark:
Reviewed 1 of 5 files at r2.
Review status: 19 of 23 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[webrender/res/ps_split_composite.vs.glsl, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigRsmK4jr3pBFH71Ym:-KigwEoGQFXxt1UpCULs:b-gki9g1) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/res/ps_split_composite.vs.glsl#L26)):*

Extracting this into a `bilerp()` function might make the intent clearer.
</blockquote></details>

thanks, fixed!

---

*[webrender/src/frame.rs, line 440 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigSSw7_t2SSkmfQqza:-KigwJtQCXjtaoyVw_jK:b-7ql1td) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame.rs#L440)):*

I think this type might be `Copy`?
</blockquote></details>

fixed

---

*[webrender/src/frame_builder.rs, line 85 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigV-jF_OjKBr-8NAth:-KigwLm_fQreiIJp1v-u:b-4mhpgl) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L85)):*

Could you explain how the anchor parameter works?
</blockquote></details>

added a comment

---

*[webrender/src/frame_builder.rs, line 254 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigSm8uK87lUs5Rr-kZ:-KigwNsBNueMJyLQVlZj:be3max7) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L254)):*

What could cause this - I guess if we have mix-blend-mode on a preserve-3d?
</blockquote></details>

yes. These interactions are not supported nicely yet

---

*[webrender/src/frame_builder.rs, line 1145 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigTLSB0GZfH_qV1lhP:-KigwU4LzgRX1YO3ygHN:b-ppnzyp) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L1145)):*

This doesn't seem right - the local bounds won't take the zoom / transform etc into effect. Shouldn't this be based on the screen bounding rect?
</blockquote></details>

I believe you answered this yourself later on: I draw primitives without transforms, then transform the layer when compositing.

---

*[webrender/src/frame_builder.rs, line 1155 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigU4F8a33vIlMSgBX7:-Kigwb4TYynLr7BnTOoH:b-lf14hs) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L1155)):*

It might be clearer to combine this if and the one above into a match statement?
</blockquote></details>

good idea! fixed

---

*[webrender/src/frame_builder.rs, line 1284 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigWEvTH_Cbb5sDDDU6:-KigwdMPX72RREcLm0FV:b2nign6) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/webrender/src/frame_builder.rs#L1284)):*

I'm not sure I fully understand this part - does this imply that any primitives in a preserve-3d context get rendered without any transform to the intermediate surface?
</blockquote></details>

yes

---

*[wrench/reftests/split/simple.yaml, line 7 at r1](https://reviewable.io:443/reviews/servo/webrender/1169#-KigM62JJP8ExxwLjpx8:-KigwecAuKgW12Ymiq85:bbb2fs3) ([raw file](https://github.com/servo/webrender/blob/c85dbb05077c8b873361165131ddcf006053a3ea/wrench/reftests/split/simple.yaml#L7)):*

CSS uses `preserve-3d` - let's use the CSS keyword in YAML files.
</blockquote></details>

agreed, fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1169)*

glennw:
Nice! The change to bake in screen space actually simplifies quite a bit of the code (conceptually at least), I think.

It's perhaps worth a couple of comments describing how the ```has_primitives_stack``` code works - I think I understand it but it would be good to clarify.

Apart from that, this looks good to me once we get the reftest issue resolved!

webrender/src/render_task.rs
```diff
@@ -51,9 +52,10 @@ pub enum RenderTaskLocation {
 
 #[derive(Debug, Clone)]
 pub enum AlphaRenderItem {
-    Primitive(ClipScrollGroupIndex, PrimitiveIndex, i32),
+    Primitive(Option<ClipScrollGroupIndex>, PrimitiveIndex, i32),
```
glennw: This ```Option``` can be removed now.

glennw:
@kvark One minor change above - the ```Option``` in the AlphaRenderItem is no longer required, now that we bake in screen space. Once that is done and CI is happy, r=me - let's get this merged!

kvark:
I kinda left this by intention - an ability to draw without transformation may be handy one day, and there is next to no maintenance cost to this feature.
I don't mind cutting it out, just making a comment :)

glennw:
That's fine - if it's intentional let's leave it :)

PR #1171, opened 2017-04-27
Handle RadialGradients with zero sized radius

rlhunt:
A RadialGradient with radius of zero should only display the final color
stop value.

Currently, a radius of zero will cause ratio_xy to be NaN, Inf, or -Inf and
the output will be blank. Forcing the radius to zero and ratio_xy to one
won't work because ps_radial_gradient can't handle a zero sized radius. So
instead we can construct an equivalent gradient with only the last color stop.

Alternatively this could be caught on the backend and the RadialGradient
could be converted to a color rect, but that would disable dithering.
This is a simple enough fix and makes reftests consistent.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1171)

### DIFF
--- a/webrender_traits/src/display_list.rs
+++ b/webrender_traits/src/display_list.rs
@@ -334,6 +334,33 @@ impl DisplayListBuilder {
                                   radius: LayoutSize,
                                   mut stops: Vec<GradientStop>,
                                   extend_mode: ExtendMode) -> RadialGradient {
+        if radius.width <= 0.0 || radius.height <= 0.0 {
+            // The shader cannot handle a non positive radius. So
+            // reuse the stops vector and construct an equivalent
+            // gradient.
+            let last_color = stops.last().unwrap().color;
+
+            stops.clear();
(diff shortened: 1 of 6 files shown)


nox:
Ain't that contradictory with https://drafts.csswg.org/css-images/#degenerate-radials?

rlhunt:
It sounds like that's the case, but it's the behavior I see in the Gecko reftests.

[test](https://hg.mozilla.org/mozilla-central/file/tip/layout/reftests/css-gradients/radial-zero-length-1a.html)
[reference](https://hg.mozilla.org/mozilla-central/file/tip/layout/reftests/css-gradients/radial-zero-length-1-ref.html)

I'd like to know more, because I was confused by the spec on that too.

nox:
Please file a spec issue if it is wrong.

rlhunt:
I'll investigate the spec more as a follow up. This PR is at the least better behavior than before and should help us pass more reftests.

r? @glennw @kvark

PR #1172, opened 2017-04-27
Add a feature for dithering

rlhunt:
In order to help pass all the reftests for enabling gradients by default in
Gecko, it would be nice to be able to disable dithering. Canvas2D gradients
and border gradients in Gecko don't use dithering, and that causes some
issues. Enabling dithering can be investigated as a follow up.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1172)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -820,6 +820,7 @@ float do_clip() {
         all(inside) ? textureLod(sCacheA8, vClipMaskUv, 0.0).r : 0.0;
 }
 
+#ifdef WR_FEATURE_DITHERING
 vec4 dither(vec4 color) {
     const int matrix_mask = 7;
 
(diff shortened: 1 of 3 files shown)


webrender/res/prim_shared.glsl
```diff
@@ -837,4 +838,6 @@ vec4 dither(vec4 color) {
 
     return color + vec4(noise, noise, noise, 0);
 }
+#endif
```
kvark: I believe all you need on the shader side is
```glsl
#else
vec4 dither(vec4 color) { return color; }
```

rlhunt: Yes that makes sense, I've updated the PR.

kvark:
Thanks!
@bors-servo r+

glennw:
This needs a rebase, then should be ready to go.

rlhunt:
Rebased.

PR #1173, opened 2017-04-27
Blob image renderer tiling and partial updates

nical:
I changed the BlobImageRenderer API so that it can handle tiled images: the request key now take the tile offset into account in addition to the image key, and an offset in image space is passed in the BlobImageDescriptor.

In the gecko integration we want to be able to send partial updates of the blob, which means that the BlobImageRenderer has to be responsible for managing and storing the blobs since the resource_cache does not know how to apply the updates.

The blob renderer example was moved to its own file and improved a bit.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1173)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -49,4 +49,7 @@ core-text = "4.0"
 name = "basic"
 
 [[example]]
+name = "blob"
+
+[[example]]
 name = "scrolling"
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -17,11 +17,10 @@ use std::env;
 use std::fs::File;
 use std::io::Read;
 use std::path::PathBuf;
-use webrender_traits::{BlobImageData, BlobImageDescriptor, BlobImageError, BlobImageRenderer, BlobImageRequest};
-use webrender_traits::{BlobImageResult, ClipRegion, ColorF, Epoch, GlyphInstance};
-use webrender_traits::{DeviceIntPoint, DeviceUintSize, DeviceUintRect, LayoutPoint, LayoutRect, LayoutSize};
-use webrender_traits::{ImageData, ImageDescriptor, ImageFormat, ImageStore, ImageRendering, ImageKey, TileSize};
-use webrender_traits::{PipelineId, RasterizedBlobImage, TransformStyle, BoxShadowClipMode};
+use webrender_traits::{ColorF, Epoch, GlyphInstance};
+use webrender_traits::{DeviceIntPoint, DeviceUintSize, LayoutPoint, LayoutRect, LayoutSize};
+use webrender_traits::{ImageData, ImageDescriptor, ImageFormat};
+use webrender_traits::{PipelineId, TransformStyle, BoxShadowClipMode};
 
 #[derive(Debug)]
 enum Gesture {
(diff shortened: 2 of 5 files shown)


kvark:
Looks good to me (minus one small thing), but I'd like to have a second look tomorrow (unless we get another pair of eyes here ;) ).

---

Reviewed 5 of 5 files at r1.

---

*[webrender/src/resource_cache.rs, line 1054 at r1](https://reviewable.io:443/reviews/servo/webrender/1173#-KilS0JVyjFWW1eoFDcy:-KilS0JWoM31PQYvlIj3:b-shhh6j) ([raw file](https://github.com/servo/webrender/blob/6fcc23f4c13a5a9c9ed9d2fafca820267ff635b5/webrender/src/resource_cache.rs#L1054)):*
> ```Rust
> 
> // Compute the width and height of a tile depending on its position in the image.
> pub fn compute_tile_size(descriptor: &ImageDescriptor,
> ```

I see it confusing. `compute_tile_size` gets `tile_size` as an argument. One of them should have a different name and/or type.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1173#-:-KilSIB91miPTvv0Gesn:b-jdjw4o)*

nical:
*[webrender/src/resource_cache.rs, line 1054 at r1](https://reviewable.io:443/reviews/servo/webrender/1173#-KilS0JVyjFWW1eoFDcy:-KilTsil3-UP70EK6xNC:bn9rkgi) ([raw file](https://github.com/servo/webrender/blob/6fcc23f4c13a5a9c9ed9d2fafca820267ff635b5/webrender/src/resource_cache.rs#L1054)):*

I see it confusing. `compute_tile_size` gets `tile_size` as an argument. One of them should have a different name and/or type.
</blockquote></details>

Yeah that's bugging me too, I am thinking of renaming the tile_size argument into base_tile_size. It is the tile size (integer) set for an image as opposed to the (Size2D) tile-width-and-height-size computed by this function that may be smaller if the tile is on the edge of the image.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1173)*

glennw:
Reviewed 5 of 5 files at r1.

---

*[webrender_traits/src/image.rs, line 114 at r1](https://reviewable.io:443/reviews/servo/webrender/1173#-Kim2XUf5HDiX_8jg1Nm:-Kim2XUf5HDiX_8jg1Nn:b-ilsid3) ([raw file](https://github.com/servo/webrender/blob/6fcc23f4c13a5a9c9ed9d2fafca820267ff635b5/webrender_traits/src/image.rs#L114)):*
> ```Rust
>     #[inline]
>     pub fn is_blob(&self) -> bool {
>         if let &ImageData::Blob(_) = self { true } else { false }
> ```

nit: might be tidier as a `match {}`, but no big deal either way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1173)*

glennw:
I've had a quick look over this - it looks OK to me too. Let me know if you want to fix up those nits now or merge and fix as a follow up. r=me.

nical:
Addressed the review nits (one day you guys gotta tell me what's not to like about ```if let```), you can merge if you don't have other changes to request.

kvark:
Thanks! 🚢 it
@bors-servo r+

PR #1174, opened 2017-04-27
make JSON frame writer use proper payload API

Gankro:
It's been broken for a while because the format changed; this should prevent that from ever happening again (without causing compiler errors).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1174)


kvark:
Sweet!
@bors-servo r+

PR #1175, opened 2017-04-28
Support dashed borders with new border path.

glennw:
Dashed border edges are handled by the standard ps_border_edge shader.

Dashed border corners instead create a clip mask for the border and
draw each dash stroke to the clip mask. The math for determining the correct
dash strokes is too complex to evaluate efficiently in a shader, doing
it this way allows us to support arbitrary kinds of dash strokes.

In the future, we will either specialize this to a border render task (to
avoid the clip mask target allocation for the whole border rect), or make
use of upcoming optimization work in the clip mask code to handle this
case more efficiently.

This will also serve as the basis for dotted border corners, which are also
too complex to evaluate inside a shader.

For now, we just treat everything as an ellipse when evaluating the dash
strokes. This gives the correct result for ellipses and circle radii,
but is very inefficient for circles. As a follow up, we will include
a fast path for corners where the x/y radii are equal.

Deciding the exact dash length etc is not well specified. The current
code is a rough approximation to what Gecko does. In the future we
can iterate on this and make it closer to Gecko's rules for dash length.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1175)


glennw:
Example of what the clip mask looks like for border dashing:

dashing

glennw:
r? @kvark 

Sorry for another large patch - hopefully the remaining border work should be much smaller patches after this one :)

leeoniya:
for posterity: https://bugzilla.mozilla.org/show_bug.cgi?id=382721

glennw:
@leeoniya Thanks! This patch will definitely need a lot more work to get the exact results and edges cases that Gecko handles - but it puts the foundation in place :)

kvark:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[webrender/res/cs_clip_border.fs.glsl, line 1 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipUVAYYjzxN-ICtNcw:-KipUVAZ6S66ZJwrUPNh:b-3nyh4b) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.fs.glsl#L1)):*
> ```GLSL
> /* This Source Code Form is subject to the terms of the Mozilla Public
> ```

let's add `#line 1` here

---

*[webrender/res/cs_clip_border.fs.glsl, line 25 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipVjGC0YuOEZCUcDtV:-KipVjGC0YuOEZCUcDtW:bkqhw4u) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.fs.glsl#L25)):*
> ```GLSL
>     // Apply AA over half a device pixel for the clip.
>     float d = smoothstep(-0.5 * afwidth,
>                          0.5 * afwidth,
> ```

I don't understand this part. So `afwidth` is an average magnitude of the local position delta per pixel. The maximum `d` would then be `0.5 * afwidth`. So if we have local positions matches with the screen (thus, `afwidth==1`), it's only going to write 0.5 there at max?

---

*[webrender/res/cs_clip_border.fs.glsl, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipVUn0Qa4WB2s0fxOt:-KipVUn0Qa4WB2s0fxOu:b-cy4zjz) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.fs.glsl#L26)):*
> ```GLSL
>     float d = smoothstep(-0.5 * afwidth,
>                          0.5 * afwidth,
>                          max(d0, -d1));
> ```

shouldn't we use the minimum of clip distances instead of the maximum?

---

*[webrender/res/cs_clip_border.vs.glsl, line 16 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipWJv8VW3XUrzLJ7AX:-KipWJv9FG5Cb8jNJJPf:b2wyjrb) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L16)):*
> ```GLSL
>     vec4 data[2] = fetch_data_2(index);
>     return BorderCorner(RectWithSize(data[0].xy, data[0].zw),
>                         vec2(data[1].xy),
> ```

nit: `vec2()` not needed

---

*[webrender/res/cs_clip_border.vs.glsl, line 49 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipWvcoKsVJjlDsqDx0:-KipWvcpmezBWy6KCdVO:bjqfzcy) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L49)):*
> ```GLSL
>     // TODO(gw): We could reduce the number of pixels written here
>     // by calculating a tight fitting bounding box of the dash itself.
>     vec2 pos = mix(corner.rect.p0,
> ```

nit: I think doing `vec2 pos = corner.rect.p0 + aPosition.xy * corner.rect.size;` is both shorter and faster, even if it doesn't matter here

---

*[webrender/res/cs_clip_border.vs.glsl, line 66 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipXoDWro4HfsljovZn:-KipXoDWro4HfsljovZo:b-uv2tkv) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L66)):*
> ```GLSL
> 
>     // Calculate the local space position for this vertex.
>     vec4 layer_pos = get_layer_pos(world_pos.xy, layer);
> ```

why is this `pos` different from the previous declaration of `pos`? we didn't do anything during the transformation to screen space

---

*[webrender/res/cs_clip_border.vs.glsl, line 66 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipXyJHKcC2noHS6tO-:-KipXyJHKcC2noHS6tO0:b-9e2cyx) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L66)):*
> ```GLSL
> 
>     // Calculate the local space position for this vertex.
>     vec4 layer_pos = get_layer_pos(world_pos.xy, layer);
> ```

in a similar call to `get_layer_pos` from `write_transform_vertex`, we are also dividing by `uDevicePixelRatio`. Is it missing here?

---

*[webrender/res/ps_border_edge.vs.glsl, line 69 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-Kip_GAmnW56iybJEXpU:-Kip_GAnExaL1qJLTtzA:b-r8l64h) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/ps_border_edge.vs.glsl#L69)):*
> ```GLSL
>         case BORDER_STYLE_DASHED: {
>             float desired_dash_length = border_width * 3.0;
>             float dash_count = ceil(0.5 * edge_length / desired_dash_length);
> ```

what's the reason for `0.5` here?

---

*[webrender/src/border.rs, line 98 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipbEcYz8OdjpCKHhiV:-KipbEcZQA1nJkX8ZH7i:b-totqqq) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/src/border.rs#L98)):*
> ```Rust
>             (BorderStyle::Dashed, BorderStyle::Dashed) => {
>                 let size = LayerSize::new(width0.max(radius.width), width1.max(radius.height));
>                 let (origin, clip_center, sign_modifier) = match corner {
> ```

it appears that you can produce only `sign_modifier` by the `match` and derive everything else from it computationally

---

*[webrender/src/ellipse.rs, line 30 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-Kiplmeax9iWWGJCVLie:-Kiplmeax9iWWGJCVLif:by6hs6s) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/src/ellipse.rs#L30)):*
> ```Rust
>     // arc length of an ellipse segment. We can probably use
>     // a faster / more accurate method!
>     fn get_simpson_length(&self, theta: f32) -> f32 {
> ```

Heavy math methods like this ask for a few unit tests to be sane.
A small benchmark would be extra bonus.
That's for the later work, of course ;)

---

*[webrender/src/renderer.rs, line 500 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipnjkccStG_U4G-hee:-KipnjkccStG_U4G-hef:b-6wzheu) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/src/renderer.rs#L500)):*
> ```Rust
>     cs_clip_rectangle: LazilyCompiledShader,
>     cs_clip_image: LazilyCompiledShader,
>     cs_clip_border: LazilyCompiledShader,
> ```

it's great to see the clip mechanics helping out in seemingly unrelated areas ;) :+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1175)*

glennw:
Review status: 13 of 18 files reviewed at latest revision, 11 unresolved discussions.

---

*[webrender/res/cs_clip_border.fs.glsl, line 1 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipUVAYYjzxN-ICtNcw:-Kj4qj0EM6j2O86enHSp:b-896fix) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.fs.glsl#L1)):*

let's add `#line 1` here
</blockquote></details>

Done.

---

*[webrender/res/cs_clip_border.fs.glsl, line 25 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipVjGC0YuOEZCUcDtV:-Kj4qlBdFuzQ_m6YppKA:b-5b17ex) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.fs.glsl#L25)):*

I don't understand this part. So `afwidth` is an average magnitude of the local position delta per pixel. The maximum `d` would then be `0.5 * afwidth`. So if we have local positions matches with the screen (thus, `afwidth==1`), it's only going to write 0.5 there at max?
</blockquote></details>

For now, I have just used the existing AA step code in other shaders. I intend to work on the AA quality of the borders (and other clip shaders) once the main border work is done.

---

*[webrender/res/cs_clip_border.fs.glsl, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipVUn0Qa4WB2s0fxOt:-Kj4rQZSfi7DTiwj8D6Q:b8962c7) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.fs.glsl#L26)):*

shouldn't we use the minimum of clip distances instead of the maximum?
</blockquote></details>

The distances are different signs inside the line, so this is a signed distance field subtract operation.

---

*[webrender/res/cs_clip_border.vs.glsl, line 16 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipWJv8VW3XUrzLJ7AX:-Kj4rX3ijXAOdEwnkKQf:b-896fix) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L16)):*

nit: `vec2()` not needed
</blockquote></details>

Done.

---

*[webrender/res/cs_clip_border.vs.glsl, line 49 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipWvcoKsVJjlDsqDx0:-Kj4rXkpbBtXMQTLS898:b-896fix) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L49)):*

nit: I think doing `vec2 pos = corner.rect.p0 + aPosition.xy * corner.rect.size;` is both shorter and faster, even if it doesn't matter here
</blockquote></details>

Done.

---

*[webrender/res/cs_clip_border.vs.glsl, line 66 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipXoDWro4HfsljovZn:-Kj4rYMS-NsZbSMqg2AS:b-4vu8vn) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L66)):*

why is this `pos` different from the previous declaration of `pos`? we didn't do anything during the transformation to screen space
</blockquote></details>

In the case of a 3d transform, this ensures we get the right Z value to do the correct interpolation in the fragment shader.

---

*[webrender/res/cs_clip_border.vs.glsl, line 66 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipXyJHKcC2noHS6tO-:-Kj4ri-77nRt4tWbxTWs:b-hqasn) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/cs_clip_border.vs.glsl#L66)):*

in a similar call to `get_layer_pos` from `write_transform_vertex`, we are also dividing by `uDevicePixelRatio`. Is it missing here?
</blockquote></details>

In ```write_transform_vertex``` we don't calculate a separate ```world_pos``` so need to scale it, but we have that available directly in this shader.

---

*[webrender/res/ps_border_edge.vs.glsl, line 69 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-Kip_GAmnW56iybJEXpU:-Kj4sI3BAlxjXMmRWDdw:b-1vqvq1) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/res/ps_border_edge.vs.glsl#L69)):*

what's the reason for `0.5` here?
</blockquote></details>

We want the length of a dash, which is half the length of an on/off segment. Added a comment.

---

*[webrender/src/border.rs, line 98 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipbEcYz8OdjpCKHhiV:-Kj4sRPdxCqvfYxDLZvc:boivf8o) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/src/border.rs#L98)):*

it appears that you can produce only `sign_modifier` by the `match` and derive everything else from it computationally
</blockquote></details>

I'm not sure how you can do that - could you explain?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1175)*

kvark:
:lgtm:

---

Reviewed 5 of 5 files at r2.

---

*[webrender/src/border.rs, line 98 at r1](https://reviewable.io:443/reviews/servo/webrender/1175#-KipbEcYz8OdjpCKHhiV:-Kj4uy-1Erts-efGcMpi:bflk9ih) ([raw file](https://github.com/servo/webrender/blob/2ef87ae469e96eedb9bb4d847d126327911ed680/webrender/src/border.rs#L98)):*

I'm not sure how you can do that - could you explain?
</blockquote></details>

Sure:
```rust
let sign_modifier = match corner {...};
let origin = /*some math on sign_modifier*/;
let clip_center = /*some math on sign_modifier*/;
```
That should be shorter, but not necessarily clearer, so I'm not pushing for it :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1175#-:-Kj4whTZbrLIWlELTVnh:bnfp4nl)*

PR #1177, opened 2017-04-28
Combined clip rect workaround

kvark:
There is a problem with `combined_local_viewport_rect` that makes it too tight for transformed layers. This PR attempts to work around this, not a proper fix. Having it would make the new reftest of #1169 to pass.

r? @mrobinson

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1177)


mrobinson:
Thanks for the workaround. I hope to fix this for real in the next couple days.

PR #1180, opened 2017-04-28
Cleanup angle, radial gradient shaders

rlhunt:
There is some code that should be shared between the shaders if possible.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1180)


kvark:
Nice one!
@bors-servo r+

PR #1185, opened 2017-05-01
An option to force no batching

kvark:
The new option allows inspecting intermediate rendering results easier with `apitrace` and other GPU debugging tools.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1185)


kvark:
There is a better way to do it, closing.

PR #1189, opened 2017-05-02
Handle gradients with multiple color stops at 0.0 or 1.0

rlhunt:
When a clamped gradient has multiple color stops at 0.0, the last one
will fill GradientData[0]. When a clamped gradient has multiple color
stops at 1.0, the first one will fill GradientData[N-1]. Both of these
conditions will cause the area outside of the gradient to not be filled
with the correct clamped color, which is the first or last gradient
color stop.

Filling the first or last color entry of GradientData with the first or
last color stop isn't exactly correct either because the first or last
color stop entry is only needed in this case for the range outside of
[0, 1].

This commit changes the GradientData so that the first color entry is
used for offset < 0, the last color entry is used for offset > 1, and
the color entries in between are for the range in between. The first and
last color entries are then filled with the correct colors for those
ranges.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1189)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -299,8 +299,20 @@ pub struct RadialGradientPrimitiveCpu {
     pub cache_dirty: bool,
 }
 
-// The number of entries in a gradient data table.
-pub const GRADIENT_DATA_RESOLUTION: usize = 128;
+// The gradient entry index for the first color stop
+pub const GRADIENT_DATA_FIRST_STOP: usize = 0;
+// The gradient entry index for the last color stop
+pub const GRADIENT_DATA_LAST_STOP: usize = GRADIENT_DATA_SIZE - 1;
+
+// The start of the gradient data table
+pub const GRADIENT_DATA_TABLE_BEGIN: usize = GRADIENT_DATA_FIRST_STOP + 1;
+// The exclusive bound of the gradient data table
+pub const GRADIENT_DATA_TABLE_END: usize = GRADIENT_DATA_LAST_STOP;
+// The number of entries in the gradient data table.
+pub const GRADIENT_DATA_TABLE_SIZE: usize = 128;
+
+// The number of entries in a gradient data: GRADIENT_DATA_TABLE_SIZE + first stop entry + last stop entry
+pub const GRADIENT_DATA_SIZE: usize = GRADIENT_DATA_TABLE_SIZE + 2;
 
 #[derive(Debug, Clone, Copy)]
 #[repr(C)]
@@ -317,10 +329,12 @@ pub struct GradientDataEntry {
 // the offset within that entry bucket is used to interpolate between the two colors in that entry.
 // This layout preserves hard stops, as the end color for a given entry can differ from the start
 // color for the following entry, despite them being adjacent. Colors are stored within in BGRA8
-// format for texture upload.
+// format for texture upload. This table requires the gradient color stops to be normalized to the
+// range [0, 1]. The first and last entries hold the first and last color stop colors respectively,
+// while the entries in between hold the interpolated color stop values for the range [0, 1].
 pub struct GradientData {
-    pub colors_high: [GradientDataEntry; GRADIENT_DATA_RESOLUTION],
-    pub colors_low: [GradientDataEntry; GRADIENT_DATA_RESOLUTION],
+    pub colors_high: [GradientDataEntry; GRADIENT_DATA_SIZE],
+    pub colors_low: [GradientDataEntry; GRADIENT_DATA_SIZE],
 }
 
 impl Default for GradientData {
@@ -342,7 +356,8 @@ impl Clone for GradientData {
 }
 
 impl GradientData {
-    // Generate a color ramp between the start and end indexes from a start color to an end color.
+    /// Generate a color ramp filling the indices in [start_idx, end_idx) and interpolating
+    /// from start_color to end_color.
     fn fill_colors(&mut self, start_idx: usize, end_idx: usize, start_color: &ColorF, end_color: &ColorF) {
         // Calculate the color difference for individual steps in the ramp.
         let inv_steps = 1.0 / (end_idx - start_idx) as f32;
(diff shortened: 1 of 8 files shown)


rlhunt:
r? @kvark @glennw 

This PR makes changes to GradientData so it collides with #1181 . I can rebase on that if it will be merged soon, or help rebase #1181 on this.

I hope this PR is clear, I got confused a few too many times by indices conversions and whatnot.

Gankro:
I could split my gradient changes out to a different pr?

rlhunt:
If that's feasible that would work, otherwise this can wait.

webrender/src/prim_store.rs
```diff
@@ -393,6 +400,10 @@ impl GradientData {
             ColorF::new(0.0, 0.0, 0.0, 0.0)
         };
 
+        // Fill in the first gradient entry with the first color stop
+        self.fill_colors(cur_idx, cur_idx + 1, &cur_color, &cur_color);
```
kvark: this is a bit confusing. Perhaps, do `fill_colors(0, 1, ...)` followed by `cur_idx = 1;`?

webrender/src/prim_store.rs
```diff
@@ -381,7 +385,10 @@ impl GradientData {
     // Compute an entry index based on a gradient stop offset.
     #[inline]
     fn get_index(offset: f32) -> usize {
-        (offset.max(0.0).min(1.0) * GRADIENT_DATA_RESOLUTION as f32).round() as usize
+        let clamped = offset.max(0.0).min(1.0);
+        let scaled = clamped * (GRADIENT_DATA_LAST - GRADIENT_DATA_FIRST) as f32 + GRADIENT_DATA_FIRST as f32;
```
kvark: So this function returns indices between the `GRADIENT_DATA_FIRST` and `GRADIENT_DATA_LAST`, inclusive. It seems a little inconsistent that we have a border element to the left of `FIRST` (under index 0), but no border to the right of `LAST`.

rlhunt:
Rebased this and added some comments around GradientData, it needed some. Whew, lots of fun exclusive and inclusive ranges.

webrender/src/prim_store.rs
rlhunt: Yes, I did something similar when I rebased.

webrender/src/prim_store.rs
rlhunt: Yes it returns from [GRADIENT_DATA_FIRST, GRADIENT_DATA_END] but the fill_colors code will only fill from [start_idx, next_idx). So we should have a border element before GRADIENT_DATA_FIRST, and starting at GRADIENT_DATA_END. I've added some comments, and they might make things clearer. Or less clear.

webrender/src/prim_store.rs
```diff
@@ -395,8 +408,13 @@ impl GradientData {
         debug_assert_eq!(first.offset, 0.0);
 
         if reverse_stops {
-            // If the gradient is reversed, then we invert offsets and draw right-to-left
-            let mut cur_idx = MAX_IDX;
+            // Fill in the first entry (for reversed stops) with the first color stop
+            self.fill_colors(GRADIENT_DATA_LAST_STOP, GRADIENT_DATA_LAST_STOP + 1, &cur_color, &cur_color);
+
+            // Fill in the center of the gradient table, generating a color ramp between each consecutive pair
+            // of gradient stops. Each iteration of a loop will fill the indices in [cur_idx, next_idx). The
```
Gankro: Should this be `(next_idx, cur_idx]`?

rlhunt: Yes I forgot to reverse that comment, it should be [next_idx, cur_idx)

rlhunt:
I've rebased and updated the comment. I just ran the gecko reftests with this patch and noticed that there are a couple of failures fixed but some new failures. So I'd like to investigate those first now.

kvark:
Thanks @rlhunt ! I marked the PR as WIP for now.

rlhunt:
The issue was that I took 2 of the 128 gradient data entries (leaving 126 left), and it looks like some reftests are sensitive to the resolution of the table. I tried 256 (254 for the table, 2 for the border) and that fixed a few but not all, so I've settled with doing 130 (128 for the table, 2 for the border) and that seems to work well.

glennw:
@rlhunt Have you run this against the Servo reftests? If not, let me know and I'll run them locally here.

kvark:
> I tried 256 (254 for the table, 2 for the border) and that fixed a few but not all

That doesn't sound good

rlhunt:
I'll run the Servo reftests now. I've tried this on the Gecko reftests and it doesn't cause any regressions and fixes a few tests.

> > I tried 256 (254 for the table, 2 for the border) and that fixed a few but not all
>
> That doesn't sound good

I can confirm that without this PR, if you change the table size to be 254 from 128, these tests fail. So I suppose these tests are sensitive to the inner table of GradientData size being a power of two.

kvark:
@rlhunt do you see why increasing the table size could make the quality worse?

rlhunt:
Increasing the table size does actually seem to improve the quality, it's just seems to be a problem with non power of two sizes (or something else). 126 and 254 causes failures, 128 and 256 seem fine.

I ran some experiments [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1352654#c5), noticeable increasing the inner table to 1024 actually resolved a test, so quality does improve, but I think 1024 is overkill.

rlhunt:
I've successfully ran the servo wpt-tests and css-tests with this change. I think it's interesting that the inner table size is very sensitive and we should do more follow up work with it, but it's a preexisting condition unrelated to this PR.

pyfisch:
It seems like I duplicated some of the work done here on the servo side. :disappointed: 

servo/servo#16666 also contains a function to handle first and last stops and it is called `fix_gradient_stops`. If this is merged please make sure this passes [testcases 2, 11, 12 and 14](https://pyfisch.org/stuff/testcases-gradients.html).

glennw:
@rlhunt Is this ready for review / merge based on your comments above about running the Servo tests? Or is there more testing required based on the comment from @pyfisch above?

rlhunt:
@glennw This should be ready for review / merge. I haven't checked whether it passes those Servo tests, but it passes more Gecko tests without regressions so I'd prefer a followup investigation.

rlhunt:
@pyfisch I ran your test case with a Servo build with this patch applied and it seems like it passes all the test cases (except for gradients with transparency which is related to #1219).

The only iffy test case is 12, where it seems that Gecko/Blink are rendering it incorrectly according to the spec and we differ with this patch.

glennw:
@bors-servo r+

Thanks! Sorry for the delay in reviewing this one.

PR #1191, opened 2017-05-02
Refactor gradient building to not assume DoubleEndedIterator

Gankro:
Future changes to how auxiliary data is stored will break this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1191)


kvark:
Apparently, @rlhunt is good with this PR too.
@bors-servo r+

PR #1192, opened 2017-05-02
Refactored vLocalRect into vLocalBounds

### LINKED ISSUE #1157: Using varying structs can cause shader errors
I get shader errors on my windows machine when using structs as in/out between VS and FS shaders.

We replaced one of the instances of this in #1136 , which fixed my startup crash. But I still get errors for other shaders, usually `ps_border` with `flat varying RectWithSize vBorderRect`.

This is happening to me with WebRender on ANGLE, I believe it happens to others on OpenGL. 

It would suck to go back to using untyped vec4's for all these cases, so it'd be nice to find a workaround. There might not be one and we might have to revert parts of c31376b.

kvark:
Fixes #1157

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1192)

### DIFF
--- a/webrender/res/clip_shared.glsl
+++ b/webrender/res/clip_shared.glsl
@@ -88,6 +88,8 @@ ClipVertexInfo write_clip_tile_vertex(RectWithSize local_clip_rect,
 
     gl_Position = uTransform * vec4(vertex_pos, 0.0, 1);
 
+    vLocalBounds = vec4(clipped_local_rect.p0, clipped_local_rect.p0 + clipped_local_rect.size);
+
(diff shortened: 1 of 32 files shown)


kvark:
On a side note, I think just using a proper naming convention (*Rect vs *Bounds) is getting most of the benefits that the type system would do for us.

glennw:
Looks good, thanks.

@bors-servo r+

jdm:
We have reports from Servo users that the symptom from #1157 remains.

kvark:
@jdm I'm not reopening #1157 yet, since there is a suspicion that it's a different problem, I requested more info on #16393

PR #1193, opened 2017-05-02
Yuv interleaved format

JerryShih:
If we use OpenGL 3.1 context at mac, there is no GL_APPLE_ycbcr_422 ext[1]. We should turn to use GL_APPLE_rgb_422 ext[2]. We could just get the rgb data from the APPLE_ycbcr_422 format because it will apply the Rec601 color space conversion automatically. The GL_APPLE_rgb_422 doesn't contain this color space conversion. We should do the conversion manually in shader.

So, this bug try to support the GL_APPLE_rgb_422 format for hardware decoded video.

[1]
https://www.khronos.org/registry/OpenGL/extensions/APPLE/APPLE_ycbcr_422.txt
[2]
https://www.khronos.org/registry/OpenGL/extensions/APPLE/APPLE_rgb_422.txt

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1193)

### DIFF
--- a/webrender/res/ps_yuv_image.fs.glsl
+++ b/webrender/res/ps_yuv_image.fs.glsl
@@ -54,6 +54,7 @@ void main(void) {
     vec2 st_y = vTextureOffsetY + clamp(
         relative_pos_in_rect / vStretchSize * vTextureSizeY,
         vHalfTexelY, vTextureSizeY - vHalfTexelY);
+#ifndef WR_FEATURE_INTERLEAVED_Y_CB_CR
     vec2 uv_offset = clamp(
         relative_pos_in_rect / vStretchSize * vTextureSizeUv,
         vHalfTexelUv, vTextureSizeUv - vHalfTexelUv);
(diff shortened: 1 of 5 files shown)


JerryShih:
r?
@glennw @kvark

kvark:
Note: breaking change

@JerryShih Looks good overall! Needs squashing though.

> We should do the conversion manually in shader.

We don't seem to be doing that at the moment. Shouldn't this be a part of the PR?

JerryShih:
@kvark 

> We don't seem to be doing that at the moment. Shouldn't this be a part of the PR?

Maybe I misunderstand your question, but we do the conversion at:

https://github.com/servo/webrender/pull/1193/commits/2e241b9d6f88ce410f801f5541b9cc52bbae533f#diff-764d64b05b62478f5dc98ce4c283be37R69

The GL_APPLE_rgb_422 format put the Y, Cb and Cr data in the same texture. And they are placed in the texture's R, G and B channel.
Then we apply the color matrix for that yuv_value.

JerryShih:
And if we use that yuvData, we will set the interleaved yuv shader at:
https://github.com/servo/webrender/pull/1193/commits/02dfa97a7dab1f8aacee7e4aed36c43ad2e46f37#diff-fa8b017f909ac133e647224ba1087a15R416

JerryShih:
@kvark 
Here are the squashed commits.

kvark:
Thanks!
@bors-servo r+

PR #1194, opened 2017-05-03
Add dotted border support to new path.

glennw:
This relies on the same basic method as the dashed border support.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1194)


glennw:
r? @kvark 

The dotted border support still needs some work - in particular, the spacing doesn't exactly match gecko and there are a few edges cases that don't work correctly. But I think it's in a good enough state to merge and incrementally make improvements to.

webrender/res/cs_clip_border.vs.glsl
```diff
@@ -38,10 +59,49 @@ void main(void) {
...
+                BorderClipDash dash = fetch_border_clip_dash(cci.data_index + cci.segment_index);
+                vPoint_Tangent0 = dash.point_tangent_0 * sign_modifier.xyxy;
+                vPoint_Tangent1 = dash.point_tangent_1 * sign_modifier.xyxy;
+                vDotParams = vec3(0.0);
+                vAlphaMask = vec2(0.0, 1.0);
+                break;
+            }
+            case CLIP_MODE_DOT: {
+                BorderClipDot cdot = fetch_border_clip_dot(cci.data_index + cci.segment_index);
+                vPoint_Tangent0 = vec4(1.0);
```
kvark: `vec4(1.0)` doesn't look like a tangent vector to me. Is this safe?

webrender/res/ps_border_corner.vs.glsl
```diff
@@ -189,6 +189,10 @@ void main(void) {
             vAlphaSelect = 1.0;
             vSDFSelect = 1.0;
             break;
+        case BORDER_STYLE_DOTTED:
+            // Disable normal clip radii for dotted corners, since
+            // all the clipping is handled by the clip mask.
+            vClipSign = vec2(0.0);
```
kvark: I recall some Android devices no appreciating `case` fall-through. Let's have a `break` just to be safe

webrender/src/border.rs
```diff
@@ -444,52 +470,157 @@ impl BorderCornerClipSource {
...
-        let dash_count = actual_count as usize;
+                // Round that up to the nearest integer, so that the dash length
+                // doesn't exceed the ratio above. Add one extra dash to cover
+                // the last half-dash of the arc.
+                (ellipse, 1 + desired_count.ceil() as usize)
+            }
+            BorderCornerClipKind::Dot => {
+                // The centers of dots follow an ellipse along the middle of the
+                // border radius.
+                let inner_radius = LayerSize::new(corner_radius.width - 0.5 * widths.width,
```
kvark: could do `corner_radius - 0.5 * widths`

webrender/src/border.rs
```diff
@@ -444,52 +470,157 @@ impl BorderCornerClipSource {
...
         *header = self.corner_data.into();
 
-        let mut current_arc_length = self.dash_arc_length * 0.5;
-        for dash_index in 0..self.dash_count {
-            let arc_length0 = current_arc_length;
-            current_arc_length += self.dash_arc_length;
+        match self.kind {
+            BorderCornerClipKind::Dash => {
+                // Get the correct dash arc length.
+                self.actual_clip_count = self.max_clip_count;
```
kvark: as an option, we could put it into a `Cell`, given that everything else stays immutable

webrender/src/border.rs
```diff
@@ -544,3 +674,43 @@ impl BorderCornerDashClipData {
...
 }
+
+/// Represents the GPU data for drawing a single dot
+/// to a clip mask.
+#[derive(Debug, Clone)]
+#[repr(C)]
+pub struct BorderCornerDotClipData {
+    pub center: LayerPoint,
+    pub radius: f32,
+    pub padding: [f32; 5],
```
kvark: does it have to be 8 floats and not just 4?

webrender/res/cs_clip_border.vs.glsl
glennw: Yep - it's just there to avoid NaN in the shader - the value gets masked out later in the shader in this code branch.

webrender/res/ps_border_corner.vs.glsl
glennw: Fixed

webrender/src/border.rs
glennw: It's placed in the same store as the Dash data struct (and the other clip-related GPU structs), so the size has to match.

glennw:
@kvark Thanks for the review! Fixes pushed and questions addressed :)

PR #1198, opened 2017-05-04
Resolve the ext-image again even though it's resolved in previous frame.

JerryShih:
r? @kvark @glennw
 
Maybe we will do the partial updating for the frame data(scrolling?). So, the SourceTexture in cpu_yuv_image is not in a invalid status. Then, we will skip that ext-image resolving. Thus, it will not be put into the deferred_resolves list. Finally, we can't resolve that ext-image in renderer.
In this patch, we will always resolve the yuv images for every frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1198)


glennw:
Seems reasonable, thanks!

@bors-servo r+

PR #1199, opened 2017-05-04
Support border corners with differing styles.

glennw:
This handles all border corners styles except those with
dashed and/or dotted styles. Those require changes to the
border clip mask shader, which will be done in a follow up.

The basic idea is to draw the corner twice, once for each style
and mask out the other side of the corner. This results in more
pixels being drawn than necessary, and can sometimes result in the
AA between the corner segments being slightly incorrect. Those
issues can be fixed at a later time.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1199)


webrender/src/tiling.rs
```diff
@@ -425,8 +426,23 @@ impl AlphaRenderItem {
                             let edge_key = AlphaBatchKey::new(AlphaBatchKind::BorderEdge, flags, blend_mode, textures);
 
                             batch_list.with_suitable_batch(&corner_key, item_bounding_rect, |batch| {
-                                for border_segment in 0..4 {
-                                    batch.add_instance(base_instance.build(border_segment, 0, 0));
+                                for (i, instance_kind) in border_cpu.corner_instances.iter().enumerate() {
+                                    let sub_index = i as i32;
+                                    match *instance_kind {
```
kvark: nit: this `match` could be simplified if it only returned `&[BorderCornerSide]`

glennw:
@kvark Thanks - fixed the ```select_style``` function - will fix the other nits up as follow ups to tidy that shader code up :)

@bors-servo r=kvark

staktrace:
@glennw I think this patch *does* actually trigger all those test failures in the gecko try push we were talking about yesterday (tracked in [bug 1363683](https://bugzilla.mozilla.org/show_bug.cgi?id=1363683)). It looks like possibly an infinite loop somewhere inside the render function - or at least something that causes the render function take upwards of 5 minutes to run, at which point the timeout kills the test.

I was able to reproduce locally on the gecko `layout/reftests/abs-pos/table-3.html` reftest. I interrupted with gdb while it was inside the render function and got a backtrace: https://gist.github.com/staktrace/22ea886d20afb334c3b80920ec7c9d0b

kvark:
@staktrace that is worrisome... I re-inspected the code and didn't see a potential cause of endless loops. Perhaps, the hang/delay is coming from the driver or the hardware doesn't like the new code for some reason. I think we should back the PR out and try reproducing your case on Gecko locally.

glennw:
@staktrace @kvark Hmm, that backtrace does indeed suggest the shader change is causing some kind of issue inside llvmpipe. I'm guessing that the shader change has triggered a bug in mesa / llvmpipe :(

glennw:
@staktrace @kvark  I will try to reproduce this locally on llvmpipe today.

glennw:
I imagine it will just be some tweak to the shader needed to avoid triggering the bug in mesa / llvm - hopefully.

glennw:
@staktrace @kvark

I reproduced this locally. It appears to be a bug in the shader compiler in mesa / llvm. I've opened a PR here [https://github.com/servo/webrender/pull/1242] which works around the hang by slightly tweaking the structure of the shader.

It shouldn't have any functional effect, but fixes the problem for me. Is the simplest way to test this just to get it merged and try to do an update?

staktrace:
If the PR can be applied (without too much rebasing) directly onto the copy of webrender in mozilla-central or graphics, that would be the best thing to push to try - that way we start from a known good revision and just test this change in isolation. However if it depends on other things that landed since the last WR update into gecko, or is otherwise problematic to rebase, then it's probably easiest to just get it merged and my next simulated WR update will test it.

PR #1201, opened 2017-05-04
Read the correct render task for composite backdrops

rlhunt:
The backdrop for a composite is a readback task not an alpha batch task.
And they have different locations for `render_target_layer_index`, which
causes problems when the source and backdrop are in different render
targets.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1201)


kvark:
Looks great!
@bors-servo r+

Issue #1178, opened 2017-04-28
Use rayon's thread pool.


nical:
The rayon crate offers a strict superset of what the threadpool crate provides, and the implementation is pretty much state of the art (compared to threadpool which is essentially a big mutex around a single mpsc channel that all of the workers contend on).
Plus, rayon has a lot of very nice things.

nical:
Hm, I got a bit carried away: spawn_async (the feature in rayon that corresponds to what we use in threadpool) is marked as unstable, so we may want to wait a little.

emilio:
On Tue, May 02, 2017 at 05:02:18AM -0700, Nicolas Silva wrote:
> Hm, I got a bit carried away: spawn_async (the feature in rayon that
> corresponds to what we use in threadpool) is marked as unstable, so we
> may want to wait a little.

I think even though it's under the unstable feature flag, it compiles
perfectly fine on rust stable. Indeed stylo uses the unstable scope()
API from rayon, and we don't have problems in m-c.

nical:
Neat! I'll look into it then.

nical:
Done in #1202 .

PR #1202, opened 2017-05-04
Use rayon's thread pool.

nical:
cf. issue #1178.

Rayon's thread pool implementation is a lot better than the one provided by the threadpool crate, and comes with a lot of very useful additional features. The spawn_async functionality which we need is currently behind the unstable feature flag. Note that stylo is also using rayon with the unstable feature flag and is bumping to version 0.7 in https://github.com/servo/servo/pull/16303.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1202)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -41,7 +41,8 @@ use std::sync::{Arc, Mutex};
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::thread;
 use texture_cache::TextureCache;
-use threadpool::ThreadPool;
+use rayon::ThreadPool;
+use rayon::Configuration as ThreadPoolConfig;
 use tiling::{AlphaBatchKind, BlurCommand, Frame, PrimitiveBatch, RenderTarget};
 use tiling::{AlphaRenderTarget, CacheClipInstance, PrimitiveInstance, ColorRenderTarget, RenderTargetKind};
 use time::precise_time_ns;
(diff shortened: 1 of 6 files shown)


nical:
I am being told that it would be best to wait for servo/servo#16303 to be merged before merging this one in order to avoid having 0.6 in servo and 0.7 in webrender.

webrender/src/renderer.rs
```diff
@@ -2215,7 +2217,7 @@ pub struct RendererOptions {
     pub enable_batcher: bool,
     pub render_target_debug: bool,
     pub max_texture_size: Option<u32>,
-    pub workers: Option<Arc<Mutex<ThreadPool>>>,
+    pub workers: Option<Arc<Mutex<rayon::ThreadPool>>>,
```
emilio: Do you need the mutex now? It seems the `spawn_async` function takes an immutable `ThreadPool`, so I don't see why off-hand.

nical:
Thanks @emilio, it's quite a lot nicer without that mutex around the thread pool!

webrender/src/renderer.rs
```diff
@@ -1046,9 +1046,11 @@ impl Renderer {
         let render_target_debug = options.render_target_debug;
         let payload_tx_for_backend = payload_tx.clone();
         let recorder = options.recorder;
+        // TODO(gw): Use a heuristic to select best # of worker threads.
+        let worker_config = rayon::Configuration::new().thread_name(|idx|{ format!("WebRender:Worker#{}", idx) })
+                                                       .num_threads(4);
```
kvark: let's not even call `num_threads`?

> If num_threads is 0, or you do not call this function, then the Rayon runtime will select the number of threads automatically. At present, this is based on the RAYON_RS_NUM_CPUS (if set), or the number of logical CPUs (otherwise).

webrender/src/renderer.rs
```diff
@@ -2215,7 +2217,7 @@ pub struct RendererOptions {
     pub enable_batcher: bool,
     pub render_target_debug: bool,
     pub max_texture_size: Option<u32>,
-    pub workers: Option<Arc<Mutex<ThreadPool>>>,
+    pub workers: Option<Arc<rayon::ThreadPool>>,
```
kvark: while I use this style in some of the personal project, I believe the convention in WR is to `use` all the types you need

glennw:
@nical I marked this as [DO NOT MERGE] based on your comment https://github.com/servo/webrender/pull/1202#issuecomment-299190032 - feel free to update the title when this is ready to go :)

stshine:
servo/servo#16303 has landed.

PR #1204, opened 2017-05-05
Make the check for a matrix preserving a 2d axis aligned rect more robust.

glennw:
Port the Skia implementation for this functionality. Previously, several
reftests were running through the transform shader path when they didn't
need to.

Also remove several unused methods from the matrix helper trait.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1204)


glennw:
r? @kvark 

I've successfully run the Servo WPT/CSS test suite with this change.

kvark:
Nice cleanup!
@bors-servo r+

PR #1205, opened 2017-05-05
Fix interaction between selecting subpixel AA and disabling text AA.

glennw:
Previously, if the render options selected subpixel AA *on*, but
text AA *off* the text rendering would be incorrect. This was due
to the glyphs being rasterized without AA, but the batching code
incorrectly selecting the subpixel blend mode.

This unifies where the selection is made of the type of glyphs
to render, meaning the resource cache just rasterizes what it
is requested to.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1205)


kvark:
Looks clean, thanks!
@bors-servo r+

PR #1208, opened 2017-05-05
Better Preserve3D support

kvark:
Follow up to #1169
Includes #1207

r? @glennw @mrobinson 

## Fixes
  - "preserve-3d" only affects children stacking contexts and contained items
  - generated polygon coordinates of non-zero local bounds
  - sorting order

## Servo 

The Servo PR is being [worked on](https://github.com/servo/servo/compare/master...kvark:preserve3d), but it's not required here, since it may safely continue using `TransformStyle::Flat`.

## Future work

There is at least one feature on the horizon to be implemented before  #739 can be truly closed.
The `TransformStyle` should be moved out of the stacking context and deserve it's own pushable item (similar to clip-scroll groups). This is required because an item without "preserve-3d" automatically becomes "flat" but does not establish a stacking context, which is [tested by Servo](https://raw.githubusercontent.com/servo/servo/master/tests/wpt/css-tests/css-transforms-1_dev/html/transform3d-sorting-004.htm).

Edit: apparently, Chrome disagrees here, so the current approach of WR might stay.
Create Bugstar [issue-1362543](https://bugzilla.mozilla.org/show_bug.cgi?id=1362543).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1208)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,7 @@ dependencies = [
  "clap 2.20.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "crossbeam 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "dwrote 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "env_logger 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "euclid 0.11.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "font-loader 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "gleam 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -30,6 +31,14 @@ dependencies = [
 name = "adler32"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "aho-corasick"
+version = "0.6.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "android_glue"
@@ -71,7 +80,7 @@ dependencies = [
 
 [[package]]
 name = "binary-space-partition"
-version = "0.1.1"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -260,6 +269,15 @@ version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "num-traits 0.1.37 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "env_logger"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -479,6 +497,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "memchr"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.20 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "metadeps"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


glennw:

---

*[webrender/src/frame_builder.rs, line 90 at r1](https://reviewable.io:443/reviews/servo/webrender/1208#-KjZrB-_EyB0J0_6DyVr:-KjZrB-acehvgFauCM69:b7s1z9m) ([raw file](https://github.com/servo/webrender/blob/ef11311c042aabe2c7647ad6d9299cc1fcf3a9d9/webrender/src/frame_builder.rs#L90)):*
> ```Rust
> fn make_polygon(sc: &StackingContext, node: &ClipScrollNode, anchor: usize)
>                 -> Polygon<f32, WorldPixel> {
>     //TODO: only work with `sc.local_bounds` worth of space
> ```

Could you expand on this comment a bit, and why ```local_bounds.origin``` is not used?

---

*[webrender/src/frame_builder.rs, line 92 at r1](https://reviewable.io:443/reviews/servo/webrender/1208#-KjZr2s23Ro9ZGfmhLqB:-KjZr2s3J79xfW5bxYXl:b-ge9v26) ([raw file](https://github.com/servo/webrender/blob/ef11311c042aabe2c7647ad6d9299cc1fcf3a9d9/webrender/src/frame_builder.rs#L92)):*
> ```Rust
>     //TODO: only work with `sc.local_bounds` worth of space
>     let size = sc.local_bounds.bottom_right();
>     let bounds = LayerRect::new(sc.reference_frame_offset, LayerSize::new(size.x, size.y));
> ```

Is size already a LayerSize?

---

*[wrench/reftests/split/reftest.list, line 3 at r1](https://reviewable.io:443/reviews/servo/webrender/1208#-KjZqU1DHww0S_rElnnu:-KjZqU1DHww0S_rElnnv:bok21cu) ([raw file](https://github.com/servo/webrender/blob/ef11311c042aabe2c7647ad6d9299cc1fcf3a9d9/wrench/reftests/split/reftest.list#L3)):*
> ```
> == simple.yaml simple-ref.yaml
> == order.yaml order-ref.yaml
> #TODO: == nested.yaml nested-ref.yaml
> ```

Is this meant to be resolved before merge, or as a follow up?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1208)*

glennw:
@kvark Looks good - just a couple minor issues.

kvark:
@glennw thanks for the review! Your concerns are hopefully addressed now. Please take a look at the last commit. 

---

Review status: 11 of 21 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/src/frame_builder.rs, line 90 at r1](https://reviewable.io:443/reviews/servo/webrender/1208#-KjZrB-_EyB0J0_6DyVr:-KjdHX8Po6k9Nn0-o3nH:bd6cvqk) ([raw file](https://github.com/servo/webrender/blob/ef11311c042aabe2c7647ad6d9299cc1fcf3a9d9/webrender/src/frame_builder.rs#L90)):*

Could you expand on this comment a bit, and why ```local_bounds.origin``` is not used?
</blockquote></details>

good idea!

---

*[webrender/src/frame_builder.rs, line 92 at r1](https://reviewable.io:443/reviews/servo/webrender/1208#-KjZr2s23Ro9ZGfmhLqB:-KjdHb5M8xy_rMnYRm9-:b-v784l8) ([raw file](https://github.com/servo/webrender/blob/ef11311c042aabe2c7647ad6d9299cc1fcf3a9d9/webrender/src/frame_builder.rs#L92)):*

Is size already a LayerSize?
</blockquote></details>

unfortunately, it's `LayerPoint`, and `euclid` doesn't seem to provide convenience conversion between these

---

*[wrench/reftests/split/reftest.list, line 3 at r1](https://reviewable.io:443/reviews/servo/webrender/1208#-KjZqU1DHww0S_rElnnu:-KjdHjQiYcWhJbpT61I3:bv9tvke) ([raw file](https://github.com/servo/webrender/blob/ef11311c042aabe2c7647ad6d9299cc1fcf3a9d9/wrench/reftests/split/reftest.list#L3)):*

Is this meant to be resolved before merge, or as a follow up?
</blockquote></details>

nicely spotted! I worked on that test and encountered a logical problem with the code, related to handling the order of planes that geometrically match each other.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1208#-:-KjdHx5QuQBpeiVleE9x:b6xgkwv)*

staktrace:
> Edit: apparently, Chrome disagrees here, so the current approach of WR might stay.
> Create Bugstar issue-1362543.

@kvark, from the discussion on bug 1362543, it looks like Chrome was correct? Are there any other WR changes needed for preserve-3d, or can we close out #739?

kvark:
@staktrace I want to get some real use cases going (outside of our artificial tests) before closing the issue.

leeoniya:
http://keithclark.co.uk/labs/css-fps/
http://www.michaelbromley.co.uk/experiments/css-space-shooter/

;)

PR #1209, opened 2017-05-08
Support elliptical clips for border-radius on primitives.

glennw:
Port the elliptical clip support from borders to the general
clip mask generation.

The AA is not quite right, but I intend to fix this as a specific
task to unify the AA code for all primitives.

This adds several new passes in the WPT/CSS test suite.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1209)


webrender/res/cs_clip_rectangle.fs.glsl
```diff
@@ -2,35 +2,51 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+void clip_against_ellipse_if_needed(vec2 pos,
+                                    inout float current_distance,
```
kvark: given that there is only one output parameter, it would look cleaner (more FP style) if it was returned directly instead of `inout`

webrender/res/cs_clip_rectangle.glsl
```diff
@@ -5,6 +5,5 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 varying vec3 vPos;
-flat varying vec4 vClipRect;
-flat varying vec4 vClipRadius;
 flat varying float vClipMode;
+flat varying vec4 vClipCenter_Radius[4];
```
kvark: Given our troubles with struct varyings, do you think arrays would be fine?

glennw:
@kvark Good points - I've made both those fixes.

kvark:
Superb! This is more verbose but the code makes me happier, thank you :)
@bors-servo r+

PR #1211, opened 2017-05-08
Eliminate need for root stacking context

### LINKED ISSUE #1195: WR renders transparent if the root of the display list is a clip rect
Apply the following patch and run the example.

```
index db65583..bdb4cf4 100644
@@ -98,16 +98,19 @@ fn main() {

     let epoch = Epoch(0);
     let root_background_color = ColorF::new(0.3, 0.0, 0.0, 1.0);

     let pipeline_id = PipelineId(0, 0);
     let mut builder = webrender_traits::DisplayListBuilder::new(pipeline_id);

     let bounds = LayoutRect::new(LayoutPoint::zero(), LayoutSize::new(width as f32, height as f32));
+    builder.push_clip_node(ClipRegion::simple(&(0, 0).to(width as i32, height as i32)),
+                           (0, 0).to(width as i32, height as i32),
+                           None);
     builder.push_stacking_context(webrender_traits::ScrollPolicy::Scrollable,
                                   bounds,
                                   None,
                                   TransformStyle::Flat,
                                   None,
                                   webrender_traits::MixBlendMode::Normal,
                                   Vec::new());

@@ -167,16 +170,17 @@ fn main() {
                           ColorF::new(0.0, 1.0, 1.0, 1.0));

         builder.pop_clip_id(); // clip_id
         builder.pop_stacking_context();
     }

     builder.pop_stacking_context();
+    builder.pop_clip_node();

     api.set_display_list(
         Some(root_background_color),
         epoch,
         LayoutSize::new(width as f32, height as f32),
         builder.finalize(),
         true);
     api.set_root_pipeline(pipeline_id);
```

Expected: the example window renders content and behaves the same as it does without the patch, since the root clip that I'm pushing includes the entire window.
Actual results: the window doesn't render anything, and I just see the desktop background through it.

@mrobinson not sure if this is expected behaviour or not. Maybe there's an assumption somewhere that the first thing int he display list is a stacking context? If so we should assert that, probably.

mrobinson:
This makes the code quit a bit cleaner and allows for a greater variety
of display lists.

Fixes #1195.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1211)

### DIFF
--- a/webrender_traits/src/api.rs
+++ b/webrender_traits/src/api.rs
@@ -8,10 +8,10 @@ use offscreen_gl_context::{GLContextAttributes, GLLimits};
 use std::cell::Cell;
 use std::fmt;
 use std::marker::PhantomData;
-use {BuiltDisplayList, BuiltDisplayListDescriptor};
-use {ClipId, ColorF, DeviceIntPoint, DeviceIntSize, DeviceUintRect, DeviceUintSize, FontKey};
-use {GlyphDimensions, GlyphKey, ImageData, ImageDescriptor, ImageKey, LayoutPoint, LayoutSize};
-use {LayoutTransform, NativeFontHandle, WorldPoint};
+use {BuiltDisplayListDescriptor, ClipId, ColorF, DeviceIntPoint, DeviceIntSize};
+use {DeviceUintRect, DeviceUintSize, DisplayListBuilder, FontKey, GlyphDimensions, GlyphKey};
+use {ImageData, ImageDescriptor, ImageKey, LayoutPoint, LayoutSize, LayoutTransform};
+use {NativeFontHandle, WorldPoint};
 #[cfg(feature = "webgl")]
 use {WebGLCommand, WebGLContextId};
 
@@ -34,10 +34,12 @@ pub enum ApiMsg {
     /// Supplies a new frame to WebRender.
     ///
     /// After receiving this message, WebRender will read the display list from the payload channel.
+    // TODO: We should consider using named members to avoid confusion.
     SetDisplayList(Option<ColorF>,
                    Epoch,
                    PipelineId,
-                   LayoutSize,
+                   LayoutSize, // viewport_size
+                   LayoutSize, // content size
                    BuiltDisplayListDescriptor,
                    bool),
     SetPageZoom(ZoomFactor),
@@ -301,21 +303,23 @@ impl RenderApi {
                             background_color: Option<ColorF>,
                             epoch: Epoch,
                             viewport_size: LayoutSize,
-                            (pipeline_id, display_list): (PipelineId, BuiltDisplayList),
+                            display_list_builder: DisplayListBuilder,
                             preserve_frame_state: bool) {
-        let (dl_data, dl_desc) = display_list.into_data();
+        let (pipeline_id, content_size, display_list) = display_list_builder.finalize();
+        let (display_list_data, display_list_descriptor) = display_list.into_data();
         let msg = ApiMsg::SetDisplayList(background_color,
-                                             epoch,
-                                             pipeline_id,
-                                             viewport_size,
-                                             dl_desc,
-                                             preserve_frame_state);
+                                         epoch,
+                                         pipeline_id,
+                                         viewport_size,
+                                         content_size,
+                                         display_list_descriptor,
+                                         preserve_frame_state);
         self.api_sender.send(msg).unwrap();
 
         self.payload_sender.send_payload(Payload {
             epoch: epoch,
             pipeline_id: pipeline_id,
-            display_list_data: dl_data,
+            display_list_data: display_list_data,
         }).unwrap();
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,7 @@ dependencies = [
  "clap 2.20.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "crossbeam 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "dwrote 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "env_logger 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "euclid 0.11.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "font-loader 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "gleam 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -21,8 +22,8 @@ dependencies = [
  "serde_json 0.9.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-glutin 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.36.0",
- "webrender_traits 0.36.0",
+ "webrender 0.37.0",
+ "webrender_traits 0.37.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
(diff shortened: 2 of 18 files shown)


mrobinson:
cc @staktrace 

This eliminates the need for the root stacking context at the API level, although when one isn't provided we create one. This internal quirk can also be removed if it becomes a problem. I have chosen not to handle multiple root stacking contexts with blend isolation. We can make that work properly later, as well, if we decide that it's necessary.

mrobinson:
Also cc @Gankro, since these changes touch his new code a bit.

kvark:

---

*[webrender/src/frame.rs, line 666 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdSvNBUIIQ_TdDdr8q:-KjdSvNCbnV0avIqtjjE:b43kyu3) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender/src/frame.rs#L666)):*
> ```Rust
>                          reference_frame_relative_offset: LayerPoint) {
>         loop {
>             let subtraversal = {
> ```

that's still hacky but looks quite easier on the comprehension side
:+1: 

---

*[webrender/src/frame_builder.rs, line 259 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdT8r5clHlYUx5IzrA:-KjdT8r5clHlYUx5IzrB:bmqy58k) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender/src/frame_builder.rs#L259)):*
> ```Rust
>     }
> 
>     pub fn started_new_pipeline(&mut self) {
> ```

nit: I think this is a little against the naming recommendations? As in - I'd prefer the functions to start with a verb, as `start_new_pipeline`, or `notify_new_pipeline_started`

---

*[webrender_traits/src/api.rs, line 40 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdViN0bWkxHtVTbwrg:-KjdViN0bWkxHtVTbwrh:bvlggz8) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender_traits/src/api.rs#L40)):*
> ```Rust
>                    Epoch,
>                    PipelineId,
>                    LayoutSize, // viewport_size
> ```

we should consider making this enum variant (and some others) to have named fields. Not necessarily for this PR, but perhaps add a `TODO` in there?

---

*[webrender_traits/src/display_list.rs, line 440 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdW3KRvKnMm8wE3vFV:-KjdW3KSlfTEidk6wiKj:b1pqj8p) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender_traits/src/display_list.rs#L440)):*
> ```Rust
>         let start_time = precise_time_ns();
> 
>         // We start at 1 here, because the root scroll id is always 0.
> ```

would be nice to have this informal dependency formalized in code

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1211)*

mrobinson:

---

*[webrender/src/frame.rs, line 666 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdSvNBUIIQ_TdDdr8q:-Kjg4khXftw4PpytwpOE:b-ucfknh) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender/src/frame.rs#L666)):*

that's still hacky but looks quite easier on the comprehension side
:+1: 
</blockquote></details>

Thanks!

---

*[webrender/src/frame_builder.rs, line 259 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdT8r5clHlYUx5IzrA:-Kjg4Cioy8alvQCwszxQ:b6l8l1f) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender/src/frame_builder.rs#L259)):*

nit: I think this is a little against the naming recommendations? As in - I'd prefer the functions to start with a verb, as `start_new_pipeline`, or `notify_new_pipeline_started`
</blockquote></details>

You're right. I've changed this to `notify_waiting_for_root_stacking_context`.

---

*[webrender_traits/src/api.rs, line 40 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdViN0bWkxHtVTbwrg:-Kjg4IRaKSZo2WBbscMo:b-etecvj) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender_traits/src/api.rs#L40)):*

we should consider making this enum variant (and some others) to have named fields. Not necessarily for this PR, but perhaps add a `TODO` in there?
</blockquote></details>

I've added a TODO here.

---

*[webrender_traits/src/display_list.rs, line 440 at r1](https://reviewable.io:443/reviews/servo/webrender/1211#-KjdW3KRvKnMm8wE3vFV:-Kjg4M4LtQoqLdV81Bgt:bkocfdu) ([raw file](https://github.com/servo/webrender/blob/9a22fffce0fd61c468d79f51b01404ade50780a6/webrender_traits/src/display_list.rs#L440)):*

would be nice to have this informal dependency formalized in code
</blockquote></details>

I agree. Ideally it would be impossible to accidentally create an id that referred to the root scrolling node. I've been rolling the problem around in my head for a while and I think it's possible to do by modifying the definition of ClipId. In this case, I'm trying to make the issue clearer in the code by moving the magic value to a constant.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1211)*

kvark:
Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1211)*

kvark:
Thanks!
@bors-servo r+

glennw:
@mrobinson I'm getting a number of panics in the Servo WPT tests when trying to do a WR update. The panic is unwrapping a stacking context index - in tests like ```meta_viewport_resize```. Is it likely related to this PR? It might just be that I ported the API changes incorrectly...?

mrobinson:
@glennw Thanks! I've uploaded https://github.com/servo/webrender/pull/1222 to fix this issue.

webrender_traits/src/api.rs
```diff
@@ -301,21 +303,23 @@ impl RenderApi {
                             background_color: Option<ColorF>,
                             epoch: Epoch,
                             viewport_size: LayoutSize,
-                            (pipeline_id, display_list): (PipelineId, BuiltDisplayList),
+                            display_list_builder: DisplayListBuilder,
                             preserve_frame_state: bool) {
-        let (dl_data, dl_desc) = display_list.into_data();
+        let (pipeline_id, content_size, display_list) = display_list_builder.finalize();
```
jrmuizel: This breaks Gecko. We finalize in a different process from the one that calls set_display_list. Is there a particular reason for this?

staktrace:
@mrobinson FYI the change to set_display_list makes this hard to use with Gecko. We create a BuiltDisplayList which is serialized from one process to another across IPC and then we pass it to set_display_list. Your change to have that function take a DisplayListBuilder makes that harder, because now we have to wrap the BuiltDisplayList back into a DisplayListBuilder.

staktrace:
Whoops, didn't see jeff's comment just above. We're complaining about the same thing :)

webrender_traits/src/api.rs
kvark: #1226

staktrace:
Could the documentation for `set_display_list` also be updated to explain what `content_size` is supposed to be?

mrobinson:
@staktrace Created https://github.com/servo/webrender/pull/1243 for your request.

staktrace:
Thanks!

PR #1212, opened 2017-05-08
Implement WebGL::VertexArrayObject extension

MortimerGoro:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1212)

### DIFF
--- a/webrender_traits/src/webgl.rs
+++ b/webrender_traits/src/webgl.rs
@@ -63,6 +63,7 @@ pub enum WebGLCommand {
     FramebufferRenderbuffer(u32, u32, u32, Option<WebGLRenderbufferId>),
     FramebufferTexture2D(u32, u32, u32, Option<WebGLTextureId>, i32),
     GetBufferParameter(u32, u32, MsgSender<WebGLResult<WebGLParameter>>),
+    GetExtensions(MsgSender<String>),
     GetParameter(u32, MsgSender<WebGLResult<WebGLParameter>>),
     GetProgramParameter(WebGLProgramId, u32, MsgSender<WebGLResult<WebGLParameter>>),
     GetShaderParameter(WebGLShaderId, u32, MsgSender<WebGLResult<WebGLParameter>>),
@@ -124,6 +125,9 @@ pub enum WebGLCommand {
     Finish(MsgSender<()>),
     Flush,
     GenerateMipmap(u32),
+    CreateVertexArray(MsgSender<Option<WebGLVertexArrayId>>),
+    DeleteVertexArray(WebGLVertexArrayId),
+    BindVertexArray(Option<WebGLVertexArrayId>),
 }
 
 #[cfg(feature = "nightly")]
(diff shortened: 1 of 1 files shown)


webrender_traits/src/webgl.rs
```diff
@@ -973,6 +994,16 @@ impl WebGLCommand {
         chan.send(shader).unwrap();
     }
 
+    fn create_vertex_array(gl: &gl::Gl, chan: MsgSender<Option<WebGLVertexArrayId>>) {
+        let vao = gl.gen_vertex_arrays(1)[0];
+        let vao = if vao == 0 {
```
kvark: is this a valid return value for the `gen_vertex_array`?

MortimerGoro: I think that it can return 0 in situations where the GLContext is not correctly set. Also, maybe some GPUs have some limits for the quantity of VAOs that can be created. 

A similar check is done in other crate calls like create_framebuffer, create_texture.

kvark: That doesn't sound like a case where we just check for the returned ID then.
> I think that it can return 0 in situations where the GLContext is not correctly set

Shouldn't it be forbidden to do any GL calls in this case?

> maybe some GPUs have some limits for the quantity of VAOs that can be created

That would be a GL error then. What is our WebGL implementation policy with regards to GL errors?

> A similar check is done in other crate calls like create_framebuffer, create_texture.

Thanks for pointing this out. I'd like to investigate further.

MortimerGoro: > Shouldn't it be forbidden to do any GL calls in this case?

Yes. It must be checked in the Servo WebGL implementation. It's possible to call WebGL methods when a context is lost (e.g. a notification when the app is in background) but those calls should be discarted by the WebGLRendeirngContext and don't arrive here.

> That would be a GL error then. What is our WebGL implementation policy with regards to GL errors?

Most of the errors are detected without calling the driver following the rules defines in WebGL spec. Other than that,  the WebGL developer should be responsible for calling getError() and act accordingly.

> Thanks for pointing this out. I'd like to investigate further.

The WebGL spec uses nullable return values for createTexture, createFramebuffer, etc. One of the reason is because of the possibility of calling the methods with a lost context. I don't find anything about other kind of errors to return null values. I think that @emilio implemented the check for create_xxx calls, do you know any other case where it can be 0/null is a valid value?

kvark:
Thanks for clarifications @MortimerGoro !
If you don't mind, I'm going to yield to @emilio in handling this.

webrender_traits/src/webgl.rs
emilio: Hmm, I unfortuntately don't remember why I added those checks.

That being said, we right now panic if we detect any gl error in the way, so I don't think this should be a safety problem.

This looks fine to me for now, since it's consistent with the rest of the code, if those checks need to be removed, and we need to check for `gl::get_error()` instead, we can do it in a separate patch.

Issue #1206, opened 2017-05-05
Shader compilation error on some windows hardware


staktrace:
In [this try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=db7e9f22bb8ab7da5820bdd071cedde86d894219) the windows 8 mochitest-bc failures happen because WR fails to startup. The log has this shader compilation error:

```
WebRender - OpenGL version new OpenGL ES 3.0 (ANGLE 2.1.0.dec065540d5f)
...
08:11:31     INFO - GECKO(4672) | Failed to link shader program: C:\fakepath(2369,1-2): error X3004: undeclared identifier 's3'
08:11:31     INFO - GECKO(4672) | Warning: D3D shader compilation failed with default flags. (vs_5_0)
08:11:31     INFO - GECKO(4672) |  Retrying with skip validation
08:11:31     INFO - GECKO(4672) | C:\fakepath(2369,1-2): error X3004: undeclared identifier 's3'
08:11:31     INFO - GECKO(4672) | Warning: D3D shader compilation failed with skip validation flags. (vs_5_0)
08:11:31     INFO - GECKO(4672) |  Retrying with skip optimization
08:11:31     INFO - GECKO(4672) | C:\fakepath(2369,1-2): error X3004: undeclared identifier 's3'
08:11:31     INFO - GECKO(4672) | Warning: D3D shader compilation failed with skip optimization flags. (vs_5_0)
08:11:31     INFO - GECKO(4672) | Failed to create D3D shaders.
```

Note however that this compilation failure happens pretty late in the log. AFAICT WebRender is initialized numerous times before that successfully. It might just be a shader that we're loading lazily and only gets loaded by particular tests.

Also note that the regular mochitest suite in the same try push must be running on different hardware because it seems to start WR just fine.

mephisto41:
I had same problem here. mark [1] and [2] fixed problem but I didn't know why. Maybe dump the d3d shader code which is generated by ANGLE to see what happened here.

[1]: https://github.com/servo/webrender/blob/master/webrender/res/ps_border_edge.vs.glsl#L45
[2]: https://github.com/servo/webrender/blob/master/webrender/res/ps_border_edge.vs.glsl#L48

mephisto41:
This is compiled shader for ```write_color```

void _write_color_float4(in float4 _color, in float _style, in bool _flip)
{
float2 _modulate = {0, 0};
switch (ivec1(_style)) {
case (6):
s3;
if (_flip)
{
[...]

PR #1214, opened 2017-05-09
Add the shader name to link errors.

nical:
Very simple change that should help with figuring out issues like #1206.

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1214)


webrender/src/device.rs
```diff
@@ -860,7 +860,7 @@ pub struct Capabilities {
 #[derive(Clone, Debug)]
 pub enum ShaderError {
     Compilation(String, String), // name, error mssage
```
kvark: should really use the named fields here instead of the comments

PR #1215, opened 2017-05-09
Expand gitignore to include common editor metadata files

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1215)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,18 @@
 target/
 *~
 *#
+
+# Editors
+*.swp
+*.swo
+
+# IntelliJ
+.idea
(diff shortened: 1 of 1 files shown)


kvark:
Looks reasonable
@bors-servo r+

kvark:
Mac is still stubborn. Retrying.
@bors-servo retry

metajack:
Not a big deal, but usually I put these in the global .gitignore on my machine instead of in every project's .gitignore.

mrobinson:
@metajack Thanks for the tip!

PR #1220, opened 2017-05-09
UNIX font rendering rework

### LINKED ISSUE #1224: Glyph width sometimes incorrectly computed in unix/font.rs
when generating a glyph with mode ```FT_PIXEL_MODE_LCD``` we appear to unconditionally add two pixels to the glyph_width for the extra sub-pixels on each side of the glyphs. It turns out freetype doesn't always render a bitmap with these two extra pixels. As a result the reported width is bigger than it should be and we only crash much farther when we are about to upload the glyph in device.rs and check that the buffer size is large enough.

I have a fix in the making.

kvark:
Addresses https://github.com/servo/servo/issues/16763
Fixes #1224

cc @changm 
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1220)

### DIFF
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -13,9 +13,9 @@ use freetype::freetype::{FT_Library, FT_Set_Char_Size};
 use freetype::freetype::{FT_Face, FT_Long, FT_UInt, FT_F26Dot6};
 use freetype::freetype::{FT_Init_FreeType, FT_Load_Glyph, FT_Render_Glyph};
 use freetype::freetype::{FT_New_Memory_Face, FT_GlyphSlot, FT_LcdFilter};
-use freetype::freetype::{FT_Done_Face};
+use freetype::freetype::{FT_Done_Face, FT_Error};
 
-use std::{mem, ptr, slice};
+use std::{cmp, mem, ptr, slice};
 use std::collections::HashMap;
 
 struct Face {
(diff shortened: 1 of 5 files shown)


webrender/src/platform/unix/font.rs
```diff
@@ -153,93 +166,81 @@ impl FontContext {
...
+        let dimensions = Self::get_glyph_dimensions_impl(slot).unwrap();
         let bitmap = unsafe { &(*slot).bitmap };
-        let metrics = unsafe { &(*slot).metrics };
-        let glyph_width = (metrics.width >> 6) as i32;
-        let glyph_height = (metrics.height >> 6) as i32;
-        let mut final_buffer = Vec::with_capacity(glyph_width as usize *
-                                                  glyph_height as usize *
+        let pixel_mode = unsafe { mem::transmute(bitmap.pixel_mode as u32) };
+        info!("Rasterizing {:?} as {:?} with dimensions {:?}", key, render_mode, dimensions);
+        // we may be filling only a part of the buffer, so initialize the whole thing with 0
```
kvark: this is a leftover comment, I'll remove it when applying the other fixes a reviewer may request

glennw:
@kvark This does seem to be causing issues in the servo reftests unfortunately. It looks like a few stray pixels in glyph runs - quite intermittent (I'm seeing failures each test run, but on different tests).

PR #1222, opened 2017-05-10
Always push a root stacking context

mrobinson:
This makes the logic simpler and fixes a bug where scrollbars were added
without a root stacking context.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1222)


mrobinson:
This fixes an issue uncovered in the Servo WebRender update.

kvark:
nice!
@bors-servo r+

PR #1223, opened 2017-05-10
Allow setting the scroll position of nodes without clamping

### LINKED ISSUE #1213: Figure out WebRender clamping of scroll offset
For gecko integration it would make my life a lot easier if WR offered a way to set a scroll position and didn't clamp it to the scroll bounds, or alternatively, allowed the caller to have more control over what it considers the scroll bounds.

Right now the `scroll_nodes` function in clip_scroll_tree.rs and the `set_scroll_origin` function in `clip_scroll_node.rs` both do a bit of clamping on the scroll position provided, but assume the minimum allowed scroll position is 0 and the maximum is the size of content rect minus the size of the viewport. For one thing this won't work for RTL pages, because on RTL pages you should start off at a scroll offset of x=0 and then scroll into negative x values for overflowing content.

But more relevant to APZ, we want to be able to set negative scroll offsets in webrender because the display list we send to WR already has a scroll offset baked into it, and a negative scroll offset would allow scrolling upwards/leftwards from that baked-in scroll offset. In other words, there is a mismatch between the displaylist that gecko produces and what WR assumes it's getting, and in order to scroll properly we need to be able to give WR negative scroll offsets. The other option of changing what gecko produces seems much more risky and error-prone.

/cc @mrobinson

mrobinson:
This change allows optionally setting arbitrary scroll positions on
scroll nodes in order to support more types of Gecko behavior. Content
will still be clipped normally, but can be moved arbitrarily. In order
to support full Gecko behavior we will need to prevent content from
being clipped as well, but that will come in a later change.

Fixes #1213.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1223)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -10,10 +10,10 @@ use renderer::VertexDataStore;
 use spring::{DAMPING, STIFFNESS, Spring};
 use tiling::PackedLayerIndex;
 use util::TransformedRectKind;
-use webrender_traits::{ClipId, ClipRegion, LayerPixel, LayerPoint, LayerRect, LayerSize};
-use webrender_traits::{LayerToScrollTransform, LayerToWorldTransform, PipelineId};
-use webrender_traits::{ScrollEventPhase, ScrollLayerRect, ScrollLocation, WorldPoint};
-use webrender_traits::{DeviceIntRect, WorldPoint4D};
+use webrender_traits::{ClipId, ClipRegion, DeviceIntRect, LayerPixel, LayerPoint, LayerRect};
+use webrender_traits::{LayerSize, LayerToScrollTransform, LayerToWorldTransform, PipelineId};
+use webrender_traits::{ScrollClamping, ScrollEventPhase, ScrollLayerRect, ScrollLocation};
+use webrender_traits::{WorldPoint, WorldPoint4D};
 
 #[cfg(target_os = "macos")]
 const CAN_OVERSCROLL: bool = true;
@@ -198,7 +198,7 @@ impl ClipScrollNode {
         LayerSize::new(overscroll_x, overscroll_y)
     }
 
-    pub fn set_scroll_origin(&mut self, origin: &LayerPoint) -> bool {
+    pub fn set_scroll_origin(&mut self, origin: &LayerPoint, clamp: ScrollClamping) -> bool {
         match self.node_type {
             NodeType::ReferenceFrame(_) => {
                 warn!("Tried to scroll a reference frame.");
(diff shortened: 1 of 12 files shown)


webrender/src/clip_scroll_node.rs
```diff
@@ -198,7 +198,7 @@ impl ClipScrollNode {
         LayerSize::new(overscroll_x, overscroll_y)
     }
 
-    pub fn set_scroll_origin(&mut self, origin: &LayerPoint) -> bool {
+    pub fn set_scroll_origin(&mut self, origin: &LayerPoint, clamp_to_bounds: bool) -> bool {
```
emilio: nit: Do you think it makes sense to have a `BoundsClampingMode` enum instead? Perhaps it's not too worth it, so just a suggestion :)

mrobinson: Yeah, I think that would be a better design. Let me make that change now...

wrench/reftests/scrolling/reftest.list
```diff
@@ -1,7 +1,8 @@
 == empty-mask.yaml empty-mask-ref.yaml
-== fixed-position.yaml fixed-position-ref.yaml
 == fixed-position-scrolling-clip.yaml fixed-position-scrolling-clip-ref.yaml
+== fixed-position.yaml fixed-position-ref.yaml
 == nested-scroll-offset.yaml nested-scroll-offset-ref.yaml
+== out-of-bounds-scroll.yaml out-of-bounds-scroll-ref.yaml
```
emilio: This test seems to be missing?

mrobinson: Okay. I've ensured that it's included now.

mrobinson:
@emilio Thanks for the review. I think my latest push fixes the issues that you mentioned.

webrender/src/clip_scroll_tree.rs
```diff
@@ -309,7 +309,7 @@ impl ClipScrollTree {
             node.finalize(&scrolling_state);
 
             if let Some(pending_offset) = self.pending_scroll_offsets.remove(clip_id) {
```
kvark: nit: we can destruct the tuple here

webrender_traits/src/api.rs
```diff
@@ -18,6 +18,12 @@ use {WebGLCommand, WebGLContextId};
 pub type TileSize = u16;
 
 #[derive(Clone, Deserialize, Serialize)]
+pub enum ScrollClamping {
```
kvark: let's move it after `ApiMsg` definition, to match what over used enums are doing

glennw:
@bors-servo r+

Let's merge this and fix the remaining nits as follow ups :)

mrobinson:
Nits addressed in https://github.com/servo/webrender/pull/1234.

PR #1227, opened 2017-05-10
Derive Hash and Eq for WebGL resource ids

MortimerGoro:
I need to use WebGL resources in a HashMap ;)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1227)


emilio:
Sounds reasonable.

@bors-servo r+

PR #1229, opened 2017-05-10
[WIP] Add coverage recording

jrmuizel:
Testing out coverage recording

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1229)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,13 +39,7 @@ script:
   - if [ $BUILD_KIND = DEBUG ]; then (cargo test --all --verbose); fi
   - if [ $BUILD_KIND = RELEASE ]; then (cargo test --all --verbose); fi
   - if [ $BUILD_KIND = DEBUG ]; then (wget https://github.com/SimonKagstrom/kcov/archive/master.tar.gz &&
-                                        tar xzf master.tar.gz &&
-                                        cd kcov-master &&
-                                        mkdir build &&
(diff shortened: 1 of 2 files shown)


jrmuizel:
This looks like it has the same segfault problem that I get when running locally. I wonder why https://github.com/codecov/example-rust/blob/master/.travis.yml works

glennw:
@jrmuizel Do you want this kept open for now?

jrmuizel:
Yeah, let's just keep it open until it works if you don't mind.

jrmuizel:
It looks like the --verify flag was needed to get this working

kvark:
> ./wrench/headless.py:70: Line is longer than 120 characters

kvark:
Why do we have two identical `coverage` notifications?

glennw:
@jrmuizel Do we still want this open? It will need to be rebased if we do...

jrmuizel:
Yeah, though I wasn't getting the most useful data. I think it should probably be run on a debug build to get better information.

glennw:
@jrmuizel semi-regular question on if you still want this open?

jrmuizel:
Yeah, I'll try again at it when I care about it.

glennw:
Closing since it hasn't been rebased for a long time. We can re-open when planning to work on this again.

PR #1230, opened 2017-05-11
Mark as needed when blob image is same epoch.

mephisto41:
Otherwise, we'll hit assertion in https://github.com/servo/webrender/blob/master/webrender/src/resource_cache.rs#L177-L180

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1230)


kvark:
I suppose that's fixed by  #1225 already

mephisto41:
Ah, you're right. Thanks for the information. Close this PR.

PR #1232, opened 2017-05-11
Make the AA for transforms more closely match that of the other shaders.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1232)


glennw:
r? @kvark 

I've run this against the CSS/WPT tests. We don't need the inset hack now that we have the more robust method of running the non-transform shader.

The AA code still needs a lot of work, but this makes it consistent - and gives us correct results for clip rects that happen to be axis aligned.

PR #1234, opened 2017-05-11
Fix some nits from the review of #1223

mrobinson:
These were requested by @kvark.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1234)


kvark:
Thank you!
@bors-servo r+

PR #1235, opened 2017-05-11
Implement WebGL::GetVertexAttribOffset

MortimerGoro:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1235)

### DIFF
--- a/webrender_traits/src/webgl.rs
+++ b/webrender_traits/src/webgl.rs
@@ -73,6 +73,7 @@ pub enum WebGLCommand {
     GetAttribLocation(WebGLProgramId, String, MsgSender<Option<i32>>),
     GetUniformLocation(WebGLProgramId, String, MsgSender<Option<i32>>),
     GetVertexAttrib(u32, u32, MsgSender<WebGLResult<WebGLParameter>>),
+    GetVertexAttribOffset(u32, u32, MsgSender<WebGLResult<isize>>),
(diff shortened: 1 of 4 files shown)


webrender_traits/src/webgl.rs
```diff
@@ -73,6 +73,7 @@ pub enum WebGLCommand {
     GetAttribLocation(WebGLProgramId, String, MsgSender<Option<i32>>),
     GetUniformLocation(WebGLProgramId, String, MsgSender<Option<i32>>),
     GetVertexAttrib(u32, u32, MsgSender<WebGLResult<WebGLParameter>>),
+    GetVertexAttribOffset(u32, u32, MsgSender<WebGLResult<isize>>),
```
kvark: I'm slightly concerned about the naming here, given that `GetVertexAttribOffset` translates to `get_vertex_attrib_pointer_v` call. I fully agree that the semantics is much better with `*Offset`, but is it fine to (i.e. is this the only case of) have the WR webgl command name not exactly matching the underlying GL call?

MortimerGoro: I used GetVertexAttribOffset because that's how the WebGL method is named: https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.10 WebGL methods generally match the ones in OpenGL ES, but not with this one.

kvark:
Makes sense, thanks!
@bors-servo r+

Issue #1001, opened 2017-03-22
the yuv image interface discussion


JerryShih:
Currently, webrender uses 3 different image keys for a planer yuv image(there are 3 y, u and v channels).
```
    pub fn push_yuv_image(&mut self,
                          rect: LayoutRect,
                          clip: ClipRegion,
                          y_key: ImageKey,
                          u_key: ImageKey,
                          v_key: ImageKey,
                          color_space: YuvColorSpace)
```
I'm considering use one image key for the 3 of y, u and v channels. It would be less error prone and would facilitate gecko integration. For this change, we need to do a bit more complicated in the resource cache(but I think it's probably not too much).

@nical @sotaroikeda @kvark @glennw 
Any idea for this change?

glennw:
My initial reaction is that it will actually turn out to be quite complicated inside WR, and I'd prefer to avoid it.

But that's just an initial reaction - it's hard to say for sure without seeing some kind of prototype / patch, which might prove me wrong!

nical:
I am keen on having the yuv planes as well as the color space associated with the image (during creation and update of the image object) for yuv images and thus associated to the same ImageKey, instead of being associated to the display item, if that can be done without complicating the code too much.

I suppose it would boil down to having resource cache items potentially associated to several texture cache items. It would take one to prototype this to see if the implementation ends up being enough simple I guess.

If it is too complicated that's not a big deal, really. It would just be nice.

JerryShih:
I have done the all implementations with this idea, but code is not clear. And it doesn't resolve the original problem for integrating gecko with WR. In gecko, we try to map a externalImage id to a textureHost. In some video case, a textureHost could contain more than one channel data(e.g. format yuv have 3 channel, and nv12 have 2 channel). So, I change to add a parameter called channel_index in ExternalImageData and ExternalImageHandler. Then, we could use both ExternalImageId and channel_index to get the specific image channel data.

```
 pub struct ExternalImageData {
     pub id: ExternalImageId,
+    pub channel_index: u8,
     pub image_type: ExternalImageType,
 }
```

```
 pub trait ExternalImageHandler {
     /// Lock the external image. Then, WR could start to read the image content.
     /// The WR client should not change the image content until the unlock()
     /// call.
-    fn lock(&mut self, key: ExternalImageId) -> ExternalImage;
+    fn lock(&mut self, key: ExternalImageId, channel_index: u8) -> ExternalImage;
     /// Unlock the external image. The WR should not read the image content
[...]

PR #1237, opened 2017-05-11
[WIP] Fix multiple channel image bug

### LINKED ISSUE #1216: Fix the channel index logic for image after #1125.
@nical @kvark @glennw 

In  #1125, we could insert different image information for the different channel with the same imageKey. It's used for some yuv image formats(e.g. NV12 has 2 channel). Currently, the ImageTemplates in resource_cache doesn't handle this situation. It only uses imageKey for the hash key. So, the previous channel data in the yuv image will be replaced by the new one because they use the same imageKey.
I will have a pull request for this problem later. In brief, the image related interfaces will have an additional channel_index parameter.

JerryShih:
~~The PR for~~ Fixes #1216

r? @kvark @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1237)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -195,6 +195,7 @@ fn push_sub_clip(api: &RenderApi, builder: &mut DisplayListBuilder, bounds: &Lay
                  -> ClipRegionToken {
     let mask_image = api.generate_image_key();
     api.add_image(mask_image,
+                  None,
                   ImageDescriptor::new(2, 2, ImageFormat::A8, true),
                   ImageData::new(vec![0, 80, 180, 255]),
                   None);
--- a/webrender/examples/blob.rs
+++ b/webrender/examples/blob.rs
@@ -180,6 +180,7 @@ fn main() {
     let blob_img1 = api.generate_image_key();
     api.add_image(
         blob_img1,
+        None,
         ImageDescriptor::new(500, 500, ImageFormat::RGBA8, true),
         ImageData::new_blob_image(serialize_blob(ColorU::new(50, 50, 150, 255))),
         Some(128),
@@ -188,6 +189,7 @@ fn main() {
     let blob_img2 = api.generate_image_key();
     api.add_image(
         blob_img2,
+        None,
         ImageDescriptor::new(200, 200, ImageFormat::RGBA8, true),
         ImageData::new_blob_image(serialize_blob(ColorU::new(50, 150, 50, 255))),
         None,
--- a/webrender/examples/yuv.rs
+++ b/webrender/examples/yuv.rs
@@ -16,7 +16,7 @@ use std::env;
 use std::path::PathBuf;
 use webrender_traits::{ColorF, Epoch};
 use webrender_traits::{DeviceIntPoint, DeviceUintSize, LayoutPoint, LayoutRect, LayoutSize};
-use webrender_traits::{ImageData, ImageDescriptor, ImageFormat};
+use webrender_traits::{ImageChannel, ImageData, ImageDescriptor, ImageFormat};
 use webrender_traits::{PipelineId, TransformStyle};
 use webrender_traits::{YuvColorSpace, YuvData};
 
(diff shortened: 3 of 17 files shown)


kvark:
@JerryShih  What is the benefit of having multiple channels/images behind an `ImageKey` versus just using multiple `ImageKey`s?

JerryShih:
> @JerryShih What is the benefit of having multiple channels/images behind an ImageKey versus just using multiple ImageKeys?

In gecko, a single compositable unit is called textureHost. The textureHost could represent a regular image(single channel) or a complicated video frame(e.g. a two-channel NV12 hardware buffer). Because it is treated as a single compositable unit, gecko just use a ImageKey for that. If we use multiple ImageKeys for all channel in video frame, it will break the concept of compositable.
Another reason it that we will not just use a Y channel for drawing. We need all Y, U and V channel together. If we use a single ImageKey to contain all of them, we could just remove all channels with that single ImageKey.

cc @nical

kvark:
@JerryShih 
> Because it is treated as a single compositable unit, gecko just use a ImageKey for that. If we use multiple ImageKeys for all channel in video frame, it will break the concept of compositable.

I'm still missing something here. Why can't a compositable be represented by a set of channels for a single image, each channel with its own `ImageKey`? The textureHost sounds to me like a good place to abstract this distinction between single and multi-channeled images.

> Another reason it that we will not just use a Y channel for drawing. We need all Y, U and V channel together. If we use a single ImageKey to contain all of them, we could just remove all channels with that single ImageKey.

Right. If textureHost is one hiding the YUV channels, it can guarantee that they are always used and deleted together.

JerryShih:
> > Because it is treated as a single compositable unit, gecko just use a ImageKey for that. If we use multiple ImageKeys for all channel in video frame, it will break the concept of compositable.
> 
> I'm still missing something here. Why can't a compositable be represented by a set of channels for a single image, each channel with its own ImageKey? The textureHost sounds to me like a good place to abstract this distinction between single and multi-channeled images.

Yes, we could use different ImageKey for all channel. But, when we try to get the channel data from gecko, we should use more information to map the ImageKey to a specific channel data in a textureHost.

e.g.
There is a NV12 format textureHost A.
imageKey1 = generateKey(); // for Y channel in textureHostA
imageKey2 = generateKey(); // for CbCr channel in textureHostA
add_image(imageKey1,...);
add_image(imageKey2,...);
When, WR try to resolve the external image using the key1 and key2.
All key1 and key2 should map to the same textureHostA. Then, we need another map to make the key1 map to channel0(the Y channel) and key2 to channel1(the CbCr channel).

And this is what I would like to do in: https://github.com/servo/webrender/issues/1001
imageKey1 = generateKey(); // generate a key for the textureHostA
add_image(imageKey1, channel0, ...);
add_image(imageKey1, channel1, ...);
With the key and channel information, it's much easier to get the proper channel data.
The external-image handler with key and channel is ready now.

Is this concept still not reasonable?

JerryShih:
We can't have a direct simple mapping from a key1 to a channel 0 of textureHost because we don't have a abstract class for "a single channel". So, both key1 and key2 should map to the same textureHost. Then, we need a second mapping to map the key1 or key2 to the channel data. This second map could be in the textureHost or other place. We could do that, but it's not so clear.

kvark:
So the `textureHost` knows about both `key1` and `key2`. It has an association of "channel" -> `ImageKey`, and at some point it may need to do the reverse association of `ImageKey` -> "channel". This seems like a doable thing to me, obviously depending on internals of textureHost.

If there are no other benefits of having WR to handle multi-channel images, I'd say let's have this logic in textureHost then.

nical:
I think that it boils down to where we want to bend the current code: Gecko tends to think of yuv images as one thing (for example a single TextureHost object), while webrender is organized around things that have individual items in the texture cache (kinda, but not quite, for example tiles) which gives a separate image per plane.
I can see the appeal of having a single Image for all planes, after all there's no reason for yuv planes to be mixed between different images, have different life cycles, etc. But the difference is not fundamental, so if it makes things needlessly complicated in wr we can easily adapt the gecko side of things, either approach can be implemented without too much trouble.

It's fine by me either way but I am confused because we already landed some work in the direction of having a single image key with plane indices: https://github.com/servo/webrender/commit/d733af2c32e8fc4c9daee59020957b838cd090e6 so I thought there was a consensus around this already.

kvark:
As per Skype discussion, @JerryShih is going to try implementing the YUV channel matching logic on the Gecko side, with potentially even reducing the `ExternalImageData` complexity as a follow-up. We'll see how difficult it is and go from there.

JerryShih:
The multiple-channel problem was handled by gecko.
https://bugzilla.mozilla.org/show_bug.cgi?id=1364922

PR #1238, opened 2017-05-11
Revamp the multi-threaded font renderer.

nical:
This is a work in progress. I am trying to address a few things here:

 - The way the glyph cache adds and removes font templates is terribly inefficient: it basically schedules N jobs (N being the number of workers) and each job waits for all of the other jobs to complete in order to insure that each worker gets exactly one of these jobs to do some thread-local storage manipulation. Bad. If the thread pool is busy, we wait for all jobs in progress to complete, then wait for all add/remove jobs to be processed on each worker, and then we can start doing useful things again. Right now it is not very noticeable because we use a thread pool dedicated to glyph rendering for each WebRender instance, but we want to share this thread pool between instances and use it for blob image rendeing as well.
 - The way things currently work is more complicated than it needs to be, It has an extra thread that does nothing other than posting jobs to the thread pool, and requires the glyph cache to move between threads. The latter means we can't easily expose the glyph cache to the blob image renderer. The extra thread isn't necessary. At best it adds a bit latency, and it is written in a way that makes it hard to share it between several webrender instances.
 - the way jobs are dispatch does not let us take advantage of the most efficient way to schedule work on the thread pool (rayon's fork-join API).
 - resource_cache.rs is quite big. It would be nice to move the font rendering bits to a separate file.

How the new font renderer works:

Font contexts are not in thread-local storage anymore, they are in a vector accessible to the workers, accessed with a worker id (provided by rayon), and protected by a mutex. each font context is almost only accessed from its dedicated worker so there should not be contention on the muteces (nothing like the current system that stops all threads anyway). It would be easy to make them really only accessed by their worker if need be.

There is no glyph cache thread anymore. if something needs to be done off the render backend thread, just schedule a job on the thread pool to do it (for example look at add_font). The glyph cache does not need to move around anymore. glyphs are just added to it in resolve_glyphs which corresponds to what we have in end_frame currently.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1238)

### DIFF
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -340,3 +340,56 @@ struct GlyphRasterJob {
     request: GlyphRequest,
     result: Option<RasterizedGlyph>,
 }
+
+#[test]
+fn raterize_200_glyphs() {
+    // This test loads a font from disc, the renders 4 requests containing
+    // 50 glyphs each, deletes the font and waits for the result.
+
+    use rayon::Configuration;
+    use std::fs::File;
+    use std::io::Read;
+
+    let workers = Arc::new(ThreadPool::new(Configuration::new()).unwrap());
+    let mut glyph_rasterizer = GlyphRasterizer::new(workers);
+    let mut glyph_cache = GlyphCache::new();
+
+    let mut font_file = File::open("../wrench/reftests/text/VeraBd.ttf").expect("Couldn't open font file");
+    let mut font_data = vec![];
+    font_file.read_to_end(&mut font_data).expect("failed to read font file");
+
+    let font_key = FontKey::new(0, 0);
+    glyph_rasterizer.add_font(font_key, FontTemplate::Raw(Arc::new(font_data), 0));
+
+    let frame_id = FrameId(1);
+
+    let mut glyph_instances = Vec::with_capacity(200);
+    for i in 0..200 {
+        glyph_instances.push(GlyphInstance {
+            index: i, // It doesn't matter which glyphs we are actually rendering.
+            point: LayoutPoint::new(0.0, 0.0),
+        });
+    }
+
+    for i in 0..4 {
+        glyph_rasterizer.request_glyphs(
+            &mut glyph_cache,
+            frame_id,
+            font_key,
+            Au::from_px(32),
+            ColorF::new(0.0, 0.0, 0.0, 1.0),
+            &glyph_instances[(50 * i)..(50 * (i + 1))],
+            FontRenderMode::Subpixel,
+            None,
+        );
+    }
+
+    glyph_rasterizer.delete_font(font_key);
+
+    glyph_rasterizer.resolve_glyphs(
+        frame_id,
+        &mut glyph_cache,
+        &mut TextureCache::new(4096),
+        &mut TextureCacheProfileCounters::new(),
+    );
+}
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -1,0 +1,342 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use app_units::Au;
+use device::TextureFilter;
+use frame::FrameId;
+use platform::font::{FontContext, RasterizedGlyph};
+use profiler::TextureCacheProfileCounters;
+use rayon::ThreadPool;
+use rayon::prelude::*;
+use resource_cache::ResourceClassCache;
+use std::sync::{Arc, Mutex, MutexGuard};
+use std::sync::mpsc::{channel, Receiver, Sender};
+use std::collections::HashSet;
+use std::mem;
+use texture_cache::{TextureCacheItemId, TextureCache};
+use internal_types::FontTemplate;
+use webrender_traits::{FontKey, FontRenderMode, ImageData, ImageFormat};
+use webrender_traits::{ImageDescriptor, ColorF, LayoutPoint};
+use webrender_traits::{GlyphKey, GlyphOptions, GlyphInstance, GlyphDimensions};
+
+pub type GlyphCache = ResourceClassCache<GlyphRequest, Option<TextureCacheItemId>>;
+
+pub struct FontContexts {
+    // These worker are mostly accessed from their corresponding worker threads.
+    // The goal is that there should be no noticeable contention on the muteces.
+    worker_contexts: Vec<Mutex<FontContext>>,
+
+    // This worker should be accessed by threads that don't belong to thre thread pool
+    // (in theory that's only the render backend thread so no contention expected either).
+    shared_context: Mutex<FontContext>,
+
+    // Stored here as a convenience to get the current thread index.
+    workers: Arc<ThreadPool>,
+}
+
+impl FontContexts {
+    /// Get access to the font context associated to the current thread.
+    pub fn lock_current_context(&self) -> MutexGuard<FontContext> {
+        let id = self.current_worker_id();
+        self.lock_context(id)
+    }
+
+    /// Get access to any particular font context.
+    ///
+    /// The id is ```Some(i)``` where i is an index between 0 and num_worker_contexts
+    /// for font contexts associated to the thread pool, and None for the shared
+    /// global font context for use outside of the thread pool.
+    pub fn lock_context(&self, id: Option<usize>) -> MutexGuard<FontContext> {
+        match id {
+            Some(index) => self.worker_contexts[index].lock().unwrap(),
+            None => self.shared_context.lock().unwrap(),
+        }
+    }
+
+    /// Get access to the font context usable outside of the thread pool.
+    pub fn lock_shared_context(&self) -> MutexGuard<FontContext> {
+        self.shared_context.lock().unwrap()
+    }
+
+    // number of contexts associated to workers
+    pub fn num_worker_contexts(&self) -> usize {
+        self.worker_contexts.len()
+    }
+
+    fn current_worker_id(&self) -> Option<usize> {
+        self.workers.current_thread_index()
+    }
+}
+
+pub struct GlyphRasterizer {
+    workers: Arc<ThreadPool>,
+    font_contexts: Arc<FontContexts>,
+
+    // Receives the rendered glyphs.
+    glyph_rx: Receiver<Vec<GlyphRasterJob>>,
+    glyph_tx: Sender<Vec<GlyphRasterJob>>,
+
+    // Maintain a set of glyphs that have been requested this
+    // frame. This ensures the glyph thread won't rasterize
+    // the same glyph more than once in a frame. This is required
+    // because the glyph cache hash table is not updated
+    // until the end of the frame when we wait for glyph requests
+    // to be resolved.
+    pending_glyphs: HashSet<GlyphRequest>,
+
+    // We defer removing fonts to the end of the frame so that:
+    // - this work is done outside of the critical path,
+    // - we don't have to worry about the ordering of events if a font is used on
+    //   a frame where it is used (although it seems unlikely).
+    fonts_to_remove: Vec<FontKey>,
+}
+
+impl GlyphRasterizer {
+    pub fn new(workers: Arc<ThreadPool>) -> Self {
+        let (glyph_tx, glyph_rx) = channel();
+
+        let num_workers = workers.current_num_threads();
+        let mut contexts = Vec::with_capacity(num_workers);
+
+        for _ in 0..num_workers {
+            contexts.push(Mutex::new(FontContext::new()));
+        }
+
+        GlyphRasterizer {
+            font_contexts: Arc::new(
+                FontContexts {
+                    worker_contexts: contexts,
+                    shared_context: Mutex::new(FontContext::new()),
+                    workers: Arc::clone(&workers),
+                }
+            ),
+            glyph_rx: glyph_rx,
+            glyph_tx: glyph_tx,
+            pending_glyphs: HashSet::new(),
+            workers: workers,
+            fonts_to_remove: Vec::new(),
+        }
+    }
+
+    pub fn add_font(&mut self, font_key: FontKey, template: FontTemplate) {
+        let font_contexts = Arc::clone(&self.font_contexts);
+        // It's important to synchronously add the font for the shared context because
+        // we use it to check that fonts have been properly added when requesting glyphs.
+        font_contexts.lock_shared_context().add_font(&font_key, &template);
+
+        // TODO: this locks each font context while adding the font data, probably not a big deal,
+        // but if there is contention on this lock we could easily have a queue of per-context
+        // operations to add and delete fonts, and have these queues lazily processed by each worker
+        // before rendering a glyph.
+        // We can also move this into a worker to free up some cycles in the calling (render backend)
+        // thread.
+        for i in 0..font_contexts.num_worker_contexts() {
+            font_contexts.lock_context(Some(i)).add_font(&font_key, &template);
+        }
+    }
+
+    pub fn delete_font(&mut self, font_key: FontKey) {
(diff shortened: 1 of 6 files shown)


nical:
The general idea/structure is implemented but I have still improvements to make before I consider it reviewable. Interested parties can have a look, I'll happily gather feedback on the high level approach.

As a side note, the code is formatted according to the new style (mostly talking about [block indents](https://github.com/rust-lang-nursery/fmt-rfcs/issues/8)), rather than the old one, if this hurts anyone's feelings, let's agree on the nature of the change and then have a style discussion, I can easily reformat this later.

cc @jrmuizel @glennw @kvark

kvark:
I wonder if the current situation with stopping all threads can be addressed as a fix, rather than a total rework. In this case, having thread-local `FontContext` would appear faster on access (than locking a `Mutex` each time) and even less hacky (since your approach also forces 1:1 relationship between contexts and threads, but does it more softly).

nical:
There's plenty of reasons to want a rewrite of this (not so large) piece of code. For example the glyph cache thread is unnecessary, it's just one thread to wait to wake up which adds a bit of latency if you have plenty of threads (gecko has like 60 of them per process). If you look at the FontRenderer, I can simply write unit tests for it while I can't do that with the current code. I also think it is much simpler (but I am partial because  wrote it).

Muteces are dirt cheap. What's not cheap is mutex contention. The way the FontRenderer is written does not have much risk of contention except when adding a font template while the workers are buzy rendering glyphs (very unlikely in fact, because we add the font templates before starting to request glyphs). We could completely remove the need to access a font context outside of its dedicated worker by having a channel per context to enqueue the additions and removals of fonts and process that lazily next time a glyph is rendered on the worker, but that's more complex and I would be surprised if it ends up faster. 

It probably personal but I don't like using thread local storage. For one it's very easy to forget to clean it up (If you destroy the font renderer whatever you have left in tls just stays there until the thread dies), and it's needlessly inflexible (you can't access it unless you are in that thread). having an ```Arc<Vec<Context>>``` means the contexts get cleaned up automatically if the font renderer is destroyed, and if we need it (or in cases where it is simpler) we can easily access the context through its id from wherever we want. And more generally it's a mutable (almost) global state with all of the badness that comes with.

I'm not sure what's hacky about the the approach I propose. Muteces are not sexy that's for sure, but resorting to tls is what really fills hacky to me. Maybe you can expand on that and I'll iterate towards something cleaner (it's not quite there yet).

glennw:
@nical Thanks for starting to look at this! Making changes similar to this (or at least, with the same end goal) has been on my low priority list for a while - it's great to get a start on this!

webrender/src/font_renderer.rs
```diff
@@ -0,0 +1,367 @@
...
+            fonts_to_remove: Vec::new(),
+        }
+    }
+
+    pub fn add_font(&mut self, font_key: FontKey, template: FontTemplate) {
+        let font_contexts = Arc::clone(&self.font_contexts);
+        // It's important to synchronously add the font for the shared context because
+        // we use it to check that fonts have been properly added when requesting glyphs.
+        font_contexts.lock_shared_context().add_font(font_key, &template);
+        self.workers.spawn_async(move || {
```
nical: Note: I am thinking of just doing this from the render backend thread instead of spawning an async job, in order to keep it simple. FontContext::add_font should be cheap (and if it is not we can easily make it so), and that way we have the guarantee that fonts are added for the subsequent glyph requests without relying on any ordering in the thread pool. otherwise we need to account for the possibility that a the thread pool process tasks out of order and run rasterize_glyph before adding the font.

glennw:
@nical I haven't looked at this PR yet at all - but wanted to add a general comment anyway.

There is (at least) one major feature the existing glyph cache code doesn't handle, which we will need to deal with. It would be great if we handle it as part of this work, but if not, we should at least think about how it would fit in when done as a follow up.

What I'd like to do is make the render backend thread have a time threshold, where it says only this much time per frame is spent on rasterizing glyphs. And if that time is exceeded, we give up on pending glyph requests and rasterize the frame with the glyphs that are currently available (i.e. pick the closest size and deal with it being slightly blurry).

The use case for this is pinch-zoom - we don't want to slow down the framerate just because we're rasterizing a heap of glyphs per frame.

Of course, that's one idea on how to implement it. If there's other ways to implement the same end result, that's worth discussing too.

nical:
@glennw yes, we'll need something like this for blob image renderer too because it will most likely involve things that are even more expensive to paint than glyphs.
I have not tackled that in this PR in order to avoid scope creep but that would be next on my list.

A simple way to do this is to just bail out of the ```while !self.pending_glyphs.is_empty()``` loop in resolve_glyphs after a certain threshold. Ideally we'd avoid blocking on a recv call and instead do some try_recv polling and steal work form the thread pool as much as possible to avoid having the render backend thread de-scheduled.

On top of that we can look into prioritizing certain glyphs over others but I would like to see if we can do without at first, because that will require some extra complexity that I'd rather avoid. I implemented a system like that in gecko and what worked best was for each worker to have separate queues for low and high priority jobs and have the workers always pull from the high priority ones first, but rayon doesn't do that so we'd have to add some extra queues on top of the ones rayon uses internally.

glennw:
@nical Yup that all sounds about the same as what I was thinking / planning. Great! :)

nical:
The code is starting to be in a pretty good shape. The PR currently contains a simple test which I temporarily turned into a benchmark. The latter helped me verify that using a ```Sender<Vec<GlyphRasterJob>>``` per glyph run was quite a bit faster than having a ```Sender<GlyphRasterJob>``` per worker which would send each glyph individually (see commit " 	Batch sending GlyphRasterJobs back to the render backend thread").

I'll remove the benchmark after I am done with the PR. It's a bit hacky and noisy, and unless we are going to watch it after each push it's not worth adding seconds of processing to the CI. It's also pretty easy to hack it up on top of the test which will remain.

nical:
The PR is in a good shape now, we can proceed with reviews. There are improvements and features that we can add on top of it but I'd rather do these in followups. Since there may have been eunexpected performance effects from moving to rayon, I checked locally that this PR does not regress the time the it takes for servo to run its css tests. The commits are pretty standalone in case it helps with the reviews, even for the merge I'd prefer to keep them separate.

r? @kvark @glennw

glennw:
Reviewed 1 of 6 files at r1, 3 of 5 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---

*[webrender/src/font_renderer.rs, line 144 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-KkhMJ41srjTYD1aNVl8:-KkhMJ42hvW_RW1S6EkM:b-uu4kn4) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/font_renderer.rs#L144)):*
> ```Rust
> 
>     pub fn request_glyphs(
>         &mut self,
> ```

Indentation here seems a bit weird

---

*[webrender/src/platform/macos/font.rs, line 29 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-KkgxxPs9Y163xyVERNP:-KkgxxPtraUvq_AMoTjw:b-5gj50r) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/macos/font.rs#L29)):*
> ```Rust
> }
> 
> unsafe impl Send for FontContext {}
> ```

Same as the comment for windows file.

---

*[webrender/src/platform/macos/font.rs, line 81 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-Kkgy3be3CpbMYijmxy5:-Kkgy3be3CpbMYijmxy6:b-naug1x) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/macos/font.rs#L81)):*
> ```Rust
>     let bounds = ct_font.get_bounding_rects_for_glyphs(kCTFontDefaultOrientation, &[glyph]);
> 
>     if bounds.origin.x.is_nan() || bounds.origin.y.is_nan() ||
> ```

Instead of returning a zero glyph metrics, perhaps this should return a Result to indicate failure?

---

*[webrender/src/platform/unix/font.rs, line 30 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-KkgxlEYZtieSjVZbSiz:-KkgxlEYZtieSjVZbSj-:b-5gj50r) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/unix/font.rs#L30)):*
> ```Rust
> }
> 
> unsafe impl Send for FontContext {}
> ```

Same as the comment for windows file.

---

*[webrender/src/platform/unix/font.rs, line 180 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-KkgxqFbtpaJJb9UdNZi:-KkgxqFbtpaJJb9UdNZj:b-5gj50r) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/unix/font.rs#L180)):*
> ```Rust
>         }
> 
>         let dimensions = match Self::get_glyph_dimensions_impl(slot) {
> ```

Same as the comment for windows file.

---

*[webrender/src/platform/windows/font.rs, line 27 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-Kkgxc2Ui1E8hl9PqytG:-Kkgxc2Ui1E8hl9PqytH:b-p0k0xf) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/windows/font.rs#L27)):*
> ```Rust
> }
> 
> unsafe impl Send for FontContext {}
> ```

Could you add a comment here explaining why this is required, and how we know it is safe?

---

*[webrender/src/platform/windows/font.rs, line 282 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-KkgxREXQVqoTYVW3TPb:-KkgxREYHULiJ6YIUn4D:bp9kex5) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/windows/font.rs#L282)):*
> ```Rust
>         // Alpha texture bounds can sometimes return an empty rect
>         // Such as for spaces
>         if width == 0 || height == 0 {
> ```

Did you hit this assert during testing? The idea here is that since get_glyph_dimensions() would return None in that case, the rasterize would never be called.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1238)*

glennw:
This generally looks good to me - a few minor comments / questions added.

glennw:
@nical I tried to test this a bit locally. There seems to be some kind of resource leak though - if you visit HN in Servo with this, and move the mouse, you'll see the CPU time in the render backend continuously increase.

glennw:
@nical It would be great to get this finished up and merged - have you had time to look into the apparent resource leak and comments above?

nical:

---

*[webrender/src/font_renderer.rs, line 144 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-KkhMJ41srjTYD1aNVl8:-KlJQHJKoNMglDri9ux6:bxwgewb) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/font_renderer.rs#L144)):*

Indentation here seems a bit weird
</blockquote></details>

It is the "new" recommended rust style guide (which switched form visual indentation to block indentation for function call and definition).

---

*[webrender/src/platform/macos/font.rs, line 81 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-Kkgy3be3CpbMYijmxy5:-KlJREGcn5-U-shj1k5b:bnw32jr) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/macos/font.rs#L81)):*

Instead of returning a zero glyph metrics, perhaps this should return a Result to indicate failure?
</blockquote></details>

Since I don't have a mac to test this type of change on, I'd rather modify platform-dependent files as little as I can in the big pull request. We can do look into this as a followup if that's alright with you.

---

*[webrender/src/platform/windows/font.rs, line 282 at r5](https://reviewable.io:443/reviews/servo/webrender/1238#-KkgxREXQVqoTYVW3TPb:-KlJRfJQhgmROVtD2RuB:bvfl367) ([raw file](https://github.com/servo/webrender/blob/f58d9907367a126bb98df1c35bb81398bff08e50/webrender/src/platform/windows/font.rs#L282)):*

Did you hit this assert during testing? The idea here is that since get_glyph_dimensions() would return None in that case, the rasterize would never be called.
</blockquote></details>

I am hitting this in my test case, which asks webrender to render lots of glyphs with indices between 0 and 200. It didn't look like get_glyph_dimensions was called in a way that would prevent this code from running. Is it that servo calls get_glyph_dimensions before submitting the glyphs (which my test case does not do)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1238)*

nical:
> I am hitting this in my test case, which asks webrender to render lots of glyphs with indices between 0 and 200.

erratum: asking the glyph renderer directly, not webrender.

glennw:
@nical I agree about changing as little platform specific code as possible in this PR. If you change your test case to only contain valid glyph indices, couldn't we remove all those platform-specific changes (in regards to invalid glyph widths etc) in this PR and do those as a separate follow up?

I did some more testing with this today. I can't reproduce the apparent resource leak I mentioned above (did you change something that should affect that?). However, I did come across a few bad CPU stalls that only appear with this PR.

In the image below, I've highlighted one of them. In some cases this causes a CPU stall of ~100ms which is really noticeable. The easiest way to reproduce is to load HN in Servo, then move the mouse around for about 10 seconds at the top of the page (to expire old cached glyphs), and then quickly scroll down to the bottom of the page. On my machine, this reproduces that stall 100% of the time. I confirmed that it occur when it rasterizes glyphs, but didn't look further into it. Perhaps a mutex stall or something like that? Could you see if you can repro that locally?

stall

nical:
I hope that it was because I had mixed up the ordering and called mark_as_needed only for glyphs that were not in the cache (causing glyphs to expire super quickly while they are in use), it's fixed in one of my latest commits, not sure if your last attempt at reproducing had it.
For the test case, only requesting valid glyphs requires the test to properly load fonts and query glyph indices (I assume with platform specific code), so prefer to harden the font rendering code so that it can handle a somewhat malicious test case.
I can remove the test case along with the relevant portion of the platform specific code from this PR and submit it in a followup if you prefer.

nical:
I'll look into the cpu stall late tomorrow or the day after, I'll have a computer with me that can build servo (relatively) quickly.

glennw:
Ah, good point about the glyph indices. If it makes it easier to get this landed, let's just leave that as it is now.

It would be good to take a look at that stall before landing this though. Once we have a fix for that stall, this should be good to merge!

nical:
I reproduced and fixed the render backend stall.
Fun story: most of the time request_glyphs is called all of the glyphs are already in the cache, but I was still dispatching an empty glyph request to the thread pool each time. Since we don't end up waiting for these glyphs as they are not in the set of pending glyph requests, we'd move on to the next frame without waiting for these empty requests to be scheduled. After a few frames of running the render backed thread at 60fps, we'd have accumulated a ton of requests in the thread pool, and the next time we'd actually have to wait for a glyph to be rendered, the thread pool would be very busy with the a gazillion of noops and spend a long time getting through them before rendering our glyph.

I think that it comes down to:
 - dispatching a task to the thread pool with spawn_async is not cheap, rayon's source code (or was it the docs) says as much.
 - and/or having tons of very ridiculously small items in the queue is a pathological case because worker threads stomp on each other while work-stealing as they all spin on that one queue which receives jobs from outside the thread pool (that was also my experience when I worked on similar stuff for gecko last year).

nical:
I cleaned up the patch queue, squashed the commits about glyph rasterization into one (but kept the other commits separate), and renamed the "font renderer" into "glyph rasterizer" since, as @kvark pointed out to me a little while back, it better describes what this is about.

Other than that, no outstanding change since the fix to the stall.

glennw:
Nice find!

@bors-servo r+

PR #1242, opened 2017-05-12
Rework the border edge shader to workaround a hang in LLVM 3.8 during shader compilation.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1242)

### DIFF
--- a/webrender/res/ps_border_edge.vs.glsl
+++ b/webrender/res/ps_border_edge.vs.glsl
@@ -108,6 +108,9 @@ void main(void) {
     // TODO(gw): Now that all border styles are supported, the switch
     //           statement below can be tidied up quite a bit.
 
+    float style;
+    bool color_flip;
+
     RectWithSize segment_rect;
     switch (sub_part) {
         case 0: {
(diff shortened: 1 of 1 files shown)


glennw:
r? @kvark 
cc @staktrace

nical:
This looks like a good candidate for the [driver issues list](https://github.com/servo/webrender/wiki/Driver-issues), @glennw could you give a some details about what the source of the issue is? Is it inlining that ends up generating a lot for the compiler to chew internally?

staktrace:
I managed to get the latest WR building with gecko, so I applied this PR on top of that to see if it fixes the issues. I just kicked it off so results aren't in yet but they will be [here](https://treeherder.mozilla.org/#/jobs?repo=try&revision=96fabc04fb70d7f8d47c8de0f8424665386fb211). For reference [here](https://treeherder.mozilla.org/#/jobs?repo=try&revision=f46bcb059cdd2ff1e3ec4256de0e83746bee3c4b) is the try push with the current WR tip for a comparison baseline.

kvark:
It would be great to have a shiny comment there warning to be careful when shuffling things around. Can be made a follow-up.
@bors-servo r+

staktrace:
So yes, this PR definitely helped. There's a handful of failures still but I'm not sure which patch they're coming from.

PR #1246, opened 2017-05-12
Make push_text accept an f32 for blur_radius.

### LINKED ISSUE #1244: DisplayListBuilder function push_text should accept float for blur_radius
[The push_text method](http://doc.servo.org/webrender_traits/struct.DisplayListBuilder.html#method.push_text) takes `Au` for `blur_radius` unlike `push_box_shadow` which takes a `f32`. The `Au` is always converted to `f32` internally. For consistency can I change the accepted type?

pyfisch:
Necessary changes to internal structures.

This is a breaking change.

Closes #1244

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1246)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -373,7 +373,7 @@ fn main() {
                           font_key,
                           ColorF::new(1.0, 1.0, 0.0, 1.0),
                           Au::from_px(32),
-                          Au::from_px(0),
+                          0.0,
                           None);
     }
 
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -741,7 +741,7 @@ impl FrameBuilder {
                     clip_region: &ClipRegion,
                     font_key: FontKey,
                     size: Au,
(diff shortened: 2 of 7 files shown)


kvark:
Would be great to get some Servo folks opinion on what approach is more idiomatic. cc @emilio @glennw

pyfisch:
Updated wrench. Do the yaml files also need updates?

kvark:
@pyfisch given that `Au` is converted to `f32` by division by 60, and we used to specify Au in YAML wrench tests, I think we need to go though them and divide by 60 manually.

glennw:
Yup, having the APIs all take ```f32``` makes sense to me. Thanks!

pyfisch:
The yaml files do not need an update since `as_px_to_au` is just converting from `f32`.

    fn as_px_to_au(&self) -> Option<Au> {
        match self.as_force_f32() {
            Some(fv) => Some(Au::from_f32_px(fv)),
            None => None
        }
    }

PR #1247, opened 2017-05-12
Fix box shadows that have a spread radius and no blur radius with a border radius.

changm:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1247)


webrender/src/frame_builder.rs
```diff
@@ -951,6 +951,7 @@ impl FrameBuilder {
                 }
             }
             BoxShadowKind::Shadow(rects) => {
+                assert!(blur_radius > 0.0);
```
kvark: It will fire if both `blur_radius == 0.0` and `border_radius == 0.0`. Where is this case handled?

changm: It breaks down into a simple rect below - https://github.com/servo/webrender/blob/master/webrender/src/frame_builder.rs#L910

We can probably merge the two cases. I'll do that in a follow up. I'm just going through the gecko reftests.

changm:
How do I make a smaller ref png? I'm just doing cargo run png

kvark:
> How do I make a smaller ref png? I'm just doing cargo run png

Ok, that's fine then.
You also need to add the test to the `reftest.list` file.

kvark:
Thanks!
@bors-servo r+

Issue #1219, opened 2017-05-09
Semi-transparent gradient stops rendered opaque


pyfisch:
I have made a few test-cases for gradients: https://pyfisch.org/stuff/testcases-gradients.html

They render fine in Servo with Webrender except for testcase 15 where I use two stops with a value of `rgba(255, 255, 255, .5)`. This should be translucent but is rendered as solid white with servo. (The intersection of the two gradients should be brighter then the rest)

Since I did not find a point where the alpha value is "forgotten" in servo I looked in webrender and found this file: https://github.com/servo/webrender/blob/8516d6c04235e684d9bf9c783ba4fc99dab3bf02/webrender/res/ps_gradient.fs.glsl It appears to min/max alpha values. Maybe this causes semi-transparent stops to be rendered as opaque?

kvark:
cc @rlhunt

rlhunt:
Does Servo premultiply the alpha of the gradient color stops? Gecko does [this](http://searchfox.org/mozilla-central/source/layout/painting/nsCSSRenderingGradients.cpp#1015) before the display list is given to WebRender. This could be done on the WebRender [side](https://github.com/servo/webrender/blob/master/webrender_traits/src/display_list.rs#L670) but isn't. It doesn't really matter too much either way.

pyfisch:
Servo does not premultiply the alpha of gradient stops. Is it possible to send a PR to webrender to premultiply gradient stops or should this be changed in servo?

mstange:
Is this fixed now?

rlhunt:
This should be fixed now.

pyfisch:
Yes it is.

PR #1250, opened 2017-05-12
Premultiply gradient stop colors

rlhunt:
CSS gradient stop colors are interpolated in premultiplied space.
Currently we assume that color stops that are passed in are
premultiplied, but because it's needed in both Servo and Gecko, it makes
sense to do the premultiplication in WebRender and expose a straight
alpha color like the rest of the API does.

The premultiplication is purposefully not done in DisplayListBuilder
to prevent it from being saved for replay and then premultiplied again
when we replay the display list.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1250)


rlhunt:
r? @glennw @kvark 
cc @pyfisch

This is for #1219 

Note: this was done on the Gecko side before, but Servo needs it too, so I think it makes sense to make the API less confusing and move the work here.

kvark:
Since we are using premultiplied alpha for gradients, shouldn't we also change the blending mode used for rendering of them?

rlhunt:
Sorry, forgot to respond. The gradients should be using a premultiplied alpha blend mode, as before. The change here is to make the API expect unpremultiplied alpha, instead of expecting the client to premultiply the stops.

rlhunt:
@kvark Thanks!

PR #1252, opened 2017-05-15
Fix binary recording and replay.

glennw:
* Add SetWindowParameters to the list of messages that get serialized.
* Ensure we wait for the three initial messages in correct order to detect the first frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1252)


jrmuizel:
This gets broken often enough that it's probably worth adding some kind of testing...

glennw:
@jrmuizel Agreed - opened https://github.com/servo/webrender/issues/1253

kvark:
Thank you!
@bors-servo r+

PR #1255, opened 2017-05-15
Enable gl::XXX_BITS constants in WebGL getParameter command.

MortimerGoro:
Required to pass some WebGL conformance tests. See https://github.com/servo/servo/pull/16867

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1255)

### DIFF
--- a/webrender_traits/src/webgl.rs
+++ b/webrender_traits/src/webgl.rs
@@ -688,20 +688,20 @@ impl WebGLCommand {
                  chan: MsgSender<WebGLResult<WebGLParameter>>) {
         let result = match param_id {
             gl::ACTIVE_TEXTURE |
-            //gl::ALPHA_BITS |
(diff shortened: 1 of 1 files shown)


emilio:
@glennw blame points at you for these commented lines (https://github.com/servo/webrender/commit/46a7cb1c2f6e22569f5f75b2135ae71bfa16c9b3).

Do you remember why was that needed?

If we can uncomment these now, r=me

glennw:
@emilio Some of those were commented due to just not being used. But some of them are not available in ES2, and cause compile errors on android. We now rely on ES3 on android, so *some* of those may no longer be an issue - but some of them are completely deprecated from ES, I think.

glennw:
@emilio It does look like all the ones that have been un-commented here should be available on Android though, I think!

Issue #1254, opened 2017-05-15
A lot of missing text-related features in wrench on non-windows platforms.


nical:
A lot of missing implementations, see for example in wrench.rs:
 - ```layout_simple_ascii```
 - ```font_key_from_name```

Right now it is hard to test any text-related feature on non-windows platforms using wrench.

jrmuizel:
font_key_from_name is unused but a patch to implement it is easy:
```patch
diff --git a/wrench/src/wrench.rs b/wrench/src/wrench.rs
index 05da112..34843c1 100644
--- a/wrench/src/wrench.rs
+++ b/wrench/src/wrench.rs
@@ -276,19 +276,21 @@ impl Wrench {
     #[cfg(any(target_os = "linux", target_os = "macos"))]
     pub fn font_key_from_yaml_table(&mut self, item: &Yaml) -> (FontKey, Option<NativeFontHandle>) {
         let family = item["family"].as_str().unwrap();
         let property = system_fonts::FontPropertyBuilder::new().family(family).build();
         let (font, index) = system_fonts::get(&property).unwrap();
         self.font_key_from_bytes(font, index as u32)
     }
 
-    #[cfg(not(target_os = "windows"))]
-    pub fn font_key_from_name(&mut self, _font_name: &str) -> (FontKey, Option<NativeFontHandle>) {
-        panic!("Can't font_key_from_name on this platform");
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    pub fn font_key_from_name(&mut self, font_name: &str) -> (FontKey, Option<NativeFontHandle>) {
+        let property = system_fonts::FontPropertyBuilder::new().family(font_name).build();
+        let (font, index) = system_fonts::get(&property).unwrap();
+        self.font_key_from_bytes(font, index as u32)
     }
 
     pub fn font_key_from_bytes(&mut self, bytes: Vec<u8>, index: u32) -> (FontKey, Option<NativeFontHandle>) {
         let key = self.api.generate_font_key();
         self.api.add_raw_font(key, bytes, index);
         (key, None)
     }
 
```

layout_simple_ascii is more involved. What I've been doing for text-releated things is just recording in servo or firefox and then using that. It would be better to have layout_simple_ascii though.

glennw:
I am planning to hook harfbuzz up for linux - it should be very simple to do. But we'll need someone to do a mac implementation of layout_simple_ascii then.

glennw:
[...]

PR #1257, opened 2017-05-16
Implement wrench font_key_from_name on mac and linux.

nical:
The code was actually written by Jeff (see issue #1254).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1257)

### DIFF
--- a/wrench/src/wrench.rs
+++ b/wrench/src/wrench.rs
@@ -281,9 +281,16 @@ impl Wrench {
         self.font_key_from_bytes(font, index as u32)
     }
 
-    #[cfg(not(target_os = "windows"))]
-    pub fn font_key_from_name(&mut self, _font_name: &str) -> (FontKey, Option<NativeFontHandle>) {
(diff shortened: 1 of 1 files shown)


kvark:
Shouldn't this line be changed as well? https://github.com/servo/webrender/pull/1257/files#diff-1821d8a90ad50120b2720d853355623dR248

glennw:
I think we'll need to either (a) include "android" in the list of target config for that function, or (b) keep a panic version of the function around for android.

An external contributor also got WR working on BSD a while ago - instead of using the linux/mac OS strings, something like ```#[cfg(any(target_os = "android", all(unix, not(target_os = "macos"))))]``` would be more appropriate - depending on whether we need a separate function for android.

glennw:
Actually, in this case just using ```unix``` would be fine, I guess - since we want it on both mac, linux and BSD. And then whatever is appropriate for android.

PR #1259, opened 2017-05-16
Let rayon automatically choose the number of threads in the ThreadPool.

nical:
This is a tiny followup from PR #1202 that I wanted to submit in isolation so that we can easily check its effect on performance (if any) and roll it back if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1259)


kvark:
easy, eh? :)
@bors-servo +

kvark:
Ah, I mistyped it.

PR #1261, opened 2017-05-17
Fix resource_cache mark_as_needed() usage.

JerryShih:
r? @glennw @kvark 
cc @nical 
 
After #1225, every image will call mark_as_needed(). Then we will hit the assert for external image in expire_old_resources().

Only the image which uses resource_cache should call mark_as_needed(). The external image doesn't need that call.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1261)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -438,8 +438,10 @@ impl ResourceCache {
             tile: tile,
         };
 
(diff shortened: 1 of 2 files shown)


JerryShih:
The external image doesn't be added into the cache list. So, the entry can't be found in expire_old_resources().

kvark:
Sounds reasonable, thank you!
@bors-servo r+

PR #1262, opened 2017-05-17
 Fix the coordinate system of the clip node in nested reftest.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1262)

### DIFF
--- a/wrench/reftests/scrolling/nested-scroll-offset.yaml
+++ b/wrench/reftests/scrolling/nested-scroll-offset.yaml
@@ -7,11 +7,11 @@ root:
       items:
       - type: scroll-layer
         bounds: [0, 300, 50, 50]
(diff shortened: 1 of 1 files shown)


glennw:
r? @mrobinson  and @kvark 

With the current clip code, the origin of the clip node is ignored.

With some local clipping changes I'm working on, this test started failing when everything else was passing.

I *think* this is because the reftest itself has the wrong coordinate system for the clip nodes. With my local changes, the origin of the clip node is now handled, which makes this test start failing.

This is based on my reading of CLIPPING.md. I wanted to open a PR for this now, so that you can check I'm correct in assuming the reftest itself is the issue, rather than the changes I'm working on locally.

Does that make sense?

mrobinson:
@glennw That sounds correct to me!

PR #1267, opened 2017-05-18
Disable subpixel mode for text runs on composited stacking contexts.

### LINKED ISSUE #1265: Incorrect subpixel AA text rendering with composite operations
See https://github.com/servo/servo/issues/16902
Take #1264 if you need to repro it on linux.

When subpixel AA text is drawn to an offscreen target because of a composite operation, it gets the RGB color messed up. See Frame 200 in the attached capture - [servo-text-aa.zip](https://github.com/servo/webrender/files/1008632/servo-text-aa.zip)

glennw:
Fixes #1265.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1267)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -759,11 +759,25 @@ impl FrameBuilder {
         // subpixel AA!
         let mut render_mode = self.config.default_font_render_mode;
 
-        // If we're using sub-pixel AA by default, but we are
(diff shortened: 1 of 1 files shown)


glennw:
A fix for this is https://github.com/servo/webrender/pull/1267 - once it lands in WR we'll need to also get the update in Servo before the fix is visible.

kvark:
Thank you!
@bors-servo r+

PR #1268, opened 2017-05-18
Show backtraces in case of unit test failure.

nical:
There is probably a more elegant way to do this, I just prefixed with 'RUST_BACKTRACE=1' the ```cargo test``` commands I came across.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1268)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ addons:
       - llvm-3.8-dev
       - libedit-dev
 env:
-  - BUILD_KIND=DEBUG
-  - BUILD_KIND=RELEASE
+  - BUILD_KIND=DEBUG RUST_BACKTRACE=1
+  - BUILD_KIND=RELEASE RUST_BACKTRACE=1
 before_install:
   - pip install mako servo-tidy
 script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,6 @@
 environment:
   PATH: 'C:\msys64\mingw64\bin;C:\msys64\usr\bin;%PATH%;C:\Rust\bin'
+  RUST_BACKTRACE: 1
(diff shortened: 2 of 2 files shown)


kvark:
@nical it would be simple to do
```yaml
env:
  - RUST_BACKTRACE=1
```
Example - https://github.com/slide-rs/hibitset/pull/4/files#diff-354f30a63fb0907d4ad57269548329e3R15

nical:
My limited understanding of travis is that adding a ``` - RUST_BACKTRACE=1``` item in the env section will add another combination of tests to run in the test matrix. I don't see an example where a single item has several environment variables set like ``` - BUILD_KIND=DEBUG, RUST_BACKTRACE=1```, maybe I should try and see.

.travis.yml
```diff
@@ -24,9 +24,9 @@ before_install:
   - pip install mako servo-tidy
 script:
   - servo-tidy
-  - if [ $BUILD_KIND = DEBUG ]; then (cd webrender_traits && cargo test --verbose --features "ipc"); fi
-  - if [ $BUILD_KIND = DEBUG ]; then (cd webrender_traits && cargo test --verbose); fi
+  - if [ $BUILD_KIND = DEBUG ]; then (cd webrender_traits && RUST_BACKTRACE=1 cargo test --verbose --features "ipc"); fi
```
kvark: let's remove `RUST_BACKTRACE=1` from here, given that `env` appears to be working?

nical: oops!

nical:
This version should [work](https://docs.travis-ci.com/user/environment-variables/#Defining-Multiple-Variables-per-Item) more elegantly.

PR #1271, opened 2017-05-18
Coalescing optimization

### LINKED ISSUE #1266: A lot of time spent in webrender::texture_cache::TextureCache::insert
I see this stand out quite a bit in profiles of Firefox scrolling cnn.com and a huge amount scrolling https://perfht.ml/2qxbfmr.

kvark:
Fixes #1266
(if it doesn't fix it yet, the PR would be a step in the right direction anyway)

This PR removes indexing and dynamic allocations from `TexturePage::coalesce`, cleans up the code, adds a basic benchmark for it. Results on my machine - time dropped by about 25%. It's not super great, but I believe that real-world parameters have improved more, given that we are able to capture intermediate coalescing results now even when timed out.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1271)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -181,54 +181,44 @@ impl TexturePage {
 
         // Iterate to a fixed point or until a timeout is reached.
         let deadline = time::precise_time_ns() + COALESCING_TIMEOUT;
-        let mut free_list = mem::replace(&mut self.free_list, FreeRectList::new()).into_vec();
+        let mut free_list = self.free_list.into_vec();
         let mut changed = false;
 
         // Combine rects that have the same width and are adjacent.
         let mut new_free_list = Vec::new();
-        free_list.sort_by(|a, b| {
-            match a.size.width.cmp(&b.size.width) {
-                Ordering::Equal => a.origin.x.cmp(&b.origin.x),
-                ordering => ordering,
-            }
-        });
+        free_list.sort_by_key(|item| (item.size.width, item.origin.x));
         for work_index in 0..free_list.len() {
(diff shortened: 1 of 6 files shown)


webrender/src/texture_cache.rs
```diff
@@ -173,92 +181,98 @@ impl TexturePage {
...
-                    break
-                }
-                if free_list[work_index].origin.y == free_list[candidate_index].max_y() ||
-                        free_list[work_index].max_y() == free_list[candidate_index].origin.y {
-                    changed = true;
-                    free_list[work_index] =
-                        free_list[work_index].union(&free_list[candidate_index]);
-                    free_list[candidate_index].size.width = 0
-                }
-                new_free_list.push(free_list[work_index])
```
kvark: this line looks like a bug to me (adding a work item during the candidates iteration), and I removed it

glennw:
@pcwalton Would you mind taking a look at this one, if you can still recall how the coalesce code works? :)

glennw:
I had a look over this and it seems right to me. Also ran the Servo tests and tried a few sites manually.

@bors-servo r+

staktrace:
@kvark, This PR seems to hit an assertion failure (or something else causing a debug-only panic) when running the gecko reftest `layout/reftests/bugs/1081185-1.html`. The panic happens in `texture_cache::FreeListBin::for_size`, see [full backtrace](https://treeherder.mozilla.org/logviewer.html#?job_id=100843761&repo=try&lineNumber=6967).

kvark:
@staktrace sorry about that! looking now

PR #1274, opened 2017-05-18
Mostly decoupled wrench from GL

### LINKED ISSUE #1269: Remove the direct GL usage from Wrench
Ideally, GL should only be is `device.rs`. Having Wrench depend on it (calling `glViewport, glReadPixels, and other) prevents Szeged team to use it for reftests (see https://github.com/szeged/webrender/issues/17).

kvark:
Fixes #1269
Note: even though `main.rs` of Wrench still depends on GL, it's only used for context initialization, so the original issue is largely addressed by the PR.
Note: this is technically a breaking API change.
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1274)

### DIFF
--- a/webrender/src/lib.rs
+++ b/webrender/src/lib.rs
@@ -140,4 +140,4 @@ extern crate plane_split;
 extern crate gamma_lut;
 
 pub use renderer::{ExternalImage, ExternalImageSource, ExternalImageHandler};
-pub use renderer::{Renderer, RendererOptions};
+pub use renderer::{GraphicsApi, GraphicsApiInfo, Renderer, RendererOptions};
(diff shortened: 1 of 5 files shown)


staktrace:
We were using that gl() function in a few places in gecko: http://searchfox.org/mozilla-central/rev/24c443a440104dabd9447608bd41b8766e8fc2f5/gfx/webrender_bindings/src/bindings.rs#781,784,789 - is there some other way to do this?

kvark:
@staktrace we have to hide GL as an implementation detail, eventually. For Gecko cases, we may need a helper lower level function at the `Renderer` level. I'll add it shortly.

staktrace:
Maybe we can just turn the `read_pixels_rgba8` function into a more generic readback function, so that we can plug it into the above bindings.rs directly. @kvark WDYT?

kvark:
@staktrace yes, that's exactly what I had in mind, see https://github.com/servo/webrender/pull/1278

PR #1275, opened 2017-05-18
Fix an assert that could incorrectly fire with subpixel AA enabled.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1275)


glennw:
Fixes https://github.com/servo/servo/issues/16777

r? @kvark

kvark:
Oh right, I totally forgot that I had to change it locally as well during the investigation.
@bors-servo r+

PR #1281, opened 2017-05-21
Update thread_profiler version.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1281)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -445,11 +445,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "json"
-version = "0.11.5"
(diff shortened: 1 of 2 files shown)


glennw:
Oh, the change in ```thread_profiler``` uses serde_json 1.0 - but it looks like everything else in WR / Servo is on v0.9. @nox Do you know if we're planning to update serde_json in Servo in the near future, or perhaps I'd be best to downgrade the serde_json version used in thread_profiler?

nox:
I intentto update serde and whatnot but was too busy with stylo things.

glennw:
Closing for now - can reopen when the serde update lands.

PR #1282, opened 2017-05-22
Update gl_generator to 0.5.3 to make generated bindings deterministic.

luser:
[I fixed gl_generator](https://github.com/brendanzab/gl-rs/pull/415) to make the bindings it generates deterministic. @jrmuizel had noticed that he was getting a lot of cache misses building with sccache, and this (+ a few sccache fixes) fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1282)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -363,10 +363,10 @@ dependencies = [
 
 [[package]]
 name = "gl_generator"
-version = "0.5.2"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "khronos_api 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "khronos_api 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "xml-rs 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


glennw:
The travis error looks unrelated - but I think there will also be a Cargo.toml entry you'll need to update.

luser:
Really? AFAICT gl_generator is only used transitively via gleam and offscreen_gl_context.

glennw:
Ah, true!

glennw:
Wait, that means you'll want to update the gleam version in the Cargo.toml then, right?

luser:
> Wait, that means you'll want to update the gleam version in the Cargo.toml then, right?

gleam doesn't have a committed Cargo.lock file, it just specifies `gl_generator = 0.5.0` in its Cargo.toml:
https://github.com/servo/gleam/blob/95ea7243cdea4da299080f649f3e0b06127a3334/Cargo.toml#L12

I could submit a PR to bump that, but by default cargo will pick the latest semver-compat version anyway, which would be 0.5.3, so that doesn't seem necessary? We only need to update projects with a checked-in Cargo.lock that has an older version of gl_generator listed.

glennw:
I was thinking more that we should update the gleam version referenced in the WR Cargo.toml. It's not technically necessary, but I think it's good practice. But we can do that as a follow up anyway. Thanks!

@bors-servo r+

PR #1285, opened 2017-05-23
Surround the statement if switch-case block contains declarations.

### LINKED ISSUE #1206: Shader compilation error on some windows hardware

mephisto41:
This fix #1206 .

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1285)

### DIFF
--- a/webrender/res/ps_border_edge.vs.glsl
+++ b/webrender/res/ps_border_edge.vs.glsl
@@ -42,11 +42,15 @@ void write_color(vec4 color, float style, bool flip) {
 
     switch (int(style)) {
         case BORDER_STYLE_GROOVE:
+        {
             modulate = flip ? vec2(1.3, 0.7) : vec2(0.7, 1.3);
             break;
+        }
         case BORDER_STYLE_RIDGE:
+        {
             modulate = flip ? vec2(0.7, 1.3) : vec2(1.3, 0.7);
             break;
+        }
         default:
             modulate = vec2(1.0);
             break;
--- a/webrender/res/ps_border_corner.vs.glsl
+++ b/webrender/res/ps_border_corner.vs.glsl
@@ -91,6 +91,7 @@ int select_style(int color_select, vec2 fstyle) {
 
     switch (color_select) {
         case SIDE_BOTH:
+        {
             // TODO(gw): A temporary hack! While we don't support
             //           border corners that have dots or dashes
             //           with another style, pretend they are solid
@@ -102,6 +103,7 @@ int select_style(int color_select, vec2 fstyle) {
             if (style.x != style.y && (has_dots || has_dashes))
                 return BORDER_STYLE_SOLID;
             return style.x;
+        }
         case SIDE_FIRST:
             return style.x;
         case SIDE_SECOND:


nical:
It would be great to add this to the [list of driver issues](https://github.com/servo/webrender/wiki/Driver-issues), even if this is not strictly coming from a driver. Or remind me to do it when I come back from vacation next week.

webrender/res/ps_border_edge.vs.glsl
```diff
@@ -42,11 +42,15 @@ void write_color(vec4 color, float style, bool flip) {
 
     switch (int(style)) {
         case BORDER_STYLE_GROOVE:
+        {
```
kvark: I don't see declarations in these switch cases. Are they also a problem?

mephisto41: Forget to explain this case. See [1], ANGLE will translate ternary operator to ```type s; if(b) s = x; else s = y;```. And ```type s``` is a declaration.

[1]: https://github.com/google/angle/blob/d7b1ab5893953b97fe5dc1b3087a9b4b50ec45ee/src/compiler/translator/UnfoldShortCircuitToIf.cpp#L139

kvark: omg, that's somewhat obscure
thanks for clarification!

kvark:
I'll add it to the wiki.
Thanks for the fix!
@bors-servo r+

kvark:
see https://github.com/servo/webrender/wiki/Driver-issues#1285---declarations-in-switch-cases

PR #1286, opened 2017-05-23
Reworked the transformed primitives logic

kvark:
The crucial bit of this fix is in `ray_plane`. It used to cull back facing layers, now it does not.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1286)


glennw:
Nice one!

@bors-servo r+

PR #1287, opened 2017-05-23
Support for allocating zero sized textures

kvark:
Addresses https://github.com/servo/webrender/pull/1271#issuecomment-303386393
cc @staktrace 
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1287)


staktrace:
Confirmed this fixes the issue from https://github.com/servo/webrender/pull/1271#issuecomment-303386393 - thanks!

PR #1292, opened 2017-05-24
Full rectangular pixel snapping

### LINKED ISSUE #1279: Pixel snapping with a stretch
We currently pass a single reference point into the primitive's vertex shader, based on which the snapping offset is computed for the whole primitive. That's not exactly how Gecko/Safari do the snapping. We should be snapping both left and right (correspondingly, top and bottom) of a primitive, potentially even stretching it.

kvark:
Fixes #1279 
r? @glennw 

WIP TODOs:
- [x] test servo WPT reftests
- [x] add local snapping tests
- [x] cover `write_transform_vertex`

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1292)


glennw:
@kvark This generally looks good to me. After running the CSS tests, I get these unexpected results:

```
  ▶ FAIL [expected PASS] /css-transforms-1_dev/html/rotate_45deg.htm
  └   → /css-transforms-1_dev/html/rotate_45deg.htm 19763a42afc94b24a1c3762e2eab58e0fc89900b
/css-transforms-1_dev/html/reference/rotate_45deg-ref.htm 44b428d2043c705a9257112a9108ae1cdec9a165
Testing 19763a42afc94b24a1c3762e2eab58e0fc89900b == 44b428d2043c705a9257112a9108ae1cdec9a165

  ▶ FAIL [expected PASS] /css-transforms-1_dev/html/rotate_y_45deg.htm
  └   → /css-transforms-1_dev/html/rotate_y_45deg.htm 3a92183be2a54f07fbbb974600b4330f5a653712
/css-transforms-1_dev/html/reference/rotate_y_45deg-ref.htm 4f53b38136f08aaf1d3bfd47fd0e693a17596dc7
Testing 3a92183be2a54f07fbbb974600b4330f5a653712 == 4f53b38136f08aaf1d3bfd47fd0e693a17596dc7

  ▶ PASS [expected FAIL] /css-transforms-1_dev/html/transform-input-019.htm

  ▶ PASS [expected FAIL] /css21_dev/html4/block-non-replaced-height-001.htm

  ▶ FAIL [expected PASS] /css21_dev/html4/c5525-fltmrgn-000.htm
  └   → /css21_dev/html4/c5525-fltmrgn-000.htm 578df2d1dbc3ffedc88fe346808ec5742b041a7c
/css21_dev/html4/reference/c5525-fltmrgn-000-ref.htm 653c6c85e4f06d453eab1cc2a46f85df8fbfe848
Testing 578df2d1dbc3ffedc88fe346808ec5742b041a7c == 653c6c85e4f06d453eab1cc2a46f85df8fbfe848
```

It's quite possible that those changes are in fact a different issue, exposed by these changes. But let's investigate them first and see if we can work out what's causing the changes.

kvark:
@Gankro thanks for giving it a try! I looked at the failing border test, and it doesn't appear to be trivial at all. Given the number of tests fixed, I think it's a step forward to trade 30 for 1 here. Hopefully, further investigation on border rendering will clear this out.

@glennw I checked rotate_45 tests, and we are rendering them now exactly like Gecko does. Unfortunately, this doesn't match WPT reference, which means Gecko fails it as well. Thus, the answer lies in the investigation of who is right between Gecko and WPT.

glennw:
@kvark Ah, interesting! If Gecko fails those reftests, we have a good case for disabling / marking as fails in the Servo test metadata.

glennw:
@kvark It'd be interesting to know if Blink also fails those tests, if it's easy to check.

kvark:
@glennw for all 3 failed tests (by this PR), I confirm both Gecko (52.0.2, Linus) and Chromium (57.0.2987.133, Linus) fail them as well (even though Chromium produces a little different result from Gecko). I took the upstream versions of the tests (from web-platform-tests repo) just to be sure.

glennw:
:shipit: 

@bors-servo r+

Gankro:
Looks like this slightly changed anti-aliasing behaviour, so you need to regenerate the test I just added (reftests/borders/degenerate-curve). Or make it fuzzy (max difference: 44, number of differing pixels: 287).

kvark:
Given that nothing in the code is changed, and the new reftest author gives a green light on changing the test, I consider this PR to still be approved.

@bors-servo r=glennw

staktrace:
@kvark This PR (or #1296, haven't done a full bisection) appears to be causing a panic in gecko crashtests, when running `dom/animation/test/crashtests/1272475-1.html`. Backtrace can be found [here](https://treeherder.mozilla.org/logviewer.html#?job_id=102467951&repo=try&lineNumber=2240)

It might also be responsible for an increase in fuzziness in the `layout/reftests/columns/column-balancing-overflow-005.html` reftest, although that's less important (but perhaps unexpected?).

kvark:
@staktrace are you sure? this PR doesn't touch any Rust code, it would unlikely trigger a panic in `TransformedRect::new`.

staktrace:
Sorry, my bad. I kicked off a bisection to confirm after posting the comment, and it looks like #1296 is the culprit. The same panic appears on [this try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=b179de62822433e57bbd214c53b4b3bc06aa1a01) which has webrender at cset 150da47b (including #1296 but not this one).

staktrace:
Also for the record the column-balancing-overflow-005.html fuzziness increase *is* from this PR.

kvark:
Thanks @staktrace ! That's the regression mentioned by @Gankro here - https://github.com/servo/webrender/pull/1292#issuecomment-303899148

PR #1293, opened 2017-05-24
don't take rectangular border fast path for curved borders

### LINKED ISSUE #1291: Webrender doesn't curve borders if both neighbours are invisible
Here are two borders that should render the same (to the human eye) but don't in webrender:

screen shot 2017-05-24 at 11 46 02 am

left (incorrect render):
```
          bounds: [28, 18, 144, 122]
          clip: [-17895698, -17895698, 35791396, 35791396]
          "clip-and-scroll": 0
          type: border
          width: [18, 0, 18, 0]
          "border-type": normal
          color: 0 0 255 1.0000
          style:
            - solid
            - none
            - solid
            - none
          radius: [61, 61]
```

right (correct render):
```
          bounds: [192, 18, 144.03334, 122]
          clip: [-17895698, -17895698, 35791396, 35791396]
          "clip-and-scroll": 0
          type: border
          width: [18, 0.016666668, 18, 0.016666668]
          "border-type": normal
          color: 0 0 255 1.0000
          style: solid
          radius: [61, 61]
```

The problem is that webrender seems to give up on applying curvature if both neighbours (in this case left and right) have no rendered size. This can be accomplished by either `border-width: 0` or `style: none` -- in this case both.

This is seen in this gecko reftest: `layout/reftests/border-radius/corner-joins-1.xhtml` -- both servo and gecko feed webrender zero-size borders.

Gankro:
fixes #1291

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1293)


webrender/src/border.rs
```diff
@@ -296,6 +296,11 @@ impl FrameBuilder {
         let top = &border.top;
         let bottom = &border.bottom;
 
+        let has_no_curve = radius.bottom_left.width  == 0.0 && radius.bottom_left.height  == 0.0
```
kvark: could be rewritten as `radius != BorderRadius::zero()`

glennw:
Can we add a reftest for this too?

webrender/res/ps_border_corner.vs.glsl
```diff
@@ -27,15 +27,19 @@ void set_radii(int style,
         case BORDER_STYLE_RIDGE:
         case BORDER_STYLE_GROOVE:
             vRadii1.xy = radii - adjusted_widths;
-            vRadii1.zw = -widths;
+            // See comment in default branch
+            vRadii1.zw = vec2(-100, -100);
```
glennw: These will need to be ```vec2(-100.0)``` - some shader compilers require the explicit float to avoid a numeric cast error.

PR #1294, opened 2017-05-25
Hold old resources in the texture cache for a bit longer.

glennw:
Previously, we were expiring resources from the texture cache
as long as they weren't used that frame. This was resulting in
a lot of extra glyph rasterization than we really want.

Instead, only expire resources that haven't been used for ~10 seconds
when running at 60fps. This drastically reduces the time worker
threads spend rasterizing glyphs, and doesn't have any noticeable
effect on the size of the texture cache from the sites I tested on.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1294)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -13,6 +13,7 @@ use clip_scroll_tree::{ClipScrollTree, ScrollStates};
 use profiler::TextureCacheProfileCounters;
 use resource_cache::ResourceCache;
 use scene::{Scene, SceneProperties};
+use std::cmp;
 use std::collections::HashMap;
 use std::hash::BuildHasherDefault;
 use tiling::{CompositeOps, DisplayListMap, PrimitiveFlags};
@@ -29,6 +30,7 @@ use webrender_traits::{StackingContext, TileOffset, TransformStyle, WorldPoint};
 pub struct FrameId(pub u32);
 
 static DEFAULT_SCROLLBAR_COLOR: ColorF = ColorF { r: 0.3, g: 0.3, b: 0.3, a: 0.6 };
+const CACHE_EXPIRY_TIME: u32 = 600;         // 10 seconds when running at 60fps.
 
 struct FlattenContext<'a> {
     scene: &'a Scene,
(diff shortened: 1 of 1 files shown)


webrender/src/frame.rs
```diff
@@ -976,7 +978,9 @@ impl Frame {
                                      display_lists,
                                      device_pixel_ratio,
                                      texture_cache_profile);
-        resource_cache.expire_old_resources(self.id);
+        // Expire any resources that haven't been used for CACHE_EXPIRY_TIME frames.
+        let expiry_frame = FrameId(cmp::max(CACHE_EXPIRY_TIME, self.id.0) - CACHE_EXPIRY_TIME);
```
kvark: I'd prefer us to not do any expiration before we reach `CACHE_EXPIRY_TIME`, since we may load a lot of stuff at frame 0, and this logic would be happy to get rid of it instantly.
This is still a big improvement though, so we can get it patched later on.

staktrace:
Just a heads up that this might be causing failures in Gecko. I'm still running some try builds to verify, but a lot of the media/gl mochitests are blue in [this try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=c2031a6f7eac2dd0129a34c242fcd60abaa2a4d7) - in the past this has been due to OOMing and/or memory leaks. Given this PR causes more memory to be held on to, it seems like it might be a similar problem.

If it turns out to be this patch I'll try reducing the expiry time from 600 frames to something smaller and see if that helps.

staktrace:
Oh, I wonder if the last 10 seconds worth never get freed?

kvark:
Yeah, resource cache only get's called `expire_old_resources` when a frame is built. We need to clean everything up upon destruction too.

staktrace:
So yes, the badness is definitely coming from this patch:
[try run without this patch](https://treeherder.mozilla.org/#/jobs?repo=try&revision=de948997878f0c2a0dcb984d4ebdba8c8b61a2ab) - green
[try run with this patch](https://treeherder.mozilla.org/#/jobs?repo=try&revision=e23099dc12b4fb1894e904009f2d49f4467659e1) - lots of blue
[with this patch, but 60 frames instead of 600](https://treeherder.mozilla.org/#/jobs?repo=try&revision=b82c3ff2202d5b36aeac901766f36e963ed5c10a) - green

staktrace:
[try run with deinit()'ing the renderer](https://treeherder.mozilla.org/#/jobs?repo=try&revision=692a904d91309f89a9ab12f1035ada788c193c94) - still lots of blue
[try run with deinit() and kvark's tweak to deinit()](https://treeherder.mozilla.org/#/jobs?repo=try&revision=315818a2b2d997518f7caaef883297460b3554bd) - blue, plus some additional reftest crashes

PR #1296, opened 2017-05-26
Ensure the result of calculation is finite number.

mephisto41:
In this [test](https://searchfox.org/mozilla-central/source/dom/animation/test/crashtests/1272475-1.html), we create a very large transform which exceed max of float. This results many NaN or infinite numbers. If we don't handle it, the partial_cmp with Nan or infinite numbers would cause crash.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1296)

### DIFF
--- a/webrender/src/util.rs
+++ b/webrender/src/util.rs
@@ -150,6 +150,15 @@ pub fn subtract_rect<U>(rect: &TypedRect<f32, U>,
         }
     }
 }
+
+pub fn get_normal(x: f32) -> Option<f32> {
+    if x.is_normal() {
+        Some(x)
+    } else {
+        None
+    }
+}
+
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 #[repr(u8)]
 pub enum TransformedRectKind {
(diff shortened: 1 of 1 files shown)


webrender/src/util.rs
```diff
@@ -150,6 +150,15 @@ pub fn subtract_rect<U>(rect: &TypedRect<f32, U>,
         }
     }
 }
+
+pub fn ensure_float_is_normal(x: f32) -> f32 {
```
kvark: This is rather minor, since the function is only used once, but if we want to make it generally useful, I think we should have an API like `get_normal(x: f32) -> Option<f32>`, so that the function clients do `unwrap_or(0.0)` (or other value, depending on the context).

mephisto41: Thanks for the suggestion. I'll update the PR soon.

staktrace:
@mephisto41 This PR appears to cause a panic in gecko crashtests, when running dom/animation/test/crashtests/1272475-1.html. Backtrace can be found at https://treeherder.mozilla.org/logviewer.html#?job_id=102467951&repo=try&lineNumber=2240

kvark:
@staktrace interesting, so the only new call this PR makes is `f32::is_normal`, which is from Rust std library: https://doc.rust-lang.org/stable/std/primitive.f32.html#method.is_normal

It doesn't appear to be unsafe to me, so I don't understand why build bots crash on it. Besides, I can't repro it either. Can you?

staktrace:
Based on the line number in the stack, it seems to be coming from [this line](https://hg.mozilla.org/try/file/b179de62822433e57bbd214c53b4b3bc06aa1a01/gfx/webrender/src/util.rs#l266) so most likely an underflow?

kvark:
Yes, see #1302
AFAIK, the test case passes an absurd transformation, which gets some of the vertices detected as abnormal (and zeroed out), which ends up subtracting -0x80000000 from 0, which is not representable as an `i32`.

PR #1298, opened 2017-05-26
Better graphics debugging experience

kvark:
There were large chunks of query-related, texture parameter-related, and other GL calls that are quite inconvenient to scroll through if looking at an `apitrace` capture. This PR makes it look at bit nicer:

wr-debug

It also adds `no-scissor` option to `wrench` so that the texture cache doesn't show pieces of old data that are distracting. We could also use it for benchmarking, given that scissoring is a controversial optimization on clears.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1298)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -65,10 +65,10 @@ pub const MAX_VERTEX_TEXTURE_WIDTH: usize = 1024;
 const GPU_TAG_CACHE_BOX_SHADOW: GpuProfileTag = GpuProfileTag { label: "C_BoxShadow", color: debug_colors::BLACK };
 const GPU_TAG_CACHE_CLIP: GpuProfileTag = GpuProfileTag { label: "C_Clip", color: debug_colors::PURPLE };
 const GPU_TAG_CACHE_TEXT_RUN: GpuProfileTag = GpuProfileTag { label: "C_TextRun", color: debug_colors::MISTYROSE };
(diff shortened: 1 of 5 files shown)


LifeIsStrange:
[HS] Just wondering, wouldn't you prefer using CodeXL which is a very powerful gpu and cpu debugger with fine-grained profiling, many apis support (vulkan), evolutivity, power consumption profiling and the no longer need to spend time to develop your own debugger ? 
It's also freesoftware so you can adapt it to your specific needs. 
http://gpuopen.com/compute-product/codexl/
Sorry for my bad english.

kvark:
@LifeIsStrange Thank you for the suggestion!

> no longer need to spend time to develop your own debugger

We don't have our own debugger. We have some debug draw functions, but that's mostly it. Part of what my PR does is - increasing the quality of GPU markers, which are standard and getting understood by all GPU profilers out there, not just `apitrace`.

LifeIsStrange:
You're welcome :) and thanks for the clarification.

glennw:
@kvark Hmmm this is giving me some weird results when profiling today. I think the issue is probably here - https://github.com/servo/webrender/pull/1298/files#diff-f5062b694b9fe53fc1757ed483d577d9L1281.

If you compare the GPU profile graph before / after this PR, they look very different - I don't think the new version is getting correct results (and may be blocking the CPU inadvertently too?).

PR #1299, opened 2017-05-29
Introduce GPU resource cache.

glennw:
This adds an incremental LRU cache that can be used to store values
that vertex shaders need to access.

The intent is to replace the use of the GPU store type, which
currently uploads the entire buffer each frame, and has a lot
of limitations (such as requiring allocation up front).

The new interface supports building GPU data on demand, as required,
which will improve CPU a lot for large pages.

For now, I have ported borders and rectangles to use the new type.
The patch is already large enough - this will allow us to port
each primitive type to use the GPU cache incrementally. The current
code also doesn't actually take advantage of building the data on
demand yet - they are just copied when needed.

Follow up work includes:
 * Porting more types to use the GPU cache.
 * Removing the old gpu store interfaces.
 * Profiling and optimizing various GL incremental update APIs.
 * Supporting resizing of the GPU cache backing store.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1299)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -18,6 +18,7 @@ use euclid::Matrix4D;
 use fnv::FnvHasher;
 use frame_builder::FrameBuilderConfig;
 use gleam::gl;
+use gpu_cache::{GpuCacheUpdate, GpuCacheUpdateList};
 use gpu_store::{GpuStore, GpuStoreLayout};
 use internal_types::{CacheTextureId, RendererFrame, ResultMsg, TextureUpdateOp};
 use internal_types::{TextureUpdateList, PackedVertex, RenderTargetMode};
@@ -38,6 +39,7 @@ use std::marker::PhantomData;
 use std::mem;
 use std::path::PathBuf;
 use std::rc::Rc;
+use std::slice;
 use std::sync::{Arc, Mutex};
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::thread;
@@ -189,6 +191,70 @@ pub enum BlendMode {
 
     // Use the color of the text itself as a constant color blend factor.
     Subpixel(ColorF),
+}
+
+/// The device-specific representation of the cache texture in gpu_cache.rs
+struct CacheTexture {
+    id: TextureId,
+}
+
+impl CacheTexture {
+    fn new(device: &mut Device) -> CacheTexture {
+        let id = device.create_texture_ids(1, TextureTarget::Default)[0];
+
+        CacheTexture {
+            id: id,
+        }
+    }
+
+    fn update(&mut self, device: &mut Device, updates: &GpuCacheUpdateList) {
+        // See if we need to create or resize the texture.
+        let current_dimensions = device.get_texture_dimensions(self.id);
+
+        if updates.height > current_dimensions.height {
+            // TODO(gw): Handle resizing an existing cache texture.
+            if current_dimensions.height > 0 {
+                panic!("TODO: Implement texture copy!!!");
+            }
+
+            // Create a f32 texture that can be used for the vertex shader
+            // to fetch data from.
+            device.init_texture(self.id,
+                                MAX_VERTEX_TEXTURE_WIDTH as u32,
+                                updates.height as u32,
+                                ImageFormat::RGBAF32,
+                                TextureFilter::Nearest,
+                                RenderTargetMode::None,
+                                None);
+        }
+
+        for update in &updates.updates {
+            match update {
+                &GpuCacheUpdate::Copy { block_index, block_count, address } => {
+                    // Apply an incremental update to the cache texture.
+                    // TODO(gw): For the initial implementation, we will just
+                    //           use update_texture() since it's simple. If / when
+                    //           we profile this and find it to be slow on some / all
+                    //           devices - we can look into other options, such as
+                    //           using glMapBuffer() with the unsynchronized bit,
+                    //           and managing the synchronization ourselves with fences.
+                    let data: &[u8] = unsafe {
+                        let ptr = updates.blocks
+                                         .as_ptr()
+                                         .offset(block_index as isize);
+                        slice::from_raw_parts(ptr as *const _, block_count * 16)
+                    };
+                    device.update_texture(self.id,
+                                          address.u as u32,
+                                          address.v as u32,
+                                          block_count as u32,
+                                          1,
+                                          None,
+                                          data);
+                }
+            }
+        }
+    }
 }
 
 struct GpuDataTexture<L> {
@@ -468,7 +534,6 @@ struct GpuDataTextures {
     data16_texture: VertexDataTexture,
     data32_texture: VertexDataTexture,
     data64_texture: VertexDataTexture,
-    data128_texture: VertexDataTexture,
     resource_rects_texture: VertexDataTexture,
     gradient_data_texture: GradientDataTexture,
     split_geometry_texture: SplitGeometryTexture,
@@ -483,7 +548,6 @@ impl GpuDataTextures {
             data16_texture: VertexDataTexture::new(device),
             data32_texture: VertexDataTexture::new(device),
             data64_texture: VertexDataTexture::new(device),
-            data128_texture: VertexDataTexture::new(device),
             resource_rects_texture: VertexDataTexture::new(device),
             gradient_data_texture: GradientDataTexture::new(device),
             split_geometry_texture: SplitGeometryTexture::new(device),
@@ -494,7 +558,6 @@ impl GpuDataTextures {
         self.data16_texture.init(device, &mut frame.gpu_data16);
         self.data32_texture.init(device, &mut frame.gpu_data32);
         self.data64_texture.init(device, &mut frame.gpu_data64);
-        self.data128_texture.init(device, &mut frame.gpu_data128);
         self.prim_geom_texture.init(device, &mut frame.gpu_geometry);
         self.resource_rects_texture.init(device, &mut frame.gpu_resource_rects);
         self.layer_texture.init(device, &mut frame.layer_texture_data);
@@ -508,7 +571,6 @@ impl GpuDataTextures {
         device.bind_texture(TextureSampler::Data16, self.data16_texture.id);
         device.bind_texture(TextureSampler::Data32, self.data32_texture.id);
         device.bind_texture(TextureSampler::Data64, self.data64_texture.id);
-        device.bind_texture(TextureSampler::Data128, self.data128_texture.id);
         device.bind_texture(TextureSampler::ResourceRects, self.resource_rects_texture.id);
         device.bind_texture(TextureSampler::Gradients, self.gradient_data_texture.id);
         device.bind_texture(TextureSampler::SplitGeometry, self.split_geometry_texture.id);
@@ -527,6 +589,7 @@ pub struct Renderer {
     result_rx: Receiver<ResultMsg>,
     device: Device,
     pending_texture_updates: Vec<TextureUpdateList>,
+    pending_gpu_cache_updates: Vec<GpuCacheUpdateList>,
     pending_shader_updates: Vec<PathBuf>,
     current_frame: Option<RendererFrame>,
 
@@ -594,6 +657,8 @@ pub struct Renderer {
 
     gdt_index: usize,
     gpu_data_textures: [GpuDataTextures; GPU_DATA_TEXTURE_POOL],
+
+    gpu_cache_texture: CacheTexture,
 
     pipeline_epoch_map: HashMap<PipelineId, Epoch, BuildHasherDefault<FnvHasher>>,
     /// Used to dispatch functions to the main thread's event loop.
@@ -1092,6 +1157,8 @@ impl Renderer {
             backend.run(backend_profile_counters);
         })};
 
+        let gpu_cache_texture = CacheTexture::new(&mut device);
+
         let gpu_profile = GpuProfiler::new(device.rc_gl());
 
         let renderer = Renderer {
@@ -1099,6 +1166,7 @@ impl Renderer {
             device: device,
             current_frame: None,
             pending_texture_updates: Vec::new(),
+            pending_gpu_cache_updates: Vec::new(),
             pending_shader_updates: Vec::new(),
             cs_box_shadow: cs_box_shadow,
             cs_text_run: cs_text_run,
@@ -1154,6 +1222,7 @@ impl Renderer {
             vr_compositor_handler: vr_compositor,
             cpu_profiles: VecDeque::new(),
             gpu_profiles: VecDeque::new(),
+            gpu_cache_texture: gpu_cache_texture,
         };
 
         let sender = RenderApiSender::new(api_tx, payload_tx);
@@ -1218,8 +1287,15 @@ impl Renderer {
         // Pull any pending results and return the most recent.
         while let Ok(msg) = self.result_rx.try_recv() {
             match msg {
-                ResultMsg::NewFrame(frame, texture_update_list, profile_counters) => {
+                ResultMsg::NewFrame(mut frame, texture_update_list, profile_counters) => {
                     self.pending_texture_updates.push(texture_update_list);
+                    if let Some(ref mut frame) = frame.frame {
+                        // TODO(gw): This whole message / Frame / RendererFrame stuff
+                        //           is really messy and needs to be refactored!!
+                        if let Some(update_list) = frame.gpu_cache_updates.take() {
+                            self.pending_gpu_cache_updates.push(update_list);
+                        }
+                    }
                     self.backend_profile_counters = profile_counters;
 
                     // Update the list of available epochs for use during reftests.
@@ -1309,6 +1385,9 @@ impl Renderer {
                         //self.update_shaders();
                         self.update_texture_cache();
 
+                        self.update_gpu_cache();
+                        self.device.bind_texture(TextureSampler::ResourceCache, self.gpu_cache_texture.id);
+
                         frame_id
                     };
 
@@ -1380,6 +1459,13 @@ impl Renderer {
         }
     }
 */
+
+    fn update_gpu_cache(&mut self) {
+        let _gm = GpuMarker::new(self.device.rc_gl(), "gpu cache update");
+        for update_list in self.pending_gpu_cache_updates.drain(..) {
+            self.gpu_cache_texture.update(&mut self.device, &update_list);
+        }
+    }
 
     fn update_texture_cache(&mut self) {
         let _gm = GpuMarker::new(self.device.rc_gl(), "texture cache update");
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -122,8 +122,8 @@ uniform sampler2D sPrimGeometry;
 uniform sampler2D sData16;
 uniform sampler2D sData32;
 uniform sampler2D sData64;
-uniform sampler2D sData128;
 uniform sampler2D sResourceRects;
+uniform sampler2D sResourceCache;
 
 // Instanced attributes
 in ivec4 aData0;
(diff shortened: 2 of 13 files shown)


glennw:
r? @kvark 

It turned out to be a large patch, sorry! Hopefully follow ups should be a lot smaller and more isolated.

I'd suggest starting by reading the comment here https://github.com/servo/webrender/pull/1299/files#diff-950eebe8d7a83d7c04825debe3b9b68fR14 in ```gpu_cache.rs``` which tries to give an overview of how it works.

glennw:
Oh, and a missing point from the follow up work is to actually make use of the on demand functionality, to avoid building items in the frame builder during flattening.

glennw:
I've successfully run the complete WPT/CSS test suite on this, and done a fair amount of manual testing on common sites.

kvark:
I'm trying to rationalize the need for that another layer of indirection that is introduced here. Given that a`GpuStoreAddress` is associated 1 to 1 with a reserved slot, couldn't we just store `(GpuStoreAddress, Generation)` directly? If `GpuStore` decides that a particular address wasn't used for too long, it would bump the internal `Generation` and re-use it for something else, which will allow you to recognize the need to re-allocate the address for the original primitive once it needs it again.

---

Review status: all files reviewed at latest revision, 19 unresolved discussions.

---

*[webrender/src/border.rs, line 262 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKCOFTIc0Ue5QVhqHG:-KlKCOFU6mwYpUza0DWF:b-36rtgc) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/border.rs#L262)):*
> ```Rust
>                 right_color.into(),
>                 bottom_color.into(),
>                 [ radius.top_left.width,
> ```

a helper function to convert from `TypedRect` to `GpuBlockData` would be useful, not just to shorten the code here but also to enforce the standard

---

*[webrender/src/device.rs, line 27 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKNoLO3movbukHB-8e:-KlKNoLO3movbukHB-8f:b-83apg9) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/device.rs#L27)):*
> ```Rust
> 
> #[derive(Debug, Copy, Clone, PartialEq, Ord, Eq, PartialOrd)]
> pub struct FrameId(pub usize);
> ```

out of curiosity, why does it need to be public now?

---

*[webrender/src/gpu_cache.rs, line 15 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFX0RDJAhanJ0nccu:-KlKFX0RDJAhanJ0nccv:b5m4jgr) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L15)):*
> ```Rust
> 
> /*
>     Overview of the GPU cache.
> ```

perhaps, we can move this block up to make it a module documentation item

---

*[webrender/src/gpu_cache.rs, line 46 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFbHDNN1_O0PQg6ha:-KlKFbHDNN1_O0PQg6hb:bm5k3nf) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L46)):*
> ```Rust
> const FRAMES_BEFORE_EVICTION: usize = 10;
> 
> // A single texel in RGBAF32 texture - 16 bytes.
> ```

let's use `///` doc comments of this public item and below

---

*[webrender/src/gpu_cache.rs, line 71 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKEgsDxmdc4dqGbcmN:-KlKEgsETJfAQhzVaI7u:bqft0pv) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L71)):*
> ```Rust
> // Any data type that can be stored in the GPU cache should
> // implement this trait.
> pub trait IntoGpuBlocks {
> ```

nit: naming convention of `Into` implies conversion of `self`, while the method takes `&self` by reference. Perhaps, it should be called `ToGpuBlocks` instead

---

*[webrender/src/gpu_cache.rs, line 131 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFO2D4XJ-8awZ50fH:-KlKFO2D4XJ-8awZ50fI:bptlwcg) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L131)):*
> ```Rust
>     // exist for this row. Allows finding the block structure quickly
>     // when free'ing a block, from its address only.
>     first_block_index: usize,
> ```

shouldn't it have `BlockIndex` type?

---

*[webrender/src/gpu_cache.rs, line 188 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFtkfkvCqMAlRAVvV:-KlKFtkgeS1p9xStHpKT:b-yf7vos) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L188)):*
> ```Rust
> 
>     fn clear(&mut self) {
>         self.free_list_1 = None;
> ```

nit: could be `*self = Self::new()`

---

*[webrender/src/gpu_cache.rs, line 217 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKGXDrjQ3LXHZYPCvi:-KlKGXDsF6J1o0t4CjyW:beong3u) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L217)):*
> ```Rust
>     blocks: Vec<Block>,
>     // Metadata about each allocated row.
>     rows: Vec<Row>,
> ```

can we just compute `height` as `rows.length()` instead of storing it?

---

*[webrender/src/gpu_cache.rs, line 345 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKNfBZMoDJz93GMuKz:-KlKNfBZMoDJz93GMuL-:bdeow3n) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L345)):*
> ```Rust
>     /// Slot is currently allocated with backing
>     /// store in the cache.
>     Occupied {
> ```

nice use of structured enum!

---

*[webrender/src/gpu_cache.rs, line 359 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKNzfOqX_iBYbFcJgP:-KlKNzfOqX_iBYbFcJgQ:b-pzgs2m) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L359)):*
> ```Rust
> struct CacheSlot<K> {
>     /// Intrusive link to the next cache slot
>     /// in the cache slot free-list (pending or
> ```

if a slot is in a free-list, why would it be `Pending`/`Occupied`?

---

*[webrender/src/gpu_cache.rs, line 373 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKOpVhkA_cIwoh8FH-:-KlKOpVisiACvX5H1o0a:b35bif6) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L373)):*
> ```Rust
>     /// Current frame ID.
>     frame_id: FrameId,
>     /// Free-list of cache slots. Each cach slot exists
> ```

"cach" typo

---

*[webrender/src/gpu_cache.rs, line 380 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKP3EuGooXZSLL-KyV:-KlKP3EuGooXZSLL-KyW:b-nar2kh) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L380)):*
> ```Rust
>     /// fast to iterate all the pending slots (for GPU upload), and
>     /// all the occupied slots (to check for cache eviction).
>     pending_list_head: Option<GpuCacheSlotId>,
> ```

ooooh, so many linked lists!

---

*[webrender/src/gpu_cache.rs, line 393 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKPloLh8BEhqOlz49b:-KlKPloMB2cCHPQlZ6Hr:b-ll92hd) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L393)):*
> ```Rust
>             frame_id: FrameId(0),
>             slots: Vec::new(),
>         	keys: HashMap::new(),
> ```

tab detected!

---

*[webrender/src/gpu_cache.rs, line 499 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKazm0_KbUit55RJMy:-KlKazm0_KbUit55RJMz:bkwyqwx) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L499)):*
> ```Rust
>                         // as empty.
>                         if self.frame_id.0 > FRAMES_BEFORE_EVICTION &&
>                            last_access_time.0 < self.frame_id.0 - FRAMES_BEFORE_EVICTION {
> ```

nit: could combine into a single condition `last_access_time.0 + FRAMES_BEFORE_EVICTION < self.frame_id.0`

---

*[webrender/src/gpu_cache.rs, line 530 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKbjP9LrI_JP4fCl4r:-KlKbjP9LrI_JP4fCl4s:bpznooi) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L530)):*
> ```Rust
>         }
> 
>         // Run through the pending list and build new items.
> ```

any reason for not combining running and pruning passes together?

---

*[webrender/src/gpu_cache.rs, line 544 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKc0NrnOZhJ5zAiMJ8:-KlKc0NrnOZhJ5zAiMJ9:b-11l68u) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L544)):*
> ```Rust
>                     // Build it!
>                     let start_index = self.texture.pending_blocks.len();
>                     f(slot.key, &mut self.texture.pending_blocks);
> ```

it's somewhat dirty to allow `f` to clear or re-fill the vector completely, but I suppose we can live with that as an implementation detail

---

*[webrender/src/gpu_cache.rs, line 574 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKcJGqVWKTgWWt-KCt:-KlKcJGrJd0r6UwKs-6g:b-pq9brc) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L574)):*
> ```Rust
>     pub fn get_address(&self, id: &GpuCacheSlotId) -> GpuCacheAddress {
>         match self.slots[id.0 as usize].details {
>             SlotDetails::Empty => panic!("Trying to access a vacant gpu cache entry!"),
> ```

nit: could include the `id` in the message

---

*[webrender/src/gpu_cache.rs, line 580 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKcM7v8CI23REExF4K:-KlKcM7v8CI23REExF4L:bb1lwhh) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L580)):*
> ```Rust
>                 // just accidentally left over in the cache from previous
>                 // frames.
>                 debug_assert!(last_access_time == self.frame_id);
> ```

nit: could use `debug_assert_eq`

---

*[webrender/src/prim_store.rs, line 1260 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKgPrhYLsJafNSTFEa:-KlKgPrhYLsJafNSTFEb:b7hhrw5) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/prim_store.rs#L1260)):*
> ```Rust
> 
>         match metadata.gpu_location {
>             GpuLocation::GpuCache(ref mut cache_id) => {
> ```

nit: looks like `if let` would be more appropriate here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1299#-:-KlKkzT_SE_4XpnMtdvZ:bzd4m3b)*

glennw:
Review status: 9 of 13 files reviewed at latest revision, 19 unresolved discussions.

---

*[webrender/src/gpu_cache.rs, line 15 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFX0RDJAhanJ0nccu:-KlLZIdiHVauJotssazO:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L15)):*

perhaps, we can move this block up to make it a module documentation item
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 46 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFbHDNN1_O0PQg6ha:-KlLZGZoSZpwRcoGcWZS:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L46)):*

let's use `///` doc comments of this public item and below
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 71 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKEgsDxmdc4dqGbcmN:-KlLZKBS9v_bDoB04w3u:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L71)):*

nit: naming convention of `Into` implies conversion of `self`, while the method takes `&self` by reference. Perhaps, it should be called `ToGpuBlocks` instead
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 131 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFO2D4XJ-8awZ50fH:-KlLZKrhV9xrchyhJYev:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L131)):*

shouldn't it have `BlockIndex` type?
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 188 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKFtkfkvCqMAlRAVvV:-KlLZLVIG72znXvHcNHw:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L188)):*

nit: could be `*self = Self::new()`
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 217 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKGXDrjQ3LXHZYPCvi:-KlLZM0mtPnH9HZhyFbR:b-9yekl9) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L217)):*

can we just compute `height` as `rows.length()` instead of storing it?
</blockquote></details>

The idea was that we'd increment height at a greater rate (perhaps powers of 2) to avoid re-allocating the texture too much. So in general the texture height is not equal to the rows.length().

---

*[webrender/src/gpu_cache.rs, line 359 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKNzfOqX_iBYbFcJgP:-KlLZadQ9JqvNPBNDG79:b-7xu4ds) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L359)):*

if a slot is in a free-list, why would it be `Pending`/`Occupied`?
</blockquote></details>

Fixed the comment - it's either in pending/occupied or a free-blocks list.

---

*[webrender/src/gpu_cache.rs, line 373 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKOpVhkA_cIwoh8FH-:-KlLZfZEF1D6JcFeFVJS:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L373)):*

"cach" typo
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 393 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKPloLh8BEhqOlz49b:-KlLZguPl32XOfPDY6jR:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L393)):*

tab detected!
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 499 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKazm0_KbUit55RJMy:-KlLZhg8O_X4E4TDc6oC:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L499)):*

nit: could combine into a single condition `last_access_time.0 + FRAMES_BEFORE_EVICTION < self.frame_id.0`
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 530 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKbjP9LrI_JP4fCl4r:-KlLZiLWcJ3F3pU7L4za:b-sorrs6) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L530)):*

any reason for not combining running and pruning passes together?
</blockquote></details>

I figured we would get better re-use of existing space if we recover all evicted blocks first, and then allocate after that.

---

*[webrender/src/gpu_cache.rs, line 544 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKc0NrnOZhJ5zAiMJ8:-KlLZvCFqP4UAKVn4Xf9:br6ymrf) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L544)):*

it's somewhat dirty to allow `f` to clear or re-fill the vector completely, but I suppose we can live with that as an implementation detail
</blockquote></details>

I agree - I don't particularly like this, but I think it's probably worth it for the efficiency. If we can come up with a cleaner way to express this that would be ideal though.

---

*[webrender/src/gpu_cache.rs, line 574 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKcJGqVWKTgWWt-KCt:-KlL_7QLyKpEgMHsUvqC:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L574)):*

nit: could include the `id` in the message
</blockquote></details>

Done.

---

*[webrender/src/gpu_cache.rs, line 580 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKcM7v8CI23REExF4K:-KlL_7pkN5Z1NMIyELb1:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L580)):*

nit: could use `debug_assert_eq`
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1260 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKgPrhYLsJafNSTFEa:-KlL_8XbD2akKJu-CoVx:b-896fix) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/prim_store.rs#L1260)):*

nit: looks like `if let` would be more appropriate here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1299)*

glennw:
@kvark Thanks for the review! I've fixed up / answered most of those questions. I pushed the changes as a separate commit for ease of review, but can squash before merge.

I haven't changed the extra indirection layer we were talking about. I *think* keeping it as is might be better in terms of possible multi-threading of the build step, and removing parts of the frame builder and building directly from DL items. We can perhaps discuss again on IRC the pros and cons? Alternatively, we could perhaps do that as a follow up - I don't think it would be a huge change, if I'm understanding your suggestion correctly.

glennw:
Hmm, it might make sense to remove that indirection layer - going to try experiment with that today.

kvark:
Reviewed 1 of 1 files at r2, 3 of 3 files at r3.

---

*[webrender/src/gpu_cache.rs, line 530 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKbjP9LrI_JP4fCl4r:-KlM2fptQZCbEbQj29Dx:bmo09vg) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L530)):*

I figured we would get better re-use of existing space if we recover all evicted blocks first, and then allocate after that.
</blockquote></details>

Ok, as long as it doesn't take too much overhead.

---

*[webrender/src/gpu_cache.rs, line 544 at r1](https://reviewable.io:443/reviews/servo/webrender/1299#-KlKc0NrnOZhJ5zAiMJ8:-KlM33ADt2uSemATmvzs:b-6xdpou) ([raw file](https://github.com/servo/webrender/blob/194ef9e63e1a919f43848598c5f7b3c3968ecb8b/webrender/src/gpu_cache.rs#L544)):*

I agree - I don't particularly like this, but I think it's probably worth it for the efficiency. If we can come up with a cleaner way to express this that would be ideal though.
</blockquote></details>

Yep, we'll fix it later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1299)*

glennw:
@kvark Rebased and pushed a new commit which simplifies the cache a lot. It's ready for review again now.

webrender/src/renderer.rs
```diff
@@ -192,6 +194,70 @@ pub enum BlendMode {
...
+        }
+    }
+
+    fn update(&mut self, device: &mut Device, updates: &GpuCacheUpdateList) {
+        // See if we need to create or resize the texture.
+        let current_dimensions = device.get_texture_dimensions(self.id);
+
+        if updates.height > current_dimensions.height {
+            // TODO(gw): Handle resizing an existing cache texture.
+            if current_dimensions.height > 0 {
```
kvark: so basically this will crash once a higher update height is requested than the first frame?

webrender/src/renderer.rs
```diff
@@ -1382,6 +1461,14 @@ impl Renderer {
     }
 */
 
+    fn update_gpu_cache(&mut self) {
+        let _gm = GpuMarker::new(self.device.rc_gl(), "gpu cache update");
+        let pending_gpu_cache_updates = mem::replace(&mut self.pending_gpu_cache_updates, vec![]);
```
kvark: can we just do this instead:
```rust
for update_list in self.pending_gpu_cache_updates.drain(..) {..}
```

glennw:
@kvark Pushed a commit to address most of those comments.

I wasn't quite sure what you meant by the ```evict_frame_id``` comment - could you clarify?

Regarding the resizing panic - that's right, but should be safe to do in a follow up. We currently use a 512 height texture as initial height, and I've never seen any page go above 2 rows so far in my testing (since we're only pushing borders and rectangles through this cache). I'd plan to do the resizing as a follow up shortly.

This will still need to be squashed once ready.

kvark:
👍 👍 👍 
@bors-servo r+

PR #1303, opened 2017-05-29
Bump euclid and heapsize and bump version

Eijebong:
---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1303)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "webrender"
-version = "0.39.0"
+version = "0.40.0"
 authors = ["Glenn Watson <gw@intuitionlibrary.com>"]
 license = "MPL-2.0"
 repository = "https://github.com/servo/webrender"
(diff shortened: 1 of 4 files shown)


webrender/Cargo.toml
```diff
@@ -17,7 +17,7 @@ app_units = "0.4"
 bincode = "1.0.0-alpha6"
 bit-set = "0.4"
 byteorder = "1.0"
-euclid = "0.11.2"
+euclid = ">= 0.11, < 0.14"
```
kvark: I'm not 100% sure about this being a non-breaking change. Is this approach established elsewhere for a reference?

Eijebong: Well I didn't change any code so unless you're reexporting euclid somewhere I don't think that this is a breaking change.

Eijebong: If you want I can force euclid 0.13 and bump to 0.40.0

kvark:
@glennw heads up - this may prevent updating WR for servo until everything is using euclid 0.13
@bors-servo r+

PR #1305, opened 2017-05-29
[DO NOT MERGE] [WIP] Removing combined local clip rect.

### LINKED ISSUE #1309: Remove/fix layer's combined clip rectangle
It's currently clipping the transformed layers incorrectly, hence the temporary hack in #1177.

glennw:
Fixes  #1309

This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1305)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -596,8 +596,7 @@ VertexInfo write_vertex(RectWithSize instance_rect,
     vec2 local_pos = instance_rect.p0 + instance_rect.size * aPosition.xy;
 
     // Clamp to the two local clip rects.
-    vec2 clamped_local_pos = clamp_rect(clamp_rect(local_pos, local_clip_rect),
-                                        layer.local_clip_rect);
+    vec2 clamped_local_pos = clamp_rect(local_pos, local_clip_rect);
 
     /// Compute the snapping offset.
     vec2 snap_offset = compute_snap_offset(clamped_local_pos, local_clip_rect, layer, snap_rect);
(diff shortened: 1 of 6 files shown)


glennw:
Writing up my notes here as @kvark says he may have time to continue this work.

My plan was to implement this in two stages:

1) Remove the combined local clip rect from the vertex shader completely, and get it working (slowly). To do this, all we need to do is make the clip mask code also handles the case of clip-scroll regions where the rectangle is aligned. In this way, a clip mask will be generated and the clipping done per-fragment.

2) Make the clip code avoid generating clip masks for most cases, by re-instating a true local clip rect in the vertex shader, where possible.

I've implemented (1) - I haven't checked since rebasing, but last time I worked on this code it was passing all the WPT/CSS test suite, but about 10-20% slower than before.

I haven't started on (2) - but my basic idea was:
 * In the vast majority of cases, the local clip rect for a primitive is entirely inside the parent clip rect. We can detect this (with SAT for transformed rects, and simpler checks for axis-aligned). In these cases, we can skip the clip mask generation altogether.

This is a very hand-wavy - but I was thinking of accumulating something like the valid local space region of the primitive (by intersecting the primitive rect with any parent clip rects that exist in the clip-scroll tree until we hit a reference frame with different coordinate system). And then, we would have a world space clip region (not necessarily rectangular). If the world space polygon of the local valid region is entirely inside the world space clip region, no mask generation is required. I think this will probably hit the vast majority of cases, avoiding clip mask generation.

For handling the other clip cases, such as masks, clip-out etc, we would then just check if they intersect with the valid world polygon of the local valid region, to see if we need to draw them into a clip mask.

glennw:
Closing for now since this was here just as a reference / starting point for some future work.

PR #1306, opened 2017-05-30
Make it possible to position the viewport anywhere

paulrouget:
Requires: https://github.com/servo/servo/pull/17091

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1306)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -21,7 +21,7 @@ use std::rc::Rc;
 //use std::sync::mpsc::{channel, Sender};
 //use std::thread;
 use webrender_traits::{ColorF, ImageFormat};
-use webrender_traits::{DeviceIntPoint, DeviceIntRect, DeviceIntSize, DeviceUintSize};
+use webrender_traits::{DeviceIntPoint, DeviceIntRect, DeviceIntSize, DeviceUintPoint, DeviceUintSize};
 
 #[derive(Debug, Copy, Clone)]
 pub struct FrameId(usize);
@@ -853,10 +853,13 @@ pub struct Device {
     // Frame counter. This is used to map between CPU
     // frames and GPU frames.
     frame_id: FrameId,
+
+    origin: DeviceUintPoint,
 }
 
 impl Device {
     pub fn new(gl: Rc<gl::Gl>,
+               origin: DeviceUintPoint,
                resource_override_path: Option<PathBuf>,
                _file_changed_handler: Box<FileWatcherHandler>) -> Device {
         //let file_watcher = FileWatcherThread::new(file_changed_handler);
@@ -899,6 +902,8 @@ impl Device {
 
             max_texture_size: max_texture_size,
             frame_id: FrameId(0),
+
+            origin: origin,
         }
     }
 
@@ -1031,7 +1036,7 @@ impl Device {
         }
 
         if let Some(dimensions) = dimensions {
-            self.gl.viewport(0, 0, dimensions.width as gl::GLint, dimensions.height as gl::GLint);
+            self.gl.viewport(self.origin.x as i32, self.origin.y as i32, dimensions.width as gl::GLint, dimensions.height as gl::GLint);
         }
     }
 
--- a/webrender/src/record.rs
+++ b/webrender/src/record.rs
@@ -76,7 +76,7 @@ pub fn should_record_msg(msg: &ApiMsg) -> bool {
         ApiMsg::Scroll(..) |
         ApiMsg::TickScrollingBounce |
         ApiMsg::WebGLCommand(..) |
-        ApiMsg::SetWindowParameters(..) =>
+        ApiMsg::SetViewGeometry(..) =>
             true,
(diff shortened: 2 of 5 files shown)


paulrouget:
I just change the viewport coordinate in `device.rs`. Is that enough?

paulrouget:
> SetViewGeometry doesn't exactly have any relation to geometry, does it?

Maybe geometry is not the right word (in french it is also used to describe a "layout" of a room).

> What makes the old name SetWindowParameters worse?

The problem is "window". From the embedder perspective, a window could be native window, or the page window.

> RenderBackend has rendering_rect and viewport_rect, but the actual glViewport call is done with rendering_rect.origin...

Yeah it's because the web viewport of the gl viewport are not the same thing :(

Definitively confusing.

Maybe:

``` rust
struct ViewParameter {
  inner_rect: … // CSS viewport. Top-left corner of this rect is 0x0 CSS pixel
  outer_rect: … // Where content is actually clipped

  //  Will eventually migrate to the a future CompositorParameter struct:
  chrome_rect: … // Chrome window dimensions. Used from window.moveTo/innerHeight/outerWidth etc.
  hidpi_factor: …
}
```

Any other ideas are welcome :)

kvark:
Trying to evolve your proposal further:
```rust
struct Viewport {
  content_rect: … // CSS viewport. Top-left corner of this rect is 0x0 CSS pixel
  outer_rect: … // Where content is actually clipped

  chrome_rect: … 
  hidpi_factor: …
}
```

paulrouget:
I don't want to bikeshed too much, but I'd rather avoid "Viewport". Like Window, it has multiple meanings (gl viewport and css viewport being 2 different things). From the embedder point of view, we usually use the term "View".

This struct will be used to initialize a `View`, but also update the View layout. So I don't want to just use use `struct View` but `struct ViewParameters` if that makes sense?

In the future, we will also add overscroll parameters (see #1314).

```
#[derive(Copy, Clone)]
pub struct ViewParameters {
    /// CSS viewport. The inner rect top left corner is the 0x0 CSS pixel (before scroll).
    pub inner_rect: TypedRect<u32, DevicePixel>,
    /// The outer rect is where content gets clipped. The outer rect
    /// is equal to or larger than the viewport area. This makes it possible to
    /// draw content outside of the CSS viewport.
    pub outer_rect: TypedRect<u32, DevicePixel>,
    /// The chrome rect is the outer coordinates of the chrome window. It is
    /// equal to or larger than the rendering area. Usually only used for the
    /// window.moveTo/resizeTo/outerWidth/outerHeight… DOM methods.
    pub chrome_rect: (Size2D<u32>, Point2D<i32>),
    /// The device pixel ratio for view.
    pub hidpi_factor: ScaleFactor<f32, DeviceIndependentPixel, DevicePixel>,
}
```

kvark:
@paulrouget ok, let's proceed with your latest scheme.

glennw:
Reviewed 5 of 5 files at r1.

---

*[webrender/src/device.rs, line 857 at r1](https://reviewable.io:443/reviews/servo/webrender/1306#-KlX6crajA9jppHOvrCb:-KlX6crajA9jppHOvrCc:b-38kazi) ([raw file](https://github.com/servo/webrender/blob/2f0194196073b0512945caf8a99f04a0362eb270/webrender/src/device.rs#L857)):*
> ```Rust
>     frame_id: FrameId,
> 
>     origin: DeviceUintPoint,
> ```

The Device code is really meant to abstract an API device, and not maintain state, where possible. We should remove this from here (explained further below).

---

*[webrender/src/device.rs, line 1039 at r1](https://reviewable.io:443/reviews/servo/webrender/1306#-KlX6olrsk7bOk6W2rzk:-KlX6olrsk7bOk6W2rzl:by8bfg5) ([raw file](https://github.com/servo/webrender/blob/2f0194196073b0512945caf8a99f04a0362eb270/webrender/src/device.rs#L1039)):*
> ```Rust
> 
>         if let Some(dimensions) = dimensions {
>             self.gl.viewport(self.origin.x as i32, self.origin.y as i32, dimensions.width as gl::GLint, dimensions.height as gl::GLint);
> ```

The bind_draw_target call is used for both the main framebuffer, and also intermediate render targets. The code as it exists will break things when intermediate targets are involved. Instead, let's change bind_draw_target() to take a rect, and pass that through from the renderer where appropriate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1306)*

glennw:
@paulrouget Added some comments above. CI is also not happy - probably samples and/or wrench need to be updated. We'll also want to bump the version in Cargo.toml (of both WR and WR_traits) since this introduces an API change.

glennw:
@paulrouget Should we keep this open for now?

paulrouget:
@glennw yes please. I'll get back to this once I'm done with some other work.

glennw:
Closing for now since there are so many conflicts. Feel free to re-open when rebased.

PR #1310, opened 2017-05-30
Add filter setting for yuv image.

JerryShih:
r? @kvark @glennw @nical 

Some reftests in gecko need the nearest sampling. Bypass the filter setting for yuv image.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1310)

### DIFF
--- a/webrender/examples/yuv.rs
+++ b/webrender/examples/yuv.rs
@@ -16,7 +16,7 @@ use std::env;
 use std::path::PathBuf;
 use webrender_traits::{ColorF, Epoch};
 use webrender_traits::{DeviceIntPoint, DeviceUintSize, LayoutPoint, LayoutRect, LayoutSize};
-use webrender_traits::{ImageData, ImageDescriptor, ImageFormat};
+use webrender_traits::{ImageData, ImageDescriptor, ImageFormat, ImageRendering};
 use webrender_traits::{PipelineId, TransformStyle};
 use webrender_traits::{YuvColorSpace, YuvData};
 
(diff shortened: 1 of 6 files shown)


kvark:
Is there ever a case where YUV would need linear sampling? I'd expect this to produce colouring artefacts.
Other than this question, the code looks good and ready to merge.

JerryShih:
@kvark 
I hit an assertion with this branch now. So, please don't merge this patch.

JerryShih:
@kvark 
> Is there ever a case where YUV would need linear sampling? I'd expect this to produce colouring artefacts.
> Other than this question, the code looks good and ready to merge.

The yuv image uses  "ImageRendering::Auto" now. That will use TextureFilter::Linear.
So, I add an additional option for the filter setting.

                let filter = match request.rendering {
                    ImageRendering::Pixelated => TextureFilter::Nearest,
                    ImageRendering::Auto | ImageRendering::CrispEdges => TextureFilter::Linear,
                };

kvark:
@JerryShih I understand what you did. My question is - do we ever need linear sampling on YUV?

JerryShih:
> @JerryShih I understand what you did. My question is - do we ever need linear sampling on YUV?

I don't know. But, maybe we will have the better quality when we enlarge the video size.

JerryShih:
@kvark 
r? again for the https://github.com/servo/webrender/pull/1310/commits/7fa2494f13a2d5a0ea5fcf9b0670ec5a359e7ce2

kvark:
I thought a bit about using linear sampling on YUV in general, and talked to Markus.
TL;DR: it should be supported, but we may need further enhancements in the shader to make it look correct.

@bors-servo r+

PR #1312, opened 2017-05-31
Switch a couple of gpu profile scopes to markers.

glennw:
These markers were outside the gpu_profile begin/end frame calls.

These changes fix the GPU stalls I've been seeing on my machine in the profiler.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1312)


kvark:
We should just assert on the markers used outside of begin/end, to prevent this behavior in the future.

PR #1313, opened 2017-05-31
Remove glyph raster profile scope, for now.

glennw:
This spams out errors to the console about running on an unregistered profiler thread.

This is a quick fix until we get the rayon worker threads to register with the profiler.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1313)


glennw:
r? @kvark 

cc @nical

nical:
Ah, indeed! I looked at the doc and you are referring to https://docs.rs/rayon/0.7.1/rayon/struct.Configuration.html#method.start_handler. That sounds perfect!

kvark:
I suppose it's safe to close in favour of #1318

PR #1315, opened 2017-05-31
Require at least plane-split 0.4.

staktrace:
Older versions of plane-split (e.g. 0.3.0) use an older version of
euclid (0.11 instead of 0.13), and so webrender doesn't actually compile
with that. The compilation failures come from a mismatch of type names,
such as euclid::rect::TypedRect vs euclid::TypedRect.

The Cargo.lock file already uses plane-split 0.4.0 but downstream
dependencies of webrender that try to use it with an older plane-split
will run into this problem.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1315)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -30,7 +30,7 @@ webrender_traits = {path = "../webrender_traits"}
 bitflags = "0.7"
 gamma-lut = "0.2"
 thread_profiler = "0.1.1"
-plane-split = ">= 0.3, < 0.5"
+plane-split = "0.4"
 
 [dev-dependencies]
 angle = {git = "https://github.com/servo/angle", branch = "servo"}


staktrace:
Shoot. The failures seem to be just OS X timeouts. Can somebody else convince bors to retry?

webrender/Cargo.toml
```diff
@@ -30,7 +30,7 @@ webrender_traits = {path = "../webrender_traits"}
 bitflags = "0.7"
 gamma-lut = "0.2"
 thread_profiler = "0.1.1"
-plane-split = ">= 0.3, < 0.5"
+plane-split = ">= 0.4, < 0.5"
```
kvark: this is equivalent to just "0.4"

staktrace: True. I'll update the patch

PR #1317, opened 2017-05-31
Debug assertions to prevent GPU markers out of frame

kvark:
Includes #1312
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1317)


kvark:
@bors-servo r=emilio
(as per IRC #servo)

PR #1318, opened 2017-05-31
Register the worker threads with the thread profiler.

nical:
As an alternative to #1313.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1318)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1068,7 +1068,9 @@ impl Renderer {
         let render_target_debug = options.render_target_debug;
         let payload_tx_for_backend = payload_tx.clone();
         let recorder = options.recorder;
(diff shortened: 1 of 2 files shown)


webrender/src/renderer.rs
```diff
@@ -1068,7 +1068,9 @@ impl Renderer {
         let render_target_debug = options.render_target_debug;
         let payload_tx_for_backend = payload_tx.clone();
         let recorder = options.recorder;
-        let worker_config = ThreadPoolConfig::new().thread_name(|idx|{ format!("WebRender:Worker#{}", idx) });
+        let worker_config = ThreadPoolConfig::new()
+            .thread_name(|idx|{ format!("WebRender:Worker#{}", idx) })
+            .start_handler(|idx| { register_thread_with_profiler(format!("WebRender:Worker#{}", idx)); });
```
kvark: the unfortunate redundancy is addressed in https://github.com/glennw/thread_profiler/pull/7

PR #1320, opened 2017-06-01
Avoid calculating a tight bounding box for text runs.

glennw:
With the previous tiling code, this was quite important. However,
it's no longer necessary and was showing up high in CPU profiles.

On some sites (such as Wikipedia and HN) that contain a lot of
text, this saves ~17% of the CPU time per frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1320)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -746,6 +746,9 @@ impl FrameBuilder {
             return
         }
 
+        // Expand the rectangle of the text run by the blur radius.
+        let rect = rect.inflate(blur_radius, blur_radius);
+
         // TODO(gw): Use a proper algorithm to select
         // whether this item should be rendered with
         // subpixel AA!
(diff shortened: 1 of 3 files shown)


webrender/src/frame_builder.rs
```diff
@@ -1754,17 +1757,11 @@ impl<'a> LayerRectCalculationAndCullingPass<'a> {
...
                                                                  &packed_layer.local_clip_rect,
                                                                  self.device_pixel_ratio) {
-                if self.frame_builder.prim_store.prepare_prim_for_render(prim_index,
-                                                                         self.resource_cache,
-                                                                         &packed_layer.transform,
-                                                                         self.device_pixel_ratio,
-                                                                         display_list) {
-                    self.frame_builder.prim_store.build_bounding_rect(prim_index,
-                                                                      self.screen_rect,
+                self.frame_builder.prim_store.prepare_prim_for_render(prim_index,
```
kvark: don't we still want to check for the returned value here?

webrender/src/prim_store.rs
```diff
@@ -1324,38 +1321,31 @@ impl PrimitiveStore {
...
                         dest_glyphs[actual_glyph_count] = GpuBlock16::from(GlyphPrimitive {
                             padding: LayerPoint::zero(),
-                            offset: local_glyph_rect.origin,
+                            offset: glyph_pos,
                         });
 
                         text.glyph_instances.push(GlyphInstance {
                             index: src.index,
-                            point: LayoutPoint::new(src.point.x, src.point.y),
+                            point: glyph_pos,
```
kvark: hmm, this does seem to change the logic. The old `src.point` is not equivalent to `(x,y)`

webrender/src/frame_builder.rs
glennw: prepare_prim_for_render no longer returns a value.

webrender/src/prim_store.rs
glennw: I don't believe it changes the logic - that value is only used to select the subpixel offset, and the dimensions fields added to it are always whole values.

kvark: ah, indeed. This asks to change `GlyphIntstance::point` to something that better reflects the semantics

webrender/src/frame_builder.rs
kvark: true, GH hid the function name in the diff due to multi-line parameters ;)

PR #1323, opened 2017-06-01
Port box shadows to use the GPU cache.

glennw:
Previously, two allocations were made - one for the main box shadow
structure and one for the instance rectangle geometry.

Now, a single allocation is used and the instance rectangles follow
the box shadow structure.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1323)


kvark:
The further we go, the less semantics we preserve for the data crossing CPU-GPU boundary (i.e. now we have a specific layout that a shader expects at a specified address), and that scares me a little.
@bors-servo r+

PR #1327, opened 2017-06-01
Move GPU cache from prim store to resource cache.

glennw:
The benefit of this is it will allow us to allocate GPU blocks
for the resource rects (the UV rects of items in the texture cache)
which will allow us to simplify and optimize the CPU and shader code
for primitives that use texture cache items.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1327)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1385,7 +1385,6 @@ impl FrameBuilder {
         let mut profile_counters = FrameProfileCounters::new();
         profile_counters.total_primitives.set(self.prim_store.prim_count());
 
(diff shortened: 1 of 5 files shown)


glennw:
r? @kvark

Instead of clearing / recycling the cache when a new DL is received, we just let items expire from the cache now.

kvark:
Looks reasonable, although it's a matter of convenience. Perhaps, at the end of the day, gpu cache should graduate by leaving parents and existing on its own? :)
@bors-servo r+

glennw:
Yup, long term it probably makes sense for it to exist standalone :)

PR #1328, opened 2017-06-02
Port aligned, angle and radial gradients to GPU cache.

glennw:
Also remove the gpu64 data texture, as it's now unused.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1328)


glennw:
Pushed a couple of follow up commits on to this PR - port the text run primitive, and remove the temporary GpuLocation enum.

webrender/src/prim_store.rs
```diff
@@ -314,20 +287,13 @@ pub struct GradientPrimitiveCpu {
     pub cache_dirty: bool,
     pub gpu_data_address: GpuStoreAddress,
     pub gpu_data_count: i32,
+    pub gpu_blocks: [GpuBlockData; 3],
```
kvark: I find it a bit inconsistent that some stuff composes `GpuBlockData` on demand while other just keeps `gpu_blocks` locally. If the semantics of `XxxPrimitiveCpu` is supposed to allow easy property manipulation, we should compose `GpuBlockData` on the fly (in `write_gpu_blocks`) for everything. If it's not supposed to be manipulated, then we can have local `[GpuBlockData]` members, and we can replace the `write_gpu_blocks(&self, &mut Vec<>)` into `get_gpu_blocks(&self) -> &[]`.

webrender/src/prim_store.rs
```diff
@@ -1200,44 +1166,57 @@ impl PrimitiveStore {
...
-                    }
-                    PrimitiveKind::YuvImage => {
-                        let yuv_image = &cpu_yuv_images[cpu_prim_index.0];
-                        yuv_image.write_gpu_blocks(blocks);
-                    }
-                    _ => {
-                        unreachable!("Encountered a type not supported by GPU cache!");
-                    }
+        // TODO(gw): Restructure this code so the borrow checker
+        //           doesn't force the locals below!
```
kvark: oh my, this is getting out of hand!
One option, perhaps the most idiomatic one, would be to have a custom iterator returned by `request_iter()`, so that we can iterate it without a closure

kvark: implemented in #1336

PR #1329, opened 2017-06-02
Fix a bug setting the gpu cache last access time.

glennw:
The last access time for a block is initialized when the free
list for a row is created.

The bug manifests in the following way:
 * New row is allocated on frame 1 - blocks access time = 1
 * New display list is received on frame 2.
 * Blocks are allocated from the same row, but the last_access_time was not set.
 * A debug assert occurs during get_address - where it checks that the block was requested this frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1329)

### DIFF
--- a/webrender/src/gpu_cache.rs
+++ b/webrender/src/gpu_cache.rs
@@ -311,6 +311,7 @@ impl Texture {
 
         // Add the block to the occupied linked list.
         block.next = self.occupied_list_head;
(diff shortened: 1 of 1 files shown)


glennw:
Oooops!!

r? @kvark

kvark:
Uh oh
@bors-servo r+

PR #1330, opened 2017-06-02
Avoid unnecessary struct field shorthand

glandium:
This was inadvertently added in 3ab7de9, and unnecessarily requires
rustc 1.17 (struct field shorthands were unstable before that).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1330)


kvark:
@glandium what is the problem with requiring Rust-0.17?
Servo is built with rust-nightly anyway, and any app using Rust should be safe to update the stable version.

PR #1331, opened 2017-06-02
Don't apply local clip rect to snap rectangle calculation.

glennw:
This fixes the remaining WPT/CSS fails in the pending WR update.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1331)


kvark:
The old code used local clip rect for snapping, and I wasn't sure why.
Good to see more tests passing!
@bors-servo r+

kvark:
OSX timeout
@bors-servo retry

PR #1333, opened 2017-06-02
Return the clip id from push_clip_node.

staktrace:
In cases where the caller doesn't pass in an explicit clip id, it is
still useful for the caller to get back the generated clip id. This is
because the caller may later need to call push_clip_and_scroll_info with
the clip id that was generated. If this function doesn't return the clip
id the caller would basically have to inline this function and define
the clip manually, which seems much less convenient.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1333)

### DIFF
--- a/webrender_traits/src/display_list.rs
+++ b/webrender_traits/src/display_list.rs
@@ -881,9 +881,11 @@ impl DisplayListBuilder {
     pub fn push_clip_node(&mut self,
                           content_rect: LayoutRect,
                           token: ClipRegionToken,
-                          id: Option<ClipId>){
+                          id: Option<ClipId>)
+                          -> ClipId {
         let id = self.define_clip(content_rect, token, id);
         self.clip_stack.push(ClipAndScrollInfo::simple(id));
+        id
     }
 
     pub fn push_clip_id(&mut self, id: ClipId) {


kvark:
Technically, a breaking change. Unlikely to break anyone though.
@bors-servo r+

mrobinson:
@staktrace Hrm. I originally thought of `push_clip_node` as a way to keep backward compatibility until everyone moved to define_clip and pushing it manually. Is there any way to Gecko can switch to this approach? Eventually, I think the clip stack might go away entirely.

kvark:
@mrobinson if you don't feel good about this change, please r-

staktrace:
@mrobinson yeah we can stop using push_clip_node if that's the preferred way forward. I thought it was here to stay as a nice convenience function.

staktrace:
(Aside: is there a way to mark functions as deprecated in rust? If not even some inline comments to that effect would be good here)

kvark:
@staktrace yeah,
```rust
#[deprecated]
fn foo() {}

fn main() {
    foo();
}
```

mrobinson:
@kvark @staktrace In the end, I didn't r- only because the clipping API is moving a bit slowly in order to make sure that it meets the needs of Gecko. If the `push_clip_node` API is very useful, I am happy to adapt the to that style. On the other hand, if the `define_clip` approach works, then let's stick with that. :smile:

kvark:
@mrobinson ok, let's re-open if @staktrace has any issues with `define_clip`

staktrace:
Yeah, no problem with using `define_clip`. `push_clip_node` is pretty small and trivial to inline into gecko where needed.

mrobinson:
@staktrace Thanks for the info. I'm tracking the removal here: https://github.com/servo/webrender/issues/1335

PR #1334, opened 2017-06-02
Run CI with rustc 1.15.1

glandium:
This is the minimum version supported by Gecko, and would avoid things
like 3ab7de9 slipping in.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1334)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: trusty
 language: rust
 rust:
-  - stable
+  - 1.15.1
   - nightly
 os:
   - linux


kvark:
This brings the total Travis job count to astonishing 12. I'm concerned about the increased build times as well as risks associated with getting OSX build failures due to timeouts. Perhaps, we could test "0.15.1" instead of "stable" to keep the job count under control?

jrmuizel:
I thought Gecko required Rust stable and not older versions like 1.15?

rillian:
Gecko currently requires rust >= 1.15.1. Our goal is stable, but we try to avoid bumping the minimum required version as a courtesy for developers who aren't working on Rust code. I.e. it reduces churn for them.

The minimum will jump to 1.17.0 in a week or two when Gecko starts building the style crate by default.

glennw:
Would it make sense to remove the ```nightly``` check, and just test against ```stable``` and the current Gecko minimum?

rillian:
That might be reasonable.

Gecko recently [added a check](https://bugzilla.mozilla.org/show_bug.cgi?id=1321847) that the code builds with the declared minimum Rust version. There are also (much less frequent) build checks against beta and nightly Rust. Having similar tests in the webrender repo will catch the issues much sooner of course. I think it depends which end of the compatibility spectrum you expect to bump into most often.

glennw:
Thoughts @kvark  ?

kvark:
@glennw I still see the value in testing Nightly, so that any compiler/stdlib changes affecting us are discovered early. For that matter, I stick to my earlier proposal to just test Rust 1.15 instead of "stable", and keep it up with Gecko needs.

glennw:
@kvark Sounds good to me.

glennw:
@bors-servo r+

Thanks!

PR #1336, opened 2017-06-02
GpuDataRequest

kvark:
based on #1328
addresses soundness of requests, refactors the usage in prim_store
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1336)


glennw:
Nice one!

r=me (I'm assuming you'll need to rebase after the included commits land).

PR #1337, opened 2017-06-03
Bump euclid to 0.14

nical:
The most notable changes in euclid are that points and vectors are now separate types, plus a whole lot of API adjustments (Transform3D instead of Matrix4D, etc).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1337)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use euclid::Point3D;
 use geometry::ray_intersects_rect;
 use mask_cache::{ClipSource, MaskCacheInfo, RegionMode};
 use prim_store::GpuBlock32;
@@ -13,7 +12,8 @@ use util::TransformedRectKind;
 use webrender_traits::{ClipId, ClipRegion, DeviceIntRect, LayerPixel, LayerPoint, LayerRect};
 use webrender_traits::{LayerSize, LayerToScrollTransform, LayerToWorldTransform, PipelineId};
 use webrender_traits::{ScrollClamping, ScrollEventPhase, ScrollLayerRect, ScrollLocation};
-use webrender_traits::{WorldPoint, WorldPoint4D};
+use webrender_traits::{WorldPoint, LayerVector2D};
+use webrender_traits::{as_scroll_parent_vector};
 
 #[cfg(target_os = "macos")]
 const CAN_OVERSCROLL: bool = true;
@@ -105,7 +105,7 @@ pub struct ClipScrollNode {
     /// The scroll offset of all the nodes between us and our parent reference frame.
     /// This is used to calculate intersections between us and content or nodes that
     /// are also direct children of our reference frame.
-    pub reference_frame_relative_scroll_offset: LayerPoint,
+    pub reference_frame_relative_scroll_offset: LayerVector2D,
 
     /// Pipeline that this layer belongs to
     pub pipeline_id: PipelineId,
@@ -138,7 +138,7 @@ impl ClipScrollNode {
             combined_local_viewport_rect: LayerRect::zero(),
             world_viewport_transform: LayerToWorldTransform::identity(),
             world_content_transform: LayerToWorldTransform::identity(),
-            reference_frame_relative_scroll_offset: LayerPoint::zero(),
+            reference_frame_relative_scroll_offset: LayerVector2D::zero(),
             parent: Some(parent_id),
             children: Vec::new(),
             pipeline_id: pipeline_id,
@@ -160,7 +160,7 @@ impl ClipScrollNode {
             combined_local_viewport_rect: LayerRect::zero(),
             world_viewport_transform: LayerToWorldTransform::identity(),
             world_content_transform: LayerToWorldTransform::identity(),
-            reference_frame_relative_scroll_offset: LayerPoint::zero(),
+            reference_frame_relative_scroll_offset: LayerVector2D::zero(),
             parent: parent_id,
             children: Vec::new(),
             pipeline_id: pipeline_id,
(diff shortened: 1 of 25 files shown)


glennw:
Ideally we'd get https://github.com/servo/servo/pull/17104 landed before merging this (hoping to finally get that landed today).

There's also a servo-tidy failure on CI we'll need to fix.

Do you have a corresponding Servo patch that can be used for the next WR update?

nical:
> Do you have a corresponding Servo patch that can be used for the next WR update?

I am working through the parts of servo that need to be updated for one by one. Unfortunately it's impractical to do without the bumped crates being published on crates.io when porting the crates that depend on them (because cargo refuses to do anything with a crate version that is not published, even if I try to `[replace]` it locally). So it is going take a while.

glennw:
OK, the WR update has landed in Servo now.

What's the best way to proceed with this so that we don't get in a spot where we are unable to update WR for some time?

nical:
> What's the best way to proceed with this so that we don't get in a spot where we are unable to update WR for some time?

I am working on this, I'm just new to the art of bumping all servo crates at once so I am progressing slowly and learning along the way. If we feel uneasy about this, now is a good time to do a WebRender update (just before we land this) in case I get stuck shaving the dependency yak for too long.

nical:
I got servo to work with euclid 0.14 (see [the enormous WIP PR](https://github.com/servo/servo/pull/17184)).

glennw:
OK, I'll try to land a WR update today, and keep landing incremental WR updates until we get this landed.

nical:
The servo part is almost ready now (only waiting for this to land). We are good to land this whenever you want. Perhaps just after a servo wr update so that if a regression is introduced in wr in the mean time, the fix does not get trapped into the euclid bump.

webrender/src/clip_scroll_node.rs
```diff
@@ -410,22 +407,16 @@ impl ClipScrollNode {
...
 
-        let p0 = inv.transform_point4d(&WorldPoint4D::new(cursor.x, cursor.y, z0, 1.0));
-        let p0 = Point3D::new(p0.x / p0.w,
-                              p0.y / p0.w,
-                              p0.z / p0.w);
-        let p1 = inv.transform_point4d(&WorldPoint4D::new(cursor.x, cursor.y, z1, 1.0));
-        let p1 = Point3D::new(p1.x / p1.w,
-                              p1.y / p1.w,
-                              p1.z / p1.w);
+        let p0 = inv.transform_point3d(&WorldPoint3D::new(cursor.x, cursor.y, z0));
```
kvark: nice transform call :)
For the vector conversion from 2d to 3d, I'm actually a big fan of `cgmath` approach: https://docs.rs/cgmath/0.14.1/cgmath/struct.Vector2.html#method.extend

You call `Vector2::extend(self, z: f32) -> Vector3`. It's more explicit then current euclid's `to_3d` and allows cases like this ^ to look somewhat nicer

webrender/src/util.rs
```diff
@@ -221,30 +221,25 @@ impl TransformedRect {
             TransformedRectKind::Complex => {
                 */
                 let vertices = [
-                    transform.transform_point4d(&LayerPoint4D::new(rect.origin.x,
+                    transform.transform_point3d(&LayerPoint3D::new(rect.origin.x,
```
kvark: could use `to_3d()` for all those cases?

webrender_traits/Cargo.toml
```diff
@@ -14,11 +14,11 @@ webgl = ["offscreen_gl_context"]
 app_units = "0.4"
 bincode = "1.0.0-alpha2"
 byteorder = "1.0"
-euclid = "0.13"
+euclid = "0.14.2"
```
kvark: btw, why `0.14.2` specifically?

webrender_traits/src/units.rs
```diff
@@ -48,8 +51,10 @@ pub struct LayerPixel;
 
 pub type LayerRect = TypedRect<f32, LayerPixel>;
 pub type LayerPoint = TypedPoint2D<f32, LayerPixel>;
```
kvark: looks inconsistent now, should be `LayerPoint2D`

webrender_traits/src/units.rs
```diff
@@ -63,6 +68,7 @@ pub struct ScrollLayerPixel;
 
 pub type ScrollLayerRect = TypedRect<f32, ScrollLayerPixel>;
 pub type ScrollLayerPoint = TypedPoint2D<f32, ScrollLayerPixel>;
+pub type ScrollLayerVector2D = TypedVector2D<f32, ScrollLayerPixel>;
```
kvark: also inconsistent

nical:
> error[E0411]: use of `Self` outside of an impl or trait

.. euclid 0.14.2 apparently uses something that made it into rustc after 1.15 :(
I'll have to make an euclid 0.14.3

webrender_traits/Cargo.toml
nical: 0.14.0  has the dreaded vector.to_point() bug (xyy). My understanding is that by specifying 0.14.2 I can get any version between 0.14.2 and 0.15.

webrender_traits/src/units.rs
nical: Yeah, I don't mind either way.
We can make this a followup if you feel strongly about it. This will be an easy PR to write but will touch pretty much anything that isn't already in the diff so I'd rather not have it in this one.

webrender/src/clip_scroll_node.rs
nical: I like it. Will add this to euclid.

webrender/src/util.rs
nical: Yes indeed.

Eijebong:
@nical I think since you added a commit it should have been try not retry :p

nical:
@Eijebong thanks
@bors-servo try

Eijebong:
What's holding this one back ?

kvark:
Because it was a try.
@bors-servo r+

PR #1338, opened 2017-06-05
Some cleanups of the texture cache API.

glennw:
This is prep work for allowing the texture cache to use the GPU cache for UV rect storage.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1338)


kvark:
much nicer!
@bors-servo r+

PR #1339, opened 2017-06-05
Revert the snap rect changes.

glennw:
I tried several fixes to avoid having to revert this - but it's causing all kinds of weird machine-specific failures on the CI machines.

It's also blocking us updating several Servo dependencies at the moment, which are blocked on the WR update for euclid, heapsize etc.

We can revisit this after we unblock and land the next WR update.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1339)


glennw:
r? @kvark 

Sorry!

kvark:
@kvark is sad 😿 
@bors-servo r+

PR #1341, opened 2017-06-05
[WIP] Immediate rectangle rendering

kvark:
That's an experiment to provide immediate mode API for drawing Gecko's UI.
It's not clear how to properly provide the textures for it. Eventually, Gecko may just use regular WR's display lists for UI.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1341)

### DIFF
--- a/webrender/res/im_rectangle.fs.glsl
+++ b/webrender/res/im_rectangle.fs.glsl
@@ -1,0 +1,8 @@
+#line 1
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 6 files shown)


glennw:
This seems quite reasonable - would we plan to add batching of the immediate items in the future? Do you want this reviewed yet or should I hold off until the [WIP] is removed?

kvark:
@glennw the PR is here for your reference (as well as Gecko folks), so no need to review it just yet. It's not clear yet if it needs to be finished, or Gecko will just use WR display items for UI.

jrmuizel:
Just to close the loop here, it looks like we're going to just append items directly to the display list for now so we don't need this.

glennw:
Closing based on comment above from @jrmuizel

PR #1342, opened 2017-06-05
Add resizing support to the GPU cache.

glennw:
When the allocated rows exceed 512, add another 512 rows to
the texture. Maintain two texture IDs and blit between them
to copy existing data when resizing.

Resizing should be very rare - most pages that I've tested on
never exceed 50 rows in the GPU cache at a time.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1342)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -194,37 +194,53 @@ pub enum BlendMode {
 
 /// The device-specific representation of the cache texture in gpu_cache.rs
 struct CacheTexture {
-    id: TextureId,
+    current_id: TextureId,
+    next_id: TextureId,
 }
 
 impl CacheTexture {
     fn new(device: &mut Device) -> CacheTexture {
-        let id = device.create_texture_ids(1, TextureTarget::Default)[0];
+        let ids = device.create_texture_ids(2, TextureTarget::Default);
 
         CacheTexture {
-            id: id,
+            current_id: ids[0],
+            next_id: ids[1],
         }
     }
 
     fn update(&mut self, device: &mut Device, updates: &GpuCacheUpdateList) {
         // See if we need to create or resize the texture.
-        let current_dimensions = device.get_texture_dimensions(self.id);
-
+        let current_dimensions = device.get_texture_dimensions(self.current_id);
         if updates.height > current_dimensions.height {
-            // TODO(gw): Handle resizing an existing cache texture.
-            if current_dimensions.height > 0 {
-                panic!("TODO: Implement texture copy!!!");
-            }
-
             // Create a f32 texture that can be used for the vertex shader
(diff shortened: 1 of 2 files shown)


glennw:
r? @kvark 

With the parameters provided, it never resizes on any page I tested. I tested this locally by setting initial size to 1 and increment size to 1. It seemed to be working correctly, looking at the apitrace output.

webrender/src/renderer.rs
```diff
@@ -194,37 +194,52 @@ pub enum BlendMode {
...
                                 ImageFormat::RGBAF32,
                                 TextureFilter::Nearest,
-                                RenderTargetMode::None,
+                                RenderTargetMode::SimpleRenderTarget,
                                 None);
+
+            // Copy the current texture into the newly resized texture.
+            if current_dimensions.height > 0 {
+                let new_dimensions = device.get_texture_dimensions(self.next_id);
+                device.bind_draw_target(Some((self.next_id, 0)), Some(new_dimensions));
```
kvark: no need to pass the viewport in there

webrender/src/renderer.rs
```diff
@@ -194,37 +194,52 @@ pub enum BlendMode {
...
+
+                let blit_rect = DeviceIntRect::new(DeviceIntPoint::zero(),
+                                                   DeviceIntSize::new(MAX_VERTEX_TEXTURE_WIDTH as i32,
+                                                                      current_dimensions.height as i32));
+
+                device.blit_render_target(Some((self.current_id, 0)),
```
kvark: We should probably use `glCopyTexSubImage2D`, given that we don't stretch/filter/resolve here.

webrender/src/renderer.rs
```diff
@@ -194,37 +194,52 @@ pub enum BlendMode {
...
 
 impl CacheTexture {
     fn new(device: &mut Device) -> CacheTexture {
-        let id = device.create_texture_ids(1, TextureTarget::Default)[0];
+        let ids = device.create_texture_ids(2, TextureTarget::Default);
 
         CacheTexture {
-            id: id,
+            current_id: ids[0],
+            next_id: ids[1],
```
kvark: given the rarity (and limitedness) of this resize operation, I wouldn't bother keeping a separate texture ID and would just create one on the spot (temporarily)

webrender/src/renderer.rs

webrender/src/renderer.rs
glennw: Added a TODO comment here to fix this in the future (not sure if that's bound in gleam yet).

glennw:
@kvark Thanks, addressed those comments.

kvark:
Thanks!
@bors-servo r+

PR #1343, opened 2017-06-06
Move GpuCache out of the ResourceCache and store in RenderBackend.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1343)


glennw:
r? @kvark 

As you suggested originally, it makes more sense for this to be separated from the resource cache.

kvark:
Closing in favour of #1344

PR #1344, opened 2017-06-06
Update glyphs to use the GPU cache.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1344)


glennw:
r? @kvark 

This includes the commit from https://github.com/servo/webrender/pull/1343.

PR #1347, opened 2017-06-06
Create an example app illustrating the use of the animations API

staktrace:
The first patch extracts a bunch of boilerplate from the scrolling.rs example app so that new example apps can be written with less effort and cargo-culting of code. Because of the way example apps are built, I resorted to using an include! to do this refactoring, but I'm open to suggestions on better ways of doing this.

The second patch actually adds the new example app that uses the animations API.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1347)

### DIFF
--- a/webrender/examples/animation.rs
+++ b/webrender/examples/animation.rs
@@ -1,0 +1,85 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+extern crate gleam;
+extern crate glutin;
+extern crate webrender;
+extern crate webrender_traits;
+
+#[macro_use]
+extern crate lazy_static;
+
+#[path="common/boilerplate.rs"]
+mod boilerplate;
+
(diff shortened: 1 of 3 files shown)


staktrace:
Added some minor edits: (1) generified the title of the window in the boilerplate code and (2) reused the `bounds` variable in animation.rs a little more.

webrender/examples/animation.rs
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+include!("boilerplate.rs");
```
kvark: this is extremely useful to have the boilerplate isolated from the examples, but I think there is a better way to share it. See this link (with a `common` folder) for an example - https://github.com/brendanzab/cgmath/tree/master/benches

glennw: Yup, agreed. This is really great to have the common code separated, but it'd be good to avoid the include!() macro.

glennw:
@bors-servo r+

Thanks!

PR #1348, opened 2017-06-07
Port gradients to use GPU cache.

glennw:
* Remove old gradient textures.
* Gradient LUTs are now stored in a single row of RGBAF32 texels.
* The interpolation is done in the fragment shader.
* Build GPU blocks on demand for the gradients.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1348)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -106,12 +106,36 @@ varying vec3 vClipMaskUv;
     flat varying vec4 vLocalBounds;
 #endif
 
+// TODO(gw): This is here temporarily while we have
+//           both GPU store and cache. When the GPU
+//           store code is removed, we can change the
+//           PrimitiveInstance instance structure to
+//           use 2x unsigned shorts as vertex attributes
+//           instead of an int, and encode the UV directly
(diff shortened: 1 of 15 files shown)


webrender/res/prim_shared.glsl
```diff
@@ -848,33 +851,34 @@ vec4 dither(vec4 color) {
...
     //     offsets < 0 use the first color entry, 0
     //     offsets from [0, 1) use the color entries in the range of [1, N-1)
     //     offsets >= 1 use the last color entry, N-1
     //     so transform the range [0, 1) -> [1, N-1)
-    float gradient_entries = 0.5 * gradient_size.x;
-    x = x * (gradient_entries - 2.0) + 1.0;
+
+    // TODO(gw): In the future we might consider making the size of the
+    // LUT vary based on number / distribution of stops in the gradient.
+    const int GRADIENT_ENTRIES = 128;
```
kvark: I guess we'll no longer have a problem of reduced quality because the entries size was reduced from 128 to 126

glennw:
Rebased.

PR #1355, opened 2017-06-09
Discard noop filters

mrobinson:
These filters can just be ignored because they should not affect
output. This change also improves the style in the filters code
a little bit.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1355)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -91,26 +91,23 @@ impl StackingContextHelpers for StackingContext {
                                   properties: &SceneProperties) -> Vec<LowLevelFilterOp> {
         let mut filters = vec![];
         for filter in display_list.get(input_filters) {
+            if filter.is_noop() {
+                continue;
+            }
+
             match filter {
                 FilterOp::Blur(radius) => {
-                    filters.push(LowLevelFilterOp::Blur(
-                        radius,
-                        AxisDirection::Horizontal));
-                    filters.push(LowLevelFilterOp::Blur(
-                        radius,
-                        AxisDirection::Vertical));
+                    filters.push(LowLevelFilterOp::Blur(radius, AxisDirection::Horizontal));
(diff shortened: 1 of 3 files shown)


webrender/src/frame.rs
```diff
@@ -93,46 +93,35 @@ impl StackingContextHelpers for StackingContext {
         for filter in display_list.get(input_filters) {
             match filter {
                 FilterOp::Blur(radius) => {
-                    filters.push(LowLevelFilterOp::Blur(
-                        radius,
-                        AxisDirection::Horizontal));
-                    filters.push(LowLevelFilterOp::Blur(
-                        radius,
-                        AxisDirection::Vertical));
+                    filters.push(LowLevelFilterOp::Blur(radius,AxisDirection::Horizontal));
```
emilio: nit: space after comma.

webrender/src/frame.rs
```diff
@@ -93,46 +93,35 @@ impl StackingContextHelpers for StackingContext {
...
                 FilterOp::Invert(amount) => {
-                    filters.push(
-                            LowLevelFilterOp::Invert(Au::from_f32_px(amount)));
+                    filters.push(LowLevelFilterOp::Invert(Au::from_f32_px(amount)));
                 }
                 FilterOp::Opacity(ref value) => {
                     let amount = properties.resolve_float(value, 1.0);
-                    filters.push(
-                            LowLevelFilterOp::Opacity(Au::from_f32_px(amount)));
+                    filters.push(LowLevelFilterOp::Opacity(Au::from_f32_px(amount)));
```
emilio: Why not avoiding pushing instead of calling `retain` below?

mrobinson:
Thanks for the review. the latest version of patch takes @emilio's suggestions and also increases the range of filters that we can discard.

PR #1356, opened 2017-06-09
Make the blob sample app reuse the boilerplate code

staktrace:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1356)

### DIFF
--- a/webrender/examples/blob.rs
+++ b/webrender/examples/blob.rs
@@ -10,18 +10,16 @@ extern crate webrender;
 extern crate webrender_traits;
 extern crate rayon;
 
-use gleam::gl;
+#[path="common/boilerplate.rs"]
+mod boilerplate;
+
 use rayon::ThreadPool;
 use rayon::Configuration as ThreadPoolConfig;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
 use std::sync::Arc;
 use std::sync::mpsc::{channel, Sender, Receiver};
-use webrender_traits::{BlobImageData, BlobImageDescriptor, BlobImageError, BlobImageRenderer, BlobImageRequest};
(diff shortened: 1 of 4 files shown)


webrender/examples/blob.rs
```diff
@@ -10,18 +10,16 @@ extern crate webrender;
...
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
 use std::sync::Arc;
 use std::sync::mpsc::{channel, Sender, Receiver};
-use webrender_traits::{BlobImageData, BlobImageDescriptor, BlobImageError, BlobImageRenderer, BlobImageRequest};
-use webrender_traits::{BlobImageResult, TileOffset, ImageStore, ColorF, ColorU, Epoch};
-use webrender_traits::{DeviceUintSize, DeviceUintRect, LayoutPoint, LayoutRect, LayoutSize};
-use webrender_traits::{ImageData, ImageDescriptor, ImageFormat, ImageRendering, ImageKey, TileSize};
-use webrender_traits::{PipelineId, RasterizedBlobImage, TransformStyle};
+use webrender_traits::*;
```
kvark: I believe this is generally non-idiomatic, but so tempting!
I'm thinking if something like `use webrender_traits as wt` for later prefixing with `wt` would make it more explicit?

staktrace: My main gripe is that we're importing so many things that any time you make any change to the sample app you have to go tweak the list of things imported. Since the app is so closely bound to the webrender_traits code I don't think it makes sense to individually list all the things it's importing, even if this is non-idiomatic.

I would wholly agree that there should be at most one wildcard import in any given app, if that helps any.

staktrace:
Added a couple of patches, the second one does what you suggested with the `wt::` prefix. Personally I prefer the wildcard import but I can live with this if you prefer.

Issue #809, opened 2017-01-30
Investigate the RGBA / BGRA weirdness in internal formats.


glennw:
Make this more consistent, which will involve sorting out PackedTexel struct. This could be part of the larger "make the image formats more sane" task.

kvark:
We explicitly use BGRA as an internal format only on GL ES, where there is `GL_BGRA_EXT` that we require. We implicitly assume that RGBA8 internal format on GL is actually BGRA, thus we expect the incoming data to always be BGRA for performance reasons (so that the driver doesn't need to swizzle).

nical:
Fixed in #1360.

PR #1360, opened 2017-06-10
Rename RGBA8 to BGRA8 to match what it actually does

Thinkofname:
~~`glTexSubImage2D`'s format parameter is the format of the data that
is being uploaded not the format of the target texture. This was
causing RGBA textures to have their red and blue channels swapped.~~

~~Edit: Servo seems to byte-swap here https://github.com/servo/servo/blob/9656341f43cc1bc415e07a0a0c919740280ad2ab/components/net_traits/image/base.rs#L74 so maybe this is intended? Seems confusing given the format is called RGBA~~

Reworked into just rename RGBA8 to BGRA8

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1360)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,8 +22,8 @@ dependencies = [
  "serde_json 0.9.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-glutin 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.41.0",
- "webrender_traits 0.41.0",
+ "webrender 0.42.0",
+ "webrender_traits 0.42.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
@@ -987,7 +987,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "webrender"
-version = "0.41.0"
+version = "0.42.0"
 dependencies = [
  "angle 0.1.2 (git+https://github.com/servo/angle?branch=servo)",
  "app_units 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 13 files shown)


glennw:
There is an existing issue here to investigate this - https://github.com/servo/webrender/issues/809.

It's likely to be quite an involved task though, as we'll need to determine what all the current callers do (primarily Servo and Gecko). For Servo, this would mean checking all the image loading paths (e.g. normal images, fallback canvas / webgl paths etc).

That would give us a better idea of what we want to do here.

Thinkofname:
In servo I can only find:

* Images: https://github.com/servo/servo/blob/9656341f43cc1bc415e07a0a0c919740280ad2ab/components/net_traits/image/base.rs#L74
* WebGL: https://github.com/servo/servo/blob/e45d7e0e2e5d47945d3d821db5767feff64d7634/components/canvas/webgl_paint_thread.rs#L251 
                https://github.com/servo/servo/blob/07b0770d21d2b374f9a30c98cdde5ae5531aaecb/components/script/dom/webglrenderingcontext.rs#L699
* Canvas: https://github.com/servo/servo/blob/e45d7e0e2e5d47945d3d821db5767feff64d7634/components/canvas/canvas_paint_thread.rs#L545 , all the `byte_swap` calls and a few places where pixels are pushed by hand.

As for gecko I wouldn't know where to start. It might be easier to rename RGBA to BGRA and fix up servo and gecko and then add RGBA back later so things can be moved across to it one piece at a time.

glennw:
The renaming plan you've suggested above sounds good to me. Thoughts @kvark  ?

kvark:
>`glTexSubImage2D`s format parameter is the format of the data that
is being uploaded not the format of the target texture. This was
causing RGBA textures to have their red and blue channels swapped.

Where did the bug occur exactly?

The current logic of WR assumes that all 8bit four-channel internal formats are BGRA, and expects the data provided in this format accordingly. So it's not a mistake in the code, it's just about assumptions that are not properly typed/documented.

I don't see a reason why we'd want to stick to the mess GL types are, and I totally welcome the rename of RGBA -> BGRA, because that would better express what we are already doing. I don't think, however, that we'll need to add RGBA afterwards.

Thinkofname:
@kvark Changed the PR to a rename from RGBA8 to BGRA8.

The main reasoning about adding RGBA8 back later was to match the fact webrender supports RG and RGB textures. 
This wouldn't effect servo (unless it wanted to let the driver do the byteswapping) and would mostly be for standalone users of webrender. 
Although they can just do the byteswapping like servo does so it isn't really an issue.

Thinkofname:
@kvark Done

kvark:
Please squash

Thinkofname:
Squashed, sorry about that.

kvark:
@jrmuizel will Gecko be fine with this change? It's not changing any logic, just clarifying semantics.

jrmuizel:
Yeah, this is fine.

PR #1364, opened 2017-06-12
Start splitting apart ScrollFrames and Clips

mrobinson:
ScrollFrames and clips are currently represented by one node type in
the ClipScrollTree, which means that many nodes are doing more work
than they need to do. This makes even less sense in a world where all
clips are added to the ClipScrollTree and handled the same way in the
API and internally. This change is the first step to splitting apart
these two types of items.

In order to minimize the changes, existing clips/ScrollFrame are
pushed as two nodes. This might almost double the depth of the tree,
but a later patch should expose API for pushing a clip without a
ScrollFrame attached.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1364)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -72,14 +72,14 @@ pub enum NodeType {
 
     /// Other nodes just do clipping, but no transformation.
     Clip(ClipInfo),
+
+    /// Other nodes just do clipping, but no transformation.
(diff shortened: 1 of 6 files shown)


mrobinson:
The test change here is just fixing a mistake in the original test, which specified a `content-size` property. This property is no longer supported.

glennw:
@mrobinson Nice one!

@bors-servo r+

PR #1365, opened 2017-06-12
Give blob images access to font data

jrmuizel:
This factors out the ImageStore trait into a more
general BlobImageResources trait and allows the renderer to
access it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1365)

### DIFF
--- a/webrender_traits/src/image.rs
+++ b/webrender_traits/src/image.rs
@@ -5,6 +5,7 @@
 use std::sync::Arc;
 use {DeviceUintRect, DevicePoint};
 use {TileOffset, TileSize};
+use font::{FontKey, FontTemplate};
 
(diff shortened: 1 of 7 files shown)


webrender_traits/src/image.rs
```diff
@@ -182,8 +190,4 @@ pub enum BlobImageError {
 pub struct BlobImageRequest {
     pub key: ImageKey,
     pub tile: Option<TileOffset>,
-}
-
-pub trait ImageStore {
-    fn get_image(&self, key: ImageKey) -> Option<(&ImageData, &ImageDescriptor)>;
-}
+}
```
kvark: nit: Missing newline

PR #1370, opened 2017-06-13
share the render backend thread

nical:
It would be good to share the render backend thread between several windows. This would waste less resources and more importantly make it a lot easier to share larger resources between render backends (fonts, etc.).

This attempt achieves that by letting you create a RenderBackendThread struct and pass it as parameter in the creation of the Renderer. All api senders push commands to the same message channel with an id to select the render backend. Having to pass an Id isn't great, I hope that there aren't cases where this could be a security issue (if, says, a compromised child process can spoof the id).

The alternative is to use a condition variable to wake up the render backend thread. It is probably better but I don't know if we have a cross-plaftorm and cross-process condvar handy.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1370)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use gleam::gl;
 use frame::Frame;
 use frame_builder::FrameBuilderConfig;
 use gpu_cache::GpuCache;
@@ -12,19 +13,23 @@ use resource_cache::ResourceCache;
 use scene::Scene;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
-use std::sync::mpsc::Sender;
+use std::sync::mpsc::{Sender, channel};
+use std::thread;
+use std::io;
+use std::rc::Rc;
+use renderer::{Renderer, RendererOptions, InitError};
 use texture_cache::TextureCache;
 use time::precise_time_ns;
 use thread_profiler::register_thread_with_profiler;
 use rayon::ThreadPool;
 use webgl_types::{GLContextHandleWrapper, GLContextWrapper};
-use api::channel::{MsgReceiver, PayloadReceiver, PayloadReceiverHelperMethods};
+use api::channel::{MsgSender, msg_channel, PayloadReceiver, PayloadReceiverHelperMethods};
 use api::channel::{PayloadSender, PayloadSenderHelperMethods};
 use api::{ApiMsg, BlobImageRenderer, BuiltDisplayList, DeviceIntPoint};
 use api::{DeviceUintPoint, DeviceUintRect, DeviceUintSize, IdNamespace, ImageData};
-use api::{LayerPoint, PipelineId, RenderDispatcher, RenderNotifier};
+use api::{LayerPoint, PipelineId, RenderDispatcher, RenderNotifier, RenderApiSender};
 use api::{VRCompositorCommand, VRCompositorHandler, WebGLCommand, WebGLContextId};
-use api::{FontTemplate};
+use api::{FontTemplate, RendererId};
 
 #[cfg(feature = "webgl")]
 use offscreen_gl_context::GLContextDispatcher;
@@ -37,7 +42,6 @@ use webgl_types::GLContextDispatcher;
 ///
 /// The render backend operates on its own thread.
 pub struct RenderBackend {
-    api_rx: MsgReceiver<ApiMsg>,
     payload_rx: PayloadReceiver,
     payload_tx: PayloadSender,
     result_tx: Sender<ResultMsg>,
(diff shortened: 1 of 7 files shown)


jrmuizel:
How does have having the condition variable help?

nical:
> How does have having the condition variable help?

The main issue is being able to push into several queues without blocking the thread on one in particular.

glennw:
Is the idea here to share an instance between the main window and the small top level windows that Gecko creates?

I have concerns about sharing a backend thread between multiple top-level windows (but maybe that's not the goal?). That would put us in a situation where the performance of rendering in one top level window directly affects the rendering performance of another window.

Could you expand a bit on the use case for this (e.g. how many instances would we have otherwise, and what are the savings we're expecting to get from this)?

nical:
Ideally I'd like to share the render backend thread for all windows. That's something that could be tweaked if it turns out that sharing the thread for too many top-level windows causes issues. Most of the consumer hardware doesn't have that many cores, so if we run into this we are already kind of in trouble. We could have a notion of focused window and prioritize render backends in the thread's event loop in function of that.
Moving render backends to the same thread would make it easier to share resources like fonts and even rasterized glyphs. Currently we serialize and sanitize all fonts for each window which is not great.

In firefox every menu, tooltip or popup is its own window, so at least these should be sharing their backend thread.

webrender/src/render_backend.rs
```diff
@@ -66,392 +66,382 @@ pub struct RenderBackend {
 
     next_webgl_id: usize,
 
-    vr_compositor_handler: Arc<Mutex<Option<Box<VRCompositorHandler>>>>
+    vr_compositor_handler: Arc<Mutex<Option<Box<VRCompositorHandler>>>>,
+
+    frame_counter: u32,
+}
+
+pub struct RenderBackendInit {
```
kvark: we need a `Default` implementation for it

webrender/src/render_backend.rs
```diff
@@ -566,3 +556,91 @@ impl GLContextDispatcher for WebRenderGLDispatcher {
...
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum BackendStatus {
+    Continue,
+    Remove,
+}
+
+impl RenderBackendThread {
+    pub fn new(thread_name: String) -> Result<Self, ::std::io::Error> {
```
kvark: nit: can we have `use std::io` instead of a global import here?

webrender/src/render_backend.rs
```diff
@@ -566,3 +556,91 @@ impl GLContextDispatcher for WebRenderGLDispatcher {
...
+        let (api_tx, api_rx) = try! { msg_channel() };
+        let (init_tx, init_rx) = channel();
+
+        try!{ thread::Builder::new().name(thread_name.clone()).spawn(move || {
```
kvark: nit: this would probably look nicer with `?`

webrender/src/render_backend.rs
```diff
@@ -566,3 +556,91 @@ impl GLContextDispatcher for WebRenderGLDispatcher {
...
+
+            register_thread_with_profiler(thread_name);
+
+            let mut render_backends = HashMap::new();
+
+            while let Ok((api_msg, id)) = api_rx.recv() {
+
+                while let Ok(init_msg) = init_rx.try_recv() {
```
kvark: This looks a bit wonky to me. How about having a separate channel for each backend instance, so that this code waits for all {backend1, backend2, ..., backendN, init} channels at once?

webrender/examples/basic.rs
```diff
@@ -253,7 +256,7 @@ fn main() {
     };
 
     let size = DeviceUintSize::new(width, height);
-    let (mut renderer, sender) = webrender::renderer::Renderer::new(gl, opts, size).unwrap();
+    let (mut renderer, sender) = webrender::renderer::Renderer::new(gl, opts, size, &mut backend_thread).unwrap();
```
kvark: we could have a `new_renderer` method on the `backend_thread`

nical:
> This looks a bit wonky to me. How about having a separate channel for each backend instance, so that this code waits for all {backend1, backend2, ..., backendN, init} channels at once?

I must have missed something. How do you wait for several channels at once?

kvark:
Ah, I was thinking about [select!](https://doc.rust-lang.org/std/macro.select.html), which is not stable, nevermind :)
Would it look cleaner if we merge `api_tx` with `init_tx` so that there is a single wait?

nical:
I didn't know about select, it's great! I wonder if we can do what the macro does manually (or if it uses unstable features under the hood), and if it would work with ipc channels.

The api_tx and init_tx might not be of the same type (api_tx beeing an ipc channel while init_tx is always a simple mpsc channel). Also, having the same channel would mean that the RenderBackendInit structure (which is currently internal to webrender) would have to be exposed in webrender_traits which isn't cleaner in my opinion.

glennw:
@nical I think this can probably be closed now, given our discussions in SF about work that will supersede this?

nical:
The SF plans diverged from something that would supersede this. We ended up with something where we could have several versions of a given scene in flight but not necessarily several scenes going to different renderers.
At the same time I have been looking into ways to not use webrender for small popup windows in firefox (in order to avoid needing this) but it turns out to be complicated due to various sources of technical debt.

So I think we should still iterate on this and get things to a point where we can share threads and resources between render backends that talk to different renderers.

kvark:
@nical The main idea, at the end of the day, is to make `RenderBackend` to process different documents (or versions of the same document) asynchronously by treating various processing phases as jobs with specified dependencies.

So while this PR can be accepted as a temporary improvement, until the new solution is in place, that new solution will essentially replace it.

nical:
There is a key difference between processing two versions of the same document and two documents going to different windows. The latter needs to be built on top of something that at least resemble what we have in this PR while the former doesn't need most of the channel changes involved here and can work with render backends working on their own threads which would not address what this PR is for.

So I propose that we keep this PR (or elements of it) for the channel changes that let us have several windows on the same render backend thread, and iterate from there to address the APZ issues.

glennw:
I don't quite follow - it seems to me that if the RB can handle different documents, that would cover the use case for separate windows? Is that your understanding of what we discussed @kvark ?

Unless we're sure we need something like this, my preference would be to not merge this until we get the multi-document support landed, since this seems like a performance optimization that shouldn't block any work in the interim. That way, once the multi-document work is complete, we can re-assess if we still need this extra functionality. Thoughts?

nical:
The two are almost orthogonal.
If we focus on the task of addressing APZ issues, the render backend will be able to handle the documents that were sent to it. Typically you would have two versions of the same display list sent through a given pipeline, and that way we can do the dance of building the display list version N+1 asynchronously while we scroll version N. In this context it does not matter that two windows have separate render backend threads since all we need is for the render backend of a given window to be able to handle several versions of the same scene for a given window.

This PR is really about having several windows sharing more resources (starting, with threads and in the future hopefully fonts etc.)
If we want the larger multi-documents project to encompass having multiple windows handled by the same render backend, then this PR is only a baby step towards the rest and we can iterate from it. 


Maybe we have slightly different idea of what the multi-document thing will look like. What are the things in this PR that you think we would have to throw away and re-implement differently in the context of the multi-document render backend? If it is just the question having a map of render backends versus having a render backend with a map of scenes, I wouldn't block on this detail.

webrender_api/src/api.rs
```diff
@@ -334,95 +348,95 @@ impl RenderApi {
...
         let msg = ApiMsg::RequestWebGLContext(*size, attributes, tx);
-        self.api_sender.send(msg).unwrap();
+        self.api_sender.send((msg, self.renderer_id)).unwrap();
         rx.recv().unwrap()
     }
 
     pub fn resize_webgl_context(&self, context_id: WebGLContextId, size: &DeviceIntSize) {
         let msg = ApiMsg::ResizeWebGLContext(context_id, *size);
-        self.api_sender.send(msg).unwrap();
+        self.api_sender.send((msg, self.renderer_id)).unwrap();
```
kvark: we could have a helper method for this

webrender/src/render_backend.rs
kvark: it turns out the `select!` is unstable, so this is a non-issue

nical:
> There are orthogonal parts, but at the core of your PR is the split of RB into multiple instances running on the same thread. This goes somewhat counter to the future plan of having a single opaque RB running on multiple threads (sounds quite the opposite, doesn't it? :) ).

This was only the simplest way to get the thread sharing up and running, I'll happily change it if that's the problem. That said it isn't necessarily in contradiction with the future APZ plans: we can either have each separate render backend manage several versions of the same Scene, or a single render backend manage several versions of all scenes. You are implying the latter which is fine by me, the former would work as well as far as APZ is concerned. We didn't go into this level of detail when we talked about the second version in SF (first version being the jobs system but I wasn't there so I don't know).

If this is the only thing blocking this PR, I much prefer to change it now than revisit this later, there are things that would be simpler in gecko if we had this, and this PR bitrots fast with rebases that aren't easy due to the extra level of indentation all over render_backend.rs that makes git think everything was rewritten.

> This is definitely helpful. I wonder, how much benefit does it do to Gecko? Is it a noticeable difference that we'll enjoy while the new RB solution is being developed?

If you think of the new RB solution as a superset of this PR, then you should think of this as step 1 of the new RB solution (I still consider them independent things with independent goals and requirements).

There is certainly overhead in duplicating fonts and threads for each window (in gecko, you ll have at least 4 windows even if you only have one main window because of popups, dropdown menus and the likes). The other thing is that sharing the thread would make it a lot simpler for gecko to shut down properly (only have to send a sync round trip to one RB + renderer thread), and moving resources from one window to another (drag a tab from a window to another) is currently more complicated than initially expected due to the fact that we can't move a resource from a render backend to another (being on different threads).

nical:
> I'm inclined to agree that we'll need to rewrite the core parts of this PR

By all means, all I ask is that we get specific about these and I'll rewrite them now.

glennw:
@nical Sorry, I didn't look through this in detail today - I'll run through it first thing in the morning and write up some more detailed info.

glennw:
@nical I tried to write up my thoughts below, but it's all very hand-wavy. Perhaps we can arrange a Vidyo conference between me, you and @kvark to try and work out the details and how to proceed?

---

First, a vague description of what I had in mind long term:

1) Add a new concept, called a "Document".
2) This document includes all the information relevant to a given page - e.g.:
	* display lists
	* pipeline map
	* animation properties map
	* window size
	* current zoom / pan
3) Modify RenderBackend to contain a list of Documents.
4) Each RenderBackend is a single thread, that contains:
	* channel rx/tx
	* gpu cache / resource cache (long term - so these are shared between documents in a single RB)
	* event loop that fires jobs to build a frame from a document, rasterize glyphs etc.
5) Each renderer object is modified to be able to handle multiple documents (that can share a single texture cache).
6) Each document may have some concept of revisions / generations (for APZ purposes).
7) Each document can only exist in a single renderer and a single RB, but there is no requirement for a 1:1 mapping between documents and renderers or RB.

---

Conceptually, this is quite similar to what's already in this PR (just with different naming).

However, from what I can see, this PR provides a shared OS thread, but doesn't address how to share the resource / glyph / texture cache.

I think we should have a rough plan for how we will do resource sharing (starting with the use cases we want to support) before we start merging PRs towards this functionality, since I think the resource sharing design will fundamentally affect how we should structure the code to share render backends / documents.

There's a lot of subtleties and assumptions with how the resource and texture cache interact between the renderer and RB structures at the moment, and I think how we solve those will fundamentally affect how we structure the code to share render backends / documents / renderers / APZ revisions.

Specifically, there's not much separation between the concept of CPU cached glyphs (that have been rasterized) vs. GPU cached glyphs (that exist in the texture cache), and there is an assumption that there is a 1:1 mapping between the RB-side texture cache and the renderer texture cache.

How we handle this depends on whether we're talking about a shared texture cache (if the documents / revisions are using the same renderer, and therefore GL context), or only a shared CPU-side glyph cache (in the case of different top-level windows).

glennw:
Hmmm, the more I think about APZ as I wrote that up, the more I think it might be a lot simpler to solve than we originally thought. Need to think about it a bit more, but I think we can get the functionality we want without a concept of revisions / generations etc...

nical:
@glennw We had a long meeting in SF where we devised another plan for APZ. The problem with the initial multi-document plan with several threads is that the core of webrender revolves around a fundamentally single-threaded and stateful use of the resource cache. If we just move documents to several threads it means that they can't share the resource cache and in particular, a second version of the same scene that is built asynchronously could not share the cache with the previous version of the scene which is too much duplicated work. The second version was based on splitting the things that need direct access to the resource cache out from the ones that don't and run the latter asynchronously while the the former was sharing the render backend thread (at least for two versions of the scene) in a more integrated way.

Let's schedule a vidyo conf and talk about this. It felt like we had it figured out in SF but it was just silly that we forgot you were not at this meeting and we didn't sync up with you. Also let's move this to another thread.

glennw:
@nical @kvark and I did some discussions on a whiteboard where we thought we had a good way to support multiple documents, despite the shared resource cache issues you mention above. We should definitely sync up on Vidyo though, it's much easier to discuss rather than via GH issue.

glennw:
@nical Should we close this one, given our discussions today and plans for the PR @kvark is working on?

nical:
Sounds good.

PR #1376, opened 2017-06-14
Optimize CPU time spent in building glyph instances.

glennw:
Previously, the CPU would query the dimensions of each glyph, and
then apply the dimensions offset from the rasterized glyph to the
instance data. This results in a lot of hash lookups to query the
dimensions (and also some additional math on the CPU).

Instead, allow storing "user data" with each texture cache item. This
allows the glyph offsets to be applied in the vertex shader instead
of on the CPU.

Because this is such a hot code path, it provides significant CPU time
wins - often up to 20% of the CPU time on real pages.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1376)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -725,13 +725,23 @@ TransformVertexInfo write_transform_vertex(RectWithSize instance_rect,
 
 #endif //WR_FEATURE_TRANSFORM
 
-struct ResourceRect {
+struct GlyphResource {
+    vec4 uv_rect;
+    vec2 offset;
+};
+
+GlyphResource fetch_glyph_resource(int address) {
+    vec4 data[2] = fetch_from_resource_cache_2(address);
+    return GlyphResource(data[0], data[1].xy);
+}
+
+struct ImageResource {
     vec4 uv_rect;
 };
 
-ResourceRect fetch_resource_rect(int address) {
+ImageResource fetch_image_resource(int address) {
     vec4 data = fetch_from_resource_cache_1(address);
-    return ResourceRect(data);
+    return ImageResource(data);
 }
 
 struct Rectangle {
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -16,11 +16,11 @@ use std::mem;
 use std::usize;
 use util::{TransformedRect, recycle_vec};
 use webrender_traits::{BuiltDisplayList, ColorF, ImageKey, ImageRendering, YuvColorSpace};
-use webrender_traits::{YuvFormat, ClipRegion, ComplexClipRegion, ItemRange, GlyphKey};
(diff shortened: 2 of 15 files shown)


webrender/src/prim_store.rs
```diff
@@ -970,33 +970,14 @@ impl PrimitiveStore {
                 let src_glyphs = display_list.get(text.glyph_range);
 
                 // Cache the glyph positions, if not in the cache already.
+                // TODO(gw): In the future, remove `glyph_instances`
+                //           completely, and just reference the glyphs
+                //           directly from the displaty list.
                 if text.glyph_instances.is_empty() {
-                    let mut glyph_key = GlyphKey::new(text.font_key,
-                                                      font_size_dp,
```
kvark: `font_size_dp` is unused now, should be issuing a warning?

webrender/res/prim_shared.glsl
```diff
@@ -727,11 +727,12 @@ TransformVertexInfo write_transform_vertex(RectWithSize instance_rect,
 
 struct ResourceRect {
     vec4 uv_rect;
+    vec2 user_data;
```
kvark: I'd like to avoid blob fields unless really necessary. Could we just pass this data in a separate resource cache entry, so that it's strongly typed?

webrender/src/prim_store.rs
glennw: It's still used in the ```request_glyphs``` call below.

webrender/res/prim_shared.glsl
glennw: I agree with the sentiment, but since glyph handling is so high in the profile (it's typically 60+% of the frame time), I think making this path as fast as possible is probably worth it here. Perhaps I could rename it to be glyph offset specifically, since I don't see it ever being used for something else?

kvark: Does it have to be `ResourceRect` then? Maybe we can have `GlyphRect` struct for this.

glennw:
@kvark Added a second commit which splits up resource rect into ImageResource and GlyphResource. It will need a (small) rebase once https://github.com/servo/webrender/pull/1383 lands, but should be ready for re-review otherwise.

glennw:
@kvark Split up the resource rects into GlyphResource and ImageResource, rebased, squashed and fixed the CI error.

webrender/res/prim_shared.glsl
```diff
@@ -725,13 +725,23 @@ TransformVertexInfo write_transform_vertex(RectWithSize instance_rect,
 
 #endif //WR_FEATURE_TRANSFORM
 
-struct ResourceRect {
+struct GlyphResource {
     vec4 uv_rect;
+    vec2 user_data;
```
kvark: can we name it "offset" or something more specific?

glennw: Yup, fixed.

kvark:
Thanks!
@bors-servo r+

PR #1380, opened 2017-06-14
Make push_prebuilt_display_list a memcpy

### LINKED ISSUE #1378: Make push_prebuilt_display_list a simple memcpy
This is necessary so that we can properly parent new clip and scroll frames in prebuilt display lists.

mrobinson:
Make push_prebuilt_display_list a memcpy by adding internal support
for nested display lists. We extend the existing replacements code to
replace references to the root scrolling node id with the id of the
node that surrounds the built display list. This a little complicated
because we maintain support for multiple levels of nesting. This can be
removed if we decide that it isn't necessary.

We also rename the method to push_nested_display_list to match the
internal naming.

Fixes #1378.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1380)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -37,6 +37,7 @@ struct FlattenContext<'a> {
     builder: &'a mut FrameBuilder,
     resource_cache: &'a mut ResourceCache,
     replacements: Vec<(ClipId, ClipId)>,
+    nested_display_list_info: Vec<(ClipId, ClipId)>,
 }
 
 impl<'a> FlattenContext<'a> {
@@ -49,14 +50,52 @@ impl<'a> FlattenContext<'a> {
             builder: builder,
             resource_cache: resource_cache,
             replacements: Vec::new(),
+            nested_display_list_info: Vec::new(),
         }
     }
 
-    fn clip_id_with_replacement(&self, id: ClipId) -> ClipId {
-        match self.replacements.last() {
+    fn push_nested_display_list_ids(&mut self, info: ClipAndScrollInfo) {
+        self.nested_display_list_info.push((info.scroll_node_id, info.clip_node_id()));
+    }
+
+    fn pop_nested_display_list_ids(&mut self) {
+        self.nested_display_list_info.pop();
+    }
+
+    fn apply_clip_and_scroll_info_replacements(&self, info: &mut ClipAndScrollInfo) {
+        info.scroll_node_id = self.apply_scroll_frame_id_replacement(info.scroll_node_id);
+
+        let mut clip_node_id = match info.clip_node_id {
+            Some(id) => id,
+            None => return,
+        };
+
+        for &(_, nested_clip_node_id) in self.nested_display_list_info.iter().rev() {
+            if clip_node_id.is_root_scroll_node() {
+                clip_node_id = nested_clip_node_id;
+            } else {
+                break;
+            }
+        }
+
+        info.clip_node_id = Some(clip_node_id);
+    }
+
+    fn apply_scroll_frame_id_replacement(&self, id: ClipId) -> ClipId {
+        let mut id = match self.replacements.last() {
             Some(&(to_replace, replacement)) if to_replace == id => replacement,
             _ => id,
+        };
+
+        for &(nested_scroll_id, _) in self.nested_display_list_info.iter().rev() {
+            if id.is_root_scroll_node() {
+                id = nested_scroll_id;
+            } else {
+                break;
+            }
         }
+
+        id
     }
 }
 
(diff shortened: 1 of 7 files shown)


webrender_traits/src/display_list.rs
```diff
@@ -909,42 +905,22 @@ impl DisplayListBuilder {
...
-            // Then reinsert the actual item, updating its clip_and_scroll
-            self.push_item(*item.item(), item.rect());
-
-            // Then handle implicit suffix items
-            match *item.item() {
-                SpecificDisplayItem::Text(_)                => self.push_range(item.glyphs(), &dl),
-                SpecificDisplayItem::PushStackingContext(_) => self.push_range(item.filters(), &dl),
-                _ => { /* do nothing */ }
-            }
-        }
```
Gankro: 🎉

webrender_traits/src/display_list.rs
```diff
@@ -909,42 +905,22 @@ impl DisplayListBuilder {
...
+    // will replace references to the root scroll frame id with the current scroll frame
+    // id.
+    //
+    // New clips and scroll frames are only supported if they can be guaranteed to have
+    // a unique id (ie a unique id must be passed when creating them).
+    pub fn push_nested_display_list(&mut self, mut built_display_list: BuiltDisplayList) {
+        self.push_clip_region(&LayoutRect::zero(), vec![], None);
+        self.push_new_empty_item(SpecificDisplayItem::PushNestedDisplayList);
+
+        self.data.append(&mut built_display_list.data);
```
Gankro: Two things:

* make this take a &BuiltDisplayList, so we won't have to change this whenever gecko decides to try to reuse DisplayList allocations.

* use `extend_from_slice` rather than `append`. Doesn't need mutability, and is guaranteed to optimize to a memcopy even in debug builds.

webrender/src/frame.rs
```diff
@@ -49,14 +50,52 @@ impl<'a> FlattenContext<'a> {
...
-        match self.replacements.last() {
+    fn push_nested_display_list_ids(&mut self, info: ClipAndScrollInfo) {
+        self.nested_display_list_info.push((info.scroll_node_id, info.clip_node_id()));
+    }
+
+    fn pop_nested_display_list_ids(&mut self) {
+        self.nested_display_list_info.pop();
+    }
+
+    fn clip_and_scroll_info_with_replacements(&self, info: &mut ClipAndScrollInfo) {
```
Gankro: This should probably be a verb like `apply_clip_and_scroll_info_replacements`.

webrender/src/frame.rs
```diff
@@ -49,14 +50,52 @@ impl<'a> FlattenContext<'a> {
...
+                clip_node_id = nested_list_info.1;
+            } else {
+                break;
+            }
+        }
+
+        info.clip_node_id = Some(clip_node_id);
+    }
+
+    fn scroll_frame_id_with_replacement(&self, id: ClipId) -> ClipId {
```
Gankro: This one's more borderline on verbness required. Your call.

webrender/src/frame.rs
```diff
@@ -49,14 +50,52 @@ impl<'a> FlattenContext<'a> {
...
+            None => return,
+        };
+
+        for nested_list_info in self.nested_display_list_info.iter().rev() {
+            if clip_node_id.is_root_scroll_node() {
+                clip_node_id = nested_list_info.1;
+            } else {
+                break;
+            }
+        }
```
Gankro: Consider changing this to `for &(_, new_clip_node_id) in ... { ... }`

webrender/src/frame.rs
```diff
@@ -49,14 +50,52 @@ impl<'a> FlattenContext<'a> {
...
             _ => id,
+        };
+
+        for nested_list_info in self.nested_display_list_info.iter().rev() {
+            if id.is_root_scroll_node() {
+                id = nested_list_info.0;
+            } else {
+                break;
+            }
         }
```
Gankro: Same for-loop clarification suggestion.

webrender/src/frame.rs
Gankro: Just checking my understanding here: it should be possible to exit this loop with `clip_node_id.is_root_scroll_node() == true`, right? That would be if there weren't any nested scrolling contexts at all?

webrender/src/frame.rs
mrobinson: I think it makes sense to call it `apply_scroll_frame_id_replacement`.

mrobinson:
So I think the answer to your design question is that this prevents the second deserialization for a little bit of complication in the pre-existing replacement code as well as in wrench. I think this will also make it a bit less awkward to add full support for clips and scroll frames in nested display lists, since we will have the option of adding dynamic clip ids without worrying about collisions.

Gankro:
r+ from me; I think it's acceptable to be a bit blurry in the API boundary here to make it easier to explore the design.

Issue #1369, opened 2017-06-13
Rename webrender_traits into webrender_api


nical:
The crate name webrender_traits is misleading at best. It would be nice to publish webrender on crates.io but before doing this I would like webrender_traits to have a less confusing name. 

Thoughts ?

cc @glennw @kvark

Gankro:
I agree, this is the right name for it.

PR #1381, opened 2017-06-14
Rename webrender_traits into webrender_api, reexport in webrender.

nical:
This does the change I proposed in #1369, and reexports webrender_api in webrender under the webrender::api namespace. This way things that depend on both webrender and webrender_api (everything except servo at the moment), can just have a dependency to webrender and use the reexported api.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1381)


appveyor.yml
```diff
@@ -12,7 +12,7 @@ install:
 build: false
 
 test_script:
-  - cd webrender_traits
+  - cd webrender_api
   - cargo test --verbose
```
kvark: we should probably be using `cargo test --all`

glennw:
Seems like a good time to land this, when there's not many large PRs in progress - any objections?

PR #1383, opened 2017-06-15
Fix a bug with UV rects for blob images in the GPU cache.

glennw:
Previously, the texture cache ID for blob images would only
get added to the requested items hash set when the blob
was new or out of date. This change ensures that blob images
get the associated texture cache ID added to the requested
items hash set unconditionally.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1383)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -392,10 +392,21 @@ impl ResourceCache {
         }
         if template.data.is_blob() {
             if let Some(ref mut renderer) = self.blob_image_renderer {
-                let same_epoch = match self.cached_images.resources.get(&request) {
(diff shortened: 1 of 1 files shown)


webrender/src/resource_cache.rs
```diff
@@ -392,11 +392,22 @@ impl ResourceCache {
...
+                    }
+                    None => {
+                        (false, None)
+                    }
                 };
 
+                // Ensure that blobs are added to the list of requested items
+                // foe the GPU cache, even if the cached blob image is up to date.
+                if let Some(texture_cache_id) = texture_cache_id {
+                    self.requested_items.insert(texture_cache_id);
```
kvark: why can't we do this right inside the previous `match`?

PR #1386, opened 2017-06-15
Add support for new clips within nested display lists

### LINKED ISSUE #1379: Properly handle clips and scroll frames in push_prebuilt_display_list
We need this so that we can remove per-item masks and complex clips.

mrobinson:
This will be necessary in order to support removal of the per-item
complex clip and mask on display items.

Fixes #1379.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1386)

### DIFF
--- a/webrender_traits/src/display_item.rs
+++ b/webrender_traits/src/display_item.rs
@@ -581,16 +581,18 @@ impl ComplexClipRegion {
     }
 }
 
+pub type NestingIndex = u64;
+
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub enum ClipId {
-    Clip(u64, PipelineId),
+    Clip(u64, NestingIndex, PipelineId),
     ClipExternalId(u64, PipelineId),
     DynamicallyAddedNode(u64, PipelineId),
 }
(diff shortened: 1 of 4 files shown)


webrender_traits/src/display_item.rs
```diff
@@ -583,14 +583,14 @@ impl ComplexClipRegion {
 
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub enum ClipId {
-    Clip(u64, PipelineId),
+    Clip(u64, u64, PipelineId),
```
kvark: let's have a `NestingLevel` typedef?

webrender_traits/src/display_item.rs
```diff
@@ -624,7 +624,14 @@ impl ClipId {
 
     pub fn is_root_scroll_node(&self) -> bool {
         match *self {
-            ClipId::Clip(id, _) if id == 0 => true,
+            ClipId::Clip(id, nesting_level, _) if id == 0 && nesting_level == 0 => true,
```
kvark: can we do `ClipId::Clip(0, 0, _)`?

PR #1387, opened 2017-06-15
Apply clip-and-scroll wrench property to stacking contexts

mrobinson:
Due to a bug in wrench the clip-and-scroll property was not being
applied to stacking contexts. This change fixes that issue.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1387)

### DIFF
--- a/wrench/reftests/scrolling/reftest.list
+++ b/wrench/reftests/scrolling/reftest.list
@@ -7,3 +7,4 @@
 == scroll-layer-with-mask.yaml scroll-layer-with-mask-ref.yaml
 == scroll-layer.yaml scroll-layer-ref.yaml
 == simple.yaml simple-ref.yaml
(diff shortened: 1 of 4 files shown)


wrench/reftests/scrolling/reftest.list
```diff
@@ -7,3 +7,4 @@
 == scroll-layer-with-mask.yaml scroll-layer-with-mask-ref.yaml
 == scroll-layer.yaml scroll-layer-ref.yaml
 == simple.yaml simple-ref.yaml
+== wrench-clip-and-scroll-property.yaml wrench-clip-and-scroll-property-ref.yaml
```
kvark: I don't think we need `wrench` in the name here, given it's in wrench reftests

mrobinson:
@kvark Thanks for the review. I've updated the branch with your suggestion.

kvark:
Sweet, thank you!
@bors-servo r+

PR #1389, opened 2017-06-15
[RFC] Add a rawtest framework

jrmuizel:
This adds a rawtesting framework that lets tests do anything
that the webrender api allows. This allows writing tests
that do more sophisticated things then the yaml can express.

The framework includes a single test that tests that when
we draw a blob image a second time we get the same results.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1389)

### DIFF
--- a/wrench/src/blob.rs
+++ b/wrench/src/blob.rs
@@ -1,0 +1,139 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// A very basic BlobImageRenderer that can only render a checkerboard pattern.
+
+use std::collections::HashMap;
+use webrender::api::*;
+
+
+// Serialize/deserialze the blob.
+
+pub fn serialize_blob(color: ColorU) -> Vec<u8> {
+    vec![color.r, color.g, color.b, color.a]
+}
+
+fn deserialize_blob(blob: &[u8]) -> Result<ColorU, ()> {
+    let mut iter = blob.iter();
+    return match (iter.next(), iter.next(), iter.next(), iter.next()) {
+        (Some(&r), Some(&g), Some(&b), Some(&a)) => Ok(ColorU::new(r, g, b, a)),
+        (Some(&a), None, None, None) => Ok(ColorU::new(a, a, a, a)),
+        _ => Err(()),
+    }
+}
+
+// This is the function that applies the deserialized drawing commands and generates
+// actual image data.
+fn render_blob(
+    color: ColorU,
+    descriptor: &BlobImageDescriptor,
+    tile: Option<TileOffset>,
+) -> BlobImageResult {
+    // Allocate storage for the result. Right now the resource cache expects the
+    // tiles to have have no stride or offset.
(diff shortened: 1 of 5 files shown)


glennw:
Neat! I've been thinking we need something like this. I wonder if it might be more useful to have raw tests (perhaps called code tests?) support comparing to an image.

It seems to me that a common pattern would be write a raw test, save the output as a png, and then compare the raw test output with the png each test run. Thoughts?

wrench/src/blob.rs
```diff
@@ -0,0 +1,145 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// A very basic BlobImageRenderer that can only render a checkerboard pattern.
+
+extern crate app_units;
```
kvark: I don't think we should be including `extern crate` items in a non-root module

jrmuizel:
Re: comparing to png. We could totally do this. I think we'll have to wait and see what kinds of tests develop with this framework. If things become repetitive it should be easy for us to build appropriate abstractions.

wrench/src/rawtest.rs
```diff
@@ -0,0 +1,106 @@
...
+        wrench.renderer.set_render_notifier(Box::new(Notifier { tx: tx }));
+
+        RawtestHarness {
+            wrench: wrench,
+            rx: rx,
+        }
+    }
+
+    pub fn run(mut self) {
+        self.retained_blob_images_test();
```
glennw: Each raw test should probably be in a separate source file.

wrench/src/rawtest.rs
```diff
@@ -0,0 +1,106 @@
...
+            LayoutSize::new(0.0, 0.0),
+            ImageRendering::Auto,
+            blob_img,
+        );
+        self.wrench.api.set_display_list(root_background_color,
+                                         Epoch(0),
+                                         layout_size,
+                                         builder.finalize(),
+                                         false);
+        self.wrench.api.generate_frame(None);
```
glennw: It seems like most of the code in this function is boilerplate that would be duplicated for each raw test. Although it's good to have complete control when necessary for a raw test, it probably makes sense to have a way of a test providing two display lists, and then some common code draws, reads and compares the two display lists?

jrmuizel: I'll clean things up a bit and rerequest review.

glennw:
@jrmuizel What's the status of this? Is it ready for re-review?

glennw:
ping @jrmuizel

jrmuizel:
It's not ready for rereview yet. I haven't had a chance to clean it up to my liking.

jrmuizel:
It's not as clean as I wanted but I'm sick of having it out of tree. Review please.

wrench/src/blob.rs
```diff
@@ -0,0 +1,139 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// A very basic BlobImageRenderer that can only render a checkerboard pattern.
+
+use std::collections::HashMap;
+use webrender::api::*;
+
+
+// Serialize/deserialze the blob.
```
glennw: nit: spelling

PR #1390, opened 2017-06-16
Optimize time spent in GPU driver for GPU cache updates.

glennw:
We keep a shadow copy of the GPU cache data in the render thread.
After all updates (patches) have been applied for this frame, scan
the rows of the texture for rows that have been modified.

Upload each dirty row to the GPU via a PBO to ensure that there
are no CPU-side stalls.

In the future, there's several more optimizations that can be made:
 * Batch consecutive row updates into a single PBO / upload call.
 * Perhaps track start/end of each row, to avoid a full row update for small changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1390)


glennw:
r? @kvark 

This gives a big improvement in CPU time spent in the render thread, and removes all driver stall warnings when running with ```INTEL_DEBUG=perf```.

glennw:
Rebased.

webrender/src/renderer.rs
```diff
@@ -191,90 +190,133 @@ pub enum BlendMode {
...
+                    self.cpu_blocks.extend_from_slice(&[GpuBlockData::empty(); MAX_VERTEX_TEXTURE_WIDTH]);
+                }
+
+                // This row is dirty (needs to be updated in GPU texture).
+                self.rows[row].is_dirty = true;
+
+                // Copy the blocks from the patch array in the shadow CPU copy.
+                let block_offset = row * MAX_VERTEX_TEXTURE_WIDTH + address.u as usize;
+                let data = &mut self.cpu_blocks[block_offset..(block_offset + block_count)];
+                for i in 0..block_count {
```
kvark: nit: could use https://doc.rust-lang.org/stable/std/primitive.slice.html#method.copy_from_slice

webrender/src/renderer.rs
```diff
@@ -191,90 +190,133 @@ pub enum BlendMode {
...
-                device.blit_render_target(Some((self.current_id, 0)),
-                                          Some(blit_rect),
-                                          blit_rect);
-
-                // Free the GPU memory for that texture until we need to resize again.
-                device.deinit_texture(self.current_id);
+                // If we had to resize the texture, just mark all rows
+                // as dirty so they will be uploaded to the texture
+                // during the next flush.
+                for row in &mut self.rows {
```
kvark: nice :)

kvark:
I'm not sure why we need to operate on rows as opposed to arbitrary slices of rows, but I can see this as a complication so the current approach is a nice approximation.
@bors-servo r+

glennw:
We operate on rows for now since the driver time overhead in glTexSubImage2D was quite high, when calling it for individual blocks.

There's still some improvements here we can make - for instance, we could track the first and last dirty block in a row. Then, for mostly dirty rows, we can update several rows with one driver call. And for rows that only have a small number of changes, we could update just those blocks to avoid blitting the entire row.

PR #1359, opened 2017-06-10
Bump serde to 1.0

Eijebong:
Still have all those replaces, but we're almost here !

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1359)

PR #1392, opened 2017-06-16
Bump serde to 1.0

SimonSapin:
This is a rebase of #1359 with `bitflags` added to the tidy duplicate crates exception list, and version numbers incremented some more to catch up with `master`. Original work by @Eijebong.

Fixes #1359.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1392)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,28 +2,27 @@
 name = "wrench"
 version = "0.2.3"
 dependencies = [
- "app_units 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "app_units 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "bincode 1.0.0-alpha6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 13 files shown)


staktrace:
@SimonSapin Is there a reason this also bumps core-text to 5.0 from 4.0? It makes webrender fail linking on OS 10.7 builders (which we are still using in m-c).

staktrace:
Argh. it looks like core-text 4.0.0 uses core-graphics 0.7.0 which uses serde 0.9.15. core-text 5.0.0 on other other hand uses core-graphics 0.8.1 which doesn't need serde.

SimonSapin:
I don’t know, I carried that particular change over from #1359. I assume that it’s a cascading effect of upgrading serde being a breaking change in public API, which causes breaking changes in public APIs of dependent crates, etc. What’s the linking issue?

staktrace:
core-text 5.0.0 uses the CTFontCopyDefaultCascadeListForLanguages symbol which is only available on OS X 10.8+. So it causes a linker failure, e.g. [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=441557f7ed0014c4f10d23af986e1fcf8827b2f2)

SimonSapin:
So this update pull in unrelated changes that had landed in the meantime in core-text-rs? Sounds like that change is what should be fixed.

PR #1393, opened 2017-06-16
Add some more fuzz so that these tests pass for me.

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1393)


glennw:
I wonder how those tests are not deterministic under osmesa?

jrmuizel:
This was running with llvmpipe instead of softpipe.

PR #1395, opened 2017-06-16
Fix FLOAT texture sampling precision problems on some mobile GPUs

MortimerGoro:
Hi,

I'm working on running WebGL/WebVR demos on a Samsung Galaxy S7. I found that WR renders a white screen in all WebGL or CSS demos, even the simple ones. It's weird because it works correctly on other old Android phones such as Nexus 4 (released 5-6 years ago). I thought that it was caused by a GPU driver issue but after heavy debugging I found that texelFetch was getting corrupt values for big float values (e.g. local_clip_rect). They are stored correctly on the texture data but clamped to maximum half_float value (65504.0) when computed in the shaders.

Adding highp precision qualifier to the FLOAT sampler2D fixes the issue. Samsung Galaxy S7 has Mali-T880 GPU. The precision qualifier wasn't need on other Android mobiles I have tested (mostly Adreno GPUs).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1395)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -10,7 +10,16 @@
         #else
         precision mediump sampler2DArray;
         #endif
+
+        // Sampler default precision is lowp on mobile GPUs.
+        // This causes RGBA32F texture data to be clamped to 16 bit floats on some GPUs (e.g. Mali-T880).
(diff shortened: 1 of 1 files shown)


glennw:
Looks good, thanks!

@bors-servo r+

PR #1404, opened 2017-06-20
Fix z-buffer usage, and calculate if aligned gradients are opaque.

glennw:
I accidentally broke z-buffering some time ago - when a change was
introduced that meant opaque batches were no longer being submitted
front-to-back.

This patch fixes that, and also correctly calculates if aligned
gradients can be part of the opaque pass. This results in big
GPU performance wins with large background gradients.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1404)


glennw:
r? @kvark 

OOOPS!!

glennw:
We really need some automated performance regression tests...

Issue #1353, opened 2017-06-09
Consider splitting the WebGL related code into a separate crate.


glennw:
Instead, we could connect WebGL up to WR via the ExternalImage APIs, which support CPU-side buffers and native surface handles.

The primary benefits would be:
 * Simplify the internals of resource cache by removing the GL context management code.
 * Remove the WebGL cargo feature, since not all clients use it.

kvark:
> we could connect WebGL up to WR via the ExternalImage APIs

WebGL needs GL context, while ExternalImage does not. Are you suggesting to read back the WebGL surfaces into CPU memory in order to bring it to WR?

emilio:
I'm assuming we'd use something like rendering to a native texture, and pass that handle to WR.

MortimerGoro:
Very good idea, it will simplify a lot of things for WebGL and WebVR. 

Could you assign this issue to me? I'd love to work on this ;)

emilio:
That'd be awesome, please do!

glennw:
For some reason, GH won't allow me to assign this to you - but that's great if you can work on this!

MortimerGoro:
Fine, I'll work on this ;) 

My first thought is to move WebGL code into a Servo component. It eases development, testability and I don't think that there are use cases to use the component outside Servo. Anyway,  it will be a standalone component, so it will be easy to move if required.

It's a good opportunity to work on https://github.com/servo/webrender/issues/607 too. I haven't forgotten about that ;)

> For some reason, GH won't allow me to assign this to you - but that's great if you can work on this!

Maybe it´s because I'm not in the Servo organization?

glennw:
That sounds good to me, and working towards solving #607 would be a great benefit too!

PR #1407, opened 2017-06-20
Use EGL Sync Objects to ensure that shared WebGL textures are ready to be used

MortimerGoro:
Currently a glFlush call is issued at the start of the frame to guarantee that shared WebGL texture data is valid. But glFlush is not enough in some GPUs because it doesn't guarantee the completion of the GL commands when the shared texture is sampled. This leads to some graphic glitches on some demos or even nothing being rendered at all (GPU Mali-T880).  glFinish guarantees the completion of the commands but it may hurt performance a lot.

Sync objects are the recommended way to ensure that textures are ready in OpenGL 3.0+. They are more performant than glFinish and guarantee the completion of the GL commands.

I added dirty checking so the sync is only performed when WebGL contexts are being used. This PR might be improved if we move the glClientWaitSync to the WR render thread (before the shared texture is used). The parallelism should be better that way. I don know where is the best place to add the sync wait in WR. If you can guide me on that I'll update the PR.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1407)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -10,7 +10,7 @@ use profiler::{BackendProfileCounters, GpuCacheProfileCounters, TextureCacheProf
 use record::ApiRecordingReceiver;
 use resource_cache::ResourceCache;
 use scene::Scene;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Sender;
 use texture_cache::TextureCache;
@@ -60,6 +60,7 @@ pub struct RenderBackend {
     notifier: Arc<Mutex<Option<Box<RenderNotifier>>>>,
     webrender_context_handle: Option<GLContextHandleWrapper>,
     webgl_contexts: HashMap<WebGLContextId, GLContextWrapper>,
+    dirty_webgl_contexts: HashSet<WebGLContextId>,
     current_bound_webgl_context_id: Option<WebGLContextId>,
     recorder: Option<Box<ApiRecordingReceiver>>,
     main_thread_dispatcher: Arc<Mutex<Option<Box<RenderDispatcher>>>>,
@@ -114,6 +115,7 @@ impl RenderBackend {
             notifier,
             webrender_context_handle,
             webgl_contexts: HashMap::new(),
+            dirty_webgl_contexts: HashSet::new(),
             current_bound_webgl_context_id: None,
             recorder,
             main_thread_dispatcher,
@@ -357,6 +359,7 @@ impl RenderBackend {
                                         self.resource_cache
                                             .add_webgl_texture(id, SourceTexture::WebGL(texture_id),
                                                                real_size);
+                                        self.dirty_webgl_contexts.insert(id);
 
                                         tx.send(Ok((id, limits))).unwrap();
                                     },
@@ -381,6 +384,7 @@ impl RenderBackend {
                                     self.resource_cache
                                         .update_webgl_texture(context_id, SourceTexture::WebGL(texture_id),
                                                               real_size);
+                                    self.dirty_webgl_contexts.insert(context_id);
                                 },
                                 Err(msg) => {
                                     error!("Error resizing WebGLContext: {}", msg);
@@ -396,6 +400,7 @@ impl RenderBackend {
                                 self.current_bound_webgl_context_id = Some(context_id);
                             }
                             ctx.apply_command(command);
+                            self.dirty_webgl_contexts.insert(context_id);
                         },
 
                         ApiMsg::VRCompositorCommand(context_id, command) => {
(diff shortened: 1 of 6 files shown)


webrender/src/render_backend.rs
```diff
@@ -490,10 +497,24 @@ impl RenderBackend {
...
+        // glFlush doesn't guarantee the completion of the GL commands when the shared texture is sampled.
+        // This leads to some graphic glitches on some demos or even nothing being rendered at all (GPU Mali-T880).
+        // glFinish guarantees the completion of the commands but it may hurt performance a lot.
+        // Sync objects are the recommended way to ensure that textures are ready in OpenGL 3.0+.
+        // They are more performant than glFinish and guarantee the completion of the GL commands.
+        for (id, webgl_context) in &self.webgl_contexts {
+            if self.dirty_webgl_contexts.contains(&id) {
+                webgl_context.make_current();
+                // Call FenceSync and ClientWaitSync to ensure that textures are ready.
+                let sync = webgl_context.gl().fence_sync(gl::SYNC_GPU_COMMANDS_COMPLETE, 0);
```
kvark: we shouldn't be calling any GL functions directly from `render_backend`

webrender/src/render_backend.rs
```diff
@@ -490,10 +497,24 @@ impl RenderBackend {
...
-            webgl_context.apply_command(WebGLCommand::Flush);
-            webgl_context.unbind();
+        // glFlush is not enough in some GPUs.
+        // glFlush doesn't guarantee the completion of the GL commands when the shared texture is sampled.
+        // This leads to some graphic glitches on some demos or even nothing being rendered at all (GPU Mali-T880).
+        // glFinish guarantees the completion of the commands but it may hurt performance a lot.
+        // Sync objects are the recommended way to ensure that textures are ready in OpenGL 3.0+.
+        // They are more performant than glFinish and guarantee the completion of the GL commands.
+        for (id, webgl_context) in &self.webgl_contexts {
+            if self.dirty_webgl_contexts.contains(&id) {
```
kvark: could be `if self.dirty_webgl_context.remove(&id) {..}`

webrender/src/render_backend.rs
MortimerGoro: Do you mean that they should be called using WebGLCommand enum? (as it was done with the existing glFlush call)

kvark: I suppose, yes. Just don't assume there is GL under the hood.

glennw:
glClientWaitSync is only available on desktop GL 3.2+, I believe. Although we currently *do* require 3.2, we're meant to be dropping required desktop GL to 3.1...

MortimerGoro:
> glClientWaitSync is only available on desktop GL 3.2+, I believe. Although we currently do require 3.2, we're meant to be dropping required desktop GL to 3.1...

Yes, glClientWaitSync is available in GL 3.2+ and GL ES 3.0+. We could check if GL_ARB_sync extension is available on GL 3.1 and fallback to glFlush/glFinish if not. Do you want to do that in this PR or create an issue or handle this in the future WebGL separat crate?

glennw:
That's fine to do it as a follow up, so long as we have a valid fallback plan (which it sounds like we do).

Thanks!

glennw:
@MortimerGoro This needs a rebase.

MortimerGoro:
Done!

webrender_api/src/webgl.rs
```diff
@@ -1040,4 +1044,13 @@ impl WebGLCommand {
         gl.shader_source(shader_id.get(), &[source.as_bytes()]);
         gl.compile_shader(shader_id.get());
     }
+
+    fn fence_and_wait_sync(gl: &gl::Gl) {
+        // Call FenceSync and ClientWaitSync to ensure that textures are ready.
+        let sync = gl.fence_sync(gl::SYNC_GPU_COMMANDS_COMPLETE, 0);
+        // SYNC_FLUSH_COMMANDS_BIT is used to automatically generate a glFlush before blocking on the sync object.
+        gl.client_wait_sync(sync, gl::SYNC_FLUSH_COMMANDS_BIT, gl::TIMEOUT_IGNORED);
```
glennw: Why does this need to be a ```glClientWaitSync``` ? Wouldn't a ```glWaitSync``` be more appropriate in this case, so that the CPU isn't blocked? I would have though a ```glWaitSync``` would be enough to sync between the two command streams?

MortimerGoro: Yes, glWaitSync is enough. Just tried and it works ok.

glennw:
Looks good, thanks!

@bors-servo r+

mrobinson:
This change is causing many Servo WebGL tests to fail, so I posted a PR to partially revert this (temporarily) in order to move the WebRender update forward. Sorry!

https://github.com/servo/webrender/pull/1471

MortimerGoro:
Ok, no problem. I'm working on https://github.com/servo/webrender/issues/1353 so the WebGL code will be refactored to a new component.  I'll check all theWebGL tests after the refactor.

PR #1409, opened 2017-06-20
Use the core-text-rs feature to ensure OS X 10.7 build compatibility,…

staktrace:
… as mozilla-central requires this

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1409)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "dwrote 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "euclid 0.15.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "font-loader 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


kvark:
I disagree on the way `core-text` was patched (see https://github.com/servo/core-text-rs/pull/61#issuecomment-309853548)
Merging though, given the urgency.
@bors-servo r+

staktrace:
Not actually that urgent right now, since I cherry-picked this change into m-c directly. If you want to do it differently we can abort this and do it better.

staktrace:
Updated the PR to pull core-text 6.0.0 with default-features=false, so that it excludes the 10.8+ APIs. Note that wrench uses font-loader which still uses core-text 5.0 so now the lockfile has both 5.0.1 and 6.0.0. I can submit a PR to font-loader to bump their version if that's something we care about. Since we don't pull wrench into gecko it isn't an issue for mozilla-central.

staktrace:
Looks like the servo-tidy doesn't like having both versions of core-text in the lockfile. I'll submit a PR for font-loader.

staktrace:
@kvark updated patch seems to be passing tests now. Not sure if you need to r+ again or maybe homu/bors need a kick?

kvark:
Thanks!
@bors-servo r+

PR #1412, opened 2017-06-21
Remove the per-item complex clipping and masking from the API

mrobinson:
This is the first step to getting all masking operations into the
ClipScrollTree. We don't completely this feature, as it's still sued
for the extra clips. These will be moved into the ClipScrollTree in a
later change. This change allows us to fully eliminate the
`DisplayListBuilder::push_clip_region` API and push its functionality
into the `define_clip` API.

Most of the changes here are:
  1. Changes to the examples, wrench, and reftests to match the new API.
  2. Adapting the "split rectangle clipped by rounded rect"
     optimization to use the ClipScrollTree. This is unfortunately a
     bit more complicated now, but now includes any rounded rect clip
     instead of per-item ones.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1412)


mrobinson:
Sorry for the large patch, but most of the changes are due to the API change as well as switching two of the examples to use the example boilerplate.

webrender/src/clip_scroll_node.rs
```diff
@@ -412,6 +411,38 @@ impl ClipScrollNode {
             _ => false,
         }
     }
+
+    pub fn find_unclipped_rectangle(&self,
```
kvark: It appears to be similar to the `MashBounds::inner`. Perhaps, you could just use `clip_info.mask_cache_info.bounds.inner` instead of having a separate method?

mrobinson: It looks like `MaskBounds::inner` is using `inner_rect_safe()` while your optimization is using `inner_rect_full()`. I'm not sure what is correct here, so would it make sense to try to combine these calculations in a later patch?

kvark:
Something scary is going on:
> REFTEST reftests/aa/rounded-rects.yaml == reftests/aa/rounded-rects-ref.png
> thread 'RenderBackend' has overflowed its stack
> fatal runtime error: stack overflow

mrobinson:
@kvark This was an issue with the rebase on top of the rust upgrade. It should be fixed now. Sorry for the churn!

kvark:
@mrobinson thanks!
@bors-servo r+

staktrace:
@mrobinson FYI the change to only treat non-zero-origin or clip-smaller-than-content clips as scroll frames resulted in a bunch of async-scrolling reftest regressions in gecko ([link](https://bugzilla.mozilla.org/show_bug.cgi?id=1374730#c13)). Presumably once the API allows us to more explicitly create scroll frames vs non-scrolling clips and we start using that this will not be a problem any more. The patch I applied to verify this can be found [here](https://hg.mozilla.org/try/rev/988ce8cb2abc98051136d0ff697e98dcb85c5bdb).

glennw:
@mrobinson We should discuss the per-item clip stuff in SF - I have a few questions / concerns about possible performance impacts of not supporting ```extra_clips``` on items.

staktrace:
> Presumably once the API allows us to more explicitly create scroll frames vs non-scrolling clips and we start using that this will not be a problem any more.

This is happening in #1428.

PR #1415, opened 2017-06-21
Mark types as repr(C)

rlhunt:
This commit marks types that can be repr(C) and are used in the Gecko
FFI to be repr(C). This also removes the repr(C) from PropertyBindingKey
as it has a type that isn't repr(C), and I don't think it is used in Gecko.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1415)

### DIFF
--- a/webrender_traits/src/api.rs
+++ b/webrender_traits/src/api.rs
@@ -507,7 +507,6 @@ impl PropertyBindingId {
 
 /// A unique key that is used for connecting animated property
 /// values to bindings in the display list.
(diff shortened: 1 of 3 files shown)


rlhunt:
Travis CI fails with
```
  = note: /usr/bin/ld: cannot find -lGL
collect2: error: ld returned 1 exit status
```
Which seems unrelated to this PR.

kvark:
@rlhunt CI should be fixed by #1417 which is about to land

PR #1417, opened 2017-06-22
Rust 1.17.0

rillian:
Gecko [bumped the minimum](https://bugzilla.mozilla.org/show_bug.cgi?id=1374807#c12), so do the same here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1417)


kvark:
Travis fails to `-lGL` for some reason...
AppVeyor fails to compile `src/perf.rs`, the errors are related to `'static` elision

rillian:
Thanks, I'll look at the test failures tomorrow.

rillian:
The `'static` elision failure was just a typo: s/str/&str/. I've pushed a fix.

rillian:
For the GL issue, I was able to fix it by installing `libGL` instead of `libGLESv2`, since that's what the build was expecting to link to. Maybe it's bitrot. It's also possible that we should be testing against `libGLES` and either the library name or sys crate's expectations changed under us, in which case this is the wrong fix, so please review carefully.

If anyone else wants to debug, I was able to reproduce on `master` in an `ubuntu:trusty` docker container:
```
apt-get update && apt-get -y upgrade
apt-get -y install git build-essential curl
apt-get -y install cmake pkg-config libfreetype6-dev libx11-dev libexpat1-dev
apt-get -y install libgles2-mesa-dev llvm-3.8-dev libedit-dev
adduser --disabled-password build
su - build
curl https://sh.rustup.rs -sSf | sh
source ~/.cargo/env
git clone https://github.com/servo/webrender
cd webrender/webrender && cargo test --all --verbose
```

kvark:
Thanks for fixing the CI!
@bors-servo r+

PR #1420, opened 2017-06-22
res: fix indentation in cs_text_run vertex shader.

emilio:
Kind of an unimpressive patch, but I saw this when investigating https://github.com/servo/servo/issues/17464 and thought it may be worth fixing.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1420)


kvark:
Thanks!
@bors-servo r+

emilio:
@bors-servo retry

 * wat

PR #1421, opened 2017-06-22
wrench: use assert_eq! so the assertion message also prints the failed test count.

emilio:
Yet another unimpressive patch.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1421)


glennw:
Needs a rebase :)

emilio:
This is actually superseded by the PR that bitrotted it, so closing instead :P

PR #1422, opened 2017-06-22
Use euclid 0.15.1

rlhunt:
TypedRect is marked repr(C) in euclid:0.15.1, which helps with the FFI.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1422)


kvark:
weird

@bors-servo retry

PR #1423, opened 2017-06-22
frame_builder: Don't use subpixel AA for non-opaque text.

emilio:
This fixes https://github.com/servo/servo/issues/17464.

Not 100% sure it's the correct fix, but it seemed reasonable given we disable it
for non-opaque stacking contexts.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1423)


glennw:
@emilio Yup, it's certainly an improvement over what we have - we can look at a "proper" solution later. Thanks!

@bors-servo r+

PR #1426, opened 2017-06-23
Robust TransformRect

kvark:
r? @glennw @nical 

The new inverse footprint calculation (sometimes) produces extremely large values (correctly), which mess up the screen bounding box calculations due to `i32` casts in `TransformRect`. This PR makes it robust, resulting in better preserve-3d handling:

rubik

URL: https://ondras.zarovi.cz/demos/rubik/

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1426)


glennw:
Nice screenshot!

@bors-servo r+

PR #1428, opened 2017-06-24
Improve the clipping and scrolling API

mrobinson:
This branch includes two improvements to the clipping and scrolling API:

1. Create `DisplayListBuilder::define_scroll_frame` API which avoids having to guess when a Clip item should create a ScrollFrame and when it shouldn't. Additionally, this removes a lot of weirdness around whether a ScrollFrame's clip starts at the origin of the ScrollFrame or not. It doesn't matter any longer, because the Clip created for the ScrollFrame is entirely independent.
2. Make all clipping coordinates relative to the StackingContext. This was whirlpool of weirdness in the API that only caused confusion. We now hide this detail, making clip coordinates the same as any other coordinates.

There are a few other small cleanups included with this, such as remove the `push_clip_node` and `pop_clip_node` APIs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1428)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,7 +21,7 @@ dependencies = [
  "serde_json 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-glutin 0.11.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.46.0",
+ "webrender 0.47.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
@@ -1022,7 +1022,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "webrender"
-version = "0.46.0"
+version = "0.47.0"
 dependencies = [
  "angle 0.1.2 (git+https://github.com/servo/angle?branch=servo)",
  "app_units 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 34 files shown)


glennw:
@mrobinson Do you have Servo branch with API updates for this and the previous PR? Would be good to be able to do some testing with Servo and land a WR update once this is reviewed and merged.

mrobinson:
I've pushed a new commit to this branch and the Servo part of this PR is available here: https://github.com/mrobinson/servo/tree/webrender-update.

staktrace:
@glennw or @kvark , can we get this one reviewed and landed? It is needed in order to do a WR update in gecko. (Well, we either need this or #1459, but this is the correct solution and #1459 is just a temp workaround).

mrobinson:
I just noticed the reftest failure here. I'm looking at this now.

mrobinson:
Okay. The latest version of the PR fixes the typo in the ref test, but one of the jobs seemed to fail due to an error on the bot. I'll rerun that particular configuration.

glennw:
Reviewed 16 of 27 files at r1, 15 of 15 files at r2, 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1428)*

glennw:
Let's get this merged, and try to get both Servo and Gecko updated to the API changes this week :)

@bors-servo r+

glennw:
@bors-servo retry
 * Looks like a nightly-specific compile error in a dependency? Let's just confirm it's consistent...

glennw:
This appears to be broken on nightly due to this change https://github.com/rust-lang/rust/issues/34537 which is causing a compile error in core-text.

A change has been made in core-text to fix this https://github.com/servo/core-text-rs/commit/6847e69397f0635746d709f8f23a561191e32cfa but it hasn't been version bumped / published yet.

If it requires a major version bump in core-text, that could take quite a while to get all deps updated. Perhaps we could disable building WR on nightly temporarily, since we really need to get this PR (and a few others) landed soon.

Thoughts @nox @jdm @kvark ?

glennw:
https://github.com/servo/webrender/pull/1462 has landed now, which should fix the CI error here.

@bors-servo retry

PR #1429, opened 2017-06-26
WIP: Text parts

Gankro:
This is only WIP because this needs backend work that @kvark or @glennw should probably handle (significant design work needed).

Notes:

* Only blob text decorations are supported. That's fine, in the sense that implementation for those is "future work" that we can do in future weeks. Doesn't need to be done to land this.

* Wrench doesn't currently save Opaque TextDecorations because I'm unclear on how that should work.

* The backend hasn't been implemented. Currently the data just gets stuffed in the TextRunPrimitive struct. This also makes the PR currently a regression, as it just makes shadows not work. (`reftests/text/shadow-single.yaml`)

* To make the backend work easier, I've left in all references to blur_radius in there. That way y'all can remove it and chase down the compilation errors to find the places that need to be updated. A dummy variable has been added in frame_builder that is incorrectly mutable, so a warning will show up to tell you to remove that too.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1429)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -292,11 +292,12 @@ fn body(api: &RenderApi,
         builder.push_text(text_bounds,
                           text_bounds,
                           &glyphs,
+                          &[],
                           font_key,
                           ColorF::new(1.0, 1.0, 0.0, 1.0),
                           Au::from_px(32),
-                          0.0,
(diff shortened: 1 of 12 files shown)


Gankro:
Closing in favour of #1454

PR #1434, opened 2017-06-29
Tidy up the GPU tags for blurs and alpha batch targets.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1434)


glennw:
r? @kvark 

Most of the diff here is just indentation changes.

PR #1435, opened 2017-06-29
Add a way to remove all resources from the resource cache that have a certain namespace

nical:
This is a tad ugly but it makes it a lot easier in firefox when a content-process crashes, to clean up every resources that were associated to it. The alternative is to maintain a map of all of the image and font keys in use in gecko-land.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1435)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -18,7 +18,7 @@ use std::mem;
 use std::sync::Arc;
 use texture_cache::{TextureCache, TextureCacheItemId};
 use webrender_traits::{Epoch, FontKey, FontTemplate, GlyphKey, ImageKey, ImageRendering};
-use webrender_traits::{FontRenderMode, ImageData, GlyphDimensions, WebGLContextId};
+use webrender_traits::{FontRenderMode, ImageData, GlyphDimensions, WebGLContextId, IdNamespace};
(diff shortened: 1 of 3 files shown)


nical:
cc @sotaroikeda @glennw

webrender_traits/src/api.rs
```diff
@@ -62,6 +62,7 @@ pub enum ApiMsg {
     /// to forward gecko-specific messages to the render thread preserving the ordering
     /// within the other messages.
     ExternalEvent(ExternalEvent),
+    RemoveAllResourcesWithNamespace(IdNamespace),
```
kvark: Perhaps, just `ClearNamespace`?

webrender/src/resource_cache.rs
```diff
@@ -764,6 +764,11 @@ impl ResourceCache {
         debug_assert_eq!(self.state, State::QueryResources);
         self.state = State::Idle;
     }
+
+    pub fn remove_all_resources_with_namespace(&mut self, namespace: IdNamespace) {
+        self.resources.image_templates.images.retain(|key: &ImageKey, _|{ key.0 != namespace.0 });
```
kvark: I'm about to make a PR to make namespaces more explicit

kvark:
Replaced by #1445
Otherwise, would need a rebase plus fix to compile on 1.17

PR #1445, opened 2017-06-30
Clear resources in a namespace

kvark:
Builds upon #1435

cc @nical 
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1445)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -18,7 +18,7 @@ use std::mem;
 use std::sync::Arc;
 use texture_cache::{TextureCache, TextureCacheItemId};
 use api::{Epoch, FontKey, FontTemplate, GlyphKey, ImageKey, ImageRendering};
-use api::{FontRenderMode, ImageData, GlyphDimensions, WebGLContextId};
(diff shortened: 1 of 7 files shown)


webrender/src/resource_cache.rs
```diff
@@ -764,6 +764,24 @@ impl ResourceCache {
         debug_assert_eq!(self.state, State::QueryResources);
         self.state = State::Idle;
     }
+
+    pub fn clear_namespace(&mut self, namespace: IdNamespace) {
+        //TODO: use `retain` when we are on Rust-1.18
```
glennw: We'll want to also remove the items from the texture cache, if they exist in there, I guess?

nical: These would eventually expire on their own, right? But yeah sounds like it would be a good thing to do as well, and also remove items from the texture cache when removing image and font templates.

kvark:
Oops, reopening

glennw:
@kvark I think this just needs an update to remove items from the texture cache, if that's feasible.

kvark:
@glennw please take another look

glennw:
@kvark Looks good! r=me, do you need to remove the text-shadow commit that is in here? Or will git handle that cleanly?

PR #1446, opened 2017-07-01
Splitter stack

kvark:
Fixes #1444

The previous code assumed that preserve-3d forms sub-trees in the stacking context tree. This is not correct, since we may encounter 2D->3D->2D->3D chains, and to handle them we need multiple BSP splitters stored in a stack.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1446)

PR #1448, opened 2017-07-01
Removal of SC's local_bounds

### LINKED ISSUE #1283: Remove use of stacking context local_bounds from make_polygon
It would be good to remove ```local_bounds``` from the StackingContext structure (and the associated public APIs).

It doesn't really have any meaning, now that we collect the bounding rect from the clip-scroll groups that make up a stacking context.

There's one piece of code that currently reads the ```local_bounds``` field in the ```make_polygon``` function.

Could we modify how that function works to not require the local bounds?

kvark:
Fixes #1283
Includes #1446

The bounds are now specifically computed for the children of isolated stacking contexts that need to be baked in (thus, excluding isolated children contexts).

Note: the `nested` test used `local_bounds` previously and considered incorrect, thus it is fixed now.
Note: breaking change

r? @mrobinson

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1448)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1180,8 +1180,8 @@ impl FrameBuilder {
         //  - ones with `ContextIsolation::Items`, for their actual items to be backed
         //  - immediate children of `ContextIsolation::Items`
         let mut preserve_3d_map: HashMap<StackingContextIndex, RenderTask> = HashMap::new();
-        // The plane splitter, using a simple BSP tree.
-        let mut splitter = BspSplitter::new();
+        // The plane splitter stack, using a simple BSP tree.
+        let mut splitter_stack = Vec::new();
 
         debug!("build_render_task()");
 
@@ -1210,6 +1210,9 @@ impl FrameBuilder {
 
                     if parent_isolation == Some(ContextIsolation::Items) ||
                        stacking_context.isolation == ContextIsolation::Items {
+                        if parent_isolation != Some(ContextIsolation::Items) {
+                            splitter_stack.push(BspSplitter::new());
+                        }
                         alpha_task_stack.push(current_task);
                         current_task = RenderTask::new_dynamic_alpha_batch(next_task_index, stacking_context_rect);
                         next_task_index.0 += 1;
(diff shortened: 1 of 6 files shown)


webrender/src/frame_builder.rs
```diff
@@ -1651,19 +1653,28 @@ impl<'a> LayerRectCalculationAndCullingPass<'a> {
...
             stacking_context.screen_bounds = stacking_context.screen_bounds
                                                              .intersection(self.screen_rect)
                                                              .unwrap_or(DeviceIntRect::zero());
-            (stacking_context.screen_bounds.clone(), stacking_context.is_visible)
+            (stacking_context.screen_bounds.clone(),
+             stacking_context.is_visible,
+             stacking_context.isolation == ContextIsolation::Items,
+             stacking_context.reference_frame_id,
+             stacking_context.isolated_items_bounds.translate(&stacking_context.reference_frame_offset),
+            )
```
mrobinson: Instead of passing out all of this new data, I wonder if it make sense to simply pass a new boolean called should_add_children_local_bounds?

webrender/src/frame_builder.rs
```diff
@@ -1783,16 +1795,18 @@ impl<'a> LayerRectCalculationAndCullingPass<'a> {
...
-                                                                          &packed_layer.local_clip_rect,
-                                                                          self.device_pixel_ratio) {
-                Some(rect) => rect,
+            let (prim_local_rect, prim_screen_rect) = match prim_store
+                .build_bounding_rect(prim_index,
+                                     &clip_bounds,
+                                     &packed_layer.transform,
+                                     &packed_layer.local_clip_rect,
+                                     self.device_pixel_ratio)
+            {
```
mrobinson: This curly brace should be snuggled, I think.

mrobinson:
Looks good to me, just a couple minor nits for now.

webrender/src/frame_builder.rs
kvark: Unfortunately, all that data is needed, and we can't have a single `should_add_children_local_bounds` because we don't have the parent information at this point

webrender/src/frame_builder.rs
kvark: fixed now

webrender/src/frame_builder.rs
mrobinson: Ah, sorry! I see now.

kvark:
Thanks!
@bors-servo r=mrobinson

Issue #1418, opened 2017-06-22
plan: text decoration blob support


Gankro:
So today if gecko has text decorations that webrender can't handle we give up completely and produce a blob image for the text. I'd like to improve this situation so that webrender handles the things it can natively, but invokes gecko via blob images for the parts that it can't.

At the high level I want webrender to receive something like:

```
text
  - glyphs
  - font/style details
  - shadows [(color, offset, blur, ...), ...]
  - underline (kindICanHandle | blob)
  - overline (kindICanHandle | blob)
  - line-through (kindICanHandle | blob)
```

And for its rendering strategy to be something like:

```
drawText:
  for shadow in shadows {
    drawTextWithEffect(shadow)
  }
  drawTextWithEffect(None)

drawTextWithEffect(effect):
  drawUnderline(effect)
  drawOverline(effect)
  drawText(effect)
  drawLineThrough(effect) 
```

Potentially the drawTextWithEffect can actually be cached like:

```
drawText:
  let text = drawTextToTemp() // like drawTextWithEffect(None), but to temporary surface
  for shadow in shadows {
    compositeAndApplyEffect(text, shadow)
  }
  compositeAndApplyEffect(text, None)
```

But I have no idea if that approach is actually practical.

Note: I am currently omitting [text-emphasis marks](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis) and selection from this description. I think selection can be handled by the client (drawing a box and tweaking colors), but I'm less certain about emphasis (they strictly should be drawn between text and linethrough -- possibly just needs to be a second list of glyphs).

# Summary of Benefits

* reduced bandwidth: currently clients are expected to just send the text again for each shadow with the effects applied. Under this system they would just send a list of shadow effects. (There's some sites that apply like 4 shadows to all their text to get a certain effect)
* a path for migrating implementation from gecko to webrender incrementally (start all decorations as blobs, then teach it basic lines, then dashed lines, then wavy lines, etc.)
* more text decoration/shadow logic can be shared between gecko and servo via webrender (e.g. Servo devs will do my job for me :P)

Gankro:
Clarification: drawTextWithEffect probably only forwards a subset of the effects. i.e. blurring of the shadow should be a finalization step, and not something each piece of the text does itself.

Gankro:
[...]

Issue #1443, opened 2017-06-29
Expose ReferenceFrames to the API


mrobinson:
Based on feedback from #1304, it might make sense to expose ReferenceFrame creation to the API. A prerequisite of this is using ReferenceFrame coordinates everywhere internally.

mrobinson:
WebRender is using reference frame coordinates everywhere internally, so I think it might be a good time to do this. It is also necessary in order to provide both a display list and hit testing area list sharing the ClipScrollTree.

mrobinson:
This was fixed by #2756.

PR #1454, opened 2017-07-05
WIP: Replace Text Blur with PushTextShadow and PopTextShadow DisplayItems

Gankro:
**WIP because it doesn't implement the backend part of this -- someone else needs to do that.** As before I've made a dummy variable that will create a warning so that the code compiles but points you to what code needs to be audited. I've also added some tests that capture some interesting cases (I suggest actually rendering them to make sure they look the way you want).
 
This is the new version of  #1418, which instead of attaching shadows/decorations to text, uses a stateful PushTextShadow and PopTextShadow.

The semantics are that everything between a PushTextShadow and PopTextShadow should be rendered to a shadow with the given properties. (Should only expect glyphs, text-decorations, and blob-image versions of decorations here) 

Why is this necessary?

Well, basically the CSS spec requires us to put every shadow in a given StackingContext on the same "layer", behind all the *actual* content. However we receive text at too low of a granularity to do this if the shadows are attached to the text. (we get roughly one TextItem per line, but more if fonts or styles change mid-line -- this can happen without any extra html/css elements due to font fallback on kanji or emoji)

The Push/Pop model gives us a simple way to say "hey all this content? yeah put a shadow on all of this". I expect the backend implementation model to look something like:

> Before rendering any content in a given stacking context, for each PopTextShadow, render the enclosed elements with the given offset, blur, and color.

The PopTextShadow approach is also beneficial because it encourages us to put as much content as possible in the "same" shadow, preventing ugly artifacts from overlapping two parts of the "same" shadow. Technically the CSS spec gives us leeway to give each glyph its own shadow rendering, but we should try to avoid this as much as possible.

This approach also avoids us having to actually add a notion of text-decorations for the time-being. Although I think webrender should eventually have them, for now they can just be implemented as blob-images from firefox or boxes from servo. When we do add them, we also won't have to bloat text -- the client can just add separate top-level items for them.

r? @glennw (still up for the backend impl work?)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1454)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -290,7 +290,6 @@ fn body(api: &RenderApi,
                           font_key,
                           ColorF::new(1.0, 1.0, 0.0, 1.0),
                           Au::from_px(32),
-                          0.0,
                           None);
     }
 
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -583,7 +583,6 @@ impl Frame {
                                          item.local_clip(),
                                          text_info.font_key,
                                          text_info.size,
-                                         text_info.blur_radius,
                                          &text_info.color,
                                          item.glyphs(),
                                          item.display_list().get(item.glyphs()).count(),
@@ -718,6 +717,12 @@ impl Frame {
 
             SpecificDisplayItem::PopStackingContext =>
                 unreachable!("Should have returned in parent method."),
+            SpecificDisplayItem::PushTextShadow(shadow) => {
+                context.builder.push_text_shadow(shadow);
+            }
+            SpecificDisplayItem::PopTextShadow => {
+                context.builder.pop_text_shadow();
+            }
         }
         None
     }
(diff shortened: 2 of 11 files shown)


glennw:
I'm up for the implementation work, but I've been thinking about this and wondering if a slightly different approach might be better.

If we go ahead and expose reference frames in the display list (https://github.com/servo/webrender/issues/1443) which I think makes sense, then having the concept of a stacking context directly in the display list is basically a superset of the PushTextShadow item, I think.

Effectively, I'm suggesting that instead of PushStackingContext and PushTextShadow, we replace these by a "PushLayerGroup" (for lack of a better name right now) that specifies a list of CSS filters / mix-blend-modes / shadows, and that this one display item is then used by the DL builder for both text shadows and stacking contexts that have filters / blends on them.

Thoughts @kvark @mrobinson @Gankro @jrmuizel ?

glennw:
Although, just thinking that through a bit more, I guess we still need the concept of a stacking context specifically, in order to pass things like ```preserve-3d```. Hmmm, I will think about this a bit more today.

Gankro:
I have no particular preference for if this is a dedicated thing or part of a broader notion of "groups" -- other than a vague desire to avoid incredibly bloated "swiss-army-knife" display items with a dozen optional properties

kvark:
@glennw yep, sounds like `Push/PopStackingContext` is here to stay. I think `Push/PopTextShadow` is fine as long as there are no stacking contexts inside those shadow sections.

glennw:
I'm working on some other text-related issues. Once those are fixed, I'll start work on the implementation side of this (hopefully this week).

glennw:
PR with the implementation (and the commits from here included) is https://github.com/servo/webrender/pull/1469.

PR #1441, opened 2017-06-29
Fixed snap rectangle for the glyph runs

kvark:
r? @glennw 
cc @mrobinson

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1441)

Issue #1453, opened 2017-07-05
Transformed text should probably be rasterized transformed


jrmuizel:
Currently it looks like we just transform the rasterized bitmap. Browsers do a transformed rasterization. An example of this is here: https://jrmuizel.github.io/tests/transformed-text.html

Servo gives a blurry mess, Gecko and Chrome give proper subpixel AA

kvark:
Related to #1451

nical:
Also related to #1542

glennw:
@lsalzman Can this be closed now?

lsalzman:
Yes, this is fixed.

PR #1455, opened 2017-07-06
Glyph snapping - revert of #1441

### LINKED ISSUE #1451: Text glyph pixel snapping
Apparently, the fix in #1441 doesn't cut it (see https://github.com/servo/webrender/pull/1441#issuecomment-312922649).

kvark:
Fixes #1451
@staktrace I didn't see much of a difference locally. Would you be able to verify if this works?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1455)

### DIFF
--- a/webrender/res/ps_text_run.vs.glsl
+++ b/webrender/res/ps_text_run.vs.glsl
@@ -23,7 +23,7 @@ void main(void) {
                                                     prim.z,
                                                     prim.layer,
                                                     prim.task,
-                                                    prim.local_rect);
+                                                    local_rect);
     vLocalPos = vi.local_pos;
     vec2 f = (vi.local_pos.xy / vi.local_pos.z - local_rect.p0) / local_rect.size;
 #else
@@ -32,7 +32,7 @@ void main(void) {
                                  prim.z,
                                  prim.layer,
                                  prim.task,
-                                 prim.local_rect);
+                                 local_rect);
     vec2 f = (vi.local_pos - local_rect.p0) / local_rect.size;
 #endif
 


glennw:
We definitely need to snap text glyphs. Otherwise fractional components in the text position result in blurry text.

jrmuizel:
Do you have an example page that needs snapping?

kvark:
@glennw Here is how I see it, sorry if some parts are obvious :)

Glyph rasterization is hard, and we don't do it. Instead, we ask a platform-specific library to do the actual rasterization. That library works within an assumption that it renders on screen, and it takes into account all the details on how glyphs are placed, including sub-pixel offsets. If we do any sort of scaling, rotation, or translation by a fractional offset, we'll ruin the precision of the rasterized glyph.
Thus, we need to map the rasterized glyphs 1:1 to screen. This means that any transformations (#1453) should be applied prior to rasterization. Snapping, if any, should only be applied on the text run as a whole, and not ever try to scale it.

There is a simple way to ensure that snapping doesn't scale - just setting the snap rectangle size to 0. This PR has been updated to do it.

staktrace:
@kvark Sadly it doesn't look like this PR fixes the problem. The reftest looks about the same as without the patch. [Try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=df1562689841ef9aed49160bf132205cb7515b20) not fully done yet but R7 is failing still: [Reftest analyzer link](https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/Oms3ZxvMRgShRQlUWP2E2A/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1)

kvark:
Thanks @staktrace ! I assume the test is done with the latest commit https://github.com/servo/webrender/pull/1455/commits/63ee6f47b40f60fe846ea77df9f1d5c2476da43b

staktrace:
@kvark yeah: https://hg.mozilla.org/try/rev/b79323ee0c9b765f79a075dbcdde70b385d8dde4

kvark:
I just noticed why this change doesn't produce desired effect:
```rust
    // Ensure that the snap rect is at *least* one device pixel in size.
    // TODO(gw): It's not clear to me that this is "correct". Specifically,
    //           how should it interact with sub-pixel snap rects when there
    //           is a layer transform with scale present? But it does fix
    //           the test cases we have in Servo that are failing without it
    //           and seem better than not having this at all.
    snap_rect.size = max(snap_rect.size, vec2(1.0 / uDevicePixelRatio));
```
@glennw what tests exactly are affected? As you mention in the comments, it doesn't make sense to adjust the size in local space anyway.

glennw:
I think that the simplest option for now is to revert that previous change which snaps to the primitive rect, and return it to how it was (snapping to the glyph rect). Then we can work out the "proper" solution after that, which will allow us to update WR in Gecko / Servo without reftest regressions?

staktrace:
Verified that the revert fixed the reftest failure - [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=dc7f89373e6c73dabdff62534bb334f75af08124). Thanks!

PR #1456, opened 2017-07-06
Disable snapping of transformed primitives

### LINKED ISSUE #1442: Consider whether / when we should do snapping on transformed primitives.
There *may* be some WPT reftests that rely on this?

It does produce visible artifacts when animating a rotated primitive - it seems we want to disable snapping on transformed primitives at least in this case.

kvark:
Fixes  #1442
(haven't run Servo tests on that yet)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1456)


glennw:
This looks good to me - I'd like to run the Servo tests first to make sure though :)

kvark:
Oops, was cleaning up the branches too aggressively :)

PR #1457, opened 2017-07-07
Use light hinting for Linux glyphs.

glennw:
This is a better default and matches glyphs exactly as Gecko draws them by default.

In the future, we'll want to make the hinting level configurable.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1457)


glennw:
Added a second commit that uses the exact same code as Gecko to convert pt size to FT char size. These two changes, plus the same changes applied in the Servo layout code (which I'll do as part of the WR update PR) make the text rendering on Linux *much* better, and very close to Gecko (I believe there is one remaining issue related to hidpi scaling).

kvark:
Looks reasonable.
@bors-servo r+

staktrace:
This PR caused a gecko reftest to fail - see https://bugzilla.mozilla.org/show_bug.cgi?id=1379604#c2 for more details.

PR #1459, opened 2017-07-07
Treat all clips as scroll frames (partial revert of #1412)

staktrace:
PR #1412 introduced a small semantic change mixed in with the removal of
per-item complex clipping and masking. The semantic change resulted in
numerous Gecko reftest failures as many scroll frames ended up being
treated as clips. Eventually WR will have an explicit API to allow
callers to distinguish scroll frames from clips, but until that is in
place we should maintain the old behaviour of just treating all clips as
potential scroll frames.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1459)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -345,8 +345,10 @@ impl Frame {
                         item: &ClipDisplayItem,
                         content_rect: &LayerRect,
                         clip: &ClipRegion) {
-        let is_scroll_frame =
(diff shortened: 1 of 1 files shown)


mrobinson:
FWIW, the API for explicit scroll frame creation is waiting at https://github.com/servo/webrender/pull/1428.

staktrace:
Ah, great!

glennw:
Closing in favor of #1428 which is merging now - please re-open if needed.

PR #1460, opened 2017-07-07
Remove incorrect warning.

### LINKED ISSUE #1424: Spurious "webrender::clip_scroll_node: Tried to scroll a non-scroll node" warnings
See [bug 1375843](https://bugzilla.mozilla.org/show_bug.cgi?id=1375843) and related bugs for backstory. It looks like the clip_scroll_tree sometimes ends up with `DynamicallyAddedNode` clip ids in its `self.nodes` map, which means when it calls `finalize` on those nodes it prints out a warning. This is happening with webrender in gecko in automation and causing job failures because the warning spews so much that it slows down jobs and triggers timeouts.

I talked with @mrobinson on IRC about this, and since gecko never tries to set a scroll offset on a `DynamicallyAddedNode` clip id we concluded this needs to be fixed internally to webrender. Filing this issue to track the problem.

staktrace:
Fixes #1424

It should be fine to call finalize() on a non-scroll node. It would not
be fine to call set_scroll_origin on a non-scroll node, but that
function emits a warning for that scenario anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1460)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -197,8 +197,11 @@ impl ClipScrollNode {
 
     pub fn finalize(&mut self, new_scrolling: &ScrollingState) {
         match self.node_type {
-            NodeType::ReferenceFrame(_) | NodeType::Clip(_) =>
-                warn!("Tried to scroll a non-scroll node."),
+            NodeType::ReferenceFrame(_) | NodeType::Clip(_) => {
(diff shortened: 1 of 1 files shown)


kvark:
I think we need to ensure that the scrolling passed into `finalize` is zero for non-scroll nodes

staktrace:
@kvark That sounds reasonable. I updated the patch.

PR #1465, opened 2017-07-10
Refined text glyph snapping

kvark:
This PR truncates the sub-pixel offset of the glyphs coming to GPU in order to fix the snapping. With  `FontRenderMode::Subpixel`, that offset is already taken into account when the glyph is rasterized. With the other rendering modes, I'm not completely sure if it's fine to truncate the fractional part, but I also don't see how the current snapping code is better anyway.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1465)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -536,10 +536,24 @@ impl TextRunPrimitiveCpu {
             // GPU block.
             let first_glyph = glyph_chunk.first().unwrap();
             let second_glyph = glyph_chunk.last().unwrap();
-            request.push([first_glyph.point.x,
-                          first_glyph.point.y,
-                          second_glyph.point.x,
-                          second_glyph.point.y]);
+            let data = match self.render_mode {
+                FontRenderMode::Mono |
+                FontRenderMode::Alpha => [
+                    first_glyph.point.x,
+                    first_glyph.point.y,
+                    second_glyph.point.x,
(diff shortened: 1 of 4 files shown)


kvark:
Fixed to pass test-wpt now. The fix was to keep the old snapping behavior for the non-subpixel text.
This PR should be ready for the review now.

webrender/src/prim_store.rs
```diff
@@ -504,10 +504,24 @@ impl ToGpuBlocks for TextRunPrimitiveCpu {
...
+                FontRenderMode::Mono |
+                FontRenderMode::Alpha => [
+                    first_glyph.point.x,
+                    first_glyph.point.y,
+                    second_glyph.point.x,
+                    second_glyph.point.y,
+                ],
+                // The sub-pixel offset has already been taken into account
+                // by the glyph rasterizer, thus the truncating here.
+                FontRenderMode::Subpixel => [
```
glennw: Instead of a conditional here, can we do this in the prepare_prim_for_render when the glyph_instances array is created? That would mean only doing it once, not each time it needs to get added to the cache.

glennw:
@kvark Looks good. We have zero coverage in Servo tests for alpha / subpixel text - we currently only run the tests with text AA disabled. We should add some test coverage to wrench for the various render modes and snap configurations.

@bors-servo r+

PR #1466, opened 2017-07-11
Only convert scroll id to nested if pipeline id is match.

mephisto41:
If nested display list is used and this nested display list contains
iframe. In current implementation, we'll convert the root scroll node
inside the iframe to the nested display list's root scroll node. Thus,
the primitive position inside the iframe is wrong. So I add a simple
check to avoid this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1466)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -66,6 +66,10 @@ impl NestedDisplayListInfo {
     }
 
     fn convert_scroll_id_to_nested(&self, id: &ClipId) -> ClipId {
+        if id.pipeline_id() != self.scroll_node_id.pipeline_id() {
+            return *id;
+        }
+
         if id.is_root_scroll_node() {
             self.scroll_node_id
         } else {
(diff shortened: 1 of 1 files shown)


webrender/src/frame.rs
```diff
@@ -67,15 +67,23 @@ impl NestedDisplayListInfo {
...
     fn convert_scroll_id_to_nested(&self, id: &ClipId) -> ClipId {
         if id.is_root_scroll_node() {
-            self.scroll_node_id
+            if id.pipeline_id() == self.scroll_node_id.pipeline_id() {
+                self.scroll_node_id
+            } else {
+                *id
+            }
         } else {
             self.convert_id_to_nested(id)
```
mrobinson: Why don't we skip the conversion for non-root scroll nodes as well? If that's the right thing to do, we can just do an early return from this method.

mephisto41: I'm not sure we need skip for non-root scroll nodes because my bug only appear in root scroll node. Do you think I should skip non-root scroll nodes as well?

mrobinson: Are you in a situation where pipelines are unique between all display lists?

mephisto41: In the gecko, yes, I think all pipelines are unique.

mrobinson: I think that in this case it makes sense to treat ClipIds from other pipelines the same way, no matter if they are the root id or not. If you can guarantee that ids from this other pipeline will be unique across the entire display list, then I think we should just do an early return in these methods.

PR #1467, opened 2017-07-11
Derive Debug in some structs/enums for easier printing

staktrace:
While trying to investigate bug 1378883 I found it useful to print out these data structures. Figured it might be worthwhile to check this in.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1467)


kvark:
Thanks!
@bors-servo r+

PR #1469, opened 2017-07-12
Implement support for multiple text runs per shadow.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1469)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -6,8 +6,8 @@ use api::{BorderDetails, BorderDisplayItem, BoxShadowClipMode, ClipAndScrollInfo
 use api::{DeviceIntPoint, DeviceIntRect, DeviceIntSize, DeviceUintRect, DeviceUintSize};
 use api::{ExtendMode, FontKey, FontRenderMode, GlyphInstance, GlyphOptions, GradientStop};
 use api::{ImageKey, ImageRendering, ItemRange, LayerPoint, LayerRect, LayerSize};
-use api::{LayerToScrollTransform, LayerVector2D, LocalClip, PipelineId, RepeatMode, TileOffset};
-use api::{TransformStyle, WebGLContextId, WorldPixel, YuvColorSpace, YuvData};
+use api::{LayerToScrollTransform, LayerVector2D, LocalClip, PipelineId, RepeatMode, TextShadow};
+use api::{TileOffset, TransformStyle, WebGLContextId, WorldPixel, YuvColorSpace, YuvData};
 use app_units::Au;
 use frame::FrameId;
 use gpu_cache::GpuCache;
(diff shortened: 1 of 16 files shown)


Gankro:
r+ on the api/wrench changes

webrender/src/frame_builder.rs
```diff
@@ -106,6 +106,37 @@ pub struct FrameBuilderConfig {
...
+}
+
+impl PendingTextShadow {
+    fn new(shadow: TextShadow,
+           clip_and_scroll: ClipAndScrollInfo,
+           local_clip: &LocalClip) -> PendingTextShadow {
+        PendingTextShadow {
+            shadow: shadow,
+            text_primitives: Vec::new(),
+            clip_and_scroll: clip_and_scroll,
```
kvark: nit: could be `clip_and_scroll,`

webrender/src/frame_builder.rs
```diff
@@ -392,6 +428,37 @@ impl FrameBuilder {
...
+        let mut text_shadow = self.pending_text_shadows
+                                  .pop()
+                                  .expect("Too many PopTextShadows?");
+        if !text_shadow.text_primitives.is_empty() {
+            let prim_cpu = TextShadowPrimitiveCpu {
+                text_primitives: text_shadow.text_primitives,
+                shadow: text_shadow.shadow,
+            };
+
+            text_shadow.local_rect = text_shadow.local_rect
```
kvark: no need to actually modify the `local_rect` here as opposed to just declaring a new variable (thus lifting `mut` on `text_shadow`)

webrender/src/prim_store.rs
```diff
@@ -493,10 +499,36 @@ pub struct TextRunPrimitiveCpu {
...
+                          device_pixel_ratio: f32,
+                          display_list: &BuiltDisplayList) {
+        // Cache the glyph positions, if not in the cache already.
+        // TODO(gw): In the future, remove `glyph_instances`
+        //           completely, and just reference the glyphs
+        //           directly from the displaty list.
+        if self.glyph_instances.is_empty() {
+            let src_glyphs = display_list.get(self.glyph_range);
+            for src in src_glyphs {
+                self.glyph_instances.push(GlyphInstance {
```
kvark: nit: could look better with `self.glyph_instances.extend(..)`

webrender/src/prim_store.rs
```diff
@@ -997,50 +1050,34 @@ impl PrimitiveStore {
...
-                                              &text.glyph_instances,
-                                              text.render_mode,
-                                              text.glyph_options);
+                // This is a text-shadow element. Create a render task that will
+                // render the text run to a target, and then apply a gaussian
+                // blur to that text run in order to build the actual primitive
+                // which will be blitted to the framebuffer.
+                let cache_width = (metadata.local_rect.size.width * device_pixel_ratio).ceil() as i32;
+                let cache_height = (metadata.local_rect.size.height * device_pixel_ratio).ceil() as i32;
+                let cache_size = DeviceIntSize::new(cache_width, cache_height);
```
kvark: could probably write via `(metadata.local_rect.size * device_pixel_ratio).ceil().to_i32()`

PR #1471, opened 2017-07-12
Partially revert #1407

mrobinson:
This change caused many WebGL tests to fail in Servo, which is blocking
the WebRender update. Partially revert the change in order to reland it
later with the issue properly fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1471)


mrobinson:
cc @nox

kvark:
It appears to revert all the changes except for dependency updates.
@MortimerGoro will hopefully take another stab at it later on (sorry about the revert!)
@bors-servo r+

MortimerGoro:
@kvark Ok, no problem. I'm working on https://github.com/servo/webrender/issues/1353 so the WebGL code will be refactored to a new component.  I'll check all the tests after the refactor.

PR #1472, opened 2017-07-12
Let iframes accept a LocalClip

mrobinson:
This will make it easier for Servo iframes to clip their contents. The
LocalClip is converted into a containing ClipScrollTree node around the
content of the iframe.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1472)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -483,6 +483,7 @@ impl Frame {
                           pipeline_id: PipelineId,
                           parent_id: ClipId,
                           bounds: &LayerRect,
+                          local_clip: &LocalClip,
                           context: &mut FlattenContext,
                           reference_frame_relative_offset: LayerVector2D) {
         let pipeline = match context.scene.pipeline_map.get(&pipeline_id) {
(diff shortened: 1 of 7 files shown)


webrender/src/frame.rs
```diff
@@ -495,6 +496,17 @@ impl Frame {
             None => return,
         };
 
+        let mut clip_region = ClipRegion::for_clip_node_with_local_clip(local_clip);
+        clip_region.origin += reference_frame_relative_offset;
+        let parent_pipeline_id = parent_id.pipeline_id();
+        let clip_id = self.clip_scroll_tree.generate_new_clip_id(parent_pipeline_id);
+        println!("clip: {:?}", clip_region);
```
kvark: leftover `println`

webrender/src/mask_cache.rs
```diff
@@ -44,6 +44,14 @@ impl ClipRegion {
         }
     }
 
+    pub fn for_clip_node_with_local_clip(local_clip: &LocalClip) -> ClipRegion {
```
glennw: The naming of this function seems a little unclear to me.

mrobinson: Okay. I changed this to create_for_clip_node_with_local_clip. I'll open a PR to make the other names coherent with it as well.

Issue #1470, opened 2017-07-12
Failed to rendered to FBO on ANGLE windows


sotaroikeda:
webrender's rendering failed on ANGLE windows in some situations. It is tracked by [Bug 1372083](https://bugzilla.mozilla.org/show_bug.cgi?id=1372083).

When the problem happened, Framebuffer validity check was failed because depth of ColorAttachment and DepthAttachment were different. The fbo was created with multiple layers in create_fbo_for_texture_if_necessary().

More information is in [Bug 1372083](https://bugzilla.mozilla.org/show_bug.cgi?id=1372083).

sotaroikeda:
"if (hasMismatchedSize(mDepthAttachment))" was failed in FramebufferState::attachmentsHaveSameDimensions().
  https://dxr.mozilla.org/mozilla-central/source/gfx/angle/src/libANGLE/Framebuffer.cpp#194

The attachmentsHaveSameDimensions() failed because dimensions of ColorAttachments and mDepthAttachment were different.

> sizes of mColorAttachments = {width=0x000004f4 height=0x00000400 depth=0x00000002 }
> size of mDepthAttachment = {width=0x000004f4 height=0x00000400 depth=0x00000001 }

depth of mDepthAttachment was always 1, since it was created with gl.renderbuffer_storage(). But depth of mColorAttachments was changed with layer_count in create_fbo_for_texture_if_necessary().

kvark:
I investigated this issue a bit, and here is what I found:
  - if one of the attachments is layered, then all others have to be layered as well. This doesn't apply to us, since our attachment is a single layer, we don't actually use layered rendering (yet).
  - GLES 3.0 and higher doesn't impose restrictions on the attachment sizes to match.
  - GLES 2.0 requires width and height of the attachments to match, which our case fulfills.

In terms of our codebase, the checks we perform on FBO creation can be improved, and I'm about to make a PR about it.

All in all, I'd bet this to be a bug in Angle, which takes the depth of the color texture as opposed to the depth of the color **attachment**.

sotaroikeda:
Thanks for the advice! Just changing attachmentsHaveSameDimensions() addressed the problem.

sotaroikeda:
The problem was addressed by changing ANGLE.

PR #1474, opened 2017-07-12
Depth RBO sharing and cleanup

kvark:
Related to  #1470
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1474)


glennw:
Nice!

@bors-servo r+

PR #1475, opened 2017-07-13
Sort the list of resource files for sharers.rs to be deterministic

glandium:
This causes sccache cache misses even when nothing changed, because the file system may not return the files in the same order.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1475)

### DIFF
--- a/webrender/build.rs
+++ b/webrender/build.rs
@@ -50,5 +50,9 @@ fn main() {
         }
     }
 
+    // Sort the file list so that the shaders.rs file is filled
+    // deterministically.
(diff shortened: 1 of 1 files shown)


webrender/build.rs
```diff
@@ -50,5 +50,9 @@ fn main() {
         }
     }
 
+    // Sort the file list so that the shaders.rs file is filled
+    // deterministically.
+    glsl_files.sort_by(|a, b| a.file_name().cmp(&b.file_name()));
```
kvark: nit: could use `sort_by_key`

Issue #1481, opened 2017-07-13
Text shadows draw over the actual text?


Gankro:
Running servo on master:

<img width="203" alt="screen shot 2017-07-13 at 7 13 00 pm" src="https://user-images.githubusercontent.com/1136864/28191747-44ecf0d2-6800-11e7-8df3-9c0b1b0dfcf2.png">

That's a green and red shadow for black text

In Gecko the black is clearly on top:

<img width="204" alt="screen shot 2017-07-13 at 7 20 36 pm" src="https://user-images.githubusercontent.com/1136864/28191778-663f6f80-6800-11e7-913b-5dd42a20a287.png">

Easy local test case: remove the transparency from the text in `reftests/text/shadow-complex.yaml`:

<img width="156" alt="screen shot 2017-07-13 at 7 25 07 pm" src="https://user-images.githubusercontent.com/1136864/28191894-08a55c6c-6801-11e7-9a97-5c133402b9eb.png">

kvark:
Fixed by #1482

Issue #1479, opened 2017-07-13
Shadow offsets don't actually offset the content


Gankro:
It appears that it only offsets the clipped/masked region:

From reftests/text/shadow-complex.yaml (with offset on red shadow set to [10, 10]) (mesa).

<img width="150" alt="screen shot 2017-07-13 at 11 12 17 am" src="https://user-images.githubusercontent.com/1136864/28173783-74fc3c58-67bd-11e7-8b60-20ed9e96f045.png">

(Note the T at the start)

From running servo on https://gankro.github.io/blah/webtests/text.html (macos sierra):

<img width="655" alt="screen shot 2017-07-13 at 10 37 51 am" src="https://user-images.githubusercontent.com/1136864/28173796-7d161b3e-67bd-11e7-869e-8de15dd0d28f.png">

(Note the `m` on "cillum")

Gankro:
Note: I tried messing around with the bounds/local_rect a bunch and nothing happened. I assume you intended to place the content according to it, but forgot to, since the only place you seem to factor in `offset` is in translating that rect.

glennw:
I'll look into these today.

kvark:
Fixed by #1482

PR #1482, opened 2017-07-14
Fix a number of issues with text-shadow.

glennw:
* Correctly handle text-shadow offset. The previous code
  was translating the local rect of the cache primitive correctly,
  however the local space offset between the text runs
  and the shadow rect itself weren't being handled.
* The paint order of non-transparent text runs added
  inside a push/pop text-shadow stack was incorrect. Now
  the code maintains a pending list of visual text runs
  encountered during a text-shadow stack, and pushes
  them as primitives after all shadows are popped.
* The text run cache shader was not correctly placing
  the glyphs in the middle of the render task that is
  used as input to the blur render tasks. This meant that
  the blur was not centered correctly, and may have
  resulted in artifacts around the edge of the blur.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1482)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -14,9 +14,9 @@ use gpu_cache::GpuCache;
 use internal_types::HardwareCompositeOp;
 use mask_cache::{ClipMode, ClipRegion, ClipSource, MaskCacheInfo};
 use plane_split::{BspSplitter, Polygon, Splitter};
-use prim_store::{GradientPrimitiveCpu, ImagePrimitiveCpu};
+use prim_store::{GradientPrimitiveCpu, ImagePrimitiveCpu, ShadowTextRun};
 use prim_store::{ImagePrimitiveKind, PrimitiveContainer, PrimitiveIndex};
-use prim_store::{PrimitiveStore, RadialGradientPrimitiveCpu};
+use prim_store::{PrimitiveStore, RadialGradientPrimitiveCpu, TextRun};
 use prim_store::{RectanglePrimitive, TextRunPrimitiveCpu, TextShadowPrimitiveCpu};
 use prim_store::{BoxShadowPrimitiveCpu, TexelRect, YuvImagePrimitiveCpu};
 use profiler::{FrameProfileCounters, GpuCacheProfileCounters, TextureCacheProfileCounters};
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

cc @Gankro 

Remaining (potential) issues:
 * I couldn't repro the edge artifacts that @Gankro was seeing on my linux or mac machines. If they're still occurring after this change, we can investigate further.
 * I don't adjust the bounds of the supplied local clip rect by the shadow offset. It's not clear to me whether we should do this or not - but it's an easy change either way.
 * The draw order when there are multiple shadows in a single stack is done for ease of implementation. We can change the ordering of this internally if it becomes a problem for callers.

glennw:
Fixes #1481

glennw:
Fixes #1479

Gankro:
Nice!

screen shot 2017-07-13 at 11 46 43 pm

I still see the artifacts on the non-blurred shadow.

Gankro:
Also as a minor point, gecko applies subpixel-aa to non-blurred shadows. So webrender's non-blurred shadows come out a bit blurrier/fainter. Not sure how important that is, but presumably that's a pretty easy to fix with a fast-path for `blur==0.0`?

glennw:
Ah, yep that's an easy fix.

Gankro:
As another minor point, gecko seems to produce a blurrier shadow for the same non-zero blur. Not sure if that's a different blur algorithm, us needing to fudge more, or servo needing to adjust our input.

screen shot 2017-07-14 at 12 01 32 am

webrender/src/frame_builder.rs
```diff
@@ -106,34 +106,123 @@ pub struct FrameBuilderConfig {
...
+        debug_assert!(self.has_shadows());
+
+        // Add the text run to each shadow in the list.
+        // TODO(gw): In the future we may be able to optimize
+        //           this to share glyph resources better.
+        for pending_shadow in &mut self.pending_shadows {
+            pending_shadow.push_text(&prim.run, &local_rect);
+        }
+
+        // If the color of the text run is not transparent,
```
kvark: probably needs to be clarified as "not fully transparent", because alpha=0.5 is transparent and the condition below handles it.

webrender/src/tiling.rs
```diff
@@ -1033,7 +1033,27 @@ impl RenderTarget for ColorRenderTarget {
...
+                          |   ...                                   |
+                          +-----------------------------------------+
+
+                        */
+
+                        // The base glyph index is multipled by 2 here
+                        // since the glyph lookup is calculated in the
+                        // vertex shader by dividing by 2 due to the
+                        // way glyphs are packed two per GPU block.
+                        let mut base_index = 2 * prim.runs.len() as i32;
```
kvark: this is rather hacky
I wish we had a way of defining such GPU layouts in the type system and then somehow enforcing the shaders to follow them at run-time...

PR #1487, opened 2017-07-15
Ensure the depth attachment is resized along with the color attachments (Fixes #1485)

### LINKED ISSUE #1485: Box shadows aren't rendering correctly at larger sizes
Found after updating stylish a large number of versions (https://github.com/Thinkofname/stylish/commit/3824c5da48c59ff68d38f2b57c170e43251feec1)

Tested on nvidia cards (970, 980 ti) with driver versions 381.22 and 384.47.

Nvidia 970: http://i.imgur.com/UBhbnqY.png
Nvidia 980 ti: http://i.imgur.com/5AxnEEe.png 
After resizing more http://i.imgur.com/PTis3XX.png

Only test case I have currently is running stylish at this commit https://github.com/Thinkofname/stylish/commit/3824c5da48c59ff68d38f2b57c170e43251feec1 with `cargo run -p stylish_webrender --example demo` and full-screening the window
Its possible i'm somehow misusing the API but I haven't spotted it yet

Thinkofname:
When the textures were resized the depth buffer kept its original size. This caused the framebuffer to render at the smallest attachment size (the depth buffer).

This fixes the shadow issue I was having in #1485. Whilst trying debug this I noticed `create_fbo_for_texture_if_necessary`  [only seems to care about the number of layers](https://github.com/servo/webrender/blob/master/webrender/src/device.rs#L1244) when deciding to resize the textures/renderbuffers and not whether the size changed. In this case it seems the number of layers does change triggering a rebuild but this may be an issue in other cases?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1487)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -354,6 +354,7 @@ struct Texture {
     format: ImageFormat,
     width: u32,
     height: u32,
+
     filter: TextureFilter,
     mode: RenderTargetMode,
     fbo_ids: Vec<FBOId>,
@@ -1170,9 +1171,11 @@ impl Device {
                         pixels: Option<&[u8]>) {
         debug_assert!(self.inside_frame);
 
+        let resized;
         {
             let texture = self.textures.get_mut(&texture_id).expect("Didn't find texture!");
             texture.format = format;
+            resized = texture.width != width || texture.height != height;
             texture.width = width;
             texture.height = height;
             texture.filter = filter;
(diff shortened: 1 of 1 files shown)


kvark:
Thanks for the PR!

>  In this case it seems the number of layers does change triggering a rebuild but this may be an issue in other cases?

I agree this is an issue. Would you want to address it as well?

Your PR looks like a step in the right direction, although there is a few concerns:
  1. You are basically changing the semantics of `create_fbo_for_texture_if_necessary`. Perhaps, let's rename it to `update_texture_storage`?
  2. The `layer_count == None` path is not covered, so you might want to update it as well

Thinkofname:
@kvark 
1. Done, it was somewhat doing this already anyway
2. Doesn't appear to have the same issue as it just has a color attachment and not a depth one (not sure why though)

Had to handle a case where the number of layers decreased and decided to delete the extra FBOs however i'm not sure how often the number of layers is expected to change given the old TODO that was there before. In my test with stylish it only ever changed on resize but this may be different in servo? If its often it might be better to just leave them instead so they can be reused.

glennw:
Looks good, thanks!

@bors-servo r+

PR #1488, opened 2017-07-17
Various improvements to text-shadow support.

### LINKED ISSUE #1483: apply subpixel-aa to non-blurred text-shadows
Gecko does this. Should hopefully be a simple fast-path.

glennw:
* Clamp UV fetches to 0.5 texels inside the UV rect for ps_cache and cs_blur shaders.
  Hopefully, this will fix the text-shadow artifacts that appear on some machines / tests.
* Add fast path for text that has blur radius of zero.
  This is a useful optimization, and also needed to pass some of the Servo reftests, which don't support fuzzy tests.
  Text runs with a zero shadow get pushed through the normal text run path, which now supports an offset parameter.
* Respect subpixel AA mode for text runs with zero blur radius.
  Fixes #1483.
* Respect disabled text AA mode for text runs with zero blur radius.
  This is needed by some of the Servo reftests.
* Simplify the GPU cache layout of text-shadow primitives.
  They now match that of a normal text run primitive, but have several runs concatenated.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1488)

### DIFF
--- a/webrender/res/cs_blur.vs.glsl
+++ b/webrender/res/cs_blur.vs.glsl
@@ -54,7 +54,8 @@ void main(void) {
             break;
     }
 
-    vUvRect = vec4(src_task.data0.xy, src_task.data0.xy + src_task.data0.zw);
+    vUvRect = vec4(src_task.data0.xy + vec2(0.5),
+                   src_task.data0.xy + src_task.data0.zw - vec2(0.5));
     vUvRect /= texture_size.xyxy;
 
     vec2 uv0 = src_task.data0.xy / texture_size;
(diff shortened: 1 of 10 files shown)


glennw:
r? @kvark 

cc @Gankro 

There's one issue remaining to get all the Servo reftests passing again - support for the underline text decoration in text shadows. I'll have a PR for that shortly, that supports non-blurred text decorations, which is all Servo tests for. This will allow us to do a WR update now, while we continue to work on the complete text-decoration support this week.

glennw:
Added a second commit - this has just enough support for text decorations to allow the Servo reftests to all pass. This will allow a WR update while we continue work on the full text decoration support. The commit message has a more detailed description of how it achieves this.

Gankro:
niiiice!
 
screen shot 2017-07-17 at 8 50 25 am

You should be able to enable the text reftest I just landed but had commented out.

glennw:
@bors-servo r=kvark

(homu queue seems incorrect).

PR #1490, opened 2017-07-17
Support resizing items in the texture cache.

nical:
I revived #925 to address #1367.

TextureCache::update now supports changing the size of the image by freeing and reallocating a rectangle in the cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1490)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -452,6 +452,8 @@ pub struct TextureCacheItem {
     // Handle to the location of the UV rect for this item in GPU cache.
     pub uv_rect_handle: GpuCacheHandle,
 
+    pub format: ImageFormat,
+
     // Some arbitrary data associated with this item.
     // In the case of glyphs, it is the top / left offset
     // from the rasterized glyph.
@@ -493,6 +495,7 @@ impl FreeListItem for TextureCacheItem {
 impl TextureCacheItem {
     fn new(texture_id: CacheTextureId,
            rect: DeviceUintRect,
+           format: ImageFormat,
            user_data: [f32; 2])
            -> TextureCacheItem {
         TextureCacheItem {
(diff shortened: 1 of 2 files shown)


webrender/src/texture_cache.rs
```diff
@@ -746,13 +778,28 @@ impl TextureCache {
...
+
+        if existing_item.allocated_rect.size.width != descriptor.width ||
+           existing_item.allocated_rect.size.height != descriptor.height {
 
-        // TODO(gw): Handle updates to size/format!
-        debug_assert_eq!(existing_item.allocated_rect.size.width, descriptor.width);
-        debug_assert_eq!(existing_item.allocated_rect.size.height, descriptor.height);
+            self.free_item_rect(existing_item.clone());
+
+            self.allocate_impl(
```
kvark: this would allocate but not really fill it with data we got in the arguments (`data`), so I think `return` should be removed

webrender/src/texture_cache.rs
```diff
@@ -746,13 +778,28 @@ impl TextureCache {
     pub fn update(&mut self,
                   image_id: TextureCacheItemId,
                   descriptor: ImageDescriptor,
+                  filter: TextureFilter,
                   data: ImageData,
                   dirty_rect: Option<DeviceUintRect>) {
-        let existing_item = self.items.get(image_id);
+        let existing_item = self.items.get(image_id).clone();
+
+        if existing_item.allocated_rect.size.width != descriptor.width ||
```
kvark: we'd also want to compare the formats

nical:
> we'd also want to compare the formats

I left the assertion that formats don't change in update_image_template. It's probably easy to add but I would rather do it as a followup.

> this would allocate but not really fill it with data we got in the arguments (data), so I think return should be removed

Oops! That and also I needed to discard the dirty rect when reallocating, and some remaining assertions that had to be removed. Fixed in the next commit which I'll squash whenever the we are done with the reviews.
I hacked together a little app to check that it works (and it does now). I'll have to figure out how properly test this. A good candidate for the rawtest framework!

Issue #1480, opened 2017-07-13
LineDisplayItem to support text-decoration-style


Gankro:
Discussing with @glennw on irc, here's our plan for supporting text-decorations, and their resulting shadows.

Add:

```rust
struct LineDisplayItem {
  baseline: f32, // LayerPixel
  start: f32,
  end: f32,
  orientation: Vertical | Horizontal, // toggles whether above values are interpreted as x/y values

  width: f32, // same units as borders? 
  // I don't *think* we need to distinguish font-weight from font-size here? (looking at gecko)
  color: Color,
  style: Solid | Dotted | Dashed | Wavy,
}
```

Producers of these items will figure out appropriate position, style, and ordering. Also in the case of a `text-decoration-style: double`, they will just emit two of these. So in total a span of text might have up to 6 LineDisplayItems pushed in with it.

In the backend, these will get converted into glyphs stored in the glyph cache, to be tiled for the length of the line (very short lines may get a dedicated glyph?). This unifies all text shadow rendering under "I render shadows of glyphs", which is nice and clean.

# Alternatives

## Attach LineDisplayItem to TextDisplayItem

This doesn't really clarify anything, other than hardcoding how many lines we can handle. Note that we can't just have a single list of LineDisplayItems, because line-through needs to come after glyphs, while underline and overline come before. Having lines be "loose" optimizes for the common case of 0 to 1 decoration lines, and makes it easy to support the different orderings. (note that the shadow rendering pass can ignore these orderings because it converts everything to the same color)

Note also that lines may span multiple TextDisplayItems because of font/color changes mid-text-span (mixing languages, emoji), and so this would reduce the amount of lines we actually need to move around.

## Overload borders

Borders already have a notion of dotted/dashed/solid, so reusing them here might make sense.

Borders are more complex than decorations (bloat) and have configurations we don't want to support in shadows. At the same time they don't have the "wavy" style text-decorations *do* need to support and that we don't want to support in borders.

(similarly, boxes only handle the solid/double cases)

## Support double lines directly

More complex than necessary to avoid minimal bloat since double-lines are incredibly rare.

glennw:
This happened.

PR #1491, opened 2017-07-17
Line item

Gankro:
This is an implementation of the API I proposed in #1480. The backend impl is basically WIP but gives me something to test, and also implements the only decoration Servo actually has (solid).

I was sitting on this hoping for feedback on the issue, but since @glennw had to hack in his own api to support Servo I figured I'd just push this up for him to use.

In implementing this, I can't help but wonder if this should just derive the line properties from the local_rect + an orientation flag. That said that's a bit weird for wavy lines. This api gives a uniform translation from CSS to webrender -- although perhaps this isn't that useful, considering the client needs to compute a larger bounding rect for the wavy lines anyway?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1491)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -11,11 +11,12 @@ use {BorderDetails, BorderDisplayItem, BorderWidths, BoxShadowClipMode, BoxShado
 use {ClipAndScrollInfo, ClipDisplayItem, ClipId, ColorF, ComplexClipRegion, DisplayItem};
 use {ExtendMode, FilterOp, FontKey, GlyphInstance, GlyphOptions, Gradient, GradientDisplayItem};
 use {GradientStop, IframeDisplayItem, ImageDisplayItem, ImageKey, ImageMask, ImageRendering};
-use {LayoutPoint, LayoutRect, LayoutSize, LayoutTransform, LayoutVector2D, LocalClip};
-use {MixBlendMode, PipelineId, PropertyBinding, PushStackingContextDisplayItem, RadialGradient};
-use {RadialGradientDisplayItem, RectangleDisplayItem, ScrollPolicy, SpecificDisplayItem};
-use {StackingContext, TextDisplayItem, TextShadow, TransformStyle, WebGLContextId, WebGLDisplayItem};
-use {YuvColorSpace, YuvData, YuvImageDisplayItem};
+use {LayoutPoint, LayoutRect, LayoutSize, LayoutTransform, LayoutVector2D, LineDisplayItem};
+use {LineOrientation, LineStyle, LocalClip, MixBlendMode, PipelineId};
+use {PropertyBinding, PushStackingContextDisplayItem, RadialGradient};
+use {RadialGradientDisplayItem, RectangleDisplayItem, ScrollPolicy};
+use {SpecificDisplayItem, StackingContext, TextDisplayItem, TextShadow, TransformStyle};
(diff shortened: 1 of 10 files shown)


Gankro:
Modified: got rid of the local_rect per @kvark's recommendation; it just encodes redundant information. In theory we can optimize this item by stuffing our data in there but not a big deal.

webrender_api/src/display_list.rs
```diff
@@ -494,6 +495,23 @@ impl DisplayListBuilder {
         self.push_item(item, rect, local_clip);
     }
 
+    pub fn push_line(&mut self,
+                     local_clip: Option<LocalClip>,
+                     baseline: f32,
+                     start: f32,
```
kvark: nit: can we have `Range<f32>` here?

wrench/src/yaml_frame_reader.rs
```diff
@@ -220,6 +220,28 @@ impl YamlFrameReader {
         self.builder().push_rect(rect, Some(local_clip), color);
     }
 
+    fn handle_line(&mut self, item: &Yaml, local_clip: LocalClip) {
+        let color = item["color"].as_colorf().unwrap_or(*BLACK_COLOR);
+        let baseline = item["baseline"].as_f32().expect("line must have baseline");
+        let start = item["start"].as_f32().expect("line must have start");
+        let end = item["end"].as_f32().expect("line must have end");
+        let width = item["width"].as_f32().expect("line must have width");
+        let orientation = match item["orientation"].as_str().expect("line must have orientation") {
```
kvark: did you consider using `define_string_enum` for implementing those enum-string conversions?

webrender_api/src/display_list.rs
Gankro: Like, core::ops::Range? Doesn't seem right since it's exclusive on one end.

kvark: All our rectangles are exclusive on one end, in a way: https://github.com/servo/euclid/blob/master/src/rect.rs#L180
So if you are using the start/end to generate a rectangle, it should play along just fine.

But really, that exclusiveness deserves some dedicated investigation, so I'm totally fine without `Range` here

glennw:
What if we remove the baseline / start / end fields, and go back to having a local rect plus an orientation?

In that case, the cross-axis size of the rect would implicitly determine the peak to peak height of wavy items, the height of a (dashed) line, or the radius of dots.

That seems simpler to me, but I haven't considered the current implementations in Gecko / Servo, or whether that doesn't cover some use cases. Thoughts?

glennw:
That also allows exact placement of the item based as the calling code determines.

Gankro:
Looking at Servo, it seems like it has font metrics that tell them where to place lines, and how thick to make them. This thickness can presumably can be fed in directly as the height of all the decorations, except for wavy lines. For wavy lines, it probably makes more sense to interpret the height as thickness, and not the bounds -- I don't think our clients need to know the "true" bounds of lines; they don't affect layout.

Gankro:
Btw this function is full of Good Stuff on lower bounds, when to apply snapping/aa, and how they derive sizes/offsets for the different styles: https://github.com/mozilla/gecko-dev/blob/master/layout/painting/nsCSSRendering.cpp#L3786

glennw:
Closing in favor of https://github.com/servo/webrender/pull/1504 which includes these commits.

PR #1498, opened 2017-07-19
Refactor how text shadows and elements are handled.

glennw:
While working on adding line decorations, I kept running into
issues and extra complexity. Instead, I've refactored how
text shadow contexts are handled. The implementation is a bit
complex in parts, but conceptually this fits in much better with
how the frame builder, prim store and GPU cache work.

Now, text-shadow primitives add text runs and decorations as
sub-primitives, using an index buffer of primitives. This means:

* The code to manage the text-element stack and maintain correct
  paint order is much simpler.
* The GPU cache layouts are much simpler. Text runs and decorations
  within a text-shadow element have the same layout as normal
  primitives (no weird offsets and packed layouts).
* Text primitives that have both visual and shadows (which is the
  common case) share the same GPU primitive - so they only get
  built and uploaded once. This is quite a large performance win
  on pages with a lot of shadows.
* The code to truncate glyph positions is moved from the CPU
  to the vertex shader, which is a significant performance win.
* The paint order of shadows is reversed from the previous
  implementation, back to what it was originally.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1498)

### DIFF
--- a/webrender/res/cs_text_run.vs.glsl
+++ b/webrender/res/cs_text_run.vs.glsl
@@ -13,6 +13,17 @@ void main(void) {
 
     int glyph_index = prim.user_data0;
     int resource_address = prim.user_data1;
+    int text_shadow_address = prim.user_data2;
+
+    // Fetch the parent text-shadow for this primitive. This allows the code
+    // below to normalize the glyph offsets relative to the original text
+    // shadow rect, which is the union of all elements that make up this
+    // text shadow. This allows the text shadow to be rendered at an
+    // arbitrary location in a render target (provided by the render
+    // task render_target_origin field).
+    PrimitiveGeometry shadow_geom = fetch_primitive_geometry(text_shadow_address);
+    TextShadow shadow = fetch_text_shadow(text_shadow_address + VECS_PER_PRIM_HEADER);
+
     Glyph glyph = fetch_glyph(prim.specific_prim_address, glyph_index);
     GlyphResource res = fetch_glyph_resource(resource_address);
 
@@ -22,7 +33,7 @@ void main(void) {
     vec2 size = res.uv_rect.zw - res.uv_rect.xy;
     vec2 local_pos = glyph.offset + vec2(res.offset.x, -res.offset.y) / uDevicePixelRatio;
     vec2 origin = prim.task.render_target_origin +
-                  uDevicePixelRatio * (text.offset + local_pos);
+                  uDevicePixelRatio * (local_pos + shadow.offset - shadow_geom.local_rect.p0);
     vec4 local_rect = vec4(origin, size);
 
     vec2 texture_size = vec2(textureSize(sColor0, 0));
@@ -34,7 +45,7 @@ void main(void) {
                    aPosition.xy);
 
     vUv = mix(st0, st1, aPosition.xy);
-    vColor = text.color;
+    vColor = shadow.color;
 
     gl_Position = uTransform * vec4(pos, 0.0, 1.0);
 }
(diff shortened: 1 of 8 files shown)


glennw:
r? @kvark

Sorry for another large patch on this stuff!

webrender/res/cs_text_run.vs.glsl
```diff
@@ -22,7 +33,7 @@ void main(void) {
     vec2 size = res.uv_rect.zw - res.uv_rect.xy;
     vec2 local_pos = glyph.offset + vec2(res.offset.x, -res.offset.y) / uDevicePixelRatio;
     vec2 origin = prim.task.render_target_origin +
-                  uDevicePixelRatio * (text.offset + local_pos);
+                  uDevicePixelRatio * (local_pos + shadow.offset - shadow_geom.local_rect.p0);
```
kvark: does everything that goes through `cs_text_run` has to have a shadow?

webrender/src/frame_builder.rs
```diff
@@ -675,12 +477,25 @@ impl FrameBuilder {
         //           text decoration support is added (via the
         //           Line display item) this can be removed, so that
         //           rectangles don't participate in text shadows.
-        if self.text_shadow_builder.has_shadows() {
-            self.text_shadow_builder.push_decoration(rect,
-                                                     color,
-                                                     local_clip,
-                                                     clip_and_scroll);
-        } else if color.a > 0.0 {
+        let mut shadows = Vec::new();
```
kvark: let's rename it to something like `trivial_shadow`?

webrender/src/frame_builder.rs
```diff
@@ -1041,36 +856,90 @@ impl FrameBuilder {
...
+        // possible. For any text shadows that have zero blur, create a normal text
+        // primitive with the shadow's color and offset. These need to be added
+        // *before* the visual text primitive in order to get the correct paint
+        // order. Store them in a Vec first to work around borrowck issues.
+        // TODO(gw): Refactor to avoid having to store them in a Vec first.
+        let mut fast_text_shadow_prims = Vec::new();
+        for shadow_prim_index in &self.shadow_prim_stack {
+            let shadow_metadata = &self.prim_store.cpu_metadata[shadow_prim_index.0];
+            let shadow_prim = &self.prim_store.cpu_text_shadows[shadow_metadata.cpu_prim_index.0];
+            if shadow_prim.shadow.blur_radius == 0.0 {
```
kvark: somewhat unfortunate that we already have the same loop at the start of the function, maybe we could combine them?

webrender/src/prim_store.rs
```diff
@@ -1058,8 +1037,33 @@ impl PrimitiveStore {
...
+            (metadata.prim_kind, metadata.cpu_prim_index)
+        };
+
+        // Recurse into any sub primitives and prepare them for rendering first.
+        // TODO(gw): This code is a bit hacky to work around the borrow checker.
+        //           Specifically, the clone() below on the primitive list for
+        //           text shadow primitives. Consider restructuring this code to
+        //           avoid borrow checker issues.
+        match prim_kind {
+            PrimitiveKind::TextShadow => {
```
kvark: nit: looks like a case for `if let`

kvark:
r=me whether or not the notes are addressed

webrender/res/cs_text_run.vs.glsl
glennw: Yep

webrender/src/frame_builder.rs
glennw: Unfortunately, we need to do part of it before adding the prim and part afterwards.

glennw:
@kvark Thanks, issued addressed or explained above.

@bors-servo r=kvark

Issue #1416, opened 2017-06-21
Need API to set a composition operator that's not "over"


mstange:
Webrender currently supports [MixBlendModes](https://github.com/servo/webrender/blob/4074d8c66cbfd01dce123aa3e36778a0036976d6/webrender_traits/src/display_item.rs#L301), but all the modes in this list don't affect the destination alpha. In Gecko we also need the modes Clear and Destination Out. Ideally we'd have all the porter duff modes so that we can support the [whole list of Moz2D CompositionOp values](http://searchfox.org/mozilla-central/rev/ae94cfb36d804727dfb38f2750bfcaab4621df6e/gfx/2d/Types.h#190).

We have two cases where we need this, and both of them are about native appearance on Mac.
 1. We need to clear the background behind vibrant regions, e.g. behind the bookmarks sidebar. In Gecko we do that using a ColorLayer whose color is transparent black and whose composition op is OP_SOURCE. But it could just as well use OP_CLEAR.
 2. We round the window corners by painting four small images in the corners with OP_DEST_OUT. This causes the destination pixels outside of the rounded shape to become transparent.

mstange:
@glennw There's one aspect of this that may be worth thinking about as part of #1774:

For optimal performance with operator source, the concept of "opaqueness" may need to be split into two dimensions. If you draw something with operator source, the old values of the destination pixels get completely ignored and replaced. So you'll want to cull any primitives underneath anything that's drawn with operator source, because anything that was drawn underneath has no impact on the final rendering anyway. However, after drawing the operator source item, the affected pixels may be transparent, and for example any text that is drawn directly on top will not be able to use subpixel AA.

The same applies to operator clear.

So there's a difference between opaque in the sense that items underneath can be culled, and opaque in the sense that items on top will draw to opaque pixels. Just something to be aware of.

kvark:
> However, after drawing the operator source item, the affected pixels may be transparent, and for example any text that is drawn directly on top will not be able to use subpixel AA.

I'm having trouble understanding this part. Why would a primitive care about what's drawn on top of it?
In general, the current difference in handling opaque vs transparent is because the latter depends on what's drawn below it, not on top.

mstange:
The primitive itself does not care about what is drawn on top of it. But the rendering system as a whole does. In general, some decisions about a given item depend on the type of items underneath it, and some depend on the type of items on top of it.

[...]

PR #1499, opened 2017-07-19
tiling: derive Default for CompositeOps.

emilio:
I was poking at #1416, and found this while at it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1499)


emilio:
@bors-servo retry

 * ICE in GCC?

PR #1500, opened 2017-07-19
Make ClipScrollGroup per ClipScrollInfo

mrobinson:
Instead of making each ClipScrollGroup per stacking context and
ClipScrollInfo combination, make them per ClipScrollInfo. This should
reduce the amount of work done per stacking context and is the first
step toward accepting all coordinates relative to reference frames.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1500)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -6,10 +6,11 @@ use api::{BorderDetails, BorderDisplayItem, BoxShadowClipMode, ClipAndScrollInfo
 use api::{DeviceIntPoint, DeviceIntRect, DeviceIntSize, DeviceUintRect, DeviceUintSize};
 use api::{ExtendMode, FontKey, FontRenderMode, GlyphInstance, GlyphOptions, GradientStop};
 use api::{ImageKey, ImageRendering, ItemRange, LayerPoint, LayerRect, LayerSize};
-use api::{LayerToScrollTransform, LayerVector2D, LineOrientation, LineStyle, LocalClip};
-use api::{PipelineId, RepeatMode, TextShadow, TileOffset, TransformStyle};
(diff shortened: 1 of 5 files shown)


webrender/src/frame_builder.rs
```diff
@@ -294,6 +295,7 @@ pub struct FrameBuilder {
 
     stacking_context_store: Vec<StackingContext>,
     clip_scroll_group_store: Vec<ClipScrollGroup>,
+    clip_scroll_group_indices: HashMap<ClipAndScrollInfo, ClipScrollGroupIndex>,
```
glennw: We should make this use Fnv as the hasher.

webrender/src/frame_builder.rs
```diff
@@ -1050,6 +1046,7 @@ impl FrameBuilder {
             run: TextRun {
                 font_key,
                 logical_font_size: size,
+                run_offset,
```
glennw: There is an offset field in text runs with this https://github.com/servo/webrender/pull/1498/ - perhaps we can make use of that here, which applies the offset in the vertex shader? That could be done as a follow up though.

webrender/src/frame_builder.rs
mrobinson: Okay. I'll replace this with FnvHasher.

webrender/src/frame_builder.rs
mrobinson: The newest version just reuses this offset. Thanks for the suggestion!

glennw:
Needs a rebase.

mrobinson:
Rebased. Going to land the new version. Thank you both for the reviews!

@bors-servo r=glennw

PR #1504, opened 2017-07-20
Basic implementation of line decoration display items.

glennw:
Includes the commits in https://github.com/servo/webrender/pull/1491 from @Gankro.

This provides a basic implementation of the line decoration styles in CSS.

I haven't specifically tried to make the styles (e.g. wavy, dotted) match other browsers, yet. I think it's reasonable to get this reviewed and merged, and then tweak the visuals as a follow up.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1504)


glennw:
Sample of the current line styles, including text-shadow blur support:

decorations-suite

webrender/src/prim_store.rs
```diff
@@ -180,6 +181,24 @@ impl ToGpuBlocks for RectanglePrimitive {
...
+pub struct LinePrimitive {
+    pub color: ColorF,
+    pub style: LineStyle,
+    pub orientation: LineOrientation,
+}
+
+impl ToGpuBlocks for LinePrimitive {
+    fn write_gpu_blocks(&self, mut request: GpuDataRequest) {
+        request.push(self.color);
+        request.push([pack_as_float(self.style as u32),
```
kvark: what's the reason to call `pack_as_float` here as opposed to just casting to `f32`?

webrender/src/renderer.rs
```diff
@@ -1878,6 +1904,22 @@ impl Renderer {
                                       &target.text_run_textures,
                                       &projection);
         }
+        if !target.line_cache_prims.is_empty() {
+            // TODO(gw): Technically, we don't need blend for solid
+            //           lines. We could check that here?
```
kvark: if that becomes a performance concern, we'd need to split those into opaque/non-opaque

Gankro:
Seems good!

I kinda lost track of what you really wanted with the line-dimension specifying stuff. From servo's perspective, boxes + orientation would be easier to send (because of how their logical->real conversion works). Gecko would probably be fine either way?

Gankro:
Also it appears that you're trying to make waves have a fixed width, and you just stretch the height. Gecko's algorithm is to have a fixed slope (45 degrees), so the width of a wave increases with its height:

screen shot 2017-07-20 at 3 46 51 pm 1

webrender/res/ps_line.fs.glsl
```diff
@@ -0,0 +1,123 @@
+#line 1
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+float det(vec2 a, vec2 b) {
```
glennw: It's a 2d cross product, I guess. It's just using the same name as the paper the algorithm was taken from.

webrender/src/prim_store.rs
glennw: ```pack_as_float``` adds 0.5 to ensure that casting back to an int definitely gets the correct value. It's more a defensive measure than anything - perhaps it's not necessary?

webrender/src/renderer.rs
glennw: Yup

Issue #1357, opened 2017-06-09
APZ doesn't seem to be async enough


jrmuizel:
I'm concerned that rasterizing blob images, glyphs and processing a new scene will block scrolling. If this is true we should do better.

nical:
We could group all blob images for a certain frame and only upload them to the texture cache when the entire group is rasterized to preserve some form of frame consistency, but not block the frame on the blob images being rasterized (or have the render backend thread block for at most a certain amount of time). This would not provide full frame consistency because we could still do some layout changes and see them on the screen before blob images from the same frame are ready. Full frame consistency while preserving APZ is hard. We could maybe just block the render backend until blob images are rasterized when there is a layout change so that we preserve APZ as much as possible and only jank when layout changes (does not sound great but still better than where we are at today).
We could also only wait for the blob images if they were created or updated this frame. This would result in checkerboarding when scrolling as blob image tiles get requested (kinda like what we have for tiled painted layer in gecko today).

For glyphs it would be cool if we could just replace missing glyphs with a placeholder when rasterization takes too long (the place holder could be an already-rasterized same letter with a different size or font. I bet it wouldn't be very noticeable (less noticeable than popping or missing the frame budget).

jrmuizel:
I don't think we'll be able to have the kind of frame consistency problems that you describe. It might be worth trying but we'll need a plain that has better consistency.

kvark:
Rough plan to proceed (adopted from @nical ):
  1. Fully detach the scene building from `ResourceCache` (part of the work went in #1527).
  2. Move the scene building to another thread (potentially, in a thread pool)
  3. Move (some of?) rasterization of images/glyphs into the scene building, or out of the main RB thread anyway.

nical:
> Move (some of?) rasterization of images/glyphs into the scene building, or out of the main RB thread anyway.

To add some precision, these are already happening on the thread pool, but the frame construction blocks until they are done. There is some tweaking we could do to stop blocking on at least a subset of the cpu rasterization, and that would probably require doing a bit of rasterization as part of the off-render-backend scene construction to avoid flickering in some cases. But we can go into the details of that when we have the infrastructure in place to build the scene asynchronously.

PR #1509, opened 2017-07-22
Multi-doc support

kvark:
A namespace can now contain multiple documents. Operations on a document are done ~~via a separate object `DocumentApi`, which created from `ResourceApi` that manages resources~~ by providing `DocumentId` parameter to `RenderApi` calls.

This PR is a step towards #1357, also relates to #1370.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1509)

### DIFF
--- a/webrender/examples/animation.rs
+++ b/webrender/examples/animation.rs
@@ -45,9 +45,7 @@ lazy_static! {
     static ref TRANSFORM: Mutex<LayoutTransform> = Mutex::new(LayoutTransform::identity());
 }
 
-fn event_handler(event: &glutin::Event,
-                 api: &RenderApi)
-{
+fn event_handler(event: &glutin::Event, document_id: DocumentId, api: &RenderApi) {
     match *event {
         glutin::Event::KeyboardInput(glutin::ElementState::Pressed, _, Some(key)) => {
             let offset = match key {
@@ -61,7 +59,7 @@ fn event_handler(event: &glutin::Event,
             // webrender using the generate_frame API. This will recomposite with
             // the updated transform.
             let new_transform = TRANSFORM.lock().unwrap().post_translate(LayoutVector3D::new(offset.0, offset.1, 0.0));
-            api.generate_frame(Some(DynamicProperties {
+            api.generate_frame(document_id, Some(DynamicProperties {
                 transforms: vec![
                   PropertyValue {
                     key: PropertyBindingKey::new(42),
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -322,17 +322,17 @@ lazy_static! {
     static ref TOUCH_STATE: Mutex<TouchState> = Mutex::new(TouchState::new());
 }
 
-fn event_handler(event: &glutin::Event, api: &RenderApi) {
+fn event_handler(event: &glutin::Event, document_id: DocumentId, api: &RenderApi) {
     match *event {
         glutin::Event::Touch(touch) => {
             match TOUCH_STATE.lock().unwrap().handle_event(touch) {
                 TouchResult::Pan(pan) => {
-                    api.set_pan(pan);
-                    api.generate_frame(None);
+                    api.set_pan(document_id, pan);
+                    api.generate_frame(document_id, None);
                 }
                 TouchResult::Zoom(zoom) => {
-                    api.set_pinch_zoom(ZoomFactor::new(zoom));
-                    api.generate_frame(None);
+                    api.set_pinch_zoom(document_id, ZoomFactor::new(zoom));
+                    api.generate_frame(document_id, None);
                 }
                 TouchResult::None => {}
             }
--- a/webrender/examples/blob.rs
+++ b/webrender/examples/blob.rs
@@ -262,9 +262,7 @@ fn body(api: &api::RenderApi,
     builder.pop_stacking_context();
 }
 
-fn event_handler(_event: &glutin::Event,
-                 _api: &api::RenderApi)
-{
+fn event_handler(_event: &glutin::Event, _document_id: api::DocumentId, _api: &api::RenderApi) {
 }
 
 fn main() {
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -56,6 +56,7 @@ pub fn main_wrapper(builder_callback: fn(&RenderApi,
                                          &PipelineId,
                                          &LayoutSize) -> (),
                     event_handler: fn(&glutin::Event,
+                                      DocumentId,
                                       &RenderApi) -> (),
                     options: Option<webrender::RendererOptions>)
 {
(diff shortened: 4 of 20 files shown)


glennw:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[webrender/src/lib.rs, line 16 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmO5Y9Yf7jvHQoNvUk:-KpmO5YA44phEnWCKmT7:b-14eodz) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/webrender/src/lib.rs#L16)):*
> ```Rust
> //! The main entry point to WebRender is the `webrender::renderer::Renderer`.
> //!
> //! By calling `Renderer::new(...)` you get a `Renderer`, as well as a `DocumentApiSender`.
> ```

It's not a DocumentApiSender you get from that?

---

*[webrender/src/lib.rs, line 19 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmODfMhtKyhFmlFOWA:-KpmODfNeVBxRfwNh12U:b2nreh6) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/webrender/src/lib.rs#L19)):*
> ```Rust
> //! Your `Renderer` is responsible to render the previously processed frames onto the screen.
> //!
> //! By calling `yourDocumentApiSenderInstance.create_api()`, you'll get a `DocumentApi` instance,
> ```

Same as above.

---

*[webrender/src/render_backend.rs, line 123 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmNZ4nJyHewaMa5MZn:-KpmNZ4nJyHewaMa5MZo:bpogl36) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/webrender/src/render_backend.rs#L123)):*
> ```Rust
> 
>     frame_config: FrameBuilderConfig,
>     documents: HashMap<DocumentId, Document>,
> ```

Let's use Fnv hasher here.

---

*[wrench/src/json_frame_writer.rs, line 73 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmOSeBjK3pxQNJNMK6:-KpmOSeBjK3pxQNJNMK7:b50oiae) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/wrench/src/json_frame_writer.rs#L73)):*
> ```Rust
>     }
> 
>     pub fn begin_write_display_list(&mut self,
> ```

I wonder if we should just remove the json from writer completely? I don't think anyone uses it...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1509)*

glennw:
Looks great! Just a few minor comments / changes.

glennw:
Also, it'd be good to have the corresponding Servo patch ready to go when we land this so we can get a WR update landed straight after.

nical:
The separation between the resource and document apis looks nice but it will make the task of bundling image updates and display lists in the same transactions difficult.

nical:
I suggest not rushing to update servo and gecko, because if I manage to be convincing enough during tomorrow's meeting, we'll do something different (not totally different but with some substantial API changes).

kvark:
Servo integration is not trivial, WIP, but other notes are addressed now.

---

Review status: 12 of 21 files reviewed at latest revision, 4 unresolved discussions.

---

*[webrender/src/lib.rs, line 16 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmO5Y9Yf7jvHQoNvUk:-Kpq19JPBPtpUuDzfVIy:b-7ql1td) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/webrender/src/lib.rs#L16)):*

It's not a DocumentApiSender you get from that?
</blockquote></details>

fixed

---

*[webrender/src/lib.rs, line 19 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmODfMhtKyhFmlFOWA:-Kpq1Aa_0EcFYGvyUfBY:b-7ql1td) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/webrender/src/lib.rs#L19)):*

Same as above.
</blockquote></details>

fixed

---

*[webrender/src/render_backend.rs, line 123 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmNZ4nJyHewaMa5MZn:-Kpq1Bx5BPbKjJ9Up93H:b-7ql1td) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/webrender/src/render_backend.rs#L123)):*

Let's use Fnv hasher here.
</blockquote></details>

fixed

---

*[wrench/src/json_frame_writer.rs, line 73 at r1](https://reviewable.io:443/reviews/servo/webrender/1509#-KpmOSeBjK3pxQNJNMK6:-Kpq1DSBLddurxVbvc0l:b-p4pqfy) ([raw file](https://github.com/servo/webrender/blob/820a8b1633de3352eaa8a1f36af56c30d8c5472c/wrench/src/json_frame_writer.rs#L73)):*

I wonder if we should just remove the json from writer completely? I don't think anyone uses it...
</blockquote></details>

agreed, but better left for a dedicated PR

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1509#-:-Kpq1IIb8Agww_FX9wdd:bywd3un)*

jrmuizel:
Previously, glennw (Glenn Watson) wrote…
>  I wonder if we should just remove the json from writer completely? I don't think anyone uses it...

It's been used for getting readable dumps of binary recordings. I think we could probably simplify it to just providing that functionality.

kvark:
@glennw servo update is in https://github.com/kvark/servo/commit/4f940df68722232a12b934f0e66054b23ae12140

glennw:
Reviewed 21 of 21 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1509)*

kvark:
Rebasing this is a constant pain...
@bors-servo r=glennw

sotaroikeda:
@kvark, can you explain about actual benefit of supporting Multi-doc support? Is it still in the middle of implementation? Gecko creates GL context for each window, but webrender Renderer instance still owns single GL context, therefore each window needs to have different Renderer instance. Then current gecko just use one document on each Renderer and each window creates each webrender Renderer by [Bug 1385003](https://bugzilla.mozilla.org/show_bug.cgi?id=1385003) Thanks!

kvark:
@sotaroikeda the idea is that WR can have multiple Renderer instances sharing the same backend and receiving frames only from the corresponding documents.

sotaroikeda:
@nical, do you have any ideas to use Multi-doc support in gecko?

nical:
> @nical, do you have any ideas to use Multi-doc support in gecko?

I haven't looked into the details since the multi-doc stuff landed, but I think that it'd be good to complete the implementation so that documents can use their own renderer. That way we can look into grouping some windows like tooltips into using a single render backend thread, and share fonts, glyph cache, etc.

> Is it still in the middle of implementation? 

Yes, as far as I can tell (it doesn't let you have a renderer per document yet for example).

kvark:
Basically, https://github.com/servo/webrender/blob/2ac2da08b77aff15cc7e38118fa20071262b07e9/webrender/src/renderer.rs#L1455

PR #1510, opened 2017-07-23
Document ColorF and ColorU

pyfisch:
Move some color code from display_item module.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1510)


pyfisch:
Btw, `PackedColor` is identical to `ColorU` (which already implements `repr(C)`). Is there a reason to keep the types separate? `PackedColor::from_color` duplicates `From<ColorF> for ColorU`.

glennw:
@pyfisch I think ```PackedColor``` just existed as an internal detail before we had a public ```ColorU``` type. We should probably remove PackedColor.

@bors-servo r+

PR #1511, opened 2017-07-23
Remove LowLevelFilterOp

pyfisch:
LowLevelFilterOp duplicated the FilterOp from Webrender API.
All types in the low level variant used Au instead of f32 but
where always converted from floats and are converted back to
floats. So there is no advantage at all to use integers.

Also use a float for blur for consistency. See also #1244.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1511)


glennw:
Nice tidy up, thanks!

@bors-servo r+

PR #1512, opened 2017-07-24
Implement basic text layout for wrench.

glennw:
Add just enough information to the WR font platforms to do basic text layout.

Tested on Linux, Mac, Windows.

This allows us to add text runs to wrench reftests and benchmarks without
having to manually specify all the glyph positions.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1512)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -19,6 +19,7 @@ use api::{ColorU, FontKey, FontRenderMode, GlyphDimensions};
 use api::{GlyphKey, GlyphOptions, SubpixelPoint};
 use api::NativeFontHandle;
 use gamma_lut::{GammaLut, Color as ColorLut};
+use std::ptr;
 
 pub struct FontContext {
     cg_fonts: HashMap<FontKey, CGFont>,
@@ -56,6 +57,7 @@ struct GlyphMetrics {
     rasterized_ascent: i32,
     rasterized_width: u32,
     rasterized_height: u32,
+    advance: f32,
 }
 
 // According to the Skia source code, there's no public API to
@@ -97,6 +99,7 @@ fn get_glyph_metrics(ct_font: &CTFont,
             rasterized_height: 0,
             rasterized_ascent: 0,
             rasterized_descent: 0,
+            advance: 0.0,
         };
     }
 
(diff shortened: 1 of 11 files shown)


glennw:
r? @kvark 

The platform code implementations are a bit messy - mostly copy and paste from old code in Servo and/or Wrench. But it works!

webrender/src/platform/macos/font.rs
```diff
@@ -210,6 +219,24 @@ impl FontContext {
         }
     }
 
+    pub fn get_glyph_index(&mut self, font_key: FontKey, ch: char) -> Option<u32> {
+        let characters: [u16; 1] = [ ch as u16 ];
```
kvark: is there a point of using fixed size arrays of size 1 here as opposed to straight values?

webrender/src/platform/windows/font.rs
```diff
@@ -223,7 +211,35 @@ impl FontContext {
...
+
+        let bounds = analysis.get_alpha_texture_bounds(texture_type);
+
+        let width = (bounds.right - bounds.left) as u32;
+        let height = (bounds.bottom - bounds.top) as u32;
+
+        // Alpha texture bounds can sometimes return an empty rect
+        // Such as for spaces
+        if width == 0 || height == 0 {
+            return None
```
kvark: what is the exact semantics of `None` for the result of this function? The lack of glyph in the font, or the lack of its bitmap?

wrench/src/wrench.rs
```diff
@@ -218,6 +193,38 @@ impl Wrench {
...
+                                    *glyph_index,
+                                    LayerPoint::zero(),
+                                    FontRenderMode::Alpha));
+        }
+        let metrics = self.api.get_glyph_dimensions(keys);
+
+        // Extract the advances from the metrics. The get_glyph_dimensions API
+        // has a limitation that it can't currently get dimensions for non-renderable
+        // glyphs (e.g. spaces), so just use a rough estimate in that case.
+        let space_advance = size.to_f32_px() / 3.0;
```
kvark: what is the point of making this rough estimate if the only usage of the function (AFAIS) is debug info output?

webrender/src/platform/windows/font.rs
```diff
@@ -214,6 +192,16 @@ impl FontContext {
                                          0.0, 0.0)
     }
 
+    pub fn get_glyph_index(&mut self, font_key: FontKey, ch: char) -> Option<u32> {
+        let face = self.fonts.get(&font_key).unwrap();
+        let indices = face.get_glyph_indices(&[ch as u32]);
+        if indices.len() == 1 {
```
kvark: nit: if the length can't be higher than `1`, we could use `first().map(..)` here

webrender/src/platform/windows/font.rs
glennw: Either of those will result in None right now - however as a follow up I'd like to make this API work for the lack of a bitmap (so spaces etc are handled).

wrench/src/wrench.rs
glennw: Just to make the benchmarks a bit nicer to inspect the output of.

glennw:
@kvark Thanks, comments addressed and nits fixed.

kvark:
Thank you!
@bors-servo r+

PR #1513, opened 2017-07-24
Allow ScrollFrame to be insensitive to input events

mrobinson:
overflow:hidden should be scrollable in Servo, but not by input events.
This change allows creating ScrollFrames which are only scrollable via
script.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1513)


mrobinson:
Here is the Servo portion of this change: https://github.com/mrobinson/servo/tree/scrollable-overflow-hidden

kvark:
Looks good, thank you!
@bors-servo r+

PR #1514, opened 2017-07-25
Various improvements to benchmark mode, plus a glyph stress benchmark.

glennw:
* Track the CPU backend time in benchmark mode.
* Add a glyph rendering stress test.
* Make subpixel AA default text mode in wrench.
* Rebuild display lists every frame in benchmark mode.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1514)


glennw:
Rebased.

wrench/src/args.yaml
```diff
@@ -24,10 +24,10 @@ args:
       long: save
       help: 'Save frames, one of: yaml, json, or binary'
       takes_value: true
-  - subpixel_aa:
+  - disable_subpixel_aa:
```
kvark: I think it should be `no_subpixel_aa` for consistency with other options

wrench/src/main.rs
```diff
@@ -320,7 +320,7 @@ fn main() {
                                  save_type,
                                  size,
                                  args.is_present("rebuild"),
-                                 args.is_present("subpixel_aa"),
+                                 !args.is_present("disable_subpixel_aa"),
```
kvark: similarly, we might want to move the inversion `!` later down the pipe in order to be consistent with the other options

wrench/src/perf.rs
```diff
@@ -188,30 +190,28 @@ impl<'a> PerfHarness<'a> {
...
-                                         .sum();
+        let composite_time_ns = extract_sample(&mut cpu_frame_profiles, |a| a.composite_time_ns);
+        let paint_time_ns = extract_sample(&mut gpu_frame_profiles, |a| a.paint_time_ns);
+        let backend_time_ns = extract_sample(&mut cpu_frame_profiles, |a| a.backend_time_ns);
 
         TestProfile {
             name: filename.to_str().unwrap().to_string(),
-            composite_time_ns: composite_time / cpu_samples.len() as u64,
-            paint_time_ns: paint_time / gpu_samples.len() as u64,
+            composite_time_ns: composite_time_ns,
```
kvark: nit: looks like you haven't gotten used to the new shiny Rust 1.16 features yet :)

wrench/src/perf.rs
```diff
@@ -188,30 +190,28 @@ impl<'a> PerfHarness<'a> {
...
+            paint_time_ns: paint_time_ns,
+            backend_time_ns: backend_time_ns,
             draw_calls,
         }
     }
 }
 
+fn extract_sample<F, T>(profiles: &mut [T], f: F) -> u64 where F: Fn(&T) -> u64 {
```
kvark: I like this one 👍

glennw:
@kvark Thanks, fixed those nits.

PR #1515, opened 2017-07-25
Improve wrench display list "replay" and DL/IPC timing.

glennw:
Previously, wrench held on to a single DLBuilder object, and copied
this and finalize'ed it each time a DL was replayed. This has several
issues, and messes up the timings of the DL profiling.

Instead, change wrench to build the DL once, and store the built
display lists. This mostly involves plumbing the DL builder through
all the yaml reader functions.

Also, modify the DL timing code to handle the situation where a DL
is built once and sent multiple times for profiling purposes. Specifically,
record an extra timestamp just before starting the send operation and
accumulate the total time as a sum of each component.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1515)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -329,13 +329,20 @@ pub struct IpcProfileCounters {
 }
 
 impl IpcProfileCounters {
-    pub fn set(&mut self, build_start: u64, build_end: u64, 
-                              consume_start: u64, consume_end: u64,
-                              display_len: usize) {
-        self.build_time.inc(build_end - build_start);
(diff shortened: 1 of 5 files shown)


glennw:
r? @kvark and @Gankro (specifically the DL/IPC timing changes to make sure I haven't messed these up).

This allows better benchmarking in wrench of processing a new DL each frame, e.g. for benchmarking stress test pages with lots of glyphs.

Gankro:
r+

Gankro:
bors r u ok...?

PR #1516, opened 2017-07-25
Make wrench yaml reader able to share font templates.

glennw:
Previously, every text run would get a different font template
loaded, preventing any glyphs being shared.

This does a minimal amount of work to share font templates, so
that text benchmarking is closer to real world usage.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1516)


glennw:
Rebased.

@bors-servo r=kvark

Issue #1503, opened 2017-07-19
Glyph cache optimizations / ideas


glennw:
A general list of ideas for improving glyph cache performance:

* We currently have way too many hash table lookups - e.g. the resource cache expiry times are maintained in a separate hash from the items. We could combine some of these hash lookups, and also could consider using a freelist in place of the hashmap in some cases.
* The glyph cache currently uses a fairly complex allocator to minimize wasted space in the atlas. Instead, it's probably worth switching to some kind of slab allocator (with different block sizes per texture page) - this will mean no fragmentation / coalescing, and make allocating items in the cache much faster. The trade-off is more texture memory used for the glyph cache, but if we choose the block sizes carefully this shouldn't be a big deal.
* The hash table to look up a glyph key is quite expensive. Since the actual used glyph keys are typically a small subset of the u32 char index, we could consider something like a page table structure to remove the hashing overhead.
* Since the cost of rasterizing a glyph on the CPU is significantly higher than the cost of uploading a glyph to the GPU, we could consider having 2-4 glyph caches, which are independent of each other (can duplicate glyphs in GPU memory), so that the text run processing can be trivially distributed across CPU threads.

glennw:
cc @kvark

Gankro:
As part of this I'm working on adding a list of all glyphs to the end of the display list, so that those can be uploaded in parallel.

glennw:
These optimizations have been applied - text processing generally doesn't show up near the top of the profile now.

PR #1518, opened 2017-07-25
add a cache of all seen glyphs to the end of the display list

Gankro:
This lets the backend call `display_list.glyphs()` to get an iterator of every glyph seen (broken down by font/color) so that the backend can prepare glyph atlases or whatever in parallel.

This is for @glennw to build off of for #1503

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1518)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1070,6 +1070,7 @@ dependencies = [
  "core-graphics 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "dwrote 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "euclid 0.15.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fxhash 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "gleam 0.4.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "heapsize 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ipc-channel 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


glennw:
Nice!

@bors-servo r+

kvark:
Needs a rebase

Gankro:
Whoops, missed this. Fixed.

glennw:
@Gankro Having recently done some testing and benchmarking of Sip / Fnv / Fx hash, this comment is a bit of a concern now:

```
FIXME: audit whether fast hashers (FNV?) are safe here
```

What would we need to do to find out if we can use FxHash here?

glennw:
I have zero relevant security knowledge, but given we use ```FxHash``` internally now for the glyph and texture cache, we're probably no worse off using it during display list construction :)

Given that - are you happy to change this to use FxHash for the hasher type and we'll get this landed?

Gankro:
Yeah sure.

glennw:
Thanks, no rush on this from my end by the way - sorry if that wasn't clear from the comment above.

Gankro:
Ok done. It didn't seem right to have webrender_api publically vendor this, so I just duplicated it in lib.rs.

glennw:
CI failure:

```
error: `pub(restricted)` syntax is experimental (see issue #32409)

  --> webrender_api/src/lib.rs:58:5

   |

58 | pub(crate) type FastHashMap<K, V> = HashMap<K, V, BuildHasherDefault<fxhash::FxHasher>>;

   |     ^^^^^

error: `pub(restricted)` syntax is experimental (see issue #32409)

  --> webrender_api/src/lib.rs:59:5

   |

59 | pub(crate) type FastHashSet<T> = HashSet<T, BuildHasherDefault<fxhash::FxHasher>>;

   |     ^^^^^

error: Could not compile `webrender_api`.
```

We are compiling against 1.17 which, I think, is the required rustc version for Gecko.

Gankro:
Fixed. Forgot submodules can see private members of parents.

PR #1521, opened 2017-07-26
Add a flag into RendererOptions to prevent render_on_scroll

### LINKED ISSUE #1497: Need a way to disable RenderBackend::render_on_scroll
Gecko with webrender does not need "render on scroll", it just spend cpu time. It is nice if we could disable it.

Gecko expects to generate new frames only with ApiMsg::GenerateFrame, since gecko generates new frames with v-sync timings. It is done by [WebRenderBridgeParent::CompositeToTarget()](https://dxr.mozilla.org/mozilla-central/source/gfx/layers/wr/WebRenderBridgeParent.cpp#1082). Frame scrolling is also controled by CompositeToTarget(). Then there is no necessity to do render on scroll by webrender. It just consumes cpu time.

gecko already does not handle new_scroll_frame_ready event by [Bug 1377688](https://bugzilla.mozilla.org/show_bug.cgi?id=1377688).

sotaroikeda:
Fixes #1497

cc  @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1521)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -68,6 +68,8 @@ pub struct RenderBackend {
 
     vr_compositor_handler: Arc<Mutex<Option<Box<VRCompositorHandler>>>>,
 
+    enable_render_on_scroll: bool,
+
     // A helper switch to prevent any frames rendering triggered by scrolling
     // messages between `SetDisplayList` and `GenerateFrame`.
     // If we allow them, then a reftest that scrolls a few layers before generating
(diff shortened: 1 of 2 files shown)


webrender/src/render_backend.rs
```diff
@@ -433,7 +437,9 @@ impl RenderBackend {
                                 });
                             }
 
-                            self.render_on_scroll = true;
+                            if self.enable_render_on_scroll {
+                                self.render_on_scroll = true;
```
kvark: could just be `self.render_on_scroll = self.enable_render_on_scroll`

webrender/src/render_backend.rs
```diff
@@ -121,6 +124,7 @@ impl RenderBackend {
             vr_compositor_handler,
             window_size: initial_window_size,
             inner_rect: DeviceUintRect::new(DeviceUintPoint::zero(), initial_window_size),
+            enable_render_on_scroll: enable_render_on_scroll,
```
kvark: nit: could just be `enable_render_on_scroll,`

sotaroikeda: Thanks, I'll update it in a next patch.

webrender/src/render_backend.rs
sotaroikeda: Yea, I'll update it in a next patch.

kvark:
Thank you!
@bors-servo r+

PR #1522, opened 2017-07-26
Move per-font-instance parameters from GlyphKey to FontInstanceKey.

glennw:
This is prep work only. It splits fields up into a separate struct
that has values that change per-font-instance vs per-glyph.

This makes several upcoming changes feasible that will reduce the
amount and cost of hash lookups we do in the resource cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1522)

### DIFF
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#[cfg(test)]
 use app_units::Au;
 use device::TextureFilter;
 use fnv::FnvHasher;
@@ -19,11 +20,11 @@ use std::collections::HashSet;
 use std::mem;
 use texture_cache::{TextureCacheItemId, TextureCache};
 #[cfg(test)]
-use api::IdNamespace;
-use api::{ColorF, LayoutPoint};
-use api::{FontKey, FontTemplate, FontRenderMode};
+use api::{ColorF, FontRenderMode, IdNamespace};
+use api::{FontInstanceKey, LayoutPoint};
+use api::{FontKey, FontTemplate};
 use api::{ImageData, ImageDescriptor, ImageFormat};
-use api::{GlyphKey, GlyphOptions, GlyphInstance, GlyphDimensions};
+use api::{GlyphKey, GlyphInstance, GlyphDimensions};
 
 pub type GlyphCache = ResourceClassCache<GlyphRequest, Option<TextureCacheItemId>>;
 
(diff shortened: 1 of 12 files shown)


glennw:
r? @kvark 

@lsalzman I hope this doesn't interfere too much with the font patches you're working on. I wanted to introduce this now, since it forms the basis of several optimizations I'm planning to make to how the resource cache handles the glyph <-> texture cache mappings.

I'm thinking that your work will basically end up containing a FontInstanceKey (or equivalent) inside each FontInstance structure.

Hopefully introducing this now is either useful for your work, or at least easier to integrate now rather than later - but let me know if it doesn't match up with what you're working on or you foresee any issues with it!

kvark:
AppVeyor got some valid errors: https://ci.appveyor.com/project/servo/webrender/build/1.0.4568#L379
```
error[E0433]: failed to resolve. Use of undeclared type or module `Au`
   --> src\glyph_rasterizer.rs:379:13
    |
379 |             Au::from_px(32),
    |             ^^^^^^^^^^^ Use of undeclared type or module `Au`
error[E0433]: failed to resolve. Use of undeclared type or module `ColorF`
   --> src\glyph_rasterizer.rs:380:13
    |
380 |             ColorF::new(0.0, 0.0, 0.0, 1.0),
    |             ^^^^^^^^^^^ Use of undeclared type or module `ColorF`
error[E0433]: failed to resolve. Use of undeclared type or module `FontRenderMode`
   --> src\glyph_rasterizer.rs:382:13
    |
382 |             FontRenderMode::Subpixel,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^ Use of undeclared type or module `FontRenderMode`
error[E0061]: this function takes 5 parameters but 9 parameters were supplied
   --> src\glyph_rasterizer.rs:376:13
    |
147 | /     pub fn request_glyphs(
148 | |         &mut self,
149 | |         glyph_cache: &mut GlyphCache,
150 | |         current_frame_id: FrameId,
...   |
213 | |         });
214 | |     }
    | |_____- defined here
...
376 |               &mut glyph_cache,
    |               ^^^^^^^^^^^^^^^^ expected 5 parameters
error: aborting due to 4 previous errors
```

webrender/src/glyph_rasterizer.rs
```diff
@@ -149,29 +148,24 @@ impl GlyphRasterizer {
...
-            let glyph_request = GlyphRequest::new(
-                font_key,
-                size,
-                color,
-                glyph.index,
-                glyph.point,
-                render_mode,
-                glyph_options,
-            );
+            glyph_request.key.index = glyph.index;
```
kvark: you aren't really saving either any lines of code or CPU cycles by re-using this `glyph_request`

webrender_api/src/font.rs
```diff
@@ -153,31 +153,61 @@ impl SubpixelPoint {
...
-impl GlyphKey {
+impl FontInstanceKey {
     pub fn new(font_key: FontKey,
                size: Au,
-               color: ColorF,
-               index: u32,
+               mut color: ColorF,
+               render_mode: FontRenderMode,
+               glyph_options: Option<GlyphOptions>) -> FontInstanceKey {
+        // In alpha/mono mode, the color of the font is irrelevant.
```
kvark: how is the color relevant to the font key when in `Subpixel` mode?
My first guess would be that the color values produced by the sub-pixel AA would just be multiplied by the given text color, which means that the rasterization is done independent of text color.

webrender_api/src/font.rs
```diff
@@ -153,31 +153,61 @@ impl SubpixelPoint {
...
+        // Forcing it to black in those cases saves rasterizing glyphs
+        // of different colors when not needed.
+        if render_mode != FontRenderMode::Subpixel {
+            color = ColorF::new(0.0, 0.0, 0.0, 1.0);
```
kvark: might add a `warn!` if the color is different

webrender/src/glyph_rasterizer.rs
glennw: Fixed

webrender_api/src/font.rs
glennw: In subpixel mode, the color gets backed into the glyphs themselves, due to the blend mode that we use.

webrender_api/src/font.rs
glennw: It's expected that the incoming color is not black in these cases - we just force it to black as an optimization.

glennw:
@kvark Thanks - nits and unit test fixed.

kvark:
Thank you!
@bors-servo r+

PR #1523, opened 2017-07-26
Remove PackedColor internal type

pyfisch:
Change some profiler fuctions to accept ColorU instead of ColorF.
Implement From<ColorU> for ColorF and get the Into for free.

Also pass Color not by reference but by value as they are Copy.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1523)


kvark:
Given that `PackedColor` as our internal implementation detail, this change will not affect WR clients (other than the fact some Color values are now passed by value, but that's minor), so it's safe to merge.

@bors-servo r+

PR #1524, opened 2017-07-26
Fix incorrectly deleted resources in clear_namespace function.

MortimerGoro:
clear_namespace function is removing resources from different namespaces from the IdNamespace argument. Image_keys and font_keys are filtered by `key.0 != namespace` instead of `key.0 == namespace`

Rust-1.18 is already available so I used the retain function to fix both the issue and the TODO.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1524)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -810,7 +810,7 @@ impl ResourceCache {
     pub fn clear_namespace(&mut self, namespace: IdNamespace) {
         //TODO: use `retain` when we are on Rust-1.18
         let image_keys: Vec<_> = self.resources.image_templates.images.keys()
-                                                                      .filter(|&key| key.0 != namespace)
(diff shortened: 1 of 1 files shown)


kvark:
Ooops!
Unfortunately, Gecko is still using Rust 1.16 AFAIK, so we can't use `retain` just yet.
Please revert those Rust-related changes to proceed.

MortimerGoro:
Ok, Done!

kvark:
Thank you!
@bors-servo r+

sotaroikeda:
By the way, rustc min version is 1.17 for gecko now with [Bug 1374807](https://bugzilla.mozilla.org/show_bug.cgi?id=1374807).

https://dxr.mozilla.org/mozilla-central/source/build/moz.configure/rust.configure#58

PR #1525, opened 2017-07-27
Text rendering optimizations and bug fixes.

glennw:
* Switch subpixel positioning to be applied on one axis only.
  Subpixel positioning only needs to be applied on the axis of the text run direction.
  Reduces the number of glyphs rasterized significantly, saving texture memory and CPU time.
* Switch shader truncating of text run vertices to be based on the subpixel positioning direction.
* Apply subpixel positioning to alpha AA text runs (including shadows).
* Fix a bug in Linux subpixel AA mode where the dimensions didn't match the bitmap.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1525)

### DIFF
--- a/webrender/res/cs_text_run.vs.glsl
+++ b/webrender/res/cs_text_run.vs.glsl
@@ -7,6 +7,10 @@
 // drawn un-transformed. These are used for effects such
 // as text-shadow.
 
+#define SUBPX_DIR_NONE        0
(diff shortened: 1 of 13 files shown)


glennw:
r? @kvark 

(I've run the tests on Linux / Mac).

glennw:
Rebased

PR #1528, opened 2017-07-28
Implement subpixel positioning for Linux / Freetype.

glennw:
Refactor how the bounding box / dimensions are calculated,
correctly handling subpixel positioning offsets.

Simplify the code to extract the rasterized bitmap, now that
we can rely on the bounding box being correct.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1528)

### DIFF
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -3,19 +3,19 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{FontInstanceKey, FontKey, FontRenderMode, GlyphDimensions};
-use api::{NativeFontHandle};
+use api::{NativeFontHandle, SubpixelDirection};
 use api::{GlyphKey};
 use internal_types::FastHashMap;
 
-use freetype::freetype::{FT_Render_Mode, FT_Pixel_Mode};
+use freetype::freetype::{FT_Render_Mode, FT_Pixel_Mode, FT_BBox, FT_Outline_Translate};
 use freetype::freetype::{FT_Done_FreeType, FT_Library_SetLcdFilter};
-use freetype::freetype::{FT_Library, FT_Set_Char_Size};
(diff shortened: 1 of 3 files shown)


glennw:
r? @kvark 

I've run the WPT / CSS tests on Linux, and done a reasonable amount of manual testing of websites with Servo.

glennw:
Rebased.

webrender/src/platform/unix/font.rs
```diff
@@ -228,100 +262,107 @@ impl FontContext {
...
-                        let mask = 0x80 >> (src_x & 0x7);
-                        let byte = unsafe {
-                            *base.offset((src_x >> 3) as isize)
-                        };
+        } as i32;
+
+        let actual_height = bitmap.rows as i32;
+        let top = unsafe { (*slot).bitmap_top };
+        let left = unsafe { (*slot).bitmap_left };
+        let mut final_buffer = vec![0; (actual_width * actual_height * 4) as usize];
```
kvark: I see how this new size (`actual_width * actual_height`) is now possible, provided your earlier changes to `RasterizedGlyph`. Basically, the metrics and rasterization are decoupled.

webrender/src/platform/unix/font.rs
```diff
@@ -228,100 +262,107 @@ impl FontContext {
...
-            }
-            for _x in dst_end .. dimensions.width as i32 {
-                final_buffer.extend_from_slice(&[0xff, 0xff, 0xff, 0]);
+                    _ => panic!("Unsupported {:?}", pixel_mode)
+                };
+                let i = 4 * (y * actual_width + x) as usize;
+                let dest = &mut final_buffer[i..i+4];
+                dest.clone_from_slice(&value);
             }
-            assert_eq!(final_buffer.len(), ((y+1) * dimensions.width * 4) as usize);
```
kvark: we need an assertion check here to ensure that `src` reached exactly the end of the source buffer

PR #1529, opened 2017-07-28
Implement updating external image buffers.

nical:
Initial implementation that ignores the dirty rect, which is correct but just misses the potential optimization of uploading a sub-rectangle of the image. Right now gecko is not wired up to send the dirty rect so I don't have a good way to test the optimization handy and I want to make progress on image updates as soon as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1529)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -816,8 +816,23 @@ impl TextureCache {
         }
 
         let op = match data {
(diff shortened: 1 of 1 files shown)


nical:
cc @JerryShih

kvark:
Looks straightforward, thank you!
@bors-servo r+

PR #1530, opened 2017-07-28
WebGL encapsulation

kvark:
Refactors our WebGL usage and fixes a regression from #1509
also related to #1353 (should make it easier)

r? @MortimerGoro 
cc @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1530)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -105,6 +105,59 @@ impl Document {
     }
 }
 
+struct WebGL {
+    last_id: WebGLContextId,
+    contexts: HashMap<WebGLContextId, GLContextWrapper, BuildHasherDefault<FnvHasher>>,
+    active_id: Option<WebGLContextId>,
+}
+
+impl WebGL {
+    fn new() -> Self {
+        WebGL {
+            last_id: WebGLContextId(0),
+            contexts: HashMap::with_hasher(Default::default()),
+            active_id: None,
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


MortimerGoro:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1530#-:-Kq8pAo_srV0EgOHCBzK:bnfp4nl)*

kvark:
not sure if this is going to work
@bors-servo r=MortimerGoro

webrender/src/render_backend.rs
```diff
@@ -105,6 +105,59 @@ impl Document {
     }
 }
 
+struct WebGL {
+    last_id: WebGLContextId,
+    contexts: HashMap<WebGLContextId, GLContextWrapper, BuildHasherDefault<FnvHasher>>,
```
emilio: nit: Could use `fnv::FnvHashMap`, and `HashMap::default()` below.

kvark: I just followed the existing code. Would be great to convert it in bulk to a shorter form as you propose, just in a separate PR

MortimerGoro: Th WebGL code in WR will be deleted after https://github.com/servo/servo/pull/17891 lands, so it's not worth the effort to work on that.

kvark: @MortimerGoro wow, +3,138 −1,417

PR #1531, opened 2017-07-31
Tidy up of the image request code in the texture cache.

glennw:
This is prep work for some upcoming work to reduce the amount
of hashing we do in the resource cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1531)


glennw:
r? @kvark  and @nical 

As far as I can tell, this shouldn't change any semantics. But please take a careful look at this one :)

kvark:
Pretty!
@bors-servo r+

PR #1532, opened 2017-07-31
Make it possible to batch updates into transactions in the API.

nical:
Add a `ResourceUpdates` struct which groups addition/updates/deletions of images and can be sent as transaction:
 -  with a display list using `api.set_display_list`
 -  or without display list using `api.update_resources` (replacing `add_image`, `update_image` and `delete_image`).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1532)

### DIFF
--- a/webrender/examples/animation.rs
+++ b/webrender/examples/animation.rs
@@ -22,6 +22,7 @@ use webrender::api::*;
 fn body(_api: &RenderApi,
         _document_id: &DocumentId,
         builder: &mut DisplayListBuilder,
+        _resouces: &mut ResourceUpdates,
         _pipeline_id: &PipelineId,
         _layout_size: &LayoutSize) {
     // Create a 100x100 stacking context with an animatable transform property.
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -176,6 +176,7 @@ fn main() {
 fn body(api: &RenderApi,
         _document_id: &DocumentId,
         builder: &mut DisplayListBuilder,
+        resources: &mut ResourceUpdates,
         _pipeline_id: &PipelineId,
         layout_size: &LayoutSize) {
     let bounds = LayoutRect::new(LayoutPoint::zero(), *layout_size);
@@ -188,10 +189,12 @@ fn body(api: &RenderApi,
                                   Vec::new());
 
     let image_mask_key = api.generate_image_key();
-    api.add_image(image_mask_key,
-                  ImageDescriptor::new(2, 2, ImageFormat::A8, true),
-                  ImageData::new(vec![0, 80, 180, 255]),
-                  None);
+    resources.add_image(
+        image_mask_key,
+        ImageDescriptor::new(2, 2, ImageFormat::A8, true),
+        ImageData::new(vec![0, 80, 180, 255]),
+        None
+    );
     let mask = ImageMask {
         image: image_mask_key,
         rect: (75, 75).by(100, 100),
@@ -231,7 +234,7 @@ fn body(api: &RenderApi,
     if false { // draw text?
         let font_key = api.generate_font_key();
         let font_bytes = load_file("res/FreeSans.ttf");
-        api.add_raw_font(font_key, font_bytes, 0);
+        resources.add_raw_font(font_key, font_bytes, 0);
 
         let text_bounds = (100, 200).by(700, 300);
         let glyphs = vec![
--- a/webrender/examples/blob.rs
+++ b/webrender/examples/blob.rs
@@ -215,10 +215,11 @@ impl api::BlobImageRenderer for CheckerboardRenderer {
 fn body(api: &api::RenderApi,
         _document_id: &api::DocumentId,
         builder: &mut api::DisplayListBuilder,
+        resources: &mut api::ResourceUpdates,
         _pipeline_id: &api::PipelineId,
         layout_size: &api::LayoutSize) {
     let blob_img1 = api.generate_image_key();
-    api.add_image(
+    resources.add_image(
         blob_img1,
         api::ImageDescriptor::new(500, 500, api::ImageFormat::BGRA8, true),
         api::ImageData::new_blob_image(serialize_blob(api::ColorU::new(50, 50, 150, 255))),
@@ -226,7 +227,7 @@ fn body(api: &api::RenderApi,
     );
 
     let blob_img2 = api.generate_image_key();
-    api.add_image(
+    resources.add_image(
         blob_img2,
         api::ImageDescriptor::new(200, 200, api::ImageFormat::BGRA8, true),
         api::ImageData::new_blob_image(serialize_blob(api::ColorU::new(50, 150, 50, 255))),
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -54,6 +54,7 @@ impl HandyDandyRectBuilder for (i32, i32) {
 pub fn main_wrapper(builder_callback: fn(&RenderApi,
                                          &DocumentId,
                                          &mut DisplayListBuilder,
+                                         &mut ResourceUpdates,
                                          &PipelineId,
                                          &LayoutSize) -> (),
                     event_handler: fn(&glutin::Event,
@@ -114,8 +115,9 @@ pub fn main_wrapper(builder_callback: fn(&RenderApi,
     let pipeline_id = PipelineId(0, 0);
     let layout_size = LayoutSize::new(width as f32, height as f32);
     let mut builder = DisplayListBuilder::new(pipeline_id, layout_size);
+    let mut resources = ResourceUpdates::new();
 
-    builder_callback(&api, &document_id, &mut builder, &pipeline_id, &layout_size);
+    builder_callback(&api, &document_id, &mut builder, &mut resources, &pipeline_id, &layout_size);
 
     api.set_display_list(
         document_id,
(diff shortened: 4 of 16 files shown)


nical:
This is starting to come together nicely. The wrench part isn't quite wired up properly yet, though.

nical:
If not a full review, I'd love to get some feedback before I write the servo and gecko patches that come with this API change.

glennw:
Reviewed 2 of 15 files at r1, 1 of 1 files at r2, 13 of 13 files at r3.

---

*[webrender_api/src/api.rs, line 22 at r3](https://reviewable.io:443/reviews/servo/webrender/1532#-Kqedg9jakF79DQIBioJ:-Kqedg9jakF79DQIBioK:b-yx21jf) ([raw file](https://github.com/servo/webrender/blob/5aeea0b9b26e5b5c6a406c531f8241f7ab3d6d8d/webrender_api/src/api.rs#L22)):*
> ```Rust
> /// The resource updates for a given transaction (they must be applied in the same frame).
> #[derive(Clone, Deserialize, Serialize)]
> pub struct ResourceUpdates {
> ```

Instead of having 5 arrays here, how about a single Vec<Update> where Update is an enum containing each of those update types? That will mean fewer allocations, and also allows expressing an exact order of operations within the transaction (that may not actually be necessary, but it seems like a nice property to have).

---

*[wrench/src/json_frame_writer.rs, line 109 at r3](https://reviewable.io:443/reviews/servo/webrender/1532#-KqeeZbrGWJ91HevM7tp:-KqeeZbrGWJ91HevM7tq:b-1kp8ow) ([raw file](https://github.com/servo/webrender/blob/5aeea0b9b26e5b5c6a406c531f8241f7ab3d6d8d/wrench/src/json_frame_writer.rs#L109)):*
> ```Rust
>     }
> 
>     fn update_resources(&mut self, updates: &ResourceUpdates) {
> ```

This seems to be duplicated - can we share the function somewhere?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1532)*

glennw:
@nical In general, this looks good to me. Added a couple of comments, but they only affect the internals - the public API change seems good. Thanks!

nical:
Reviewed 2 of 15 files at r1, 1 of 1 files at r2, 13 of 13 files at r3.

---

*[webrender_api/src/api.rs, line 22 at r3](https://reviewable.io:443/reviews/servo/webrender/1532#-Kqedg9jakF79DQIBioJ:-KqgPApndtk0LQfAJks_:bmm13wp) ([raw file](https://github.com/servo/webrender/blob/5aeea0b9b26e5b5c6a406c531f8241f7ab3d6d8d/webrender_api/src/api.rs#L22)):*

Instead of having 5 arrays here, how about a single Vec<Update> where Update is an enum containing each of those update types? That will mean fewer allocations, and also allows expressing an exact order of operations within the transaction (that may not actually be necessary, but it seems like a nice property to have).
</blockquote></details>

The exact order of destruction/update within a transaction should not matter. The idea was that the possibility to batch and sort by size insertions/deletions in the texture cache could help with fragmentation. That said we can do the sorting in the resource cache.
I'll  make it a single vec for now and maybe turn that into separate vecs later in the resource cache when I get to playing with the texture packing algorithm. 

---

*[wrench/src/json_frame_writer.rs, line 109 at r3](https://reviewable.io:443/reviews/servo/webrender/1532#-KqeeZbrGWJ91HevM7tp:-KqgVoC-UM6ar3tLC6XT:bzgadn9) ([raw file](https://github.com/servo/webrender/blob/5aeea0b9b26e5b5c6a406c531f8241f7ab3d6d8d/wrench/src/json_frame_writer.rs#L109)):*

This seems to be duplicated - can we share the function somewhere?
</blockquote></details>

The whole `*_frame_writer` stuff is duplicated. To share this (and the rest) we'd be better off refactoring it completely.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1532)*

nical:
I added a commit to make it a single vector of a ResourceUpdate enum. I still prefer having separate vectors but it's not very hard to switch between the two if we settle for something different later. Up to you!

glennw:
Needs a rebase again, sorry. Will get this merged as soon as rebased.

glennw:
This is an interesting point. We should definitely consider this as part of the design. Perhaps we could try to set up some benchmarks or test cases that demonstrate the issue?

kvark:
@glennw basically, scrolling a page with many images (either small or big - doesn't matter), aka tumblr/instagram.

nical:
> If we promise the WR client that the updates will happen right for this next frame, then rendering would have to be stalled for the updated GPU data to transfer. This means hitches and frame time spikes during scrolling...

You don't have a choice unless you want to artificially introduce frames of latency to compensate for the upload time. I really don't think that's a good idea, we already have plenty of latency between the DOM and the screen.
These frames of latency also come with the obligatory memory overhead of having both the previous and new versions of the image until the switch has been made.

The way I see it, the implicit versioning scheme takes something that is fairly doable out of webrender to put it in the hands of the client where it is a lot harder to handle. One of the reasons it is fundamentally harder to do for the client is that it is not in the thread where all of the smart things happen (the render backend). It means that whatever needs to figure out versioning never has an up to date view of the information it needs. The other reason is that adding versioning is more complicated is that it amounts to adding a problem (as in, having to do versioning at all) that we currently do not have with the transactions.

I am not found of introducing extra latency, but since you bring it up, I'll argue that it is a lot easier to implement a scheme where applying the transaction is delayed until its updated textures are transferred to the GPU. We already are adding a simple mechanism to delay applying transactions, we can trivially extend it wait until some uploads are done.

> With explicit versioning, WR could shove the responsibility to separate resource updates from their use to WR client. With implicit approach (this PR), we'll need to do something tricky from within WR.

We are WebRender's clients. Adding complexity in the client is not a particularly good thing (you end up with complex code to maintain in both servo and gecko instead of just in webrender). It only has value if something is a lot easier to implement in the client than in webrender. But really, adding versioning will make things more complicated in both the client and webrender.

kvark:
> You don't have a choice unless you want to artificially introduce frames of latency to compensate for the upload time. I really don't think that's a good idea, we already have plenty of latency between the DOM and the screen.

We can only start uploading the data when we are sure that it is used, thus - when the layer culling is done. We don't have that much time between layer culling and frame rendering at the moment.

> These frames of latency also come with the obligatory memory overhead of having both the previous and new versions of the image until the switch has been made.

I don't think there is any way around having copies. If you want the new frame to instantly have all the new data, you'll stall.

> We already are adding a simple mechanism to delay applying transactions, we can trivially extend it wait until some uploads are done.

It would be great if we can piggy-back on that delay.

---

I realized one way of reducing the transfer latency. Our texture cache should just be a set of persistently mapped manually synchronized staging textures (or PBOs). The CPU side would then need to potentially keep multiple versions of textures in flight (yes, that comes up again, Iknowright?). When an image is needed for rendering (i.e. it passed the culling phase and needs to be uploaded), we can just issue a texture-to-texture (or buffer-to-texture for PBOs) copy, thus saving the driver quite a bit of headache.

nical:
>  We can only start uploading the data when we are sure that it is used, thus - when the layer culling is done. We don't have that much time between layer culling and frame rendering at the moment.

This doesn't need to be the case. But maybe around the end of your comment we are thinking about the same thing, because you mentioned PBOs.
I think that it should be more efficient to have a pool of "transfer buffers" using PBOs to transfer the data to the GPU and then blit these into the texture cache using the GPU. Some GPUs/drivers do atrocious things with glTexSubImage and although it would sorta amount to re-implementing glTexSubImage, I we'd have more knobs to tweak like more async transfer and we could avoid poor driver implementations.

> I don't think there is any way around having copies. If you want the new frame to instantly have all the new data, you'll stall.

There is a trade-off here, and for sure if we have more uploads than what we can possibly handle within a frame, then we isn't much we can do other than delay applying changes to the texture cache which inevitably creates some buffering and uses more memory. But I think that in the common case we can actually fit these uploads within a frame without causing jank, so it would be great to not pay the cost of buffering when we don't need it. We could easily delay frames based some heuristic of the cost of uploads (like the number of pixels to transfer being above a certain threshold).

Keeping memory usage reasonable is hard work. Out of memory is the biggest source of crash in firefox (after plugins like flash), and it's one of those things where we are known to fare better than the competition, it would be great to stay that way.

Sorry about the long posts. There several topics here:
- The way we upload textures I think can be improved (regardless of whether to add delay).
- If we can't upload everything we need within a frame, the only ways to not jank is upload asynchronously with a frame of delay at the cost of some form of buffering, or break transactions (I believe that we can't rely on the latter at least not at this point). I think we agree here, although I would rather make this a special case than do that every time there is something to upload.
- Whether or not the transaction-based architecture is preventing us from implementing async texture transfers with delayed frames efficiently. I believe that we can do that optimally with this architecture.

kvark:
Correction:
> Our texture cache should just be a set of persistently ...

I mean not the texture cache, of course, but the image/font templates.

> Some GPUs/drivers do atrocious things with glTexSubImage and although it would sorta amount to re-implementing glTexSubImage, I we'd have more knobs to tweak like more async transfer and we could avoid poor driver implementations.

Even if the driver is doing the most optimal thing there, it would still have to do an extra copy of the data, and we'll get the result quite a bit later (than doing it manually).

> I think that it should be more efficient to have a pool of "transfer buffers" using PBOs to transfer the data to the GPU and then blit these into the texture cache using the GPU.

Worth pointing out that it's not blitting that we are going to do, it's data transfer. And it's not on GPU, it's using DMA. That's the main reason it stalls the GPU - because the compute unites have to wait for the DMA transfer to finish.

All in all, I think we mostly agree here. I raised the concern that we just need to keep in mind. I'll try to convert WR to use PBOs aggressively and check if that's going to be enough.

PR #1533, opened 2017-08-01
Export max texture size in renderer.

JerryShih:
There are some assert for the texture size checking.
https://hg.mozilla.org/mozilla-central/annotate/7ce557b85b61/gfx/webrender/src/texture_cache.rs#l693

I'm going to export the max texture size and handle the size problem in user space.

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -667,6 +667,8 @@ pub struct Renderer {
 
     notifier: Arc<Mutex<Option<Box<RenderNotifier>>>>,
 
+    max_texture_size: u32,
+
     enable_profiler: bool,
     max_recorded_profiles: usize,
     clear_framebuffer: bool,
@@ -1223,6 +1225,7 @@ impl Renderer {
             backend_profile_counters: BackendProfileCounters::new(),
             profile_counters: RendererProfileCounters::new(),
             profiler: Profiler::new(),
+            max_texture_size: max_texture_size,
             enable_profiler: options.enable_profiler,
             max_recorded_profiles: options.max_recorded_profiles,
             clear_framebuffer: options.clear_framebuffer,
@@ -1252,6 +1255,10 @@ impl Renderer {
 
         let sender = RenderApiSender::new(api_tx, payload_tx);
         Ok((renderer, sender))
+    }
+
+    pub fn get_max_texture_size(&self) -> u32 {
+        self.max_texture_size
     }
 
     pub fn get_graphics_api_info(&self) -> GraphicsApiInfo {


kvark:
> There are some assert for the texture size checking.

I thought we are auto-tiling textures not fitting into the max size. @nical ?

nical:
I am not sure what type of texture @JerryShih is referring to. External image buffers should be auto-tiled we have code specifically for to handle that case (not saying it is bug-free).

JerryShih:
I mean the "external image buffers". Not the hardware-specific one(e.g. macIOSurface or d3d texture).
One crash report[1] shows that wr hit the texture size checking[2].

I will talk with @nical and update the status later.

[1]
https://crash-stats.mozilla.com/report/index/da5b5deb-e838-488f-920a-b76380170723
[2]
https://hg.mozilla.org/mozilla-central/annotate/7ce557b85b61/gfx/webrender/src/texture_cache.rs#l693

JerryShih:
After some testing and talking with @nical, the "external image buffer" could use tiling. But if we use mask, there will be no tiling.

Here is a test page:
https://github.com/JerryShih/test_page/tree/gh-pages
The gecko still does the clipping for painted layer, so I can't get a large size of mask image with this test page. So, I set the maximum texture size into 2048[1]. Then, I could hit the texture size assertion.

[1]
https://hg.mozilla.org/mozilla-central/annotate/7ce557b85b61/gfx/webrender/src/texture_cache.rs#l693

sotaroikeda:
The following URL in [Bug 1370649](https://bugzilla.mozilla.org/show_bug.cgi?id=1370649) could hit the assert. In this case, height=8922 > max texture size=8192 (format: A8) on my linux pc.

http://www.mmo-champion.com/content/6669-Interview-Darkmoon-Faire-Class-Mounts-Power-Ascended-Raid-Testing-Armory-Update

JerryShih:
Thanks! I will use this link to test my patch.

sotaroikeda:
@JerryShih, how do you use this max texture size info? This max texture size is different from TextureCache::max_texture_size.  The TextureCache::max_texture_size was limited to SQRT_MAX_RGBA_PIXELS_PER_TEXTURE(8192) by the following.
  https://github.com/servo/webrender/blob/master/webrender/src/texture_cache.rs#L565

This pull request expose the device_max_size, but it does not expose TextureCache::max_texture_size that actually cause the crash. On my linux pc, device_max_size was 16384, but TextureCache::max_texture_size was reduced to 8192.

sotaroikeda:
gecko side just use  device_max_size(16384) to create Mask Layer, therefore there could be a case that exceed TextureCache::max_texture_size(8192).

sotaroikeda:
gecko limits MaskImage's max size with ContainerState::CreateMaskLayer()
   https://dxr.mozilla.org/mozilla-central/source/layout/painting/FrameLayerBuilder.cpp#6423

It uses  device_max_size(16384) not TextureCache::max_texture_size(8192), then it could hit the assert.

JerryShih:
wow, that might be the cause of the crash. In gecko side, I saw we check the texture max size from layer manager to create the mask layer which is just the device max texture size I export here. Let me check with the test link.

sotaroikeda:
So, I am not sure to assert with TextureCache::max_texture_size in direct texture binding case.

Issue #1403, opened 2017-06-19
Weird ps_border_edge program linker error on Mali-T880 GPU


MortimerGoro:
ps_border_edge program fais to link in Samsung Galaxy S7 (Mali-T880 GPU). GetProgramInfoLog returns a empty string for this link error

```
Failed to link shader program: "ps_border_edge".
ERROR:servo: Failed to link shader program: "ps_border_edge". Error ""
```

It's really weird because after some testing it seems that the link error is triggered when both vColor0 and vColor1 variants are set in the same function (it also happens moving all the write_color function to main()) . If I comment one of the vColor0/vColor1 variant assignment it just links correctly. I tried to rename the variants or to separate the assignment to another small function but it fails to link too:

```
void write_color1(vec4 color, vec2 modulate) {
    vColor0 = vec4(color.rgb * modulate.x, color.a);
}

void write_color2(vec4 color, vec2 modulate) {
    vColor1 = vec4(color.rgb * modulate.y, color.a);
}

void write_color(vec4 color, float style, bool flip) {
    vec2 modulate;

    switch (int(style)) {
        case BORDER_STYLE_GROOVE:
            modulate = flip ? vec2(1.3, 0.7) : vec2(0.7, 1.3);
            break;
        case BORDER_STYLE_RIDGE:
            modulate = flip ? vec2(0.7, 1.3) : vec2(1.3, 0.7);
            break;
        default:
            modulate = vec2(1.0);
            break;
    }

    write_color1(color, modulate);
    write_color2(color, modulate);
}
```

The only combination that I found to work is to duplicate the entire write_color func for each variant. Probably it helps that both funcs are a bit more difficult to inline than the small function attempt mentioned before. It's unfortunate because the switch is computed twice and I don want to affect performance on other platforms to fix a issue on this GPU...

```
void write_color1(vec4 color, float style, bool flip) {
    vec2 modulate;

    switch (int(style)) {
        case BORDER_STYLE_GROOVE:
            modulate = flip ? vec2(1.3, 0.7) : vec2(0.7, 1.3);
            break;
        case BORDER_STYLE_RIDGE:
            modulate = flip ? vec2(0.7, 1.3) : vec2(1.3, 0.7);
            break;
        default:
            modulate = vec2(1.0);
            break;
    }

    vColor0 = vec4(color.rgb * modulate.x, color.a);
}

void write_color2(vec4 color, float style, bool flip) {
    vec2 modulate;

    switch (int(style)) {
[...]

PR #1534, opened 2017-08-01
Work around for Mali-T880 GPU driver program link error

MortimerGoro:
See https://github.com/servo/webrender/issues/1403

About performance: Glenn was ok with this. The vertex shader is taking a lot less time compared to the fragment shader.


leeoniya:
candidate for https://github.com/servo/webrender/wiki/Driver-issues ?

glennw:
@MortimerGoro Yep, please add this to the driver issues wiki link above. Thanks!

PR #1535, opened 2017-08-01
Pass the dirty rect down into the blob image

jrmuizel:
This lets us know which area of the image has changed and is useful
for merging the old blob data and the new blob data.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1535)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -350,7 +350,8 @@ impl ResourceCache {
             if let ImageData::Blob(ref mut blob) = data {
                 self.blob_image_renderer.as_mut().unwrap().update(
                     image_key,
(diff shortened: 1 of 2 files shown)


glennw:
The examples needs to be updated.

```
error[E0050]: method `update` has 3 parameters but the declaration in trait `webrender::<unnamed>::BlobImageRenderer::update` has 4

   --> webrender/examples/blob.rs:145:52

    |

145 |     fn update(&mut self, key: api::ImageKey, cmds: api::BlobImageData) {

    |                                                    ^^^^^^^^^^^^^^^^^^ expected 4 parameters, found 3

```

glennw:
ping @jrmuizel

glennw:
Closing for now due to no activity. Please re-open when ready to go.

PR #1537, opened 2017-08-01
Add a benchmark / stress test that has 8192 image templates.

glennw:
This is a good test case for resource cache / hashing performance.


kvark:
Can we have a script in the repo that generates this test?

glennw:
Do you mean instead of checking this file in, or in addition to having the file checked in?

I did generate it with a python script, so can tidy that up and either add it here or as a follow up.

kvark:
I think it's fine to check it in, but having the generating source would be useful. So yes, let's have this python script somewhere (`/script/`?)

glennw:

PR #1538, opened 2017-08-01
Switch hash table and maps to use FxHash instead of Fnv.

glennw:
Also introduce a FastHashMap and FastHashSet type alias
that is used to allow easily benchmarking different hash
functions, and should be used for all HashMaps and HashSets
that are used by WR.

Some representative benchmarks below showing the total backend
CPU time for various hash types:

Wrench text benchmark
 Sip: 2.06 ms
 Fnv: 1.32 ms
 Fx:  1.12 ms

https://news.ycombinator.com
 Sip: 1.40 ms
 Fnv: 1.02 ms
 Fx:  0.92 ms

https://en.wikipedia.org/wiki/Rust
 Sip: 3.14 ms
 Fnv: 2.72 ms
 Fx:  2.39 ms


glennw:
r? @kvark 

I still want to work on reducing the *amount* of hashing we do, but this is a significant and easy win we can get right now.

webrender/build.rs
```diff
@@ -11,10 +11,10 @@ fn write_shaders(glsl_files: Vec<PathBuf>, shader_file_path: &Path) {
     let mut shader_file = File::create(shader_file_path).unwrap();
 
     write!(shader_file, "/// AUTO GENERATED BY build.rs\n\n").unwrap();
-    write!(shader_file, "use std::collections::HashMap;\n").unwrap();
+    write!(shader_file, "use internal_types::FastHashMap;\n").unwrap();
```
emilio: This could use PHF, btw. Not sure how hot this lookup is, but given they're static keys seems like it could be better.

glennw: They're actually only fetched once - when they need to be compiled (lazily) for the first time. So not a high priority, but something we could consider as a follow up.

emilio:
@bors-servo r-

Err, forgot that you asked for explicit review by @kvark

kvark:
@emilio I reviewed #1540, which includes this PR

glennw:

PR #1539, opened 2017-08-02
Move last access time from separate hash table to inline with resource.

glennw:
This reduces a lot of hash table lookups in the resource cache. It
often results in a total CPU time improvement of ~10-15%, especially
on sites that are text or image heavy.


glennw:

PR #1540, opened 2017-08-02
Use multi-level hash tables for glyph cache.

glennw:
Do one hash lookup per text run to find a glyph key cache, then
fetch each glyph from the glyph cache.

This drastically reduces the size of keys that we typically hash
and lookup, resulting in large performance wins for text heavy sites.

The text-rendering benchmark drops from 0.84ms CPU to 0.38ms CPU
on my laptop.


glennw:
r? @kvark 

This includes https://github.com/servo/webrender/pull/1539, https://github.com/servo/webrender/pull/1538 and https://github.com/servo/webrender/pull/1537.

I can close those off, if you want to review these all as one PR. Alternatively, feel free to review those individually and I can remove those commits from here once they are landed.

glennw:
@jrmuizel The combination of patches in this PR typically drops the total WR CPU time by ~30-50% on real-world pages (those with significant amounts of text).

It'd be interesting to see what effect it has on Gecko profiles once this is all reviewed and landed (although, perhaps not as drastic if the Gecko display lists still contain a lot of stacking contexts).

webrender/src/resource_cache.rs
```diff
@@ -801,33 +804,28 @@ impl ResourceCache {
...
 
         self.cached_images.clear_keys(&mut self.texture_cache, |request| request.key.0 == namespace);
-        self.cached_glyphs.clear_keys(&mut self.texture_cache, |request| request.font.font_key.0 == namespace);
+        self.cached_glyphs.clear_fonts(&mut self.texture_cache, |font| font.font_key.0 == namespace);
     }
 }
 
 pub trait Resource {
-    fn texture_cache_item_id(&self) -> Option<TextureCacheItemId>;
+    fn free(&self, texture_cache: &mut TextureCache);
```
kvark: The reverse in responsibility here (the resource doesn't know what texture cache it belongs to, but it has to know how to delete itself with any?..) is a little concerning. Perhaps, we could store some unique Id of the owning texture cache in the resource, so that we can at least assert at run time if the resource is owned by a cache that it needs to remove from. That would be fine to add in follow-ups, unless you don't see a problem with this internal API.

webrender/src/glyph_cache.rs
```diff
@@ -0,0 +1,86 @@
...
+    where for<'r> F: Fn(&'r &FontInstanceKey) -> bool
+    {
+        let caches_to_destroy = self.glyph_key_caches.keys()
+            .filter(&key_fun)
+            .cloned()
+            .collect::<Vec<_>>();
+        for key in caches_to_destroy {
+            let mut cache = self.glyph_key_caches.remove(&key).unwrap();
+            cache.clear(texture_cache);
+            self.glyph_key_caches.remove(&key);
```
glennw: Oops, fixed.

webrender/src/resource_cache.rs
```diff
@@ -105,50 +103,56 @@ impl ImageTemplates {
...
-    pub fn insert(&mut self, key: K, value: V, frame: FrameId) {
-        self.last_access_times.insert(key.clone(), frame);
+    pub fn insert(&mut self, key: K, value: V) {
         self.resources.insert(key, value);
     }
 
     pub fn entry(&mut self, key: K, frame: FrameId) -> Entry<K,V> {
-        self.last_access_times.insert(key.clone(), frame);
-        self.resources.entry(key)
+        let mut entry = self.resources.entry(key);
```
glennw: The API returns the entry to the caller anyway, so it seems simpler to use the retrieved entry to update the timestamp?

webrender/src/resource_cache.rs
glennw: There is a 1:1 mapping between a texture cache and a resource cache (which contains these items), so I'm not sure there's an issue here. We could definitely consider tidying this up as a follow up if you're still unsure of the API.

glennw:
@kvark Thanks! Fixed the bug and commented on the others.

kvark:
Thank you!
@bors-servo r+

PR #1541, opened 2017-08-02
Remove UBO code from device - we don't intend to use it.

glennw:


kvark:
I'm not completely sure if we'll never use UBO, but we can always resurrect this code if we do.
@bors-servo r+

kvark:
bors-servo retry

PR #1545, opened 2017-08-02
Remove the requested_images and requested_glyphs hash sets.

glennw:
These were used to work out which texture cache items to add
to the GPU cache for this frame.

However, we already run through the cache items to expire old
resources, so we can just check if the resource was accessed
during this frame while doing that, and push those items to
the GPU cache.

This results in some very good performance wins:

many-images benchmark:
 Before: 6.66 ms
 After:  5.32 ms

text-rendering benchmark:
 Before: 0.41 ms
 After:  0.36 ms


glennw:
I think this is the last of the obvious wins in the resource cache hashing code. I have some ideas for more improvements but the hashing no longer shows up near the top of the profile with this and the previous changes...

kvark:
Oh fun:

ICC when building Angle

PR #1546, opened 2017-08-03
Allow flexible vertex formats, remove hard coded formats from device.rs.

glennw:


glennw:
r? @kvark 

This shouldn't have any noticeable changes, however it's a much cleaner implementation. This will make it simpler in the future to do things like a smaller vertex format for glyphs, which should help a lot with CPU usage on real world pages.

glennw:
Rebased

webrender/src/debug_render.rs
```diff
@@ -3,14 +3,69 @@
...
 use internal_types::{ORTHO_NEAR_PLANE, ORTHO_FAR_PLANE, TextureSampler};
-use internal_types::{DebugFontVertex, DebugColorVertex, RenderTargetMode};
+use internal_types::RenderTargetMode;
 use std::f32;
 use api::{ColorU, ImageFormat, DeviceUintSize};
 
+const DESC_FONT: VertexDescriptor = VertexDescriptor {
+    vertex_attributes: &[
+        VertexAttribute { name: "aPosition", count: 2, kind: VertexAttributeKind::F32 },
+        VertexAttribute { name: "aColor", count: 4, kind: VertexAttributeKind::U8 },
```
kvark: We should clearly differentiate between normalized and non-normalized integer semantics. So, I suggest using something like `U8Norm` here.

webrender/src/device.rs
```diff
@@ -144,146 +153,97 @@ pub trait FileWatcherHandler : Send {
...
+impl VertexDescriptor {
+    fn bind(&self,
+            gl: &gl::Gl,
+            main: VBOId,
+            instance: VBOId) {
+        main.bind(gl);
+
+        let vertex_stride = self.vertex_attributes
+                                .iter()
+                                .fold(0, |stride, attr| stride + attr.size_in_bytes());
```
kvark: nit: could be `map(|attr| attr.size_in_bytes()).sum()`

glennw:
Thanks! Comments addressed.

kvark:
Awesome, thanks!
@bors-servo r+

PR #1547, opened 2017-08-03
Remove shaders hashmap from device. Also add shader deinit path.

glennw:
Lots of the surrounding code (particularly in renderer.rs) are
still far from ideal. However, this does remove the hash lookups
when switching shaders, and also ensures all shaders are correctly
deleted when shutting down a renderer.

We can work on incremental improvements to the code in renderer.rs.

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -173,4 +173,6 @@ pub fn main_wrapper(builder_callback: fn(&RenderApi,
         renderer.render(DeviceUintSize::new(width, height));
         window.swap_buffers().ok();
     }
+
+    renderer.deinit();
 }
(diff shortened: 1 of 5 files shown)


kvark:
@glennw sorry, needs a rebase

glennw:
Rebased.

@bors-servo r=kvark

kvark:
>bors-servo retry force

force

staktrace:
@glennw FYI the assertion in Program::drop is failing in gecko. https://treeherder.mozilla.org/logviewer.html#?job_id=121988470&repo=try&lineNumber=1975

staktrace:
@glennw nvm, kvark pointed out we were missing the call to deinit the renderer before dropping it.

PR #1552, opened 2017-08-04
React to memory pressure.

nical:
Memory pressure events are sent when the amount of available memory is so low that we risk crashing very soon if we don't do anything about it. When this happens we don't really care about frame budgets and smoothness anymore, we can only do a best effort at freeing memory and hope to not crash.

This PR adds memory pressure notifications to the API and reacts to it by completely clearing the texture cache and clearing some of the allocations that we try to recycle for performance purposes.

### DIFF
--- a/webrender/src/glyph_cache.rs
+++ b/webrender/src/glyph_cache.rs
@@ -89,6 +89,15 @@ impl GlyphCache {
         }
     }
 
+    pub fn clear(&mut self, texture_cache: &mut TextureCache) {
+        for (_, glyph_key_cache) in &mut self.glyph_key_caches {
+            glyph_key_cache.clear(texture_cache)
+        }
+        // We use this in on_memory_pressure where retaining memory allocations
+        // isn't desirable, so we completely remove the hash map instead of clearing it.
+        self.glyph_key_caches = FastHashMap::default();
+    }
+
     pub fn clear_fonts<F>(&mut self, texture_cache: &mut TextureCache, key_fun: F)
     where for<'r> F: Fn(&'r &FontInstanceKey) -> bool
     {
--- a/webrender/src/internal_types.rs
+++ b/webrender/src/internal_types.rs
@@ -268,6 +268,7 @@ impl RendererFrame {
 pub enum ResultMsg {
     RefreshShader(PathBuf),
     NewFrame(DocumentId, RendererFrame, TextureUpdateList, BackendProfileCounters),
+    UpdateResources { updates: TextureUpdateList, cancel_rendering: bool },
 }
 
 #[derive(Debug, Clone, Copy, Eq, Hash, PartialEq)]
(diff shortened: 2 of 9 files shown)


nical:
This built on top of #1532 so it has some extra commits until that PR is merged.

nical:
Rebased and tested. You can use the 'm' key in the sample and in wrench to trigger memory pressure events.
I reverted the patch that was clearing the frame builder's allocations during memory pressure, because frames would not be rendered until a new frame is built from a new display list which is probably not the behavior one would expect. Fixing this to re-build the frame automatically wouldn't be very hard but it requires to reorganize some more code and I wasn't sure it was worth it at this point. We can revisit that later.

webrender/src/glyph_cache.rs
```diff
@@ -89,6 +90,14 @@ impl GlyphCache {
         }
     }
 
+    pub fn clear(&mut self, texture_cache: &mut TextureCache) {
+        let caches = mem::replace(&mut self.glyph_key_caches, FastHashMap::default());
```
kvark: this could be rewritten cleaner without `mem::replace`, which I consider highly overused in our codebase, e.g.:
```rust
for (_, mut glyph_key_cache) in &mut self.glyph_key_caches {
  glyph_key_cache.clear(texture_cache);
}
self.glyph_key_caches.clear();
```
Bonus points - space is kept reserved.

webrender/src/internal_types.rs
```diff
@@ -268,6 +268,7 @@ impl RendererFrame {
 pub enum ResultMsg {
     RefreshShader(PathBuf),
     NewFrame(DocumentId, RendererFrame, TextureUpdateList, BackendProfileCounters),
+    UpdateResources(TextureUpdateList, bool /* cancel rendering */),
```
kvark: let's just use named fields here

nical:
> What happened to the idea of having the expiration time being a non-linear function of the available memory? It seems like it would ultimately replace the proposed memory pressure mechanics.

I think that we should do both. There are other places where we could free up memory beyond the texture cache (I started with it because it was the obvious one, but we should extend this to other places).

Because of memory fragmentation, it can be hard to tell from the amount of available memory whether we are really in trouble and allocations can start failing (if we are lucky causing memory pressure events) before we detect that we are really in trouble if we only look at the available memory.

I see the idea of adjusting expiration frames in function of the available memory more as a way to keep the memory usage "reasonable" (whatever that means) in general, while memory pressures are a last resort stability thing.

webrender/src/glyph_cache.rs
nical: > Bonus points - space is kept reserved.

Well, this is the one situation where keeping the space reserved is undesirable. But I can just do the loop as you suggest and assign an empty hash map to it instead of clearing if you prefer.

kvark: hah, good point :)

PR #1553, opened 2017-08-04
Expose the namespace id from RenderApi

staktrace:
As a result of the multi-document changes in #1509, it is necessary for
Gecko to stop just inventing its own namespace ids and instead use ones
that match what WR is using internally. Therefore it is necessary for
the RenderApi to expose the namespace id.


kvark:
This is very nice to hear!
@bors-servo r+

PR #1556, opened 2017-08-07
Add debug flag to see the texture cache.

nical:
This PR adds a way to see what's in the texture cache for debugging purposes. In the process of doing this I merged the three debugging options (profiler, render target and texture cache) into a single bitflags and exposed it (previously only the profiler could be toggled after initialization).
In wrench and the sample programs, use "i" to toggle showing the texture cache, "o" to toggle the intermediate render targets, and "p" for the profiler.
Textures are justified to the right of the screen (less likely to overlap with the profiler output) and their sizes are adjusted to fit the width of the screen if they overflow.

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -8,6 +8,7 @@ use std::env;
 use std::path::PathBuf;
 use webrender;
 use webrender::api::*;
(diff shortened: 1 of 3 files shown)


glennw:
This will be super handy for some work I have planned to improve / refactor the texture cache allocator, thanks!

@bors-servo r+

nical:
I didn't know about `enable_batcher`. I think that should be added to the debug flags (as `DISABLE_BATCHER` since in the default non-debug state appears to be `true`).

PR #1558, opened 2017-08-08
Refactor freelist and fix texture updates.

glennw:
Refactor the freelist implementation to be simpler. Specifically,
the handles returned from the freelist are strong handles, that
cannot be cloned or copied. This is prep work for some upcoming
refactoring of the texture cache allocator and implementation.

While doing this, I found some bugs in the texture update APIs
that are now fixed. There were two issues:
 * When reusing an existing texture ID, the texture cache info
   was not being updated in the item cache. This meant the old
   dimensions were being used.
 * When a standalone texture was updated, it wasn't getting
   initialized in the code path where the dimensions are different.

It's a bit difficult to reftest this right now (until we get
raw tests), so I've added an example that checks both texture
update code paths. Neither of those code paths were working
correctly before this change.

### DIFF
--- a/webrender/src/freelist.rs
+++ b/webrender/src/freelist.rs
@@ -2,104 +2,105 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
-pub struct FreeListItemId(u32);
+use std::marker::PhantomData;
+use std::mem;
 
-impl FreeListItemId {
-    #[inline]
-    pub fn new(value: u32) -> FreeListItemId {
-        FreeListItemId(value)
-    }
+// TODO(gw): Add a weak free list handle. This is like a strong
+//           free list handle below, but will contain an epoch
+//           field. Weak handles will use a get_opt style API
+//           which returns an Option<T> instead of T.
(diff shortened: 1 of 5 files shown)


glennw:
@nical I don't think the texture updates with differing dimensions could have worked before this fix. Is it possible we only tested format change during the implementation and not dimensions? I could also be missing something completely as that code is a bit convoluted :)

nical:
The PR was a refresh of an old one that had been more thoroughly tested at the time but I only put together a quick test the second time and it looked like it worked. It is possible I just Ianded on a lucky alignment of things that made it look ok.

webrender/src/freelist.rs
```diff
@@ -2,104 +2,105 @@
...
+#[derive(Debug)]
+pub struct FreeListHandle<T> {
+    index: u32,
+    _marker: PhantomData<T>,
 }
 
-struct FreeListIter<'a, T: 'a> {
-    items: &'a [T],
-    cur_index: Option<FreeListItemId>,
+enum SlotValue<T> {
```
kvark: Any reason we don't just use `Option` instead?

glennw: In the future (when we have weak handles) the ```Free``` enum type will have some information as well (the previous epoch used in that slot).

kvark:
I see, thanks for explanation!
@bors-servo r+

PR #1559, opened 2017-08-08
Fix push_nested_display_list's glyph cache handling (and also the glyph cache parser)

Gankro:
* Cache stores indices, not full GlyphInstances
* push_nested needs to split off the cache and data and handle them seperately

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -9,11 +9,11 @@ use serde::ser::{SerializeSeq, SerializeMap};
 use time::precise_time_ns;
 use {BorderDetails, BorderDisplayItem, BorderWidths, BoxShadowClipMode, BoxShadowDisplayItem};
 use {ClipAndScrollInfo, ClipDisplayItem, ClipId, ColorF, ComplexClipRegion, DisplayItem};
-use {ExtendMode, FilterOp, FontKey, GlyphInstance, GlyphOptions, Gradient, GradientDisplayItem};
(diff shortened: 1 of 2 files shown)


Gankro:
Post-mortem: 

* nothing in our codebase tests push_nested_display_list
* nothing in our codebase currently parses the cache (should be fixed when @glennw tries to use it)

webrender_api/src/display_list.rs
```diff
@@ -289,15 +297,16 @@ impl<'a> BuiltDisplayListIter<'a> {
 }
 
 impl<'a> Iterator for GlyphsIter<'a> {
-    type Item = (FontKey, ColorF, ItemRange<GlyphInstance>);
+    // NOTE: u32 = GlyphInstance.index
```
kvark: let's have a typedef?

webrender_api/src/display_list.rs
```diff
@@ -289,15 +297,16 @@ impl<'a> BuiltDisplayListIter<'a> {
...
+    type Item = (FontKey, ColorF, ItemRange<u32>);
```
kvark: wouldn't it be simpler to have `AuxIter<>` as the last element in this tuple?

glennw:
@Gankro I think bors got confused by the travis mac issues earlier today. I usually do a ```git commit --amend``` and ```git push -f``` to re-trigger the CI. There might be a simpler way, but that usually does the trick.

PR #1561, opened 2017-08-09
Refactor the boilerplate code and update examples.

glennw:
Instead of free functions, there is now an Example trait that
each example must implement. This allows examples to store
state without resorting to globals, mutexes and such things.

The trait has two methods:

 * render - the example must create a display list and resource
            transaction update matching the current application state.

 * on_event - called when a glutin event is received. This can return
              true if the example wants render() to be called again
              after the event is handled.

Previously, it was not possible to modify or rebuild the display list
after the initial display list was created. Now, examples can choose
to rebuild a display list as required.

### DIFF
--- a/webrender/examples/animation.rs
+++ b/webrender/examples/animation.rs
@@ -6,76 +6,83 @@ extern crate gleam;
 extern crate glutin;
 extern crate webrender;
 
-#[macro_use]
-extern crate lazy_static;
-
 #[path="common/boilerplate.rs"]
 mod boilerplate;
 
-use boilerplate::HandyDandyRectBuilder;
-use std::sync::Mutex;
+use boilerplate::{Example, HandyDandyRectBuilder};
 use webrender::api::*;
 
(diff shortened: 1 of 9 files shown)


glennw:
r? @kvark 

The diff stat looks much bad - but it's mostly indentation changes, I think.

I'll probably need to rebase this, or the open PR which has a new example, depending on which lands first. But this should be ready for review now anyway.

glennw:
Oh, the reason for doing this is so I can write samples that create and destroy lots of different resources and display lists, to stress test the texture change changes I am working on.

kvark:
> webrender/examples/image_resize.rs:103:31
>> boilerplate::main_wrapper(body, event_handler, None);

> expected 2 parameters

glennw:
Rebased and updated the new example.

@bors-servo r=kvark

glennw:
I'll fix the borrowing of the app as a follow up :)

PR #1562, opened 2017-08-09
Freetype saving of bytes

### LINKED ISSUE #1560: Possible use after free in FreeType `FontContext::add_raw_font()`
https://github.com/servo/webrender/blob/caf35a2e58aca13e8352802826d41297e478dac4/webrender/src/platform/unix/font.rs#L100

According to the FreeType documentation, `bytes` must not be freed until `FT_Done_Face` is called. How do we guarantee this?

kvark:
Fixes #1560

### DIFF
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -296,7 +296,7 @@ impl FontContext {
     fn add_font(&mut self, font_key: &FontKey, template: &FontTemplate) {
         match template {
             &FontTemplate::Raw(ref bytes, index) => {
-                self.add_raw_font(&font_key, &**bytes, index);
+                self.add_raw_font(&font_key, bytes.clone(), index);
             }
             &FontTemplate::Native(ref native_font_handle) => {
                 self.add_native_font(&font_key, (*native_font_handle).clone());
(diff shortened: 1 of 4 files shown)


webrender/src/platform/unix/font.rs
```diff
@@ -87,17 +90,18 @@ impl FontContext {
     pub fn add_raw_font(&mut self, font_key: &FontKey, bytes: &[u8], index: u32) {
```
nical: Raw fonts are passed around as an `Arc<Vec<u8>>`. I think that you could change add_raw_font to take the Arc nstead of a slice and just store the arc to make sure it lives long enough (and that way avoid the copy).

kvark:
@nical thanks, code is updated now.
Let's see if CI likes it.

webrender/src/glyph_rasterizer.rs
```diff
@@ -296,7 +296,7 @@ impl FontContext {
     fn add_font(&mut self, font_key: &FontKey, template: &FontTemplate) {
         match template {
             &FontTemplate::Raw(ref bytes, index) => {
-                self.add_raw_font(&font_key, &**bytes, index);
+                self.add_raw_font(&font_key, bytes.clone(), index);
```
nical: nit: `Arc::clone(&bytes)` ftw (it's even [documented as the idiomatic way](https://doc.rust-lang.org/std/sync/struct.Arc.html#cloning-references) now.

kvark: ok, I'll fix it if CI breaks ;)

Issue #1260, opened 2017-05-16
assertion failed: requested_size.width <= self.max_texture_size (running wpt tests with broken driver)


MortimerGoro:
Running filtered servo wpt tests such as "`./mach test-wpt --release dom`" or "`./mach test-wpt --release webgl`" crash on my desktop linux machine.

They only crash when --no-pause-after-test is enabled (running test categories or a single test using "`./mach test-wpt --release tests/wpt/web-platform-tests/webgl/conformance-1.0.3/conformance/extensions/oes-vertex-array-object.html --no-pause-after-test`"). Running any single test without --no-pause-after-test works ok.

It seems like something broke in headless mode in a webrender change. I was able to run tests with no problem in the past.

Here is the backtrace, any idea?

```
 0:04.74 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "assertion failed: requested_size.width <= self.max_texture_size (thread main, at /home/mortimer/Projects/extensions/servo/.cargo/git/checkouts/webrender-c3596abe1cf4f320/4ab9b9b/webrender/src/texture_cache.rs:618)"
 0:05.09 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "stack backtrace:"
 0:05.09 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "   0:     0x55e58ff2decc - backtrace::backtrace::trace::haa53f4880ff627c6"
 0:05.09 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "   1:     0x55e58ff2e392 - backtrace::capture::Backtrace::new::h5f3de869c0349b3d"
 0:05.09 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "   2:     0x55e58f06cd15 - servo::main::{{closure}}::hed617678bcb97b95"
 0:05.09 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "   3:     0x55e59100a0ea - std::panicking::rust_panic_with_hook"
 0:05.09 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "                        at /checkout/src/libstd/panicking.rs:550"
 0:05.09 PROCESS_OUTPUT: Thread-TestrunnerManager-1 (pid:9988) "   4:     0x55e58fcf058f - std::panicking::begin_panic::h2cdb5c8f557073e7"
[...]

PR #1563, opened 2017-08-09
Fail initialization if the max texture size is insufficient.

nical:
This works around #1260 ([driver issues entry](https://github.com/servo/webrender/wiki/Driver-issues#1260---broken-gl-context-reports-max-texture-size-equal-to-zero)) by gracefully failing the initialization instead of panicking the first time we try to allocate a texture cache page.
This should let us fallback to another compositor backend and more easily identify the issue when it happens.

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -810,6 +810,7 @@ pub struct Renderer {
 pub enum InitError {
     Shader(ShaderError),
     Thread(std::io::Error),
+    MaxTextureSize,
 }
 
 impl From<ShaderError> for InitError {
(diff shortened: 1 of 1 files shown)


nical:
Moved this sanity check to the beginning of the function to not bother with creating shaders, threadsd and other resources.

r? @kvark

kvark:
Thanks, looking good!
@bors-servo r+

kvark:
@nical y u no compile? :)
all CI is broken

> 847 |             result_tx: result_tx.clone(),
> |                        ^^^^^^^^^ not found in this sco

nical:
oops, did a last minute change and apparently forgot to build it. This version does build locally now.

Issue #1405, opened 2017-06-20
Implement the fallback path for the big size image


JerryShih:
@kvark @glennw @nical @sotaroikeda 
In webrender, we have the limit size of gl texture. Currently, there is an assertion if the image is to large[1]. When we integrate gecko with WR, there are a lot of crash for this assertion.
I'm trying to create a global shared black texture. If we hit the limitation, show the global shared texture instead.
 
[1]
https://github.com/servo/webrender/commit/cb510c5c9a4ce78eb1c02b5bd01af2fc6c5c7510

glennw:
cc @nical 

I thought we handled this with tiling support now?

glennw:
Ah, probably just not for external images, right?

nical:
The plan is to use tiling for this, I thought we already supported tiled external images, but there might be a bit of missing glue to get the two to work together (I vaguely remember looking into this a while back but now I a not sure what came out of it). It should not be hard to do.

nical:
So I had another look and we already have code to support tiling external buffer images the same way we tile regular images. This is done by specifying the proper stride and offset in the texture cache that is passed to `TextureUpdateOp::UpdateForExternalBuffer`. `ResourceCache::should_tile` allows tiling external buffer images under the same conditions as regular images.

@JerryShih Are you referring to other types of external images (like external texture handles) that are bigger than webrender's maximum texture size setting?

JerryShih:
I think the "external buffer" could be used with a large size[1]. But I think this type of buffer could be tiled.

https://dxr.mozilla.org/mozilla-central/rev/e0b0865639cebc1b5afa0268a4b073fcdde0e69c/gfx/layers/composite/TextureHost.cpp#590

nical:
@JerryShih I did some testing and as far as I can tell, external image buffers are properly tiled inside WebRender, with two exceptions: masks and yuv images. Gecko currently doesn't support tiling masks and yuv images either (but maybe that doesn't trigger a fatal assertion).

Could you provide some more details about how you run into this issue (when you have some time, no hurry)?

jrmuizel:
We're seeing crashes like: https://bugzilla.mozilla.org/show_bug.cgi?id=1383731
where we fail this assert: https://hg.mozilla.org/mozilla-central/annotate/52285ea5e54c/gfx/webrender/src/texture_cache.rs#l693

Shouldn't tiling be preventing this from happening?

nical:
[...]

PR #1564, opened 2017-08-09
Handle big size image

JerryShih:
@nical @glennw @kvark @sotaroikeda 
r?

This patch try to handle the #1405.
If we hit the maximum texture size, turn to use a 1x1 rgba(255,255,255,255) dummy texture.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1564)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1058,7 +1058,6 @@ impl PrimitiveStore {
         }
     }
 
-    /// Returns true if the bounding box needs to be updated.
     fn prepare_prim_for_render_inner(
         &mut self,
         prim_index: PrimitiveIndex,
@@ -1326,7 +1325,6 @@ impl PrimitiveStore {
         true
     }
 
-    /// Returns true if the bounding box needs to be updated.
     pub fn prepare_prim_for_render(
         &mut self,
         prim_index: PrimitiveIndex,
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -22,6 +22,7 @@ use internal_types::{FastHashMap, FastHashSet, SourceTexture, TextureUpdateList}
 use profiler::{ResourceProfileCounters, TextureCacheProfileCounters};
 use rayon::ThreadPool;
 use std::collections::hash_map::Entry::{self, Occupied, Vacant};
+use std::cmp;
 use std::fmt::Debug;
 use std::hash::Hash;
 use std::mem;
@@ -113,8 +114,15 @@ struct CachedImageInfo {
     epoch: Epoch,
 }
 
+#[derive(Debug)]
+pub enum ResourceClassCacheError {
+    OverLimitSize,
+}
+
+pub type ResourceCacheResult<V> = Result<V, ResourceClassCacheError>;
+
 pub struct ResourceClassCache<K, V> {
-    resources: FastHashMap<K, V>,
+    resources: FastHashMap<K, ResourceCacheResult<V>>,
 }
 
 impl<K, V> ResourceClassCache<K, V>
@@ -127,21 +135,21 @@ where
         }
     }
 
-    fn get(&self, key: &K) -> &V {
-        self.resources
-            .get(key)
+    fn get(&self, key: &K) -> &ResourceCacheResult<V> {
+        self.resources.get(key)
             .expect("Didn't find a cached resource with that ID!")
     }
 
-    pub fn insert(&mut self, key: K, value: V) {
+    pub fn insert(&mut self, key: K, value: ResourceCacheResult<V>) {
         self.resources.insert(key, value);
     }
 
-    pub fn get_mut(&mut self, key: &K) -> Option<&mut V> {
+    pub fn get_mut(&mut self, key: &K) -> &mut ResourceCacheResult<V> {
         self.resources.get_mut(key)
+            .expect("Didn't find a cached resource with that ID!")
     }
 
-    pub fn entry(&mut self, key: K) -> Entry<K, V> {
+    pub fn entry(&mut self, key: K) -> Entry<K, ResourceCacheResult<V>> {
         self.resources.entry(key)
     }
 
@@ -159,7 +167,7 @@ where
             .cloned()
             .collect::<Vec<_>>();
         for key in resources_to_destroy {
-            self.resources.remove(&key).unwrap();
+            let _ = self.resources.remove(&key).unwrap();
         }
     }
 }
@@ -483,25 +491,38 @@ impl ResourceCache {
                     return;
                 }
 
+                let side_size =
+                    template.tiling.map_or(cmp::max(template.descriptor.width, template.descriptor.height),
+                                           |tile_size| tile_size as u32);
+                if side_size > self.texture_cache.max_texture_size() {
+                    // The image or tiling size is too big for hardware texture size.
+                    warn!("Dropping image, image:(w:{},h:{}, tile:{}) is too big for hardware!",
+                          template.descriptor.width, template.descriptor.height, template.tiling.unwrap_or(0));
+                    self.cached_images.insert(request, Err(ResourceClassCacheError::OverLimitSize));
+                    return;
+                }
+
                 // If this image exists in the texture cache, *and* the epoch
                 // in the cache matches that of the template, then it is
                 // valid to use as-is.
                 let (entry, needs_update) = match self.cached_images.entry(request) {
                     Occupied(entry) => {
-                        let needs_update = entry.get().epoch != template.epoch;
+                        let needs_update = entry.get().as_ref().unwrap().epoch != template.epoch;
                         (entry.into_mut(), needs_update)
                     }
                     Vacant(entry) => (
-                        entry.insert(CachedImageInfo {
-                            epoch: template.epoch,
-                            texture_cache_handle: TextureCacheHandle::new(),
-                        }),
+                        entry.insert(Ok(
+                            CachedImageInfo {
+                                epoch: template.epoch,
+                                texture_cache_handle: TextureCacheHandle::new(),
+                            }
+                        )),
                         true,
                     ),
                 };
 
                 let needs_upload = self.texture_cache
-                    .request(&mut entry.texture_cache_handle, gpu_cache);
+                    .request(&mut entry.as_mut().unwrap().texture_cache_handle, gpu_cache);
 
                 if !needs_upload && !needs_update {
                     return;
@@ -600,7 +621,7 @@ impl ResourceCache {
         debug_assert!(fetch_buffer.is_empty());
 
         for (loop_index, key) in glyph_keys.iter().enumerate() {
-            if let Some(ref glyph) = *glyph_key_cache.get(key) {
+            if let Ok(Some(ref glyph)) = *glyph_key_cache.get(key) {
                 let cache_item = self.texture_cache.get(&glyph.texture_cache_handle);
                 if current_texture_id != cache_item.texture_id {
                     if !fetch_buffer.is_empty() {
@@ -648,15 +669,24 @@ impl ResourceCache {
         image_key: ImageKey,
         image_rendering: ImageRendering,
         tile: Option<TileOffset>,
-    ) -> CacheItem {
+    ) -> Result<CacheItem, ()> {
         debug_assert_eq!(self.state, State::QueryResources);
         let key = ImageRequest {
             key: image_key,
             rendering: image_rendering,
             tile,
         };
-        let image_info = &self.cached_images.get(&key);
-        self.texture_cache.get(&image_info.texture_cache_handle)
+
+        // TODO(Jerry): add a debug option to visualize the corresponding area for
+        // the Err() case of CacheItem.
+        match *self.cached_images.get(&key) {
+          Ok(ref image_info) => {
+              Ok(self.texture_cache.get(&image_info.texture_cache_handle))
+          }
+          Err(_) => {
+              Err(())
+          }
+        }
     }
 
     pub fn get_image_properties(&self, image_key: ImageKey) -> Option<ImageProperties> {
@@ -816,7 +846,7 @@ impl ResourceCache {
                 image_template.descriptor.clone()
             };
 
-            let entry = self.cached_images.get_mut(&request).unwrap();
+            let entry = self.cached_images.get_mut(&request).as_mut().unwrap();
             self.texture_cache.update(
                 &mut entry.texture_cache_handle,
                 descriptor,
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -482,6 +482,7 @@ impl AlphaRenderItem {
                         );
 
                         if color_texture_id == SourceTexture::Invalid {
+                            warn!("Warnings: skip a PrimitiveKind::Image at {:?}.\n", item_bounding_rect);
                             return;
                         }
 
@@ -540,6 +541,8 @@ impl AlphaRenderItem {
                             glyph_fetch_buffer,
                             gpu_cache,
                             |texture_id, glyphs| {
+                                debug_assert_ne!(texture_id, SourceTexture::Invalid);
+
                                 let textures = BatchTextures {
                                     colors: [
                                         texture_id,
@@ -633,6 +636,7 @@ impl AlphaRenderItem {
                             );
 
                             if texture == SourceTexture::Invalid {
+                                warn!("Warnings: skip a PrimitiveKind::YuvImage at {:?}.\n", item_bounding_rect);
                                 return;
                             }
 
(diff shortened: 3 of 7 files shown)


glennw:
I'm working on a large refactoring of the texture cache at the moment - if possible, I'd like to hold off on merging this until that's done. Is that going to be OK? (I would expect it to be done early next week). Handling this case should be *much* simpler with the changes I'm working on, too.

JerryShih:
@glennw 
That's fine. If the refactor is done, please show the note in this thread.

glennw:
@JerryShih The texture cache refactor has landed now. Instead of trying to use a dummy texture for this case, could we just not add the item to any batches when we find an invalid texture?

JerryShih:
> @JerryShih The texture cache refactor has landed now. Instead of trying to use a dummy texture for this case, could we just not add the item to any batches when we find an invalid texture?

@glennw 
That could be another approach. I think I will still use a special id for the big size allocation(or failed allocation). Then, we could skip the item with that id in batch list or draw with a special color texture for debugging(with a 1x1 black dummy texture).

glennw:
I think the simplest approach will be to early exit from ```request_image``` in resource cache, before even getting to texture cache. Then, ```get_cached_image``` could just return ```SourceTexture::Invalid```. But let's see how this looks after rebasing to the new texture cache and decide from there :)

JerryShih:
@glennw r?

And I need the help for the wrench reftest. I got the different result from the standalone yaml replay and the reftest result.

standalone:
I could see a black Rect. That's what I want.
`cargo run show reftests/image/very-big-tile-size.yaml`

reftest:
I got a gray Rect.
`./headless.py reftest reftests/image/very-big-tile-size.yaml`

I don't know the difference between them.

JerryShih:
@glennw r? again.
rebase.
I still hit the reftest problem with wrench. Maybe you have some hints for this problem.

glennw:
@JerryShih I'm not really sure why it would be different between standalone and reftest. Perhaps the wrong texture is getting bound or something like that?

I wrote up a prototype of what I was discussing above here https://github.com/glennw/webrender/commit/9ca1214d73e658499e599e73ec200668ad6d17fc

I think if we use a solution like this, where it doesn't involve the texture cache at all, the code is quite a bit simpler. And when we want to display the bad texture, we could do that as a follow up in the renderer (handle the case where a texture resolves to SourceTexture::Invalid during batching). What do you think?

JerryShih:
> @JerryShih I'm not really sure why it would be different between standalone and reftest. Perhaps the wrong texture is getting bound or something like that?
> 
> I wrote up a prototype of what I was discussing above here glennw/webrender@9ca1214
> 
> I think if we use a solution like this, where it doesn't involve the texture cache at all, the code is quite a bit simpler. And when we want to display the bad texture, we could do that as a follow up in the renderer (handle the case where a texture resolves to SourceTexture::Invalid during batching). What do you think?

Your prototype is much simpler than me. So, yours is a better solution. I add a comment in glennw/webrender@9ca1214. I will create another follow up if we need to show the skipped image.

glennw:
@JerryShih Sounds good, thanks! Do you want me to follow up your comments and open a PR, or should I leave that for you?

JerryShih:
@glennw 
I would like to do that!

JerryShih:
@glennw has a better solution. Close this pr.

JerryShih:
@glennw @kvark r? again.
This version just skip the image request and return a invalid CacheItem for the big size image request.

webrender/src/resource_cache.rs
```diff
@@ -450,6 +460,23 @@ impl ResourceCache {
                     return;
                 }
 
+                if let Some(tile_size) = template.tiling {
+                    // If the tiling size is too big for hardware texture size then
+                    // just early out and drop the image.
+                    if tile_size as u32 > self.texture_cache.max_texture_size() {
+                        warn!("Dropping image, tile size:{} is too big for hardware!", tile_size);
+                        return;
```
kvark: I'm a bit uncomfortable with these `return` statements as they are different from other return code paths from this method in that they are failures, while others are successes.
How about we return true/false (at least), and then also modify `prepare_prim_for_render` to return `Option<Metadata>` so that a primitive that failed to prepare can be discarded in time?

webrender/src/resource_cache.rs
```diff
@@ -546,13 +573,14 @@ impl ResourceCache {
...
 
         for (loop_index, key) in glyph_keys.iter().enumerate() {
-            let glyph = glyph_key_cache.get(key);
-            let cache_item = glyph.as_ref().map(|info| self.texture_cache.get(&info.texture_cache_handle));
-            if let Some(cache_item) = cache_item {
-                f(loop_index, &cache_item.uv_rect_handle);
-                debug_assert!(texture_id == None ||
-                              texture_id == Some(cache_item.texture_id));
-                texture_id = Some(cache_item.texture_id);
+            if let Some(glyph) = glyph_key_cache.get(key) {
```
kvark: We need to handle the case of failure here. Does it make sense to return as usual if some of the glyphs failed to be obtained?

webrender/src/resource_cache.rs
```diff
@@ -588,7 +616,15 @@ impl ResourceCache {
...
         };
         let image_info = &self.cached_images.get(&key);
-        self.texture_cache.get(&image_info.texture_cache_handle)
+
+        // If an image is not in cached_images, just return an invalid CacheItem
+        // for that request. That invalid item will be skipped during the rendering.
+        //
+        // TODO(Jerry): add a debug option to fill the corresponding area for
+        // that invalid CacheItem.
+        image_info.map_or(CacheItem::invalid(), |image_info| {
```
kvark: we call `get_glyphs` when populating batches. Wouldn't it be simpler to just return an error here, so that a corresponding instance can be discarded?

webrender/src/resource_cache.rs
JerryShih: @kvark 
I'm not familiar with the frame_builder::build process. The prepare_prim_for_render() is called by frame_builder::handle_primitive_run()[1]. And the handle_primitive_run() looks like to handle the mask and clip related thing. Are you thinking about skipping the render-task creation in handle_primitive_run() for the case of None Option<Metadata> return value? 

[1]
https://dxr.mozilla.org/mozilla-central/rev/c959327c6b75cd4930a6ea087583c38b805e7524/gfx/webrender/src/frame_builder.rs#1996

kvark: Yes

glennw:
@JerryShih @kvark What's the status of this?

JerryShih:
@glennw I'm check an animation artifact problem with WR. I will go back to finish this soon.

webrender/src/tiling.rs
```diff
@@ -813,6 +817,11 @@ impl ClipBatcher {
                     ClipSource::Image(ref mask) => {
                         let cache_item =
                             resource_cache.get_cached_image(mask.image, ImageRendering::Auto, None);
+                        // Skip the invalid image mask.
+                        if cache_item.texture_id == SourceTexture::Invalid {
```
JerryShih: This image mask will be rendered at [1]. So, I think we should skip this image if the texture id is not valid.

[1]
https://github.com/servo/webrender/blob/e11297a240b29bd57154d37ecb402d0b9aa299e6/webrender/src/renderer.rs#L2912

webrender/src/resource_cache.rs
JerryShih: We call get_glyphs at [1]. If there is any error inside get_glyphs()[2], the lambda expression at [3] will not be called(which will push the instance to draw list).

[1]
https://github.com/servo/webrender/blob/e11297a240b29bd57154d37ecb402d0b9aa299e6/webrender/src/tiling.rs#L513
[2]
https://github.com/servo/webrender/pull/1564/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R630
[3]
https://github.com/servo/webrender/blob/e11297a240b29bd57154d37ecb402d0b9aa299e6/webrender/src/tiling.rs#L518

webrender/src/resource_cache.rs
JerryShih: The prepare_prim_for_render() is changed. The "Metadata" can't be remove now.

From the first comment:

> I'm a bit uncomfortable with these return statements as they are different from other return code paths from this method in that they are failures, while others are successes.

We still have a case with a failure at [1] in request_image(). So, the new return statements are not just the new failed case. If they are still not looked good, I need to check the  prepare_prim_for_render() for the "Metadata" removing.

[1]
https://github.com/servo/webrender/blob/e11297a240b29bd57154d37ecb402d0b9aa299e6/webrender/src/resource_cache.rs#L553

JerryShih:
@glennw @kvark r?

webrender/src/resource_cache.rs
```diff
@@ -127,10 +136,11 @@ where
         }
     }
 
-    fn get(&self, key: &K) -> &V {
-        self.resources
-            .get(key)
-            .expect("Didn't find a cached resource with that ID!")
+    /// In some cases, we can't handle the resource request(e.g. the user
```
kvark: I think we should still differentiate between keys that didn't end up in the map for some other logical reasons, and keys that were just failing to be added. Perhaps, the resources map should contain `Result<V>` instead of `V`, so that we can handle those failures in a more explicit way?

webrender/src/prim_store.rs
```diff
@@ -1057,7 +1057,6 @@ impl PrimitiveStore {
         }
     }
 
-    /// Returns true if the bounding box needs to be updated.
```
JerryShih: This comment is not correct now.

webrender/src/resource_cache.rs
```diff
@@ -113,8 +122,14 @@ struct CachedImageInfo {
...
 
+pub enum ResourceClassCacheError {
+    OverLimitSize,
+}
+
+pub type ResourceCacheCacheResult<V> = Result<V, ResourceClassCacheError>;
+
 pub struct ResourceClassCache<K, V> {
-    resources: FastHashMap<K, V>,
+    resources: FastHashMap<K, ResourceCacheCacheResult<V>>,
```
JerryShih: Use Result<V> instead of V.

webrender/src/resource_cache.rs
```diff
@@ -483,25 +498,46 @@ impl ResourceCache {
                     return;
                 }
 
+                if let Some(tile_size) = template.tiling {
+                    // If the tiling size is too big for hardware texture size then
+                    // just early out and drop the image.
+                    if tile_size as u32 > self.texture_cache.max_texture_size() {
+                        warn!("Dropping image, tile size:{} is too big for hardware!", tile_size);
+                        self.cached_images.insert(request, Err(ResourceClassCacheError::OverLimitSize));
```
JerryShih: If we hit the size limitation, just insert an Err() entry.

JerryShih:
@kvark r?
Try to use Result<T> in ResourceClassCache.

JerryShih:
I haven't handled the Err() case for each get/get_mut() call, but I will just add some logs for these paths if we need to have these logs.

webrender/src/resource_cache.rs
```diff
@@ -113,8 +122,14 @@ struct CachedImageInfo {
     epoch: Epoch,
 }
 
+pub enum ResourceClassCacheError {
+    OverLimitSize,
+}
+
+pub type ResourceCacheCacheResult<V> = Result<V, ResourceClassCacheError>;
```
kvark: perhaps you meant `ResourceClassCacheResult`?

webrender/src/resource_cache.rs
```diff
@@ -159,7 +174,7 @@ where
             .cloned()
             .collect::<Vec<_>>();
         for key in resources_to_destroy {
-            self.resources.remove(&key).unwrap();
```
kvark: why is this removed? I assume we get a warning about unused `Result` or `Option`

webrender/src/resource_cache.rs
```diff
@@ -483,25 +498,46 @@ impl ResourceCache {
...
+                    // If the tiling size is too big for hardware texture size then
+                    // just early out and drop the image.
+                    if tile_size as u32 > self.texture_cache.max_texture_size() {
+                        warn!("Dropping image, tile size:{} is too big for hardware!", tile_size);
+                        self.cached_images.insert(request, Err(ResourceClassCacheError::OverLimitSize));
+                        return;
+                    }
+                } else {
+                    // The image is too big for hardware texture size.
+                    if template.descriptor.width > self.texture_cache.max_texture_size() ||
```
kvark: should probably merge this block with the one above, using `let side_size = image.tiling.unwrap_or(cmp::max(template.descriptor.width, template.descriptor.height));`

webrender/src/resource_cache.rs
```diff
@@ -483,25 +498,46 @@ impl ResourceCache {
...
+                    }
+                }
+
                 // If this image exists in the texture cache, *and* the epoch
                 // in the cache matches that of the template, then it is
                 // valid to use as-is.
                 let (entry, needs_update) = match self.cached_images.entry(request) {
                     Occupied(entry) => {
-                        let needs_update = entry.get().epoch != template.epoch;
+                        let needs_update = entry.get().as_ref().ok().unwrap().epoch != template.epoch;
```
kvark: no need for `ok()` here

webrender/src/resource_cache.rs
```diff
@@ -483,25 +498,46 @@ impl ResourceCache {
...
+                                texture_cache_handle: TextureCacheHandle::new(),
+                            }
+                        )),
                         true,
                     ),
                 };
 
                 let needs_upload = self.texture_cache
-                    .request(&mut entry.texture_cache_handle, gpu_cache);
+                    .request(&mut entry.as_mut().ok().unwrap().texture_cache_handle, gpu_cache);
```
kvark: same here - no need for `ok()`

webrender/src/resource_cache.rs
```diff
@@ -656,7 +692,15 @@ impl ResourceCache {
             tile,
         };
         let image_info = &self.cached_images.get(&key);
```
kvark: would probably be cleaner with a simple `match` statement

webrender/src/resource_cache.rs
```diff
@@ -816,7 +860,7 @@ impl ResourceCache {
                 image_template.descriptor.clone()
             };
 
-            let entry = self.cached_images.get_mut(&request).unwrap();
+            let entry = self.cached_images.get_mut(&request).as_mut().ok().unwrap();
```
kvark: no need for `ok()`

webrender/src/tiling.rs
```diff
@@ -516,6 +517,8 @@ impl AlphaRenderItem {
                             glyph_fetch_buffer,
                             gpu_cache,
                             |texture_id, glyphs| {
+                                debug_assert!(texture_id != SourceTexture::Invalid);
```
kvark: nit: `debug_assert_ne!()`

webrender/src/resource_cache.rs
```diff
@@ -656,7 +692,15 @@ impl ResourceCache {
...
         };
         let image_info = &self.cached_images.get(&key);
-        self.texture_cache.get(&image_info.texture_cache_handle)
+
+        // If an image is not in cached_images, just return an invalid CacheItem
+        // for that request. That invalid item will be skipped during the rendering.
+        //
+        // TODO(Jerry): add a debug option to fill the corresponding area for
+        // that invalid CacheItem.
+        image_info.as_ref().ok().map_or(CacheItem::invalid(), |image_info| {
```
kvark: let's just return `Result<CacheItem, _>` from this function and match it on the other side, there is no need for nullable `CacheItem` semantic (that `invalid()` does here)

webrender/src/resource_cache.rs
```diff
@@ -174,7 +167,7 @@ where
             .cloned()
             .collect::<Vec<_>>();
         for key in resources_to_destroy {
-            self.resources.remove(&key);
+            self.resources.remove(&key).unwrap();
```
JerryShih: I revert the change, but the CI test is failed.

error: unused `std::result::Result` which must be used
   --> src/resource_cache.rs:170:13
    |
170 |             self.resources.remove(&key).unwrap();
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D unused-must-use` implied by `-D warnings`

JerryShih:
@kvark r? for https://github.com/servo/webrender/pull/1564/commits/d68cf51bd7d9065e24688d15b772ea5fd58ebeab

kvark:
Thank you for addressing my comments!
 >self.resources.remove(&key).unwrap();

Let's do `let _ = self.resources.remove(&key).unwrap();`

Please squash the commits together, and we can merge once CI is happy.

JerryShih:
@kvark 
Update for CI "self.resources.remove(&key).unwrap();" failed, and squash the patch set.

kvark:
@JerryShih awesome, :shipit: ! 👍 
@bors-servo r+

PR #1565, opened 2017-08-09
Introduce texture_allocator.rs, and use it for render target allocation.

glennw:
I'm working on a new allocator for the texture cache, that is better
suited to a dynamic glyph / image cache.

However, the new allocator is not suitable for the needs of the
render task tree, which currently shares the same allocator.

This patch copies the TexturePage allocator implementation from
the texture cache, into texture_allocator.rs, renames it to
guillotine allocator, and switches the render task tree to use it.

It is an exact copy of the code, but with unnecessary features
removed (e.g. all the freeing and coalescing support).

This means we have a bit of code duplication temporarily, but
allows changing the texture cache allocator without affecting the
allocator used by the render task tree.

In the future, we will likely want to tune / modify the allocator
implementation for render tasks as well.

### DIFF
--- a/webrender/src/lib.rs
+++ b/webrender/src/lib.rs
@@ -75,6 +75,7 @@ mod render_task;
 mod resource_cache;
 mod scene;
 mod spring;
+mod texture_allocator;
 mod texture_cache;
 mod tiling;
 mod util;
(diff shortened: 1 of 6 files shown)


glennw:
r? @kvark 

I'm trying to land parts of the texture cache refactor I'm working on as individual PRs, to simplify reviews. This is mostly just a rename and copy of existing code, so the allocator implementation itself here doesn't need review.

kvark:
Thanks for splitting the work!
@bors-servo r+

PR #1566, opened 2017-08-10
Export texture cache size limit

JerryShih:
In TextureCache, there is another formula to calculate the max texture size. So, we should use that value instead of the device max texture size.

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -868,7 +868,7 @@ impl Renderer {
             println!("Device reporting insufficient max texture size ({})", device_max_size);
             return Err(InitError::MaxTextureSize);
         }
-        let max_texture_size = cmp::max(
+        let max_device_size = cmp::max(
             cmp::min(device_max_size, options.max_texture_size.unwrap_or(device_max_size)),
             min_texture_size
         );
@@ -1131,7 +1131,9 @@ impl Renderer {
                                      options.precache_shaders)
         };
 
-        let texture_cache = TextureCache::new(max_texture_size);
+        let texture_cache = TextureCache::new(max_device_size);
+        let max_texture_size = texture_cache.max_texture_size();
+
         let backend_profile_counters = BackendProfileCounters::new();
 
         let dummy_cache_texture_id = device.create_texture_ids(1, TextureTarget::Array)[0];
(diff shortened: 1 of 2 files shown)


JerryShih:
@kvark @glennw r?
@sotaroikeda @nical

webrender/src/renderer.rs
```diff
@@ -1131,7 +1117,22 @@ impl Renderer {
...
+        // gracefully fail now than panic as soon as a texture is allocated.
+        let min_texture_size = 512;
+        if device_max_size < min_texture_size {
+            println!("Device reporting insufficient max texture size ({})", device_max_size);
+            return Err(InitError::MaxTextureSize);
+        }
+        let mut max_texture_size = cmp::max(
+            cmp::min(device_max_size, options.max_texture_size.unwrap_or(device_max_size)),
+            min_texture_size
+        );
```
nical: nit: I would prefer that this block stays towards the top of this function because it can fail sometimes with really bad drivers and i'd rather exit before trying building all of the resources if that happens.

Maybe we could move it into a static `TextureCache::compute_max_texture_size(&Device) -> u32` function so that all of the constants and logic to determine the max size is in one place.

JerryShih: Thanks, I move that block back.

The TextureCache is passed to render_backend. Maybe we can't always get a device when we try to call TextureCache::compute_max_texture_size().

nical:
> The TextureCache is passed to render_backend. Maybe we can't always get a device when we try to call TextureCache::compute_max_texture_size().

I meant a static function just to compute the texture size during initialization. We don't need to access the texture cache itself to compute the max texture size, just some global constants that are defined in texture_cache.rs (I am just proposing this so as to put the logic in the file that has these constants but it's not of prime importance). This static method would be used when initializing the renderer and we would store the result in the texture cache just like we currently do (the texture cache object would not call the static method since as you pointed out it doesn't have access to the device).

kvark:
I agree with @nical that having a static function would be cleaner. In the meantime, this PR is a step in the right direction, and it doesn't break anything.
@bors-servo r+

JerryShih:
I need this PR asap. So, I will try to have another PR for that static function.

JerryShih:
@glennw r?

PR #1569, opened 2017-08-10
tune wavy line style

Gankro:
Browsers generally give wavy lines a fixed "slope" so that the period is a linear function of the the height. In addition they tend to prefer thinner lines (in fact gecko disables AA until font-size is  > 50px, which is larger than anything in decorations-suite.yaml).


Gankro:
I left in the TODO since we probably want to tune it more eventually. This is a fairly coarse "make it look reasonable" pass.

PR #1570, opened 2017-08-10
Add a texture cache stress test playground.

glennw:
Also add support to examples for specifying an external image handler.


glennw:
r? @kvark 

Another isolated part of the texture cache work we can merge separately.

This just contains a range of scenarios that test out texture cache allocation patterns (such as deleting and updating images, creating huge amounts of image keys etc).

kvark:

ugh..
@bors-servo r+

PR #1572, opened 2017-08-14
Refactor and improve the texture cache, implement new allocator.

glennw:
Refactor and improve the texture cache implementation.

* All textures using the sColor* samplers are now array textures. This will allow us to simplify the large image tiling code in the future.
* All texture updates are performed via PBO, to eliminate CPU driver stalls when updating textures.

* The texture cache now acts more like a traditional cache, where the cache is a fixed size, and can evict items as required.
* The allocator is switched to a slab style allocator, allowing deterministic behavior when allocating and freeing.
* The slab allocator does not suffer from fragmentation like the previous allocator, and has no coalesce requirements.
* The eviction policy is a lot smarter than the previous implementation. Items are evicted from the cache as required.
* Since the shared cache is a fixed / bounded size, any allocations that are too large for the cache are created as standalone textures.
* In the rare / edge case of the fixed sized cache being full, and the cache not being able to evict items, these items also fallback to the standalone texture allocator.

* Add support for array textures that don't have FBOs.
* Simplify render targets (always array textures).
* Remove resize_texture which is now unused.
* Remove support for texture upload without PBO.
* Add support for weak handles to freelist.
* Update gleam to 0.4.8 (allows upload of array textures via PBO).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1572)


glennw:
r? @kvark

I'm *so* sorry for the size of this patch. It really snowballed. Let me know if you want to discuss details on IRC / Vidyo etc.

I'm happy with the end result though :)

I've done a fair amount of manual testing in both Servo and Gecko. It also passes the Gecko sanity reftests, the Servo WPT/CSS tests and the Wrench reftests.

glennw:
@staktrace @sotaroikeda @JerryShih  I've run the gecko sanity tests locally with this, and they pass. It doesn't change any external interfaces, but it does expect that external textures which use the normal texture2D sampler switch over to place them into the texture2DArray sampler. It doesn't affect anything that goes via the TextureRect or TextureExternal image types. Do you foresee any issues with this? I'll ping you later during the week with some questions I have on running a try build with this change.

glennw:
@kvark PS: The ```texture_cache.rs``` file is different enough now that it probably is easier to just review it as a new file rather than trying to diff it...

glennw:
To get a feel for how the cache works, the image below may help. This is what the cache looks like after browsing a heap of wikipedia and mozilla pages in Servo.

cache

You can see that within each layer of a texture array, there are number of "regions". Each region acts as a slab allocator with a fixed size, that allows fast and deterministic allocating and freeing.

We'll probably want to tweak the slab sizes, but the initial implementation seems to work well on all the pages I've tried so far.

kvark:
Overall, I think moving to slab allocation is a great step forward! The amount of code is make is rather difficult to get a full picture even when using Reviewable, but I got a general idea, and I like it (despite the fact I was cleaning up the coalescing not so long ago) ;)

> In the rare / edge case of the fixed sized cache being full, and the cache not being able to evict items, these items also fallback to the standalone texture allocator.

Hmm, not sure about this one. I feel like we should not be hard-coding the texture cache size. If there is a workload that needs a little more than we have, we'll see consistent drop in performance there, where optimally I'd like our performance curve to be smooth.

So, instead of allocating a standalone texture, can we just allocate a new texture page to be added to the cache (and reused later on)?

> To get a feel for how the cache works, the image below may help.

I love those, thank you! Although, perhaps it would be visually more appealing to render them Y-flipped, so that all the textures look exactly as they are on screen.

---

Reviewed 32 of 34 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[webrender/res/debug_font.fs.glsl, line 10 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWmyRlhB0yG0mjqeYv:-KrWmyRmDkKWVoIY3fmP:b-iidb5) ([raw file](https://github.com/servo/webrender/blob/9e1a85e8226d7746e82902f44467293d1a9938ca/webrender/res/debug_font.fs.glsl#L10)):*
> ```GLSL
> void main(void)
> {
> #ifdef SERVO_ES2
> ```

is that check intended for removal?

---

*[webrender/src/device.rs, line 1644 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWnldvun6JxBA6GUjN:-KrWnldvun6JxBA6GUjO:b-ol5t2f) ([raw file](https://github.com/servo/webrender/blob/9e1a85e8226d7746e82902f44467293d1a9938ca/webrender/src/device.rs#L1644)):*
> ```Rust
>         let (gl_format, bpp, data, data_type) = match self.textures.get(&texture_id).unwrap().format {
>             ImageFormat::A8 => {
>                 if cfg!(any(target_arch="arm", target_arch="aarch64")) {
> ```

is this path intended for removal as well?

---

*[webrender/src/freelist.rs, line 94 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWoIpOc8j3Zg8YlU26:-KrWoIpOc8j3Zg8YlU27:bv4a7pa) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/freelist.rs#L94)):*
> ```Rust
>     }
> 
>     pub fn upsert(&mut self,
> ```

please add a comment explaining what `upsert` means in this context

---

*[webrender/src/glyph_cache.rs, line 14 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWogoFMYvVtwVqoMbc:-KrWogoG5rmXsd_QeyMf:bxjbgjp) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_cache.rs#L14)):*
> ```Rust
>     pub texture_cache_handle: TextureCacheHandle,
>     pub glyph_bytes: Arc<Vec<u8>>,
>     pub width: u32,
> ```

should it use `euclid` primitives, perhaps? I.e. `TypedSize<u32>` and `TypedPoint2D<f32>`

---

*[webrender/src/glyph_rasterizer.rs, line 161 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWp5-4Jsx3hNinhPMJ:-KrWp5-4Jsx3hNinhPMK:b-i7rtmz) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_rasterizer.rs#L161)):*
> ```Rust
>             match glyph_key_cache.entry(key.clone()) {
>                 Entry::Occupied(mut entry) => {
>                     if let &mut Some(ref mut glyph_info) = entry.get_mut() {
> ```

nit: probably easier to have `if let Some(...) = *entry.get_mut()`

---

*[webrender/src/glyph_rasterizer.rs, line 273 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWq-bbP3K1cF5K4O4X:-KrWq-bcBdaVH6OHd8hJ:b-wjf6dg) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_rasterizer.rs#L273)):*
> ```Rust
>                     let glyph_bytes = Arc::new(glyph.bytes);
>                     let mut texture_cache_handle = TextureCacheHandle::new();
>                     texture_cache.request(&mut texture_cache_handle, gpu_cache);
> ```

shouldn't we check for the `request` return value?

---

*[webrender/src/glyph_rasterizer.rs, line 274 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWq2sEXv_OllvROrs_:-KrWq2sEXv_OllvROrsa:b-idrck8) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_rasterizer.rs#L274)):*
> ```Rust
>                     let mut texture_cache_handle = TextureCacheHandle::new();
>                     texture_cache.request(&mut texture_cache_handle, gpu_cache);
>                     texture_cache.update(
> ```

is update always following `request`?

---

*[webrender/src/internal_types.rs, line 100 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWqGqBbpqVMdD-htAd:-KrWqGqBbpqVMdD-htAe:byysel6) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/internal_types.rs#L100)):*
> ```Rust
> 
> #[derive(Copy, Clone, Debug, PartialEq)]
> pub enum RenderTargetMode {
> ```

I feel like we'll eventually (not now!) need to have a separate `TargetView` type (instead of this enumeration) that would capture the internal FBOs (per slice) and the depth attachment.

---

*[webrender/src/renderer.rs, line 2417 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWs_1aUHgVQ6X0_zH2:-KrWs_1bzwMa5wq0mcnz:b-74jypq) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/renderer.rs#L2417)):*
> ```Rust
>         let mut size = 512;
>         let fb_width = framebuffer_size.width as i32;
>         let num_textures: i32 = self.cache_texture_id_map
> ```

shouldn't this be `num_layers`?

---

*[webrender/src/resource_cache.rs, line 108 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWskewDCQvtoPQomGf:-KrWskewDCQvtoPQomGg:b-y1xw8f) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/resource_cache.rs#L108)):*
> ```Rust
> }
> 
> pub struct ResourceClassCache<K,V> {
> ```

is there any point in this struct now as oppose to just typedeffing to `FastHashMap`?

---

*[webrender/src/resource_cache.rs, line 144 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWswXxYZclKyXjKC51:-KrWswXxYZclKyXjKC52:bsqpkgk) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/resource_cache.rs#L144)):*
> ```Rust
>     where for<'r> F: Fn(&'r &K) -> bool
>     {
>         let resources_to_destroy = self.resources.keys()
> ```

I think it can be easily done on the caller site now that we have Rust-1.19 (used by Gecko) and all those pretty [retain](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.retain) methods in libstd

---

*[webrender/src/texture_cache.rs, line 82 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWu08fv2q7ixBq2Ruo:-KrWu08gJBeRoHR7FWPU:bkzhc4u) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L82)):*
> ```Rust
> // cache or a standalone texture.
> #[derive(Debug)]
> struct CacheEntry {
> ```

nice struct!

---

*[webrender/src/texture_cache.rs, line 231 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWuuUk6JbB7B2iAFgM:-KrWuuUlvDylh3tYVRag:beuoml3) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L231)):*
> ```Rust
> 
>     // Request an item in the texture cache. All images that will
>     // be used on a frame *must* have request() called on their
> ```

do we have any detection of a forgotten `request`?

---

*[webrender/src/texture_cache.rs, line 404 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWwcqajWa-ErxJHXCJ:-KrWwcqbsibq5zJXnLvK:b-w5hj4w) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L404)):*
> ```Rust
> 
>         // Sort by access time so we remove the oldest ones first.
>         eviction_candidates.sort_by_key(|handle| {
> ```

I think this procedure can be done simpler and a bit faster without temporary arrays:
  - sort `standalone_entry_handles` in descending order of `last_access`. Given that it's likely to be temporarily coherent, a quicksort-like method would be efficient at skipping the work on already sorted parts
  - check 32th from the back for early return
  - pop 32 items, calling `free()` for each

Although, this doesn't seem to be a performance critical place here, so we can look into it later. The current code is correct.

---

*[webrender/src/texture_cache.rs, line 477 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWx_Ps_8-X48r4IYWX:-KrWx_PtGPv_VSDpEw6-:bgt3zcg) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L477)):*
> ```Rust
>                 let entry = self.entries.free(handle);
>                 if let Some(region) = self.free(entry) {
>                     found_matching_slab = found_matching_slab ||
> ```

can we use `|=` operator?

---

*[webrender/src/texture_cache.rs, line 870 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWzJJwOeuSNNkhL-Jf:-KrWzJJwOeuSNNkhL-Jg:b-h0uwbh) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L870)):*
> ```Rust
>         // allocate from.
>         let slab_size = SlabSize::new(width, height);
>         let slab_size_bytes = slab_size.get_size();
> ```

I thought that `get_size()` returns the width of a slab region, so it seems confusing to see `_bytes` here

---

*[webrender/src/texture_cache.rs, line 906 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWznI_gQEGX_REN6jt:-KrWznIao3xnCMEkCFZG:bix9grl) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L906)):*
> ```Rust
>                 region.init(slab_size);
>                 if let Some(location) = region.alloc() {
>                     entry_kind = Some(EntryKind::Cache {
> ```

nit: could be `entry_kind = region.alloc().map(|location| {...});`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1572#-:-KrWXoH9qPraMAAaIMMX:b-9h3ty8)*

glennw:
Review status: 29 of 34 files reviewed at latest revision, 17 unresolved discussions.

---

*[webrender/res/debug_font.fs.glsl, line 10 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWmyRlhB0yG0mjqeYv:-KrXiZa-VibCbnVJ5erI:b-fl0lf2) ([raw file](https://github.com/servo/webrender/blob/9e1a85e8226d7746e82902f44467293d1a9938ca/webrender/res/debug_font.fs.glsl#L10)):*

is that check intended for removal?
</blockquote></details>

Yep, not needed now that we require es3.

---

*[webrender/src/device.rs, line 1644 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWnldvun6JxBA6GUjN:-KrXibdlytfvV27kOwS8:bwapn85) ([raw file](https://github.com/servo/webrender/blob/9e1a85e8226d7746e82902f44467293d1a9938ca/webrender/src/device.rs#L1644)):*

is this path intended for removal as well?
</blockquote></details>

Yep - es3 supports alpha texture format natively.

---

*[webrender/src/freelist.rs, line 94 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWoIpOc8j3Zg8YlU26:-KrXigfTdJ0imw6jh5Al:b-896fix) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/freelist.rs#L94)):*

please add a comment explaining what `upsert` means in this context
</blockquote></details>

Done.

---

*[webrender/src/glyph_cache.rs, line 14 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWogoFMYvVtwVqoMbc:-KrXihWRCQ4qFfOX0En_:b-896fix) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_cache.rs#L14)):*

should it use `euclid` primitives, perhaps? I.e. `TypedSize<u32>` and `TypedPoint2D<f32>`
</blockquote></details>

Done.

---

*[webrender/src/glyph_rasterizer.rs, line 161 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWp5-4Jsx3hNinhPMJ:-KrXiiF6YWEAbrJ1pb8t:b-896fix) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_rasterizer.rs#L161)):*

nit: probably easier to have `if let Some(...) = *entry.get_mut()`
</blockquote></details>

Done.

---

*[webrender/src/glyph_rasterizer.rs, line 273 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWq-bbP3K1cF5K4O4X:-KrXiivUgv2180jd6PWJ:b-p5ovjj) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_rasterizer.rs#L273)):*

shouldn't we check for the `request` return value?
</blockquote></details>

In this case (a newly rasterized glyph), we know that we will definitely want to update.

---

*[webrender/src/glyph_rasterizer.rs, line 274 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWq2sEXv_OllvROrs_:-KrXinf0Il16Ivox8eBh:beq501d) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/glyph_rasterizer.rs#L274)):*

is update always following `request`?
</blockquote></details>

Only if request returns true, or the caller knows it wants to update the image data.

---

*[webrender/src/internal_types.rs, line 100 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWqGqBbpqVMdD-htAd:-KrXiuPx5qUwNBe1VaNU:brxp9sq) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/internal_types.rs#L100)):*

I feel like we'll eventually (not now!) need to have a separate `TargetView` type (instead of this enumeration) that would capture the internal FBOs (per slice) and the depth attachment.
</blockquote></details>

Sounds good.

---

*[webrender/src/renderer.rs, line 2417 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWs_1aUHgVQ6X0_zH2:-KrXivuX7bXy3RQLVZe_:b-896fix) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/renderer.rs#L2417)):*

shouldn't this be `num_layers`?
</blockquote></details>

Done.

---

*[webrender/src/resource_cache.rs, line 108 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWskewDCQvtoPQomGf:-KrXiwc60ie6aOhYSEGD:bmh6pti) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/resource_cache.rs#L108)):*

is there any point in this struct now as oppose to just typedeffing to `FastHashMap`?
</blockquote></details>

Probably not - we could perhaps do as a follow up though?

---

*[webrender/src/resource_cache.rs, line 144 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWswXxYZclKyXjKC51:-KrXj1uXhepeAjlgXMxe:b9f2oz3) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/resource_cache.rs#L144)):*

I think it can be easily done on the caller site now that we have Rust-1.19 (used by Gecko) and all those pretty [retain](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.retain) methods in libstd
</blockquote></details>

Yup, we could do that as a follow up when we remove the resource class cache struct.

---

*[webrender/src/texture_cache.rs, line 231 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWuuUk6JbB7B2iAFgM:-KrXj7cvAuMUQv7Fh9S9:bdd14ei) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L231)):*

do we have any detection of a forgotten `request`?
</blockquote></details>

Yes, get() asserts that the last_frame matches the current frame.

---

*[webrender/src/texture_cache.rs, line 404 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWwcqajWa-ErxJHXCJ:-KrXjF5fWc8zpDqMg6CV:brxp9sd) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L404)):*

I think this procedure can be done simpler and a bit faster without temporary arrays:
  - sort `standalone_entry_handles` in descending order of `last_access`. Given that it's likely to be temporarily coherent, a quicksort-like method would be efficient at skipping the work on already sorted parts
  - check 32th from the back for early return
  - pop 32 items, calling `free()` for each

Although, this doesn't seem to be a performance critical place here, so we can look into it later. The current code is correct.
</blockquote></details>

Sounds good!

---

*[webrender/src/texture_cache.rs, line 477 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWx_Ps_8-X48r4IYWX:-KrXjGzcCP_H-zfSk8rA:b-896fix) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L477)):*

can we use `|=` operator?
</blockquote></details>

Done.

---

*[webrender/src/texture_cache.rs, line 870 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWzJJwOeuSNNkhL-Jf:-KrXjHqJNeZOEhlhGIlo:b-896fix) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L870)):*

I thought that `get_size()` returns the width of a slab region, so it seems confusing to see `_bytes` here
</blockquote></details>

Done.

---

*[webrender/src/texture_cache.rs, line 906 at r2](https://reviewable.io:443/reviews/servo/webrender/1572#-KrWznI_gQEGX_REN6jt:-KrXjIZhBNct46hXqoWa:b-896fix) ([raw file](https://github.com/servo/webrender/blob/869babf069e1532981bbe977204c074883e71be7/webrender/src/texture_cache.rs#L906)):*

nit: could be `entry_kind = region.alloc().map(|location| {...});`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1572)*

glennw:
@kvark Thanks for the review. I think all comments / questions addressed - I still haven't run on the gecko try build yet. I agree with you about the resizing - in the case where we evict everything not used on this frame and *still* can't fit in the cache, we could resize and add more layers. Are you happy for this to be done as a follow up?

kvark:
Yes, I'm happy with it. I assume you are going to test Gecko before merging?
r=me then

---

Review status: 29 of 34 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1572#-:-KrY_xCKwog5gWuz2dX0:b-o7uk91)*

glennw:
Yup, I'll test against Gecko today. Thanks!

JerryShih:
>  I've run the gecko sanity tests locally with this, and they pass. It doesn't change any external interfaces, but it does expect that external textures which use the normal texture2D sampler switch over to place them into the texture2DArray sampler. It doesn't affect anything that goes via the TextureRect or TextureExternal image types. Do you foresee any issues with this? I'll ping you later during the week with some questions I have on running a try build with this change.

I think there is no problem if we try to use texture2DArray for texture2D sampler with your fix.

glennw:
OK, gecko test run results here - https://treeherder.mozilla.org/#/jobs?repo=try&revision=727e26960bc131783b36d0fe4cd6dc143af414b3

I think that's all done correctly and tested OK - but I'll get @staktrace to confirm since I haven't tried this before.

If that's all OK, and with the sign-off from @JerryShih above, this should be good to merge.

glennw:
@staktrace The required changes in gecko are:

* Update gleam to 0.4.8.
* Remove the ```cache_expiry_frames``` field when initializing WR, it's no longer needed.

staktrace:
@glennw The try push looks good, thanks! And also thanks for listing the changes required in gecko, that's perfect!

kvark:
Ok, sounds like a fairy tale, :shipit: 
@bors-servo r+

PR #1576, opened 2017-08-15
Handle image keys in display lists that calling code has deleted.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1576)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1198,10 +1198,11 @@ impl PrimitiveStore {
                         // It's cheap enough that it's not worth introducing a cache for images
                         // right now, but if we introduce a cache for images for some other
                         // reason then we might as well cache this with it.
-                        let image_properties = resource_cache.get_image_properties(image_key);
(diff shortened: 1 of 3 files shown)


glennw:
r? @kvark 

This is mostly indentation changes.

jdm:
https://github.com/servo/webrender/pull/1576/files?w=1

kvark:
Thank you for taking care of those crashes!
@bors-servo r+

jrmuizel:
Why do we want to warn instead of panic here? Isn't this basically a malicious client?

glennw:
This situation occurs if, for example, the client submits a display list referencing an image key, then deletes the image key, and then calls generate frame.

jrmuizel:
What's the use case for supporting that sequence and why would consumer want to draw the display list with the images missing? (that's what the patch does right?)

glennw:
Right now, at least Servo (and perhaps Gecko) don't correctly use transactions to ensure atomically updating image keys vs. display lists. Certainly, once that happens, we could consider making this a panic condition again.

PR #1579, opened 2017-08-16
Avoid Sending an empty message if there is no resource to update.

nical:
I tiny detail, but it is more convenient to check here and always call `update_resources` in gecko/servo than check whether the update list isn't empty before calling `update_resources`. Typically (in particular in gecko), the `ResourceUpdates` is passed around and the code that adds updates is different from the one that sends the transaction.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1579)


kvark:
Honestly, I'm not convinced, but it's not worth arguing about:)
@bors-servo r+

PR #1581, opened 2017-08-16
Add external image support for both 2D and 2D_ARRAY samplers.

glennw:
This fixes an oversight on my part when adding the new texture
cache implementation. The webgl / offscreen rendering code has
not been updated to create array textures.

Add explicit external image support for both 2D and 2D_ARRAY
samplers. In the future, we may deprecate / remove the 2D
image buffer kind, once we have switched everything over to
use array textures, but this is a good interim fix for now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1581)


nical:
Not sure why the error in wrench.rs is showing up in this PR. @bors-servo r+

PR #1582, opened 2017-08-17
Various tidy ups and deinit path improvements.

glennw:
* Change delete_program to move the program.
* Remove hashmap for VAOs.
* Add deinit path for VAOs similar to programs.
* Update renderer and debug render code.
* Remove shader watch code (can re-add later if we get it working again).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1582)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -5,7 +5,6 @@
 use euclid::Transform3D;
 use gleam::gl;
 use internal_types::{RenderTargetMode, TextureSampler, DEFAULT_TEXTURE, FastHashMap};
-//use notify::{self, Watcher};
 use super::shader_source;
 use std::fs::File;
 use std::io::Read;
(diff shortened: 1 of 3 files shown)


webrender/src/device.rs
```diff
@@ -366,32 +364,18 @@ impl Drop for Program {
     }
 }
 
-struct VAO {
-    gl: Rc<gl::Gl>,
```
kvark: niiice

webrender/src/device.rs
```diff
@@ -366,32 +364,18 @@ impl Drop for Program {
...
-            // todo(gw): maybe make these their own type with hashmap?
-            self.gl.delete_buffers(&[self.ibo_id.0]);
-        }
-        if self.owns_vertices {
-            self.gl.delete_buffers(&[self.main_vbo_id.0]);
-        }
-        if self.owns_instances {
-            self.gl.delete_buffers(&[self.instance_vbo_id.0])
-        }
+        debug_assert!(thread::panicking() || self.id == 0);
```
kvark: need to add a message saying that `deinit` was not called
not a blocked, can do as follow up

glennw:
Rebased and updated the deinit assert comment.

@bors-servo r=kvark

PR #1584, opened 2017-08-17
Switch clip code and shaders to use typed GpuCacheAddress directly.

glennw:
This saves a bit of work converting address types in both CPU and
vertex shader code.

It also makes the vertex types use the correct type for cache
addresses, instead of casting to an i32.

The intent is to incrementally move all vertex formats over to
using the 2D style addresses directly, and then remove the
duplicated code from prim_shared.glsl which allows fetching via
an i32.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1584)

### DIFF
--- a/webrender/res/cs_clip_border.vs.glsl
+++ b/webrender/res/cs_clip_border.vs.glsl
@@ -21,8 +21,8 @@ struct BorderCorner {
     int clip_mode;
 };
 
-BorderCorner fetch_border_corner(int index) {
-    vec4 data[2] = fetch_from_resource_cache_2(index);
+BorderCorner fetch_border_corner(ivec2 address) {
+    vec4 data[2] = fetch_from_resource_cache_2_direct(address);
     return BorderCorner(RectWithSize(data[0].xy, data[0].zw),
                         data[1].xy,
                         int(data[1].z),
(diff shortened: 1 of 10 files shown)


webrender/res/cs_clip_border.vs.glsl
```diff
@@ -85,15 +85,15 @@ void main(void) {
         switch (corner.clip_mode) {
             case CLIP_MODE_DASH: {
                 // Fetch the information about this particular dash.
-                BorderClipDash dash = fetch_border_clip_dash(cci.data_index + 2 + 2 * (cci.segment_index - 1));
+                BorderClipDash dash = fetch_border_clip_dash(cci.clip_data_address, 2 + 2 * (cci.segment - 1));
```
kvark: nit: now that the offset has a separate semantics from the base address, it might be reasonable to move the segment -> offset computation logic into the corresponding `fetch_border_clip*` functions, so that here you only call it as `fetch_border_clip_dash(cci.clip_data.address, cci.segment)`

webrender/res/cs_clip_rectangle.vs.glsl
```diff
@@ -18,8 +18,9 @@ struct ClipCorner {
     vec4 outer_inner_radius;
 };
 
-ClipCorner fetch_clip_corner(int index) {
-    vec4 data[2] = fetch_from_resource_cache_2(index);
+ClipCorner fetch_clip_corner(ivec2 address, int index) {
+    address += ivec2(2 + 2 * index, 0);
```
kvark: ... especially since we have the similar semantics here, where the offset is computed inside the `fetch*` function

glennw:
@kvark Sounds good, fixed those functions.

PR #1585, opened 2017-08-17
Add support for primitive shaders referencing the render target caches.

glennw:
This switches ps_cache_image over to use SourceTexture::CacheRGBA8.

The idea here is to enable primitives added during batch creation
to select a render cache target as a source texture. This is possible
now since the texture cache uses array textures, and therefore shares
the same sampler target as the render target cache.

The goal is to begin unifying some of the shaders - for example,
the ps_cache_image shader can be replaced by the normal ps_image
shader, which will reduce the number of shaders and shader switches.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1585)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -239,6 +239,94 @@ impl CpuProfile {
     }
 }
 
+struct SourceTextureResolver {
+    /// A vector for fast resolves of texture cache IDs to
+    /// native texture IDs. This maps to a free-list managed
+    /// by the backend thread / texture cache. We free the
+    /// texture memory associated with a TextureId when its
+    /// texture cache ID is freed by the texture cache, but
+    /// reuse the TextureId when the texture caches's free
+    /// list reuses the texture cache ID. This saves having to
+    /// use a hashmap, and allows a flat vector for performance.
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

Most of the changes here are moving a few pre-existing fields from the ```Renderer``` struct to the new ```SourceTextureResolver``` struct.

It's still a bit untidy how we reference the cache texture in places, but I consider this an incremental improvement.

webrender/src/renderer.rs
```diff
@@ -776,31 +864,18 @@ pub struct Renderer {
...
-    /// by the backend thread / texture cache. We free the
-    /// texture memory associated with a TextureId when its
-    /// texture cache ID is freed by the texture cache, but
-    /// reuse the TextureId when the texture caches's free
-    /// list reuses the texture cache ID. This saves having to
-    /// use a hashmap, and allows a flat vector for performance.
-    cache_texture_id_map: Vec<TextureId>,
-    texture_cache_upload_pbo: PBOId,
+    // Manages and resolves source textures IDs to real texture IDs.
+    texture_resolver: SourceTextureResolver,
```
kvark: nice abstraction!

PR #1586, opened 2017-08-17
Use the proper window dimension when setting up wrench.

jrmuizel:
This makes sure we use the same dimensions for the first frame
as for subsequent ones. It also fixes the offset of things
when on hidpi on macOS.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1586)


glennw:
Seems reasonable.

@bors-servo r+

PR #1587, opened 2017-08-17
Remove WebGL implementation, it's no longer needed.

### LINKED ISSUE #607: Work out a better way to handle webgl / webvr.
Playing back webgl and webvr compositor commands on the backend thread is not ideal, for a number of reasons.

Specifically, the backend thread should generally be doing minimal work so it is free to respond to query requests etc. Having things like webvr vsync called on that thread could be very bad for UI latency in some cases.

Investigate the possible ways to handle webgl / webvr in WR and write up a plan for how to integrate it in order to alleviate these performance issues.

### LINKED ISSUE #1353: Consider splitting the WebGL related code into a separate crate.

glennw:
Clients use the external image APIs to provide native
WebGL handles or readback buffers.

Fixes #1353.
Fixes #607.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1587)


glennw:
@kvark Yup, got a servo patch ready to go for when this lands :)

@bors-servo r=kvark

emilio:
Neat, thanks for getting rid of this! :)

glennw:
Thanks to @MortimerGoro and @emilio for their work to make this feasible :)

emilio:
(Mostly to @MortimerGoro, tbh :P)

glennw:
Fixed tidy script.

@bors-servo r=kvark

MortimerGoro:
i-love-the-smell-of-napalm-in-the-morning-i-love-the-smell-of-deleted-code-in-the-morning

PR #1588, opened 2017-08-18
Add support for shader includes.

glennw:
If the first line of a shader is //import followed by a comma
separated list of glsl filenames, prepend those as includes
for the shader. The idea here is that we'll start splitting
up prim_shared.glsl etc into smaller blocks that are faster
to parse and compile, and just generally more manageable.

This also supports having the VS and FS in the same glsl file. This
commit ports the debug_font and debug_color shaders to this style.

If we go ahead with this format, we'll incrementally port other
shaders to use this style, which requires only one file per shader.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1588)


glennw:
r? @kvark and @nical 

It's always a bit of an open question how to manage shader includes and common code with GLSL. This seems to work reasonably well, and would allow us to modularize our shared GLSL code a lot. What do you think?

webrender/src/device.rs
```diff
@@ -46,10 +46,12 @@ const GL_FORMAT_BGRA_GL: gl::GLuint = gl::BGRA;
...
 
 const SHADER_VERSION_GL: &str = "#version 150\n";
-
 const SHADER_VERSION_GLES: &str = "#version 300 es\n";
 
-static SHADER_PREAMBLE: &str = "shared";
+const SHADER_KIND_VERTEX: &str = "#define WR_VERTEX_SHADER\n";
+const SHADER_KIND_FRAGMENT: &str = "#define WR_FRAGMENT_SHADER\n";
+const SHADER_IMPORT: &str = "//import ";
+const SHADER_LINE_MARKER: &str = "#line 1\n";
```
nical: Neat!

webrender/src/device.rs
```diff
@@ -770,23 +748,13 @@ impl Device {
...
-        for prefix in shader_preamble {
-            s.push_str(prefix);
-        }
-        s.push_str(source_str);
-
         let id = gl.create_shader(shader_type);
-        let mut source = Vec::new();
-        source.extend_from_slice(s.as_bytes());
-        gl.shader_source(id, &[&source[..]]);
+        let sources: Vec<&[u8]> = sources.iter().map(|s| s.as_bytes()).collect();
```
kvark: I'd prefer us building a single source chunk per shader (as opposed to having an array of chunks), because OpenGL is the only API supporting multiple sources.

webrender/src/device.rs
```diff
@@ -1190,128 +1158,137 @@ impl Device {
...
-                    program: &mut Program,
-                    include: String,
-                    descriptor: &VertexDescriptor) -> Result<(), ShaderError> {
-        debug_assert!(self.inside_frame);
+        // Parse the main .glsl file, including any imports
+        // and append them to the list of sources.
+        let mut shared_strings = Vec::new();
+        if let Some(shared_source) = get_shader_source(base_filename, &self.resource_override_path) {
+            parse_shader_source(shared_source,
+                                &self.resource_override_path,
```
kvark: time to learn the new formatting style ;)

glennw:
OK, updated to use ```#include```, support includes on any line, use a single source string, and also updated the shader validation test to be more representative of what we actually compile (and use the same shared code to build the shader string).

kvark:
Thank you!
@bors-servo r+

PR #1594, opened 2017-08-21
Renderer coverage profiling

kvark:
Computes the fraction of screen getting covered by the opaque/transparent passes.
Also hides `renderer` module from the public eyes (breaking change).
r? @glennw

~~WIP until the FF patch is ready~~
[wr-profile.diff.zip](https://github.com/servo/webrender/files/1240226/wr-profile.diff.zip)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1594)

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -8,8 +8,6 @@ use std::env;
 use std::path::PathBuf;
 use webrender;
 use webrender::api::*;
-use webrender::renderer::{PROFILER_DBG, RENDER_TARGET_DBG, TEXTURE_CACHE_DBG};
-use webrender::renderer::ExternalImageHandler;
 
 struct Notifier {
     window_proxy: glutin::WindowProxy,
@@ -65,7 +63,7 @@ pub trait Example {
                 event: glutin::Event,
                 api: &RenderApi,
                 document_id: DocumentId) -> bool;
-    fn get_external_image_handler(&self) -> Option<Box<ExternalImageHandler>> {
+    fn get_external_image_handler(&self) -> Option<Box<webrender::ExternalImageHandler>> {
         None
     }
 }
@@ -113,7 +111,7 @@ pub fn main_wrapper(example: &mut Example,
     };
 
     let size = DeviceUintSize::new(width, height);
-    let (mut renderer, sender) = webrender::renderer::Renderer::new(gl, opts).unwrap();
+    let (mut renderer, sender) = webrender::Renderer::new(gl, opts).unwrap();
     let api = sender.create_api();
     let document_id = api.add_document(size);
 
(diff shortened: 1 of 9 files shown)


glennw:

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1594)*

kvark:
@glennw thanks!
I agree that GL queries would be more universal to implement and useful as a metric. Will look into it.

glennw:
@kvark Cool, sounds good. ```GL_SAMPLES_PASSED``` should give us what we want, I think.

kvark:
@glennw please take another look

wr-sampler-profile

glennw:
Reviewed 4 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1594)*

glennw:
Nice, this looks great! Could you just confirm that the reported compositor CPU time with this change is the same as without. I couldn't see anything that would introduce a stall, but the CPU compositor time in that screenshot seems very high (although maybe that's caused by the screenshot tool or something?).

Otherwise, r=me.

kvark:
@glennw thanks for taking a look!
Good point about compositor timing. I'll compare and get back here.

glennw:
Once this is all in and working, these would be good counters to add to the values tracked on the perf dashboard.

kvark:
@glennw both master and this PR code show equal numbers for CPU compositor timing on that sample:
  - debug: ~7ms
  - release: ~14ms

glennw:
@kvark Interesting - I guess you're running with vsync on, perhaps? The reported time in CPU for that test on my machine is ~0.2 ms with vsync off.

@bors-servo r+

kvark:
FF patch
```diff
diff --git a/gfx/webrender_bindings/src/bindings.rs b/gfx/webrender_bindings/src/bindings.rs
index 758e8945bf71..e2f41dccc1b6 100644
--- a/gfx/webrender_bindings/src/bindings.rs
+++ b/gfx/webrender_bindings/src/bindings.rs
@@ -7,9 +7,9 @@ use std::os::raw::{c_void, c_char, c_float};
 use gleam::gl;
 
 use webrender_api::*;
-use webrender::renderer::{ReadPixelsFormat, Renderer, RendererOptions};
-use webrender::renderer::{ExternalImage, ExternalImageHandler, ExternalImageSource};
-use webrender::renderer::DebugFlags;
+use webrender::{ReadPixelsFormat, Renderer, RendererOptions};
+use webrender::{ExternalImage, ExternalImageHandler, ExternalImageSource};
+use webrender::DebugFlags;
 use webrender::{ApiRecordingReceiver, BinaryRecorder};
 use thread_profiler::register_thread_with_profiler;
 use moz2d_renderer::Moz2dImageRenderer;
```

PR #1597, opened 2017-08-22
Handle the long text for dl::push_text()

JerryShih:
@kvark @glennw r?

There is an crash[1] to show that we hit the gpu_cache's size limit for a long text. We could just split the long text into several parts to prevent this problem.

[1]
https://crash-stats.mozilla.com/report/index/ebd69828-67ed-4d94-8acf-413c00170814

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1597)


webrender/src/gpu_cache.rs
```diff
@@ -468,9 +468,8 @@ impl<'a> GpuDataRequest<'a> {
         self.texture.pending_blocks.extend_from_slice(blocks);
     }
 
-    /// Consume the request and return the number of blocks written
-    pub fn close(self) -> usize {
```
glennw: I'd prefer to keep this as ```close``` taking ```self``` as a parameter, that prevents it from being used again after the close is called.

JerryShih: I will keep the close() function and create a new one for regular block num checking.

JerryShih:
@glennw 
r? again for the GpuDataRequest updating.

edunham:
@bors-servo r-

please check the appveyor build

JerryShih:
@edunham 
The windows appveyor build failed is a known issue, and that's not related to this PR. Please check https://github.com/servo/webrender/pull/1598

> @kvark Thanks! The windows / appveyor failures are occurring on all PRs at the moment. I haven't had a chance to look into it - I suspect the rustc version might be the cause?

Please merge the https://github.com/servo/webrender/pull/1599. That could fix the build problem.

edunham:
@JerryShih Thanks for the clarification. The temporary r- is to try to get that PR through. I may need to try closing this and reopening after #1599 lands as well.

kvark:
Closing temporarily to resolve Homu

kvark:
oh, homu is still testing the same commit even after closing

PR #1600, opened 2017-08-23
Refactor and small optimization of batching code.

glennw:
This is prep work for some upcoming changes to support
different vertex formats per batch type. This will allow:

1) Much smaller vertex format for glyph instances.
2) Different vertex attribute types per batch type (e.g. using 2x u16 for GPU cache addresses).

There's also a small optimization - we no longer store the bounding
rects for items in the opaque batch lists, since we don't need
to check for overlap there.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1600)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -101,39 +101,20 @@ struct DynamicTaskInfo {
     rect: DeviceIntRect,
 }
 
-pub struct BatchList {
-    pub alpha_batches: Vec<PrimitiveBatch>,
-    pub opaque_batches: Vec<PrimitiveBatch>,
+pub struct AlphaBatchList {
+    pub batches: Vec<AlphaPrimitiveBatch>,
 }
 
-impl BatchList {
-    fn new() -> BatchList {
-        BatchList {
-            alpha_batches: Vec::new(),
-            opaque_batches: Vec::new(),
+impl AlphaBatchList {
+    fn new() -> AlphaBatchList {
+        AlphaBatchList {
+            batches: Vec::new(),
         }
-    }
-
-    fn with_suitable_batch<F>(&mut self,
-                              key: &AlphaBatchKey,
-                              item_bounding_rect: &DeviceIntRect,
-                              f: F) where F: Fn(&mut PrimitiveBatch) {
-        let batch = self.get_suitable_batch(key, item_bounding_rect);
-        f(batch)
     }
 
     fn get_suitable_batch(&mut self,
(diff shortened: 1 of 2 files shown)


glennw:
r? @kvark 

Passes WPT / wrench, gecko try here:

https://hg.mozilla.org/try/rev/5e75903c6a43b8fd756ab43a3fe966d4bce18da8
https://treeherder.mozilla.org/#/jobs?repo=try&revision=5e75903c6a43b8fd756ab43a3fe966d4bce18da8

glennw:
Rebased

webrender/src/tiling.rs
```diff
@@ -143,37 +124,74 @@ impl BatchList {
...
-        let batch = &mut batches[selected_batch_index.unwrap()];
+        let batch = &mut self.batches[selected_batch_index.unwrap()];
         batch.item_rects.push(*item_bounding_rect);
 
-        batch
+        &mut batch.instances
+    }
+}
+
+pub struct OpaqueBatchList {
```
kvark: hmm, there appears to be quite a bit of duplicated code, on both the API and the implementation side.
Might be worth uniting those structs:
```rust
trait PrimitiveBatch: Sized {
  type Bound;
  fn new(&AlphaBatchKey) -> Self;
  fn check_intersections(&self, &Self::Bound) -> bool;
  fn add_bound(&mut self, &Self::Bound) -> &mut Vec<PrimitiveInstance>;
}
struct BatchList<B> {
  pub batches: Vec<B>,
}
impl<B: PrimitiveBatch> for BatchList<B> {
...
}
```

kvark:
Sorry about the late review!
The PR does simplify a lot of things nicely, the least part I like is the duplication of `get_suitable_batch` code

webrender/src/tiling.rs
glennw: Although the code is mostly duplicated right now, I'm planning to experiment with changing the batching strategies for opaque and alpha - I think they'll end up being quite different.

In particular, while the alpha batcher would stay quite similar, the opaque batcher would ensure very large primitives get rendered first (even if a batch break is involved), while smaller primitives could get added to a per-batch-type list, without having to search for an appropriate batch.

So I think it makes sense to leave the code duplicated for now - does that seem reasonable?

kvark:
Ok, looking forward to see the next steps ;) 👍 
@bors-servo r+

glennw:
Rebased.

@bors-servo r=kvark

PR #1602, opened 2017-08-23
Add font instance keys which associate size and other options with a key

lsalzman:
This adds FontInstanceKey to the API, which alleviates the need to pass both
a FontKey and other options parameterizing a font to any functions using that
font. FontInstanceKeys are managed via the new add_font_instance and
delete_font_instance API functions. This enables validation of these parameters
to be done up front when add_font_instance is called, rather than every time
text is submitted. It also future-proofs FontInstances by allowing more
parameters than might otherwise be tenable if one had to supply them all when
submitting text.

At a minimum, one needs to specify a size, much like a ScaledFont in Gecko.
One can also optionally supply FontInstanceOptions for parameters common across
all platforms, and FontInstancePlatformOptions for parameters that are specific
to a given platform (which will be developed further in future patches).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1602)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -95,6 +95,7 @@ impl NestedDisplayListInfo {
 struct FlattenContext<'a> {
     scene: &'a Scene,
     builder: &'a mut FrameBuilder,
+    resource_cache: &'a ResourceCache,
     tiled_image_map: TiledImageMap,
     replacements: Vec<(ClipId, ClipId)>,
     nested_display_list_info: Vec<NestedDisplayListInfo>,
@@ -104,11 +105,12 @@ struct FlattenContext<'a> {
 impl<'a> FlattenContext<'a> {
     fn new(scene: &'a Scene,
            builder: &'a mut FrameBuilder,
-           resource_cache: &ResourceCache)
+           resource_cache: &'a ResourceCache)
            -> FlattenContext<'a> {
         FlattenContext {
             scene,
             builder,
+            resource_cache,
             tiled_image_map: resource_cache.get_tiled_image_map(),
             replacements: Vec::new(),
             nested_display_list_info: Vec::new(),
@@ -565,12 +567,12 @@ impl Frame {
                                               info.image_rendering);
             }
             SpecificDisplayItem::Text(ref text_info) => {
+                let instance = context.resource_cache.get_font_instance(text_info.font_key).unwrap();
                 context.builder.add_text(clip_and_scroll,
                                          reference_frame_relative_offset,
                                          item_rect_with_offset,
                                          &clip_with_offset,
-                                         text_info.font_key,
-                                         text_info.size,
+                                         instance,
                                          &text_info.color,
                                          item.glyphs(),
                                          item.display_list().get(item.glyphs()).count(),
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -4,7 +4,8 @@
 
 use api::{BorderDetails, BorderDisplayItem, BoxShadowClipMode, ClipAndScrollInfo, ClipId, ColorF};
 use api::{DeviceIntPoint, DeviceIntRect, DeviceIntSize, DeviceUintRect, DeviceUintSize};
-use api::{ExtendMode, FontKey, FontRenderMode, GlyphInstance, GlyphOptions, GradientStop};
+use api::{ExtendMode, FontInstance, FontInstanceKey, FontKey, FontRenderMode};
+use api::{GlyphInstance, GlyphOptions, GradientStop};
 use api::{ImageKey, ImageRendering, ItemRange, LayerPoint, LayerRect, LayerSize};
 use api::{LayerToScrollTransform, LayerVector2D, LayoutVector2D, LineOrientation, LineStyle};
 use api::{LocalClip, PipelineId, RepeatMode, ScrollSensitivity, SubpixelDirection, TextShadow};
@@ -875,24 +876,37 @@ impl FrameBuilder {
                     run_offset: LayoutVector2D,
                     rect: LayerRect,
                     local_clip: &LocalClip,
-                    font_key: FontKey,
-                    size: Au,
+                    font: &FontInstance,
                     color: &ColorF,
                     glyph_range: ItemRange<GlyphInstance>,
                     glyph_count: usize,
                     glyph_options: Option<GlyphOptions>) {
         // Trivial early out checks
-        if size.0 <= 0 {
+        if font.size.0 <= 0 {
+            return
+        }
+
+        // Sanity check - anything with glyphs bigger than this
+        // is probably going to consume too much memory to render
+        // efficiently anyway. This is specifically to work around
+        // the font_advance.html reftest, which creates a very large
+        // font as a crash test - the rendering is also ignored
+        // by the azure renderer.
+        if font.size >= Au::from_px(4096) {
             return
         }
 
         // TODO(gw): Use a proper algorithm to select
         // whether this item should be rendered with
         // subpixel AA!
-        let mut normal_render_mode = self.config.default_font_render_mode;
+        let mut default_render_mode = self.config.default_font_render_mode.limit_by(font.render_mode);
+        if let Some(options) = glyph_options {
+            default_render_mode = default_render_mode.limit_by(options.render_mode);
+        }
 
         // There are some conditions under which we can't use
         // subpixel text rendering, even if enabled.
+        let mut normal_render_mode = default_render_mode;
         if normal_render_mode == FontRenderMode::Subpixel {
             if color.a != 1.0 {
                 normal_render_mode = FontRenderMode::Alpha;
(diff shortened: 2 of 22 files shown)


glennw:

---

*[webrender/src/frame_builder.rs, line 902 at r1](https://reviewable.io:443/reviews/servo/webrender/1602#-KsHwUVud-WrCsB9Ivba:-KsHwUVud-WrCsB9Ivbb:b-8u89ls) ([raw file](https://github.com/servo/webrender/blob/2dbd0a68cc536a9214e61f76c68cb6e852157c40/webrender/src/frame_builder.rs#L902)):*
> ```Rust
>         // whether this item should be rendered with
>         // subpixel AA!
>         let mut default_render_mode = cmp::min(self.config.default_font_render_mode,
> ```

I'd prefer we don't use cmp() here - it's not clear to me on reading it what the semantics are (even though it works numerically).

---

*[webrender/src/frame_builder.rs, line 974 at r1](https://reviewable.io:443/reviews/servo/webrender/1602#-KsHwmKr3pOBYNVxIfRH:-KsHwmKr3pOBYNVxIfRI:b-hwihi) ([raw file](https://github.com/servo/webrender/blob/2dbd0a68cc536a9214e61f76c68cb6e852157c40/webrender/src/frame_builder.rs#L974)):*
> ```Rust
>                 let mut text_prim = prim.clone();
>                 text_prim.font.color = shadow_prim.shadow.color.into();
>                 text_prim.color = shadow_prim.shadow.color;
> ```

Why do we need to store the color in two places here?

---

*[webrender/src/prim_store.rs, line 532 at r1](https://reviewable.io:443/reviews/servo/webrender/1602#-KsHu-Gn2THNcxrpsqyy:-KsHu-Go6wsgBOCmuQy3:b-6rnvyq) ([raw file](https://github.com/servo/webrender/blob/2dbd0a68cc536a9214e61f76c68cb6e852157c40/webrender/src/prim_store.rs#L532)):*
> ```Rust
>         let mut font = self.font.clone();
>         font.size = font.size.scale_by(device_pixel_ratio);
>         if run_mode == TextRunMode::Shadow {
> ```

Let's stick with the match that was here - that means we'll get a compile error here should anyone ever add a field to the TextRunMode enum.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1602)*

glennw:
This generally looks fine to me - just a few comments, and CI seems unhappy too (examples need updating, I think).

lsalzman:

---

*[webrender/src/frame_builder.rs, line 902 at r1](https://reviewable.io:443/reviews/servo/webrender/1602#-KsHwUVud-WrCsB9Ivba:-KsKFgBhDqA5ksQRc7wT:b-zhmie8) ([raw file](https://github.com/servo/webrender/blob/2dbd0a68cc536a9214e61f76c68cb6e852157c40/webrender/src/frame_builder.rs#L902)):*

I'd prefer we don't use cmp() here - it's not clear to me on reading it what the semantics are (even though it works numerically).
</blockquote></details>

I added a slightly more explicit function for combining the render modes that doesn't use cmp::min in my upcoming review patch. The semantics are meant to agree with Gecko, in that when an individual text run comes in, with its own AA override, that AA override is meant to disable some AA functionality that was otherwise enabled by default. So Gecko's AA default is permissive, and sometimes we need to disallow it.

---

*[webrender/src/frame_builder.rs, line 974 at r1](https://reviewable.io:443/reviews/servo/webrender/1602#-KsHwmKr3pOBYNVxIfRH:-KsKG9aj1z6lALqUqQKf:bizk6x6) ([raw file](https://github.com/servo/webrender/blob/2dbd0a68cc536a9214e61f76c68cb6e852157c40/webrender/src/frame_builder.rs#L974)):*

Why do we need to store the color in two places here?
</blockquote></details>

Color is quantized in FontInstance via ColorU to reduce the amount of hash keys and rasterized glyphs we generate, and turning that back into a ColorF will not yield the same fidelity when we access the color to pass it in to the shader or blending. In the interests of preserving both constraints, I had to leave the ColorF field around.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1602)*

lsalzman:

---

*[webrender/src/prim_store.rs, line 532 at r1](https://reviewable.io:443/reviews/servo/webrender/1602#-KsHu-Gn2THNcxrpsqyy:-KsKGvDu8pbymWr7kVCM:b-72mme2) ([raw file](https://github.com/servo/webrender/blob/2dbd0a68cc536a9214e61f76c68cb6e852157c40/webrender/src/prim_store.rs#L532)):*

Let's stick with the match that was here - that means we'll get a compile error here should anyone ever add a field to the TextRunMode enum.
</blockquote></details>

Okay, can do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1602)*

webrender/src/frame.rs
```diff
@@ -557,12 +559,12 @@ impl Frame {
                                               info.image_rendering);
             }
             SpecificDisplayItem::Text(ref text_info) => {
+                let instance = context.resource_cache.get_font_instance(text_info.font_key).unwrap();
```
glennw: It might make sense to store the font instance key in the text primitive here. Then, we wouldn't need the resource cache at all at this part of the code (it could be fetched later), and the text run primitive structure could be made significantly smaller. We could do this as a follow up though...

webrender/src/frame_builder.rs
```diff
@@ -913,30 +927,32 @@ impl FrameBuilder {
 
         // Shadows never use subpixel AA, but need to respect the alpha/mono flag
         // for reftests.
-        let (shadow_render_mode, subpx_dir) = match self.config.default_font_render_mode {
+        let (shadow_render_mode, subpx_dir) = match default_render_mode {
             FontRenderMode::Subpixel | FontRenderMode::Alpha => {
                 // TODO(gw): Expose subpixel direction in API once WR supports
```
glennw: This comment can be removed now, I think?

webrender/src/prim_store.rs
```diff
@@ -504,21 +504,17 @@ pub struct TextShadowPrimitiveCpu {
 
 #[derive(Debug, Clone)]
 pub struct TextRunPrimitiveCpu {
-    pub font_key: FontKey,
+    pub font: FontInstance,
```
glennw: This is the bit I was mentioning above that could just store the FontInstanceKey here.

webrender_api/src/font.rs
```diff
@@ -115,6 +115,20 @@ impl FontRenderMode {
             _ => panic!("Should only be given the fractional part"),
         }
     }
+
+    // Combine two font render modes such that the lesser amount of AA limits the AA of the result.
+    pub fn limit_by(self, other: FontRenderMode) -> FontRenderMode {
+        match self {
```
glennw: Might be tidier if written as ```match (self, other)``` but can be left as is if you prefer.

lsalzman:
Review status: 4 of 22 files reviewed at latest revision, 7 unresolved discussions.

---

*[webrender/src/frame.rs, line 562 at r2](https://reviewable.io:443/reviews/servo/webrender/1602#-KsEdoo3nW3TiZzXzJ7_-r2-562:-KsnfbVCE0ijjEHP5N17:b-gztqyx) ([raw file](https://github.com/servo/webrender/blob/c04d0a40aa70025200c734f6646846f7452d0899/webrender/src/frame.rs#L562)):*

It might make sense to store the font instance key in the text primitive here. Then, we wouldn't need the resource cache at all at this part of the code (it could be fetched later), and the text run primitive structure could be made significantly smaller. We could do this as a follow up though... 
</blockquote></details>

There's some nastiness with respect to needing to access the size and render mode when building the primitive, which in the simplest form would require multiple lookups into the font instance hashmap to deal with. It might be possible to restructure the code to avoid that, but I think that is best for a followup.

---

*[webrender/src/frame_builder.rs, line 932 at r2](https://reviewable.io:443/reviews/servo/webrender/1602#-KsEdoo4r82bTeQYCNYD-r2-932:-Ksng3ZPWRn8LzB3MwvF:b-c5u84r) ([raw file](https://github.com/servo/webrender/blob/c04d0a40aa70025200c734f6646846f7452d0899/webrender/src/frame_builder.rs#L932)):*

This comment can be removed now, I think?
</blockquote></details>

Not quite, as still haven't really exposed subpixel direction in the api. I just merely routed some plumbing there from the FontInstance to not just hardcode the subpixel direction, but there's still no particular way of manipulating that for the FontInstance just yet.

---

*[webrender_api/src/font.rs, line 121 at r2](https://reviewable.io:443/reviews/servo/webrender/1602#-KsEdoo4r82bTeQYCNYM-r2-121:-KsngfNj1CBEsT4d-C8Q:bwygxsx) ([raw file](https://github.com/servo/webrender/blob/c04d0a40aa70025200c734f6646846f7452d0899/webrender_api/src/font.rs#L121)):*

Might be tidier if written as ```match (self, other)``` but can be left as is if you prefer.
</blockquote></details>

Noted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1602)*

glennw:
Servo patch for the API update is here - https://github.com/glennw/servo/commit/26b580963e3048a99b0c634742bbeb192897d74c

I'm assuming @lsalzman will have something similar ready for Gecko as a patch :)

PR #1603, opened 2017-08-23
Tile decomposition crash

jrmuizel:
This adds a test case that causes a panic in render_backend.rs with 'RenderBackend' panicked at 'assertion failed: image_size.width <= tile_size'. This is a reduction from a crash in Gecko with tiled blob images turned on.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1603)

### DIFF
--- a/wrench/src/rawtest.rs
+++ b/wrench/src/rawtest.rs
@@ -38,12 +38,51 @@ impl<'a> RawtestHarness<'a> {
 
     pub fn run(mut self) {
         self.retained_blob_images_test();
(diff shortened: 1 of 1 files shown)


jrmuizel:
It might be worth fixing this before landing it.

glennw:
I figured it was OK to land since the raw tests are only run manually right now - but we can hold off if you prefer?

jrmuizel:
I don't feel strongly either way.

PR #1605, opened 2017-08-23
Add some basic testing for the binary record/replay infrastructure

jrmuizel:
This also ends up testing the png command.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1605)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,5 +30,6 @@ script:
   - if [ $BUILD_KIND = DEBUG ]; then (cd webrender_api && cargo test --verbose --features "ipc"); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo build --verbose --no-default-features); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cargo test --all --verbose); fi
+  - if [ $BUILD_KIND = DEBUG ]; then (cd wrench && cargo run -- --save binary png reftests/text/decorations.yaml && cargo run -- replay --play wr-record.bin); fi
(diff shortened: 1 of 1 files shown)


glennw:
This is failing on CI, that new command is not exiting cleanly.

glennw:
Oh, I'm guessing the failure was the png command deinit fix, which has landed now...

@bors-servo r+

kvark:
Looks like travis fails to parse our script now?
> $ if [ $BUILD_KIND = DEBUG ]; then (cd wrench && cargo run -- --save binary png reftests/text/decorations.yaml && cargo run -- replay --play wr-record.bin)

>/home/travis/.travis/job_stages: eval: line 58: syntax error: unexpected end of file

glennw:
@jrmuizel I think this is just missing the ```fi``` at the end of the line?

glennw:
Failed on the Linux + Debug build config:

```
Running `/home/travis/build/servo/webrender/target/debug/wrench --save binary png reftests/text/decorations.yaml`

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NoBackendAvailable(XOpenDisplayFailed)', /checkout/src/libcore/result.rs:859

```

It worked on mac and Linux release though...

glennw:
@jrmuizel The OSMesa backend is only available in release builds, I think.

We could change this to use the ```headless``` command and run on a release config. That would allow the ```png``` command to work. I'm not sure what would happen with the replay command if run under headless though?

Alternatively, we could just run this line on the mac build config, since that has a GL backend available?

glennw:
Closing for now - please re-open when ready for re-review.

PR #1607, opened 2017-08-24
[RFC] Introduce a web based debugger and some basic examples.

glennw:
This introduces a simple web based debugger for WR, as an optional feature.

The initial implementation allows toggling the debug flags (e.g. profiler, texture cache debug view) and dumping a list of draw call batches to the browser.

In the future, I'd plan to expand the feature set significantly - for example, a render tree visualization, and more detailed timing. We can also use this as a debugging tool - for example, adding support for stepping through a single frame etc.

The main things to answer are:
 * Should the client live in this repo or elsewhere?
 * Is the current setup of the feature going to work OK with vendoring?
 * Is the current implementation / threading strategy reasonable?

The implementation uses Vue.js, websockets and Bulma.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1607)


glennw:
debugger

glennw:
Thoughts @kvark @jrmuizel @nical

glennw:
BTW, it's listed as [RFC], but it's actually ready to go from my perspective, if everyone is happy with it as-is.

nical:
I would like this to be in the repo. Hopefully, having it behind a feature flag like you did should be enough for vendoring.

webrender/src/renderer.rs
```diff
@@ -1481,6 +1517,93 @@ impl Renderer {
...
+    }
+
+    #[cfg(feature = "debugger")]
+    fn update_debug_server(&self) {
+        while let Ok(msg) = self._debug_server.debug_rx.try_recv() {
+            match msg {
+                DebugMsg::FetchBatches(sender) => {
+                    let mut batch_list = BatchList::new();
+
+                    if let Some(ref frame) = self.current_frame {
```
kvark: nit: could combine this pattern match with the next one to avoid nesting

webrender/src/renderer.rs
```diff
@@ -1481,6 +1517,93 @@ impl Renderer {
...
+                        if let Some(ref frame) = frame.frame {
+                            for pass in &frame.passes {
+                                for target in &pass.alpha_targets.targets {
+                                    batch_list.push("[Clip] Clear", target.clip_batcher.border_clears.len());
```
kvark: can we add (to JSON) any information about the target itself?

webrender/src/debug_server.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Serialize)]
+pub struct BatchInfo {
+    kind: &'static str,
+    count: usize,
+}
+
+#[derive(Serialize)]
+pub struct BatchList {
+    kind: &'static str,
+    batches: Vec<BatchInfo>,
```
kvark: we might consider having a bit stronger types here, e.g.
```rust
#[derive(Serialize)]
pub struct ClipBatcherInfo {
  clears: usize,
  borders: usize,
  rectangles: usize,
}
#[derive(Serialize)]
pub struct CacheBatcherInfo {
...
}
pub struct BatchList {
    kind: &'static str,
    clip: ClipBatcherInfo,
    cache: CacheBatcherInfo,
    alpha: HashMap<&'static str, usize>,
}
```

webrender/src/render_backend.rs
```diff
@@ -461,6 +461,15 @@ impl RenderBackend {
...
                     self.notifier.lock().unwrap().as_mut().unwrap().new_frame_ready();
                 }
+                ApiMsg::DebugCommand(option) => {
+                    let msg = ResultMsg::DebugCommand(option);
+                    self.result_tx.send(msg).unwrap();
+                    let notifier = self.notifier.lock();
+                    notifier.unwrap()
+                            .as_mut()
+                            .unwrap()
+                            .new_frame_ready();
```
kvark: looks like `new_frame_ready` gets abused more. Maybe we should just add a `ping` method to be used both here and for `MemoryPressure`?

webrender/src/renderer.rs
```diff
@@ -1481,6 +1517,93 @@ impl Renderer {
...
+                    let json = serde_json::to_string(&batch_list).unwrap();
+                    sender.send(json).ok();
+                }
+            }
+        }
+    }
+
+    fn handle_debug_command(&mut self, command: DebugCommand) {
+        match command {
+            DebugCommand::EnableProfiler(enable) => {
```
kvark: since we are at it, maybe we can remove the duplicated logic from wrench and examples boilerplate

webrender/src/renderer.rs
```diff
@@ -1481,6 +1517,93 @@ impl Renderer {
...
+                ResultMsg::DebugCommand(command) => {
+                    self.handle_debug_command(command);
+                }
+            }
+        }
+    }
+
+    #[cfg(not(feature = "debugger"))]
+    fn update_debug_server(&self) {
+    }
```
kvark: perhaps, would be cleaner to do `let _ = &self.debug_server` than having underscore in `_debug_server`

webrender/src/renderer.rs

webrender/src/renderer.rs
glennw: Definitely - the batch list is very much a proof of concept - I'll expand on the information with a more detailed view as a follow up.

webrender/src/debug_server.rs
glennw: Yep - sounds good. OK to do as a follow up?

webrender/src/render_backend.rs
glennw: Sounds like a good idea.

webrender/src/renderer.rs
glennw: Do you mean remove the keyboard shortcuts? I think it makes sense to keep those for quick debugging - or did you mean something else?

webrender/src/renderer.rs

glennw:
@kvark Thanks for the review! Fixed up some of those nits. I think it's OK to do the rest as follow ups - does that sound OK? Also added the debugger feature to wrench, which I had missed previously.

kvark:
@glennw 
> Do you mean remove the keyboard shortcuts?

yeah, I don't see much point in keeping the shortcuts if you can do the same thing by just enabling `debugger` feature. I won't lose my sleep if the shortcuts are still there though ;)

> Fixed up some of those nits. I think it's OK to do the rest as follow ups - does that sound OK?

Absolutely!

One last concern before r=me is CI-testing the debugger feature. I don't see you modifying the travis/appveyor scripts.

glennw:
@kvark Ah, good point - I've rebased and updated the travis file to build wrench in release + debugger mode.

kvark:
Wonderful, it's shipping time!
@bors-servo r+

kvark:
CI has legitimate errors:
> error: no field `opaque_batches` on type `tiling::BatchList`
>    --> /home/travis/build/servo/webrender/webrender/src/renderer.rs:1564:53
>     |
> 1564 |                                                    .opaque_batches
>     |                                                     ^^^^^^^^^^^^^^ did you mean `opaque_batch_list`?

glennw:
Ah, it needed a rebase after the batching PR landed - done.

kvark:
thanks!
@bors-servo r+

PR #1608, opened 2017-08-24
Fix assertion when decomposing tiled repeating images.

nical:
The code was asserting that if an image is repeated, its tile decomposition should already have been handled, but the decomposition is actually handled after the assertion.
The condition of the assertion does interact with whether we can do the repetition in the shader or if we have to do it on the CPU, so it should be part of that branch.
In addition, this PR renames a few variables to make it a bit easier to read.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1608)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -994,22 +994,15 @@ impl Frame {
         // This can happen with very tall and thin images used as a repeating background.
         // Apparently web authors do that...
 
-        let mut repeat_x = false;
-        let mut repeat_y = false;
+        let needs_repeat_x = info.stretch_size.width < item_rect.size.width;
+        let needs_repeat_y = info.stretch_size.height < item_rect.size.height;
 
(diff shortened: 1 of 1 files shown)


webrender/src/frame.rs
```diff
@@ -994,22 +994,15 @@ impl Frame {
...
+        let needs_repeat_y = info.stretch_size.height < item_rect.size.height;
 
-        if info.stretch_size.height < item_rect.size.height {
-            // If this assert blows up it means we haven't properly decomposed the image in decompose_image.
-            debug_assert!(image_size.height <= tile_size);
-            // we don't actually tile in this dimension so repeating can be done in the shader.
-            repeat_y = true;
-        }
+        let tiled_in_x = image_size.width > tile_size;
+        let tiled_in_y = image_size.height > height;
```
kvark: should this be `tile_size`?

nical: darn.

Issue #1593, opened 2017-08-21
Some GL states should be reset after using external image hooks


nical:
in [bug 1391481](https://bugzilla.mozilla.org/show_bug.cgi?id=1391481), @sotaroikeda investigated a bug where an external image implementation changes the gl texture binding. Since `Device` caches the currently bound texture to avoid binding it again several times we run into issues because it did not see the binding change.

One solution is to try to restore the state in the external image hook (gecko-side), but I think that it'd be more bullet-proof to clobber the cached texture binding in the Device (and maybe other cached states) after calling external image hooks in webrender.

sotaroikeda:
@nical, do you expect the code like [this](https://bug1391481.bmoattachments.org/attachment.cgi?id=8900599)? Or other implementation?

nical:
Either something like this or just resetting `Device.bound_textures` to containing 16 `TextureId::invalid()` so that the next time something needs a texture the device will re-do the binding. I suspect it would be more efficient because that way we don't need to touch the texture slots that we don't use.
cc @kvark

kvark:
The external API can return a set of bits indicating which states are fixed: textures, shaders, VAOs, etc. That, if there is a concern that state resets after external blobs are affecting the perf. Otherwise - we can just reset "everything", starting with those `bound_textures`.
We've got to have certain level of trust that these flags are correct, at least as long as we are using GL, which doesn't have built-in isolation primitives like command buffers.

nical:
Fixed in #1609

PR #1609, opened 2017-08-24
Reset cached gl state after calling into external image hooks.

nical:
A naive fix for #1593.

If always resetting cached states turns out to be too expensive can easily extend this to do what @kvark suggests in  #1593 and let the handler communicate through a bitfield what states needs to be clobbered if any.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1609)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -805,6 +805,14 @@ impl Device {
         &self.capabilities
     }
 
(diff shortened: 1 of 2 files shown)


glennw:
It would be good to add a rawtest for this in the future.

PR #1610, opened 2017-08-24
Add the mut keyword for font_key_from_native_handle() in windows plat…

glennw:
…form.

Fix the build break in windows. The add_native_font() call in this
function needs a mutable variable.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1610)


glennw:
Pushed PR from @JerryShih to my fork, to try and work around bors problems.

@bors-servo p=100 r=glennw

PR #1617, opened 2017-08-27
Make RendererKind public since Servo needs that enum.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1617)


glennw:
r? anyone

PR #1618, opened 2017-08-28
Add a basic render pass / target / batch view to the debugger.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1618)


glennw:
debug1

nical:
Nice!
@bors-servo r+

PR #1620, opened 2017-08-28
Tidy up how sampler slots are handled by device.

glennw:
This moves sampler binding from the device to the application
level, which seems a bit tidier. It allows different modules
to declare their own sampler type enums.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1620)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -20,15 +20,14 @@ use debug_render::DebugRenderer;
 use debug_server::{BatchList, DebugMsg, DebugServer};
 use device::{DepthFunction, Device, FrameId, Program, TextureId, VertexDescriptor, GpuMarker, GpuProfiler, PBOId};
 use device::{GpuTimer, TextureFilter, VAO, VertexUsageHint, FileWatcherHandler, TextureTarget, ShaderError};
-use device::{get_gl_format_bgra, VertexAttribute, VertexAttributeKind};
+use device::{get_gl_format_bgra, TextureSlot, VertexAttribute, VertexAttributeKind};
 use euclid::{Transform3D, rect};
 use frame_builder::FrameBuilderConfig;
 use gleam::gl;
 use gpu_cache::{GpuBlockData, GpuCacheUpdate, GpuCacheUpdateList};
 use internal_types::{FastHashMap, CacheTextureId, RendererFrame, ResultMsg, TextureUpdateOp};
 use internal_types::{TextureUpdateList, RenderTargetMode, TextureUpdateSource};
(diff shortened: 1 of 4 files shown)


webrender/src/renderer.rs
```diff
@@ -768,9 +809,23 @@ fn create_prim_shader(name: &'static str,
...
 
-    device.create_program(name,
-                          &prefix,
-                          &vertex_descriptor)
+    let program = device.create_program(name,
+                                        &prefix,
+                                        &vertex_descriptor);
+
+    if let Ok(ref program) = program {
+        device.bind_shader_sampler(program, "sColor0", TextureSampler::Color0);
```
kvark: Not sure what being generic over `Into<TextureSlot>` gives us here. Would it be simpler to just do:
```rust
device.bind_shader_textures(program, &[
  ("sColor0", 0),
  ("sColor1", 1),
  ... // this table could be defined as a static const on top and re-used, if needed, for different programs
]);
```

webrender/src/renderer.rs
```diff
@@ -780,7 +835,16 @@ fn create_clip_shader(name: &'static str, device: &mut Device) -> Result<Program
...
     debug!("ClipShader {}", name);
 
-    device.create_program(name, &prefix, &DESC_CLIP)
+    let program = device.create_program(name, &prefix, &DESC_CLIP);
+
+    if let Ok(ref program) = program {
+        device.bind_shader_sampler(program, "sColor0", TextureSampler::Color0);
+        device.bind_shader_sampler(program, "sLayers", TextureSampler::Layers);
+        device.bind_shader_sampler(program, "sRenderTasks", TextureSampler::RenderTasks);
+        device.bind_shader_sampler(program, "sResourceCache", TextureSampler::ResourceCache);
```
kvark: for cases like this where you need a subset of the table, we can either have 2 separate tables or group the items so that it's easy to select a range with slices

webrender/src/renderer.rs
glennw: It means that there is only one place in the code that does the mapping from a sampler enum to an integer. It seems like that might make it harder to accidentally get things out of sync - a minor detail, certainly.

webrender/src/renderer.rs
glennw: Changed it to ```bind_shader_samplers``` which takes an array - good idea.

glennw:
@kvark Pushed an update and commented above. Happy to just change it to be integer based if you think there's no benefits.

kvark:
thank you!
I think we are inconsistent with terminology (samplers vs textures), but otherwise looks solid.
@bors-servo r+

glennw:
Yes, there are definitely inconsistencies with the terminology that we should fix. Thanks!

PR #1621, opened 2017-08-28
Remove the instance stride parameter from VAO creation.

glennw:
Now that we have vertex descriptors, we can calculate the
instance stride internally, so there is no need for the caller
to specify it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1621)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -284,6 +284,12 @@ impl VertexAttribute {
 }
 
 impl VertexDescriptor {
+    fn instance_stride(&self) -> u32 {
+        self.instance_attributes
+            .iter()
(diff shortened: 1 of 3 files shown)


webrender/src/device.rs
```diff
@@ -307,9 +313,7 @@ impl VertexDescriptor {
 
         if !self.instance_attributes.is_empty() {
             instance.bind(gl);
-            let instance_stride: u32 = self.instance_attributes
-                                           .iter()
-                                           .map(|attr| attr.size_in_bytes()).sum();
```
nical: This assumes we are packing attributes in a way that does not leave any padding. It'd be nice to have a check that the assumption holds true.

PR #1625, opened 2017-08-28
Switch box shadow cache primitives to use custom vertex format.

glennw:
Cache primitives should use a custom vertex format. This will
make it easier to make changes to the PrimitiveInstance struct
in the future (e.g. compressing some of the fields, and switching
over to use native cache addresses).

Also switch box shadow shader to be in a single file.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1625)


glennw:
Rebased

kvark:
Looks good!
@bors-servo r+

PR #1626, opened 2017-08-29
Use PBO for updating vertex textures.

glennw:
This means we don't need to manually cycle through a number of
textures to avoid driver stalls.

Also simplify the vertex texture logic, since we only use it
for a small number of data types now, which are all floats.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1626)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -40,7 +40,6 @@ use std;
 use std::cmp;
 use std::collections::VecDeque;
 use std::f32;
-use std::marker::PhantomData;
 use std::mem;
 use std::path::PathBuf;
 use std::rc::Rc;
@@ -62,7 +61,6 @@ use api::{BlobImageRenderer, channel, FontRenderMode};
 use api::{YuvColorSpace, YuvFormat};
 use api::{YUV_COLOR_SPACES, YUV_FORMATS};
 
-pub const GPU_DATA_TEXTURE_POOL: usize = 5;
 pub const MAX_VERTEX_TEXTURE_WIDTH: usize = 1024;
 
 const GPU_TAG_CACHE_BOX_SHADOW: GpuProfileTag = GpuProfileTag { label: "C_BoxShadow", color: debug_colors::BLACK };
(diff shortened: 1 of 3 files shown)


webrender/src/renderer.rs
```diff
@@ -514,107 +512,79 @@ impl CacheTexture {
...
+        let texels_per_item = mem::size_of::<T>() / 16;
+        let items_per_row = MAX_VERTEX_TEXTURE_WIDTH / texels_per_item;
 
         // Extend the data array to be a multiple of the row size.
         // This ensures memory safety when the array is passed to
         // OpenGL to upload to the GPU.
         if items_per_row != 0 {
             while data.len() % items_per_row != 0 {
-                data.push(T::default());
+                data.push(unsafe { mem::uninitialized() });
```
kvark: I never understood the obsession with `uninitialized` as opposed to `zeroed`. It reads like a premature optimization to me, which may end up wasting debugging time when something empty would look like it has data.

webrender/src/renderer.rs
```diff
@@ -514,107 +512,79 @@ impl CacheTexture {
...
+                                RenderTargetMode::None,
+                                1,
+                                None);
+        }
 
-    fn texture_width<T>() -> usize {
-        MAX_VERTEX_TEXTURE_WIDTH - (MAX_VERTEX_TEXTURE_WIDTH % Self::texels_per_item::<T>())
-    }
+        // Bind a PBO to do the texture upload.
+        // Updating the texture via PBO avoids CPU-side driver stalls.
```
kvark: I don't think the comment really applies to this use case: if your `update_texture_from_pbo` follows `update_pbo_data`, then the actual work the driver has got to do is no different from a regular `glTexImage*`, and it will still stall.
In order to get the benefit from PBOs, we need to:
  1. do the `update_pbo_data` as early as possible, giving it more time until the PBO -> texture copy is issued
  2. don't re-use the same PBO space until the *GPU* is done with the previous frame, which means having a number of those PBO buffers cycled over

webrender/src/renderer.rs
glennw: Fair point - will change to zeroed() as a follow up.

webrender/src/renderer.rs
glennw: Intuitively, I agree with your comment above - but the ```perf``` output from the intel driver suggests otherwise.

With the path above, I see no performance stalls reported, and nothing in the profile graphs. With the previous code, we see stalls in the perf logs, and noticeable stalls in the frame profiler.

I guess the driver has a stall-free path for PBOs but not for textures, perhaps?

kvark: Thanks for confirming this! I'll try to get a close look what's going on there when I get a chance.

PR #1627, opened 2017-08-29
Remove textures hashmap in device.

glennw:
Also add a proper texture deinit path, and ensure that all
textures created are deleted at shutdown.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1627)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -4,7 +4,7 @@
 
 use euclid::Transform3D;
 use gleam::gl;
-use internal_types::{RenderTargetMode, FastHashMap};
+use internal_types::RenderTargetMode;
 use super::shader_source;
 use std::fs::File;
 use std::io::Read;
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

There's a couple of messy bits in here that could still be tidied up, but it's a net improvement, I think.

glennw:
I'll do a gecko try run for this patch in the morning.

webrender/src/device.rs
```diff
@@ -1250,6 +1217,13 @@ impl Device {
         texture.width = 0;
         texture.height = 0;
         texture.layer_count = 0;
+        texture.id = 0;
```
kvark: hmm, not sure I understand the semantics here: is it valid for `Texture` to have `id==0`, effectively making it a null/invalid texture? In this case having `Option<Texture>` anywhere would be confusing. Otherwise (and preferred), can we not allow `id == 0` here?

glennw: Oops, that's just a bug - well spotted! That line was left there from a previous iteration of this patch I was working on. Indeed, it's not valid for a texture to have id == 0 until it has been deleted.

glennw:
Rebased

glennw:
Try run looks good:

https://hg.mozilla.org/try/rev/cc2d771cf6a71512fe345f10ca90ca059dd2ac08
https://treeherder.mozilla.org/#/jobs?repo=try&revision=cc2d771cf6a71512fe345f10ca90ca059dd2ac08

kvark:
Thank you!
@bors-servo r+

PR #1628, opened 2017-08-29
Using stack for storing preserve_3d_map.

mephisto41:
If we have nested preserve 3d stacking contexts, the preserve_3d_map
need using stack otherwise we'll handle wrong preserve_3d_map and cause
crash.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1628)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1374,7 +1374,7 @@ impl FrameBuilder {
         // The stacking contexts that fall into this category are
         //  - ones with `ContextIsolation::Items`, for their actual items to be backed
         //  - immediate children of `ContextIsolation::Items`
-        let mut preserve_3d_map: FastHashMap<StackingContextIndex, RenderTaskId> = FastHashMap::default();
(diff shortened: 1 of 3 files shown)


kvark:
@mephisto41 could you add a simple reftest for the crash?

mephisto41:
gecko [test case](https://searchfox.org/mozilla-central/source/layout/reftests/transform-3d/preserves3d-nested.html) causes this crash. But sure, I'll convert it to wrench reftest.

wrench/reftests/reftest.list
```diff
@@ -10,3 +10,4 @@ include scrolling/reftest.list
 include snap/reftest.list
 include split/reftest.list
 include text/reftest.list
+include transform3d/reftest.list
```
kvark: please put it into `split` folder

mephisto41: done.

kvark:
thank you!
@bors-servo r+

PR #1629, opened 2017-08-29
Add an API to remove pipelines.

### LINKED ISSUE #691: struct Scene doesn't support removing pipelines
The HashMaps in here will just accumulate pipeline forever.

staktrace:
Fixes #691 

I'm not at all sure this code is doing the right thing (it's largely copy-paste), so please don't hesitate to point out all the things that are wrong with this code. In particular I don't know (a) if this will leave things in an inconsistent state and (b) how to exercise the wrench changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1629)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -275,6 +275,12 @@ impl RenderBackend {
                     DocumentOp::Nop
                 }
             }
+            DocumentMsg::RemovePipeline(pipeline_id) => {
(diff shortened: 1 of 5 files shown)


staktrace:
@glennw I was under the impression that ClearNamespace removes all the documents associated with a namespace ID, and since the documents are getting dropped, their `Scene` objects are getting dropped as well, along with the pipeline tables that this patch is touching.

PR #1630, opened 2017-08-29
Add ResourceUpdates::merge and clear.

nical:
Nothing fancy or surprising here, `merge` consumes another `ResourceUpdates` object and merges it into self, `clear` just clears the internal vector.
Gecko needs something like merge in a few places when two separate things are building up updates that have to be merged into a single list before sending to webrender, and there is also a place where we clear the vector if updates were built while webrender was shutdown for whatever expected or unexpected reason.

At least `merge` should also make sense for servo if/when frame consistency is implemented (updates built on various threads (layout, canvases, etc.) could be merged and then sent to webrender.

These can be implemented in the bindings (since the updates member is public), but since the use cases are sound, I would rather avoid logic in the bindings.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1630)

### DIFF
--- a/webrender_api/src/api.rs
+++ b/webrender_api/src/api.rs
@@ -70,6 +70,14 @@ impl ResourceUpdates {
     pub fn delete_font(&mut self, key: FontKey) {
         self.updates.push(ResourceUpdate::DeleteFont(key));
     }
+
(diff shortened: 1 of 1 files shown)


kvark:
timeout again, retrying
@bors-servo retry

PR #1633, opened 2017-08-30
Propagate errors during rendering WebRender instead of panicking

### LINKED ISSUE #1631: Propagate errors during rendering WebRender instead of panicking
#768 propagates errors when initializing WebRender. But it is not enough to propagate errors  instead of panicking, since actual shader compilations happen when they are used under Renderer::render() in normal use cases. Then the render() panics when the compilations failed. gecko want to avoid the panic.

The problem of gecko is tracked by [Bug 1390138](https://bugzilla.mozilla.org/show_bug.cgi?id=1390138)

sotaroikeda:
Fixes #1631

I am not sure this approach is OK. It makes render() as not to panic when a  shader error happens.

@glennw @kvark  @nical

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1633)

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -201,7 +201,7 @@ pub fn main_wrapper(example: &mut Example,
         }
 
         renderer.update();
-        renderer.render(DeviceUintSize::new(width, height));
+        renderer.render(DeviceUintSize::new(width, height)).unwrap();
         window.swap_buffers().ok();
     }
 
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -716,9 +716,14 @@ impl LazilyCompiledShader {
         Ok(shader)
     }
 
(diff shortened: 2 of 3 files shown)


glennw:
Good question how we should handle this case!

I think that propagating the errors back to ```render()``` this way might be a bit error prone - because if we return mid-frame due to an error we won't necessarily clean up correctly for the frame (e.g. return the render targets for this frame to the render target pool). So I think returning mid-frame might be a bit hard to reason about.

Instead, how about we maintain a list of errors that were encountered (e.g. during shader ```bind```) and store them inside the ```Renderer``` struct. Each time we encounter a ```bind``` error we just don't issue that draw call. Then, at the end of the frame, we return a ```Result``` which contains the list of errors that were encountered, or ```Ok```.

Thoughts?

sotaroikeda:
Thanks for the advice! It looks good approach!

sotaroikeda:
I updated the pull request based on the comment. How about it?

glennw:
Nice!

@bors-servo r+

PR #1635, opened 2017-08-30
Switch box shadows to render to A8 cache instead of RGBA8.

glennw:
Box shadows are just an alpha mask - we may as well save GPU
bandwidth by drawing them into the A8 cache target.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1635)


Gankro:
I binary searched this to be the change that is causing https://bugzilla.mozilla.org/show_bug.cgi?id=1395982 (mac is constantly flickering).

Gankro:
Correction: this seems to just expose an older issue (see linked discussion in previous comment).

glennw:
@Gankro fix is here https://github.com/servo/webrender/pull/1666

PR #1636, opened 2017-08-30
Add a basic display list treeview to the debugger.

glennw:
This allows viewing the input set of display lists and documents
for a given WR instance.

The treeview itself is very basic, and the information provided
per item is minimal. Consider this a proof of concept which can
be expanded upon to provide more information.

Importantly, it demonstrates how to provide information to the
debugger websocket that is only available in the render backend
(the previous debug items were all available in the renderer
instance).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1636)

### DIFF
--- a/webrender/src/debug_server.rs
+++ b/webrender/src/debug_server.rs
@@ -14,8 +14,9 @@ use ws;
 // debug command queue. These are sent in a separate queue so
 // that none of these types are exposed to the RenderApi interfaces.
 // We can't use select!() as it's not stable...
-pub enum DebugMsg {
-    FetchPasses(ws::Sender),
+enum DebugMsg {
+    AddSender(ws::Sender),
+    RemoveSender(ws::util::Token),
 }
 
 // Represents a connection to a client.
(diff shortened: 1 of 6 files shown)


glennw:
doc

webrender/src/render_backend.rs
```diff
@@ -514,4 +528,134 @@ impl RenderBackend {
...
+
+    #[cfg(not(feature = "debugger"))]
+    fn get_docs_for_debugger(&self) -> String {
+        String::new()
+    }
+
+    // TODO(gw): The display list traversal below is a copy of the skeleton
+    //           from frame.rs. We might like to try and unify the traversal
+    //           functions so that this code could be shared, and invoke
+    //           a callback as items are encountered?
```
nical: Or maybe implement Display or Debug or some specific method on display items such that you could do something like:

```rust
match *item.item() {
    SpecificDisplayItem::PushStackingContext(..) => {
        let mut subtraversal = item.sub_iter();
        let mut child_node = debug_server::TreeNode::new("stacking context");
        self.traverse_items(&mut subtraversal, &mut child_node);
        node.add_child(child_node);
        return Some(subtraversal);
    }
    other_item => { node.add_item(format!("{}", other_item)); }
}
```

webrender/src/debug_server.rs
```diff
@@ -101,6 +114,42 @@ impl DebugServer {
...
+                    self.senders.retain(|sender| {
+                        sender.token() != token
+                    });
+                }
+            }
+        }
+
+        // Broadcast the message to all senders. Keep
+        // track of the ones that failed, so they can
+        // be removed from ths active sender list.
```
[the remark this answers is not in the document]
glennw: Fixed

webrender/src/render_backend.rs
glennw: Ah, good idea - that's much tidier.

glennw:
@nical Comments addressed and rebased, thanks!

glennw:
Serializing from the debugger to a yaml file should be totally doable.

PR #1638, opened 2017-08-31
Fix texture cache profiler display.

### LINKED ISSUE #1637: The texture cache does not update profile counters anymore.
Since #1572 the texture cache info in the profiler view is not updated, making it look like there is no image allocated, which creates some confusion like in this bug report https://bugzilla.mozilla.org/show_bug.cgi?id=1394806#c0

I think that this info was useful and it'd be nice to get it back somewhere (be it in the HUD profiler view or the web debugger). If we are not going to restore it the profiler view we should we should remove it to avoid the confusion.

glennw:
This was disabled when the new texture cache allocator was merged.
The numbers aren't super meaningful right now, since the caches
don't grow, but they at least show which texture caches are
allocated and what the fixed size of them is.

Fixes #1637

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1638)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -131,6 +131,11 @@ impl ResourceProfileCounter {
         self.value += 1;
         self.size += size;
     }
+
+    pub fn set(&mut self, count: usize, size: usize) {
+        self.value = count;
+        self.size = size;
+    }
 }
 
 impl ProfileCounter for ResourceProfileCounter {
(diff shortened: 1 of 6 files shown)


nical:
Thanks!

@bors-servo r+

staktrace:
@glennw The `bytes_per_pixel` change caused gecko build bustage.

staktrace:
@glennw (btw i already fixed the above issue, no need for you to do anything - that was just an fyi to consider for future changes)

glennw:
@staktrace Ah, I missed that one. Apologies!

PR #1639, opened 2017-08-31
Make sure no PBO is bound on texture deletion

kvark:
Attempt to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1395327
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1639)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -1172,6 +1172,7 @@ impl Device {
 
     pub fn free_texture_storage(&mut self, texture: &mut Texture) {
         debug_assert!(self.inside_frame);
+        debug_assert_eq!(self.bound_pbo, PBOId(0));
 
         if texture.format == ImageFormat::Invalid {
             return;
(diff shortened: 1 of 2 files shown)


glennw:
Looks sane, thanks!

@bors-servo r+

jrmuizel:
FWIW, I still get crashes with this manually applied.

kvark:
@jrmuizel thanks for testing this!
Perhaps, you could also try changing the width/height/depth we pass to `tex_image_3d` in `free_texture_storage`? E.g. pass 1s instead of 0s.

glennw:
@jrmuizel Are you able to test with the integrated GPU force enabled? It sounds like it may be an issue in the nvidia driver.

Gankro:
I applied the tip of this branch on top of @staktrace's latest WR update (https://hg.mozilla.org/try 8117e084974a) and can no longer reproduce the crashes.

jrmuizel:
Do we have any idea what code paths would've triggered this so that we can get a test in the webrender repo?

Gankro:
Scratch that, reproduced reliably this morning. Guess I got lucky last night. 😖

kvark:
@jrmuizel my hypothesis (for this PR) was `TextureUpdateOp::Update` followed by `TextureUpdateOp::Free`, which was trying to deallocate texture space while a PBO is bound.
Now that you confirmed it's not fixing the bug for you, we are back to square one.

PR #1642, opened 2017-08-31
Introduce shared A8 target that is available as input to all subsequent passes.

glennw:
Many tasks (such as box shadows) are simple tasks that don't
have dependencies on other tasks in the tree. This change
ensures that those tasks are always drawn in the first render pass.

A new sampler is introduced that is always bound to the shared
target as an input for all subsequent passes, for tasks that
opt in.

This is a performance win where a cacheable task would previously
be rendered in multiple passes. For example, a box shadow that is
used by pass 1 and 2 would previously be drawn twice, but with this
change the box shadow is drawn only once (if the box shadow cache
key is the same).

This is a performance win by itself, but is also some prep work
for some ideas I want to prototype related to how we manage
clip masks as part of the clip stack.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1642)

### DIFF
--- a/webrender/src/render_task.rs
+++ b/webrender/src/render_task.rs
@@ -68,6 +68,16 @@ impl RenderTaskTree {
             }
         }
 
+        // If this task can be shared between multiple
+        // passes, render it in the first pass so that
+        // it is available to all subsequent passes.
+        let pass_index = if task.is_shared() {
+            debug_assert!(task.children.is_empty());
+            0
+        } else {
+            pass_index
+        };
+
         let pass = &mut passes[pass_index];
         pass.add_render_task(id);
     }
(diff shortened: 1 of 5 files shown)


glennw:
I've run all the Servo WPT/CSS tests on this change.

webrender/src/render_task.rs
```diff
@@ -514,6 +524,26 @@ impl RenderTask {
         }
     }
 
+    // Check if this task wants to be made available as an input
+    // to all passes in the render task tree. To qualify for this,
+    // the task needs to have no children / dependencies.
```
kvark: why can't we just check for `task.children.is_empty()` here then?
plus, we need to mention the A8 requirement

webrender/src/render_task.rs
```diff
@@ -514,6 +524,26 @@ impl RenderTask {
         }
     }
 
+    // Check if this task wants to be made available as an input
+    // to all passes in the render task tree. To qualify for this,
```
kvark: correction: all passes *except* the first one

webrender/src/renderer.rs
```diff
@@ -394,14 +399,35 @@ impl SourceTextureResolver {
...
+    fn end_pass(&mut self,
+                pass_index: usize,
+                pass_count: usize,
+                mut a8_texture: Option<Texture>,
+                mut rgba8_texture: Option<Texture>,
+                a8_pool: &mut Vec<Texture>,
+                rgba8_pool: &mut Vec<Texture>) {
+        // If we have cache textures from previous pass, return them to the pool.
+        if let Some(texture) = self.cache_rgba8_texture.take() {
+            rgba8_pool.push(texture);
```
kvark: nit: could be `rgba8_pool.extend(self.cache_rgba8_texture.take())`;

glennw:
@kvark Thanks! Comments addressed and rebased.

webrender/src/render_task.rs
glennw: Fixed

webrender/src/render_task.rs
glennw: Right now that's the only requirement but in the future there will be more involved logic here.

webrender/src/renderer.rs
glennw: Nice!

PR #1641, opened 2017-08-31
Fix some warnings that crept in.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1641)

PR #1644, opened 2017-08-31
Fix nightly compile warnings and add fail-on-warnings annotations

staktrace:
To prevent more things like #1641 . I tested that this builds on {Mac, Linux, Windows} using {stable, nightly} compilers, in the default configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1644)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ addons:
       - llvm-3.8-dev
       - libedit-dev
 env:
-  - BUILD_KIND=DEBUG RUST_BACKTRACE=1
-  - BUILD_KIND=RELEASE RUST_BACKTRACE=1
+  - BUILD_KIND=DEBUG RUST_BACKTRACE=1 RUSTFLAGS='--deny warnings'
+  - BUILD_KIND=RELEASE RUST_BACKTRACE=1 RUSTFLAGS='--deny warnings'
 before_install:
   - pip install mako servo-tidy
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: trusty
 language: rust
 rust:
-  - 1.17.0
+  - 1.19.0
   - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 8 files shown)


kvark:
@staktrace I don't think this would be a strictly positive change. When tinkering with WR, it's often desirable to leave the room for warnings in order to iterate faster - they are harmless after all.
I'd vote to only fail on warnings in CI builds, leaving the local development more flexible.

staktrace:
@kvark Yeah that sounds like a better idea. I'll see if I can get that working.

staktrace:
@kvark updated patches to only run with --deny warnings in CI builds. I also threw in a patch to build on stable instead of 1.17 so we stay up to date.

staktrace:
For posterity, another reason to not bake in the #![deny(warnings)] into lib.rs files is that right now webrender CI doesn't run on Windows. This means warnings in Windows-only code will not get caught by CI but can then fail local builds for people building on Windows.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 dist: trusty
 language: rust
 rust:
-  - 1.17.0
+  - stable
```
kvark: I think we should be tracking the version Gecko uses here, which is not always `stable`

staktrace: @kvark see https://groups.google.com/d/msg/mozilla.dev.platform/IL6OrN_5ifg/TaUcsWvYBAAJ - Gecko is tracking stable rust, but with a two week delay. Unless we want to keep bumping this every 6 weeks I think it's better to just leave it at `stable`. Considering we're now at 1.19 clearly nobody remembered to update this for the last two rust version bumps, I don't think it's reasonable to expect people to remember to update this for each new rust release.

kvark: So you are willing to allow every 2/6 weeks (so, 33% of time) for WR to base on the Rust version that has more features than Gecko sees? I'd rather keep the current version fixed until someone  is bored or want to use specific feature that they can confirm Gecko's Rust supports.

staktrace: Ok, fair enough. I've updated the patch to just bump it from 1.17 to 1.19.

kvark:
Thank you!
@bors-servo r+

Issue #1648, opened 2017-08-31
MotionMark CSS bouncing tagged images is slower in Gecko with WebRender than with out


jrmuizel:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1395728

I see this running on my Mac

http://browserbench.org/MotionMark/developer.html?test-interval=30&display=minimal&tiles=big&controller=ramp&frame-rate=50&kalman-process-error=1&kalman-measurement-error=4&time-measurement=performance&suite-name=HTMLsuite&test-name=CSSbouncingtaggedimages&complexity=278

With WebRender I get 93.54 vs 279.69 without.

This seems like something that WebRender should be able to handle well.

jrmuizel:
This is running with webrendest and layers-free turned on.

jrmuizel:
Interestingly I get painted layers for the images when I run this on Linux.

glennw:
Does that explain the performance issue? Or are you saying that you *don't* get painted layer on other platforms so there is a different issue there?

jrmuizel:
It does not explain the performance issue. It seems slow on Linux and Mac (regardless of if the images show up as blob images or not).

glennw:
OK, I'll do some profiling in both Servo and Gecko and report results here.

jrmuizel:
Here's a standalone version: https://jrmuizel.github.io/perf-tests/motion-mark-tagged-images-still.html. I get 37 ms of GPU time with it.

glennw:
@jrmuizel OK, in that standalone version I see 36ms in Gecko, and 3ms in Servo. I'll find out where the discrepancy is today.

jrmuizel:
I suspect it's due to additional clips that gecko's doing

glennw:
https://github.com/servo/webrender/pull/1649 drops the GPU time in Gecko from 36ms down to 5ms on my machine. This is still quite a bit higher than Servo, but makes it reasonable at least.

jrmuizel:
So I tried to have a look at what the differences were between Servo and Gecko's display lists for this page. I wasn't able to get binary replay of recordings working from servo working so that made things more difficult. However, there were some differences:
1. Gecko has redundant clips: https://bugzilla.mozilla.org/show_bug.cgi?id=1396131
2. The bounds of the clips on the stacking contexts are smaller in servo's case

Manually removing the redundant clips did not seem to improve the performance so perhaps it's the clip bounds on the stacking context.

[...]

PR #1649, opened 2017-08-31
Skip clip masks where the inner rect encloses the primitive rect.

glennw:
This fixes some performance issues in the Motionmark images demo.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1649)

### DIFF
--- a/webrender/src/render_task.rs
+++ b/webrender/src/render_task.rs
@@ -244,7 +244,8 @@ impl RenderTask {
     pub fn new_mask(key: Option<ClipId>,
                     task_rect: DeviceIntRect,
                     raw_clips: &[ClipWorkItem],
-                    extra_clip: Option<ClipWorkItem>)
+                    extra_clip: Option<ClipWorkItem>,
+                    prim_rect: DeviceIntRect)
                     -> Option<RenderTask> {
         // Filter out all the clip instances that don't contribute to the result
         let mut inner_rect = Some(task_rect);
@@ -283,13 +284,20 @@ impl RenderTask {
         //           In the future, we'll expand this to handle the
         //           more complex types of clip mask geometry.
         let mut geometry_kind = MaskGeometryKind::Default;
-        if inner_rect.is_some() && clips.len() == 1 {
-            let (_, ref info) = clips[0];
-            if info.border_corners.is_empty() &&
-               info.image.is_none() &&
-               info.complex_clip_range.get_count() == 1 &&
-               info.layer_clip_range.get_count() == 0 {
-                geometry_kind = MaskGeometryKind::CornersOnly;
+        if let Some(inner_rect) = inner_rect {
+            // If the inner rect completely contains the primitive
+            // rect, then this mask can't affect the primitive.
+            if inner_rect.contains_rect(&prim_rect) {
+                return None;
+            }
+            if clips.len() == 1 {
+                let (_, ref info) = clips[0];
+                if info.border_corners.is_empty() &&
+                   info.image.is_none() &&
+                   info.complex_clip_range.get_count() == 1 &&
+                   info.layer_clip_range.get_count() == 0 {
+                    geometry_kind = MaskGeometryKind::CornersOnly;
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


glennw:
I'll run the Servo and Gecko tests against this before merging, but it seems like it should be a safe optimization.

r? @kvark

glennw:
WPT/CSS tests in Servo pass.

Gecko try run here looks good (nearly complete):

https://hg.mozilla.org/try/rev/d0893ae2eb0a051db8819254054df695056dcf96
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d0893ae2eb0a051db8819254054df695056dcf96

Fixes #1648

webrender/src/render_task.rs
```diff
@@ -252,6 +253,20 @@ impl RenderTask {
 
             match clip_info.bounds.inner {
                 Some(ref inner) if !inner.device_rect.is_empty() => {
+                    // If this clip item has an inner rect that completely
+                    // encloses the screen rect of the primitive, it can't
+                    // possibly affect the result, so remove it during
+                    // this filtering pass.
+                    if inner.device_rect.contains_rect(&prim_rect) {
```
kvark: There is only one case where `task_rect` is larger than `prim_rect`: when we are processing clips specified for parent stacking contexts, while the primitive itself doesn't have a specific clip.

But in this case it should be desirable to have this task even if the particular primitive rect is not affected, because the generated mask is to be re-used for multiple primitives inside this stacking context.

I think we need to sort this out at a higher level: have a heuristic for a stacking context to share it's mask with children primitives. The heuristic can be as follows:
  1. compute the bounding box of all primitives inside this stacking context that don't have specific clips already
  2. compute the sum of bounding box squares of all the same primitives
  3. if the sum is smaller than the bounding box of all, then pass `None` here for `clip_id` and `prim_rect` for `task_rect`
  4. if the sum is larger, then do what we are doing now

As you can see, with this approach we'll not need a hack inside `new_mask`, and we'll not pass the primitive rectangle in here (just the task rectangle).

glennw: That does sound a bit more elegant conceptually, but I think the implementation may be a bit complicated.

We'd have to either scan the primitive list inside a stacking context first to get the sum, and know which path to choose when generating the masks, or we'd have to keep track of the primitives that need a clip mask and then revisit them to set up the appropriate clip mask, wouldn't we? It's possible I'm over-thinking this :)

I think we could land this first, since it's a simple win, and then revisit the heuristic above as a possible follow up. What do you think?

glennw: Huh, I have an idea that might fix this and drastically simplify the way we manage shared masks. Will need some more thought and prototyping to see if it works though. @jrmuizel Do you want an urgent fix for that performance issue, or is it fine to hold off for a few days and work out the best long term solution?

jrmuizel:
There's no urgency for a fix

glennw:
@kvark I wrote up what I've been thinking about for clip masks - https://github.com/servo/webrender/wiki/Improved-clipping.

When you have time available, it'd be great to hear any feedback or issues you can think of related to it. It seems like it should just work, but it's quite likely there's something subtle I haven't thought though...

kvark:
@glennw Thank you! I moved the discussion to #1656

glennw:
@kvark In terms of this PR / optimization - I realized a simpler way to achieve it. The patch now just skips creating a mask altogether if the final inner rect contains the primitive rect.

This removes any hacks regarding modifying the cache key, and still fixes up the performance of this benchmark.

Tests run OK in Servo, and Gecko try run here looks good:

https://hg.mozilla.org/try/rev/23927113bc7051c5779d3ed5c5dd5a0e16f2e689
https://treeherder.mozilla.org/#/jobs?repo=try&revision=23927113bc7051c5779d3ed5c5dd5a0e16f2e689

(Also rebased).

webrender/src/render_task.rs
```diff
@@ -283,13 +284,20 @@ impl RenderTask {
...
-            let (_, ref info) = clips[0];
-            if info.border_corners.is_empty() &&
-               info.image.is_none() &&
-               info.complex_clip_range.get_count() == 1 &&
-               info.layer_clip_range.get_count() == 0 {
-                geometry_kind = MaskGeometryKind::CornersOnly;
+        if let Some(inner_rect) = inner_rect {
+            // If the inner rect completely contains the primitive
+            // rect, then this mask can't affect the primitive.
+            if inner_rect.contains_rect(&prim_rect) {
```
kvark: hmm, so now of all the primitives sharing the same task, some will just cop out by returning `None` here and others will process as usual? seems logically correct to me

glennw: Yup, that's right :)

PR #1651, opened 2017-09-01
Remove some clones that are no longer needed after render task refactor.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1651)


glennw:
r? anyone

PR #1660, opened 2017-09-05
Introduce clip.rs, and tidy up some clip related code.

glennw:
* Move some of the clip types into clip.rs.
* Introduce ClipSources, which contains both a list of ClipSource items, along with the MaskCacheInfo.
* Remove unused screen_inner_rect from ClipScrollNode.
* Remove some Clone derives that are no longer needed.
* Move the image mask request code into shared update method.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1660)


kvark:
Looks somewhat cleaner, thank you!
@bors-servo r+

PR #1666, opened 2017-09-05
Fix box shadow regression.

glennw:
The box shadow shader vertex format was incorrectly being set
to primitive instances instead of the new vertex format.

On some implementations, this happened to work correctly. On
mac, the default vertex attribute binding meant that this bug
showed up there.

This change removes the vertex format enum, which is redundant
with the vertex array kind enum, and fixes the box shadow shader
to bind to the correct vertex format.

As a follow up, I'll make the shader compilation code assert
if a vertex attribute binding fails.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1666)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -226,18 +226,13 @@ const DESC_CACHE_BOX_SHADOW: VertexDescriptor = VertexDescriptor {
     ]
 };
 
+#[derive(Debug, Copy, Clone)]
 enum VertexArrayKind {
     Primitive,
     Blur,
     Clip,
 
     CacheBoxShadow,
-}
-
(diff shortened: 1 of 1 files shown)


glennw:
r? @kvark 

cc @Gankro @jrmuizel

glennw:
This fixes the recent box shadow issues on mac. It was a genuine bug, but the default attribute binding order meant the bug didn't show up on Linux/Intel or Mesa.

kvark:
Thanks for tracking this down!
@bors-servo r+

glennw:
I ... don't know how we ended up with two enums. Must have been asleep that day :)

PR #1667, opened 2017-09-06
Change internal clip sources to be a simple array of clips.

glennw:
Instead of maintaining a copy of a ClipRegion etc, flatten
any input clip structures into a list of simple clips.

This simplifies some of the mask cache code, and is also
prep work for some further simplifications and optimizations
to how we handle the mask cache structures.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1667)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{ComplexClipRegion, ImageMask, ImageRendering};
+use api::{BorderRadius, ComplexClipRegion, ImageMask, ImageRendering};
 use api::{LayerPoint, LayerRect, LayerToWorldTransform, LocalClip};
 use border::BorderCornerClipSource;
 use gpu_cache::GpuCache;
@@ -49,20 +49,6 @@ impl ClipRegion {
         };
         ClipRegion::create_for_clip_node(*local_clip.clip_rect(), complex_clips, None)
     }
-
-    pub fn create_for_local_clip(local_clip: &LocalClip) -> ClipRegion {
-        let complex_clips = match local_clip {
-            &LocalClip::Rect(_) => Vec::new(),
-            &LocalClip::RoundedRect(_, ref region) => vec![region.clone()],
-        };
-
-        ClipRegion {
-            origin: LayerPoint::zero(),
-            main: *local_clip.clip_rect(),
-            image_mask: None,
-            complex_clips,
-        }
-    }
 }
 
 #[repr(C)]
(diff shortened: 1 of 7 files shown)


glennw:
r? @kvark

I think this is a net improvement, but it may not be super clear what the benefits are until the follow ups are completed. So we can leave this sit until the rest is done, if you prefer.

The main bit to check is the mask_cache ```update``` method. I think it's correct, and tests seem OK, but that code is a bit subtle so I might have missed something.

glennw:
Servo and wrench tests are OK. Gecko try:

https://hg.mozilla.org/try/rev/4434224b2b0b2b310b6bd58fcac6caae23d4935a
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4434224b2b0b2b310b6bd58fcac6caae23d4935a

webrender/src/clip.rs
```diff
@@ -83,17 +69,36 @@ impl Not for ClipMode {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub enum ClipSource {
-    Complex(LayerRect, f32, ClipMode),
-    Region(ClipRegion),
+    Rectangle(LayerRect),
```
kvark: I'm glad we are getting rid of the complex clips :)

webrender/src/mask_cache.rs
```diff
@@ -114,17 +114,16 @@ impl MaskCacheInfo {
         // and if we have an image mask.
         for clip in clips {
             match *clip {
-                ClipSource::Complex(..) => {
+                ClipSource::RoundedRectangle(..) => {
                     complex_clip_count += 1;
```
kvark: I suppose we need to rework the `complex_clip_count`, at least rename it

webrender/src/mask_cache.rs
```diff
@@ -149,46 +148,39 @@ impl MaskCacheInfo {
...
                                                      LayerSize::new(2.0 * MAX_CLIP, 2.0 * MAX_CLIP)));
-            let mut local_inner: Option<LayerRect> = None;
+            let mut local_inner = local_rect;
             let mut has_clip_out = false;
             let has_border_clip = !self.border_corners.is_empty();
 
             for source in sources {
                 match *source {
-                    ClipSource::Complex(rect, radius, mode) => {
+                    ClipSource::Image(ref mask) => {
```
kvark: now that we have 1:1 relationship between `local_rect`/`local_inner` modification and a `source` item, we may go fancy and use `fold()` here ;)

kvark:
@glennw 
>  it may not be super clear what the benefits are until the follow ups are completed

oh no, sir, it's totally clear that this is a good simplification to me
Thank you!

@bors-servo r+

kvark:

@bors-servo retry

Issue #1646, opened 2017-08-31
Handle the zero length of data for external raw data.


JerryShih:
@kvark @glennw @nical 
If we pass a empty slice to the wr, we will pass a nullptr data to device.update_pbo_data()[1]. Then, we might see the previous content in texture cache for that area.
In this case, I would like to create a small 1x1 texture and blit this texture to the corresponding texture cache area for initialization.

[1]
https://github.com/servo/webrender/blob/91671498bc1b454d9e271df79ca41fc94d3ef309/webrender/src/renderer.rs#L1662

kvark:
>If we pass a empty slice to the wr, we will pass a nullptr data to device.update_pbo_data()[1]

Could you expand on the specific of where and how you pass the empty slice exactly?

I don't think `update_pbo_data` ever gets a nullptr. According to [playground](https://play.rust-lang.org/?gist=12c5a0a0c3d888e459b76dcda122f065&version=stable), an empty slice pointer resolves to 1.

JerryShih:
@kvark 
Sorry, not the addr but the size.
https://play.rust-lang.org/?gist=72e634c90b5b56dae46532a7eca48205&version=stable

JerryShih:
Since the size is 0, we will not upload any data into that pbo. Then, we will see the old content for that area.
Gecko might pass a nullptr to the ExternalImageHandler. Then, we construct a empty slice for this nullptr and pass it to update_pbo_data().

kvark:
@JerryShih 

Why would Gecko pass zero data into `ExternalImageSource::RawData` ?

Clearly, WR is missing any checks to ensure that the amount of external data matches the transfer operation with do with PBO. I'd just add an assert/error in there, but you are saying that this case (with zero size) is legitimate?

nical:
I think that @JerryShih  is referring to an error scenario where we fail to map the shared memory in the gecko side and end up with an external image callback that doesn't have anything to give to webrender.

I think that the use case is legit, although it could probably be worked around in gecko (by allocating a small buffer and pass ing a pointer to that instead of the shared memory). From where I stand (from far away, so take with a grain of salt) it would feel cleaner to me if we had a way for the external image callback to fail (doesn't need to be zero size) and handle that in webrender (probably just not render the thing or sample from a placeholder texture). 

Maybe Jerry also has other use cases in mind.

kvark:
[...]

PR #1668, opened 2017-09-06
Handle the failed case of ext-image callback.

JerryShih:
@kvark @nical 
#1646
This patch try to create a local buffer for the pbo initialization.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1668)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1907,6 +1907,16 @@ impl Renderer {
                                     ExternalImageSource::RawData(data) => {
                                         self.device.update_pbo_data(&data[offset as usize..]);
                                     }
+                                    ExternalImageSource::Invalid => {
+                                        // Create a local buffer to fill the pbo.
+                                        let bpp = texture.get_bpp();
+                                        let width = stride.unwrap_or(rect.size.width * bpp);
+                                        let total_size = width * rect.size.height;
+                                        // WR haven't support RGBAF32 format in texture_cache, so
+                                        // we use u8 type here.
+                                        let dummy_data: Vec<u8> = vec![255; total_size as usize];
+                                        self.device.update_pbo_data(&dummy_data);
+                                    }
                                     _ => panic!("No external buffer found"),
                                 };
                                 handler.unlock(id, channel_index);
@@ -2432,6 +2442,11 @@ impl Renderer {
 
                 let texture = match image.source {
                     ExternalImageSource::NativeTexture(texture_id) => ExternalTexture::new(texture_id, texture_target),
+                    ExternalImageSource::Invalid => {
+                        warn!("Invalid ext-image for ext_id:{:?}, channel:{}.", ext_image.id, ext_image.channel_index);
+                        // Just use 0 as the gl handle for this failed case.
+                        ExternalTexture::new(0, texture_target)
+                    }
                     _ => panic!("No native texture found."),
                 };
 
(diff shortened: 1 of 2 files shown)


webrender/src/renderer.rs
```diff
@@ -1908,6 +1908,16 @@ impl Renderer {
...
                                     }
+                                    ExternalImageSource::Invalid => {
+                                        // Create a local buffer to fill the pbo.
+                                        let bpp = texture.get_bpp();
+                                        let width = stride.unwrap_or(rect.size.width);
+                                        let total_size = width * rect.size.width * bpp;
+                                        // WR haven't support RGBAF32 format in texture_cache, so
+                                        // we use u8 type here.
+                                        let dummy_data: Vec<u8> = vec![255; total_size as usize];
+                                        self.device.update_pbo_data(&dummy_data);
```
JerryShih: I'm not sure whether we have alignment problem for the glBufferData() call here.

kvark: > Clients must align data elements consistently with the requirements of the client platform, with an additional base-level requirement that an offset within a buffer to a datum comprising N bytes be a multiple of N.

JerryShih: @kvark 
I have read these from the document, but I'm not sure I understand the meaning here.
For modern，data alignment could be 1, 2, 4 and more. So, what's the "requirement of the client platform"? And could we set the alignment hint for vector's buffer allocation?

kvark: I wasn't able to find a concrete answer to that question. It might be just bad wording on the specification side, as noted by [this SO answer](https://gamedev.stackexchange.com/a/54876/6426).
In rust, there is `repr(align = XXX)` attribute but it's experimental and doesn't work in stable channel...

My conclusion is that the alignment is only required for data that is going to be directly accessed by GPU and/or DMA operations, which your `glBufferData` one is *not*. This data gets copied by the driver in the appropriate place, so alignment requirements don't really apply to your data source.

glennw:
@kvark Is this good to go then, from your perspective?

webrender/src/renderer.rs
```diff
@@ -1907,6 +1907,16 @@ impl Renderer {
                                     ExternalImageSource::RawData(data) => {
                                         self.device.update_pbo_data(&data[offset as usize..]);
                                     }
+                                    ExternalImageSource::Invalid => {
+                                        // Create a local buffer to fill the pbo.
+                                        let bpp = texture.get_bpp();
+                                        let width = stride.unwrap_or(rect.size.width * bpp);
```
JerryShih: Sorry, I made a mistake for the image width.
The stride already contains the bpp.

webrender/src/renderer.rs
```diff
@@ -1907,6 +1907,16 @@ impl Renderer {
...
+                                        let total_size = width * rect.size.height;
```
JerryShih: I also have a typo here.
The size should be width * height. Not the width * width.

webrender/src/renderer.rs
```diff
@@ -2432,6 +2442,11 @@ impl Renderer {
 
                 let texture = match image.source {
                     ExternalImageSource::NativeTexture(texture_id) => ExternalTexture::new(texture_id, texture_target),
+                    ExternalImageSource::Invalid => {
```
kvark: can we just skip/ignore this element here instead of trying to add it to the `external_images` list (that will be unlocked later on, which doesn't make sense for a failed lock in the first place)?
It seems hazardous to me to just proceed with TextureId==0 as if it's OK.

Issue #1650, opened 2017-09-01
Add a page to the web debugger that displays the clip-scroll tree.


glennw:

lqd:
I'd like to take a stab at this :)

glennw:
@lqd Great! I'd suggest starting with looking at ```print_node``` and then porting that to be serializable to a JSON object (see some of the other web debugger pages with the ```debugger``` feature).

After that, it would be a case of writing the Vue template in the debugger to display it (should be able to use the existing pages as a reference).

lqd:
@glennw yeah I have already done the Vue etc — but was wondering about a couple things (the reason why I pinged you on irc) and basically those questions were answered by your mentioning `print_node` :)
Just to be sure, as I'm not that familiar with WR or its clipscroll tree, my thoughts were that this serialization work should be done in the `render_backend`, serializing/traversing the clip_scroll_tree of each document's frame, right ?

glennw:
@lqd Yep, that sounds reasonable - there are a couple of existing commands (the one that serializes the display lists) that could be useful as a reference.

lqd:
This issue was indeed fixed by #1669 :)

PR #1669, opened 2017-09-06
Add a basic clip-scroll tree view to the debugger

lqd:
This allows viewing the frames' clip_scroll_tree data in the debugger, using the existing "pretty-print" display feature.

Since it's ultimately a tree, I thought it had to look more like the documents view rather than the passes (since passes seem only 2-deep, whereas the clip-scroll tree depth is not fixed):

image

toughts @glennw ?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1669)

### DIFF
--- a/debugger/js/app.js
+++ b/debugger/js/app.js
@@ -3,6 +3,7 @@ var state = {
     page: "options",
     passes: [],
     documents: [],
+    clipscrolltree: [],
 }
 
 class Connection {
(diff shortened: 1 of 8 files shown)


lqd:
oh btw @glennw I came across this https://github.com/servo/webrender/blob/master/webrender/src/renderer.rs#L1639 where/how did you get this warning ? (I couldn't see it when I tested without the workaround)

glennw:
Nicely done, thanks!

It's possible that warning was from an earlier iteration of the debug code, and may not be needed any more.

@bors-servo r+

glennw:
Fixes #1650

Issue #1670, opened 2017-09-06
ArsTechnica takes 8 ms of GPU time


jrmuizel:
STR: scroll about halfway down the front pages of ArsTechnica in Gecko with layers-free=true,blob-images=true,webrendest=true

I see this on my Haswell 4770k on Linux running at 1080p:
- 216 visible primitives
- 5 passes
- 5 color targets
- 1 alpha targets

glennw:
Wow, that's a lot of color passes.

The first step will to investigate this will be to enable the render target debugger, and see if anything stands out as part of the 5 color targets that looks obviously weird.

glennw:
It does seem that each of the color targets is basically a clear / draw / composite of the previous target. My guess would be that it's doing blends / composites when it doesn't actually need to.

Of the 8ms GPU time, around 6ms is spent doing those clears and composites which appear to be unnecessary at first glance. I'll dig deeper and find out why it's doing all those copies.

![at](https://user-images.githubusercontent.com/2350882/30141193-1d537314-93bd-11e7-90d4-3691780111c3.png)

glennw:
As a reference point, the Servo GPU time on that page is ~3.7ms on my machine (usual incomplete Servo caveats apply, but it looks very similar).

glennw:
Yup, the debugger shows the issue:

Passes 1 and 2 appear to contain a large number of (hopefully) redundant blends and composites.

There's also a lot of images being drawn in the alpha pass, when they appear to be opaque. There might be a bug there which will be causing slower GPU time than necessary (does Gecko correctly check and pass through whether an image is opaque I wonder?).

```
Pass 0

Target 0 (A8)

Batch 0 (Rectangles, Clip, 16 instances)

Batch 1 (Box Shadow, Cache, 2 instances)

Target 1 (RGBA8)

Batch 0 (Image (Array), Opaque, 6 instances)

Batch 1 (Image (Array), Opaque, 1 instances)

Batch 2 (Image (Array), Alpha, 1 instances)

Pass 1

Target 0 (RGBA8)

Batch 0 (Blend, Alpha, 8 instances)

Pass 2

Target 0 (RGBA8)

Batch 0 (Rectangle, Opaque, 8 instances)

Batch 1 (Composite, Alpha, 1 instances)

Batch 2 (Composite, Alpha, 1 instances)

Batch 3 (Composite, Alpha, 1 instances)

Batch 4 (Composite, Alpha, 1 instances)

[...]

PR #1673, opened 2017-09-07
Fix bug in handling redundant opacity filters.

glennw:
If a filter had a property binding, and that value resolved
to opacity=1, WR can skip applying an expensive filter
operation on that stacking context.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1673)


glennw:
r? @kvark 

This helps quite a bit with #1670.

PR #1679, opened 2017-09-08
Various simplifications, optimizations and features related to clipping.

glennw:
* Introduce ClipStore, an arena for storing clip sources.
    This reduces a lot of memory copying and allocations, since the clip stack built each frame now just contains handles into the clip store, rather than copying the entire mask cache.
* Move remaining mask cache code into ClipSources, and simplify it.
    We no longer need to copy the border sources etc into the mask cache structures - they can be referenced directly from the clip sources.
* Use a separate GPU cache handle for each clip, so ClipAddressRange code is no longer needed.
    This is possible since we use the GpuCache now, which doesn't need to allocate space during initialization.
* Simplify the code in the clip batcher that adds clip instances to the render targets.
* Add support for multiple image masks (a by-product of removing the init step of the mask cache code).
* Move MaskBounds into the ClipSources (from the MaskCacheInfo struct).
* Add recycle() to free list, and support cloning weak handles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1679)


glennw:
r? @kvark 

https://hg.mozilla.org/try/rev/c67da9f751701d22a7a7b7c2b9ea4d024b99f661
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c67da9f751701d22a7a7b7c2b9ea4d024b99f661

glennw:
This change is probably easier to review by considering each commit as a separate improvement...

webrender/src/clip.rs
```diff
@@ -109,18 +109,16 @@ impl From<ClipRegion> for ClipSources {
 
 #[derive(Debug)]
 pub struct ClipSources {
-    clips: Vec<ClipSource>,
-    pub mask_cache_info: MaskCacheInfo,
+    pub clips: Vec<ClipSource>,
+    pub handles: Vec<GpuCacheHandle>,
```
kvark: why don't we want to either simply include `GpuCacheHandle` into `ClipSource` or have it stored nearby, e.g. `Vec<(ClipSrouce, GpuCacheHandle)>`?

glennw: I did consider including it in ```ClipSource``` but it seemed a bit cleaner conceptually to separate them. No real reason it couldn't be in a tuple though!

kvark: Great, let's tuple it up then ;)

glennw:
@kvark OK, pushed a commit that moves ```handles``` into a tuple in ```clips```.

kvark:
Thank you!
@bors-servo r+

PR #1684, opened 2017-09-08
Switch appveyor from Rust nightly to 1.19

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1684)


larsbergstrom:
Did you want to do both instead? Is there a reason you don't also want testing against nightly?

cc @glennw @edunham

kvark:
Doing both sounds good to me.

PR #1685, opened 2017-09-08
Don't install freetype on Windows.

jrmuizel:
I don't think we need it for anything anymore

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1685)


emilio:
@bors-servo r+

Thanks!

PR #1687, opened 2017-09-11
Introduce frame output support.

glennw:
This adds a new feature to WR, that allows copying the output
of a given pipeline ID to an external texture. This is a feature
request from the webvr prototyping team.

The included example demonstrates how to use the API to copy the
output of an iframe to an external texture, which is then copied
onto the screen via custom GL commands.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1687)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -20,7 +20,7 @@ use debug_render::DebugRenderer;
 use debug_server::{self, DebugServer};
 use device::{DepthFunction, Device, FrameId, Program, Texture, VertexDescriptor, GpuMarker, GpuProfiler, PBO};
 use device::{GpuTimer, TextureFilter, VAO, VertexUsageHint, FileWatcherHandler, TextureTarget, ShaderError};
-use device::{ExternalTexture, get_gl_format_bgra, TextureSlot, VertexAttribute, VertexAttributeKind};
+use device::{ExternalTexture, FBOId, get_gl_format_bgra, TextureSlot, VertexAttribute, VertexAttributeKind};
 use euclid::{Transform3D, rect};
 use frame_builder::FrameBuilderConfig;
 use gleam::gl;
@@ -37,6 +37,7 @@ use render_task::RenderTaskTree;
 use serde_json;
 use std;
 use std::cmp;
+use std::collections::hash_map::Entry;
 use std::collections::VecDeque;
 use std::f32;
 use std::mem;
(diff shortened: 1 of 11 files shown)


glennw:
r? @kvark 

cc @MortimerGoro

webrender/src/renderer.rs
```diff
@@ -2679,17 +2746,19 @@ impl Renderer {
         }
 
         for (i, texture) in self.color_render_targets.iter().chain(self.alpha_render_targets.iter()).enumerate() {
+            let dimensions = texture.get_dimensions();
+            let src_rect = DeviceIntRect::new(DeviceIntPoint::zero(),
+                                              DeviceIntSize::new(dimensions.width as i32,
```
kvark: nit: could probably do some casting of `euclid::TypedSize2D`

webrender/src/renderer.rs
```diff
@@ -2860,6 +2938,16 @@ pub trait ExternalImageHandler {
     fn unlock(&mut self, key: ExternalImageId, channel_index: u8);
 }
 
+/// Allows callers to receive a texture with the contents of a specific
+/// pipeline copied to it. Lock should return the native texture handle
+/// and the size of the texture. Unlock will only be called if the lock()
+/// call succeeds, when WR has issued the GL commands to copy the output
+/// to the texture handle.
+pub trait OutputImageHandler {
+    fn lock(&mut self, pipeline_id: PipelineId) -> Option<(u32, DeviceIntSize)>;
```
kvark: we might want to revisit the lock/unlock semantics here later (is that how VR team wants to work with WR?), especially since WR pretty much guarantees that `unlock` is going to be called right away after some blitting (contrary to external image locks, where unlocking can happen at any time later).

kvark:
Oh, and I forgot to mention that this would be really nice to hook up via the document API: having VR content in a separate WR document that's routed into a separate target, as opposed to hacking the `PipelineId` handling.

webrender/src/renderer.rs
glennw: The vr team have a demo working with the lock/unlock - but yes, we may want to revisit these semantics later on once we work out the exact synchronization story.

glennw:
@kvark I would like to hook up to the document API too - I think the pipeline API is fine for now (allows easy prototyping in Servo). But certainly once we make better use of the document API in Servo, we could consider switching frame output to support both documents and pipelines, or switch to documents exclusively.

glennw:
Rebased and fixed cast nit.

kvark:
Ok, merging on the merit that this is going to be ported to document API later on (or at least considered :) )
@bors-servo r+

glennw:
Rebased.

@bors-servo r=kvark

PR #1689, opened 2017-09-11
Handle text runs that need split batches, and reduce allocations.

glennw:
Previously, both the normal text run and text shadow code assumed
that all the glyphs in a single text run were placed in the same
texture. This is generally true, but not *always*. In particular,
a text run that contains a glyph that fits in the texture cache, and
one that is too large for the texture cache would trigger this condition.

Now, create a text shadow batch per texture ID. For normal text runs,
flush and create a new batch each time a glyph is encountered that
belongs to a different source texture.

Also, remove an allocation per text run by using a shared buffer
to store glyphs as they are fetched for a text run. This is a bit
untidy - the caller must provide the fetch buffer since the resource
cache is considered immutable at this point (batching may be
spread across threads in the future). However, since this is such
a hot code path on most pages, it's worth it (typically a 10% CPU
time saving on pages I tested).

The reftests just need to not crash - the content is identical.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1689)

### DIFF
--- a/webrender/src/internal_types.rs
+++ b/webrender/src/internal_types.rs
@@ -74,6 +74,14 @@ impl BatchTextures {
             ]
         }
     }
+
+    pub fn color(texture: SourceTexture) -> Self {
+        BatchTextures {
+            colors: [ texture,
+                      SourceTexture::Invalid,
+                      SourceTexture::Invalid ]
+        }
+    }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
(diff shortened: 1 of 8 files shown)


JerryShih:
With this, do we still need the https://github.com/servo/webrender/pull/1597?

glennw:
Yup, we still need #1597 - that solves a different issue.

@bors-servo retry

glennw:
@bors-servo retry
 * more ssl builder issues

PR #1691, opened 2017-09-12
Use strong types for instance vertex structures, and fix some inconsistencies.

glennw:
This just makes naming a bit more consistent, uses strong types in some of the instance structures, and moves a bit more code out of ```tiling.rs```. The main benefit of using strong types here is to simplify some follow up work to compress these instance fields (e.g. start using u16 instead of i32 where possible, for layer and render task addresses).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1691)


nical:
Nice!

@bors-servo r+

PR #1692, opened 2017-09-12
Add a debug flag to see alpha primitive rects.

nical:
I extracted @glennw's [quick hack to see alpha primitive rects](https://github.com/glennw/webrender/commit/b5f36cd3a83ce622411c62543393328ef1566ffa) and added it to the debug flags (as ALPHA_PRIM_DBG) because I think this is a useful tool to have.

r? @kvark or @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1692)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -127,6 +127,7 @@ bitflags! {
         const PROFILER_DBG      = 1 << 0;
         const RENDER_TARGET_DBG = 1 << 1;
         const TEXTURE_CACHE_DBG = 1 << 2;
+        const ALPHA_PRIM_DBG    = 1 << 3;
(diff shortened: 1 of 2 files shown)


webrender/src/renderer.rs
```diff
@@ -2303,6 +2304,13 @@ impl Renderer {
                     prev_blend_mode = batch.key.blend_mode;
                 }
 
+                if self.debug_flags.contains(ALPHA_PRIM_DBG) {
+                    let color = ColorF::new(1.0, 0.0, 0.0, 1.0).into();
```
kvark: we could change the color based on what `BlendMode` is used

PR #1697, opened 2017-09-12
Fix SSL errors on travis CI.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1697)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,7 +15,7 @@ dependencies = [
  "gleam 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "image 0.14.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


glennw:
Not for merging - just testing right now.

glennw:
@bors-servo retry
 * seemingly unrelated error (memory?)

glennw:
@bors-servo r=metajack

Also includes an update to ```osmesa-src``` which seems to fix an inexplicably new c++ compiler error. The ```osmesa``` version also matches what we are using in Servo now, which is a good thing anyway.

Cargo.lock
```diff
@@ -655,8 +655,8 @@ dependencies = [
 
 [[package]]
 name = "osmesa-src"
-version = "12.0.1"
-source = "git+https://github.com/servo/osmesa-src#42509fbd70d0e54e1278849b771ea1cb1c3386f4"
+version = "17.2.0-devel"
```
kvark: oh wow, I wonder if we can use LLVM4 now

Issue #1351, opened 2017-06-08
Implement blur filter


mrobinson:
This can be observed by running the following wrench test:
```yaml
root:
  items:
    - type: stacking-context
      bounds: [10, 10, 500, 500]
      filters: blur(5)
      items:
        - type: rect
          bounds: [10, 10, 50, 50]
          color: green
```

cc @glennw @jrmuizel

glennw:
It's not implemented yet, that's why :) Most of the work is done in other areas, so it shouldn't be a huge amount of work.

There is a (slow) blur render task and shader that is used for text shadow blurs.

There is also some code to render a stacking context to a target with identity transform.

We basically need to hook up / abstract those bits of code, to get the blur filter working.

We also need to (at some point) write an optimized blur shader that doesn't have awful performance, but that can be done later.

glennw:
This happened.

PR #1699, opened 2017-09-13
Implement blur filter.

glennw:
In addition to implementing the blur filter, this PR adds support for intermediate targets that are larger than the frame buffer.

Now, we track the largest render task size requirement, and base the size of the render target array texture off this. This is only rarely needed - e.g. when a blur has a large radius. Normally, we can just clip the contents of render tasks to the visible screen size (which is what all other render tasks do).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1699)


glennw:
Fixes #1351.

glennw:
r? @kvark 

No rush on reviewing this PR - it passes the Servo and wrench tests, but I still need to do a gecko try run.

glennw:
Note that the output of the blur shader doesn't exactly match the gecko blur, yet. But I think it's in a reasonable state to add the initial support, and then work on making the blur algorithm match.

glennw:
OK, gecko try looks good I think (2 remaining jobs):

https://hg.mozilla.org/try/rev/d21a576c342ae7f7cbf1be28d21be43ef573f7f6
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d21a576c342ae7f7cbf1be28d21be43ef573f7f6

webrender/src/tiling.rs
```diff
@@ -1156,22 +1162,44 @@ pub struct RenderPass {
...
     }
 
-    pub fn add_render_task(&mut self, task_id: RenderTaskId) {
+    pub fn add_render_task(&mut self,
+                           task_id: RenderTaskId,
+                           size: DeviceIntSize,
+                           target_kind: RenderTargetKind) {
+        match target_kind {
+            RenderTargetKind::Color => {
+                self.max_color_target_size.width = cmp::max(self.max_color_target_size.width,
```
kvark: @nical - here is a case for missing on `TypedSize2D::max()`

wrench/headless.py
```diff
@@ -67,4 +67,9 @@ def set_osmesa_env(bin_path):
 
 subprocess.check_call(['cargo', 'build', '--release', '--verbose', '--features', 'headless'])
 set_osmesa_env('../target/release/')
-subprocess.check_call(['../target/release/wrench', '-h'] + sys.argv[1:])
+# TODO(gw): We have an occasional accuracy issue or bug (could be WR or OSMesa)
+#           where the output of a previous test that uses intermediate targets can
+#           cause 1.0 / 255.0 pixel differences in a subsequent test. For now, we
```
kvark: somewhere, UVs are not clamped properly before sampling?..

wrench/src/premultiply.rs
```diff
@@ -6,9 +6,9 @@
 pub fn premultiply(data: &mut [u8]) {
     for pixel in data.chunks_mut(4) {
         let a = pixel[3] as u32;
-        let r = pixel[2] as u32;
+        let b = pixel[2] as u32;
```
kvark: is this change intentional? it doesn't match the `unpremultiply` code below

wrench/headless.py
glennw: I did a quick scan of the shaders involved and didn't see anything obvious, but I agree this is the most likely candidate.

wrench/src/premultiply.rs
glennw: Yep - without this change the colors in the Firefox image logo I am using in the tests is wrong. I updated ```unpremultiply``` to match.

glennw:
@kvark Thanks for the review - comments addressed.

kvark:
Thank you!
@bors-servo r+

PR #1701, opened 2017-09-13
Make StickySideConstraint repr(C) for use in gecko

staktrace:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1701)


glennw:
@bors-servo retry
 * Looks like OOM.

PR #1708, opened 2017-09-15
Add some factories for PrimitiveInfo

mrobinson:
This will make it easier to add new members in the future (for hit
testing). We also make local_clip a LocalClip instead of an
Option<LocalClip> because this should never be None.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1708)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -195,11 +195,7 @@ impl Example for App {
         _document_id: DocumentId,
     ) {
         let bounds = LayoutRect::new(LayoutPoint::zero(), layout_size);
-        let info = LayoutPrimitiveInfo {
-            rect: bounds,
-            local_clip: None,
-            is_backface_visible: true,
-        };
+        let info = LayoutPrimitiveInfo::new(bounds);
         builder.push_stacking_context(
             &info,
             ScrollPolicy::Scrollable,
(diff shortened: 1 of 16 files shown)


webrender/examples/basic.rs
```diff
@@ -334,12 +313,7 @@ impl Example for App {
             let spread_radius = 0.0;
             let simple_border_radius = 8.0;
             let box_shadow_type = BoxShadowClipMode::Inset;
-            let info = LayoutPrimitiveInfo {
-                rect: rect,
-                local_clip: Some(LocalClip::from(bounds)),
-                is_backface_visible: true,
-            };
-
+            let info = LayoutPrimitiveInfo::new_with_clip_rect(rect, bounds);
```
kvark: should probably just be `with_clip_rect`?

webrender/examples/blob.rs
```diff
@@ -253,11 +249,7 @@ impl Example for App {
                                       api::MixBlendMode::Normal,
                                       Vec::new());
 
-        let info = api::LayoutPrimitiveInfo {
-            rect: (30, 30).by(500, 500),
-            local_clip: Some(api::LocalClip::from(bounds)),
-            is_backface_visible: true,
-        };
+        let info = api::LayoutPrimitiveInfo::new((30, 30).by(500, 500));
```
kvark: missing local clip?

webrender/examples/blob.rs
```diff
@@ -266,11 +258,7 @@ impl Example for App {
             blob_img1,
         );
 
-        let info = api::LayoutPrimitiveInfo {
-            rect: (600, 600).by(200, 200),
-            local_clip: Some(api::LocalClip::from(bounds)),
-            is_backface_visible: true,
-        };
+        let info = api::LayoutPrimitiveInfo::new((600, 600).by(200, 200));
```
kvark: missing local clip?

webrender/examples/basic.rs
mrobinson: Okay. I'll make this change.

webrender/examples/blob.rs
mrobinson: I'll fix this.

webrender/examples/blob.rs
mrobinson: Ditto!

mrobinson:
Okay. I've updated the branch with the changes and rebased it on top of the rustfmt changes.

kvark:
Thanks @mrobinson , and sorry about the rebase pain!
@bors-servo r+

staktrace:
@mrobinson This broke the gecko build, please see https://bugzilla.mozilla.org/show_bug.cgi?id=1400216#c5 and provide a gecko-side patch. Thanks!

mrobinson:
@staktrace Thanks for the heads up. I've posted a patch to Bugzilla which should fix the build.

PR #1709, opened 2017-09-15
Add reftest for shadow colors being added

Gankro:
Current master has a bug where we seem to be blending shadows in by adding the color, which isn't correct.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1709)


glennw:
Closed in favor of https://github.com/servo/webrender/pull/1716, which includes this commit.

Issue #1655, opened 2017-09-01
Implement font-variation support (needed for -apple-system fonts)


Gankro:
If you navigate to a github issue that mentions a user on macos with gecko, it should be bold:

Hello @Gankro this is a test

But with webrendest layers-free (which hands the text to webrender) it doesn't show up as bold. I can't find anything going wrong in gecko here (it seems to be natively bold), so my current hunch is that we're processing this font incorrectly somehow. It's possible there's extra information we need gecko to send us? I don't know enough about CGFont stuff to say.

Gankro:
Discussing with @lsalzman, it seems like this is probably the result of not sending font variations. Needs gecko and webrender work.

Gankro:
Twin bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1397458

Gankro:
Fixed!

PR #1714, opened 2017-09-18
support font variations for font instances

lsalzman:
This adds support for font variations in an effort to correct the problem identified in https://github.com/servo/webrender/issues/1655 and https://bugzilla.mozilla.org/show_bug.cgi?id=1397458

Apple system fonts may use font variations internally to identify modifications of a base raw font for styling, namely bold. To correctly supply these fonts to WebRender, we need to not only send over the raw font, but any internal font variation information. This makes this possible to receive it on the WebRender side, in cooperation with changes on the Gecko side to supply it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1714)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ addons:
       - libedit-dev
       - python
 env:
-  - BUILD_KIND=DEBUG RUST_BACKTRACE=1 RUSTFLAGS='--deny warnings'
-  - BUILD_KIND=RELEASE RUST_BACKTRACE=1 RUSTFLAGS='--deny warnings'
+  - BUILD_KIND=DEBUG RUST_BACKTRACE=1
+  - BUILD_KIND=RELEASE RUST_BACKTRACE=1
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install zlib; fi
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,7 +21,7 @@ dependencies = [
  "serde_json 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-glutin 0.11.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.36 (registry+https://github.com/rust-lang/crates.io-index)",
- "webrender 0.50.0",
+ "webrender 0.51.0",
  "yaml-rust 0.3.4 (git+https://github.com/vvuk/yaml-rust)",
 ]
 
@@ -1037,13 +1037,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "webrender"
-version = "0.50.0"
+version = "0.51.0"
 dependencies = [
  "angle 0.5.0 (git+https://github.com/servo/angle?branch=servo)",
  "app_units 0.5.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "bincode 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitflags 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "core-foundation 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-text 6.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "dwrote 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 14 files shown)


Gankro:
rustfmt'd... bummer :(

glennw:
Reviewed 12 of 13 files at r1.
Review status: 12 of 13 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[webrender/src/platform/macos/font.rs, line 181 at r1](https://reviewable.io:443/reviews/servo/webrender/1714#-KuIJxUHzau4NKQU17Vm:-KuIJxUI7pn-_IQve750:b6x5q3z) ([raw file](https://github.com/servo/webrender/blob/f7566cf039eab160688bacc864e55fc22187c6e4/webrender/src/platform/macos/font.rs#L181)):*
> ```Rust
>             let axis: CFDictionary = TCFType::wrap_under_get_rule(axis_ptr as CFDictionaryRef);
>             if !axis.instance_of::<CFDictionaryRef, CFDictionary>() {
>               return ct_font;
> ```

nit: indentation

---

*[webrender/src/platform/macos/font.rs, line 249 at r1](https://reviewable.io:443/reviews/servo/webrender/1714#-KuIKFzM-sSmz1UJm9gk:-KuIKFzM-sSmz1UJm9gl:b-2clcei) ([raw file](https://github.com/servo/webrender/blob/f7566cf039eab160688bacc864e55fc22187c6e4/webrender/src/platform/macos/font.rs#L249)):*
> ```Rust
>             };
> 
>             if val < min_val {
> ```

use stdlib min/max here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1714)*

glennw:
Reviewed 1 of 13 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1714)*

glennw:
In general this looks good to me. I know almost nothing of the mac interfaces - is there anything we need to be careful of in that unsafe code regarding leaks?

It's going to need a rebase too, sorry.

@kvark Do you know much about mac APIs? :)

kvark:
@glennw no, unfortunately, I don't know a dime about OSX APIs (well, other than Metal, but that doesn't count;) )

lsalzman:
Review status: 3 of 13 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/platform/macos/font.rs, line 181 at r1](https://reviewable.io:443/reviews/servo/webrender/1714#-KuIJxUHzau4NKQU17Vm:-KuLKUZt9MNIayMtvnYf:b-lx4fjp) ([raw file](https://github.com/servo/webrender/blob/f7566cf039eab160688bacc864e55fc22187c6e4/webrender/src/platform/macos/font.rs#L181)):*

nit: indentation
</blockquote></details>

Fixed.

---

*[webrender/src/platform/macos/font.rs, line 249 at r1](https://reviewable.io:443/reviews/servo/webrender/1714#-KuIKFzM-sSmz1UJm9gk:-KuLKVpSs0WpoR3vRwfM:bbx1gzx) ([raw file](https://github.com/servo/webrender/blob/f7566cf039eab160688bacc864e55fc22187c6e4/webrender/src/platform/macos/font.rs#L249)):*

use stdlib min/max here?
</blockquote></details>

Can't, because those require their arguments to satisfy Ord, but floats do not.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1714)*

lsalzman:
@glenww, in principle all that TCFType wrapping that is going on is dealing with the underlying reference counts of those objects, via the wrap_under_(get/create)_rule and the Drop trait. In fact, it is actually doing more reference counting than is needed in those cases (harmless, but maybe a small performance nit). Otherwise, the code is a fairly faithful to what the Gecko side is doing here:  https://dxr.mozilla.org/mozilla-central/source/gfx/2d/ScaledFontMac.cpp#361

glennw:
@lsalzman For the ```min/max``` there are methods implemented on the ```f32``` and ```f64``` types, since they don't satisfy Ord. But we can do that as a follow up, it's minor.

@bors-servo r+

lsalzman:
Cleaned up the min/max usage and updated the core-graphics version to fix the CGFontRef zero-sized struct error.

glennw:
Ugh, dependency issues with multiple core-graphics versions - I'll take a look locally and see what needs to be updated :/

glennw:
Looks like we will need a heap of dependency bumps in other repos to land this :( Waiting to hear back on https://github.com/servo/cocoa-rs/pull/170, then I'll get that updated and start updating the other repos.

glennw:
@lsalzman @kvark @staktrace I think it's going to take a while (maybe 2-4 days?) for me to get the dependency version bumps filtered through all the mac crates. Instead, I think if we revert the core text version update, and allow warning on CI builds (temporarily) we could land this today. Then we could re-enable deny-warnings after I get all the versions bumped. Thoughts?

staktrace:
@glennw I don't have a strong opinion either way. I'm not sure if we need to hurry this along - it sounds like you have to do the version bumps anyway so I'd lean towards just waiting until that's done. If people are blocked on this and have nothing else to do, then I don't have any objections to landing it sooner.

glennw:
@staktrace OK. @lsalzman has disabled the warnings - let's see if that passes CI. If it does, we'll merge this now well I work on getting the dependency updates done. Thanks!

lsalzman:
Sigh, not sure how the FontVariation import got lost. Must have happened during the rustfmt-pocalypse somehow. Fixed now.

glennw:
Merging manually due to CI issues.

PR #1717, opened 2017-09-18
Fix frame output garbage collection

MortimerGoro:
Frame output FBOs are regenerated each frame causing OOM issues. This is caused because the output_targets.retain() function is not correctly implemented

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1717)


glennw:
Oops!

@bors-servo r+

kvark:
Uh oh, that's quite easy to overlook. Gotta be careful with those `bool` types (note to myself for reviewing next time).

PR #1719, opened 2017-09-18
Enhance the position:sticky example to demonstrate top and bottom sti…

staktrace:
…ckiness on a single element.

r? @mrobinson 

I wrote this to get a better idea of how the sticky API works when dealing with multiple sticky properties on a single element, figured we might as well land it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1719)


glennw:
Merging manually due to travis issues. Boom.

PR #1720, opened 2017-09-19
Allow modifying the device pixel ratio at runtime.

### LINKED ISSUE #1616: Way to change device pixel ratio at runtime
On macOS, when a window is dragged between a Retina display and an external non-Retina display or vice-versa, the device pixel ratio changes. I don't see any API to update Webrender's pixel ratio after creating the renderer.

Is the correct solution to recreate the renderer when the device pixel ratio changes, or to add a feature to Webrender to update the ratio?

glennw:
Add an extra parameter to the SetWindowParameters message. This is
useful for situations such as when a window is dragged from one
monitor to another monitor that has a different device pixel ratio.

Fixes #1616.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1720)

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -216,6 +216,28 @@ pub fn main_wrapper(example: &mut Example, options: Option<webrender::RendererOp
                 glutin::Event::KeyboardInput(
                     glutin::ElementState::Pressed,
                     _,
+                    Some(glutin::VirtualKeyCode::Key1),
+                ) => {
+                    api.set_window_parameters(document_id,
+                        size,
+                        DeviceUintRect::new(DeviceUintPoint::zero(), size),
+                        1.0
+                    );
+                }
+                glutin::Event::KeyboardInput(
+                    glutin::ElementState::Pressed,
+                    _,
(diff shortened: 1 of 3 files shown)


glennw:
r? @nical or @kvark 

@staktrace This adds a new parameter to the ```set_window_parameters``` API. I'm not sure if Gecko has that in bindings yet. The extra parameter is an f32 that is the device pixel ratio for the window.

staktrace:
@glennw It does, yes. Should I just hard-code that to 1.0 for now? It's not clear to me from the patch what the semantics of the device pixel ratio are. Gecko has multiple factors (widget scaling, page full zoom) that combine to produce the `window.devicePixelRatio` exposed to JS and it's not obvious which of those factors WR is expecting.

nical:
@staktrace I think that corresponds to the widget scaling (webrender has separate values for zoom factors). This parameter replaces the device_pixel_ratio member in RendererOptions that we don't touch and is equal to 1.0 by default so if you hard-code 1.0 it should keep the current behavior.

glennw:
@staktrace @nical Yep, using whatever you currently use in RendererOptions is fine.

PR #1721, opened 2017-09-19
Implement synthetic italics on Linux.

glennw:
Mac and Windows implementations still need to be implemented.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1721)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1065,6 +1065,7 @@ impl FrameBuilder {
             subpx_dir,
             font.platform_options,
             font.variations.clone(),
+            font.synthetic_italics,
         );
         let prim = TextRunPrimitiveCpu {
             font: prim_font,
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -411,6 +411,7 @@ fn raterize_200_glyphs() {
         SubpixelDirection::Horizontal,
         None,
         Vec::new(),
+        false,
     );
 
     let mut glyph_keys = Vec::with_capacity(200);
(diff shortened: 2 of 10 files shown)


glennw:
italics

glennw:
r? @nical or @kvark 

cc @Gankro - this is just the Linux implementation, for now.

glennw:
Rebased.

@bors-servo r=nical

staktrace:
@glennw @nical This broke the gecko build, see https://bugzilla.mozilla.org/show_bug.cgi?id=1401244#c3

The problem appears to be that [Option<FontRenderMode> inside FontInstanceOptions](https://hg.mozilla.org/try/rev/cf2222460d5a282ab33bc3e4e263b7c9e75a33dd#l21.79) is not supported by cbindgen; it generates a dummy "struct Option_FontRenderMode;" in the [generated bindings](https://hg.mozilla.org/try/rev/76e63d91a66a6079e6bba0dc9becbf7a5728e687) but that's not sufficient.

@eqrion Thoughts? Can we make cbindgen support this? ^

eqrion:
An Option of and enum is tricky/unsafe to use in FFI. We'd need something like the work @Gankro was doing in [rustc](https://github.com/eqrion/cbindgen/issues/24)

staktrace:
@nical Should we back this change out then? It's going to block updating gecko to anything past WR 2a005f1

staktrace:
Sure, if you have something in mind to work around this I'm happy to wait a bit.

glennw:
Oh, it didn't even occur to me that cbindgen would have trouble with this. One easy way is I could add a "Default" variant to FontRenderMode, and then we use FontRenderMode::Default instead of None, which would mean it doesn't need to be an option. Want me to make that change?

staktrace:
The patch nical posted to bugzilla seems to solve the problem. I don't have a strong opinion as to which solution is preferable.

glennw:
OK, I'll wait to hear @nical's opinion before making any change.

nical:
The WebRender update landed last Friday, no need to make the FontRenderMode change.

PR #1722, opened 2017-09-19
Fix the brightness and invert filters to work with premultiplied alpha.

### LINKED ISSUE #1507: Brightness filter with transparent contents renders incorrectly
Testcase: https://bugzilla.mozilla.org/attachment.cgi?id=8888896

Both boxes should render the same. They look different in Servo and Gecko+webrender+webrendest.

Screenshot:

screen shot 2017-07-21 at 3 32 04 pm

### LINKED ISSUE #1508: Invert filter with transparent contents renders incorrectly
This is very similar to #1507.

Testcase: https://bugzilla.mozilla.org/attachment.cgi?id=8888897

Both boxes should render the same. They look different in Servo and Gecko+webrender+webrendest.

Screenshot:

screen shot 2017-07-21 at 3 33 49 pm

glennw:
Fixes #1507.
Fixes #1508.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1722)

### DIFF
--- a/webrender/res/ps_blend.glsl
+++ b/webrender/res/ps_blend.glsl
@@ -116,7 +116,12 @@ vec4 HueRotate(vec4 Cs, float amount) {
 }
 
 vec4 Invert(vec4 Cs, float amount) {
-    return mix(Cs, vec4(1.0, 1.0, 1.0, Cs.a) - vec4(Cs.rgb, 0.0), amount);
+    Cs.rgb /= Cs.a;
+
+    vec3 color = mix(Cs.rgb, vec3(1.0) - Cs.rgb, amount);
+
+    // Pre-multiply the alpha into the output value.
+    return vec4(color.rgb * Cs.a, Cs.a);
 }
 
 vec4 Saturate(vec4 Cs, float amount) {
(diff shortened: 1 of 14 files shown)


glennw:
r? @kvark 

Gecko try (pending):

https://hg.mozilla.org/try/rev/2b0c4b777b82695064d1247e2eb7ac50876d644d
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2b0c4b777b82695064d1247e2eb7ac50876d644d

glennw:
Servo/wrench tests pass OK.

glennw:
@nical Happy to review this one too? :)

nical:
@glennw I think i am, but the gecko reftests don't look too happy. Might it be because we are making wrong assumptions in Gecko about pre-multiplication when using the bightness filter?

glennw:
@nical Ah, I hadn't noticed the try failures. Pushed an updated version, and added several new reftests to cover the cases that Gecko tests for.

Try run (successful this time) here:

https://hg.mozilla.org/try/rev/bf466f065c0db5379a707e459ca184357f792544
https://treeherder.mozilla.org/#/jobs?repo=try&revision=bf466f065c0db5379a707e459ca184357f792544

webrender/res/ps_blend.glsl
```diff
@@ -116,7 +116,10 @@ vec4 HueRotate(vec4 Cs, float amount) {
 }
 
 vec4 Invert(vec4 Cs, float amount) {
-    return mix(Cs, vec4(1.0, 1.0, 1.0, Cs.a) - vec4(Cs.rgb, 0.0), amount);
+    vec4 color = mix(Cs, vec4(Cs.a) - vec4(Cs.rgb, 0.0), amount);
+
+    // Pre-multiply the alpha into the output value.
+    return vec4(color.rgb * clamp(color.a, 0.0, 1.0), color.a);
```
mstange: `color` is already premultiplied; here you're multiplying it a second time. Your yaml reftest doesn't catch it because it uses a black result, and multiplying black one more time still stays black.

glennw: Ah, I guess we don't have an (enabled) gecko reftest that catches that either. Will add a specific reftest for this to wrench.

glennw:
@mstange Switched the invert to do a normal un/pre-multiply and added a reftest with alpha. Does that look right to you?

mstange:
Yes, looks good now. It took me a second to understand the test; I assume you're putting transparent blue on top of opaque blue in order to avoid the need to fuzz?

glennw:
Yep, thanks.

r? @nical or @kvark

PR #1726, opened 2017-09-20
Tidy up a few parts of frame.rs

glennw:
This is mostly just moving code out of frame.rs to more appropriate places. It also removes a hashmap in the Scene structure, since we had two hashmaps with 1:1 usage.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1726)


glennw:
r? @mrobinson 

(Happy to redo this after your hit testing PR, if this will be a pain to rebase on top of).

PR #1727, opened 2017-09-20
Support creating weak freelist handles directly from strong handle.

glennw:
This makes creating weak handles a bit easier - no need to pass the
owning freelist around. This will make the code a bit cleaner as
we start to use this in more places.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1727)


glennw:
Merging manually due to CI issues.

Issue #1730, opened 2017-09-20
Gradients which are clipped to rounded corners are *additionally* drawn unclipped, with operator ADD


mstange:
>> [Testcase](https://bug1401639.bmoattachments.org/attachment.cgi?id=8910378) <<

Expected vs actual:
<img width="519" alt="screen shot 2017-09-20 at 1 24 08 pm" src="https://user-images.githubusercontent.com/961291/30658113-b1f05e2a-9e07-11e7-89d5-1855cbfbdf3a.png">

It looks like the gradient is drawn twice, both clipped and unclipped. And the unclipped drawing seems to use operator ADD so it's not visible on white backgrounds.

https://bugzilla.mozilla.org/show_bug.cgi?id=1401639

```html
<!DOCTYPE html>
<html lang="en">
<meta charset="utf-8">
<title>Rounded corners with gradients</title>

<style>

div {
  display: inline-block;
  width: 50px;
  padding: 10px;
}

p {
  margin: 0;
  width: 50px;
  height: 50px;
  background: linear-gradient(lime, lime);
  border-radius: 100%;
}

</style>

<div style="background:white">
  <p></p>
</div>

<div style="background:blue">
  <p></p>
</div>

<div style="background:black">
  <p></p>
</div>
```

PR #1733, opened 2017-09-21
Fix pre-multiplied alpha for clip masks in aligned gradients.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1733)


glennw:
Fixes #1730 

r? @nical or @kvark

glennw:
We probably have several other shaders with this issue, as fall out from the switch to pre-multiplied alpha. We can fix them up and add reftests as we come across them...

PR #1735, opened 2017-09-21
Collapse LayerRectCalculationAndCullingPass into FrameBuilder.

glennw:
Although this still looks quite messy, it should allow us to start
splitting up this functionality in a more fine-grained way in the
future. Specifically, the next step is to split off the way that
the clip stack is handled, so that this can be done once during
clip-scroll tree update - for performance reasons.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1735)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -93,6 +93,15 @@ pub struct FrameBuilder {
 
     /// Whether or not we've pushed a root stacking context for the current pipeline.
     has_root_stacking_context: bool,
+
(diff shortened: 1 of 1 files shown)


glennw:
r? @mrobinson

This is the change we discussed yesterday. Sorry for the large patch size! It doesn't need super careful review - it's just shuffling code around without functional changes at this stage.

Also, happy to just redo this after your hit-test changes if it's going to be a difficult rebase.

glennw:
OK, I'll merge now then. Thanks!

@bors-servo r=mrobinson

PR #1743, opened 2017-09-22
Fix subpixel text positioning with negative coordinates.

### LINKED ISSUE #1739: subpixel_quantize_offset seems wrong for negative values
https://bugzilla.mozilla.org/show_bug.cgi?id=1402088

Glyphs to the left of the coordinate system origin should be treated the same way as glyphs to the right of the coordinate system origin. The current implementation of  subpixel_quantize_offset doesn't seem to do that, because it uses `.frac()` and `.abs()`, instead of, for example, `x - x.floor()`.

Or at least it behaves differently than Skia does, as can be seen in this testcase, when running with Firefox+webrender:

\>> [Testcase](https://bug1402088.bmoattachments.org/attachment.cgi?id=8910915) <<

```html
<!DOCTYPE html>
<html lang="en">
<meta charset="utf-8">
<title>Glyphs with negative local positions are misaligned</title>

<style>

p:hover {
  filter: blur(0px);
}

</style>

<div style="transform: translateX(100px); width: 500px">
  <div style="margin-left: -100px">
    <p>
      Here is some text which will shift when you move your mouse over this paragraph.<br>
      More specifically, only glyphs within the leftmost 100px of this element will
      shift, because those glyphs have negative local positions, and webrender's
      subpixel quantization isn't based on <code>x - floor(x)</code>.
    </p>
  </div>
</div>
```

glennw:
Fixes #1739.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1743)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -294,11 +294,11 @@ Glyph fetch_glyph(int specific_prim_address,
             // Glyphs positioned [-0.125, 0.125] get a
             // subpx position of zero. So include that
             // offset in the glyph position to ensure
-            // we truncate to the correct whole position.
-            glyph.x = trunc(glyph.x + 0.125);
+            // we round to the correct whole position.
+            glyph.x = floor(glyph.x + 0.125);
             break;
         case SUBPX_DIR_VERTICAL:
-            glyph.y = trunc(glyph.y + 0.125);
+            glyph.y = floor(glyph.y + 0.125);
             break;
     }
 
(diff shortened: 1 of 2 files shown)


glennw:
@mstange Your suggested fix works perfectly, nice!

r? @nical or @jrmuizel 

I'm not sure how best to regression test stuff like this, but it would be good to automate it. For now, I did a pixel perfect comparison with Gecko, using the test case in #1739.

Gankro:
Can you not just test this by comparing text at X to text at -Y in a container at X+Y?

PR #1744, opened 2017-09-22
Add support for hit testing

### LINKED ISSUE #1575: Expose a hit-test API in webrender
Gecko tracking bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1389146

For Gecko/APZ we want webrender to expose a hit-testing API. The API would take a point and return these two pieces of information: (1) the innermost scrollable frame that contains the point. This will needs to take into account any applied animated/async transforms. (2) the event region information associated with the point. How we do this is still up in the air, but #411 is tracking providing webrender with the event region information. This will likely be on a per-display-item granularity.

mrobinson:
This adds a new API point for doing hit testing against the display
list. This version supports the features Servo needs to do hist
testing, except for text index support (which will come in a later PR).

Fixes #1575.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1744)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1075,6 +1075,7 @@ version = "0.52.0"
 dependencies = [
  "app_units 0.5.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "bincode 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "bitflags 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-foundation 0.4.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -2,9 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{BorderRadius, ComplexClipRegion, ImageMask, ImageRendering};
-use api::{DeviceIntRect, LayerPoint, LayerRect, LayerSize, LayerToWorldTransform, LocalClip};
+use api::{BorderRadius, ComplexClipRegion, DeviceIntRect, ImageMask, ImageRendering, LayerPoint};
+use api::{LayerRect, LayerSize, LayerToWorldTransform, LayoutPoint, LayoutVector2D, LocalClip};
 use border::BorderCornerClipSource;
+use ellipse::Ellipse;
 use freelist::{FreeList, FreeListHandle, WeakFreeListHandle};
 use gpu_cache::{GpuCache, GpuCacheHandle, ToGpuBlocks};
 use prim_store::{ClipData, ImageMaskData};
@@ -110,6 +111,22 @@ impl From<ClipRegion> for ClipSources {
 
         ClipSources::new(clips)
     }
+}
+
+impl ClipSource {
+    pub fn contains(&self, point: &LayerPoint) -> bool {
+        // We currently do not handle all types of clip sources, because they
+        // aren't used for ClipScrollNodes and this method is only used during hit testing.
+        match self {
+            &ClipSource::Rectangle(ref rectangle) => rectangle.contains(point),
+            &ClipSource::RoundedRectangle(rect, radii, ClipMode::Clip) =>
+                rounded_rectangle_contains_point(point, &rect, &radii),
+            &ClipSource::RoundedRectangle(rect, radii, ClipMode::ClipOut) =>
+                !rounded_rectangle_contains_point(point, &rect, &radii),
+            _ => unreachable!("Tried to call contains on an unsupported ClipSource."),
+        }
+    }
+
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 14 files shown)


mrobinson:
cc @staktrace

glennw:

---

*[webrender/src/clip_scroll_tree.rs, line 129 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KuphI8_CQjdlpgE2RYv:-KuphI8_CQjdlpgE2RYw:b-pguyhv) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender/src/clip_scroll_tree.rs#L129)):*
> ```Rust
>         clip_store: &ClipStore
>     ) -> bool {
>         match cache.get(node_id) {
> ```

if let

---

*[webrender/src/ellipse.rs, line 78 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KupfQoi1XppwpMmIo_M:-KupfQoi1XppwpMmIo_N:b-s899ba) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender/src/ellipse.rs#L78)):*
> ```Rust
>     /// Find the signed distance from this ellipse given a point.
>     /// Taken from http://www.iquilezles.org/www/articles/ellipsedist/ellipsedist.htm
>     fn signed_distance(&self, point: LayerVector2D) -> f32 {
> ```

We should probably add some unit tests for this - as a low priority follow up.

---

*[webrender/src/frame_builder.rs, line 1591 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KupiGib85-XgctBAa8s:-KupiGicQm6YfcP6uGIg:b-ln0ksa) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender/src/frame_builder.rs#L1591)):*
> ```Rust
>         }
> 
>         result.items.dedup();
> ```

dedup() will only work if the items are sorted - that won't be the case here, right?

---

*[webrender_api/src/api.rs, line 149 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-Kupen2bB8Lr8D1WOv1V:-Kupen2cA5cEbzVvTqcj:b2rnkln) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender_api/src/api.rs#L149)):*
> ```Rust
>     pub pipeline: PipelineId,
>     pub tag: ItemTag,
>     pub point_in_viewport: LayoutPoint,
> ```

Could we add a comment here with more details on the coordinate space here?

---

*[webrender_api/src/display_item.rs, line 46 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KupeFGvMtoHT4vzSwHC:-KupeFGwl-gmC-G95vqb:bhgikry) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender_api/src/display_item.rs#L46)):*
> ```Rust
> /// second is used to select the cursor that should be used during mouse
> /// movement.
> pub type ItemTag = (u64, u8);
> ```

The size of this tuple seems quite implementation specific. I imagine that struct alignment will mean there is probably 7 bytes of padding added to the type anyway (although I haven't checked this). Is there any way we could make this a single u64 (or 2x u32) and pack the tags a bit tighter on the Servo side?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1744)*

glennw:
Also - CI is failing, due to warnings as errors. So we'll need to fix those up before landing this too.

mrobinson:

---

*[webrender/src/clip_scroll_tree.rs, line 129 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KuphI8_CQjdlpgE2RYv:-KuuH2iFfxqxRDDJ4oeY:b-896fix) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender/src/clip_scroll_tree.rs#L129)):*

if let
</blockquote></details>

Done.

---

*[webrender/src/ellipse.rs, line 78 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KupfQoi1XppwpMmIo_M:-KuuLZxy7wWwQO7imQkU:bns5trj) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender/src/ellipse.rs#L78)):*

We should probably add some unit tests for this - as a low priority follow up.
</blockquote></details>

Okay. I will create unit tests in a followup PR.

---

*[webrender/src/frame_builder.rs, line 1591 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KupiGib85-XgctBAa8s:-KuuH5225dEHH_73qM4i:b-cbxpyi) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender/src/frame_builder.rs#L1591)):*

dedup() will only work if the items are sorted - that won't be the case here, right?
</blockquote></details>

This is to avoid having a bunch of repeated items with the same information (for instance the border and background of the same div). In this case, the items with equal tags will almost certainly be together (CSS stacking works per Flow and not per display item). Deduplicating here should leave one entry per Flow (if the cursor is equal).

---

*[webrender_api/src/api.rs, line 149 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-Kupen2bB8Lr8D1WOv1V:-KuuHkFitvS3_UmIjOZD:b-896fix) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender_api/src/api.rs#L149)):*

Could we add a comment here with more details on the coordinate space here?
</blockquote></details>

Done.

---

*[webrender_api/src/display_item.rs, line 46 at r1](https://reviewable.io:443/reviews/servo/webrender/1744#-KupeFGvMtoHT4vzSwHC:-KuuHks1Bkol-ZRBlCkK:bh7wdqd) ([raw file](https://github.com/servo/webrender/blob/be11310324126c683cb8d6a68d2a2508ac558441/webrender_api/src/display_item.rs#L46)):*

The size of this tuple seems quite implementation specific. I imagine that struct alignment will mean there is probably 7 bytes of padding added to the type anyway (although I haven't checked this). Is there any way we could make this a single u64 (or 2x u32) and pack the tags a bit tighter on the Servo side?
</blockquote></details>

I took a look at this and having metadata stored in the layout process introduces a race condition. The WebRender thread is the only thread with display list information for all pipelines. If we need to make a second request to a pipeline's layout thread to fetch the metadata, conceivably the layout thread may have produced a new display list(s) in the meantime. If the ids that we use to locate the metadata are produced by a simple counter we are likely to produce an incorrect hit test result.

My idea here is to make a future patch which turns the ItemTag into a simple u32 index and allows the display list to include a side table of display item metadata.  This metadata structure will be defined in some way by the client, but importantly, it will stay with the display list to avoid race conditions. The tag will be used to access the data from the side table and return it to the compositor.

If you are okay with this idea, I'd like to land this version and do the "fix" later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1744)*

Issue #1724, opened 2017-09-19
Use sccache for builds


jrmuizel:
https://github.com/mozilla/sccache/issues/179 has some instructions.

When I last looked at these I believe I couldn't figure out how exactly to communicate the encrypted AWS key to sccache. The details are fuzzy though.

jrmuizel:
In talking with kats I realized that we might be better off just trying to get sccache to work with task-cluster directly instead of getting it working with travis first.

jrmuizel:
Here's an example of using secrets with taskcluster:
https://dxr.mozilla.org/mozilla-central/source/taskcluster/ci/upload-generated-sources/kind.yml#35
https://dxr.mozilla.org/mozilla-central/source/build/upload_generated_sources.py#94

and upload the secret using https://tools.taskcluster.net/secrets

jrmuizel:
Also:
jrmuizel: you can set env vars for sccache for AWS creds (assuming your CI won't leak them in logs), or you could have a little script to rewrite the JSON secret into ~/.aws/credentials format
ted
https://github.com/mozilla/sccache/blob/master/src/simples3/credential.rs
ted
is the code that sccache uses to find AWS creds (forked from rusoto)

and https://dxr.mozilla.org/mozilla-central/source/taskcluster/scripts/builder/build-linux.sh#55

glennw:
@jrmuizel @metajack What is the likelihood of getting someone at Mozilla who works on this stuff to look into this officially (running CI on task cluster)? I certainly don't know enough about that stuff to look into it.

We haven't been able to merge anything for ~3 days with the latest travis issues, which is following another fortnight of similar issues...

jrmuizel:
@staktrace is looking at this a bit right now and we can probably get @luser to help before he goes on PTO next week.

staktrace:
At the moment I only have basic linux64 taskcluster integration working. There are two steps involved:
1. Install the integration github app on the repo at https://github.com/apps/taskcluster (I have this app installed for my fork, `staktrace/webrender`)
2. Apply the change from my [taskcluster-ci](https://github.com/staktrace/webrender/commits/taskcluster-ci) branch - this just adds the .taskcluster.yml file and tweaks one of the reftest fuzz numbers.

[...]

PR #1746, opened 2017-09-22
Add a .taskcluster.yml file to run CI using taskcluster

staktrace:
This is a test PR to see if (a) taskcluster correctly picks up the PR and schedules the CI jobs and (b) to see how bors/homu deal with this extra CI job.

This is related to #1724

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1746)


staktrace:
Ok, that was anticlimactic. I'm not sure what's the best way forward here. Somebody who knows more about how the CI checks work with the bors/homu bots should chime in. Otherwise I'd say we should merge this PR but be ready to back it out if it causes problems. In theory the taskcluster jobs should always be green for good PRs so it should be safe to run them side-by-side with the travis CI until we're confident it doesn't break stuff.

glennw:
@staktrace I'm fine with merging this - we can back it out if required. But why the change to fuzzy results in one of the reftests?

staktrace:
@glennw On the taskcluster Linux machines that reftest gives slightly different
results so I had to tweak the reftest.list to make the test pass.

staktrace:
I continued looking into what we need to complete the integration and posted to https://github.com/servo/webrender/issues/1724#issuecomment-331917517. Even without the integration complete, I think it would be valuable to merge this PR, because we'll get taskcluster running CI and can get results faster than via Travis. Even if bors/homu don't yet recognize those results, the human doing the PR can get feedback faster which would be valuable.

PR #1752, opened 2017-09-26
Support platform specific reftests.

glennw:
Enable the synthetics italics reftest on Linux.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1752)

### DIFF
--- a/wrench/src/reftest.rs
+++ b/wrench/src/reftest.rs
@@ -19,6 +19,15 @@ use webrender::api::*;
 use wrench::{Wrench, WrenchThing};
 use yaml_frame_reader::YamlFrameReader;
 
+#[cfg(target_os = "windows")]
+const PLATFORM: &str = "win";
+#[cfg(target_os = "linux")]
+const PLATFORM: &str = "linux";
+#[cfg(target_os = "macos")]
+const PLATFORM: &str = "mac";
+#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))]
+const PLATFORM: &str = "other";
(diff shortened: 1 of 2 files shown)


wrench/src/reftest.rs
```diff
@@ -19,6 +19,15 @@ use webrender::api::*;
 use wrench::{Wrench, WrenchThing};
 use yaml_frame_reader::YamlFrameReader;
 
+#[cfg(target_os = "windows")]
+const PLATFORM: &str = "win";
+#[cfg(target_os = "linux")]
```
kvark: Do we alienate non-linux UNIX systems into `other` category?
Perhaps, we can use `cfg(windows)`, `cfg(unix)`, and `cfg(target_os="macos")` instead, see [target flags](https://doc.rust-lang.org/reference/attributes.html#conditional-compilation) for more info

wrench/src/reftest.rs
```diff
@@ -19,6 +19,15 @@ use webrender::api::*;
...
+const PLATFORM: &str = "linux";
+#[cfg(target_os = "macos")]
+const PLATFORM: &str = "mac";
+#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))]
```
kvark: sometimes I wish there was `if-else` in `cfg` things

wrench/src/reftest.rs
glennw: Yea - you're right. I'll update it to be unix.

PR #1753, opened 2017-09-26
Allow sub-primitives to have individual clip masks and calculate bounding rect.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1753)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1707,17 +1707,7 @@ impl FrameBuilder {
         for i in 0 .. prim_count {
             let prim_index = PrimitiveIndex(base_prim_index.0 + i);
 
-            let (prim_local_rect, prim_screen_rect) = match self.prim_store.build_bounding_rect(
-                prim_index,
-                &prim_context,
-            ) {
-                Some(rects) => rects,
-                None => continue,
-            };
-
-            debug!("\t\t{:?} bound is {:?}", prim_index, prim_screen_rect);
-
-            let visible = self.prim_store.prepare_prim_for_render(
+            if let Some((prim_local_rect, prim_screen_rect)) = self.prim_store.prepare_prim_for_render(
                 prim_index,
                 &prim_context,
                 resource_cache,
@@ -1726,11 +1716,10 @@ impl FrameBuilder {
                 TextRunMode::Normal,
                 render_tasks,
                 &mut self.clip_store,
-            );
-
-            if visible {
-                stacking_context.screen_bounds =
-                    stacking_context.screen_bounds.union(&prim_screen_rect);
+            ) {
+                stacking_context.screen_bounds = stacking_context
+                    .screen_bounds
+                    .union(&prim_screen_rect);
                 stacking_context.isolated_items_bounds = stacking_context
                     .isolated_items_bounds
                     .union(&prim_local_rect);
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1719,7 +1719,6 @@ impl FrameBuilder {
 
             let visible = self.prim_store.prepare_prim_for_render(
                 prim_index,
-                prim_screen_rect,
                 &prim_context,
                 resource_cache,
                 gpu_cache,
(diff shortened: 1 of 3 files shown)


glennw:
r? @mrobinson 

The main goal here is to move the bounding rect and clip mask calculation into ```prepare_prim_for_render```, since this is called recursively for sub-primitives.

In some ways this is a bit tidier, I think, with the new ```PrimitiveContext```. In other ways, it's a bit messier (the current length of ```prepare_prim_for_render```).

glennw:
Servo/wrench tests OK.

Gecko try:

https://hg.mozilla.org/try/rev/9ee389afdafd783d4bfd3ac2b1c45fcfeef8bace
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9ee389afdafd783d4bfd3ac2b1c45fcfeef8bace

I *think* the failures that occurred on the try run are infrastructure / intermittent issues - @staktrace  ?

staktrace:
@glennw Yeah the gtest failure on the try run looks like a pre-existing intermittent.

webrender/src/frame_builder.rs
```diff
@@ -112,15 +112,96 @@ pub struct FrameBuilder {
...
+
+impl<'a> PrimitiveContext<'a> {
+    fn new(
+        packed_layer_index: PackedLayerIndex,
+        packed_layer: &'a PackedLayer,
+        clip_id: ClipId,
+        screen_rect: &DeviceIntRect,
+        clip_scroll_tree: &ClipScrollTree,
+        clip_store: &ClipStore,
+        device_pixel_ratio: f32) -> Option<PrimitiveContext<'a>> {
```
kvark: nit: should be
```rust
    `device_pixel_ratio: f32,
) -> Option<Self> {
```

webrender/src/frame_builder.rs
```diff
@@ -1684,96 +1678,54 @@ impl FrameBuilder {
...
-            "\tclip_bounds {:?}, layer_local_clip {:?}",
-            clip_bounds,
-            packed_layer.local_clip_rect
-        );
 
         if !stacking_context.is_backface_visible && packed_layer.transform.is_backface_visible() {
             return;
         }
 
+        let prim_context = PrimitiveContext::new(
```
kvark: hmm, so this clip-scroll tree traversal used to be cached in `current_clip_info`, and now it's performed unconditionally. Seems like a perf regression?

webrender/src/frame_builder.rs
```diff
@@ -1684,96 +1678,54 @@ impl FrameBuilder {
...
-                device_pixel_ratio,
-            ) {
-                Some(rects) => rects,
-                None => continue,
-            };
-
-            debug!("\t\t{:?} bound is {:?}", prim_index, prim_screen_rect);
 
-            let prim_metadata = self.prim_store.prepare_prim_for_render(
+            if let Some((prim_local_rect, prim_screen_rect)) = self.prim_store.prepare_prim_for_render(
```
kvark: note that we have a special type `clip::Geometry` for having both the local and global rectangles

webrender/src/prim_store.rs
```diff
@@ -970,6 +973,7 @@ impl PrimitiveStore {
                     local_rect: *local_rect,
```
kvark: given how often this struct gets repeated, we should make one before the `match` and then do
```rust
let metadata = PrimitiveMetadata {
    cpu_prim_index: SpecificPrimitiveIndex(self.cpu_yuv_images.len()),
    .. base_metadata
};
```

webrender/src/frame_builder.rs
glennw: It *is* a performance regression, but only minor (and temporary). Previously, the cache would get invalidated as soon as the clip ID changed. Now, we do it for every primitive run. But primitive runs get broken when the clip and/or scroll ID changes. In the vast majority of cases, the cause for a primitive run break is when the clip node ID has changed, which would invalidate the cache anyway.

glennw:
@kvark Thanks for the reviews - nits and comments addressed.

mrobinson:
I'm late to the party, but I just finished my review. Looks good to me as well. :)

Issue #1749, opened 2017-09-25
Shouldn't be applying color to emoji (colored glyphs)


Gankro:
If you set `color: #0366d6` on my good friend https://gankro.github.io/blah/webtests/text.html the emoji turn purple (at least on macos). Webrender should be ignoring the RGB color values attached to colored fonts. Although it does need to respect the alpha.

You can also see this on reactions in github issues. (which I will add to this issue as an easy example)

Possibly the same root cause as #1747

Gankro:
Interestingly, doesn't happen for mozilla's emoji: http://mozilla.github.io/fxemoji/dist/FirefoxEmoji/index.html

So it looks like a funky interaction with apple's emoji set. Did font variations suddenly unlock this...?

glennw:
@Gankro I think this one can be closed?

Gankro:
I agree, I was just waiting for this land upstream in gecko (a few more hours) before checking the box.

kvark:
@Gankro for the reference, please post the associated Gecko issue here

Gankro:
There isn't one, which is why I was using this for it 😅

(we can't test it in this repo, so it made sense to do final verification once it was integrated with something that can)

Gankro:
Confirmed fixed.

Issue #1747, opened 2017-09-22
Webrender mishandling shadows/transparency on emoji (color glyphs)


Gankro:
This crashes webrender:

```text
data:text/html,<html><head><meta charset="utf8"></head><body style="text-shadow: 0px 0px 1px black;">🙃</body></html>
```

This fails to render as transparent in webrender:

```text
data:text/html,<html><head><meta charset="utf8"></head><body style="color:rgba(0,0,0,0.1);">🙃</body></html>
```

Seems related? text-shadows bypassing a safety guard that causes the second one to be opaque?

backtrace:

```
thread 'WebRender:Worker#5' panicked at 'assertion failed: `(left == right)` (left: `4`, right: `8`)', /Users/ABeingessner/dev/gecko/gfx/webrender/src/platform/macos/font.rs:426
stack backtrace:
   0:        0x11a82a993 - std::sys::imp::backtrace::tracing::imp::unwind_backtrace::h884a721e113c3303
   1:        0x11a836e53 - std::panicking::default_hook::{{closure}}::h7a7d734b2824d103
   2:        0x11a836a9c - std::panicking::default_hook::h3eb11bd6cbfdc331
   3:        0x11a838fc7 - std::panicking::rust_panic_with_hook::h8b9b25777425677b
   4:        0x11a838e64 - std::panicking::begin_panic::h69fe06a9d67b6165
   5:        0x11a838dd2 - std::panicking::begin_panic_fmt::haf8695bfb6c52436
   6:        0x11a14d1e5 - webrender::glyph_rasterizer::GlyphRasterizer::request_glyphs::{{closure}}::{{closure}}::he3c67689b6433db9
   7:        0x11a115f52 - rayon::iter::internal::bridge_producer_consumer::helper::he869ddeb5dd1c5df
   8:        0x11a0fc509 - std::panicking::try::do_call::h3b50a9cff7cadfa3
Redirecting call to abort() to mozalloc_abort
```

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1401798

Gankro:
The shadow crash only happens with Apple's emoji font. For instance, there's no problem with applying shadows to mozilla's emoji font: http://mozilla.github.io/fxemoji/dist/FirefoxEmoji/index.html

Although we still don't apply transparency.

Gankro:
I removed the assertion for fun, and got this haunting specter:
[...]

PR #1754, opened 2017-09-26
Add color emoji support on mac.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1754)

### DIFF
--- a/webrender/res/ps_text_run.glsl
+++ b/webrender/res/ps_text_run.glsl
@@ -70,15 +70,14 @@ void main(void) {
     //note: the blend mode is not compatible with clipping
     oFragColor = texture(sColor0, tc);
 #else
-    float alpha = texture(sColor0, tc).a;
+    vec4 color = texture(sColor0, tc) * vColor;
 #ifdef WR_FEATURE_TRANSFORM
     float a = 0.0;
     init_transform_fs(vLocalPos, a);
-    alpha *= a;
+    color.a *= a;
 #endif
-    vec4 color = vColor;
-    alpha = min(alpha, do_clip());
-    oFragColor = vec4(vColor.rgb, vColor.a * alpha);
+    color.a = min(color.a, do_clip());
+    oFragColor = color;
 #endif
 }
 #endif
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -11,7 +11,7 @@ use core_foundation::base::TCFType;
 use core_foundation::dictionary::{CFDictionary, CFDictionaryRef};
 use core_foundation::number::{CFNumber, CFNumberRef};
 use core_foundation::string::{CFString, CFStringRef};
-use core_graphics::base::{kCGImageAlphaNoneSkipFirst, kCGImageAlphaPremultipliedLast};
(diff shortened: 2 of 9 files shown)


glennw:
emoji

cc @Gankro The blur is not quite the same as in Gecko - but I think that's more an issue with the blur than the glyph. Nonetheless, it's better than it was.

Gankro:
screen shot 2017-09-26 at 10 50 45 am
screen shot 2017-09-26 at 10 50 57 am
screen shot 2017-09-26 at 10 51 18 am
screen shot 2017-09-26 at 10 52 26 am

Lookin' good!

Gankro:
Fixes #1749 and #1747 (except we still don't make them transparent with `color: rgba(0,0,0,0.5)`)

mstange:
https://dbaron.org/log/20110225-blur-radius may provide insight.

webrender/res/ps_text_run.glsl
```diff
@@ -70,15 +70,14 @@ void main(void) {
...
 #else
-    float alpha = texture(sColor0, tc).a;
+    vec4 color = texture(sColor0, tc) * vColor;
 #ifdef WR_FEATURE_TRANSFORM
     float a = 0.0;
     init_transform_fs(vLocalPos, a);
-    alpha *= a;
+    color.a *= a;
 #endif
-    vec4 color = vColor;
```
kvark: This code path used to ignore `sColor0` texture fetch, and now it does not. Is this change intentional?

webrender/src/platform/macos/font.rs
```diff
@@ -422,6 +422,14 @@ impl FontContext {
         }
     }
 
+    pub fn is_bitmap_font(&mut self, font_key: FontKey) -> bool {
+        self.get_ct_font(font_key, Au(16 * 60), &[])
+            .map_or(false, |ref ct_font| {
```
kvark: nit: a simple match would look cleaner

glennw: Fixed

webrender/res/ps_text_run.glsl
glennw: It used to fetch just the ```a``` component for non-subpixel fonts. Now, it fetches and scales the color by the entire ```rgba``` of the texture. This is necessary to support full color glyphs. Normal alpha glyphs that go through this path always have ```(1,1,1)``` in their ```rgb``` components, so it all works the same for those fonts (I verified this visually on all platforms).

glennw:
@kvark Thanks - fixed and comments addressed.

mbrubeck:
@bors-servo r-

testing

glennw:
Merging manually due to bors issues.

webrender_api/src/font.rs
```diff
@@ -93,6 +93,7 @@ pub enum FontRenderMode {
     Mono = 0,
     Alpha,
     Subpixel,
+    Bitmap,
```
mstange: Would `Color` be a better name? When I read the word `Bitmap` I think of a monochrome pixel font. @jrmuizel felt the same when I asked him about it.

PR #1758, opened 2017-09-26
Initialize render targets to the required size, saving memory and clear time.

glennw:
We have the information available that tells us the worst case
*actual* size requirement for each render pass array texture. Keep
track of that and only create targets of that size.

This is a quite significant GPU memory saving, and also a large
performance win on *some* GPU architectures (seems to be a good
benefit on the Intel GPUs I've tested with).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1758)


glennw:
r? @kvark 

(Still needs a try run before we can consider merge).

glennw:
This does trigger one failure in the Servo WPT suite. It looks like some kind of rounding / snapping error related to filters. Will investigate and fix that before doing a gecko try.

glennw:
In each of the test images, there are a small number (~50?) of pixels that change by a value of 1.0/255.0. The changed pixels aren't along the edge, they are in random places in the images. I *think* this is an accuracy issue in OSMesa, shown up by the different frame buffer size.

staktrace:
@glennw This caused some debug reftest failures, one of the assertions is tripping. See https://bugzilla.mozilla.org/show_bug.cgi?id=1403971#c1 for links to the try pushes.

glennw:
@staktrace  This actually got merged before I had a chance to do a gecko try run :/ It's fine with me if someone wants to revert this change for now (if it's urgent) - otherwise I'll look into the failures on Monday.

kvark:
Oh, that would be my fault. I misread the earlier comment for rounding issue to be the only one.

glennw:
My fault too - I forgot that I hadn't run the try on it. Sorry!

kvark:
I looked at the code, and the link between computed max size and the assertion is not obvious. Figured we can keep looking while the code is reverted, see #1782

kvark:
The only path I came up with so far is that we got `TextureAllocator::allocate` calls with size width or height being zero (on that non-framebuffer target).

I believe the simplest solution would then be to allocate 1x1 target for such an extreme case. It's not a performance critical use case, and placing more workarounds/handling code along the way (earlier) would just unnecessarily complicate the code.

kvark:
Implemented in #1782
Optimally, we'd want a test for that, but I'm not sure how to create one with zero sized task.

Issue #1732, opened 2017-09-20
Transformed images with transformed rectangle clips are not clipped properly


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1401665

\>> [Testcase](https://bug1401665.bmoattachments.org/attachment.cgi?id=8910430) <<

Expected vs actual rendering:
<img width="413" alt="screen shot 2017-09-20 at 3 59 33 pm" src="https://user-images.githubusercontent.com/961291/30664776-d130ff00-9e1c-11e7-91a3-cf86097e6f06.png">

It looks like the image is only clipped to the axis-aligned bounds of the rotated rectangle instead of to the actual rotated rectangle.

In the YAML recording, the `image: "res/1505936895-img-59.png"` item has `"clip-and-scroll": 4`, and the clip with `id: 4` is defined inside the rotated stacking context, so I think this is a webrender bug and not a Gecko bug.

```
---
root: 
  items: 
    -
      bounds: [0, 0, 0, 0]
      "clip-rect": [0, 0, 0, 0]
      "clip-and-scroll": 0
      "backface-visible": true
    - 
      bounds: [0, 0, 2880, 1482]
      "clip-rect": [0, 0, 2880, 1482]
      "clip-and-scroll": 0
      "backface-visible": true
      type: rect
      color: white
    - 
      bounds: [0, 0, 2880, 1482]
      "clip-rect": [0, 0, 2880, 1482]
      "clip-and-scroll": 0
      "backface-visible": true
      type: clip
      id: 1
      "content-size": [2880, 1482]
    - 
      "clip-rect": [0, 0, 2880, 1482]
      "clip-and-scroll": 1
      "backface-visible": true
      type: "scroll-frame"
      id: 2
      "content-size": [2880, 1482]
      bounds: [0, 0, 2880, 1482]
    - 
      bounds: [0, 0, 2880, 1482]
      "clip-rect": [0, 0, 2880, 1482]
      "clip-and-scroll": 2
      "backface-visible": true
      type: clip
      id: 3
      "content-size": [2880, 1482]
    - 
      bounds: [0, 0, 2880, 1482]
      "clip-rect": [0, 0, 2880, 1482]
      "clip-and-scroll": 3
      "backface-visible": true
      type: rect
      color: white
    -
      bounds: [0, 0, 0, 0]
      "clip-rect": [0, 0, 0, 0]
      "clip-and-scroll": 3
      "backface-visible": true
      type: "stacking-context"
[...]

PR #1760, opened 2017-09-27
Fix local clip rect being applied in transform shader.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1760)


glennw:
r? @kvark 

Fixes #1732.

This still needs a gecko try run. Passes the servo/wrench tests ok.

glennw:
There is a gecko try run here, that has some failures.

https://hg.mozilla.org/try/rev/2b2948c12b1d516c5d7bb49a07cd6b1d9ffd1d00
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2b2948c12b1d516c5d7bb49a07cd6b1d9ffd1d00

I *think* all the failures are intermittent, except for one permanent failure in R4 on ```layout/reftests/bugs/579323-1.html```. I briefly looked at that test and there's no transform elements on it. Given that, I wouldn't expect this test to be affected (since it should be running through the non-transform shader path).

I'll debug it tomorrow, unless @staktrace or @jrmuizel have any ideas about that failure?

glennw:
Hmmm, that same test is failing (in the same way) for me locally, *without* this patch applied (just running with current tip). So I think that means this is fine to merge - @staktrace  ?

glennw:
I also verified that none of the items in that test are running through the transform shader path.

staktrace:
@glennw Yeah this appears to be a pre-existing failure that cropped up sometime last night. I'm chasing down what caused it but it is unrelated to your patches so feel free to land your stuff. Bug 1403457 is tracking the failure if you're interested.

PR #1763, opened 2017-09-27
Move content_size to ScrollingState and store a scrolling size

mrobinson:
This is only used for scroll frames, so it doesn't make sense to have
it on the shared data structure.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1763)


mrobinson:
All WebRender reftests are passing with this change.

kvark:
unrelated OSX CI error:
> error: RPC failed; curl 56 SSLRead() return error -36

@bors-servo retry

PR #1766, opened 2017-09-27
Remove a redundant field from ClipInfo.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1766)


glennw:
I also removed the ```screen_bounding_rect``` field from ```ClipInfo``` since it's handled by the clip sources bounding rect check in the same place.

PR #1771, opened 2017-09-28
Add clip mask sampling to the angle gradient shader.

### LINKED ISSUE #1770: Most gradients don't respect rounded clips
https://bugzilla.mozilla.org/show_bug.cgi?id=1403976

Making any of the following changes to `wrench/reftests/gradient/linear-aligned-border-radius.yaml` causes the gradients to no longer respect the rounded corners of the clip:

 - Adding `repeat: true`
 - Changing the position of the first color stop to something non-zero, e.g. `0.2`
 - Making the gradient non-axis-aligned, for example using `end: 51 100`

glennw:
Fixes #1770.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1771)


webrender/res/ps_angle_gradient.glsl
```diff
@@ -57,8 +59,12 @@ void main(void) {
 
     float offset = dot(pos - vStartPoint, vScaledDir);
 
-    oFragColor = sample_gradient(vGradientAddress,
+    vec4 color = sample_gradient(vGradientAddress,
                                  offset,
                                  vGradientRepeat);
+
+    color *= do_clip();
```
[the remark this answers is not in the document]
glennw: Because the gradients use pre-multiplied alpha.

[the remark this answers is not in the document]
kvark: hmm, but then the alpha from the clip is not handled in the same way as it is for other primitives.
here it's basically multiplying original_alpha * clip_alpha
elsewhere it's min(original_alpha, clip_alpha)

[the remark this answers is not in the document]
glennw: Hmmm, there's something strange with the existing pre-multiplied gradient tests. I'll investigate and fix those up first, then modify this patch.

glennw:
@kvark Updated - please re-review when you have time :)

webrender/res/ps_angle_gradient.glsl
```diff
@@ -57,8 +59,18 @@ void main(void) {
 
     float offset = dot(pos - vStartPoint, vScaledDir);
 
-    oFragColor = sample_gradient(vGradientAddress,
+    vec4 color = sample_gradient(vGradientAddress,
                                  offset,
                                  vGradientRepeat);
+
+    // Un-premultiply the color from sampling the gradient.
+    color.rgb /= color.a;
```
kvark: what happens if `color.a == 0.0`? I assume that `color.rgb` becomes INF, and then when we re-multiply by `color.a` we'll get INF * 0.0, which is not actually 0.0.

mstange: If the answer here is "we cannot rely on this becoming 0.0 after the multiplication", then the Invert and Brightness functions will need to be fixed as well.

kvark: AFAIK, IEEE 32bit float define INF * 0.0 = NaN, so we can't rely on division by `color.a` to be reversible

kvark: Test in Rust - https://play.rust-lang.org/?gist=dc6143962fd7ea794266d9decd417202&version=stable

glennw: You're right - I added an ```if``` check to only apply this if the alpha is > 0. We will also need to fix up the blend functions that @mstange mentioned - I'll do those in a separate PR.

glennw:
@kvark Rebased and fixed up. Thanks!

kvark:
Thanks!
@bors-servo r+

PR #1775, opened 2017-09-29
ClipScrollGroup reference & visibility

### LINKED ISSUE #1751: Introducing a regular clip inside a sticky clip results in bad behaviour and stack overflow
Run the scrolling.rs example, and scroll the nested frame. Observe how the dark blue rect (which is initially not visible, but becomes visible as you scroll the nested frame down) behaves. In particular, it is sticky at two different ranges, once when it is near the bottom of the nested scroll frame, and once more when it is near the top of the nested scroll frame.

Now apply this patch, which just introduces a single clip inside the sticky frame. Note that the clip rect is the same as the bounds of the rect inside the sticky frame, so this clip should have no visual effect.
```patch
index d335206b..f9747c03 100644
@@ -116,18 +116,21 @@ impl Example for App {
                     Some(StickySideConstraint {
                         margin: 10.0,
                         max_offset: -40.0,
                     }),
                     None,
                 ),
             );
             builder.push_clip_id(sticky_id);
+            let dummy_clip = builder.define_clip(None, (50, 350).by(50, 50), vec![], None);
+            builder.push_clip_id(dummy_clip);
             let info = LayoutPrimitiveInfo::new((50, 350).by(50, 50));
             builder.push_rect(&info, ColorF::new(0.5, 0.5, 1.0, 1.0));
+            builder.pop_clip_id(); // dummy_clip
             builder.pop_clip_id(); // sticky_id

             // just for good measure add another teal square further down and to
             // the right, which can be scrolled into view by the user
             let info = LayoutPrimitiveInfo::new((250, 350).to(300, 400));
             builder.push_rect(&info, ColorF::new(0.0, 1.0, 1.0, 1.0));

```

Build with this and scroll. Observe how (a) the rect doesn't stick any more and (b) if you scroll far enough the app crashes with this:

```
thread 'RenderBackend' has overflowed its stack
fatal runtime error: stack overflow

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)
```

/cc @mrobinson

kvark:
This PR makes us properly ignore the primitive runs culled out by the clip-scroll tree hierarchy.
Fixes #1751

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1775)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -94,7 +94,9 @@ pub struct FrameBuilder {
 
     stacking_context_store: Vec<StackingContext>,
     clip_scroll_group_store: Vec<ClipScrollGroup>,
-    clip_scroll_group_indices: FastHashMap<ClipAndScrollInfo, ClipScrollGroupIndex>,
+    // Note: value here is meant to be `ClipScrollGroupIndex`,
+    // but we already have `ClipAndScrollInfo` in the key
+    clip_scroll_group_indices: FastHashMap<ClipAndScrollInfo, usize>,
     packed_layers: Vec<PackedLayer>,
 
     // A stack of the current text-shadow primitives.
@@ -136,7 +138,8 @@ impl<'a> PrimitiveContext<'a> {
         screen_rect: &DeviceIntRect,
         clip_scroll_tree: &ClipScrollTree,
         clip_store: &ClipStore,
-        device_pixel_ratio: f32) -> Option<Self> {
+        device_pixel_ratio: f32,
+    ) -> Option<Self> {
 
         let mut current_clip_stack = Vec::new();
         let mut clip_bounds = *screen_rect;
@@ -242,15 +245,6 @@ impl FrameBuilder {
         }
     }
 
-    pub fn create_clip_scroll_group_if_necessary(&mut self, info: ClipAndScrollInfo) {
-        if self.clip_scroll_group_indices.contains_key(&info) {
-            return;
-        }
-
-        let group_index = self.create_clip_scroll_group(info);
-        self.clip_scroll_group_indices.insert(info, group_index);
-    }
-
     /// Create a primitive and add it to the prim store. This method doesn't
     /// add the primitive to the draw list, so can be used for creating
     /// sub-primitives.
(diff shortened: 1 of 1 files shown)


glennw:
I haven't looked through this in detail yet - but how does this differ from the ```is_visible()``` method that exists on ```ClipScrollGroup``` now?

glennw:
Put another way, do we need this extra visibility flag because that method isn't applicable here? Or can we make use of that and not store an extra visibility flag?

webrender/src/frame_builder.rs
```diff
@@ -94,7 +94,7 @@ pub struct FrameBuilder {
 
     stacking_context_store: Vec<StackingContext>,
     clip_scroll_group_store: Vec<ClipScrollGroup>,
-    clip_scroll_group_indices: FastHashMap<ClipAndScrollInfo, ClipScrollGroupIndex>,
+    clip_scroll_group_indices: FastHashMap<ClipAndScrollInfo, usize>,
```
glennw: Any reason to switch to an untyped index here? I guess because the ```ClipScrollGroupIndex``` type has extra information?

webrender/src/frame_builder.rs
```diff
@@ -1551,8 +1547,11 @@ impl FrameBuilder {
         &self,
         clip_and_scroll: &ClipAndScrollInfo
     ) -> Option<PackedLayerIndex> {
-        let group_index = self.clip_scroll_group_indices.get(&clip_and_scroll).unwrap();
-        let clip_scroll_group = &self.clip_scroll_group_store[group_index.0];
+        let group_id = self.clip_scroll_group_indices[&clip_and_scroll];
+        //Note: this ignores `group_ref.visibility` intentionally
```
glennw: This comment doesn't seem right with this version of the patch?

kvark: good eye! fixed now

webrender/src/frame_builder.rs
kvark: yes

kvark:
Thanks for detailed review!
@bors-servo r=glennw

PR #1782, opened 2017-09-29
Temporary revert of #1758 (Initialize render targets to the required size)

kvark:
This reverts commit 3a5b79233f5ccb9ac0bc65b22ba6578b85094fa1 due to Gecko failures in https://github.com/servo/webrender/pull/1758#issuecomment-333112046

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1782)

PR #1783, opened 2017-09-29
Unrevert #1758, fix the target allocation size to be at least 1

kvark:
This is an alternative to #1782
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1783)


kvark:
@glennw I believe the reason was [mentioned here](https://github.com/servo/webrender/pull/1758#issuecomment-333234083). If you can see a clean way to prevent 0x0 coming through, I'd be happy to agree with it ;)

glennw:
Yep - I think we should be able to find out which render task results in a 0 sized allocation and handle that case. I'll take a look this week at the test failures.

glennw:
I traced this in the failing Gecko test. We end up in a situation where a stacking context is considered visible, even though the screen bounding rect for the stacking context is zero. This means we create a dynamic alpha batch task with zero size, which triggers the assertion. We'll need to find why the SC is considered visible in this case.

glennw:
I'll close this for now - since the original patch causes a test failure in Servo too, that I haven't investigated yet. We can re-open once these issues are sorted out.

PR #1787, opened 2017-10-02
Fix some calculation errors for ps_split_composite and make_polygon.

mephisto41:
There are some bugs when we have a transform-style:flat in the middle of transform-style:preserve-3d. So I fixed it by following changes.

1. ps_split_composite shader did not take position of dest_task into
account.
2. make_polygon didn't work if isolated_items_bounds is empty.

r? kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1787)

### DIFF
--- a/webrender/res/ps_split_composite.glsl
+++ b/webrender/res/ps_split_composite.glsl
@@ -38,11 +38,15 @@ void main(void) {
     CompositeInstance ci = fetch_composite_instance();
     SplitGeometry geometry = fetch_split_geometry(ci.user_data0);
     AlphaBatchTask src_task = fetch_alpha_batch_task(ci.src_task_index);
+    AlphaBatchTask dest_task = fetch_alpha_batch_task(ci.render_task_index);
+
+    vec2 dest_origin = dest_task.render_target_origin -
+                       dest_task.screen_space_origin;
 
     vec3 world_pos = bilerp(geometry.points[0], geometry.points[1],
                             geometry.points[3], geometry.points[2],
                             aPosition.y, aPosition.x);
-    vec4 final_pos = vec4(world_pos.xy * uDevicePixelRatio, ci.z, 1.0);
+    vec4 final_pos = vec4((world_pos.xy + dest_origin) * uDevicePixelRatio, ci.z, 1.0);
 
     gl_Position = uTransform * final_pos;
 
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -8,7 +8,7 @@ use api::{DeviceIntPoint, DeviceIntRect, DeviceIntSize, DeviceUintRect, DeviceUi
 use api::{ExtendMode, FIND_ALL, FilterOp, FontInstance, FontRenderMode};
 use api::{GlyphInstance, GlyphOptions, GradientStop, HitTestFlags, HitTestItem, HitTestResult};
 use api::{ImageKey, ImageRendering, ItemRange, ItemTag, LayerPoint, LayerPrimitiveInfo, LayerRect};
-use api::{LayerSize, LayerToScrollTransform, LayerVector2D, LayoutVector2D, LineOrientation};
+use api::{LayerPixel, LayerSize, LayerToScrollTransform, LayerVector2D, LayoutVector2D, LineOrientation};
 use api::{LineStyle, LocalClip, POINT_RELATIVE_TO_PIPELINE_VIEWPORT, PipelineId, RepeatMode};
 use api::{ScrollSensitivity, Shadow, TileOffset, TransformStyle};
 use api::{WorldPixel, WorldPoint, YuvColorSpace, YuvData, device_length};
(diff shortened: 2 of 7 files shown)


webrender/src/frame_builder.rs
```diff
@@ -46,12 +46,21 @@ fn make_polygon(
     stacking_context: &StackingContext,
     node: &ClipScrollNode,
     anchor: usize,
+    device_pixel_ratio: f32,
 ) -> Polygon<f32, WorldPixel> {
-    //TODO: only work with `isolated_items_bounds.size` worth of space
```
kvark: why is this comment removed?

webrender/res/ps_split_composite.glsl
```diff
@@ -38,11 +38,15 @@ void main(void) {
     CompositeInstance ci = fetch_composite_instance();
     SplitGeometry geometry = fetch_split_geometry(ci.user_data0);
     AlphaBatchTask src_task = fetch_alpha_batch_task(ci.src_task_index);
+    AlphaBatchTask dest_task = fetch_alpha_batch_task(ci.render_task_index);
+
+    vec2 dest_origin = dest_task.render_target_origin -
```
kvark: nice!

webrender/src/frame_builder.rs
```diff
@@ -46,12 +46,21 @@ fn make_polygon(
...
     // from the primitive local coordinates into the layer transformation.
     // Which in turn needs it to be a render task property obeyed by all primitives
     // upon rendering, possibly not limited to `write_*_vertex` implementations.
+    if stacking_context.isolated_items_bounds.is_empty() {
+        // isolated_items_bounds is empty means there is no item for this sc.
+        // in this case, we use screen_bounds instead.
+        let rect = stacking_context.screen_bounds.to_f32() / device_pixel_ratio;
+        let bounds = LayerRect::new(LayerPoint::new(rect.origin.x, rect.origin.y),
+                                    LayerSize::new(rect.size.width, rect.size.height));
+        return Polygon::from_transformed_rect(bounds, LayerToWorldTransform::identity(), anchor);
```
kvark: hmm, I'm not seeing the point here. If a stacking context has no items to isolate, why would we even want it to participate in the plane splitting (and hence, generate a polygon here)?

webrender/src/frame_builder.rs
mephisto41: I removed this accidentally. I'll add it back.

webrender/src/frame_builder.rs
mephisto41: In the reftest we have following display list.
```
sc1 (preserve3d)
  sc2 (preserve3d)
    rect1
  sc3 (flat)
    sc4 (preserve3d)
      sc5 (preserve3d)
        rect2
      sc6 (preserve3d)
        rect3
```

We'll create plane splitting for rect 2 and rect 3, then render it in sc3. Next, we'll create plane splitting for sc3 and render it on the screen. Thus, we have a sc with no item bound but need a plane splitting. Does it makes sense?

kvark: I believe the `isolated_items_bounds` for SC3 would then need to include the bounds of SC4. Given that SC4 is transformed, its bounds would need to be projected onto SC3 (transformed).

mephisto41:
@kvark , Now if the SC doesn't have any primitive I'll use the union of all children's bounds as isolated_items_bounds. Can you review it again? Thanks.

kvark:
@mephisto41 looks good to me now, thank you!
Do you mind launching a try push to see if any Gecko issues appear before we proceed?

mephisto41:
Sure, I'll submit a try.

mephisto41:
Here is try.  https://treeherder.mozilla.org/#/jobs?repo=try&revision=9f31a21048f83634bed711b15e2a6b558e894805

kvark:
@mephisto41 looks good, thank you!
Please rebase and r=me

mephisto41:
@kvark, I found the original solution did not resolved the issue completely. If the SC is not isolation and it contains primitives and preserve-3d SCs. The bounds for make_polygon should include primitives bounds and all SC children bounds. My original solution only works with the non-isolation SC without any primitives. So I add a new test case for this problem and add a condition in the ```handle_pop_stacking_context```. I think this modification need your review.

```
else if stacking_context.isolation != ContextIsolation::Items {
    stacking_context.isolated_items_bounds = stacking_context
        .isolated_items_bounds
        .union(&stacking_context.children_sc_bounds);
}
```

kvark:
> If the SC is not isolation and it contains primitives and preserve-3d SCs

Could you have a test case for that? AFAIS, both your reftests have all the contexts being isolated.

mephisto41:
I think the SC in line 19 of intermediate-2.yaml is non-isolated SC which has a primitive and a preserve-3d SC.

kvark:
@mephisto41 ok, I think we are good to go then, thank you!
@bors-servo r+

PR #1793, opened 2017-10-03
Support ahem reftests in wrench.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1793)


glennw:
r? @kvark or @nical 

@Gankro Once I disabled subpx AA in the tests, the results are pixel perfect (on Linux at least, pending CI on mac) :)

Gankro:
Wow, super surprising that the difference is that high!

I assume the path forward for gecko reftests is to wait for @lsalzman's FontInfo patches to land?

glennw:
If Gecko already reads information from FontConfig as to whether subpixel AA is enabled, you could potentially modify where Gecko calls ```add_font_instance``` to create fonts without subpixel AA (or even just pass through the global enable subpixel AA option in ```RendererOptions``` at init time).

Otherwise, waiting for the patches @lsalzman is working on should resolve the issue.

PR #1798, opened 2017-10-03
Extra care of the upper bound for image tiling shader

### LINKED ISSUE #1797: Thin line visible around a transformed primitive
See #1732 for the test case image/code.
The used image is attached here:
1505936895-img-59

I can repro it on Linux/Intel.

kvark:
Fixes  #1797
r? @glennw

Gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=8a192063236587f0d93b5b85b0bf1c16108c3a6d

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1798)


glennw:
Shader validation is failing on CI. Is it possible to add a reftest for this? It would be good to have test coverage for an edge case like this.

kvark:
@glennw should be better now

kvark:
Weird, I generated the image from `headless.py`, then cropped it, and it doesn't match O_o

glennw:
@kvark I hate to be pedantic on this stuff - but we probably shouldn't include that image in the repo unless we know the licensing on it. Would it be possible to replace with a simple image we create that tests the issue in question?

kvark:
@glennw sure, that would be most optimal, just going to take a bit more effort.

kvark:
Replaced the image with a hand-generated one.
Reftest produces 1 pixel difference of value 1 when running all the reftests (with osmesa) for some reason, but not when running transform tests only.

glennw:
Thanks!

@bors-servo r+

PR #1799, opened 2017-10-03
Use an unsafe serializer.

jrmuizel:
This gives a noticable impact on serialization performance in Gecko.
wr_dp_push_text() goes from 35.6% of
nsDisplayText::CreateWebRenderCommands down to 24%.

The generated code is still pretty bad but hopefully adding
proper noalias information to rust will fix that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1799)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -17,6 +17,8 @@ use YuvImageDisplayItem;
 use bincode;
 use serde::{Deserialize, Serialize, Serializer};
 use serde::ser::{SerializeMap, SerializeSeq};
+use std::io::Write;
+use std::{io, ptr};
 use std::marker::PhantomData;
 use time::precise_time_ns;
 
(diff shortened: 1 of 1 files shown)


Gankro:
I'm currently building a rustc with the noalias flag we discussed, would like to hold off on merging that until we've looked at the codegen of the 4 combos of `alias x unsafe`

webrender_api/src/display_list.rs
```diff
@@ -483,6 +485,48 @@ impl<'a, 'b> Serialize for DisplayItemRef<'a, 'b> {
     }
 }
 
+// This is a replacement for bincode::serialize_into(&vec)
+// The default implementation Write for Vec will basically
+// call extend_from_slice(). Serde ends up calling that for ever
```
Gankro: s/ever/every

Gankro:
Some codgen tests here. Looks like this is an all around win, and even positions us for bigger wins with mutable noalias:

https://github.com/rust-lang/rust/pull/45012#issuecomment-334203385

webrender_api/src/display_list.rs
```diff
@@ -483,6 +485,71 @@ impl<'a, 'b> Serialize for DisplayItemRef<'a, 'b> {
...
+        self.0 += buf.len();
+        Ok(buf.len())
+    }
+    fn flush(&mut self) -> io::Result<()> { Ok(()) }
+}
+
+fn serialize_fast<T: Serialize>(vec: &mut Vec<u8>, e: &T) {
+    // manually counting the size is faster than vec.reserve(bincode::serialized_size(&e) as usize) for some reason
+    let mut size = SizeCounter(0);
+    bincode::serialize_into(&mut size,e , bincode::Infinite).unwrap();
```
kvark: missing space before `e`

webrender_api/src/display_list.rs
```diff
@@ -483,6 +485,71 @@ impl<'a, 'b> Serialize for DisplayItemRef<'a, 'b> {
...
+//
+
+struct UnsafeVecWriter<'a>(&'a mut Vec<u8>);
+
+impl<'a> Write for UnsafeVecWriter<'a> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        unsafe {
+            let old_len = self.0.len();
+            self.0.set_len(old_len + buf.len());
+            ptr::copy_nonoverlapping(buf.as_ptr(), self.0.as_mut_ptr().offset(old_len as isize), buf.len());
```
kvark: need some debug assertions for the capacity guarantee

PR #1801, opened 2017-10-04
Rename text-shadow APIs to shadow instead.

glennw:
In the near future, the text-shadow APIs will be used to
implement other types of shadows (e.g. box-shadows).

This patch just renames all functions and structures, which
will make it easier to review the follow up patches that
modify box shadows to work via the current text-shadow API.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1801)


glennw:
r? @nical or @kvark 

@Gankro @staktrace Heads up that this modifies the public API and will require updating WR bindings. It's a straight rename - ```push_text_shadow``` to ```push_shadow``` etc.

Gankro:
Ok, should be very easy to update gecko.

Just checking, will this refactoring only lead to the addition of rects as "things that go in a (text)-shadow"? (currently only glyphs and lines?)

glennw:
Yep, that's right (for now, anyway!).

staktrace:
Thanks for the heads-up (re WR bindings update needed)!

PR #1806, opened 2017-10-04
Remove the (now unused) nested display list APIs.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1806)


glennw:
r? @mrobinson and @jrmuizel 

(Should probably not merge until @jrmuizel confirms that https://bugzilla.mozilla.org/show_bug.cgi?id=1405819 has landed).

It wasn't clear to me whether the different types of ClipIds (external, dynamic etc) are related to nesting. Is that for something else or should they be modified too?

kvark:
One less thing in WR that I don't fully understand :D

kvark:
The Gecko change has landed (with `RESOLVED FIXED` status).
@bors-servo r=mrobinson

Issue #1738, opened 2017-09-21
On my integrated GPU, a non-blurred rounded box shadow breaks all future clipping


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1402073

\>> [Testcase](https://bug1402073.bmoattachments.org/attachment.cgi?id=8910894) <<

This testcase contains two things:
 * a box-shadow with rounded corners and a zero blur radius, and
 * a background color that's clipped to a rounded shape

On my integrated GPU, the background is not clipped to the rounded shape.

Rendering with integrated GPU (this build already contains the `!=` workarounds for text and gradients):
<img width="790" alt="screen shot 2017-09-21 at 2 56 40 pm" src="https://user-images.githubusercontent.com/961291/30714149-cf166c8c-9edf-11e7-8570-2e0dfacc9cea.png">

Rendering with discrete GPU:
<img width="790" alt="screen shot 2017-09-21 at 2 56 35 pm" src="https://user-images.githubusercontent.com/961291/30714173-e0520d1c-9edf-11e7-82ea-9e0863e8825e.png">

```html
<!DOCTYPE html>
<html lang="en">
<meta charset="utf-8">
<title>Two box shadows create garbage</title>

<style>

div {
  display: inline-block;
  margin: 10px;
  border-radius: 20px;
  width: 100px;
  height: 100px;
}

</style>

<div style="box-shadow: black 0 1px;"></div>
<div style="background: black"></div>
```

@glennw, any ideas where I should be searching for another integer equality comparison?

glennw:
I wonder if it could be related to the for loop in the blur shader, which uses integers? Argh, what a frustrating shader compiler / driver you have!

mstange:
This testcase doesn't have a blur, though.

glennw:
Oh, right. Not immediately clear to me where there would be an int comparison. I'll take a quick look at the shaders today and let you know if I spot anything...

mstange:
It turns out that the alternative workaround for #1809 that I gave in https://github.com/servo/webrender/pull/1810#issuecomment-334480233 actually fixes this bug. I have no idea why though.

kvark:
@mstange could you make a PR then? thanks!

Issue #1728, opened 2017-09-20
Glyphs are all over the place on my integrated GPU


mstange:
After the most recent [webrender update in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1400216), which included #1715, text rendering is now completely garbled on my machine while using the integrated GPU.

When switching to the discrete GPU, text rendering is fixed.

Integrated:
<img width="935" alt="screen shot 2017-09-20 at 9 08 11 am" src="https://user-images.githubusercontent.com/961291/30645387-9d2bb2ce-9de3-11e7-869f-648651d4b241.png">

Discrete:
<img width="935" alt="screen shot 2017-09-20 at 9 08 20 am" src="https://user-images.githubusercontent.com/961291/30645393-a30f6d2a-9de3-11e7-80a1-a47f5ce120fb.png">

I'm on macOS 10.12.6 (16G29), my integrated GPU is an Intel HD Graphics 530 1536 MB.

mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1401633

glennw:
@mstange Was the text working fine on your mac previous to that update? Do you have a git SHA1 regression range I could take a look at?

mstange:
Text is working fine on the integrated GPU with 1407b1e1cd0631527f2e686b16aed38c0feabaaf and broken with 68e2e7cd0e39d216bb6c35dfb353dc0700bb6948 .

For 1407b1e1cd0631527f2e686b16aed38c0feabaaf I tested the macOS build from the first try push in https://bugzilla.mozilla.org/show_bug.cgi?id=1400216#c4 (which is at https://queue.taskcluster.net/v1/task/B-4OuL9RTNOUpW3_4CkDUA/runs/0/artifacts/public/build/target.dmg ).

glennw:
Wow, that's really strange. The only related change I can see in that range is the subpx text PR, which has only a minor shader change (https://github.com/servo/webrender/commit/ec26fb9221502d21126a52a7a227f54bc48f7c72).

If it's easy to do, it would be handy to revert that locally and confirm that is trigger what is (probably) a driver bug. Otherwise, I have a mac laptop being shipped to me now - so hopefully I'll receive that within the next week or so and be able to reproduce locally.

paulrouget:
Same here. Just regular Servo.

Intel Iris Graphics 550 1536 MB

[...]

PR #1810, opened 2017-10-05
Work around an apparent driver bug that caused wrong corner radii.

### LINKED ISSUE #1809: Misassigned corner radii on my integrated GPU
https://bugzilla.mozilla.org/show_bug.cgi?id=1405955

[Testcase](https://bug1405955.bmoattachments.org/attachment.cgi?id=8915473)

Expected vs actual:

screen shot 2017-10-05 at 2 20 25 am

```
  border-top-left-radius: 0px;
  border-top-right-radius: 10px;
  border-bottom-left-radius: 20px;
  border-bottom-right-radius: 40px;
```

mstange:
Fixes #1809.

Using the driver of the Intel GPU in my Macbook Pro (Intel HD Graphics 530
1536 MB, 0x191b, Skylake GT2) on macOS 10.12.6, the integer calculation that was
used in order to compute the lookup address gave wrong results, and the corner
radius for the wrong corner was looked up.
The only way to get correct results from it was to move the entire adjustment of
the address ivec2 variable into the caller. Now all values in the calculations
were integer literals.
I then replaced the multiplications with gradual "+= 2" adjustments of the
address variable in fetch_clip, because that looked better and might even be
easier to reason about than the original code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1810)

### DIFF
--- a/webrender/res/cs_clip_rectangle.glsl
+++ b/webrender/res/cs_clip_rectangle.glsl
@@ -27,8 +27,7 @@ struct ClipCorner {
     vec4 outer_inner_radius;
 };
 
-ClipCorner fetch_clip_corner(ivec2 address, int index) {
-    address += ivec2(2 + 2 * index, 0);
+ClipCorner fetch_clip_corner(ivec2 address) {
     vec4 data[2] = fetch_from_resource_cache_2_direct(address);
     return ClipCorner(RectWithSize(data[0].xy, data[0].zw), data[1]);
 }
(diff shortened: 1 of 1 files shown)


glennw:
Fun! Could you add a comment in that shader code just outlining roughly what's in the commit message and then r=me. Thanks for digging into it!

nical:
I added this to the [driver issues](https://github.com/servo/webrender/wiki/Driver-issues#1810---incorrect-integer-arithmetic-in-shader-on-mac-under-certain-conditions) page. Unfortunately my descriptions of the problem and resolution are super vague, if you have any idea how to better explain this for future reference  don't hesitate to fill in the blanks.

kvark:
omg, this is an aweful bug
thanks @mstange for exposing and fixing it!

mstange:
Here's an alternative fix that also works:

```diff
diff --git a/webrender/res/cs_clip_rectangle.glsl b/webrender/res/cs_clip_rectangle.glsl
--- a/webrender/res/cs_clip_rectangle.glsl
+++ b/webrender/res/cs_clip_rectangle.glsl
@@ -27,8 +27,8 @@ struct ClipCorner {
     vec4 outer_inner_radius;
 };

-ClipCorner fetch_clip_corner(ivec2 address, int index) {
-    address += ivec2(2 + 2 * index, 0);
+ClipCorner fetch_clip_corner(ivec2 address, float index) {
+    address += ivec2(2 + 2 * int(index), 0);
     vec4 data[2] = fetch_from_resource_cache_2_direct(address);
     return ClipCorner(RectWithSize(data[0].xy, data[0].zw), data[1]);
 }
```

It looks like what's broken is: reading from *integer variables* (or integer arguments) when doing any kind of integer calculation, in the right circumstances. If I store the number in a float, everything works. If I do the calculation without making use of an integer variable as part of the inputs of that calculation, things work as well.

I'll stick with the original fix though, and I'll add a comment.

leeoniya:
those pesky integers are at it again? https://github.com/servo/webrender/issues/1728#issuecomment-331234126

mstange:
They sure are. And there's still at least one more instance of this problem left to find (#1738).

PR #1813, opened 2017-10-05
Remove glyph cache

Gankro:
It was never used, and @glennw has said it isn't as interesting anymore. Let's get   f r e e   p e r f.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1813)


glennw:
@bors-servo r+

I still think this might be useful in the future, but it's low priority. Removing now might simplify performance related work on the DL (de)serialization too.

PR #1816, opened 2017-10-05
add more per-platform font options to FontInstancePlatformOptions

lsalzman:
This fleshes out FontInstancePlatformOptions in the ways previously discussed with Glenn:

Mainly that now FontInstancePlatformOptions actually has varying sets of options per-platform.

Revised FontInstanceOptions to rather implement Default trait instead of using Option all over the place to make Gecko integration simpler and saner.

Especially notable is that I heavily revised the set of FreeType options to be sufficient for Gecko to pass in all of its Fontconfig options.

As part of this, there was a lot of reworking of how bitmap fonts are managed with FreeType, and to allow them to be scaled in the shader from fixed-size strikes. To support bitmap fonts, I also needed to migrate us from using BlendMode::Alpha to using BlendMode::PremultipliedAlpha to avoid decode hell in FreeType (this also better agreeing with our image handling too). This required some heavy revision of the FreeType backend, and I optimized some decoding there while I was at it.

I cleaned up the FontInstance mechanics so that they are only really normalized once they are ready for lookup into the cache. This prevents us having to litter code everywhere to zero out the color, the subpixel direction, etc. This also slims down TextRunPrimitiveCpu by letting us remove some options that were only there because they were getting overwritten in FontInstance.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1816)

### DIFF
--- a/webrender_api/src/font.rs
+++ b/webrender_api/src/font.rs
@@ -137,6 +137,16 @@ impl FontRenderMode {
     }
 }
 
+impl SubpixelDirection {
+    // Limit the subpixel direction to what is supported by the render mode.
+    pub fn limit_by(self, render_mode: FontRenderMode) -> SubpixelDirection {
+        match render_mode {
+            FontRenderMode::Mono | FontRenderMode::Bitmap => SubpixelDirection::None,
+            FontRenderMode::Alpha | FontRenderMode::Subpixel => self,
+        }
+    }
+}
+
 #[repr(u8)]
 #[derive(Hash, Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub enum SubpixelOffset {
@@ -196,16 +206,101 @@ pub struct GlyphOptions {
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Deserialize, Hash, Eq, PartialEq, PartialOrd, Ord, Serialize)]
 pub struct FontInstanceOptions {
-    pub render_mode: Option<FontRenderMode>,
+    pub render_mode: FontRenderMode,
+    pub subpx_dir: SubpixelDirection,
     pub synthetic_italics: bool,
 }
 
+impl Default for FontInstanceOptions {
+    fn default() -> FontInstanceOptions {
+        FontInstanceOptions {
+            render_mode: FontRenderMode::Subpixel,
+            subpx_dir: SubpixelDirection::Horizontal,
+            synthetic_italics: false,
+        }
+    }
+}
+
+#[cfg(target_os = "windows")]
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Deserialize, Hash, Eq, PartialEq, PartialOrd, Ord, Serialize)]
 pub struct FontInstancePlatformOptions {
-    // These are currently only used on windows for dwrite fonts.
     pub use_embedded_bitmap: bool,
     pub force_gdi_rendering: bool,
+}
+
+#[cfg(target_os = "windows")]
+impl Default for FontInstancePlatformOptions {
+    fn default() -> FontInstancePlatformOptions {
+        FontInstancePlatformOptions {
+            use_embedded_bitmap: false,
+            force_gdi_rendering: false,
+        }
+    }
+}
+
+#[cfg(target_os = "macos")]
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Deserialize, Hash, Eq, PartialEq, PartialOrd, Ord, Serialize)]
+pub struct FontInstancePlatformOptions {
+    pub unused: u32,
+}
+
+#[cfg(target_os = "macos")]
+impl Default for FontInstancePlatformOptions {
+    fn default() -> FontInstancePlatformOptions {
+        FontInstancePlatformOptions {
+            unused: 0,
+        }
+    }
+}
+
+pub const FONT_FORCE_AUTOHINT: u16  = 0b1;
+pub const FONT_NO_AUTOHINT: u16     = 0b10;
+pub const FONT_EMBEDDED_BITMAP: u16 = 0b100;
+pub const FONT_EMBOLDEN: u16        = 0b1000;
+pub const FONT_VERTICAL_LAYOUT: u16 = 0b10000;
+pub const FONT_SUBPIXEL_BGR: u16    = 0b100000;
+
+#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize)]
+pub enum FontLCDFilter {
+    None,
+    Default,
+    Light,
+    Legacy,
+}
+
+#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize)]
+pub enum FontHinting {
+    None,
+    Mono,
+    Light,
+    Normal,
+    LCD,
+}
+
+#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Deserialize, Hash, Eq, PartialEq, PartialOrd, Ord, Serialize)]
+pub struct FontInstancePlatformOptions {
+    pub flags: u16,
+    pub lcd_filter: FontLCDFilter,
+    pub hinting: FontHinting,
+}
+
+#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+impl Default for FontInstancePlatformOptions {
+    fn default() -> FontInstancePlatformOptions {
+        FontInstancePlatformOptions {
+            flags: 0,
+            lcd_filter: FontLCDFilter::Default,
+            hinting: FontHinting::LCD,
+        }
+    }
 }
 
 #[derive(Clone, Hash, PartialEq, Eq, Debug, Deserialize, Serialize, Ord, PartialOrd)]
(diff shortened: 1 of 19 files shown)


glennw:
Reviewed 12 of 19 files at r1.
Review status: 12 of 19 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1816)*

glennw:
(Partial review only so far - will be working through this today).

webrender_api/src/font.rs
```diff
@@ -196,18 +206,103 @@ pub struct GlyphOptions {
...
+        }
+    }
+}
+
+pub const FONT_FORCE_AUTOHINT: u16  = 1 << 0;
+pub const FONT_NO_AUTOHINT: u16     = 1 << 1;
+pub const FONT_EMBEDDED_BITMAP: u16 = 1 << 2;
+pub const FONT_EMBOLDEN: u16        = 1 << 3;
+pub const FONT_VERTICAL_LAYOUT: u16 = 1 << 4;
+pub const FONT_SUBPIXEL_BGR: u16    = 1 << 5;
```
Gankro: Note: Rust has byte literals: `0b00100`

glennw:
Reviewed 7 of 19 files at r1.

---

*[webrender/res/cs_text_run.glsl, line 42 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KviyvtPjuo3ruSQrwcs:-KviyvtQs63kkzruB5Kc:b61v1wu) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/res/cs_text_run.glsl#L42)):*
> ```GLSL
>     // the glyph offset, relative to its primitive bounding rect.
>     vec2 size = (res.uv_rect.zw - res.uv_rect.xy) * res.scale;
>     vec2 local_pos = glyph.offset + vec2(res.offset.x, -res.offset.y) * res.scale / uDevicePixelRatio;
> ```

Why does the scale need to be applied to the offset?

---

*[webrender/res/ps_text_run.glsl, line 31 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-Kviye3_-PYHSSI28L_j:-Kviye3aNwHx5BWDJvqP:b888or3) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/res/ps_text_run.glsl#L31)):*
> ```GLSL
>     vec2 local_pos = glyph.offset +
>                      text.offset +
>                      vec2(res.offset.x, -res.offset.y) * scale;
> ```

I don't follow why the scale gets applies to the offset?

---

*[webrender/src/glyph_rasterizer.rs, line 153 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KviyJMGQngsc2_sadwU:-KviyJMGQngsc2_sadwV:br6zt4k) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/src/glyph_rasterizer.rs#L153)):*
> ```Rust
>         match font.render_mode {
>             FontRenderMode::Mono | FontRenderMode::Bitmap => {
>                 font.color = ColorF::new(1.0, 1.0, 1.0, 1.0).into();
> ```

nit: could construct ColorU here directly.

---

*[webrender/src/prim_store.rs, line 605 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KviyRrSDR1ssetE54Dh:-KviyRrSDR1ssetE54Di:b-by35vh) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/src/prim_store.rs#L605)):*
> ```Rust
>             self.offset.x,
>             self.offset.y,
>             self.font.subpx_dir.limit_by(self.font.render_mode) as u32 as f32,
> ```

It seems a bit weird that we have to limit_by here - shouldn't the subpx_dir value be set correctly earlier?

---

*[webrender/src/platform/unix/font.rs, line 143 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KvizLYPRp7Vf5WYFcNs:-KvizLYPRp7Vf5WYFcNt:b-kdwdbt) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/src/platform/unix/font.rs#L143)):*
> ```Rust
>             match (hinting, font.render_mode) {
>                 (FontHinting::None, _) => load_flags |= FT_LOAD_NO_HINTING,
>                 (FontHinting::Mono, _) => load_flags = FT_LOAD_TARGET_MONO,
> ```

It it deliberate that most of these are = but the first one is |= ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1816)*

glennw:
@lsalzman This mostly looks good to me - but there are definite rendering issues when I run this patch in Servo.

Without:

old

With:

new

If you view those images at full resolution, there's definite subpixel fringing on the second image that looks wrong. This is running on Linux with a device-pixel-ratio of 2 (not sure if that's relevant).

Once those minor issues above are sorted, and the subpx fringing is fixed, I'll do a gecko try run with this and compare to native gecko rendering. Then this should be good to go (although I should probably run over the freetype changes again, since there's a lot there).

glennw:
If I hard code the filter to DEFAULT, the fringing looks better. But the text is much heavier now:

olds

news

lsalzman:
Review status: 13 of 19 files reviewed at latest revision, 6 unresolved discussions.

---

*[webrender/res/cs_text_run.glsl, line 42 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KviyvtPjuo3ruSQrwcs:-KvjuAa3-upEf-iFIu4T:b-8cs51o) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/res/cs_text_run.glsl#L42)):*

Why does the scale need to be applied to the offset?
</blockquote></details>

I left the offset in strike space, which I had made some sense to me as width/height/bytes are in that space. I guess that is arguable, though. But I just changed it now so the left/top are output from the backends in the scaled space, that way the scaling of the offset no longer takes place in the shader. Just the rect.

---

*[webrender/res/ps_text_run.glsl, line 31 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-Kviye3_-PYHSSI28L_j:-Kvju_7pPi3YlbAp3Y4Y:b-q64pkt) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/res/ps_text_run.glsl#L31)):*

I don't follow why the scale gets applies to the offset?
</blockquote></details>

See above.

---

*[webrender/src/glyph_rasterizer.rs, line 153 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KviyJMGQngsc2_sadwU:-KvjubWQEnbzUmiU6V92:bb71kmx) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/src/glyph_rasterizer.rs#L153)):*

nit: could construct ColorU here directly.
</blockquote></details>

Okay

---

*[webrender/src/prim_store.rs, line 605 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KviyRrSDR1ssetE54Dh:-KvjudqWgcR82bOa71Dl:boc1r3c) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/src/prim_store.rs#L605)):*

It seems a bit weird that we have to limit_by here - shouldn't the subpx_dir value be set correctly earlier?
</blockquote></details>

It's not anymore because I pushed that back to when prepare_font is called before lookup into the cache. Otherwise, prepare_font would override fields in the font we depend on. This way of doing things also prevents us from having to litter the code everywhere with the burden of maintaining these constraints. prepare_font is called to fixup the copy of the font used to index the cache, and that's the only place. If we modified the font that is directly stored in TextRunPrimitiveCpu, we'd also have to store values like the color separately, and it still wouldn't help us for TextRunMode::Shadow. It just seemed cleaner to do this way, just with one small snag of having to limit the subpx_dir there. So the font we see there possibly has a subpx_dir that is incorrect for the render_mode. 

---

*[webrender/src/platform/unix/font.rs, line 143 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KvizLYPRp7Vf5WYFcNs:-KvjvCRCpFV2xvQ-OArE:b-ivfhi2) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender/src/platform/unix/font.rs#L143)):*

It it deliberate that most of these are = but the first one is |= ?
</blockquote></details>

Yes, FT_LOAD_DEFAULT has the default load target. So in some cases we don't want to override that, just amend it with some flags. Other hinting modes require us to change the target from the default.

---

*[webrender_api/src/font.rs, line 263 at r1](https://reviewable.io:443/reviews/servo/webrender/1816#-KvhrmGUF9nlCV0yygq_-r1-263:-KvjvNWKqJ7WiOm3Wnvy:b-lu5fl9) ([raw file](https://github.com/servo/webrender/blob/f51a8fb914943bef6b9c8ed4ce97870d848fffbc/webrender_api/src/font.rs#L263)):*

Note: Rust has binary literals: `0b00100`
</blockquote></details>

I find this less readable than the C convention for specifying bit flags, but if you prefer it this way, okay.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1816)*

lsalzman:
Fixed it to now properly pull FontInstancePlatformOptions Default::default() values in case None is supplied.

---

Review status: 13 of 19 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1816#-:-Kvjvc8oAodzUb8HJTuh:bzhl0tu)*

glennw:
WPT and wrench tests seem OK. Fringing is resolved with latest patch (Servo update will still need to set default hinting options to match previous behavior).

This is a really nice patch - it adds a heap of new functionality *and* simplifies the existing code - thanks!

Gecko try is here:
https://hg.mozilla.org/try/rev/84cb9319744c83e467024fab5e313bb2e867e98f
https://treeherder.mozilla.org/#/jobs?repo=try&revision=84cb9319744c83e467024fab5e313bb2e867e98f

If that is green, r=me.

glennw:
Try looks good, :rocket: 

@bors-servo r+

Issue #1812, opened 2017-10-05
DisplayListBuilder save-restore


Gankro:
In gecko, there's some display items which we'll probably ship with fallbacks (e.g. anything that can randomly use svg).

Under the current architecture, this forces us to "collect up" all facts about the item's rendering before actually using the DisplayListBuilder. This means a bunch of allocations and copies. If it instead had a save-restore functionality, we could push into it as we analyze the display item, and use a restore to undo our changes if we discover that the item needs fallback.

I am picturing an API like

```
let save_state = builder.save();
// ...
if found_something_bad {
  builder.restore(save_state);
}
```

There are currently 3 pieces of state in the builder: the dl buffer, the clip stack, and the glyph cache. I'm about to submit a PR to remove the glyph cache, so let's ignore that. 

The dl buffer is very easy: save the length, and truncate on restore.

The clip stack is similarly easy; save next_clip_id and the stack len. (I don't think we need to support restoring a clip that is popped). So basically this would be 3 u64's, and O(1) work.

PR #1819, opened 2017-10-05
Add save-restore functionality to builder

Gankro:
Implementation of #1812

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1819)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -51,6 +51,14 @@ impl<T> ItemRange<T> {
     }
 }
 
+#[repr(C)]
+#[derive(Debug)]
+pub struct DisplayListBuilderSaveState {
+    dl_len: usize,
+    clip_stack_len: usize,
+    next_clip_id: u64,
+}
+
 /// A display list.
 #[derive(Clone, Default)]
 pub struct BuiltDisplayList {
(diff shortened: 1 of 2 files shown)


glennw:
servo-tidy script fails with:

```
./wrench/src/rawtest.rs:193: trailing whitespace
./wrench/src/rawtest.rs:216: trailing whitespace
./wrench/src/rawtest.rs:224: trailing whitespace
./wrench/src/rawtest.rs:227: trailing whitespace
./wrench/src/rawtest.rs:228: trailing whitespace
./wrench/src/rawtest.rs:229: trailing whitespace
./wrench/src/rawtest.rs:233: trailing whitespace
./wrench/src/rawtest.rs:241: trailing whitespace
./wrench/src/rawtest.rs:259: trailing whitespace
./wrench/src/rawtest.rs:190: found an empty line following a {
```

Once those are resolved, r=me.

Gankro:
fixed

webrender_api/src/display_list.rs
```diff
@@ -594,6 +602,20 @@ impl DisplayListBuilder {
         }
     }
 
+    pub fn save(&self) -> DisplayListBuilderSaveState {
+        DisplayListBuilderSaveState {
+            clip_stack_len: self.clip_stack.len(),
```
kvark: wait, you are saving the clip stack length but not the actual contents?

Gankro: Yes, it will only "restore" by removing clips that were pushed since the save. This is sufficient for all actual use-cases, I think? Having trouble picturing a case where you'd save and then pop something that was pushed before the save.

glennw:
Looks like one remaining servo-tidy issue:

```
./wrench/src/rawtest.rs:258: trailing whitespace
```

webrender_api/src/display_list.rs
kvark: I don't understand. This is a stack, not a growing-only vector. We don't know what have been pushed/popped from the stack in the meantime. Having the stack containing newer items in place of the older ones after resetting would open a whole can of worms, sounds very unsafe to me.

PR #1821, opened 2017-10-06
Insert the shader name in the source.

nical:
This makes life easier when looking at apitrace recordings.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1821)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -183,6 +183,11 @@ pub fn build_shader_strings(
     vs_source.push_str(gl_version_string);
     fs_source.push_str(gl_version_string);
 
+    // Insert the shader name to make debugging easier.
+    let name_string = format!("// {}\n", base_filename);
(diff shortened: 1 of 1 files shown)


webrender/src/device.rs
```diff
@@ -183,6 +183,11 @@ pub fn build_shader_strings(
     vs_source.push_str(gl_version_string);
     fs_source.push_str(gl_version_string);
 
+    // Insert the shader name to make debugging easier.
+    let name_string = format!("// {}\n", base_filename);
```
kvark: perhaps, it would make sense to do in debug only?

nical: I use apitrace with release builds too and I don't think this adds any noticeable overhead so I lean towards leaving it in release builds as well unless you feel strongly about it.

kvark:
Okey
@bors-servo r+

PR #1825, opened 2017-10-06
Use panic=abort instead of panic=unwind

jrmuizel:
This is the same mode as Gecko uses and will get better codegen
once https://github.com/rust-lang/rust/pull/45012 has landed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1825)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,4 @@ members = [
 
 [profile.release]
 debug = true
+panic = "abort"


nox:
What does that imply for other users of webrender?

jrmuizel:
They will get worse performance until Rust supports emitting noalias metadata in code that can unwind. Although even then, having to unwind will always constrain the optimizer.

Gankro:
I think the top crate wins, so servo will choose to ignore this directive. However this will make wrench use panic=abort.

Gankro:
Actually I think servo/gecko won't see this at all, since it's in the workspace, and not the crates.

nox:
@Gankro Ok, I thought that this would set `panic=abort` for any user of this crate, and thus that any panic in WR would have made their program abort. That would have been a blocker in my book, but it seems I was wrong anyway so disregard me.

Gankro:
Yeah I totally get it, but it *should* be fine?

glennw:
I confirmed in a local servo build that a panic in WR doesn't abort the process with this change, so I'm assuming that means it's OK to merge.

@bors-servo r+

PR #1830, opened 2017-10-06
Make the writer even more unsafe.

jrmuizel:
Instead of changing the length every write we just adjust the pointer.
This avoids https://github.com/rust-lang/rust/issues/45068.
However we now need to ensure that we set the length when we are done.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1830)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -492,15 +492,13 @@ impl<'a, 'b> Serialize for DisplayItemRef<'a, 'b> {
 // }
 //
 
-struct UnsafeVecWriter<'a>(&'a mut Vec<u8>);
+struct UnsafeVecWriter(*mut u8);
 
-impl<'a> Write for UnsafeVecWriter<'a> {
+impl Write for UnsafeVecWriter {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
         unsafe {
-            let old_len = self.0.len();
-            self.0.set_len(old_len + buf.len());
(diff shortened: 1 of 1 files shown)


webrender_api/src/display_list.rs
```diff
@@ -523,7 +521,11 @@ fn serialize_fast<T: Serialize>(vec: &mut Vec<u8>, e: &T) {
     bincode::serialize_into(&mut size,e , bincode::Infinite).unwrap();
     vec.reserve(size.0);
 
-    bincode::serialize_into(&mut UnsafeVecWriter(vec), e, bincode::Infinite).unwrap();
+    let old_len = vec.len();
+    let ptr = unsafe { vec.as_mut_ptr().offset(vec.len() as isize) };
```
Gankro: r=me with this changed to old_len

jrmuizel:
I fixed up old_len and added a debug_assert to make sure things are sane.

jrmuizel:
With this patch only 1.2% of WebRender display list building is spent in serialization.

Gankro:
Nice!

pcwalton:
What was the fraction of time spent in serialization before?

jrmuizel:
I have too few samples to get a good comparison of the unsafe serializers. We have some patches coming that get rid of the cpu time spent elsewhere so that should raise the percentage of time spent in serialization. The safe serializer accounted for roughly 5.1% of the display list construction time.

Gankro:
r=me, don't have review rights yet

PR #1820, opened 2017-10-05
Add a workaround for inner rects on transformed clips.

glennw:
Fixes #1814.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1820)

PR #1833, opened 2017-10-09
Introduce ClipChain

mrobinson:
Clip chains are a linked list of ClipWorkItems that are used to avoid
having to recalculate the stack of clips necessary to render items when
switching between different ClipScrollNodes. The ClipChains are
calculated once on a per node basis during update_transform.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1833)

### DIFF
--- a/webrender/src/render_task.rs
+++ b/webrender/src/render_task.rs
@@ -6,10 +6,12 @@ use api::{ClipId, DeviceIntLength, DeviceIntPoint, DeviceIntRect, DeviceIntSize}
 use api::{FilterOp, MixBlendMode};
 use api::PipelineId;
 use clip::{ClipSource, ClipSourcesWeakHandle, ClipStore};
+use clip_scroll_tree::CoordinateSystemId;
 use gpu_cache::GpuCacheHandle;
 use internal_types::HardwareCompositeOp;
 use prim_store::{BoxShadowPrimitiveCacheKey, PrimitiveIndex};
 use std::{cmp, usize, f32, i32};
+use std::rc::Rc;
 use tiling::{ClipScrollGroupIndex, PackedLayerIndex, RenderPass, RenderTargetIndex};
 use tiling::{RenderTargetKind, StackingContextIndex};
 
@@ -26,6 +28,31 @@ pub struct RenderTaskAddress(pub u32);
 pub struct RenderTaskTree {
     pub tasks: Vec<RenderTask>,
     pub task_data: Vec<RenderTaskData>,
+}
+
+pub type ClipChain = Option<Rc<ClipChainNode>>;
+
+#[derive(Debug)]
+pub struct ClipChainNode {
+    pub work_item: ClipWorkItem,
+    pub prev: ClipChain,
+}
+
+struct ClipChainNodeIter {
+    current: ClipChain,
+}
+
+impl Iterator for ClipChainNodeIter {
+    type Item = Rc<ClipChainNode>;
+
+    fn next(&mut self) -> ClipChain {
+        let previous = self.current.clone();
+        self.current = match self.current {
+            Some(ref item) => item.prev.clone(),
+            None => return None,
+        };
+        previous
+    }
 }
 
 impl RenderTaskTree {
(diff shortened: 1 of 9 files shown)


mrobinson:
I did a bit of performance testing for this change. For nasty situations (a thousand deep nested clip), `ClipScrollNode::update_all_transforms` now takes about as long as `ClipScrollNode::update_all_transforms` and `FrameBuilder::recalculate_clip_scroll_nodes` together (as one would expect). Meanwhile a per-clip node cost has been eliminated, meaning that rendering should be slightly faster overall.

glennw:
I didn't get to this today, sorry. I'll look at it first thing tomorrow.

The CI seems to be failing on mac - it might be a float accuracy issue with the clip changes that requires re-generating the reference images?

mrobinson:
@glennw I'm pretty sure the failure here is due to your recent fix for the behavior of `inner_rect`. To keep this fix, I'll need to implement the thing we discussed which was the id assignment for each axis-aligned group of coordinate spaces. I've almost got that finished, but I need to fix one more issue. Hopefully I can have an updated PR for this tomorrow.

glennw:
OK, I'll hold off on reviewing until the updated patch. @kvark pointed out to me that one of my assumptions in that fix was incorrect - so if there's a better way to fix it that still passes the new reftest, go for it.

mrobinson:
@glennw Okay. I think the branch is ready now. The good thing about the change is that now we can start using the `coordinate_system_id` for more kinds of optimizations.

webrender/src/render_task.rs
```diff
@@ -303,53 +341,54 @@ impl RenderTask {
...
-            .chain(extra_clip.iter())
-            .filter(|work_item| {
+        let clips: Vec<_> = ClipChainNodeIter { current: raw_clips }
+            .chain(ClipChainNodeIter { current: extra_clip })
+            .filter_map(|node| {
+                let work_item = node.work_item.clone();
+
+                // We cannot apply any kind of optimization based on inner rectangles when the
+                // clips exists in another axis-aligned coordinate system.
+                if node.work_item.has_compatible_coordinate_system(current_coordinate_system_id) {
```
kvark: I don't think this is correct. Inner rects are valid regardless of the coordinate systems.

mrobinson: This is just a replacement for the workaround in https://github.com/servo/webrender/pull/1820. Would it be better for me to add a FIXME and open an issue?

kvark: yes, please

glennw:
This looks good to me - but we should do a gecko/servo try run before merging too.

mrobinson:
Here's the try run for this PR: https://treeherder.mozilla.org/#/jobs?repo=try&revision=8e48d88f50d99977ba93d86b7a36dd9c58c9d5cd&selectedJob=136489766

There are a few tests failing, but it seems like the differences are subpixel. Yet, this PR should change rendering at all. I will investigate a bit more.

glennw:
@mrobinson Looking at the reference image that was updated here. There appears to be AA being applied along the axis-aligned edges of the rectangle. That shouldn't be occurring - those edges should be pixel perfect. It's quite possible that fixing that will fix the gecko reftest errors, I think.

I wonder if an extra clip / mask is being applied that wasn't previously?

glennw:
It's probably worth fixing that ref test in wrench before looking into the gecko failures, since they are probably the same issue.

mrobinson:
Okay. After looking at this a bit I found two bugs in my original PR. The first was that the condition for the inner_rect workaround was reversed (!). Additionally the outer bounds of the clip were not being passed properly down the ClipScrollTree hierarchy. I have fixed these issues and have done the rebase.

Here is the try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=604e9e610ea14c65039f2a5b9e258ce23585b982&selectedJob=136791157

Of the unexpected results:
 1. Some are due sticky fixes I made recently
 1. Two unexpected passes appear to be intermittent tests deviations
 1. One is an accessibility failure that looks completely unrelated

kvark:
@mrobinson awesome, thank you!
@bors-servo r+

kvark:
>  Downloading euclid v0.15.2
warning: spurious network error (2 tries remaining): failed to get 200 response from `https://crates.io/api/v1/crates/euclid/0.15.2/download`, got 503
warning: spurious network error (1 tries remaining): failed to get 200 response from `https://crates.io/api/v1/crates/euclid/0.15.2/download`, got 503
error: unable to get packages from source

@bors-servo retry

mrobinson:
The failure looks like it was related to the merge conflict. I rebased the branch.

@bors-servo retry

mrobinson:
@bors-servo retry
 * "The job exceeded the maximum time limit for jobs, and has been terminated." Seems that this was an issue with the build infrastructure.

PR #1834, opened 2017-10-10
Improve deserialization performance

jrmuizel:
I haven't yet tested how much this improves deserialization performance but it should be noticeable.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1834)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -17,9 +17,10 @@ use YuvImageDisplayItem;
 use bincode;
 use serde::{Deserialize, Serialize, Serializer};
 use serde::ser::{SerializeMap, SerializeSeq};
-use std::io::Write;
+use std::io::{Read, Write};
 use std::{io, ptr};
 use std::marker::PhantomData;
+use std::slice;
 use time::precise_time_ns;
 
 // We don't want to push a long text-run. If a text-run is too long, split it into several parts.
@@ -171,6 +172,7 @@ fn skip_slice<T: for<'de> Deserialize<'de>>(
     (range, count)
 }
 
+
 impl<'a> BuiltDisplayListIter<'a> {
     pub fn new(list: &'a BuiltDisplayList) -> Self {
         Self::new_with_list_and_data(list, list.item_slice())
@@ -221,7 +223,7 @@ impl<'a> BuiltDisplayListIter<'a> {
                 return None;
             }
 
-            self.cur_item = bincode::deserialize_from(&mut self.data, bincode::Infinite)
+            self.cur_item = bincode::deserialize_from(&mut UnsafeReader::new(&mut self.data), bincode::Infinite)
                 .expect("MEH: malicious process?");
 
             match self.cur_item.item {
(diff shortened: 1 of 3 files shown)


webrender_api/src/display_list.rs
```diff
@@ -179,6 +179,45 @@ fn skip_slice<T: for<'de> Deserialize<'de>>(
     (range, count)
 }
 
+struct UnsafeReader<'a: 'b, 'b> {
```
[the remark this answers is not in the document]
Gankro: Also why it's desirable.

jrmuizel:
With this version and an updated bincode (https://github.com/TyOverby/bincode/pull/207) I go from 140fps to 200fps on my gmail yaml

webrender_api/src/display_list.rs
```diff
@@ -525,6 +527,73 @@ fn serialize_fast<T: Serialize>(vec: &mut Vec<u8>, e: &T) {
     debug_assert!(((w.0 as usize) - (vec.as_ptr() as usize)) == vec.len());
 }
 
+// This uses a (start, end) representation instead of (start, len) so that
+// only need to update a single field as we read through it. This gives
```
Gankro: "gives" at end of line is junk

webrender_api/src/display_list.rs
```diff
@@ -525,6 +527,73 @@ fn serialize_fast<T: Serialize>(vec: &mut Vec<u8>, e: &T) {
...
+    // Second, we only need to adjust 'start' after reading and it's only adjusted by a
+    // constant. This allows LLVM to avoid adjusting the length field after ever read
+    // and lets it be aggregated into a single adjustment.
+    #[inline(always)]
+    fn read_internal(&mut self, buf: &mut [u8]) {
+        // this is safe because we panic if start + buf.len() > end
+        unsafe {
+            if self.start.offset(buf.len() as isize) > self.end {
+                panic!();
+            }
```
Gankro: This if should just be 

`assert!(self.start.offset(buf.len() as isize) <= self.end, "UnsafeReader: wrote past end of target");`

webrender_api/src/display_list.rs
```diff
@@ -525,6 +527,73 @@ fn serialize_fast<T: Serialize>(vec: &mut Vec<u8>, e: &T) {
     debug_assert!(((w.0 as usize) - (vec.as_ptr() as usize)) == vec.len());
 }
 
+// This uses a (start, end) representation instead of (start, len) so that
+// only need to update a single field as we read through it. This gives
+// makes it easier for llvm to understand what's going on. (https://github.com/rust-lang/rust/issues/45068)
+// we update the slice only once we're done reading
+struct UnsafeReader<'a: 'b, 'b> {
```
Gankro: This is actually totally save to use, so maybe just ExactReader?

glennw:
Wow, that's a nice FPS improvement!

Gankro:
r=me with requested changes

hherman1:
Hi! Rust noob here,

Would it make sense to make this available in the rust stl if it’s so much faster than the built in?

Gankro:
Probably not, for a few reasons. It's specific to how we're using serde (which isn't part of std), and is also unsound if you're not using serde exactly like us. Basically we're relying on the fact that all of our serialization code is machine-generated, so we can't ever get expected and actual size wrong.

PR #1835, opened 2017-10-10
Move hardware composite shader to a single file, to match others.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1835)


glennw:
r? anyone

emilio:
@bors-servo r+

Makes sense.

PR #1836, opened 2017-10-10
Optimize skip_slice by recording serialized length

Gankro:
For @jrmuizel to test.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1836)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -152,23 +152,25 @@ fn skip_slice<T: for<'de> Deserialize<'de>>(
     data: &mut &[u8],
 ) -> (ItemRange<T>, usize) {
     let base = list.data.as_ptr() as usize;
-    let start = data.as_ptr() as usize;
 
-    // Read through the values (this is a bit of a hack to reuse logic)
-    let mut iter = AuxIter::<T>::new(*data);
-    let count = iter.len();
-    for _ in &mut iter {}
-    let end = iter.data.as_ptr() as usize;
+    let byte_size: usize = bincode::deserialize_from(data, bincode::Infinite)
+                                    .expect("MEH: malicious input?");
+    let start = data.as_ptr() as usize;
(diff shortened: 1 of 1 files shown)


glennw:
@Gankro do we want to review / merge this?

glennw:
@Gankro do we want to review / merge this?

Gankro:
Still need to do perf eval, gotten distracted the last few days.

jrmuizel:
I tried performance testing this. I couldn't get solid numbers. WebRender seems to have some weird bimodal timing behaviour that makes measuring header.

glennw:
@Gankro I can try to do some profiling with this tomorrow, if that would be helpful?

glennw:
I did some simple profiling on one of our benchmarks, with the following:

```
vblank_mode=0 ../target/release/wrench -r show benchmarks/text-rendering.yaml
```

This uses ```-r``` to replay the display list every frame, and ```vblank_mode=0``` disables vsync on Intel GPUs. The benchmark draws 64 reasonably long text runs of varying colors and font sizes, so it contains a lot of glyphs in the display list.

Without: ~325 fps.
With: ~385 fps.

So, at least in this test case, it's a very significant win. It'd be interesting to see if others can reproduce similar results.

With this patch, this is what the profile graph looks look:

profile

The two areas I've marked with a red dashed line are time spent inside deserialization functions / iterators, so hopefully there's still quite a lot of wins to be found there. There's certainly a couple of other big targets in that profile I can look at performance-wise in the future too.

Gankro:
Wow that's really surprising since text is one of the slices that shouldn't actually need this optimization (with a Sufficiently Smart Compiler).

Seems pretty convincing to me it's a good win. Let's ship it.

jrmuizel:
https://github.com/serde-rs/serde/issues/855 is probably the next place where we're going to get deserialization wins. @Gankro will look at it when he gets a chance.

Issue #1832, opened 2017-10-08
YAML reader confuses order of border colors


mstange:
The order should be "top right bottom left" (and that's what yaml_frame_writer.rs uses), but yaml_frame_reader reads it in the order of "top left bottom right".

https://github.com/servo/webrender/blob/59502ed15ade1441e4e86c1a64c2826baa937296/wrench/src/yaml_frame_reader.rs#L705-L720

hgallagher1993:
Not sure if I can self assign but I'll take this anyway

glennw:
@hgallagher1993 Thanks!

glennw:
This happened.

PR #1837, opened 2017-10-10
Fix order of border colors

hgallagher1993:
From #1832

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1837)


mstange:
Could you add a test for this?

hgallagher1993:
I wouldn't really know where to start tbh, this is only my second or third task on Webrender and I've never written any sort of test for it before. . .at the same time though I'd be happy to try with a couple of pointers :-)

glennw:
What we'd generally use is a reftest in wrench (the WR testing tool). However, in this case there are already several reftests that have a border with a different color for each edge. Since these tests fail with this change, just updating the test expectations should be sufficient.

You can see the test failures here https://travis-ci.org/servo/webrender/jobs/286108271. You can also see them locally by running:

```cd wrench
./headless reftest
```

In the case of this PR, the reference images are PNGs, so you'll need to re-generate them. You can do that for an individual reference by running:

```
./headless png <path to individual reference test>
```

Once that's done, running the ```reftest``` command above should pass all tests. One additional complication is that I usually manually crop the output PNG reference images to the approximate size of the test.

hgallagher1993:
@glennw That's perfect thanks, I'll get to it later on this evening

glennw:
@hgallagher1993 Sorry, I missed that you had updated this! There seems to be a difference in the degenerate-curve png - perhaps the border color needs to be specified for a different side in that test?

glennw:
@hgallagher1993 Will you have time in the next few days to fix up that last reftest image?

hgallagher1993:
@glennw Yes, sorry about the delay I can look at it tomorrow evening

glennw:
@hgallagher1993 No worries, thank you!

hgallagher1993:
@glennw Hi so I have had a look at the degenerate-curve but didn't get it fixed but then something came up in the last week that I'd rather not go into here but anyway if there's no rush with this I can get back to it next week or I can enable edits by maintainers either if there is.

glennw:
@hgallagher1993 It's not super urgent, next week is fine, thanks!

hgallagher1993:
@glennw I've finally got back to this now and the degenerate curve looks the same as before and 1 or 2 tests failed after a rebase so they should be fixed as well, reftests pass locally anyway.

glennw:
@hgallagher1993 Looks good, thanks! Would you mind squashing those commits into one and then we'll get this merged?

hgallagher1993:
@glennw Squashed now

Issue #1831, opened 2017-10-07
Box shadow edge radius needs to be affected by the spread radius


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1406608

[Testcase](https://bug1406608.bmoattachments.org/attachment.cgi?id=8916204)

Expected vs actual results:
<img width="395" alt="screen shot 2017-10-07 at 1 58 04 am" src="https://user-images.githubusercontent.com/961291/31305244-066da77a-ab03-11e7-89da-46f758d9e4af.png">

glennw:
This will be covered by the changes coming in https://github.com/servo/webrender/issues/1774.

glennw:
Fixed by #1841 - if there's remaining edge cases, we can open specific issues.

Issue #130, opened 2016-01-07
Some box shadows aren't drawn correctly.


glennw:
Loading http://thany.nl/apps/BoxShadows/ with WR shows several incorrect box-shadow renderings.

glennw:
@pcwalton I was looking for a good box shadow test page to check the current state of the box-shadow before I switch how it does clipping and came across this. It has a few layout bugs, but it's fairly easy to compare a heap of box shadow examples to other browser implementations.

pygy:
Inset shadows look fine, but outer ones are all made of a solid rectangle, no tapering, no rounding.

glennw:
Fixed by #1841. We can open specific bugs if there are any more correctness issues.

PR #1841, opened 2017-10-10
Switch box-shadows to use brush masks (and add basic brush prim).

glennw:
This introduces a new brush primitive type. Right now, this is
extremely basic and doesn't provide much new functionality. As
more items are ported to use it, the functionality of the brush
shader type will expand (mostly the VS parts).

The major change in this patch is to remove the old box-shadow
shader and switch box shadows to be drawn via a brush mask +
gaussian blur approach.

This *is* a performance regression for small / simple box shadows
compared to previously, but this is only temporary. The goal of
this (large) patch is to switch over to a different box-shadow
technique and address correctness issues first, and then we can
apply a number of optimizations on top of that. It *is* a performance
win on very large box shadows, which was the performance worst
case for the previous box shadow shader.

The changes included:
 * Extend blur shader to support both A8 and RGBA8 blurs.
 * Basic brush shader and primitive code.
 * Brush mask shader.
 * Remove old box shadow shader.
 * Move some common clip functions to ellipse.glsl.
 * Extend cache image shader to support both A8 and RGBA8 sources.
 * Support clip masks in cache image shader.
 * Remove box shadow primitive type.
 * Add reftests for spread, offset, blur, border radius box-shadow variations.
 * Support adjusting the border radii of box shadows as defined by spec.

Next steps:
 * Extend the mask shader to support arbitrary border radii per corner.
 * Apply various optimizations to the brush and blur code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1841)


glennw:
r? @kvark 

There's a couple of blockers before this can be merged - but it can probably be reviewed now.

The main issue is that the clip mask shader used for fast path box shadows (with zero blur) introduces a faint line of AA in some cases where it shouldn't. I'm a little unsure if I should leave this here until @nical finishes up his work on the AA changes, which may just resolve that problem. That causes a few test failures in the Servo WPT suite. @nical Do you think your AA work is likely to fix that, and how soon do you think it will be ready?

kvark:
Reviewed 30 of 30 files at r1.

---

*[webrender/res/brush.glsl, line 63 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBmAiQ_OQA0JVqnvav:-KwBmAiR3MFHdzDVYgkD:bfkx8s8) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/brush.glsl#L63)):*
> ```GLSL
>     // will be expanded to support transform picture types (the common kind).
>     vec2 pos = pic_task.target_rect.p0 + aPosition.xy * pic_task.target_rect.size;
>     vLocalPos = aPosition.xy * pic_task.target_rect.size / uDevicePixelRatio;
> ```

I'm concerned that this makes `vLocalPos` to live in a separate space from `vLocalRect`, which is confusing

---

*[webrender/res/cs_text_run.glsl, line 23 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBosmj8DE4KOrTl01M:-KwBosmkzPRujdrXU-RP:b1hshd7) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/cs_text_run.glsl#L23)):*
> ```GLSL
>     int picture_address = prim.user_data2;
> 
>     // Fetch the parent text-shadow for this primitive. This allows the code
> ```

nit: comment still implies the shadow

---

*[webrender/res/ellipse.glsl, line 70 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBoy_1tBOkwX--aZC4:-KwBoy_1tBOkwX--aZC5:b-wncfc4) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/ellipse.glsl#L70)):*
> ```GLSL
> 
> float clip_against_ellipse_if_needed(vec2 pos,
>                                      float current_distance,
> ```

nit: would be nicer with the new formatting style

---

*[webrender/res/ps_cache_image.glsl, line 68 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBuCu8ksvnPKHt7Hq-:-KwBuCu8ksvnPKHt7Hq0:b-vdbf8n) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/ps_cache_image.glsl#L68)):*
> ```GLSL
> 
>         // Apply the clip mask
>         color.a = min(color.a, do_clip());
> ```

I guess this will now become easier as @mstange convinced us to use multiplication ;)

---

*[webrender/res/ps_line.glsl, line 102 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBuTjCL1-CJzlxf1Ts:-KwBuTjCL1-CJzlxf1Tt:b5a0kwb) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/ps_line.glsl#L102)):*
> ```GLSL
> #ifdef WR_FEATURE_CACHE
>     int picture_address = prim.user_data0;
>     PrimitiveGeometry shadow_geom = fetch_primitive_geometry(picture_address);
> ```

should `shadow_geom` be renamed to `picture_geom`?

---

*[webrender/src/frame_builder.rs, line 1330 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBvmpt2_CAzx_pGcBb:-KwBvmpt2_CAzx_pGcBc:bpqdnr0) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/src/frame_builder.rs#L1330)):*
> ```Rust
>                         shadow_rect,
>                         BorderRadius::uniform(shadow_radius),
>                         ClipMode::ClipOut
> ```

Shouldn't this be `ClipIn`?

---

*[webrender/src/frame_builder.rs, line 1341 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBvpLyg8FoH5eJF0uW:-KwBvpLyg8FoH5eJF0uX:b-lsrwt5) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/src/frame_builder.rs#L1341)):*
> ```Rust
>                     );
> 
>                     self.add_primitive(
> ```

`add_primitive` could possibly go outside the `match`

---

*[webrender/src/frame_builder.rs, line 1431 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBwAJTkxymFktgkoE3:-KwBwAJTkxymFktgkoE4:b-nuq2iq) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/src/frame_builder.rs#L1431)):*
> ```Rust
>                     ));
> 
>                     self.add_primitive(
> ```

similarly, this could possibly be moved out of the `match`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1841)*

glennw:
Review status: 26 of 30 files reviewed at latest revision, 8 unresolved discussions.

---

*[webrender/res/brush.glsl, line 63 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBmAiQ_OQA0JVqnvav:-KwEG2VSGH7GHen-bPsZ:beo30xj) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/brush.glsl#L63)):*

I'm concerned that this makes `vLocalPos` to live in a separate space from `vLocalRect`, which is confusing
</blockquote></details>

Now that I look at this again, I'm not sure it correctly accounts for multiple text shadows with different offsets. I'll take a detailed look in the morning and either explain further with more comments, or fix if incorrect.

---

*[webrender/res/cs_text_run.glsl, line 23 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBosmj8DE4KOrTl01M:-KwEGEVxmbCPkFtbyw7W:b-896fix) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/cs_text_run.glsl#L23)):*

nit: comment still implies the shadow
</blockquote></details>

Done.

---

*[webrender/res/ellipse.glsl, line 70 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBoy_1tBOkwX--aZC4:-KwEGFKlVojjYi5CBpLg:b-896fix) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/ellipse.glsl#L70)):*

nit: would be nicer with the new formatting style
</blockquote></details>

Done.

---

*[webrender/res/ps_line.glsl, line 102 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBuTjCL1-CJzlxf1Ts:-KwEGJUgbFRW9lkF1_QT:b-896fix) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/res/ps_line.glsl#L102)):*

should `shadow_geom` be renamed to `picture_geom`?
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 1330 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBvmpt2_CAzx_pGcBb:-KwEGKa3ByrLut7_jxBc:b-16kq4z) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/src/frame_builder.rs#L1330)):*

Shouldn't this be `ClipIn`?
</blockquote></details>

I think ClipOut is named somewhat confusingly - it is the opposite of Clip (that is, it allows pixels outside the specific region).

---

*[webrender/src/frame_builder.rs, line 1341 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBvpLyg8FoH5eJF0uW:-KwEGSY_Jg6DAgf5ea1w:b-896fix) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/src/frame_builder.rs#L1341)):*

`add_primitive` could possibly go outside the `match`
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 1431 at r1](https://reviewable.io:443/reviews/servo/webrender/1841#-KwBwAJTkxymFktgkoE3:-KwEGTOLSaiPhtSnI8my:b-896fix) ([raw file](https://github.com/servo/webrender/blob/718901a79af180dfc6cde97cf240dc2ce3b53be0/webrender/src/frame_builder.rs#L1431)):*

similarly, this could possibly be moved out of the `match`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1841)*

glennw:
@kvark Thanks for the review! I've addressed most of those comments. The remaining issues blocking this are:

* I need to revisit the shadow offset comment above.
* I have a local fix for the AA edge + reftest issue I mentioned above. I think I'll wait until the clip chain stuff from @mrobinson lands first though, to make sure I don't break anything there.
* I still need to do a gecko try run once I get all the Servo tests passing with the AA fix mentioned above.

webrender/res/ellipse.glsl
```diff
@@ -66,4 +66,56 @@ float distance_to_ellipse(vec2 p, vec2 radii) {
...
+                                                      vec2(-1.0),
+                                                      afwidth);
+
+    current_distance = clip_against_ellipse_if_needed(pos,
+                                                      current_distance,
+                                                      clip_center_radius_bl,
+                                                      vec2(1.0, -1.0),
+                                                      afwidth);
+
+    return smoothstep(0.0, afwidth, 1.0 - current_distance);
```
nical: This implementation has a few bugs (see my ramblings in #1882), you should replace it by the corrected rounded_rect implementation from #1822.
I don't know if that's related to the aa glitch you referred to, though.

glennw: Yes, the patch in #1822 fixes the test failures in this patch - it's due to the change in ```init_transform_fs```.

glennw: If we land #1822 soon-ish, I'll rebase this patch on top of that.

glennw:
Fixes #130 
Fixes #1831

Makes it trivial to implement #1430 once this lands.

glennw:
I've rebased this over the AA change from @nical.

Gecko try run is here:
https://hg.mozilla.org/try/pushloghtml?changeset=58922e35932a38a0e9247e6f3a50845d67b02774
https://treeherder.mozilla.org/#/jobs?repo=try&revision=58922e35932a38a0e9247e6f3a50845d67b02774

There are 6 new passes, and 6 new fails.

It's likely some of these are related to other recent changes in WR (I don't think the fuzziness has been updated for the AA patch yet).

I'll do some further investigation on those failures and see if they are fuzziness issues or bugs that need to be resolved before landing this.

glennw:
Looking at the unexpected passes / fails:

Probably related to @Gankro 's text-shadow change:
```background-clip-text-2.html``` PASS.
```basic-opacity.html``` PASS.

Probably related to @nical 's AA change:
```preserve3d-6a.html``` FAIL (more fuzziness needed)
```backface-visibility-2.html``` FAIL (more fuzziness needed)
```split-intersect1.html``` PASS
```bugs/289480.html#top``` PASS

New box-shadow PASS from this patch:
```boxshadow-inner-basic.html``` PASS
```boxshadow-border-radius-int.html``` PASS

New box-shadow FAIL from this patch:
```box-shadow/boxshadow-large-border-radius.html``` 
```boxshadow-color-rounding-middle.html``` 
```boxshadow-inset-neg-spread.html```
```boxshadow-inset-large-offset.html```
```box-decoration-break-with-inset-box-shadow-1.html```

None of those box-shadow fails looks drastically wrong, but some of them have more fuzziness than I would *expect* to need (although most of them have fuzziness on skia / d2d too). I'll investigate some of those now.

Gankro:
Can confirm those two passes are what I was targeting with my patch.

glennw:
New try run:

https://hg.mozilla.org/try/rev/007c2de84545170b1b25b627c8a0149d6cff5350
https://treeherder.mozilla.org/#/jobs?repo=try&revision=007c2de84545170b1b25b627c8a0149d6cff5350&selectedJob=137389129

New PASSes related to this change:
```boxshadow-inner-basic.html```
```boxshadow-border-radius-int.html```

New FAILs related to this change:
```box-decoration-break-with-inset-box-shadow-1.html``` - imperceptible differences. Already marked fuzzy - needs a bit more.
```boxshadow-large-border-radius.html``` - blur has changed slightly. Already marked fuzzy - needs a bit more.
```boxshadow-inset-neg-spread.html``` - imperceptible differences. Already marked fuzzy - needs a bit more.
```boxshadow-inset-large-offset.html``` - Looks odd - already has a huge fuzzy range on WR, D2D and Skia. I think we could mark this with extra fuzziness, or as an expected FAIL for now.
 
Test differences *unrelated* to this change:
PASS - ```background-clip-text-2.html```
PASS - ```fallback-mark-stacking-1.html```
PASS - ```basic-opacity.html```
PASS - ```289480.html#top```
PASS - ```split-intersect1.html```
PASS - ```615121-2.html```
FAIL - ```preserve3d-6a.html```
FAIL - ```backface-visibility-2.html```

@staktrace @nical @kvark Given the above, I think this is probably ready to go now. Are you happy with the fuzziness changes described above?

staktrace:
@glennw I agree with your reftest analysis and am fine with this landing. Thanks for doing the try push!

staktrace:
Also for future reference, if you look at the most recent try runs in the wr-future-update bug the links to try pushes there will show which reftest differences were caused by stuff already landed and that I'm aware of. So in this case all the unrelated reftest failures that you were seeing I already have patches to adjust the fuzziness for.

kvark:
@staktrace @glennw thanks for being extremely cautious here!
:shipit: 🚀 🎢 
@bors-servo r+

glennw:
@staktrace A heads up - you need to update the bindings for this enum change - https://github.com/servo/webrender/pull/1841/files#diff-937670fa1a68b9bb7dab70622773f50d - otherwise, you get serde panics when sending a display list with box shadows.

@jrmuizel Also a heads up - this does add a perf regression for box shadows. If you see large pink lines in the GPU profiler, don't worry, it's only temporary. I'll be working on the optimizations on top of this PR this week.

staktrace:
So in the [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=2e33837af5fcbafe277ac73b35973dae3dfb987a) that included this update the reftest results are slightly different. There was no change to `boxshadow-inner-basic.html` which was supposed to be a PASS. And looking at the analyzer results I feel like `boxshadow-inset-neg-spread.html` and `boxshadow-inset-large-offset.html` should both be marked as failing as the difference between the test and reference cases is quite visible. For the former there's a light grey rect that's visible once you see the diff and know what to look for, and for the latter even though there's already a large fuzz applied, the files look visibly the same when I load them in non-WR firefox.

`boxshadow-large-border-radius.html` already had a visible difference and it's made worse but I'll fuzz it, and same with `box-decoration-with-inset-box-shadow-1.html`. In terms of gecko reftests it seems like this change overall made things worse and only fixed one reftest.

PR #1845, opened 2017-10-11
Fix nested clips in sticky frames

### LINKED ISSUE #1777: Introducing a regular clip inside a sticky clip results in lack of sticky behaviour
(This is mostly a clone of #1751, since part (a) of that issue, the lack of sticky behaviour, is still not fixed)

Run the scrolling.rs example, and scroll the nested frame. Observe how the dark blue rect (which is initially not visible, but becomes visible as you scroll the nested frame down) behaves. In particular, it is sticky at two different ranges, once when it is near the bottom of the nested scroll frame, and once more when it is near the top of the nested scroll frame.

Now apply this patch, which just introduces a single clip inside the sticky frame. Note that the clip rect is the same as the bounds of the rect inside the sticky frame, so this clip should have no visual effect.
```patch
index d335206b..f9747c03 100644
@@ -116,18 +116,21 @@ impl Example for App {
                     Some(StickySideConstraint {
                         margin: 10.0,
                         max_offset: -40.0,
                     }),
                     None,
                 ),
             );
             builder.push_clip_id(sticky_id);
+            let dummy_clip = builder.define_clip(None, (50, 350).by(50, 50), vec![], None);
+            builder.push_clip_id(dummy_clip);
             let info = LayoutPrimitiveInfo::new((50, 350).by(50, 50));
             builder.push_rect(&info, ColorF::new(0.5, 0.5, 1.0, 1.0));
+            builder.pop_clip_id(); // dummy_clip
             builder.pop_clip_id(); // sticky_id

             // just for good measure add another teal square further down and to
             // the right, which can be scrolled into view by the user
             let info = LayoutPrimitiveInfo::new((250, 350).to(300, 400));
             builder.push_rect(&info, ColorF::new(0.0, 1.0, 1.0, 1.0));

```

Build with this and scroll. Observe how the rect doesn't stick any more.

/cc @mrobinson

mrobinson:
Pass on sticky offsets as scrolling offsets in a way that ensures that
nested clips move properly with their parent sticky frame.

Fixes #1777.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1845)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -41,7 +41,6 @@ pub struct ClipScrollTree {
 pub struct TransformUpdateState {
     pub parent_reference_frame_transform: LayerToWorldTransform,
     pub parent_combined_viewport_rect: LayerRect,
-    pub parent_scroll_offset: LayerVector2D,
     pub parent_accumulated_scroll_offset: LayerVector2D,
     pub nearest_scrolling_ancestor_offset: LayerVector2D,
     pub nearest_scrolling_ancestor_viewport: LayerRect,
@@ -312,7 +311,6 @@ impl ClipScrollTree {
                 0.0,
             ),
             parent_combined_viewport_rect: root_viewport,
-            parent_scroll_offset: LayerVector2D::zero(),
             parent_accumulated_scroll_offset: LayerVector2D::zero(),
             nearest_scrolling_ancestor_offset: LayerVector2D::zero(),
             nearest_scrolling_ancestor_viewport: LayerRect::zero(),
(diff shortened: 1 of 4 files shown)


webrender/src/clip_scroll_tree.rs
```diff
@@ -338,27 +336,34 @@ impl ClipScrollTree {
...
+                NodeType::Clip(..) => TransformUpdateState {
                     parent_combined_viewport_rect: node.combined_local_viewport_rect,
-                    parent_scroll_offset: LayerVector2D::zero(),
                     ..*state
                 },
                 NodeType::ScrollFrame(ref scrolling) => TransformUpdateState {
-                    parent_combined_viewport_rect: node.combined_local_viewport_rect,
-                    parent_scroll_offset: scrolling.offset,
+                    parent_combined_viewport_rect:
+                        node.combined_local_viewport_rect.translate(&-scrolling.offset),
```
kvark: wouldn't this change the semantics of scroll frames? I thought the PR only targeting sticky frames logic

mrobinson: Previously, `parent_scroll_offset` and  `parent_combined_viewport_rect`  were combined to calculate `scrolled_parent_combined_clip` at https://github.com/servo/webrender/pull/1845/files#diff-6e285308c5268349a08cb330b8f08781L259. Since this no longer happens, we can remove `parent_scroll_offset`. I need to pass along the combined rectangle, because sticky frames need the scroll offset passed along (in `accumulated_scroll_offset`), but `parent_combined_viewport_rect` already had the offset applied as part of the `combined_local_viewport_rect` calculation.

All this means that the semantics of scroll frames don't change, but the value that is passed in `parent_combined_viewport_rect` does. `update_transform` now interprets this values as the parent combined viewport rect in the coordinate system of the children.

kvark:
Thanks @mrobinson ! Do we need a try push for Gecko?

mrobinson:
Here's the try output for this PR. Looks like the other than one intermittent there is one unexpected pass related to sticky behavior. This is a good sign, I think. :)

https://treeherder.mozilla.org/#/jobs?repo=try&revision=5488e453de1f0d108637163bbf632a3446186a1c&selectedJob=136461343

PR #1849, opened 2017-10-11
Make the _with_parent variants of clip/scrollframe creation functions…

staktrace:
… actually do something useful

Turns out the parent_id field on ClipDisplayItem and
ScrollFrameDisplayItem are never even used, and what I really wanted to
be able to override was the clip_and_scroll field on the display item.

So this deletes the parent_id fields and modifies the override functions
to allow overriding the clip_and_scroll field instead.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1849)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -699,6 +699,22 @@ impl DisplayListBuilder {
         )
     }
 
+    fn push_item_with_scrollinfo(
+        &mut self,
+        item: SpecificDisplayItem,
+        info: &LayoutPrimitiveInfo,
+        scrollinfo: ClipAndScrollInfo
+    ) {
+        serialize_fast(
+            &mut self.data,
+            &DisplayItem {
+                item,
+                clip_and_scroll: scrollinfo,
+                info: *info,
+            },
+        )
+    }
+
     fn push_new_empty_item(&mut self, item: SpecificDisplayItem) {
         let info = LayoutPrimitiveInfo::new(LayoutRect::zero());
         serialize_fast(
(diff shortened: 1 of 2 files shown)


glennw:
This seems reasonable to me - let's merge it now. Would be good for @mrobinson to just double check when he comes online though.

@bors-servo r+

webrender_api/src/display_list.rs
```diff
@@ -1112,21 +1128,21 @@ impl DisplayListBuilder {
...
+            scrollinfo,
             content_rect,
             clip_rect,
             complex_clips,
             image_mask,
             scroll_sensitivity)
     }
 
-    pub fn define_scroll_frame_with_parent<I>(
+    pub fn define_scroll_frame_with_scrollinfo<I>(
```
mrobinson: A ClipScrollNode can only have one parent, so I'm not sure I understand the benefit of passing in a ClipScrollInfo here. I think this should probably just accept a ClipId to avoid ambiguity. This would mean these API points would maintain with "with_parent" naming.

webrender_api/src/display_list.rs
```diff
@@ -699,6 +699,22 @@ impl DisplayListBuilder {
         )
     }
 
+    fn push_item_with_scrollinfo(
```
mrobinson: This should probably be called either `push_item_with_scroll_info` or `push_item_with_clip_scroll_info.`

webrender_api/src/display_list.rs
staktrace: Ah that would explain why my previous attempt to define the clip chains and scroll chains more independently didn't work. This seems like an inconsistent API, because if you push a rect inside a ClipAndScrollInfo it means one thing (scroll by this clip id, clip by this other clip id) but if you define a clip inside a ClipAndScrollInfo it means something else (scroll by this clip, ignore this other clip id).

Issue #1848, opened 2017-10-11
non-blurred text shadows aren't painted atomically


Gankro:
`data:text/html,<u style="font:99px serif;color:red;text-shadow:black 10px 20px 0px">test`

The shadow glyphs are painted over the real line:

![screen shot 2017-10-11 at 3 19 16 pm](https://user-images.githubusercontent.com/1136864/31461497-acf9d45a-ae97-11e7-981d-fe4b3733b681.png)

Adding a 0.01 blur (forcing slow-path rendering) fixes it:

![screen shot 2017-10-11 at 3 19 37 pm](https://user-images.githubusercontent.com/1136864/31461511-b8f50edc-ae97-11e7-9439-98d1dd0cbc83.png)

Gankro:
Filed in bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1407652

Gankro:
Seems to be because underlines appear before glyphs (writing reftest now).

Gankro:
Ah yeah, the fast path code basically does

```
// same for line/rect
add_text:
  let prim = ...;
  for shadow in fast_shadows:
    self.add_prim(prim)
  self.add_prim(prim)
```

But it needs to be creating a picture (or whatever equivalent) to collect up all of a shadow's contents into, just like the slow path does.

PR #1850, opened 2017-10-11
Add reftest for atomic shadow painting, issue #1848

Gankro:
(fails on master)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1850)


Gankro:
Replacing with actual fix (and more complex test)

PR #1852, opened 2017-10-11
Add an in-tree copy of the reftest analyzer.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1852)


glennw:
r? anyone

Gankro:
What's the motivation?

glennw:
Many of our contributors don't have a copy of Gecko cloned, and it's useful to have the correct analyzer version available (the WPT analyzer version is subtly different and won't load Gecko reftest files).

Gankro:
You don't need a checkout of gecko, it's live at https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml

glennw:
It seems useful to me to have it self contained in tree, rather than pointing people to a URL. But if everyone would prefer to just use the live version, I guess that's fine.

nical:
That's fine by me as well. It's self-contained, small and we can easily remove it if we change our minds. @bors-servo r+

PR #1853, opened 2017-10-12
Ensure shadows are painted in the right order and mode

Gankro:
It feels like *something* better could be done here, but I'm not really sure what.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1853)


glennw:
It seems like we could still get things out of order - but I'm probably just not thinking it through properly. What would happen, for example, if ```has_fast_shadows``` is false, but then a new shadow is pushed on after that which is a fast shadow? Is there any situation we could get ordering issues depending on whether fast / blur shadows are pushed on before / after another shadow? Would it be simpler to just always buffer every shadow element?

Gankro:
I’d like to avoid always buffering if possible, but you’re right I didn’t fully think about it. I was abusing my knowledge that gecko/servo never manage to do anything other than pushpushpush-content-poppoppop. 

I think the Right solution is to just check in pop if there are no fast shadows in the stack rather is_empty, but I’m gonna sleep on it.

glennw:
Cool, thanks. FWIW always buffering *might* actually be a win in this case. Instead of having the fast shadows vec be allocated each time, if we just use vecs that are in the frame builder and ```clear``` them after each shadow context, that might avoid more allocations than we do now anyway? Haven't though that through - but mentioning just in case it's relevant here.

Gankro:
Ok new version up with another reftest to prove that this works right.

As discussed with @glennw this replaces PopShadow with PopAllShadows, which will require a trivial update from gecko/servo (replacing a for loop with a single PopAllShadows).

I also found a bug in how we set the mode for fast shadows, which is fixed by the last patch in the series.

Gankro:
@glennw @staktrace 

Servo patch for breakage: https://github.com/Gankro/servo/commit/b2561faf49efca4659fdd2e9483d41fc9718b01b

Gecko patch for breakage (also needs cbindgen run):

<details>

```patch
diff --git a/gfx/webrender_bindings/WebRenderAPI.cpp b/gfx/webrender_bindings/WebRenderAPI.cpp
index cfaf88842e0f..77b6665cfbb5 100644
--- a/gfx/webrender_bindings/WebRenderAPI.cpp
+++ b/gfx/webrender_bindings/WebRenderAPI.cpp
@@ -1084,9 +1084,9 @@ DisplayListBuilder::PushShadow(const wr::LayoutRect& aRect,
 }
 
 void
-DisplayListBuilder::PopShadow()
+DisplayListBuilder::PopAllShadows()
 {
-  wr_dp_pop_shadow(mWrState);
+  wr_dp_pop_all_shadows(mWrState);
 }
 
 void
diff --git a/gfx/webrender_bindings/WebRenderAPI.h b/gfx/webrender_bindings/WebRenderAPI.h
index 6ba3652f1a5a..0e521a608314 100644
--- a/gfx/webrender_bindings/WebRenderAPI.h
+++ b/gfx/webrender_bindings/WebRenderAPI.h
@@ -374,7 +374,7 @@ public:
                       bool aIsBackfaceVisible,
                       const wr::Shadow& aShadow);
 
-  void PopShadow();
+  void PopAllShadows();
 
 
 
index d8e889f2873e..1747b57e08dd 100644
@@ -1458,10 +1458,10 @@ pub extern "C" fn wr_dp_push_shadow(state: &mut WrState,
 }
 
 #[no_mangle]
-pub extern "C" fn wr_dp_pop_shadow(state: &mut WrState) {
+pub extern "C" fn wr_dp_pop_all_shadows(state: &mut WrState) {
     debug_assert!(unsafe { is_in_main_thread() });
 
-    state.frame_builder.dl_builder.pop_shadow();
+    state.frame_builder.dl_builder.pop_all_shadows();
 }
 
 #[no_mangle]
diff --git a/layout/generic/TextDrawTarget.h b/layout/generic/TextDrawTarget.h
index 90d2eb1088d8..ba0a78dc922f 100644
--- a/layout/generic/TextDrawTarget.h
+++ b/layout/generic/TextDrawTarget.h
@@ -137,16 +137,16 @@ public:
   AppendShadow(const wr::Shadow& aShadow)
   {
     mBuilder.PushShadow(mBoundsRect, mClipRect, mBackfaceVisible, aShadow);
-    mShadowCount++;
+    mHasShadows = true;
   }
 
   void
   TerminateShadows()
   {
-    for (size_t i = 0; i < mShadowCount; ++i) {
-      mBuilder.PopShadow();
+    if (mHasShadows) {
+      mBuilder.PopAllShadows();
+      mHasShadows = false;
     }
-    mShadowCount = 0;
   }
 
   void
@@ -220,8 +220,8 @@ private:
   // * Text stroke
   bool mHasUnsupportedFeatures = false;
 
-  // Number of shadows currently pushed, so we can pop them later.
-  size_t mShadowCount = 0;
+  // Whether PopAllShadows needs to be called
+  bool mHasShadows = false;
 
   // Things used to push to webrender
   wr::DisplayListBuilder& mBuilder;
```

</details>


(haven't tried compiling these, but this should be ~95% of the work if there's any issue)

glennw:
Reviewed 4 of 5 files at r1, 19 of 19 files at r2, 5 of 5 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1853)*

glennw:
Looks good, thanks!

@bors-servo r+

PR #1855, opened 2017-10-12
Support clip masks for text runs.

glennw:
This switches our subpixel rendering over to a two pass method
that supports normal alpha blending.

There's some tradeoffs involved here:
 * +++ We get clip masks on subpixel text.
 * +++ CPU batching is much better (no batch breaks on text color change).
 * --- We draw the subpixel text pixels twice.

On balance, this is likely to be a reasonable performance win, because:
 (a) We are typically CPU limited rather than GPU limited.
 (b) The time in text run FS is typically a small portion of GPU time.

More importantly, this gives us correct clip masking on text, and also
allows us to apply some optimizations in the future:

* Take advantage of dual source blending where available. This is
  available on most recent hardware, and allows clip masking plus
  subpixel rendering in a single pass.
* Consider using the old constant color blend method when there
  is no clip mask present, if the two pass method does show up
  as a significant cost.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1855)


glennw:
r? @kvark 

(Passes servo/wrench tests OK, but I still need to do a gecko try with this).

glennw:
Gecko try:

https://hg.mozilla.org/try/rev/2d2e7a960008f0724f2f2799233d21bac923763d
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2d2e7a960008f0724f2f2799233d21bac923763d

glennw:
Gecko try is green on all configs.

kvark:
Code looks good. I haven't evaluated the performance impact yet though.

---

Reviewed 16 of 16 files at r1.

---

*[webrender/res/ps_text_run.glsl, line 88 at r1](https://reviewable.io:443/reviews/servo/webrender/1855#-KwFmn7jZJa10f_PM-Yl:-KwFmn7kahJYWnt2rOZg:bdgy5qc) ([raw file](https://github.com/servo/webrender/blob/dfaa0deb0d6fafec6426330afd5822d2748952c7/webrender/res/ps_text_run.glsl#L88)):*
> ```GLSL
>     switch (uMode) {
>         case MODE_ALPHA:
>             oFragColor = alpha * color * vColor;
> ```

I vaguely recall driver issues with output variables assigned in a switch

---

*[webrender/src/device.rs, line 1466 at r1](https://reviewable.io:443/reviews/servo/webrender/1855#-KwFnsrYEt1knMOYcDtS:-KwFnsrYEt1knMOYcDtT:b2kz71u) ([raw file](https://github.com/servo/webrender/blob/dfaa0deb0d6fafec6426330afd5822d2748952c7/webrender/src/device.rs#L1466)):*
> ```Rust
>             .uniform_1f(program.u_device_pixel_ratio, self.device_pixel_ratio);
>         self.gl
>             .uniform_1i(program.u_mode, mode);
> ```

I wonder if we can get any perf from caching those uniforms. Device pixel ratio and the mode don't change often.

---

*[webrender/src/renderer.rs, line 900 at r1](https://reviewable.io:443/reviews/servo/webrender/1855#-KwFnQnxKS-fkOxmnrm8:-KwFnQnxKS-fkOxmnrm9:b9181cn) ([raw file](https://github.com/servo/webrender/blob/dfaa0deb0d6fafec6426330afd5822d2748952c7/webrender/src/renderer.rs#L900)):*
> ```Rust
>         mode: M,
>         renderer_errors: &mut Vec<RendererError>,
>     ) where M: Into<i32> {
> ```

would not harm to have a proper name for this type instead of `i32`, even if just a typedef

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1855#-:-KwFoMTR3xgiEcBPl1Sy:b-8a798e)*

kvark:
> More importantly, this gives us correct clip masking on text

I guess there is no alternative atm :)
@bors-servo r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1855#-:-KwGnNjvhwnXa7AGedPO:b-s2noj3)*

kvark:
Actually, Glenn is here soon, so I'll wait to let him address the nits.
@bors-servo r-

glennw:
Review status: 15 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/res/ps_text_run.glsl, line 88 at r1](https://reviewable.io:443/reviews/servo/webrender/1855#-KwFmn7jZJa10f_PM-Yl:-KwHLGfvQWeNCLzqW_CB:b-w1fqjp) ([raw file](https://github.com/servo/webrender/blob/dfaa0deb0d6fafec6426330afd5822d2748952c7/webrender/res/ps_text_run.glsl#L88)):*

I vaguely recall driver issues with output variables assigned in a switch
</blockquote></details>

I *think* that was only in a very specific case, not in any switch. Maybe merge and see if we encounter any issues?

---

*[webrender/src/device.rs, line 1466 at r1](https://reviewable.io:443/reviews/servo/webrender/1855#-KwFnsrYEt1knMOYcDtS:-KwHLOzLqGXxSqo00ukD:b-9jojgm) ([raw file](https://github.com/servo/webrender/blob/dfaa0deb0d6fafec6426330afd5822d2748952c7/webrender/src/device.rs#L1466)):*

I wonder if we can get any perf from caching those uniforms. Device pixel ratio and the mode don't change often.
</blockquote></details>

https://github.com/servo/webrender/issues/1864

---

*[webrender/src/renderer.rs, line 900 at r1](https://reviewable.io:443/reviews/servo/webrender/1855#-KwFnQnxKS-fkOxmnrm8:-KwHLbq2hDy46pR3sx-_:b-896fix) ([raw file](https://github.com/servo/webrender/blob/dfaa0deb0d6fafec6426330afd5822d2748952c7/webrender/src/renderer.rs#L900)):*

would not harm to have a proper name for this type instead of `i32`, even if just a typedef
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1855)*

glennw:
@kvark Thanks for the quick review!

Comments addressed.

kvark:
>I think that was only in a very specific case, not in any switch. Maybe merge and see if we encounter any issues?

I think we should make it an informal rule to have a single code path for all output writes. It makes drivers happy, and allows easier code navigation and reasoning.

Anyhow, it's fine to follow-up with. Thanks for addressing my nits!
@bors-servo r+

glennw:
Yea, that sounds like a good idea - perhaps we need a style guide for shader writing...

staktrace:
@glennw  I believe this change is causing shader compilation failures on Windows:

thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Link("debug_font", "Precisions for uniform \'uMode\' differ between vertex and fragment shaders\n")', src\libcore\result.rs:860

Full regression range is in https://bugzilla.mozilla.org/show_bug.cgi?id=1407213#c10 but this PR is the one that touches uMode.

kvark:
@staktrace could you try with #1872 ?

staktrace:
Thanks for the quick PR! Try push is in progress at https://treeherder.mozilla.org/#/jobs?repo=try&revision=911f826e93460538f7906168377958a3259b30af

staktrace:
I screwed up the reftest.list files in that try push so it didn't run the reftests. Here's another one (this includes #1873 as well): https://treeherder.mozilla.org/#/jobs?repo=try&revision=1f38daa67ff395256f26bc280b099a930dc26c38

staktrace:
Fix verified, thanks!

PR #1859, opened 2017-10-12
Extract a helper method to build a checkerboard image

staktrace:
This makes it a little easier to exercise push_image calls in other
sample apps by allowing some code reuse.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1859)


Gankro:
r=me

PR #1858, opened 2017-10-12
Update to bincode 0.9

jrmuizel:
bincode 0.8.1 has been yanked because of API breakage

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1858)

PR #1866, opened 2017-10-13
Update to bincode 0.9

staktrace:
Same as #1858 but with the ipc-channel update as welll. I'm basically stealing this from jeff in the hope of landing it sooner as it is blocking even attempted WR updates in gecko.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1866)


jrmuizel:
And I'm stealn' the review

PR #1867, opened 2017-10-13
Expose ClipMode in public API.

### LINKED ISSUE #1862: Expose negative clip rects publically
https://mdn.mozillademos.org/files/4563/editable_select.html requires us to draw a border everywhere but where the text is. Internally we already have this notion for box shadows. We should consider exposing it publicly.

glennw:
Fixes #1862.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1867)


glennw:
r? @mrobinson 

cc @Gankro

Gankro:
Are we ok with doing this before moving ComplexClips out of local clips?

glennw:
I think so - it's only one extra int, so I don't think it's too bad. And an extra win to be had when to *do* move complex clips out :) But it's up to you - happy to hold off on it if you prefer.

mrobinson:
This seems okay to me. I don't think it will affect removing rounded rectangles from LocalClips.

glennw:
Rebased.

r? @kvark or @mrobinson

staktrace:
@glennw @mrobinson This PR appears to be causing RenderThread panics in gecko, looks like: `GECKO(1114) | thread 'RenderBackend' panicked at 'MEH: malicious input?: Custom("invalid value: integer `3857049061`, expected variant index 0 <= i < 2")', /checkout/src/libcore/result.rs:906:4` for example. Do we need to change things on the gecko side?

staktrace:
Ah, looks like we need to actually initialize this new clipmode member. I have a patch that I'm testing now.

PR #1868, opened 2017-10-13
Fix diacritic glyphs on Linux.

### LINKED ISSUE #1846: Some diacritics not rendering on linux
https://gankro.github.io/blah/reftests/text/fallback-mark-stacking-1.html
https://gankro.github.io/blah/reftests/text/fallback-mark-stacking-1-notref.html

works fine on mac/windows, which *suggests* it's a bug in webrender's font code. 

On mac with webrendest gecko, these render as:

screen shot 2017-10-11 at 2 16 23 pm
screen shot 2017-10-11 at 2 16 30 pm

On linux, you only see the Xs.

glennw:
Diacritic glyphs have a valid width and height, but a
zero advance. Previously we were discarding these but
they are valid glyphs. The code also checks for a width
and height of 0 below, which catches non-printable glyphs
such as spaces.

Fixes #1846.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1868)

### DIFF
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -278,12 +278,6 @@ impl FontContext {
     ) -> Option<GlyphDimensions> {
         let metrics = unsafe { &(*slot).metrics };
 
-        // If there's no advance, no need to consider this glyph
-        // for layout.
-        if metrics.horiAdvance == 0 {
-            return None
-        }
-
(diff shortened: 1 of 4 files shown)


glennw:
r? @nical or @kvark 

cc @Gankro

kvark:
Looks good!
@bors-servo r+

PR #1869, opened 2017-10-13
Fix plane splitting precision error.

mephisto41:
We use f32 for plane splitting calculation. But f32 would have some precision error. I use f64 instead to prevent precision error.

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1869)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -17,7 +17,7 @@ use border::ImageBorderSegment;
 use clip::{ClipMode, ClipRegion, ClipSource, ClipSources, ClipStore, Contains};
 use clip_scroll_node::{ClipInfo, ClipScrollNode, NodeType};
 use clip_scroll_tree::ClipScrollTree;
-use euclid::{SideOffsets2D, vec2, vec3};
+use euclid::{SideOffsets2D, TypedTransform3D, vec2, vec3};
 use frame::FrameId;
 use gpu_cache::GpuCache;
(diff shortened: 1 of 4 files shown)


webrender/src/frame_builder.rs
```diff
@@ -48,15 +48,32 @@ fn make_polygon(
...
     //TODO: only work with `isolated_items_bounds.size` worth of space
     // This can be achieved by moving the `origin` shift
     // from the primitive local coordinates into the layer transformation.
     // Which in turn needs it to be a render task property obeyed by all primitives
     // upon rendering, possibly not limited to `write_*_vertex` implementations.
     let size = stacking_context.isolated_items_bounds.bottom_right();
     let bounds = LayerRect::new(LayerPoint::zero(), LayerSize::new(size.x, size.y));
-    Polygon::from_transformed_rect(bounds, node.world_content_transform, anchor)
+    let mat = TypedTransform3D::row_major(
+        node.world_content_transform.m11 as f64,
```
kvark: uh, that is unfortunate... @nical can we get a `cast()` here please?
At the very least, we may assign the source matrix to a temporary variable so that it's not repeated all over

kvark:
Thank you!
@bors-servo r+

webrender/src/frame_builder.rs
nical: Ouch! yeah, that would be handy.

PR #1871, opened 2017-10-13
Add routes to publish pulse messages

staktrace:
This will cause the CI tasks to be indexed, which makes them easily
viewable at https://tools.taskcluster.net/index/garbage.webrender.ci/
But more importantly, it will cause TaskCluster to send pulse messages
when these tasks are completed, which we can listen for. One way is to
use the pulse inspector at https://tools.taskcluster.net/pulse-inspector
with the route "#.webrender.#" on the following exchanges:
  exchange/taskcluster-queue/v1/task-completed (for successful tasks)
  exchange/taskcluster-queue/v1/task-failed (for failed tasks)
  exchange/taskcluster-queue/v1/task-exception (for taskcluster errors)
We can also listen for these in any program using the pulse API (e.g.
via the mozillapulse python library).
This can eventually be used to notify bors/homu if the test jobs succeed
so it can merge without waiting for the slower travis runs to complete.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1871)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -66,6 +66,8 @@ tasks:
           servo-tidy &&
           (cd wrench && python headless.py reftest) &&
           (cd wrench && cargo build --release --features=debugger)
+    routes:
+      - "index.garbage.webrender.ci.{{event.head.user.login}}.{{event.head.repo.branch}}.linux-release"
   - metadata:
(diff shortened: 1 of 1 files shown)


glennw:
@bors-servo r+

:smile:

PR #1873, opened 2017-10-13
Fix aa clamping for transparent text

Gankro:
My previous patch accidentally resulted in upgrading mono to alpha.

Currently causing two failures in the linux CI since it uses Mono everywhere.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1873)


staktrace:
Verified that this fixes the gecko reftest failures, thanks!

PR #1879, opened 2017-10-15
Bump bitflags to 1.0

Eijebong:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1879)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,6 +106,11 @@ version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "bitflags"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "block"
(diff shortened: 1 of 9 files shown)


glennw:
Why do we need to allow duplicate packages in the ```servo-tidy``` script?

Eijebong:
Because of clap which still depends on 0.9 :/ https://github.com/kbknapp/clap-rs/pull/1063

glennw:
@Eijebong Ah, thanks for the info. ```clap``` is only used by wrench, so I guess this should be fine for Servo.

@bors-servo r+

glennw:
@Eijebong Oh, wait - is the rustc stable version requirement going to cause an issue for Gecko compiling against WR?

Eijebong:
https://github.com/servo/servo/pull/18809#issuecomment-336662383 Shouldn't be an issue anymore no :)

glennw:
@bors-servo r+

:+1:

Issue #1794, opened 2017-10-03
IdNamespace is unique only within webrender instance


sotaroikeda:
IdNamespace allocation is managed within webrender instance . Therefore there could be a case that different RenderApi could have same namespace_id between different webrender instances.

It causes the problem like [Bug 1405241](https://bugzilla.mozilla.org/show_bug.cgi?id=1405241) on gecko since gecko assumes unique IdNamespace even between different webrender instances.

I am not sure if the problem should be addressed gecko side or webrender side.

kvark:
I believe the end goal is to have a single WR instance anyway, that is managing multiple independent documents/scenes. In that case, WR will not need to have any built-in protection from multiple instances collision. However, in order to allow this to be patched on the Gecko side, WR does need to provide some way to make those `IdNamespace` different.

glennw:
https://github.com/servo/webrender/pull/1882 has landed.

PR #1882, opened 2017-10-16
Use a global static variable for IdNamespace counting.

JerryShih:
#1794
Try to use a simple global static variable for that id counting. That's not thread safe. All render-backend should be created in the same thread. If we try to handle the thread problem, I will turn to use thread_local.

cc @sotaroikeda 
r? @kvark @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1882)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -22,6 +22,7 @@ use resource_cache::ResourceCache;
 use scene::Scene;
 #[cfg(feature = "debugger")]
 use serde_json;
+use std::sync::atomic::{ATOMIC_USIZE_INIT, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Sender;
 use std::u32;
@@ -124,6 +125,9 @@ enum DocumentOp {
     Rendered(RendererFrame),
 }
 
+/// The unique id for WR resource identification.
+static NEXT_NAMESPACE_ID: AtomicUsize = ATOMIC_USIZE_INIT;
+
 /// The render backend is responsible for transforming high level display lists into
 /// GPU-friendly work which is then submitted to the renderer in the form of a frame::Frame.
 ///
@@ -133,7 +137,6 @@ pub struct RenderBackend {
     payload_rx: PayloadReceiver,
     payload_tx: PayloadSender,
     result_tx: Sender<ResultMsg>,
-    next_namespace_id: IdNamespace,
     default_device_pixel_ratio: f32,
 
     gpu_cache: GpuCache,
@@ -163,6 +166,9 @@ impl RenderBackend {
         blob_image_renderer: Option<Box<BlobImageRenderer>>,
         enable_render_on_scroll: bool,
     ) -> RenderBackend {
+        // The namespace_id should start from 1.
+        NEXT_NAMESPACE_ID.fetch_add(1, Ordering::Relaxed);
+
         let resource_cache = ResourceCache::new(texture_cache, workers, blob_image_renderer);
 
         register_thread_with_profiler("Backend".to_string());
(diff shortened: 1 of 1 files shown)


nical:
> All render-backend should be created in the same thread. 

This is currently not the case. With gecko, each window creates a separate render backend thread, so this code is running on multiple thread.

nical:
Another possibility could be to have some of the bits of the id namespace representing a unique id per webrender instance, and the other bits representing a unique id within the instance (although a global atomic variable is simple enough as well).

JerryShih:
This version try to use a simple global atomic variable for the id counting. https://github.com/servo/webrender/pull/1882/commits/10f79745525367e86f52980b6de85cb934f41fe0

webrender/src/render_backend.rs
```diff
@@ -22,6 +22,7 @@ use resource_cache::ResourceCache;
 use scene::Scene;
 #[cfg(feature = "debugger")]
 use serde_json;
+use std::sync::atomic::{ATOMIC_USIZE_INIT, AtomicUsize, Ordering};
```
JerryShih: The u32is not stable now. Use usize instead.

error: use of unstable library feature 'integer_atomics' (see issue #32976)
  --> src\render_backend.rs:25:25
   |
25 | use std::sync::atomic::{AtomicU32, Ordering};

JerryShih:
r? @glennw @kvark @nical 
https://github.com/servo/webrender/pull/1882/commits/10f79745525367e86f52980b6de85cb934f41fe0

sotaroikeda:
We want to avoid to use 0 as id name space. But current patch seems to do it.

JerryShih:
Thanks. If this approach looks good, I will change the init value to 1.

webrender/src/render_backend.rs
```diff
@@ -124,6 +125,9 @@ enum DocumentOp {
     Rendered(RendererFrame),
 }
 
+/// The unique id for WR resource identification.
+static mut NEXT_NAMESPACE_ID: AtomicUsize = ATOMIC_USIZE_INIT;
```
glennw: This doesn't need the ```mut``` for an atomic.

webrender/src/render_backend.rs
```diff
@@ -452,9 +454,10 @@ impl RenderBackend {
                     tx.send(glyph_indices).unwrap();
                 }
                 ApiMsg::CloneApi(sender) => {
-                    let namespace = self.next_namespace_id;
-                    self.next_namespace_id = IdNamespace(namespace.0 + 1);
-                    sender.send(namespace).unwrap();
+                    unsafe {
```
glennw: Once the ```mut``` above is removed, the ```unsafe``` can also be removed.

webrender/src/render_backend.rs
```diff
@@ -163,6 +166,9 @@ impl RenderBackend {
         blob_image_renderer: Option<Box<BlobImageRenderer>>,
         enable_render_on_scroll: bool,
     ) -> RenderBackend {
+        // The namespace_id should start from 1.
```
JerryShih: I can't set the value to 1 directly, so I just increase the value in RenderBackend::new().

error: `std::sync::atomic::AtomicUsize::new` is not yet stable as a const fn
   --> src\render_backend.rs:129:41
    |
129 | static NEXT_NAMESPACE_ID: AtomicUsize = AtomicUsize::new(1);
    |                                         ^^^^^^^^^^^^^^^^^^^
    |
    = help: in Nightly builds, add `#![feature(const_atomic_usize_new)]` to the crate attributes to enable

JerryShih:
@glennw 
r? again.

glennw:
Looks good, thank you!

@bors-servo r+

PR #1883, opened 2017-10-16
Allow resource updates to affect pipeline epochs.

nical:
This adds `DocumentMsg::UpdatePipelineResources` which allows updating resources *and* update a pipeline's epoch (useful when the resources being updated are as meaningful as layout updates, for gecko this is typically image updates).

This doesn't remove the global (as in not associated to a given pipeline) `UpdateResources` message because it is easier for servo to keep using that one for now and even for gecko there are several cases where the global resource updates message is a better fir like adding fonts and cleaning up after a pipeline is removed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1883)

### DIFF
--- a/webrender_api/src/api.rs
+++ b/webrender_api/src/api.rs
@@ -192,6 +192,11 @@ pub enum DocumentMsg {
         preserve_frame_state: bool,
         resources: ResourceUpdates,
     },
+    UpdatePipelineResources {
+        resources: ResourceUpdates,
+        pipeline_id: PipelineId,
+        epoch: Epoch,
+    },
     SetPageZoom(ZoomFactor),
     SetPinchZoom(ZoomFactor),
     SetPan(DeviceIntPoint),
@@ -214,6 +219,7 @@ impl fmt::Debug for DocumentMsg {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_str(match *self {
             DocumentMsg::SetDisplayList { .. } => "DocumentMsg::SetDisplayList",
+            DocumentMsg::UpdatePipelineResources { .. } => "DocumentMsg::UpdatePipelineResources",
             DocumentMsg::HitTest(..) => "DocumentMsg::HitTest",
             DocumentMsg::SetPageZoom(..) => "DocumentMsg::SetPageZoom",
             DocumentMsg::SetPinchZoom(..) => "DocumentMsg::SetPinchZoom",
@@ -461,7 +467,7 @@ impl RenderApi {
         ImageKey::new(self.namespace_id, new_id)
     }
 
-    /// Adds an image identified by the `ImageKey`.
+    /// Add/remove/update resources such as images and fonts.
     pub fn update_resources(&self, resources: ResourceUpdates) {
         if resources.updates.is_empty() {
             return;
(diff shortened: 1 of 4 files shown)


nical:
I still have to do some testing with this but feedback is welcome.
cc @kvark @sotaroikeda

webrender_api/src/api.rs
```diff
@@ -471,6 +477,24 @@ impl RenderApi {
...
+    /// This is similar to update_resources with the addition that it allows updating
+    /// a pipeline's epoch.
+    pub fn update_pipeline_resources(
+        &self,
+        resources: ResourceUpdates,
+        document_id: DocumentId,
+        pipeline_id: PipelineId,
+        epoch: Epoch,
+    ) {
+        self.send(document_id, DocumentMsg::UpdatePipelineResources {
```
kvark: Not sure I understand the semantics of this. What is the relation of `pipeline_id` and `document_id` with the given resource updates?
If the resources are supposed to be used by this pipeline/document only, we need to reflect it somehow. If not, then it's not clear why would the user want to bump the epoch of an arbitrary pipeline while doing some of the resource updates.

> This is similar to update_resources with the addition that it allows updating a pipeline's epoch.

This begs for a question why it's not just a separate method to bump the pipeline epoch.

nical:
> If the resources are supposed to be used by this pipeline/document only, we need to reflect it somehow.

We can look into doing that for the aesthetics of it, but there is no real need for it.

> This begs for a question why it's not just a separate method to bump the pipeline epoch.

Right. I think I'd be happiest with a separate message to bump an epoch. The problem with that is the atomicity of transactions that comes back. We'd need the guarantee that this message is processed with the rest of the transaction without interleaving any other messages.

It would be a lot easier handle this if instead of sending `DocumentMsg`s, one by one, we'd send them together (maybe just in an array):

```rust
enum DocumentMsg {
    SetDisplayList(..),
    UpdateResources(..),
    UpdateEpoch(..),
    ... 
}

enum ApiMsg {
    ...
    UpdateDocument(DocumentId, Vec<DocumentMsg>),
    ...
}
```
WebRender would only have to guarantee that all of the DocumentMsg in the array are processed before rendering a frame or looking at another ApiMsg.

I don't know if there would be an implicit cost to serializing large things like the display list in another array (I recently found out that there is on gecko's IPC system so I'm cautious). With this model I would have all of the flexibility I need to properly synchronize things in gecko.

I thought I'd have better chances of moving forward with this through a minor API addition (most shared textures are leaking in gecko's wr integration until we figure this out so while I don't mean to rush this, I wouldn't let mind coming with a fix reasonably soon), but if you'd be OK with a more fundamental change like this one, that would be even better.

nical:
I have an icomplete prototype of the transaction approach suggested above in a branch: https://github.com/nical/webrender/commit/b4928efab7fb9c9d438b93eaed96907f6f3a9862

In addition to letting me fix the issues related to bumping the epoch when a canvas or a video frame changes, this lets us generically handle the following things:

- update several pipelines in a single transaction and rebuild the scene only once. We currently have this use case in gecko because each video element is in a separate pipeline.
- update the window parameters in a transaction (I don't know how to reliably resize the window synchronously otherwise since we need to relayout synchronously).
- ...whatever else we'll end up find out requires being in a transaction along the way.

kvark:
We discussed this on today's call. Change is fine as a temporary measure.
@bors-servo r+

PR #1885, opened 2017-10-16
Make push_text accept an iterator, to avoid copies in gecko

Gankro:
This will probably be a **[breaking change]** for servo/gecko, but trivially fixed as seen in the wrench/example case.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1885)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -313,7 +313,7 @@ impl Example for App {
             let info = LayoutPrimitiveInfo::new(text_bounds);
             builder.push_text(
                 &info,
-                &glyphs,
+                glyphs.iter().cloned(),
                 font_instance_key,
                 ColorF::new(1.0, 1.0, 0.0, 1.0),
                 None,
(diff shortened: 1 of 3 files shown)


Gankro:
(don't merge quite yet, confirming that this can be done properly on the gecko side)

JerryShih:
Could you tell more why there has the copies from the original code?

`glyphs: &[GlyphInstance] // <= We use slice here, right?`

Gankro:
We have to create a temporary array/copy for the conversion from Gecko glyph to Wr glyph. By accepting an iterator we can do the transform element-by-element and avoid that.

glennw:
@Gankro Any updates on if we want this?

jrmuizel:
We're doing some changes on the Gecko side that are worth validating before we pursue this right now.

glennw:
Can this be closed for now?

Gankro:
Yeah I'll come back to this later, unless @jrmuizel wants to pick it up.

Gankro:
Status update: this is being handled upstream, by there no longer needing to be any space transform that we need to hide with an iterator. so this is wontfix

Issue #1765, opened 2017-09-27
Image masks and rounded clips not respected on text


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1403333

The text in this testcase is drawn without any fade-out effect:

```
---
root:
  items:
    - type: clip
      id: 1
      bounds: [0, 0, 240, 50]
      image-mask:
        image: "linear-gradient.png"
        rect: [0, 0, 240, 50]
        repeat: false
    - type: rect
      bounds: [0, 0, 240, 50]
      color: blue
      "clip-and-scroll": 1
    - bounds: [0, 0, 240, 50]
      glyphs: [55, 75, 76, 86, 3, 76, 86, 3, 87, 75, 72, 3, 69, 72, 86, 87]
      offsets: [16, 33, 35.533333, 33, 51.533333, 33, 60.4, 33, 72.833336, 33, 80.833336, 33, 89.7, 33, 102.13333, 33, 110.13333, 33, 119, 33, 135, 33, 149.2, 33, 157.2, 33, 173.2, 33, 187.4, 33, 196.26666, 33]
      size: 18
      color: black
      font: "../text/VeraBd.ttf"
      "clip-and-scroll": 1
```
Save to `wrench/reftests/mask/text.yaml`, and add this image with the name `linear-gradient.png` next to it: ![linear-gradient.png
<img width="502" alt="screen shot 2017-09-27 at 5 20 28 pm" src="https://user-images.githubusercontent.com/961291/30938505-349cca84-a3a8-11e7-8a42-83e7a8fcf087.png">
](https://user-images.githubusercontent.com/961291/30938420-f4213b34-a3a7-11e7-9ae8-de81ce83cf2f.png)

Screenshot:
<img width="502" alt="screen shot 2017-09-27 at 5 20 28 pm" src="https://user-images.githubusercontent.com/961291/30938520-3d73f88a-a3a8-11e7-96e2-5ea8da11d1cb.png">

You can see that the blue rectangle respects the mask but the text does not.

mstange:
Text also doesn't appear to respect rounded clip rects. So the testcase can probably be a lot simpler.

glennw:
We need to draw subpixel text that has a clip mask with a different blend mode. I'm going to hold off on this until the text-related work that @lsalzman is doing lands, so we don't conflict.

mstange:
@lsalzman's work has landed, so this is can now be worked on, I think.

glennw:
[...]

Issue #1881, opened 2017-10-16
Transparent text does not use subpixel AA


mstange:
Not sure if this bug is already covered by a different issue, but I'll file it anyway:

At the moment, only text with an opaque text color uses subpixel anti-aliasing. Text which is slightly transparent does not.

Testcase: `data:text/html,<p style="color:rgba(0,0,0,0.9)">Text`

glennw:
With the old subpixel blend method, this caused issues. But it should be easy to re-enable now.

PR #1886, opened 2017-10-17
Fix alpha fading on subpixel text.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1886)


glennw:
Fixes #1765.

glennw:
This is a squashed commit with fixes for #1765 and #1881.

PR #1887, opened 2017-10-17
Remove redundant nested cfg statements

staktrace:
Since the entire impl is wrapped in the same cfg, it seems unnecessary to put it inside the function bodies as well. This causes cbindgen to fail when parsing the file.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1887)


kvark:
@staktrace correct, those are leftovers from refactoring.
Thanks!
@bors-servo r+

PR #1889, opened 2017-10-17
Support different radii and elliptical radii in box shadows.

### LINKED ISSUE #1430: Support different box-shadow radii per corner.

glennw:
Now that box shadows are rendered via a normal mask + blur,
we can just pass through radii for each corner to handle
this case. Right now, the brush mask assumes it will handle
the entire rect. The next step is to support segmenting
parts of the mask generation, which will greatly help with
the performance of the new box shadow implementation.

Fixes #1430.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1889)


glennw:
@staktrace Heads up - this PR changes the public API of ```add_box_shadow``` - it now takes a complete ```BorderRadius``` struct instead of ```f32```.

staktrace:
@glennw Thanks for the heads-up. It's nice to have example files in the tree where I can just copy the changes that need to be applied on the caller side :)

PR #1891, opened 2017-10-18
[DO NOT MERGE] Box shadow performance optimization.

glennw:
Use a 9-patch for outset box shadows to improve performance.

This drastically reduces the size of the region that the blur
shader operates on. The blur shader itself is already much
slower than it should be, so this is a big performance win
on large box shadows.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1891)


glennw:
The performance of box shadows can be quite bad with the new blur approach on very large shadows, due to a simplistic initial implementation aiming for correctness only.

My plan is to work on the proper segment / brush support discussed in previous issues over the next few days. Once that lands, the performance of the box shadows will be significantly faster than previously, and support all the new features of this implementation.

My preference would be to hold off for a few more days until the "proper" solution is in place to reduce the size of box shadows clips and blurs, but this is here in case we decide the current box shadow performance is intolerable, as a stopgap solution.

glennw:
(This also includes https://github.com/servo/webrender/pull/1889 which is still in the merge queue, so the patch isn't as big as it looks).

glennw:
Note to self: doesn't scale with device pixel ratio correctly - need to update the cache image shader.

kvark:
looks pretty good! :heart: 

---

Reviewed 7 of 10 files at r2.
Review status: 7 of 18 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/picture.rs, line 162 at r2](https://reviewable.io:443/reviews/servo/webrender/1891#-KwktVzL0z4zD8yxiKp9:-KwktVzL0z4zD8yxiKpA:b5xgxba) ([raw file](https://github.com/servo/webrender/blob/d93feb4ae2c498894ae3e2bc4694e516fef7a35c/webrender/src/picture.rs#L162)):*
> ```Rust
>                 let cache_height =
>                     (prim_metadata.local_rect.size.height * prim_context.device_pixel_ratio).ceil() as i32;
>                 let cache_size = DeviceIntSize::new(cache_width, cache_height);
> ```

nit:
if we treat `prim_context.device_pixel_ratio` like `ScaleFactor<f32, LayerSpace, DeviceSpace>`, then this could be as simple as
```rust
let cache_size = (prim_metadata.local_rect.size * prim_context.device_pixel_ratio).ceil().to_i32();
```

---

*[webrender/src/tiling.rs, line 593 at r2](https://reviewable.io:443/reviews/servo/webrender/1891#-KwkuQuWMSVazCp6eIhe:-KwkuQuXREGaguhY240K:bhnogco) ([raw file](https://github.com/servo/webrender/blob/d93feb4ae2c498894ae3e2bc4694e516fef7a35c/webrender/src/tiling.rs#L593)):*
> ```Rust
>                             PictureKind::BoxShadow(..) => {
>                                 for i in 0..9 {
> ```

nit: would be nice to have some indication here on what `9` means. Perhaps, we could have an `enum XXX: u32` that we can cast to `u32` here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1891#-:-KwkupKM5S8Wtw_JR8Jr:bul53dz)*

glennw:
Closing in favor of https://github.com/servo/webrender/pull/1896 which is a simpler solution.

PR #1896, opened 2017-10-19
Optimize box shadows - only apply blurs where they can affect the shadow.

glennw:
Add support for regions to blur render tasks. These allow render tasks
to specify one or more regions where the blur should be applied. If
no blur region is supplied, the entire render task is blurred.

Also start expanding the Picture struct, with methods to prepare
it for rendering etc.

Finally, add the basic infrastructure for render tasks to specify
what color they should be cleared to. This allows tasks to specify
a non-default color, if required. This is useful for inset shadows
where we will avoid blurring the inside of the shadow rect.

For outset box shadows with a large blur radius and/or border radius
this patch makes them ~3-4x faster than previously. A follow up will
be to add the same blur regions for inset box shadows.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1896)


glennw:
r? @kvark 

I still need to do a try run and WPT tests - but it should be OK to review this when you have time.

kvark:
Reviewed 9 of 9 files at r1.

---

*[webrender/src/frame_builder.rs, line 581 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq1XQCjsGnltodItLZ:-Kwq1XQCjsGnltodItL_:b1fwsp7) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/frame_builder.rs#L581)):*
> ```Rust
> 
>         // Borrowcheck dance
>         let mut shadows = mem::replace(&mut self.shadow_prim_stack, Vec::new());
> ```

I'm quite surprised to see this. If we just iterated on `self.shadow_prim_stack.drain(..)` it wouldn't prevent us to use the prim_store. Could this dance be a leftover from the previous code?

---

*[webrender/src/frame_builder.rs, line 590 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq1OXhEt0ubVcwESHt:-Kwq1OXiu9cYv7qvmoig:br9kr6u) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/frame_builder.rs#L590)):*
> ```Rust
>                 let metadata = &mut self.prim_store.cpu_metadata[prim_index.0];
>                 let prim = &self.prim_store.cpu_pictures[metadata.cpu_prim_index.0];
>                 let shadow = prim.as_text_shadow();
> ```

all shadows are considered to be text shadows now?

---

*[webrender/src/frame_builder.rs, line 1364 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq2gos3A9w1EnF69xM:-Kwq2gouFcgZzcMy1c9G:b-6wgpct) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/frame_builder.rs#L1364)):*
> ```Rust
> 
>                     let pic_rect = shadow_rect.inflate(blur_offset, blur_offset);
>                     let blur_range = 3.0 * blur_radius;
> ```

what is `3.0` here? would be nice to make it a constant with a comment on top

---

*[webrender/src/frame_builder.rs, line 1385 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq3L4s-yGWe2YnjSVS:-Kwq3L4tfFIoFJy7Tw_o:b-l7u013) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/frame_builder.rs#L1385)):*
> ```Rust
>                     ) * 3.0;
> 
>                     blur_regions.push(LayerRect::from_floats(0.0, 0.0, tl.width, tl.height));
> ```

should we detect intersections here to avoid extra work on pixels covered by multiple regions?

---

*[webrender/src/picture.rs, line 35 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq3lnYQkKvU-BQh5lY:-Kwq3lnYQkKvU-BQh5lZ:bom4e90) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/picture.rs#L35)):*
> ```Rust
>         shadow: Shadow,
>     },
>     BoxShadow {
> ```

that's what I've been thinking of when reviewing the previous version of the PR, nice one!

---

*[webrender/src/prim_store.rs, line 1060 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq4m-xMtyEiwliPA37:-Kwq4m-xMtyEiwliPA38:b-7m9pri) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/prim_store.rs#L1060)):*
> ```Rust
>             PrimitiveKind::Picture => {
>                 self.cpu_pictures[metadata.cpu_prim_index.0]
>                     .prepare_for_render(
> ```

neat!

---

*[webrender/src/renderer.rs, line 2923 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq5D6dOMflCH7TrrxY:-Kwq5D6e0UyRPkOoPXOY:bwaxh6t) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/renderer.rs#L2923)):*
> ```Rust
>                 let (rect, _) = task.get_target_rect();
>                 self.device
>                     .clear_target_rect(Some(zero_color), None, rect);
> ```

so we clear on top of the previous clear here?

---

*[webrender/src/tiling.rs, line 1908 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq6t4z-76emDUTak1D:-Kwq6t5-36m56514S2sp:b-y2ia8x) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/tiling.rs#L1908)):*
> ```Rust
>         } else {
>             for region in &self.regions {
>                 instances.push(BlurInstance {
> ```

nit: could create a single `BlurInstance` before the branch, then doing `BlurInstance { region, .. blur_instance }` here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1896)*

kvark:
> For outset box shadows with a large blur radius and/or border radius
> this patch makes them ~3-4x faster than previously.

curious, what exactly made this case faster?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1896#-:-Kwq7MDB9g7UCfVxuU3E:bgywyf0)*

glennw:
Review status: 7 of 9 files reviewed at latest revision, 8 unresolved discussions.

---

*[webrender/src/frame_builder.rs, line 581 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq1XQCjsGnltodItLZ:-Kwr3ouKXx5ewAXQSUlG:bwm39qr) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/frame_builder.rs#L581)):*

I'm quite surprised to see this. If we just iterated on `self.shadow_prim_stack.drain(..)` it wouldn't prevent us to use the prim_store. Could this dance be a leftover from the previous code?
</blockquote></details>

It's the borrow on self for the call to add_primitive_to_draw_list() that is the problem. This could definitely be restructured and tidied up in the future though...

---

*[webrender/src/frame_builder.rs, line 590 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq1OXhEt0ubVcwESHt:-Kwr443J4TDGJtmMK3A5:b1k2jo3) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/frame_builder.rs#L590)):*

all shadows are considered to be text shadows now?
</blockquote></details>

For now, yes. The box shadows don't currently make use of the offset field in Shadow. So to avoid confusion I'm not using the Shadow struct. We might end up changing how this works in the future though.

---

*[webrender/src/frame_builder.rs, line 1364 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq2gos3A9w1EnF69xM:-Kwr6N5uSX1lbmX9VChR:b-896fix) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/frame_builder.rs#L1364)):*

what is `3.0` here? would be nice to make it a constant with a comment on top
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 2923 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq5D6dOMflCH7TrrxY:-Kwr6QQrYUFn4oADyTdf:bqubj9d) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/renderer.rs#L2923)):*

so we clear on top of the previous clear here?
</blockquote></details>

Yea. We might like to do something smarter here in the future.

---

*[webrender/src/tiling.rs, line 1908 at r1](https://reviewable.io:443/reviews/servo/webrender/1896#-Kwq6t4z-76emDUTak1D:-Kwr6lSAV9n9xXdJo0CV:b-896fix) ([raw file](https://github.com/servo/webrender/blob/509075efa9601a22c5799e7058f64c8f4e806651/webrender/src/tiling.rs#L1908)):*

nit: could create a single `BlurInstance` before the branch, then doing `BlurInstance { region, .. blur_instance }` here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1896)*

glennw:
@kvark The speed win comes because we only run the blur shader along the edges and corners of the mask rect. This won't be important once we have proper segmenting of the box shadows into a 9-patch (or even just a single corner in the case of uniform border radii), but it's a good win in the interim.

I think the review comments are all addressed now.

WPT tests on Servo are good and Gecko try run looks green:

https://hg.mozilla.org/try/rev/d23b7c67075bf7d553dbad12fd0b7390f4bfcc77
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d23b7c67075bf7d553dbad12fd0b7390f4bfcc77

kvark:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1896)*

kvark:
thank you!
@bors-servo r+

PR #1903, opened 2017-10-20
Pass the dirty rect down into the blob image

jrmuizel:
This lets us know which area of the image has changed and is useful
for merging the old blob data and the new blob data.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1903)


Gankro:
Note from discussion: this is just setting up public API for gecko to use in its trait impl.

@bors-servo r+

Gankro:
Bluh, https://github.com/servo/saltfs/pull/740 still needs-deploy

PR #1904, opened 2017-10-20
Require immutable, threadsafe render notifier at renderer initialization

jdm:
This addresses one of the top intermittent test failures in Servo and cleans up an awkward part of the renderer API. Instead of storing an optional notifier inside of a shared mutex, we require that the notifier is cloneable and is responsible for its own mutation requirements. This puts the onus on the implementation of a particular notifier to deal with threadsafety concerns, rather than making all consumers pay for it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1904)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -145,7 +145,7 @@ pub struct RenderBackend {
     frame_config: FrameBuilderConfig,
     documents: FastHashMap<DocumentId, Document>,
 
-    notifier: Arc<Mutex<Option<Box<RenderNotifier>>>>,
+    notifier: Arc<Mutex<Box<RenderNotifier>>>,
     recorder: Option<Box<ApiRecordingReceiver>>,
 
     enable_render_on_scroll: bool,
@@ -160,7 +160,7 @@ impl RenderBackend {
         default_device_pixel_ratio: f32,
         texture_cache: TextureCache,
         workers: Arc<ThreadPool>,
-        notifier: Arc<Mutex<Option<Box<RenderNotifier>>>>,
+        notifier: Arc<Mutex<Box<RenderNotifier>>>,
         frame_config: FrameBuilderConfig,
         recorder: Option<Box<ApiRecordingReceiver>>,
         blob_image_renderer: Option<Box<BlobImageRenderer>>,
@@ -445,7 +445,7 @@ impl RenderBackend {
                 }
                 Err(..) => {
                     let notifier = self.notifier.lock();
-                    notifier.unwrap().as_mut().unwrap().shut_down();
+                    notifier.unwrap().shut_down();
                     break;
                 }
             };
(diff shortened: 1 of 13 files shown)


webrender/src/render_backend.rs
```diff
@@ -160,12 +160,12 @@ impl RenderBackend {
...
         texture_cache: TextureCache,
         workers: Arc<ThreadPool>,
-        notifier: Arc<Mutex<Option<Box<RenderNotifier>>>>,
+        notifier: T,
         frame_config: FrameBuilderConfig,
         recorder: Option<Box<ApiRecordingReceiver>>,
         blob_image_renderer: Option<Box<BlobImageRenderer>>,
         enable_render_on_scroll: bool,
-    ) -> RenderBackend {
+    ) -> RenderBackend<T> {
```
kvark: nit: `Self`

webrender_api/src/api.rs
```diff
@@ -872,11 +872,11 @@ pub struct DynamicProperties {
     pub floats: Vec<PropertyValue<f32>>,
 }
 
-pub trait RenderNotifier: Send {
-    fn new_frame_ready(&mut self);
-    fn new_scroll_frame_ready(&mut self, composite_needed: bool);
-    fn external_event(&mut self, _evt: ExternalEvent) {
+pub trait RenderNotifier: Send + Clone {
```
kvark: would it make sense to provide an empty implementation, e.g. for `()`?

webrender/src/render_backend.rs
```diff
@@ -132,7 +132,7 @@ static NEXT_NAMESPACE_ID: AtomicUsize = ATOMIC_USIZE_INIT;
 /// GPU-friendly work which is then submitted to the renderer in the form of a frame::Frame.
 ///
 /// The render backend operates on its own thread.
-pub struct RenderBackend {
+pub struct RenderBackend<T> {
```
jrmuizel: Is there a reason this needs to be generic instead of using a trait object? Having it generic is going to increase wrench's already bad build times and we don't have any need to inline or devirtualize the notifier.

webrender_api/src/api.rs
jdm: I don't see the purpose if no code makes use of it.

webrender/src/render_backend.rs
jdm: I have rewritten the changes to retain the trait object.

glennw:
Reviewed 7 of 13 files at r1, 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1904)*

glennw:
@bors-servo r+

Thanks!

@staktrace Heads up - this will need an update to Gecko bindings. The render notifier is provided at initialization now and has slightly different requirements. Once this lands, I'll write a Gecko patch for that and post it here.

staktrace:
@glennw thanks for the heads up. Please post the patch to https://bugzil.la/wr-future-update when you have it, thanks!

glennw:
@staktrace OK, attached a patch to that bug. I'm unfamiliar with bugzilla workflow, but hopefully that should be what you need to update for the notifier API changes (I've tested them locally).

staktrace:
@glennw Perfect, thanks!

PR #1905, opened 2017-10-22
Only respect the text color's alpha component for Bitmap glyphs.

### LINKED ISSUE #1780: Transparent apple emoji aren't transparent
```
data:text/html,<html><head><meta charset="utf8"></head><body><span style="color: blue">%F0%9F%8E%89</span><span style="color: rgba(0,0,0,0.7)">%F0%9F%8E%89</span><span style="color: rgba(0,0,255,0.4)">%F0%9F%8E%89</span><span style="color: transparent">%F0%9F%8E%89</span>
```

```html
<html>
<head><meta charset="utf8"></head>
<body>
    <span style="color: blue">🎉</span>
    <span style="color: rgba(0,0,0,0.7)">🎉</span>
    <span style="color: rgba(0,0,255,0.4)">🎉</span>
    <span style="color: transparent">🎉</span>
</body>
```

Chrome:
screen shot 2017-09-29 at 3 00 01 pm

Webrendest Firefox:
screen shot 2017-09-29 at 3 00 22 pm

Partially this is my fault, because I missed that firefox treats this the same as synthetic bold, doing the transparency as a compositing affect (and so wr was being sent opaque colors). However this should be fixed by https://bugzilla.mozilla.org/show_bug.cgi?id=1404370, and yet with that patch we still get opaque.

### LINKED ISSUE #1876: Emoji incorrectly respect the text color again
See https://github.com/servo/webrender/issues/1780#issuecomment-335628687

mstange:
Fixes #1780.
Fixes #1876.

This restores a chunk of code that was added in #1754 and removed in #1816, with the difference that I use `color.a` as the alpha component instead of `1.0`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1905)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1155,10 +1155,17 @@ impl FrameBuilder {
             }
         }
 
+        let color = match font.render_mode {
+            FontRenderMode::Bitmap => ColorF::new(1.0, 1.0, 1.0, color.a),
+            FontRenderMode::Subpixel |
+            FontRenderMode::Alpha |
+            FontRenderMode::Mono => *color,
+        };
+
         let prim_font = FontInstance::new(
             font.font_key,
             font.size,
-            *color,
+            color,
             render_mode,
             font.subpx_dir,
             font.platform_options,


Gankro:
In the context of AppleColorEmoji, this change seem reasonable and correct.

However a big part of #1816 was refactoring how bitmap fonts actually work. Was that all just for that one font? Are there other fonts that this mode now applies to and wouldn't be correct?

webrender/src/frame_builder.rs
```diff
@@ -1155,10 +1155,17 @@ impl FrameBuilder {
             }
         }
 
+        let color = match font.render_mode {
+            FontRenderMode::Bitmap => ColorF::new(1.0, 1.0, 1.0, color.a),
+            FontRenderMode::Subpixel |
+            FontRenderMode::Alpha |
+            FontRenderMode::Mono => *color,
+        };
```
Gankro: As a minor style point, this could just be a ternary:

`let color = if font.render_mode == FontRenderMode::Bitmap { ... } else { ... };`

I don't expect this should need to be updated for other font modes.

mstange:
I don't know the answers to these questions. I'm hoping Glenn does :)

mstange:
I can add a new `FontRenderMode` value called `Color` and use that for Apple emojis. How does that sound, @glennw?

glennw:
@mstange How about instead we have something like (with better naming perhaps):

```
enum BitmapColorMode {
    Color,
    Alpha,
    // None if we need in the future to ignore color *and* alpha??
}

enum FontRenderMode {
    ...
    Bitmap(BitmapColorMode)
}
```

This allows us to specify how to treat the color for a bitmap font, without needing a new font render mode.

mstange:
@lsalzman had some concerns about this change. He is going to come up with a better fix.

PR #1906, opened 2017-10-23
Add a bit more fuzziness to subpx color test, for Ubuntu 17.10.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1906)


glennw:
r? anyone

PR #1908, opened 2017-10-23
Fix premultiplied problem for box shadow.

demo99:
There is a gecko[ bug 1410305](https://bugzilla.mozilla.org/show_bug.cgi?id=1410305) about the wrong rendering result of the box-shadow. It should be premultiplied alpha problem. We take the picture's blend mode as PremultipliedAlpha[1]. So I multiply the color's alpha when applying the blur texture. I'm not sure if it's a good place to do this.

[1]https://github.com/servo/webrender/blob/d741f472dd3d6c3441646f7bf4e714c71bea39b7/webrender/src/tiling.rs#L70

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1908)


mstange:
I think an alternate fix would be to use `color.premultiplied()` here: https://github.com/servo/webrender/blob/9e1180d124b450ca4493a7cf560e7aae84451119/webrender/src/picture.rs#L74-L79

I wonder what Glenn prefers.

kvark:
I'd prefer `color.premultiplied()` since it would mean less GPU work per pixel.

glennw:
A wrench reftest would also be good to have, so we don't regress this.

demo99:
Thanks. I will use color.premultiplied() and add a wrench test.

glennw:
Looks good, thanks!

@bors-servo r+

PR #1910, opened 2017-10-23
Always multiply alpha with the clip's alpha.

### LINKED ISSUE #1838: Change clip masks to multiply alpha instead of using min()
We have this sort of code in a bunch of places:

```glsl
  alpha = min(alpha, do_clip());
```

I'm surprised by this. Isn't applying a mask usually done by multiplying the alphas? If you use min(), aren't you going to lose some of the anti-aliasing of the shape if you apply the clip to something that's, say, only 10% opaque?

mstange:
Fixes #1838.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1910)


webrender/res/ps_angle_gradient.glsl
```diff
@@ -63,17 +63,6 @@ void main(void) {
...
-
-        // Apply the clip mask
-        color.a = min(color.a, do_clip());
-
-        // Pre-multiply the result.
-        color.rgb *= color.a;
-    }
-
-    oFragColor = color;
+    oFragColor = color * do_clip();
```
kvark: heh, nice :)

webrender/res/ps_border_corner.glsl
```diff
@@ -322,7 +322,7 @@ void main(void) {
     vec2 local_pos = vLocalPos;
 #endif
 
-    alpha = min(alpha, do_clip());
+    alpha *= do_clip();
```
kvark: nit: in shaders like this one, we can simply start with `float alpha = do_clip();`, remove that awkward `alpha = 0.0;` and end up with fewer lines ;)

mstange: As discussed on IRC, this won't work because then `init_transform_fs` would overwrite the value from `do_clip()`.

mstange:
I've started a try push at https://treeherder.mozilla.org/#/jobs?repo=try&revision=2b67e0b20991426902c97609c8be25c49ad9e44c .

mstange:
The try push shows one failure in box-decoration-break-with-inset-box-shadow-1.html. It's otherwise green. However, I don't understand the failure, and if I run that test on my machine, it already fails in the same way without my change. (It looks like a snapping-related issue that causes some box-shadows in the reference to be drawn 1px to the right of where they're drawn in the test.)

Most of the box-shadows in that test render very incorrectly anyway, which I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1411074 and https://github.com/servo/webrender/issues/1916 about. I think until those bugs are resolved, we can mark the test as random.

mstange:
Oh, I understand now: the offset between the test and the reference is already accounted for with a fuzzy annotation: `fuzzy-if(webrender,17-17,6785-6785)`. And the change in this PR just causes four more pixels to be slightly different: `max difference: 17, number of differing pixels: 6789`. So we can just adjust the fuzz.

glennw:
@bors-servo r+

That seems reasonable, thanks! @staktrace Heads up on a slight fuzziness change from this PR.

PR #1911, opened 2017-10-23
Draw white on black if the font is bright, because bright text uses a different amount of font dilation.

### LINKED ISSUE #1767: Consider using different glyph masks for dark and bright text on macOS, in order to match native font rendering
https://bugzilla.mozilla.org/show_bug.cgi?id=1410971

(Some of the text below is liberally stolen from @lsalzman's email to the Skia team.)

Somewhere between 10.9 and 10.11, Apple changed the apparent weighting of fonts when rendering with LCD smoothing depending on whether the foreground color of the text is light or dark. Light text now renders thinning than it used to, while dark text renders the same as it always has.

If WebRender wants to match this, it'll need to rasterize every glyph twice; once with bright text and once with dark text. Then it needs to select the correct mask based on the text color.

This is something that Firefox lost when we switched from CoreGraphics to Skia, because Skia does not implement this brightness-based shape switching. As a consequence, bright text on dark backgrounds appears thicker than in Safari. With the switch to WebRender, we have the option to restore the native behavior.

More details on how this seems to work:

On 10.9, if you render a glyph as white foreground on black background, decode the gamma to linear, it gives you the same result as if you render black foreground on white background, decode the gamma to linear, then invert the result.

However, on 10.11, this no longer holds true. The white foreground on black background case is less heavily dilated than it used to be, so the inverted black-on-white is no longer the same as white-on-black.

We further verified by animated text foreground color, and there is a single distinct pop once the text color reaches a certain threshold, so we're pretty sure there are only two sets of weightings being chosen between here.

We did a lot of playing around with the threshold condition, and we're pretty sure given R, G, B between 0..1: If min(R,G,B) < 1/3 or R+G+B < 2, then use the normal dark text weighting. Otherwise, use the light text weighting.

It seems this strategy was introduced to deal with a perceptual glitch like outlined here: http://www.lighterra.com/articles/macosxtextaabug/

mstange:
Fixes #1767.

I can hardly see a difference, but it still seems like the right thing to do. On the [testcase](https://bug1410971.bmoattachments.org/attachment.cgi?id=8921146) I see about 8 distinctive "pops" over the course of the transition from 0 to 255 which are more visible than the new, intended, "pop" that's added by this patch. But those eight pops also happen with Skia.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1911)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -406,6 +406,14 @@ impl FontContext {
         }
     }
 
+    fn should_use_white_on_black(&self, color: ColorU) -> bool {
+        let r = color.r as f32 / 255.0;
+        let g = color.g as f32 / 255.0;
+        let b = color.b as f32 / 255.0;
+        // These thresholds were determined on 10.12 by observing what CG does.
+        r >= 0.333 && g >= 0.333 && b >= 0.333 && r + g + b >= 2.0
+    }
+
     #[allow(dead_code)]
     fn print_glyph_data(&mut self, data: &[u8], width: usize, height: usize) {
         // Rust doesn't have step_by support on stable :(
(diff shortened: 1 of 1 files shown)


webrender/src/platform/macos/font.rs
```diff
@@ -497,13 +505,22 @@ impl FontContext {
...
-        // when we draw the glyphs.
-        let (antialias, smooth, bg_color) = match font.render_mode {
-            FontRenderMode::Subpixel => (true, true, 1.0),
-            FontRenderMode::Alpha => (true, false, 1.0),
-            FontRenderMode::Bitmap => (true, false, 0.0),
-            FontRenderMode::Mono => (false, false, 1.0),
+        // Also note that WR expects text to be white text on black bg, so invert
+        // when we draw the glyphs as black on white.
+        let use_white_on_black = self.should_use_white_on_black(font.color);
+        let (antialias, smooth, text_color, bg_color, bg_alpha, invert) = match font.render_mode {
```
glennw: This could possibly be simplified by deriving some of those values after the match, but it can be done as a follow up, if at all.

PR #1914, opened 2017-10-23
Use premultiplied blending for rectangle primitives.

mstange:
It would be nice to drop unpremultiplied blending completely at some point. This change goes one step on the way there.

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1914)

### DIFF
--- a/webrender/res/ps_rectangle.glsl
+++ b/webrender/res/ps_rectangle.glsl
@@ -49,6 +49,6 @@ void main(void) {
 #ifdef WR_FEATURE_CLIP
     alpha = min(alpha, do_clip());
 #endif
-    oFragColor = vColor * vec4(1.0, 1.0, 1.0, alpha);
+    oFragColor = vColor * alpha;
 }
 #endif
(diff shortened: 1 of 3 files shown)


kvark:
> It would be nice to drop unpremultiplied blending completely at some point.

Not sure if we should completely drop it. Is everything in this world is going to be pre-multiplied? :)

Anyway, the PR looks good. Please launch a try push to see if it breaks anything.

mstange:
> Is everything in this world is going to be pre-multiplied? :)

If I get my way, then yes, everything is going to be premultiplied. :)

mstange:
Try push is at https://treeherder.mozilla.org/#/jobs?repo=try&revision=e5801178a9af020793157655564fcb0801a4d573 . It also contains the changes from #1910, so if it goes orange, it's probably the other PR's fault.

glennw:
I tend to agree with @mstange - I think we should probably just move to pre-multiplied alpha for everything. @kvark Are there are use cases you can think of where it's required / better / faster to *not* use premultiplied?

mstange:
The try push shows that the `fuzzy-if(webrender,1-1,13-13)` annotation on `backface-visibility-2.html` needs to be adjusted to `fuzzy-if(webrender,1-1,16-16)` and is otherwise green.

glennw:
@bors-servo r+

That seems reasonable, thanks! @staktrace Heads up on a slight fuzziness change from this PR.

PR #1918, opened 2017-10-24
Fix inset box shadows with spread radius > border radius.

### LINKED ISSUE #1916: Inset box shadows with spread_radius > border_radius explode
https://bugzilla.mozilla.org/show_bug.cgi?id=1411074

[Testcase](https://bug1411074.bmoattachments.org/attachment.cgi?id=8921264)

Expected vs actual results:
screen shot 2017-10-23 at 7 15 51 pm

This seems to be caused by the `sharpness_scale` adjustment that's done when computing the impact of the spread radius on the corner radius. This adjustment is in the spec, but neither Firefox nor Chrome implement it, so I think we can just drop it again.

```html
<!DOCTYPE html>
<meta charset="utf-8">

<title>Box-shadows with spread radius needs to affect non-zero border-radius</title>

<style>

body {
  margin: 0;
}

div {
  display: inline-block;
  width: 80px;
  height: 80px;
  margin: 20px;
  box-shadow: 0 0 0 10px inset blue;
}
.withRadius {
  border-radius: 20px;
}

</style>

<div style="border-radius: 20px"></div>
<div style="border-radius: 25px"></div>
<div style="border-radius: 10px"></div>
<div style="border-radius: 9px"></div>
<div style="border-radius: 8px"></div>
<div style="border-radius: 7px"></div>
<div style="border-radius: 6px"></div>
<div style="border-radius: 5px"></div>
<div style="border-radius: 4px"></div>
```

glennw:
Fixes #1916.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1918)


webrender/src/frame_builder.rs
```diff
@@ -2449,15 +2440,10 @@ fn adjust_corner_for_box_shadow(
 fn adjust_radius_for_box_shadow(
     border_radius: f32,
     spread_amount: f32,
-    spread_radius: f32,
 ) -> f32 {
-    // Adjust the shadow box radius as per:
-    // https://drafts.csswg.org/css-backgrounds-3/#shadow-shape
```
kvark: why isn't this code following the CSS spec any longer?

glennw: @mstange found that neither Chrome nor Gecko follow that part of the spec. He's opened a bug to remove or reconsider that text in the spec - https://github.com/w3c/csswg-drafts/issues/1900.

mstange:
The reason that this adjustment caused glitches was actually due to a bug in the implementation, or maybe lack of clarity in the spec. For inset box shadows, the adjustment should only apply for negative spread radii.

In any case, I think for now it's a good idea to be consistent with Firefox. If we want to adopt this behavior, we should do it in both Gecko's regular rendering path and in webrender at the same time.

For future reference, this patch would have fixed it:

```diff
diff --git a/webrender/src/frame_builder.rs b/webrender/src/frame_builder.rs
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -1275,17 +1275,16 @@ impl FrameBuilder {
             BoxShadowClipMode::Inset => {
                 -spread_radius
             }
         };

         let shadow_radius = adjust_border_radius_for_box_shadow(
             border_radius,
             spread_amount,
-            spread_radius
         );
         let shadow_rect = prim_info.rect
                                    .translate(box_offset)
                                    .inflate(spread_amount, spread_amount);

         if blur_radius == 0.0 {
             let mut clips = Vec::new();

@@ -2395,68 +2394,59 @@ impl FrameBuilder {
             gpu_cache_updates: Some(gpu_cache_updates),
         }
     }
 }

 fn adjust_border_radius_for_box_shadow(
     radius: BorderRadius,
     spread_amount: f32,
-    spread_radius: f32,
 ) -> BorderRadius {
     BorderRadius {
         top_left: adjust_corner_for_box_shadow(
             radius.top_left,
-            spread_radius,
             spread_amount,
         ),
         top_right: adjust_corner_for_box_shadow(
             radius.top_right,
-            spread_radius,
             spread_amount,
         ),
         bottom_right: adjust_corner_for_box_shadow(
             radius.bottom_right,
-            spread_radius,
             spread_amount,
         ),
         bottom_left: adjust_corner_for_box_shadow(
             radius.bottom_left,
-            spread_radius,
             spread_amount,
         ),
     }
 }

 fn adjust_corner_for_box_shadow(
     corner: LayoutSize,
     spread_amount: f32,
-    spread_radius: f32,
 ) -> LayoutSize {
     LayoutSize::new(
         adjust_radius_for_box_shadow(
             corner.width,
-            spread_radius,
             spread_amount
         ),
         adjust_radius_for_box_shadow(
             corner.height,
-            spread_radius,
             spread_amount
         ),
     )
 }

 fn adjust_radius_for_box_shadow(
     border_radius: f32,
     spread_amount: f32,
-    spread_radius: f32,
 ) -> f32 {
     // Adjust the shadow box radius as per:
     // https://drafts.csswg.org/css-backgrounds-3/#shadow-shape
-    let sharpness_scale = if border_radius < spread_radius {
+    let sharpness_scale = if border_radius < spread_amount {
         let r = border_radius / spread_amount;
         1.0 + (r - 1.0) * (r - 1.0) * (r - 1.0)
     } else {
         1.0
     };
     (border_radius + spread_amount * sharpness_scale).max(0.0)
 }
```

PR #1919, opened 2017-10-24
Switch ps_cache_image to be a brush shader.

glennw:
This starts to expand on how the brush shader works, and
converts one of the simplest primitive shaders over to use it.

The fundamental difference between a primitive and brush shader is:

Primitive shaders have two versions: transform and axis-aligned.

Brush shaders have an opaque and alpha version. In the future,
the opaque brush shader can run on a transformed primitive to
handle the inner part of the primitive which doesn't need AA.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1919)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -4,6 +4,13 @@
 
 #ifdef WR_VERTEX_SHADER
 
+void brush_vs(
+    int prim_address,
+    vec2 local_pos,
+    RectWithSize local_rect,
+    ivec2 user_data
+);
+
 // Whether this brush is being drawn on a Picture
 // task (new) or an alpha batch task (legacy).
 // Can be removed once everything uses pictures.
@@ -102,7 +109,7 @@ void main(void) {
         //           shaders that don't clip in the future,
         //           but it's reasonable to assume that one
         //           implies the other, for now.
-#ifdef WR_FEATURE_ALPHA
+#ifdef WR_FEATURE_ALPHA_PASS
         write_clip(
             vi.screen_pos,
             clip_area
@@ -121,11 +128,14 @@ void main(void) {
 #endif
 
 #ifdef WR_FRAGMENT_SHADER
+
+vec4 brush_fs();
+
 void main(void) {
     // Run the specific brush FS code to output the color.
     vec4 color = brush_fs();
 
-#ifdef WR_FEATURE_ALPHA
+#ifdef WR_FEATURE_ALPHA_PASS
     // Apply the clip mask
     color *= do_clip();
 #endif
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -2,14 +2,21 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-varying vec2 vLocalPos;
-flat varying vec4 vLocalRect;
-
 #ifdef WR_VERTEX_SHADER
+
+// Whether this brush is being drawn on a Picture
+// task (new) or an alpha batch task (legacy).
+// Can be removed once everything uses pictures.
+#define BRUSH_FLAG_USES_PICTURE     (1 << 0)
 
 struct BrushInstance {
     int picture_address;
     int prim_address;
+    int layer_address;
+    int clip_address;
+    int z;
+    int flags;
+    ivec2 user_data;
 };
 
 BrushInstance load_brush() {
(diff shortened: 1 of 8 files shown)


glennw:
r? @kvark 

(Still needs to be run against WPT and Gecko try, but ready for review).

glennw:
All WPT tests are OK on Servo.

Gecko try:
https://hg.mozilla.org/try/rev/92c2487d793bf66b146d02f676d62cc63aa1aabe
https://treeherder.mozilla.org/#/jobs?repo=try&revision=92c2487d793bf66b146d02f676d62cc63aa1aabe

I'm fairly certain the orange tests are either unrelated intermittents, or expected differences given the other PRs we landed today (the pre-multiplied rectangle and clip multiply PRs).

webrender/res/brush.glsl
```diff
@@ -44,36 +56,79 @@ void main(void) {
...
+        local_pos = vi.local_pos;
+
+        // For brush instances in the alpha pass, always write
+        // out clip information.
+        // TODO(gw): It's possible that we might want alpha
+        //           shaders that don't clip in the future,
+        //           but it's reasonable to assume that one
+        //           implies the other, for now.
+#ifdef WR_FEATURE_ALPHA
+        write_clip(
```
kvark: could technically use a `BRUSH_FLAG_XXX` for that

webrender/res/brush_image.glsl
```diff
@@ -60,6 +59,8 @@ void main(void) {
     vec4 color = vColor * texture(sColor1, vec3(uv, vUv.z)).r;
 #endif
 
-    oFragColor = color * do_clip();
+    return color;
 }
 #endif
+
+#include brush
```
kvark: why are we including it at the end rather than at the start?

webrender/src/renderer.rs
```diff
@@ -1296,6 +1374,20 @@ impl Renderer {
                                       options.precache_shaders)
         };
 
+        let brush_image_a8 = try!{
+            BrushShader::new("brush_image",
+                             &mut device,
+                             &["ALPHA"],
```
kvark: hmm, the `alpha` shader already gets this feature inside the `new()`. Is this a collision of features names, or just to reflect the fact that the opaque version of the shader makes no sense here?

webrender/src/renderer.rs
```diff
@@ -2344,6 +2422,34 @@ impl Renderer {
                     .bind(&mut self.device, projection, 0, &mut self.renderer_errors);
                 GPU_TAG_PRIM_BLEND
             }
+            BatchKind::Brush(brush_kind) => {
+                match brush_kind {
+                    BrushBatchKind::Image(target_kind) => {
+                        match target_kind {
+                            RenderTargetKind::Alpha => {
+                                self.brush_image_a8.bind(
```
kvark: nit: could move the actual `bind` call outside of that match

webrender/src/tiling.rs
```diff
@@ -583,13 +584,22 @@ impl AlphaRenderItem {
...
-                        let kind = BatchKind::Transformable(
-                            transform_kind,
-                            TransformBatchKind::CacheImage(picture.target_kind()),
+                        let kind = BatchKind::Brush(
+                            BrushBatchKind::Image(picture.target_kind()),
                         );
                         let key = BatchKey::new(kind, blend_mode, textures);
                         let batch = batch_list.get_suitable_batch(key, item_bounding_rect);
-                        batch.push(base_instance.build(0, cache_task_address.0 as i32, 0));
+                        let instance = BrushInstance::new(
```
kvark: I feel like we'd be better off withing this kind of `new` constructor that's only purpose is to erase the field names. Constructing `BrushInstance {...}` manually here and in other places would read better.

webrender/res/brush.glsl
glennw: I think it makes sense to have separate shaders for this (since it affects the FS too), but we could certainly revisit later on after profiling.

webrender/res/brush_image.glsl
glennw: Being lazy and avoiding a forward declaration of the brush_vs and brush_fs functions :) Fixed.

webrender/src/renderer.rs
glennw: Well spotted - I made the feature names a bit more specific.

webrender/src/renderer.rs
glennw: Done

webrender/src/tiling.rs
glennw: Done

glennw:
@kvark Thanks for the review. All comments addressed and pushed, I think.

kvark:
@glennw awesome, thank you!
@bors-servo r+

Issue #1827, opened 2017-10-06
Divide shadow blur_radius by 2?


Gankro:
Gecko seems to do this internally for its shadow painting code, and our shadows seem a lot blurrier.

glennw:
Does Gecko use a gaussian blur or approximate it with some kind of box filter blur? In the WR gaussian blur shader, I've picked an arbitrary constant for the sigma value, which affects the blurriness amount. It's possible that just tweaking this will be enough to get the results close enough.

mstange:
Gecko uses three box blurs.

Here's the Gecko code + comments that deal with the blur radius for box-shadows:
http://searchfox.org/mozilla-central/rev/b53e29293c9e9a2905f4849f4e3c415e2013f0cb/layout/painting/nsCSSRendering.cpp#4268-4293
http://searchfox.org/mozilla-central/rev/b53e29293c9e9a2905f4849f4e3c415e2013f0cb/gfx/thebes/gfxBlur.cpp#221-226
http://searchfox.org/mozilla-central/rev/b53e29293c9e9a2905f4849f4e3c415e2013f0cb/gfx/2d/Blur.cpp#839-866
http://searchfox.org/mozilla-central/rev/b53e29293c9e9a2905f4849f4e3c415e2013f0cb/gfx/2d/Blur.cpp#293-336

And here's a post that dbaron wrote about the topic: https://dbaron.org/log/20110225-blur-radius

kvark:
I got confused by the subject and initial discussion.

> Divide shadow blur_radius by 2

This is not about changing the blur radius. What Gecko does is dividing the radius by 2 to get the standard deviation, from which the blur kernel operates. The radius stays the same.

glennw:
I believe this is fixed by #1920.

PR #1920, opened 2017-10-24
Tweak blur filter so the result is close to gecko.

mephisto41:
We actually pass the std deviation into cs_blur shader for gaussian
blur. So, correct some naming which causing confusion. Also, according
to the spec, the std deviation of box and text shadow is half of blur
radius.

Here are some gecko try for comparison.
[With this patch](https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/TTHMc9DqR-m6i0qnZo5oCw/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1) 

[Without this patch](https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/JureAaHrT2qty-tYG0BOHg/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1)

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1920)


glennw:
Nice work!

There is a CI failure in the shader validation - you can test this locally by running ```cargo test``` in the root webrender directory.

We should also do a gecko try run with this to see what effect it has on all the existing tests.

mephisto41:
Thanks. I have fixed the shader validation.

gecko try is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9311537999c72f68f92dc4912ce9f576776a5abf&selectedJob=139079239

Only fuzzy changes required I think.

kvark:
Travis seems OK:

Taskcluster is showing one of the reftest failures: https://tools.taskcluster.net/groups/EtuwFR_uQd-l9S7d7QDiow/tasks/EtuwFR_uQd-l9S7d7QDiow/runs/0/logs/public%2Flogs%2Flive.log#L1152

Also, I suppose this fixes #1827

glennw:
@mephisto41  For some reason, the TaskCluster CI often seems to require slightly more fuzziness than the Travis CI. It might just be a case of looking at the TC logs and updating the fuzziness on that one wrench test that is failing.

mephisto41:
>@mephisto41 It does look like quite a lot of unexpected fails on the try run - could you detail the ones that have changed and what the differences are?

Yap, we have R3, R4 and R7 reftests failure. R3 has some box-shadow tests, R4 has both box-shadow and css blur filter tests and R7 has css blur filter tests.

We only need change the fuzzy value for those tests since we change the blur shader.

glennw:
Most of them do appear to be fairly small differences. There's one (maybe in R7?) where one of the reftests has no blur but the reference has a blur.

I *think* this is probably OK to merge, if it makes the blur look closer to Gecko (I haven't checked that myself). But I'd like to get opinion from @staktrace first on whether those failures are reasonable to fuzz (for now).

mephisto41:
Ah, I saw it. it is in R4 (mix-blend-mode-and-filter.html). I'll check it again to see if it is regressed by this patch. Thanks for reminder.

mephisto41:
mix-blend-mode-and-filter.html is regressed by gecko bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1370564). Not this bug, I think this bug is ready to go. I'll create a gecko bug for fixing mix-blend-mode-and-filter.html.

staktrace:
> gecko try is here: https://treeherder.mozilla.org/#/jobs?repo=try&
> revision=9311537999c72f68f92dc4912ce9f576776a5abf&selectedJob=139079239

Yeah it looked like the only non-fuzzy change here was mix-blend-mode-and-filter.html which already got annotated in m-c via bug 1370564. The rest of these changes look fine to me.

glennw:
@staktrace @mephisto41 OK, thanks!

@bors-servo r+

glennw:
@mephisto41  Looks like this has compile errors as a result of some other PRs that landed in the meantime, so will need to be updated, sorry!

mephisto41:
I've updated the changesets.

webrender/src/frame_builder.rs
```diff
@@ -1861,10 +1862,12 @@ impl FrameBuilder {
                         match *filter {
                             FilterOp::Blur(blur_radius) => {
                                 let blur_radius = device_length(blur_radius, device_pixel_ratio);
+                                let blur_std_deviation = blur_radius.0 as f32;
```
glennw: This doesn't seem right - the std deviation shouldn't be the same as the blur radius?

mephisto41: I know this is confusing. But for css blur filter, the blur radius is the std deviation. See the spec https://drafts.fxtf.org/filter-effects/#FilterProperty

"Applies a Gaussian blur to the input image. The passed parameter (the radius) defines the value of the standard deviation to the Gaussian function."

glennw:
Ah, right. Thanks!

@bors-servo r+

PR #1923, opened 2017-10-24
rewrite push_line interface, and make patterned lines match gecko closer

Gankro:
Edit: now ready to review.

See commits for details.

I pushed this up now because I ran into an interesting problem in the gecko test-suite that we need to figure out a solution for. It will also effect partial glyphs.

Basically, gecko likes doing some cute tricks where it draws parts of the text multiple times with different clips to create the illusion of a single glyph/line across multiple runs with possible style changes.

Mostly this works fine, and can be naively lowered with no problem. However this causes a problem for fast shadows, because they inherit the clip from the original element. Therefore fast shadows with an offset will be clipped incorrectly and render broken.

So as a level-1 problem, when we clone the display item to push it in as a fast shadow, we also need to clone its clip and apply the shadow offset. However this creates a level-2 problem: translating the clip isn't always correct to do -- you don't want to translate a clip used for, say, `overflow:hidden`.

Under the current architecture there's a simple heuristic we can apply: clips that are pushed *inside* a shadow should be cloned and translated when pushing fast-shadows. These clips are necessarily just "part of drawing that part of the text", and not "real" clips.

Slow (blurred) shadows are unaffected by all of this (apparently).

TODO:

* [x] Fix up dotted/dashed line periods to match gecko
* [x] Regenerate decoration-suite.png
* [ ] ~~Fix fast-shadow clipping~~
* [ ] ~~Add test for fast-shadow clipping~~

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1923)


Gankro:
For reference, this attempts to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1401653

Gankro:
The clipping thing is also technically a pre-existing orthogonal issue, so maybe we should just land this as-is?

glennw:
If it's a pre-existing issue, then it shouldn't block landing this, I agree.

Gankro:
Fixed shader issues. (all float/int stuff)

Gankro:
@staktrace  heads up for webrender update breakage. 

Most of the update work is done in https://bugzilla.mozilla.org/show_bug.cgi?id=1401653

However it will need an additional fixup atomically with the webrender update that accepts this patch:

```diff
diff --git a/gfx/webrender_bindings/src/bindings.rs b/gfx/webrender_bindings/src/bindings.rs
index 7eab0c1e68ee..35a3b84fe97f 100644
--- a/gfx/webrender_bindings/src/bindings.rs
+++ b/gfx/webrender_bindings/src/bindings.rs
@@ -1487,35 +1487,20 @@ pub extern "C" fn wr_dp_push_line(state: &mut WrState,
                                   clip: &LayoutRect,
                                   is_backface_visible: bool,
                                   bounds: &LayoutRect,
-                                  _wavy_line_thickness: f32,
+                                  wavy_line_thickness: f32,
                                   orientation: LineOrientation,
                                   color: &ColorF,
                                   style: LineStyle) {
     debug_assert!(unsafe { is_in_main_thread() });
 
-    // TODO: remove this once WR is updated to just check the bounds.
-    let (baseline, start, end, width) = match orientation {
-        LineOrientation::Horizontal => (bounds.origin.y,
-                                        bounds.origin.x,
-                                        bounds.origin.x + bounds.size.width,
-                                        bounds.size.height),
-        LineOrientation::Vertical => (bounds.origin.x,
-                                      bounds.origin.y,
-                                      bounds.origin.y + bounds.size.height,
-                                      bounds.size.width),
-    };
-
     let mut prim_info = LayoutPrimitiveInfo::with_clip_rect(*bounds, (*clip).into());
     prim_info.is_backface_visible = is_backface_visible;
     state.frame_builder
          .dl_builder
          .push_line(&prim_info,
-                    baseline,
-                    start,
-                    end,
+                    wavy_line_thickness,
                     orientation,
-                    width,
-                    *color,
+                    color,
                     style);
 
 }
```

glennw:
Reviewed 8 of 10 files at r1, 1 of 1 files at r2, 1 of 2 files at r3.
Review status: 10 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/prim_store.rs, line 209 at r3](https://reviewable.io:443/reviews/servo/webrender/1923#-KxFOzY-UfVdF9Zc0nPz:-KxFOzY-UfVdF9Zc0nQ-:bt12lgl) ([raw file](https://github.com/servo/webrender/blob/678098d76aa7c1e83e209f6127e3fd6557a1b700/webrender/src/prim_store.rs#L209)):*
> ```Rust
>         request.push(self.color);
>         request.push([
>             self.wavy_line_thickness,
> ```

This doesn't seem to match up with fetch_line() in the shader - the line thickness is in the Z component, but it's pushed in the X component here?

---

*[webrender_api/src/display_item.rs, line 152 at r3](https://reviewable.io:443/reviews/servo/webrender/1923#-KxFOaGpmk5mdE1bpbYY:-KxFOaGpmk5mdE1bpbYZ:bwv36vh) ([raw file](https://github.com/servo/webrender/blob/678098d76aa7c1e83e209f6127e3fd6557a1b700/webrender_api/src/display_item.rs#L152)):*
> ```Rust
> pub struct LineDisplayItem {
>     pub orientation: LineOrientation, // toggles whether above values are interpreted as x/y values
>     pub wavy_line_thickness: f32,
> ```

Probably a bit cleaner if this was a field in the LineStyle::Wavy enum?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1923)*

glennw:
@Gankro Looks good in general - two issues above to follow up and then we can merge this.

Gankro:
re: shader component ordering: no if you look at the struct it's the x; the patch only adds .z because now there's 3 elements.

re: Wavy(thickness): that would be a big pain for gecko, as it would make the line style enum non-c-like.

glennw:
You're right - I misread that, thanks!

@bors-servo r+

staktrace:
> @staktrace heads up for webrender update breakage.
> 
> Most of the update work is done in https://bugzilla.mozilla.org/show_bug.cgi?id=1401653
> 
> However it will need an additional fixup atomically with the webrender update that accepts this 
> patch:

Thanks for the patch! However it appears this patch applies *on top* of bug 1401653 which just got backed out of autoland. So unless I do a bunch of manual fixup we're stuck for now.

PR #1924, opened 2017-10-24
Use premultiplied blending for borders.

mstange:
Pretty straightforward. The only interesting parts are the ones where we need to clamp colors after modulation so that the rgb values do not exceed the alpha value (because that would be a superluminous color).

And the reftest change shows a positive difference that we get by using premultiplied colors: The diagonal divide in the border corner between two adjacent border sides now looks better, because when we have anti-aliasing between e.g. red and transparency, halfway values are now just half-transparent red instead of half-transparent *dark* red.

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1924)


mstange:
The try push at https://treeherder.mozilla.org/#/jobs?repo=try&revision=c70e2309103731e155ada91c64780fca09142e11 shows no new failures.

PR #1925, opened 2017-10-24
generate grayscale AA from subpixel AA on Mac

lsalzman:
This makes dilation of the font in the grayscale AA case more closely match the subpixel AA case. This is the default behavior in Gecko.

We don't want to ask OS X for real grayscale AA unless -moz-osx-font-smoothing is used. To get this behavior, a new flag is added in FontInstancePlatformOptions which can be used to toggle font smoothing on or off for the grayscale case.

This is a necessary correctness followup for https://bugzilla.mozilla.org/show_bug.cgi?id=1408481

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1925)


mstange:
I've reviewed this and it looks good to me. Thanks! Looking forward to trying it out in Firefox.

glennw:
This fails some of the wrench reftests, which will need further investigation.

You can run them locally with:

```./headless.py reftest```

lsalzman:
Fixed an issue with the font color not accounting for the different light/dark dilations when using FontRenderMode::Alpha. This fixed the shadow-grey test.

Otherwise, since rendering changed, Markus helped with updating the reftests/text/white-opacity.png reference to match.

All tests pass now.

PR #1929, opened 2017-10-24
Add a ClearRectangle item type.

### LINKED ISSUE #1926: Need a clear-rect display item type
This will be used by Firefox on macOS to make areas with vibrancy transparent. See #1416 for background.

I'm not sure how items of this type should behave with respect to transforms and clips. It would be easiest to just say that transforms and clips on are going to be ignored on these items, but that seems hacky and unexpected from an API user's perspective. And it's totally possible to make it respect transforms and clips.

Proposed implementation:
 - If there is no clip or transform, draw a rectangle with color (0, 0, 0, 0) and blending disabled.
 - If there is a clip or a transform, draw a rectangle subject to those things with color (0, 0, 0, 1) and blending set as:
```rust
  self.gl.blend_func(gl::ZERO, gl::ONE_MINUS_SRC_ALPHA);
  self.gl.blend_equation(gl::FUNC_ADD);
```
(i.e. operator dest-out)

mstange:
Fixes #1926.

I haven't extensively tested this yet. This draws ClearRectangle items as Rectangle primitives with opaque black and operator destination out.
I'm not really sure how to optimize the case where we don't need blending, because the color is set on the primitive before we decide how to blend it. But if there's no blending, we need to set the color to transparent black instead of opaque black.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1929)

### DIFF
--- a/webrender/src/border.rs
+++ b/webrender/src/border.rs
@@ -8,7 +8,7 @@ use clip::ClipSource;
 use ellipse::Ellipse;
 use frame_builder::FrameBuilder;
 use gpu_cache::GpuDataRequest;
-use prim_store::{BorderPrimitiveCpu, PrimitiveContainer, TexelRect};
+use prim_store::{BorderPrimitiveCpu, RectangleContent, PrimitiveContainer, TexelRect};
 use tiling::PrimitiveFlags;
 use util::{lerp, pack_as_float};
 
@@ -385,7 +385,7 @@ impl FrameBuilder {
                 self.add_solid_rectangle(
                     clip_and_scroll,
                     &info,
-                    &border.top.color,
+                    &RectangleContent::Fill(border.top.color),
                     PrimitiveFlags::None,
                 );
             }
@@ -398,7 +398,7 @@ impl FrameBuilder {
                 self.add_solid_rectangle(
                     clip_and_scroll,
                     &info,
-                    &border.left.color,
+                    &RectangleContent::Fill(border.left.color),
                     PrimitiveFlags::None,
                 );
             }
@@ -411,7 +411,7 @@ impl FrameBuilder {
                 self.add_solid_rectangle(
                     clip_and_scroll,
                     &info,
-                    &border.right.color,
+                    &RectangleContent::Fill(border.right.color),
                     PrimitiveFlags::None,
                 );
             }
@@ -424,7 +424,7 @@ impl FrameBuilder {
                 self.add_solid_rectangle(
                     clip_and_scroll,
                     &info,
-                    &border.bottom.color,
+                    &RectangleContent::Fill(border.bottom.color),
                     PrimitiveFlags::None,
                 );
             }
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -17,6 +17,7 @@ use euclid::rect;
 use frame_builder::{FrameBuilder, FrameBuilderConfig};
 use gpu_cache::GpuCache;
 use internal_types::{FastHashMap, FastHashSet, RendererFrame};
+use prim_store::RectangleContent;
 use profiler::{GpuCacheProfileCounters, TextureCacheProfileCounters};
 use resource_cache::{FontInstanceMap,ResourceCache, TiledImageMap};
 use scene::{Scene, StackingContextHelpers, ScenePipeline};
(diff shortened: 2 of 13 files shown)


mstange:
Somehow this change causes my back button's white fill to ignore the rounded clip. Will investigate.

webrender/src/border.rs
```diff
@@ -385,7 +385,7 @@ impl FrameBuilder {
                 self.add_solid_rectangle(
                     clip_and_scroll,
                     &info,
-                    &border.top.color,
+                    &FillOrClear::Fill(border.top.color),
```
kvark: "XxxxOrYyy" pattern sounds a bit awkward to me. Perhaps, we could go with something like `Fill::Color` versus `Fill::Clear` ?

webrender/src/frame_builder.rs
```diff
@@ -616,17 +616,23 @@ impl FrameBuilder {
...
-        let prim = RectanglePrimitive { color: *color };
-
-        // Don't add transparent rectangles to the draw list, but do consider them for hit
-        // testing. This allows specifying invisible hit testing areas.
-        if color.a == 0.0 {
-            self.add_primitive_to_hit_testing_list(info, clip_and_scroll);
-            return;
-        }
+        let prim = match operation {
+            &FillOrClear::Fill(ref color) => {
```
kvark: Could also be `RectangleContent::Fill/Clear`

webrender/src/frame_builder.rs
```diff
@@ -616,17 +616,23 @@ impl FrameBuilder {
...
+                // Don't add transparent rectangles to the draw list, but do consider them for hit
+                // testing. This allows specifying invisible hit testing areas.
+                if color.a == 0.0 {
```
kvark: should probably move this condition up into the matching, maybe even turn it into an if/let i.e.
```rust
if let &FillOrClear::Fill(ColorF{a: 0.0, ..}) = operation {
   (something something return)
}
let prim = RectanglePrimitive { operation: *operation };
```

webrender/src/frame_builder.rs
mstange: I went with this suggestion, `RectangleContent::Fill/Clear`. Thanks!

mstange:
I've addressed the comments. I added a TODO for the missing optimization but didn't address it. I suppose I could make `AlphaRenderItem::add_to_batch` (which knows the chosen blend mode, and whether blending is needed) write a piece of information into the data that's passed to the shader, and then I could modify the rectangle shader to respect that piece of information to switch between opaque black and transparency... but that seems awful.

r? @glennw

webrender/src/frame.rs
```diff
@@ -442,13 +443,27 @@ impl<'a> FlattenContext<'a> {
...
                         clip_and_scroll,
                         &prim_info,
-                        &info.color,
+                        &RectangleContent::Fill(info.color),
+                        PrimitiveFlags::None,
+                    );
+                }
+            }
+            SpecificDisplayItem::ClearRectangle => {
+                if !self.try_to_add_rectangle_splitting_on_clip(
```
glennw: Let's not bother trying to split clear rectangles - we can just directly call ```add_solid_rectangle``` here.

webrender/src/prim_store.rs
```diff
@@ -138,6 +138,7 @@ pub struct PrimitiveMetadata {
     pub cpu_prim_index: SpecificPrimitiveIndex,
     pub gpu_location: GpuCacheHandle,
     pub clip_task_id: Option<RenderTaskId>,
+    pub is_clear: bool,
```
glennw: This should be in the ```RectanglePrimitive``` type since it's not common to all primitives.

mstange: How do I get access to the `RectanglePrimitive` inside `get_blend_mode`?

glennw: Something like:

```let rect = &self.cpu_rectangles[metadata.cpu_prim_index.0];```

should work. The ```TextRun``` primitive in the same function does something similar.

mstange: Thanks!

mstange:
I've addressed the comments.

glennw:
@bors-servo r+

I think the opacity type for a clear rectangle does work - we can easily change it if required, since this is a new interface.

It would be good to add some wrench reftests as a follow up.

mstange:
Rebased. r? @glennw

PR #1930, opened 2017-10-24
Fix render target debugger to show all layers

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1930)


glennw:
The render target debugger wasn't correctly handling targets that have multiple layers in an array texture. This meant that some targets weren't being displayed.

r? anyone

PR #1936, opened 2017-10-25
Add auto crop when generating PNG reftest images.

### LINKED ISSUE #1934: Auto-crop for generating png reference for reftests
It is tedious and error-prone to have to manually crop the reference images generated by wrench. It would be nice if there were some option to auto-crop the output png so that we can skip the extra manual intervention here for the common case.

glennw:
When using the PNG command, if a PNG with the same filename
already exists, read the dimensions from that and crop the
new file to those dimensions.

This means:
 * To create a full size image, or a new image size, delete the
   existing file.
 * Common case - the file size will be cropped to existing size.
 * When creating a new image, crop it manually to set the default
   size.

Fixes #1934.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1936)

### DIFF
--- a/wrench/src/main.rs
+++ b/wrench/src/main.rs
@@ -631,7 +631,7 @@ fn main() {
             gl::UNSIGNED_BYTE,
         );
 
-        save_flipped("screenshot.png", &pixels, size);
+        save_flipped("screenshot.png", pixels, size);
     }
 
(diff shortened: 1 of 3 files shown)


glennw:
r? anyone

PR #1937, opened 2017-10-25
Move box shadow building code to a separate source file.

glennw:
This is prep work for some of the upcoming box shadow optimizations. It makes sense to move this into a separate file since there will be quite a lot of box shadow specific code.

This PR itself has no functional changes. It just moves code around.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1937)


webrender/src/box_shadow.rs
```diff
@@ -0,0 +1,291 @@
+use api::{ColorF, LayerPoint, LayerRect, LayerSize, LayerVector2D};
```
nical: Don't we need a license header at the top of this file?

glennw: Oops, fixed.

PR #1940, opened 2017-10-25
Improve the sticky API and documentation

mrobinson:
This should hopefully fix a lot of issues with the previous version of
the API which made things more complicated than necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1940)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -4,10 +4,11 @@
 
 use api::{ClipId, DeviceIntRect, LayerPixel, LayerPoint, LayerRect, LayerSize};
 use api::{LayerToScrollTransform, LayerToWorldTransform, LayerVector2D, PipelineId};
-use api::{ScrollClamping, ScrollEventPhase, ScrollLocation, ScrollSensitivity, StickyFrameInfo};
+use api::{ScrollClamping, ScrollEventPhase, ScrollLocation, ScrollSensitivity, StickyOffsetBounds};
 use api::WorldPoint;
 use clip::{ClipRegion, ClipSources, ClipSourcesHandle, ClipStore};
 use clip_scroll_tree::{CoordinateSystemId, TransformUpdateState};
+use euclid::SideOffsets2D;
 use geometry::ray_intersects_rect;
 use gpu_cache::GpuCache;
 use render_task::{ClipChain, ClipChainNode, ClipWorkItem};
@@ -54,6 +55,29 @@ impl ClipInfo {
 }
 
 #[derive(Debug)]
+pub struct StickyFrameInfo {
+    pub margins: SideOffsets2D<Option<f32>>,
+    pub vertical_offset_bounds: StickyOffsetBounds,
+    pub horizontal_offset_bounds: StickyOffsetBounds,
+    pub current_offset: LayerVector2D,
+}
+
+impl StickyFrameInfo {
+    pub fn new(
+        margins: SideOffsets2D<Option<f32>>,
+        vertical_offset_bounds: StickyOffsetBounds,
+        horizontal_offset_bounds: StickyOffsetBounds
+    ) -> StickyFrameInfo {
+        StickyFrameInfo {
+            margins,
+            vertical_offset_bounds,
+            horizontal_offset_bounds,
+            current_offset: LayerVector2D::zero(),
+        }
+    }
+}
+
+#[derive(Debug)]
 pub enum NodeType {
     /// A reference frame establishes a new coordinate space in the tree.
     ReferenceFrame(ReferenceFrameInfo),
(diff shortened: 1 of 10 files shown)


mrobinson:
@staktrace Do you see this as an improvement? I'd love to get some feedback on this from you.

webrender_api/src/display_item.rs
```diff
@@ -113,19 +113,50 @@ pub struct ClipDisplayItem {
...
+#[repr(C)]
 #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize)]
-pub struct StickyFrameDisplayItem {
-    pub id: ClipId,
-    pub sticky_frame_info: StickyFrameInfo,
+pub struct StickyOffsetConstraints {
+    /// The minimum offset for this frame, typically a negative value, which specifies how
+    /// far in the negative direction the sticky frame can offset its contents in this
+    /// dimension.
+    pub min_offset: f32,
```
staktrace: I think this would be slightly clearer if we called these fields `lower_bound` and `upper_bound` rather than `min_offset` and `max_offset`. The 'bound' naming also seems more consistent with the 'Constraints' naming on the struct. The comment could end up like this:
```
/// The lower bound for the frame's sticky offset. This is typically a negative value
/// which constrains how far in the negative direction the sticky frame can offset its
/// contents in this dimension.
```

webrender/src/clip_scroll_node.rs
```diff
@@ -436,40 +457,71 @@ impl ClipScrollNode {
...
+        if let Some(margin) = info.margins.top {
+            // First we calculate the potential offset if we are scrolled against the top edge
+            // of the viewport. This would be enough to move us to the edge minus the margin.
+            // Recall that we are working with the position of the viewport (which is not scrolled)
+            // relative the scrolled sticky rect.
+            sticky_offset.y = viewport_rect.min_y() + margin - sticky_rect.min_y();
+
+            // If the offset is less than zero, that means it is already below the top of the
+            // viewport rectangle.
+            sticky_offset.y = sticky_offset.y.max(0.0);
```
staktrace: This code is much clearer than it was before, thanks! It makes sense to me, although as I was reading it it occurred to me it sometimes computes a thing for sticky_offset, but then cancels the thing it just computed in the max call. It's effectively equivalent to this:

```
// The desired position of the item assuming we are scrolled against the top edge of the viewport
let desired_position = viewport_rect.min_y() + margin;
// If sticky rect is closer to the edge than desired, adjust the offset to put it at the desired location
if sticky_rect.min_y() < desired_position {
    sticky_offset.y = desired_position - sticky_rect.min_y();
}
```
I'm not sure if they compile down to the same thing but avoiding the setting and re-setting of sticky_offset.y might be good.

staktrace:
My brain is quite tired so I could be wrong but I think this change will still not address [bug 1411627](https://bugzilla.mozilla.org/show_bug.cgi?id=1411627). The reason is the same as before: the code assumes that the sticky item as provided in the display list is already in it's "normal" position with no sticky adjustment applied. In gecko this is not always the case. So when `sticky_rect.min_y() > desired_position` that can either mean that (a) the sticky_rect in normal position is still sufficiently far away from the edge of the viewport or (b) the sticky_rect with some sticky adjustment applied actually needs to reduce that applied sticky adjustment. And in the (b) case we actually do want to carry a negative sticky_offset.y.

mrobinson:
@staktrace That's correct. My plan was to fix the current API and then update it to support the Gecko situation.

webrender_api/src/display_item.rs
mrobinson: The only issue I have the with `bounds` terminology is that it isn't totally clear that this is dealing with offsets and not absolute or relative positions. Perhaps it makes sense to change the name of the data structure to something like StickyOffsetRange?

webrender/src/clip_scroll_node.rs
mrobinson: Sure. This seems like a reasonable change. I think I was just trying to avoid another level of indentation here.

webrender_api/src/display_item.rs
mrobinson: I ended up going with `StickyOffsetBounds` here with `min` and `max`. I didn't want to use upper and lower, because their meanings are reversed visually (the lower bound specifies how far upward the sticky item can move). I wanted to avoid any possible confusion there.

staktrace: Ok, fair enough. I hadn't thought about the visual interpretation of upper and lower - you're right that it could get confusing :)

staktrace:
Sounds good, I'm fine with adding support for the Gecko situation in a different PR.

glennw:
Is this ready for review / merge?

mrobinson:
@glennw Yes. I believe it is.

glennw:
@mrobinson Great, I'll look at this today. Is there a corresponding Servo patch available? Or is it fairly easy to see what needs to change on the Servo side?

mrobinson:
@glennw I think on the servo side the changes just involve putting the same bits of data into the new data structure. I can write up a patch tomorrow if you like!

glennw:
@mrobinson OK, sounds good. I'll try to do the Servo update today if this gets merged, otherwise one of us can do it tomorrow. Thanks!

glennw:
Reviewed 1 of 10 files at r1, 9 of 9 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1940)*

PR #1945, opened 2017-10-25
remember GlyphFormat in glyph cache and pass to text run shader

lsalzman:
When glyphs are rasterized, this makes sure the font backends output the actual format of the glyph. This is then passed down through fetch_glyphs into the shader, which can use this information to properly interpret things like color.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1945)

### DIFF
--- a/webrender/res/ps_text_run.glsl
+++ b/webrender/res/ps_text_run.glsl
@@ -13,6 +13,12 @@ varying vec3 vLocalPos;
 #endif
 
 #ifdef WR_VERTEX_SHADER
+
+#define MODE_ALPHA          0
+#define MODE_SUBPX_PASS0    1
+#define MODE_SUBPX_PASS1    2
+#define MODE_COLOR_BITMAP   3
+
 void main(void) {
     Primitive prim = load_primitive();
     TextRun text = fetch_text_run(prim.specific_prim_address);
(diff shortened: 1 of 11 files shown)


lsalzman:
Fixed undeclared uMode in vertex shader.

lsalzman:
Vertex shader had a leftover vColor assignment that was messing results up. Removed it.

lsalzman:
Fixed the debugger build issue... Once more with feeling.

glennw:

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1945)*

glennw:
This looks great. I'm just going to do a gecko try run with it, and if that passes, we can merge it.

glennw:
Nice performance win in the text shader! This should also help me when starting to unify the shaders as brush shaders.

glennw:
Gecko try run here:

https://hg.mozilla.org/try/rev/7692b8b8281e38035c40ba9a1e35e589cada75c9
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7692b8b8281e38035c40ba9a1e35e589cada75c9&selectedJob=139693056

There is a lot of orange - but they are all either expected from other PRs, or caused by me disabling text decorations in this local run, due to the API differences.

staktrace:
@lsalzman This seems to be causing shader compilation failures on windows. See https://bugzilla.mozilla.org/show_bug.cgi?id=1410893#c14

glennw:
@staktrace Opened https://github.com/servo/webrender/issues/1950 regarding the windows shader errors we don't seem to be catching in the shader validation check.

lsalzman:
Dzmitry's pull request is a reasonable fix for the shader error: https://github.com/servo/webrender/pull/1952

Another reasonable approach might be to set an explicit precision for uMode in both definitions, though depending on the minimum GLSL spec, that could be dicey.

PR #1947, opened 2017-10-26
Tidy up picture/brush code in preparation for box shadow optimizations.

glennw:
There was some confusing code in the existing text shadow implementation
that was making it difficult to optimize the box shadows to use sub-parts
of a picture image. Specifically, in the cs_text_run and ps_line shader,
we were fetching the owning Picture primitive. Instead, now the following
conventions are observed:

* When drawing *on* a picture, fetch the picture render task and the
  specific primitive only. This should have all information to draw
  on to a picture task.
* When *compositing* a picture into the scene, fetch the picture
  render task for sampling information (e.g. UV rect, color mask),
  and the picture primitive for geometry information.

In the second case, we don't currently need to fetch the picture
primitive, since it's always just the entire primitive local rect.
However, with the upcoming box shadow optimizations, this will
typically be sub-parts or a 9-patch of the picture image.

Importantly, this also decouples the geometry of the picture to be drawn
from the geometry of the picture render task. This will allow box
shadow optimizations such as rendering just a corner of the box
shadow when uniform radii are present, and mirroring the blur task
using brush segments.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1947)


glennw:
r? @kvark 

There's a couple of hacks here (e.g. the ```#ifdef``` on the ```Primitive``` struct) but it's mostly a lot tidier, I think, and will greatly simplify the work involved in doing the planned box shadow optimizations.

kvark:
Looks great for the most part! Just one concern to answer/address.

---

---

*[webrender/src/picture.rs, line 128 at r1](https://reviewable.io:443/reviews/servo/webrender/1947#-KxOKgj31v-E-E1ojJBp:-KxOKgj31v-E-E1ojJBq:b-mh6x30) ([raw file](https://github.com/servo/webrender/blob/188fb02cd3287b6c36211e1838268aedb93535a9/webrender/src/picture.rs#L128)):*
> ```Rust
>         match self.kind {
>             PictureKind::TextShadow { ref shadow } => {
>                 let blur_offset = shadow.blur_radius * BLUR_SAMPLE_SCALE;
> ```

the old code used to inflate by `blur_radius` as opposed to `blur_offset` here. Is this a logic fix? (I thought the PR is only about refactoring)

---

*[wrench/reftests/text/decorations-suite.png, line 0 at r1](https://reviewable.io:443/reviews/servo/webrender/1947#-KxOLCVancVBG2mgY4ID:-KxOLCVancVBG2mgY4IE:bywkafj) ([raw file](https://github.com/servo/webrender/blob/188fb02cd3287b6c36211e1838268aedb93535a9/wrench/reftests/text/decorations-suite.png#L0)):*
why is this reftest changed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1947#-:-KxOLG4sVyplXTJI6yh0:bruz0q7)*

glennw:
@kvark Yep, it's a logic fix - I hadn't realized I had fixed that inadvertently. The reftest update is because that logic fix allows a slightly wider (correct) range for the blur on a text shadow to occur over.

kvark:
Thanks!
@bors-servo r+

PR #1951, opened 2017-10-26
Make UnsafeVecWriter faster

jrmuizel:
I was seeing write() not being inlined as well as the default write_all() implementation. Neither of those are ever good things for us to have.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1951)


Gankro:
@bors-servo r+

seems reasonable.

PR #1953, opened 2017-10-26
When rasterizing FontRenderMode::Alpha text, use a byte order that CG supports subpixel AA for.

mstange:
When we rasterize fonts on macOS with `FontRenderMode::Alpha` and with `smooth == true`, we want to get subpixel AA text from CG and convert that into grayscale text. However, we were creating a CGContext that does not support drawing subpixel AA text, so instead of recovering grayscale from subpixel we were recovering grayscale from grayscale. Using `kCGBitmapByteOrder32Little` fixes that, and now `FontRenderMode::Alpha` text finally looks like `FontRenderMode::Subpixel` text instead of thinned out.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1953)


glennw:
@mstange No rush, but it'd be great to have some wrench reftests that can test each of the mac rendering modes, to make sure we don't regress in the future.

lsalzman:
With this patch, we now have the opposite problem, that the context flags are wrong when font_smoothing is off. We need to be checking the value of font_smoothing (could just move the definition of use_font_smoothing up there).

When use_font_smoothing is true, we need the subpixel context flags. Otherwise, when it is false, we need to use the old ones.

mstange:
> Otherwise, when it is false, we need to use the old ones.

What, no, why? When it's false, we still call `cg_context.set_should_smooth_fonts(false)`, so CG renders thin grayscale fonts. It can do that with any byte ordering.

mstange:
I've tested this patch in Firefox and confirmed that it produces the expected results both with and without `-moz-osx-font-smoothing: grayscale` (with the patch in [bug 1408481](https://bugzilla.mozilla.org/show_bug.cgi?id=1408481) applied) and with text inside opacity.

mstange:
Lee and I had a conversation about this and came to the conclusion that the patch as-is would work, but that some of the comments need to be adjusted, and that it would be a good idea to make Mono use the new flags as well, because we access the buffer in a little-endian order further down, so we really always want to specify little-endian.

mstange:
I've made the following changes to the patch:

 - Switched over the Mono font render mode
 - Rewrote many comments
 - Slightly changed the match statement in the middle to make it clearer that the `Subpixel` and the `Alpha + smoothing` modes are treated the same way

I'm now looking into the reftest failures.

mstange:
I fixed up the reftest failures:
 - `white-opacity.yaml` has text inside opacity and hits the alpha mode. It now uses the grayscale-from-subpixel mask instead of the thin-grayscale mask, and `white-opacity.png` needed to be generated.
 - `shadow-grey.yaml` needed a slight increase of the existing fuzz. It uses the alpha mode for the text-shadow. The new text is thicker, has more pixels in the anti-aliasing, and has more pixels that encounter the apparent rounding error.
 - `ahem.yaml` hits the alpha mode because it has `options(disable-subpixel)` in the reftest.list. However, it seems that CG adds anti-aliasing on ahem glyphs when drawing with subpixel AA but not when drawn with font_smoothing == false. Now that we're asking CG to draw the glyph with subpixel AA and then creating the grayscale mask from that, we also have anti-aliasing in the Alpha mode. I've changed the reftest.list annotation to `options(disable-aa)`.

lsalzman:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1953#-:-KxaXAgUOKxx1LvGjoQE:bnfp4nl)*

PR #1954, opened 2017-10-27
First round of optimizations for box shadows.

### LINKED ISSUE #1571: High gpu time on http://github.com when not signed in
STR:
- Turn on layers-free, webrendest, and profiling in Gecko
- Scroll down to the large orange blue and purple circles containing numbers.
- The GPU time goes up to 53ms

glennw:
This applies two optimizations to outset box shadows:

* If border radii are uniform, render only the top left corner
  of the box shadow, and mirror that across the primitive.
* If border radii are non-uniform, render a minimal mask
  rect and stretch that across the primitive.

There are several remaining box shadow / clip related optimizations
to come, but this provides some large immediate wins on large
outset box shadows.

Fixes #1571.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1954)


glennw:
@jrmuizel This provides a large performance win on https://bugzilla.mozilla.org/show_bug.cgi?id=1408612.

In the profile bars below, the pink segments are blurs for box shadows. They are an almost negligible amount of time with this optimization.

profile


On my HD4600, this drops the GPU time on bugzilla from ~20ms to ~8ms. The performance with this fix is reasonable now, but there's three major issues outstanding:

* No z-buffer wins (known issue being worked on in WR, due to Gecko not using local clips).
* Way too many clip masks being drawn (still nearly 3 2048x2048 A8 targets being filled per frame - the purple bars). Need to investigate this more - probably some issues in both Gecko and WR here.
* High number of draw calls / bad batching. Needs a bit of investigation - it's worse than I'd expect. Will be resolved anyway as we unify more shaders to brush types.

Follow ups:
 * Optimize inset shadows.
 * Specifiy render task cache key for blurs - this will save doing the same blur on different primitives.

glennw:
Pending gecko try:

https://hg.mozilla.org/try/rev/92d0786b9e62d6b6e5ddbb741b1628e17057aaa5
https://treeherder.mozilla.org/#/jobs?repo=try&revision=92d0786b9e62d6b6e5ddbb741b1628e17057aaa5

glennw:
Servo WPT sanity tests pass, I need to do a full run through.

kvark:
Amazing work :+1: ! I just got a few concerns to address.

---

Reviewed 15 of 15 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[webrender/res/brush_image.glsl, line 90 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTUib6PfNtrC7EHcdy:-KxTUib7AGp9gnsTdHRN:bju07bl) ([raw file](https://github.com/servo/webrender/blob/2c9b4145ff594d4fe4652d7a0dff961093b19fc6/webrender/res/brush_image.glsl#L90)):*
> ```GLSL
>         }
>         case BRUSH_IMAGE_NINEPATCH: {
>             uv = clamp(vUv.xy, vec2(0.0), vParams.xy);
> ```

this seems incorrect, given that `vParams.xy == vec2(0.5)` at this point

---

*[webrender/res/brush_image.glsl, line 98 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTV8wlnPCG2_dnS2dF:-KxTV8wmmoRnqpl-Gxgk:b-bi62s) ([raw file](https://github.com/servo/webrender/blob/2c9b4145ff594d4fe4652d7a0dff961093b19fc6/webrender/res/brush_image.glsl#L98)):*
> ```GLSL
>             // Mirror and stretch the box shadow corner over the entire
>             // primitives.
>             uv = vParams.xy - abs(vUv.xy - vParams.xy);
> ```

given that `vParams.xy` is half-size of the local rect, this would produce `uv` between the local rect origin and it's half-size. It doesn't appear to be desired?

---

*[webrender/res/brush_mask_rounded_rect.glsl, line 72 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTVkxvssS6KKni_v1n:-KxTVkxvssS6KKni_v1o:bnl38pk) ([raw file](https://github.com/servo/webrender/blob/2c9b4145ff594d4fe4652d7a0dff961093b19fc6/webrender/res/brush_mask_rounded_rect.glsl#L72)):*
> ```GLSL
>     // be made much faster.
> 
>     // NOTE: The AA range must be computed outside the if statement,
> ```

I think we should rather move the AA stuff out of `rounded_rect` (than call `rounded_rect` all the time), so that we can avoid some work here if we are outside of the clip region.

---

*[webrender/src/box_shadow.rs, line 20 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTWEmT1-rkTRmkKUJZ:-KxTWEmT1-rkTRmkKUJ_:bdtq8jk) ([raw file](https://github.com/servo/webrender/blob/2c9b4145ff594d4fe4652d7a0dff961093b19fc6/webrender/src/box_shadow.rs#L20)):*
> ```Rust
> // mask. This ensures that we get a few pixels past the corner that can be
> // blurred without being affected by the border radius.
> pub const MASK_CORNER_PADDING: f32 = 4.0;
> ```

hmm, should this depend on the border radius then?

---

*[webrender/src/picture.rs, line 41 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTWyVX1v58dlrs-5Am:-KxTWyVX1v58dlrs-5An:bx82cjs) ([raw file](https://github.com/servo/webrender/blob/2c9b4145ff594d4fe4652d7a0dff961093b19fc6/webrender/src/picture.rs#L41)):*
> ```Rust
>         blur_regions: Vec<LayerRect>,
>         clip_mode: BoxShadowClipMode,
>         has_uniform_radii: bool,
> ```

would be nice to have it as an `enum` for better calling semantics

---

*[webrender/src/picture.rs, line 142 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTXAkPtwVgOXAEwuzs:-KxTXAkPtwVgOXAEwuzt:bt0akid) ([raw file](https://github.com/servo/webrender/blob/2c9b4145ff594d4fe4652d7a0dff961093b19fc6/webrender/src/picture.rs#L142)):*
> ```Rust
>             PictureKind::BoxShadow { blur_radius, clip_mode, has_uniform_radii, .. } => {
>                 // We need to inflate the content rect if outset.
>                 if clip_mode == BoxShadowClipMode::Outset {
> ```

nit: `match` would be more clear

---

*[webrender/src/prim_store.rs, line 190 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTXtUEAuInX-nXcsOv:-KxTXtUFL45oAosHdrrI:b1ww9m4) ([raw file](https://github.com/servo/webrender/blob/2c9b4145ff594d4fe4652d7a0dff961093b19fc6/webrender/src/prim_store.rs#L190)):*
> ```Rust
>     fn write_gpu_blocks(&self, mut request: GpuDataRequest) {
>         match self.kind {
>             BrushKind::Mask { clip_mode, ref kind } => {
> ```

nit: given the paths are so different, might as well match `BrushMaskKind` in place here, i.e. `BrushKind::Mask { clip_mode, kind: BrushMaskKind::Corner(radius) } => {...}`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1954#-:-KxTYOgU5voSARqM9peC:byxboz0)*

glennw:
Review status: 9 of 17 files reviewed at latest revision, 7 unresolved discussions.

---

*[webrender/res/brush_image.glsl, line 90 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTUib6PfNtrC7EHcdy:-Kxe9HKjCWIn5tOwtJZO:bikfxxh) ([raw file](https://github.com/servo/webrender/blob/0c3bb11a7ab92ba10952156132821e269d07686e/webrender/res/brush_image.glsl#L90)):*

this seems incorrect, given that `vParams.xy == vec2(0.5)` at this point
</blockquote></details>

The UV coords passed are divided by the size of the patch, so 0.5 is the UV when we reach half of the source patch.

---

*[webrender/res/brush_image.glsl, line 98 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTV8wlnPCG2_dnS2dF:-KxeA8tWIO8gIrC_X_-2:b-87auge) ([raw file](https://github.com/servo/webrender/blob/0c3bb11a7ab92ba10952156132821e269d07686e/webrender/res/brush_image.glsl#L98)):*

given that `vParams.xy` is half-size of the local rect, this would produce `uv` between the local rect origin and it's half-size. It doesn't appear to be desired?
</blockquote></details>

Same as above - the UV space is local_rect_size / patch_size.

---

*[webrender/res/brush_mask_rounded_rect.glsl, line 72 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTVkxvssS6KKni_v1n:-KxeAGzIdfwQEWF9zCX6:b-896fix) ([raw file](https://github.com/servo/webrender/blob/0c3bb11a7ab92ba10952156132821e269d07686e/webrender/res/brush_mask_rounded_rect.glsl#L72)):*

I think we should rather move the AA stuff out of `rounded_rect` (than call `rounded_rect` all the time), so that we can avoid some work here if we are outside of the clip region.
</blockquote></details>

Done.

---

*[webrender/src/box_shadow.rs, line 20 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTWEmT1-rkTRmkKUJZ:-KxeAMsdgLxtt4iLM5rQ:b-q8ol7s) ([raw file](https://github.com/servo/webrender/blob/0c3bb11a7ab92ba10952156132821e269d07686e/webrender/src/box_shadow.rs#L20)):*

hmm, should this depend on the border radius then?
</blockquote></details>

It's added as padding in addition to the border radii sizes below.

---

*[webrender/src/picture.rs, line 41 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTWyVX1v58dlrs-5Am:-KxeARv3YI2_rT7Rc9Rn:b-896fix) ([raw file](https://github.com/servo/webrender/blob/0c3bb11a7ab92ba10952156132821e269d07686e/webrender/src/picture.rs#L41)):*

would be nice to have it as an `enum` for better calling semantics
</blockquote></details>

Done.

---

*[webrender/src/picture.rs, line 142 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTXAkPtwVgOXAEwuzs:-KxeASYNwFUXXIXYTy78:b-896fix) ([raw file](https://github.com/servo/webrender/blob/0c3bb11a7ab92ba10952156132821e269d07686e/webrender/src/picture.rs#L142)):*

nit: `match` would be more clear
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 190 at r1](https://reviewable.io:443/reviews/servo/webrender/1954#-KxTXtUEAuInX-nXcsOv:-KxeATLTx9lKFJTlwdsH:b-896fix) ([raw file](https://github.com/servo/webrender/blob/0c3bb11a7ab92ba10952156132821e269d07686e/webrender/src/prim_store.rs#L190)):*

nit: given the paths are so different, might as well match `BrushMaskKind` in place here, i.e. `BrushKind::Mask { clip_mode, kind: BrushMaskKind::Corner(radius) } => {...}`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1954)*

glennw:
@kvark Thanks! All comments addressed, I think.

This is ready for re-review, but I still need to do a gecko try run to see what the differences are and if they are within expected fuzziness tolerance.

glennw:
Looks like there are a few failures in a gecko try run - will investigate those today.

glennw:
This is the try run:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=c5f8f1ac6a590d976ba59e3ca8e5be9315509528&selectedJob=140569327

It has 1 new PASS and 3 new FAILs.

Unfortunately those fails seem to pass locally on my machine. Still investigating...

glennw:
There is a 3rd commit which fixes / works around some of the Gecko try failures above. Specifically:

* Handle the case of an inset shadow rect becoming invalid and skip trying to draw it.
* Add a temporary workaround for cases where the inset box shadow is computed to be a fractional number of device pixels. We'll need a proper fix for this as a follow up.

With those changes, we have:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=ecc653837134cedc319859af3de8f01040d102dd&selectedJob=140625740

The ```a11y``` and ```mda2``` failures are unrelated.

Two new PASS in ```R3``` and ```R4```:

```boxshadow-large-border-radius.html```
```box-decoration-break-with-inset-box-shadow-1.html```

One new FAIL in ```R3```:

```boxshadow-skiprect.html``` - In this test, we clip one extra vertical line of pixels than it should be. This appears to be a bug in the clipping code, so technically unrelated to the box shadow code. I'm also unable to reproduce this locally - I get pixel-perfect results both running under OSMesa or on Intel GPU.

With @staktrace 's blessing, I think we should add some fuzziness to ```boxshadow-skiprect.html``` and get these changes re-reviewed and merged - they provide a really significant performance improvement on a lot of sites.

r? @kvark

glennw:
(or @nical for the last 2 commits, if @kvark is on PTO this week?)

PR #1956, opened 2017-10-27
Remove isolation conflict log.

### LINKED ISSUE #1955: Remove isolation conflict log in push_stacking_context
There are some cases that a stacking context can contain multiple child stacking contexts with blend mode:
  In the test in[1], we have two background-images applied to blend mode in the same item.

If my understanding is correct, this error log in push_stacking_context assume a stacking context can only have one child stacking context which has blend mode.

Perhaps we should remove this assumption?

[1] https://searchfox.org/mozilla-central/source/layout/reftests/css-blending/background-blending-gradient-gradient.html
[2] https://github.com/servo/webrender/blob/4b8493d6bdc64d2d83202ac15b06b0d4b14c6e76/webrender/src/frame_builder.rs#L369

GuanWen-Chen:
This fixes #1955, there are some cases that a stacking context can contain multiple child stacking contexts with blend mode.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1956)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -365,13 +365,6 @@ impl FrameBuilder {
                 // must be drawn with a transparent background, unless the parent stacking context
                 // is the root of the page
                 let isolation = &mut self.stacking_context_store[parent_index.0].isolation;
-                if *isolation != ContextIsolation::None {
-                    error!(
-                        "Isolation conflict detected on {:?}: {:?}",
-                        parent_index,
-                        *isolation
-                    );
-                }
                 *isolation = ContextIsolation::Full;
             }
         }


webrender/src/frame_builder.rs
```diff
@@ -365,13 +365,6 @@ impl FrameBuilder {
                 // must be drawn with a transparent background, unless the parent stacking context
                 // is the root of the page
                 let isolation = &mut self.stacking_context_store[parent_index.0].isolation;
-                if *isolation != ContextIsolation::None {
```
kvark: what if it's `ContextIsolation::Items`? overwriting it with `Full` would screw up the follow up logic

glennw:
Yep, we need to fix the underlying bug / issue that's causing this warning. I'll close this for now, since I think it's quite an involved change to fix this problem. Please feel free to re-open otherwise though.

webrender/src/frame_builder.rs
GuanWen-Chen: Thank you for the feedback.

Maybe we could relax the condition here instead of removing it and make the stacking context able to contain multiple child stacking context with blend mode?

According to the code, we only use ContextIsolation::Items for elements which contain preserve-3d transform-style.
And in gecko, an additional stacking context will be created to contain stacking contexts with blend mode.
With this mechanism, we can prevent isolation conflict in a stacking context.
I am not sure if Servo has this mechanism, but I think it might be reasonable to do that.

I will push another pr for this.

glennw: There is a PR nearly ready to merge which should solve most (or all?) of the isolation issues - https://github.com/servo/webrender/pull/2031. Ideally, we'd get that to land, and then perhaps you could investigate if there are any remaining isolation failures and fix them on top of that patch (since it changes the way we handle stacking contexts internally very significantly)?

GuanWen-Chen: Oh, thank you for the feedback! Then I will cancel the pr and wait for the issue solved.

glennw: Sounds good, thank you!

PR #1958, opened 2017-10-28
ensure font colors are quantized to preblend table precision

lsalzman:
When building the preblend tables, the input color that we preblend against is actually quantized to only 3 bits of precision per color component, rather than 8. Skia ensures that the color it is using to key the glyph cache is similarly quantized, so as not to needly generate more glyph entries than needed.

I have added the necessary functionality to quantize colors to gamma-lut-rs in version 0.2.2.

This just ensures we actually use it now. For subpixel we just need to quantize the color. For alpha, we need to first convert to luminance, then quantize. On FreeType, we don't use preblends, so this is unnecessary there. This matches what Skia is doing in all these cases now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1958)


lsalzman:
Funny story: reftests/text/shadow-ordering was failing because grayscale preblending is broken before the changes I've made in this PR. The G color channel is used to pick the preblend table for grayscale, assuming the input color is actually luminance/grayscale. But we never actually converted the color to luminance for FontRenderMode::Alpha.

So certain colors like red, blue, black, etc. have no G channel value, meaning they disable preblending effectively. This test managed to use those colors. The ref used transparent color (0, 0, 0, 0) instead of those colors. So it suddenly broke when my changes here in prepare_font actually set up font.color to be a proper luminance value, since the preblend for those colors and transparency no longer matched.


After conversing with Gankro, the right/simple thing to do for the shadow-ordering test is to disable AA on it, since that's not what the test actually tests for anyway, and altering the colors to get matching preblends would be tricky and silly.

Should pass now...

glennw:
@lsalzman has a cleaner version of this patch, but we'll let this merge since it's in the queue anyway. We can do the cleaner version as a follow up.

Issue #1817, opened 2017-10-05
https://www.w3.org/conf/2013sf/ uses too much GPU time


mstange:
With the integrated GPU on my Macbook Pro, https://www.w3.org/conf/2013sf/ uses about 12ms of GPU time. I exported a frame as a YAML + resources zip: http://tests.themasta.com/w3.zip

Here's a screenshot of the webrender profiler on `cargo run --release -- -p 1 show w3/w3.yaml`:

<img width="1552" alt="screen shot 2017-10-05 at 4 33 17 pm" src="https://user-images.githubusercontent.com/961291/31251134-f7185392-a9ea-11e7-9488-21390ee5dfa6.png">

glennw:
There are a lot of clips and blurs occurring in that profile - we'll need to investigate why (or why they are slow).

This is also (probably) a good example of how in the Gecko integration we're not currently getting any z-buffer benefits - most pixels on here are opaque, so we'll probably benefit a lot from early z-reject once this is fixed in WR.

mstange:
Adjusting the text shadow blur radius to match Gecko might also improve this slightly.

mstange:
The inner rect workaround from #1820 regressed this testcase by about 50%: The GPU time on my machine with the integrated GPU went from 10.91ms to 16.11ms.

kvark:
Thanks @mstange . Time to revisit my proposal of #1828 to address this properly?

mstange:
The shadow changes increased this to 340ms.

mstange:
With the blur optimizations from #1896, it goes down to 60ms.

glennw:
Wow! I'll take a look at this today, and write up any findings here.

metajack:
cc @metajack

glennw:
I logged out all the blur render tasks that are added for this test case:

```
new_blur r=24 t=Alpha s=1756×13711
new_blur r=4 t=Color s=810×74
new_blur r=4 t=Color s=865×74
new_blur r=4 t=Color s=602×74
new_blur r=4 t=Color s=280×74
new_blur r=4 t=Color s=224×74
new_blur r=4 t=Color s=198×74
new_blur r=4 t=Color s=207×74
new_blur r=4 t=Color s=255×74
new_blur r=4 t=Color s=216×74
new_blur r=4 t=Color s=219×74
new_blur r=4 t=Color s=222×74
new_blur r=4 t=Color s=255×74
new_blur r=4 t=Color s=338×74
[...]

PR #1960, opened 2017-10-30
Use primitive clip masks only, instead of sharing clip chain masks.

glennw:
We're not getting any noticeable benefit from trying to share clip
masks. In fact, we are often seeing very significant slowdowns
since we over-estimate the size needed for a shared clip mask.

This change drops the GPU time in #1817 from 40ms to 7ms on
my machine - this is an extreme case, but it's faster in every
site that I tested on locally.

This is also a good simplification for upcoming changes, where
we will be more aggressively removing clip masks when they
are redundant, and drawing them in small segments where possible.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1960)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1191,7 +1191,7 @@ impl PrimitiveStore {
         // Try to create a mask if we may need to.
         let prim_clips = clip_store.get(&metadata.clip_sources);
         let is_axis_aligned = prim_context.packed_layer.transform.preserves_2d_axis_alignment();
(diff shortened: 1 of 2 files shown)


glennw:
Ran this through the Servo WPT suite with no problems, and Gecko try looks good too:

https://hg.mozilla.org/try/rev/f25d829728e34eadb9e76c3b57fd28ab21592dab
https://treeherder.mozilla.org/#/jobs?repo=try&revision=f25d829728e34eadb9e76c3b57fd28ab21592dab

PR #1961, opened 2017-10-30
 Downscaling the render target when having large std deviation. 

mephisto41:
Some performance numbers.

Without downscaling
```
      "name": "benchmarks/large-blur-radius.yaml",
      "backend_time_ns": 54474,
      "composite_time_ns": 373885,
      "paint_time_ns": 79356138,
```

With downscaling
```
      "name": "benchmarks/large-blur-radius.yaml",
      "backend_time_ns": 34355,
      "composite_time_ns": 150301,
      "paint_time_ns": 527546,
```

We have 150x faster for large blur in paint time.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1961)


glennw:
Reviewed 7 of 14 files at r1.
Review status: 7 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1961)*

glennw:
@mephisto41 Amazing result! In general the patch looks good to me (although it will need a rebase).

But it's not clear to me why all those reftest images changed? Most of them don't involve a blur so shouldn't be affected by this change? Were you running the reftests in the software rendering mode (with ```./headless.py reftest```) to ensure consistent results between machines?

We will also need to do a gecko try run before merging this, since it could affect a lot of gecko tests.

mephisto41:
@glennw, ```degenerate-curve.png``` and ```linear-aligned-border-radius.png``` are affected by switching linear filter for CacheA8. And rest of tests are boxshadow and css blur filter which are affected by this change.

I'll submit a gecko try later.

mephisto41:
gecko try:  https://treeherder.mozilla.org/#/jobs?repo=try&revision=d54ca6b1a9a46a0f2ff82652f30f9511bae42229

glennw:
@mephisto41 Ah, of course - that makes sense, thanks!

mephisto41:
Looks like boxshadow position is wrong in boxshadow-rotated.html. I'll debugging it. And after rebase, the boxshadow results are weird. Also checking.

mephisto41:
Latest gecko try. https://treeherder.mozilla.org/#/jobs?repo=try&revision=e7f53805fa83730513bb1e778025766ef85dbdcc&selectedJob=141865448

All failures only need fuzzy value change. So it looks good.

glennw:
(especially since this gives a very significant performance win on large blurs).

glennw:
@mephisto41 As a follow up, could you add a reftest that tests both small blurs (with no downscale applied), and also large blurs where the downscale takes effect?

mephisto41:
> As a follow up, could you add a reftest that tests both small blurs (with no downscale applied), and also large blurs where the downscale takes effect?

Sure, but I'm wondering how to check the downscale takes effect or not? Can I just compare a generated png?

staktrace:
Looking at the reftest analyzer on the try push, I'm a little uncertain about the changes in box-shadow/boxshadow-skiprect.html and box-shadow/boxshadow-rotated.html - in both of those the border of the box itself looks blurry in the test image compared to the reference image, and I'm not sure if that's acceptable. @jrmuizel - any thoughts on if we're ok to fuzz these differences, or if they should be considered failures?

mstange:
Good catch. I think those should be considered failures. The border box clip should still be sharp.

staktrace:
Ok. Do we still want to merge this PR, or try and fix that before merging? I don't have a strong opinion either way but in general I don't like regressing correctness for performance...

glennw:
Let's try to fix it before merging - @mephisto41 could you take a look at the reftests that @staktrace mentioned above and find out what causes those differences?

mephisto41:
Yes, I'll check.

jrmuizel:
https://jrmuizel.github.io/webrender/qmusic.zip from https://bugzilla.mozilla.org/show_bug.cgi?id=1414691 is another blur related performance problem that will hopefully be fixed by this.

mephisto41:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=8ac0b402101a29ef6c74f05f1b13439bd29483e1&selectedJob=142356934

Looks like the failures are all fixed.

staktrace:
LGTM!

mstange:
I tested the patch (by pulling it from the try build) locally and found a problem with the inset box shadow in this testcase:
`data:text/html,<div style="margin:100px;height:200px;box-shadow:inset 0 0 30px black;border-radius:50px">`

before this patch:
screen shot 2017-11-06 at 9 42 36 am

after this patch:
screen shot 2017-11-06 at 9 39 40 am

mstange:
I don't mind landing the patch now because it's such a big performance improvement, and the inset box shadow bug probably isn't buying us any artificial performance, so it can be fixed later.

glennw:
Great, thanks!

@bors-servo r+

PR #1963, opened 2017-10-30
Make the API transaction-based

nical:
The idea is to replace individual `DocumentMsg`s in the api messages by vectors of `DocumentMsg` to provide a way to express the constraint that several DocumentMsg must be handled atomically without any other message interleaving in between.
This also lets us avoid redundant work when we build transactions that contain several operations that would cause the frame and/or scene to be rebuilt.

At the API level, users create a `Transaction` object and accumulate commands like `set_display_list`, `scroll` or `update_resources` into it and only touch the `RenderApi` at the end when calling `send_transaction`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1963)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -457,12 +457,7 @@ impl RenderBackend {
                 tx.send(doc.frame_ctx.get_scroll_node_state()).unwrap();
                 DocumentOps::nop()
             }
-            DocumentMsg::GenerateFrame(property_bindings) => {
-                profile_scope!("GenerateFrame");
-                let _timer = profile_counters.total_time.timer();
-
-                let mut op = DocumentOps::nop();
-
+            DocumentMsg::UpdateDynamicProperties(property_bindings) => {
                 // Ideally, when there are property bindings present,
                 // we won't need to rebuild the entire frame here.
                 // However, to avoid conflicts with the ongoing work to
@@ -473,10 +468,13 @@ impl RenderBackend {
                 //           are completed, optimize the internals of
                 //           animated properties to not require a full
                 //           rebuild of the frame!
-                if let Some(property_bindings) = property_bindings {
-                    doc.scene.properties.set_properties(property_bindings);
-                    op.build = true;
-                }
+                doc.scene.properties.set_properties(property_bindings);
+                DocumentOps::build()
+            }
+            DocumentMsg::GenerateFrame => {
+                let _timer = profile_counters.total_time.timer();
+
+                let mut op = DocumentOps::nop();
 
                 if let Some(ref mut ros) = doc.render_on_scroll {
                     *ros = true;
@@ -659,12 +657,12 @@ impl RenderBackend {
         let doc = self.documents.get_mut(&document_id).unwrap();
 
         if op.build {
-            profile_scope!("build_scene");
+            profile_scope!("build scene");
             doc.build_scene(&mut self.resource_cache);
         }
 
         if op.render {
-            profile_scope!("render");
+            profile_scope!("generate frame");
 
             *frame_counter += 1;
             let rendered_document = doc.render(
@@ -683,7 +681,6 @@ impl RenderBackend {
             );
             self.result_tx.send(msg).unwrap();
             profile_counters.reset();
-
         }
 
         if op.render || op.scroll {
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -145,12 +145,33 @@ impl Document {
     }
 }
 
-enum DocumentOp {
-    Nop,
-    Built,
-    ScrolledNop,
-    Scrolled(RenderedDocument),
-    Rendered(RenderedDocument),
+struct DocumentOps {
+    scroll: bool,
+    build: bool,
+    render: bool,
+}
+
+impl DocumentOps {
+    fn nop() -> Self {
+        DocumentOps {
+            scroll: false,
+            build: false,
+            render: false,
+        }
+    }
+
+    fn build() -> Self {
+        DocumentOps {
+            build: true,
+            ..DocumentOps::nop()
+        }
+    }
+
+    fn combine(&mut self, other: Self) {
+        self.scroll = self.scroll || other.scroll;
+        self.build = self.build || other.build;
+        self.render = self.render || other.render;
+    }
 }
 
 /// The unique id for WR resource identification.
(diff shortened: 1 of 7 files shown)


nical:
This is a work in progress, it's not too far from being ready but there are still a few unresolved TODOs in the code. For now I haven't removed the RenderApi methods that are in Transaction, so that we can hopefully first land this, update gecko and servo and then remove the deprecated APIs. Hopefully that should decouple the task of updating WebRender and moving to the new API.
Feedback very welcome.

cc @kvark @glennw

glennw:
I haven't looked at this in super detail, but the general patch looks good - the bits I looked at so far make sense to me.

nical:
Gecko try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9224869560da7e7c60097b52ba966c0c8c3a2195

The PR is in a good state to be reviewed now.

webrender/src/render_backend.rs
```diff
@@ -128,17 +128,46 @@ impl Document {
...
-    Built,
-    ScrolledNop,
-    Scrolled(RendererFrame),
-    Rendered(RendererFrame),
-}
-
 /// The unique id for WR resource identification.
 static NEXT_NAMESPACE_ID: AtomicUsize = ATOMIC_USIZE_INIT;
 
+struct DocumentOp {
```
kvark: should be renamed to `DocumentOps` now

webrender/src/render_backend.rs
```diff
@@ -422,23 +444,22 @@ impl RenderBackend {
                 //           rebuild of the frame!
                 if let Some(property_bindings) = property_bindings {
                     doc.scene.properties.set_properties(property_bindings);
-                    doc.build_scene(&mut self.resource_cache);
+                    op.build = true;
```
kvark: we should probably do the same for scroll & render events, instead of having `and_render` boolean

webrender/src/render_backend.rs
```diff
@@ -573,37 +570,69 @@ impl RenderBackend {
...
+                &mut profile_counters.resources,
+            );
+
+            // Publish the frame
+            let pending_update = self.resource_cache.pending_updates();
+            let msg = ResultMsg::NewFrame(document_id, frame, pending_update, profile_counters.clone());
+            self.result_tx.send(msg).unwrap();
+            profile_counters.reset();
+
+            if !op.scroll {
```
kvark: hacky!

webrender_api/src/api.rs
```diff
@@ -120,6 +120,170 @@ impl ResourceUpdates {
     }
 }
 
+/// A Transaction is a group of commands to apply atomically to a document.
+///
+/// This mechanism ensures that:
+///  - no other message can be interleaved between two commands that need to be applied together.
+///  - no redundant work is performed if two commands in the same tranasction cause the scen or
```
kvark: "scen", "tranasction"

glennw:
Needs a rebase.

glennw:
@nical Are you still working on this?

nical:
I do plan to finish this but unless there we identify cases that we care about for Austin where this can help performance, it is not in the list of high priority items until the workweek.

glennw:
Closing for now - let's re-open once we're ready to resume work on this.

kvark:
@glennw I don't see much value in closing the issue if it's still to be worked on. Our goal for now is to have performance up, not close as many issues as we can.

nical:
Resuming work on this (rebase in progress).

kvark:
looking good so far :+1: 

---

Reviewed 7 of 7 files at r2.
Review status: 6 of 7 files reviewed at latest revision, 4 unresolved discussions.

---

*[webrender/src/render_backend.rs, line 488 at r2](https://reviewable.io:443/reviews/servo/webrender/1963#-L1x95XXZvmKtKCclFwx:-L1x95XY7IFvexN706gc:b7rqs50) ([raw file](https://github.com/servo/webrender/blob/f7e7ba6cda6f92673140402f6ae627a70ce4d04e/webrender/src/render_backend.rs#L488)):*
> ```Rust
>                     op.render = true;
>                 } else {
>                     // TODO(nical) - Eeeeeek! The previous code did the equivalent of this, but it doesn't
> ```

we need to clear up that semantics bit, it's inherited from long ago. perhaps, nop-scroll was used like a "wakeup" so now we can remove it since we have explicit wakeup

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1963#-:-L1x9aHkvvvCJjcGmMl1:bojkfh6)*

nical:
I got the PR to build again but before merging I'd like to revive the gecko side as well and do some testing. I may move the resource updates in a separate array because they will likely need some special handling for the async scene building stuff.

nical:
The PR is in a reviewable state now. There are some more cosmethic changes that would be nice to have (such as making SetDisplayList less of a uber-message, and make epochs truly up to the api user, but that can wait. There might be some reorg inside of the Transaction struct to help with the upcoming async scene building, but that should be doable without changing the API itself.

Next is deciding:

 - whether or not to keep the old API on the side. Personally I'd rather not but that's open for bikeshed.
 - whether to keep ResourceUpdateQueue. With a dedicated transaction object we can just merge the two, I don't feel strongly either way.

These can be decided later too.

kvark:
:lgtm:

---

Reviewed 4 of 4 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1963#-:-L2MGJyYPKgoVgjgbIKL:bnfp4nl)*

webrender/src/render_backend.rs
```diff
@@ -392,66 +425,66 @@ impl RenderBackend {
...
             DocumentMsg::GetScrollNodeState(tx) => {
                 profile_scope!("GetScrollNodeState");
                 tx.send(doc.frame_ctx.get_scroll_node_state()).unwrap();
-                DocumentOp::Nop
+                DocumentOps::nop()
             }
-            DocumentMsg::GenerateFrame(property_bindings) => {
-                profile_scope!("GenerateFrame");
+            DocumentMsg::UpdateDynamicProperties(property_bindings) => {
+                // Ideally, when there are property bindings present,
```
glennw: This comment and code is obsolete, as of https://github.com/servo/webrender/pull/2043. We shouldn't need to rebuild the scene when the property bindings change.

glennw:
Looks good to me! I'll delegate r+ and leave it to nical to determine if we need a gecko try run for this change, and merge when ready (unless @kvark has any remaining questions).

@bors-servo delegate+

nical:
I had some test timeouts which ended up me messing up the gecko side integration. Works now!

@bors-servo r+

staktrace:
@nical Please post the gecko integration patch to bug 1428766 so I can include it in my testing try pushes after this merges. Thanks!

staktrace:
Ah cool, that works too :)

PR #1965, opened 2017-10-30
Set the subpixel text mask's alpha so that drawing text on transparency doesn't create black boxes.

mstange:
Drawing subpixel text to non-opaque destinations isn't really supported, but in some cases we do it anyway. One of those cases will be the Firefox tab bar on macOS as soon as we start putting a `ClearRect` primitive in there, in order to support vibrancy.

This patch is basically a quick hack to get rid of the black boxes in that case. The resulting subpixel AA will only look correct if the destination surface is later composited on top of a black background. But that's the case for the Firefox tab bar anyway (the dark vibrancy is close to black), so we're lucky.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1965)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -612,12 +612,15 @@ impl FontContext {
                         pixel[2] = 255 - pixel[2];
                     }
 
-                    pixel[3] = match font.render_mode {
-                        FontRenderMode::Subpixel => 255,
-                        _ => {
-                            pixel[0]
-                        }
-                    }; // end match
+                    // Set alpha to the value of the green channel. For grayscale
+                    // text, all three channels have the same value anyway.
+                    // For subpixel text, the mask's alpha only makes a difference
+                    // when computing the destination alpha on destination pixels
+                    // that are not completely opaque. Picking an alpha value
+                    // that's somehow based on the mask at least ensures that text
+                    // blending doesn't modify the destination alpha on pixels where
+                    // the mask is entirely zero.
+                    pixel[3] = pixel[1];
                 } // end row
             } // end height
 


mstange:
I should mention that this is only a stopgap solution; for the full solution, we need a way of text blending to transparent destinations that takes into account a "font smoothing background color". I'm currently trying to work out the math for such a blend method.

mstange:
@lsalzman asked me whether this change had any potential to break regular subpixel blending into opaque destinations. I double checked the math and I'm pretty sure it doesn't break anything, because the mask's alpha channel is never considered when computing the rgb values for text blending result, it only has influence on the alpha value of the text blending result:

```
How does current subpixel AA blending work?

Two passes:

pub fn set_blend_mode_subpixel_pass0(&self) {
    self.gl.blend_func(gl::ZERO, gl::ONE_MINUS_SRC_COLOR);
}
pub fn set_blend_mode_subpixel_pass1(&self) {
    self.gl.blend_func(gl::ONE, gl::ONE);
}

Pass0: vColor = vec4(text.color.a); oFragColor = vColor * mask;
Pass1: vColor = text.color; oFragColor = vColor * mask;

so:
result_after_pass0.x = 0 * text_color.a * mask.x + (1 - text_color.a * mask.x) * dest.x
result_after_pass1.x = 1 * text_color.x * mask.x + 1 * result_after_pass0.x
=> result.x = text_color.x * mask.x + (1 - text_color.a * mask.x) * dest.x

result_after_pass0.a = 0 * text_color.a * mask.a + (1 - text_color.a * mask.a) * dest.a
result_after_pass1.a = 1 * text_color.a * mask.a + 1 * result_after_pass0.a
=> result.a = text_color.a * mask.a + (1 - text_color.a * mask.a) * dest.a
``` 

More specifically, because we only use the GL blend funcs `ZERO`, `ONE` and `ONE_MINUS_SRC_COLOR`, no mixing between the rgb and the alpha channels occurs. We'd only have that problem if we used, for example, `ONE_MINUS_SRC_ALPHA` somewhere.

glennw:
@bors-servo r+

@mstange As a follow up, I guess we should do the same to Linux / Windows implementations?

PR #1966, opened 2017-10-30
merge gamma-lut-rs into webrender

lsalzman:
Given that gamma-lut-rs is so small and the original owner is no longer with us, it makes more sense to merge this directly into WR. This makes it far easier to make changes that simultaneously dig into the font backends and that need to modify this module. It also simplifies the module ownership issue here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1966)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -358,14 +358,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "gamma-lut"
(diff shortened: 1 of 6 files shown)


kvark:
The proper way to do this would be preserving git history, as well as possibly having it a separate crate inside this repo.

PR #1967, opened 2017-10-30
Always enable query + debugger features for wrench.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1967)


glennw:
I got tired of enabling these manually - wrench is a testing and benchmarking tool, it makes sense to have the profiler and debugger always enabled here.

r? @nical or @Gankro

jrmuizel:
@bors-servo r+
I'm tired of it too.

PR #1969, opened 2017-10-31
cleanup of subpixel glyph format

lsalzman:
This initially started as an idea between Markus and I to just store the grayscale value in the alpha channel of subpixel glyphs on all platforms (not just macOS). That, unfortunately, led to digging up a number of bugs and inconsistencies in the glyph code surrounding the subpixel glyph format, making this more of an omnibus patch to clean that all up.

Among the things addressed:
1) DWrite subpixel layout was reversed - now fixed to be BGRA. Gamma lut code simplified now that all backends are consistently BGRA.
2) FreeType bitmap code wasn't advancing destination row.
3) All platforms now output alpha for subpixel, so that we no longer need to limit text shadows to alpha format, so text shadows should generate faster and use less memory now.
4) Removed all the TextRunMode::Shadow junk since it is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1969)

### DIFF
--- a/webrender/src/platform/windows/font.rs
+++ b/webrender/src/platform/windows/font.rs
@@ -253,46 +253,46 @@ impl FontContext {
             })
     }
 
-    // DWRITE gives us values in RGB. WR doesn't really touch it after. Note, CG returns in BGR
-    // TODO: Decide whether all fonts should return RGB or BGR
-    fn convert_to_rgba(&self, pixels: &[u8], render_mode: FontRenderMode) -> Vec<u8> {
+    // DWrite ClearType gives us values in RGB, but WR expects BGRA.
+    fn convert_to_bgra(&self, pixels: &[u8], render_mode: FontRenderMode) -> Vec<u8> {
         match render_mode {
             FontRenderMode::Bitmap => {
                 unreachable!("TODO: bitmap fonts");
             }
             FontRenderMode::Mono => {
-                let mut rgba_pixels: Vec<u8> = vec![0; pixels.len() * 4];
+                let mut bgra_pixels: Vec<u8> = vec![0; pixels.len() * 4];
                 for i in 0 .. pixels.len() {
-                    rgba_pixels[i * 4 + 0] = pixels[i];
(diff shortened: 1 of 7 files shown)


webrender/src/platform/windows/font.rs
```diff
@@ -253,46 +253,46 @@ impl FontContext {
...
+                let mut bgra_pixels: Vec<u8> = vec![0; length * 4];
                 for i in 0 .. length {
-                    rgba_pixels[i * 4 + 0] = pixels[i * 3 + 0];
-                    rgba_pixels[i * 4 + 1] = pixels[i * 3 + 1];
-                    rgba_pixels[i * 4 + 2] = pixels[i * 3 + 2];
-                    rgba_pixels[i * 4 + 3] = 0xff;
+                    bgra_pixels[i * 4 + 0] = pixels[i * 3 + 0];
+                    bgra_pixels[i * 4 + 1] = pixels[i * 3 + 1];
+                    bgra_pixels[i * 4 + 2] = pixels[i * 3 + 2];
+                    bgra_pixels[i * 4 + 3] = 0xff;
```
mstange: Should this be changed to use green as well?

lsalzman: The gamma-lut now does that during the preblend (for mac and windows). This Windows code now converts to BGRA *before* the preblend happens, rather than after, so that value on input is just useless for the most part.

mstange: oh I see.

glennw:
Reviewed 6 of 7 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/1969)*

glennw:
Gecko try looks good:

https://hg.mozilla.org/try/rev/6c5a4afbe09261fa8ec2a3876c04cdc13063806c
https://treeherder.mozilla.org/#/jobs?repo=try&revision=6c5a4afbe09261fa8ec2a3876c04cdc13063806c

@bors-servo r+

PR #1970, opened 2017-10-31
Add serialize_iter_fast

Gankro:
Frequent calls to `reserve` in the profile are showing up.

NOTE: This may cause perf regressions for things that pass a Vec directly, rather than vec.iter().cloned(), but that should be easy to fix.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1970)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -532,6 +532,12 @@ impl<'a> Write for SizeCounter {
     fn flush(&mut self) -> io::Result<()> { Ok(()) }
 }
 
+/// Serializes a value assuming the Serialize impl has a stable size across two 
+/// invocations.
+///
+/// If this assumption is incorrect, the result will be Undefined Behaviour. This
+/// assumption should hold for all derived Serialize impls, which is all we currently
+/// use.
 fn serialize_fast<T: Serialize>(vec: &mut Vec<u8>, e: &T) {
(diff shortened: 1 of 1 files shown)


jrmuizel:
It's not beautiful but it looks like it should do the trick.

glennw:
Is this still WIP or ready for review?

Without having looked at this in detail, is it worth considering using some kind of custom container rather than a Vec which has better allocation patterns / performance?

jrmuizel:
This turns the glyph hotloop into:
```assembly
+0x260	movl                (%r15), %edx
+0x263	movl                %edx, (%rcx)
+0x265	movl                4(%r15), %edx
+0x269	movl                %edx, 4(%rcx)
+0x26c	movl                8(%r15), %edx
+0x270	movl                %edx, 8(%rcx)
+0x273	addq                $12, %r15
+0x277	addq                $12, %rcx
+0x27b	cmpq                %r13, %r15
+0x27e	jne                 +0x260
```
which is acceptable. Compiling with -O opt-level=3 may even fuse the load/stores.

It looks like any further reduction in time here will come from eliminating one of the copies for the glyph buffer that we do.

Gankro:
Added docs, ready for review.

Gankro:
@glennw note that the overhead was literally just *calling* reserve. Actual allocation wasn't the problem, since gecko reserves ample space on construction.

PR #1972, opened 2017-10-31
Revert "Stop creating a root scrolling node"

mrobinson:
This reverts commit 0904622fa930ea37d474871b89a5763514053af9.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1972)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -47,10 +47,9 @@ pub struct ClipScrollTree {
     /// this ID is not valid, which is indicated by ```node``` being empty.
     pub root_reference_frame_id: ClipId,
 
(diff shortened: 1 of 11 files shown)


mrobinson:
It seems like this change had a subtle, fundamental problem. Currently the size of scroll frames is important, since WebRender automatically makes a clipping region around scroll frames added via the API. Since we don't know the outer frame (either viewport or iframe) clipping size at the moment we create the display list, we cannot create a properly sized clipping region.

We are probably going to need to expose the ability to create unclipped scroll frames in the display list to make this work. Since there is a good chance that scroll frames and clips will at some point be in totally different trees, I think it makes sense to reapproach this change at that time.

glennw:
Makes sense, thanks for the explanation.

@bors-servo r+

PR #1973, opened 2017-10-31
Fix taskcluster config, after debugger feature change.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1973)


glennw:
Closed in favour of #1979.

PR #1974, opened 2017-11-01
Add support for specifying an estimated background color for subpixel text

mstange:
This is almost ready for review, but not quite.

I've written some extensive documentation for the new text blending at `webrender/doc/text-rendering.md`: [Rendered](https://github.com/mstange/webrender/blob/subpixel-with-bg-color/webrender/doc/text-rendering.md)

Remaining issues:
 - [x] rebase on top of Lee's changes to GlyphFormat
 - [x] decide whether it's reasonable to have individual font rasterization backends do `mask.a = mask.max_rgb` or whether that's cheap enough to do in the shader - conclusion: do it in the backends
 - [x] if the former, also fix up the linux and windows backends
 - [x] decide whether the middle pass should use a different shader so that regular text rendering doesn't have the overhead of the extra color field and the extra `if` in the fragment shader - conclusion: use different shader through an #ifdef in the text shader
 - ~~[ ] add some tests~~ filed #1980 to get test support

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1974)

### DIFF
--- a/webrender/doc/text-rendering.md
+++ b/webrender/doc/text-rendering.md
@@ -1,0 +1,684 @@
+# Text Rendering
+
+This document describes the details of how WebRender renders text, particularly the blending stage of text rendering.
+We will go into grayscale text blending, subpixel text blending, and "subpixel text with background color" blending.
+
+### Prerequisites
+
+The description below assumes you're familiar with regular rgba compositing, operator over,
+and the concept of premultiplied alpha.
+
+### Not covered in this document
+
+We are going to treat the origin of the text mask as a black box.
+We're also going to assume we can blend text in the device color space and will not go into the gamma correction and linear pre-blending that happens in some of the backends that produce the text masks.
+
+## Grayscale Text Blending
+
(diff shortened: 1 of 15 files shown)


glennw:
@mstange Wow, impressive doc!

mstange:
The release tests are running into `thread 'RenderBackend' panicked at 'assertion failed: request.current_used_block_num() <= MAX_VERTEX_TEXTURE_WIDTH', /Users/tcworker/webrender/webrender/src/prim_store.rs:623:8`. Looking into it.

mstange:
The test failures should be fixed now.

glennw:
@bors-servo r+

We should add some reftests for this too.

glennw:
@staktrace @mstange Heads up - this PR adds a field to FontInstanceOptions struct that silently compiles in Gecko but results in random data for the background color field. I guess the FFI bindings will need to be regenerated?

glennw:
Oh, I see there's a patch ready for this in the wr-future bug, never mind me.

Issue #1805, opened 2017-10-04
border-radius produces additional lines


Darkspirit:
Please feel free to edit the title of this issue.

https://bugzilla.mozilla.org/show_bug.cgi?id=1404158
* [Screenshot](https://bug1403976.bmoattachments.org/attachment.cgi?id=8913256)
* [Video](https://bug1404158.bmoattachments.org/attachment.cgi?id=8913466)
* [Testcase](https://bug1404158.bmoattachments.org/attachment.cgi?id=8913468)

> Nightly 58 x64 20170928100123 de_DE @ Debian Testing (KDE / Radeon RX480)
> gpu-process + layers force accel + webrender (layersfree) + webrendest
> Its looking like a "T" that's lying on the side. If I resize the window, the lines may hide and reappear.

Also seen on:
* Twitter (login button)
![screenshot_20171004_202358](https://user-images.githubusercontent.com/17962859/31192622-0abec34e-a942-11e7-82c3-6921b0ff3a05.png)
* Youtube (loading circle in the center of the video player)
![screenshot_20171004_202749](https://user-images.githubusercontent.com/17962859/31192823-94e397c0-a942-11e7-84db-c9a80f7d218d.png)

* Cloudflare (Caching page: Custom radio buttons have a horizontal line that flickers while scrolling)
![screenshot_20171004_202534](https://user-images.githubusercontent.com/17962859/31192682-3a76361c-a942-11e7-88a6-7c2237c1d4ff.png)

subsevenx2001:
I found that the bogus lines disappeared after I bypass this if to make all pixel run in the if-block anyway. Maybe there is something wrong in the else block, I suspect it is some kind of rounding issue.
if (all(lessThan(local_pos * vClipSign, vClipCenter * vClipSign))) -> if (true)
https://dxr.mozilla.org/mozilla-central/source/gfx/webrender/res/ps_border_corner.glsl#335

Darkspirit:
[Verified fixed](https://bugzilla.mozilla.org/show_bug.cgi?id=1404158#c16). Thank you! (Can we close this?)

PR #1975, opened 2017-11-01
Only draw vertical and horizontal edge when border radius is less then border width

subsevenx2001:
#1805 is caused by fragment shader incorrectly processes corner pixels outside the clip reigon which is originally written for handling border width > border radius case. 

This patch checks whether border radius is less than border width or not before drawing pixel.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1975)


subsevenx2001:
screen shot 2017-11-02 at 2 55 50 pm

I ran reftest locally and corrected some logic, now this one is the only failed reftest.

But the difference is so little that I cannot see it unless with "Circle the Difference" checked.

Hi @glennw , is that the case which need ref image being regenerated?

glennw:
In that case, you can run ```./headless.py png <reftest name>``` and it will generate and crop a new baseline image.

glennw:
@subsevenx2001 Is there any way we can add a new reftest that covers this change, to ensure we don't regress in the future?

subsevenx2001:
@glennw  It seems that the bug itself only triggered in certain zoom ratio, and the ratio varies among different screen resolution, I am not sure how to trigger it with reftest since we don't have screen resolution in headless mode when performing reftest.

glennw:
@subsevenx2001 Yea, it could be tricky. I *think* I ported the test case at https://bug1404158.bmoattachments.org/attachment.cgi?id=8913468 to a wrench yaml file and was able to reproduce it when I was doing some testing. I've lost that file, but it's probably worth just converting that test case to a wrench YAML file and seeing if it reproduces.

subsevenx2001:
@glennw I see 1920x1080 when running the script generating reference image.

I think maybe it is the base resolution when running headless mode, perhaps we can try reproducing it on a 1080p screen and convert it in YAML.

subsevenx2001:
Add a reftest for testing bogus line in border radius.

glennw:
I submitted a gecko try run for this:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=a374b822433151bd6b29fabb26b78d7fe649cfc9

If it's all green, then this is good to merge.

glennw:
Try looks good, thanks!

@bors-servo r+

Issue #1912, opened 2017-10-23
StickySideConstraint needs documentation


staktrace:
It appears that I've been laboring under a misunderstanding about how `StickySideConstraint` is supposed to work. Some documentation here would probably have saved me a couple of weeks of wasted time (and will save me from wasting more time). In particular, the way `max_offset` is used makes no sense to me. Here's my current understanding based on the code in `calculate_sticky_offset`:

This line:
```rust
sticky_offset.y = viewport_rect.min_y() + info.margin - sticky_rect.min_y();
```
seems to be computing the position the item would be in when "stuck", relative to `sticky_rect`'s current y-offset. And this line:
```rust
sticky_offset.y = sticky_offset.y.max(0.0).min(info.max_offset);
```
then tries to clamp the item so that it doesn't go outside the container it's supposed to be inside. However it's not clear to me why the min bound of the clamp range is 0 because the item might have a normal position somewhere further down inside the sticky container.

Looking at the "bottom" case we have:
```rust
                sticky_offset.y = (viewport_rect.max_y() - info.margin) -
                    (sticky_offset.y + sticky_rect.min_y() + sticky_rect.size.height);
```
which is a roundabout way of writing `sticky_offset.y = (viewport_rect.max_y() - info.margin - sticky_rect.size.height) - sticky_rect.min_y()` which is again the position the item would be in when "stuck", relative to `sticky_rect`'s current y-offset. But then once we've done that, the clamping should be exactly the same as in the previous case, but for some reason it's inverted:
```rust
sticky_offset.y = sticky_offset.y.min(0.0).max(info.max_offset);
```
and this makes no sense to me. I must be misunderstanding something but I can't find an interpretation of `max_offset` that makes this code make any sense.

@mrobinson  can you please document this API so that users can build a model of what's supposed to happen and what needs to be passed in? Thanks!

mrobinson:
@staktrace Thanks for the feedback and sorry for the pain! My plan for today is to:
1. Rework the API so the arguments are clearer. In particular I think it's better to think of the two max_offsets as the "minimum negative offset" and the "maximum positive offset". These offsets define the edges of the sticky constraint rectangle relative to the sticky rect. The "bottom" and "top" terminology just makes it very confusing.
2. Document the new API very well.
3. Add comments to this section of the code.

staktrace:
[...]

PR #1977, opened 2017-11-01
Allow gecko to provide the already-applied sticky offset to WR

staktrace:
Fixes the remaining issue in #1912.

r? @mrobinson

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1977)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{ClipId, DeviceIntRect, LayerPixel, LayerPoint, LayerRect, LayerSize};
-use api::{LayerToScrollTransform, LayerToWorldTransform, LayerVector2D, PipelineId};
+use api::{LayerToScrollTransform, LayerToWorldTransform, LayerVector2D, LayoutVector2D, PipelineId};
 use api::{ScrollClamping, ScrollEventPhase, ScrollLocation, ScrollSensitivity};
 use api::{StickyOffsetBounds, WorldPoint};
 use clip::{ClipRegion, ClipSources, ClipSourcesHandle, ClipStore};
@@ -55,6 +55,7 @@ pub struct StickyFrameInfo {
     pub margins: SideOffsets2D<Option<f32>>,
     pub vertical_offset_bounds: StickyOffsetBounds,
     pub horizontal_offset_bounds: StickyOffsetBounds,
+    pub previously_applied_offset: LayoutVector2D,
     pub current_offset: LayerVector2D,
 }
 
@@ -62,12 +63,14 @@ impl StickyFrameInfo {
     pub fn new(
         margins: SideOffsets2D<Option<f32>>,
         vertical_offset_bounds: StickyOffsetBounds,
-        horizontal_offset_bounds: StickyOffsetBounds
+        horizontal_offset_bounds: StickyOffsetBounds,
+        previously_applied_offset: LayoutVector2D
     ) -> StickyFrameInfo {
         StickyFrameInfo {
             margins,
             vertical_offset_bounds,
             horizontal_offset_bounds,
+            previously_applied_offset,
             current_offset: LayerVector2D::zero(),
         }
     }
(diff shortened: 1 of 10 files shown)


webrender/src/clip_scroll_node.rs
```diff
@@ -454,6 +457,14 @@ impl ClipScrollNode {
         }
     }
 
+    fn clamp_to_bounds(
```
mrobinson: It probably makes more sense for this to be a trait on f32 if possible. The self value is unused here.

webrender/src/clip_scroll_node.rs
```diff
@@ -478,54 +489,73 @@ impl ClipScrollNode {
...
+                // cancel out the already-applied offset, but should never end up adjusting the
+                // position the other way.
+                sticky_offset.y = top_viewport_edge - sticky_rect.min_y();
+                sticky_offset.y = sticky_offset.y.max(-info.applied_offset.y);
             }
-            debug_assert!(sticky_offset.y >= 0.0);
+            debug_assert!(sticky_offset.y + info.applied_offset.y >= 0.0);
         }
 
         if sticky_offset.y == 0.0 {
```
mrobinson: Hrm. We don't need to test sticky_offset.y plus the applied offset?

wrench/src/yaml_frame_writer.rs
```diff
@@ -999,6 +999,9 @@ impl YamlFrameWriter {
 
                     yaml_node(&mut v, "horizontal-offset-bounds", Yaml::Array(horizontal));
                     yaml_node(&mut v, "vertical-offset-bounds", Yaml::Array(vertical));
+
+                    f32_node(&mut v, "applied-offset-x", item.applied_offset.x);
```
mrobinson: Thanks for adding wrench support! I think it would be good to add a test for this as well, since there is a risk of breaking such a particular use case.

webrender/src/clip_scroll_node.rs
```diff
@@ -478,54 +489,73 @@ impl ClipScrollNode {
...
 
-        sticky_offset.y = sticky_offset.y.max(info.vertical_offset_bounds.min);
-        sticky_offset.y = sticky_offset.y.min(info.vertical_offset_bounds.max);
-        sticky_offset.x = sticky_offset.x.max(info.horizontal_offset_bounds.min);
-        sticky_offset.x = sticky_offset.x.min(info.horizontal_offset_bounds.max);
+        // The total "sticky offset" (which is the sum that was already applied by
+        // the calling code, stored in info.applied_offset, and the extra amount we
+        // computed as a result of scrolling, stored in sticky_offset) needs to be
+        // clamped to the provided bounds.
+        sticky_offset.y = self.clamp_to_bounds(sticky_offset.y + info.applied_offset.y, &info.vertical_offset_bounds) - info.applied_offset.y;
```
mrobinson: These lines extend beyond 100 characters.

webrender_api/src/display_item.rs
```diff
@@ -155,6 +155,13 @@ pub struct StickyFrameDisplayItem {
     /// position is scrolled out of view. Constraints specify a maximum and minimum offset from the
     /// original position relative to non-sticky content within the same scrolling frame.
     pub horizontal_offset_bounds: StickyOffsetBounds,
+
+    /// The amount of offset that has already been applied to the sticky frame. A positive y
+    /// component in this field means that a top-sticky item was in a scrollframe that
+    /// has been scrolled down, such that the sticky item's position needed to be offset
+    /// downwards by `applied_offset.y`. A negative y component corresponds to the upward
+    /// offset applied due to bottom-stickiness. The x-axis works analogously.
+    pub applied_offset: LayoutVector2D,
```
mrobinson: Maybe this should be called something like `previously_applied_offset` to avoid confusing it with the current offset provided by WebRender?

webrender/src/clip_scroll_node.rs
staktrace: You're right, we do. When I thought it through before I thought this was correct but thinking it through again now you're right. (For posterity, the specific thought-example is the case where gecko paints something with a top-offset already applied, and then with the async scroll we completely unapply that offset and apply a bottom-offset).

webrender_api/src/display_item.rs
staktrace: I was thinking of that too but wasn't sure how about variable name length. I can change it.

wrench/src/yaml_frame_writer.rs
staktrace: I'm not entirely sure how to do this but I'll ask around.

staktrace:
Updated. For the clamp function I just made a little lambda inside the main function which I think is cleaner than before. While writing the reftests I also discovered that neither `sticky_offset.y == 0` nor `sticky_offset.y + info.applied_offset.y == 0` were correct, so I fixed that up better and added a comment to indicate what happened there. Hopefully this version addresses all of your comments - let me know if there's anything else you'd like to see changed!

staktrace:
Travis failures are just timeouts; please ask bors to retry if the updated patch looks good to you.

staktrace:
Rebased ^. Can somebody with permissions stamp this with r=mrobinson please? Thanks!

glennw:
@bors-servo r=mrobinson

Thanks!

PR #1979, opened 2017-11-01
Follow-up to #1967, update taskcluster config as well

staktrace:
TC release tests are currently broken because of this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1979)


glennw:
I had already filed https://github.com/servo/webrender/pull/1973 to fix this, but it hasn't been reviewed. I'll close that and merge this instead.

@bors-servo r+

Issue #1774, opened 2017-09-29
[RFC] A rough plan for future optimizations.


glennw:
After doing some recent profiling and optimization work, it's now clear to me where our major bottlenecks are. I've tried to write up a basic summary and plan for fixing these below. Additionally, there's a number of outstanding bugs / issues that should become significantly easier to fix once some of these changes are in place.

A lot of this is quite vague and hand-wavy but hopefully it's useful to give some context to the upcoming changes I'd like to work on.

I don't think solving most of the problems is particularly difficult. The key is working out a way to do them incrementally, so we can continue working on other bugs and issues at the same time.

First, a list of the most obvious performance issues and major bugs I'm aware of:

CPU bottlenecks
---------------

* The DL deserialization time is a big problem. It's often ~50% of the total frame time when I profile. Fixing that is unrelated to the other things I'll mention, but I list it here since it has such a massive potential effect on CPU performance.

* Apart from DL time, the major CPU bottleneck is dealing with clip hierarchies and the clip-scroll tree. This has more to do with the size of the clip-scroll tree being larger than (we think) is required - there is some ongoing work in Gecko to try and improve this. Nonetheless, we do a lot of hashing here, and a lot of memory allocations (not in the clip-scroll tree itself, but in processing masks and building clip tasks). We also do a lot of CPU work due to an impedance mismatch between the clip-scroll tree and the batching / shader code (specifically, the way we handle clip-scroll groups, packed layers etc).

* Those two major areas listed above typically account for ~80% of the CPU time processing a frame. After those, batch creation is typically the next most expensive area. More details on this in the GPU section.

GPU bottlenecks
---------------

* On the sites we're seeing in Gecko, our draw call counts are higher than ideal. They are still good (typically ~50 per frame), but improving this would be a good win for both CPU time (batching) and GPU time (shader switches etc).

* It occurs to me that many of the effects we now have separate shaders for can in fact be unified into a smaller set of shaders. For example, the line decoration shader can just be a rectangle with a clip mask. Similarly, many of the specialized border shaders we have now can be vastly simplified by moving the shaping into a clip mask - often collapsing to a rectangle shader with a clip mask.

[...]

PR #1981, opened 2017-11-02
Remove ClipScrollGroup and PackedLayer.

glennw:
Instead, we provide access to both the clip node and scroll
node for a given primitive in the shader. The shader then
calculated the appropriate local clip rect.

There's a few hacks in here - for example, the way that
I pack the clip/scroll IDs into a single u32 for the
shaders. In the future, we'll be switching the vertex formats
to work on u16 size data fields, so this will become a lot
simpler then.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1981)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -71,14 +71,14 @@ vec4[2] fetch_from_resource_cache_2(int address) {
 
 #ifdef WR_VERTEX_SHADER
 
-#define VECS_PER_LAYER              9
+#define VECS_PER_LAYER              10
 #define VECS_PER_RENDER_TASK        3
 #define VECS_PER_PRIM_HEADER        2
 #define VECS_PER_TEXT_RUN           3
 #define VECS_PER_GRADIENT           3
 #define VECS_PER_GRADIENT_STOP      2
 
-uniform HIGHP_SAMPLER_FLOAT sampler2D sLayers;
+uniform HIGHP_SAMPLER_FLOAT sampler2D sClipScrollNodes;
 uniform HIGHP_SAMPLER_FLOAT sampler2D sRenderTasks;
 
 // Instanced attributes
@@ -143,14 +143,16 @@ vec4 fetch_from_resource_cache_1(int address) {
     return texelFetch(sResourceCache, uv, 0);
 }
 
-struct Layer {
+struct ClipScrollNode {
     mat4 transform;
     mat4 inv_transform;
-    RectWithSize local_clip_rect;
+    vec4 local_clip_rect;
+    vec2 reference_frame_relative_scroll_offset;
+    vec2 scroll_offset;
 };
 
-Layer fetch_layer(int index) {
-    Layer layer;
+ClipScrollNode fetch_clip_scroll_node(int index) {
+    ClipScrollNode node;
 
     // Create a UV base coord for each 8 texels.
     // This is required because trying to use an offset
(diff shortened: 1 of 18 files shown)


glennw:
r? @mrobinson 

This is not *quite* ready for merge yet - there are a couple of failures in the Gecko try and and a couple of failures in the Servo WPT suite I need to fix. But I think those should be fairly minor fixes, so this is probably ready for review.

Gecko try:

https://hg.mozilla.org/try/rev/933c76557503d46ff73501b31ebbda5be1eefe30
https://treeherder.mozilla.org/#/jobs?repo=try&revision=933c76557503d46ff73501b31ebbda5be1eefe30

glennw:
This simplifies a lot of the CPU code that deals with primitives, and makes it feasible to work on the changes proposed in #1774.

mrobinson:
@gw I absolutely love this change. I've taken a quick look and don't see any obvious problems, but I'll do a more thorough review today.

mrobinson:
Okay. I have one question about the ClipScrollNodeData. It seems that it is associated to primitives via their assigned scroll node, yet it contains information about the local clip rect. The local clip rect is a property of the primitive's assigned clip node. Perhaps that is why you are seeing failures on Gecko.

glennw:
@mrobinson OK, there are three small fix-up commits added. With those changes, all WR / Servo / Gecko tests pass.

Gecko try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=844bc16b48eea4096461c405e96547755590517b

glennw:
@mrobinson Thanks, fixed those comments up in the last commit. Once you review that change, I'll squash before merge.

glennw:
Squashed fixup commits.

@bors-servo r=mrobinson

webrender/res/prim_shared.glsl
```diff
@@ -160,18 +162,46 @@ Layer fetch_layer(int index) {
...
+}
+
+struct Layer {
+    mat4 transform;
+    mat4 inv_transform;
+    RectWithSize local_clip_rect;
+};
+
+Layer fetch_layer(int clip_node_id, int scroll_node_id) {
+    ClipScrollNode clip_node = fetch_clip_scroll_node(clip_node_id);
```
kvark: we should have a fast path for `clip_node_id == scroll_node_id` instead of fetching the same data again

webrender/res/prim_shared.glsl
```diff
@@ -160,18 +162,46 @@ Layer fetch_layer(int index) {
...
+    vec4 offsets = TEXEL_FETCH(sClipScrollNodes, uv1, 0, ivec2(1, 0));
+    node.reference_frame_relative_scroll_offset = offsets.xy;
+    node.scroll_offset = offsets.zw;
+
+    return node;
+}
+
+struct Layer {
+    mat4 transform;
+    mat4 inv_transform;
```
kvark: we should gate the `inv_transform` fetches with the `WR_FEATURE_TRANSFORM` to save on fetches for a more popular (aligned) case

webrender/res/prim_shared.glsl
kvark: also note: the `clip_note.inv_transform` is not used here at all, so we shouldn't fetch it

webrender/src/clip_scroll_node.rs
```diff
@@ -452,6 +434,39 @@ impl ClipScrollNode {
                     info.current_offset + state.parent_accumulated_scroll_offset;
             }
         }
+
+        // Store coord system ID, and also the ID used for shaders to reference this node.
+        self.coordinate_system_id = state.current_coordinate_system_id;
+        self.id = ClipScrollNodeIndex(node_data.len() as u32);
```
kvark: might be useful to have the name of this type reflecting the fact it's for GPU/shader consumption only

webrender/src/frame_builder.rs
```diff
@@ -113,11 +113,6 @@ pub struct FrameBuilder {
     pub config: FrameBuilderConfig,
 
     stacking_context_store: Vec<StackingContext>,
-    clip_scroll_group_store: Vec<ClipScrollGroup>,
```
kvark: sweeeet!

PR #1989, opened 2017-11-05
Write out stacking context filters

jrmuizel:
Not writing out opacity explains why our yaml wrench numbers
sometimes looked better than our binary recording numbers.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1989)


glennw:
@jrmuizel I had been meaning to investigate this, thanks!

@bors-servo r+

PR #1996, opened 2017-11-06
Add support for handling property bindings in the yaml writer

jrmuizel:
This doesn't try and do anything fancy, it just writes out the current
value. The code is very similar to the code from webrender/scene.rs but
probably not worth sharing, because it's pretty trivial.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1996)

### DIFF
--- a/wrench/src/scene.rs
+++ b/wrench/src/scene.rs
@@ -5,6 +5,70 @@
 use std::collections::HashMap;
 use webrender::api::{BuiltDisplayList, ColorF, Epoch};
 use webrender::api::{LayerSize, PipelineId};
+use webrender::api::{PropertyBinding, PropertyBindingId, LayoutTransform, DynamicProperties};
+
+/// Stores a map of the animated property bindings for the current display list. These
+/// can be used to animate the transform and/or opacity of a display list without
+/// re-submitting the display list itself.
+#[derive(Default)]
+pub struct SceneProperties {
+    transform_properties: HashMap<PropertyBindingId, LayoutTransform>,
+    float_properties: HashMap<PropertyBindingId, f32>,
+}
+
+impl SceneProperties {
+    /// Set the current property list for this display list.
+    pub fn set_properties(&mut self, properties: &DynamicProperties) {
+        self.transform_properties.clear();
+        self.float_properties.clear();
+
+        for property in &properties.transforms {
+            self.transform_properties
+                .insert(property.key.id, property.value);
+        }
+
+        for property in &properties.floats {
+            self.float_properties
+                .insert(property.key.id, property.value);
+        }
+    }
+
+    /// Get the current value for a transform property.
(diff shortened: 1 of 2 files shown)


wrench/src/scene.rs
```diff
@@ -5,6 +5,76 @@
...
+/// Stores a map of the animated property bindings for the current display list. These
+/// can be used to animate the transform and/or opacity of a display list without
+/// re-submitting the display list itself.
+pub struct SceneProperties {
+    transform_properties: HashMap<PropertyBindingId, LayoutTransform>,
+    float_properties: HashMap<PropertyBindingId, f32>,
+}
+
+impl SceneProperties {
+    pub fn new() -> SceneProperties {
```
kvark: would be easier to just put `#[derive(Default)]` on top of the struct

wrench/src/scene.rs
```diff
@@ -5,6 +5,76 @@
...
+        }
+
+        for property in &properties.floats {
+            self.float_properties
+                .insert(property.key.id, property.value);
+        }
+    }
+
+    /// Get the current value for a transform property.
+    pub fn resolve_layout_transform(
```
kvark: so why can't we share this code with the `scene.rs`? It's not complex, but still a few pages

wrench/src/scene.rs
```diff
@@ -16,6 +86,7 @@ pub struct ScenePipeline {
 
 /// A complete representation of the layout bundling visible pipelines together.
 pub struct Scene {
+    pub properties: SceneProperties,
```
kvark: this `Scene` struct could also benefit from `Default` being derived

wrench/src/yaml_frame_writer.rs
```diff
@@ -179,14 +179,10 @@ fn maybe_radius_yaml(radius: &BorderRadius) -> Option<Yaml> {
     }
 }
 
-fn write_sc(parent: &mut Table, sc: &StackingContext, filter_iter: AuxIter<FilterOp>) {
+fn write_sc(parent: &mut Table, sc: &StackingContext, scene: &Scene, filter_iter: AuxIter<FilterOp>) {
```
kvark: nit: looks like we actually only need the properties here and below, not the whole scene

wrench/src/scene.rs
jrmuizel: It's really just setting and getting from hash tables, and there are differences in the the signatures of the resolve methods due to the kind of references passed in. Also one piece of code uses warn! vs println!. Sharing it really just seems like a hassle.

glennw:
I'm not sure if there's further follow ups needed to this, but I figure we can do them after merge if required, so I'll get this merged while the travis CI queue is relatively empty.

@bors-servo r+

PR #1998, opened 2017-11-06
Use constant color blend for subpx text without clips.

### LINKED ISSUE #1984: Very high draw call count on simple sites in Gecko.
For example, in HN the Gecko + WR is submitting around 275 draw calls per frame. In Servo, this is typically less than 10.

Investigating now since this is likely to be a simple bug and a significant performance win once fixed.

glennw:
The current subpx rendering mode that supports clip masks is a
two pass technique. With the current batcher implementation,
this can result in a lot of batches on sites where text
bounding rects are a bit loose, and are detected as potentially
overlapping. This most often occurs on sites that use a lot
of small text.

As an interim solution, only use the clip mask supporting method
when a clip mask (or translucent text) is active. Otherwise, use
the old color blending method.

In the near(ish) future, we'll start using the dual source blending
extension for subpixel rendering on hardware that supports it. This
extension gives the best of both worlds - supporting clip masks and
being a single pass technique. It's supported on the vast majority
of our target hardware (perhaps even all of it).

On hardware that doesn't support dual source blending, using the
modes here which invoke a different technique depending on need
for alpha masking is a very reasonable tradeoff.

Additionally, future improvements to batching will improve this
fallback solution by being a bit smarter about clumping together
non-overlapping batches.

Fixes #1984.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1998)


webrender/src/device.rs
```diff
@@ -1925,6 +1925,11 @@ impl Device {
         self.gl.blend_func_separate(gl::ONE, gl::ONE, gl::ONE, gl::ONE_MINUS_SRC_ALPHA);
         self.gl.blend_equation(gl::FUNC_ADD);
     }
+    pub fn set_blend_mode_subpixel_opaque(&self, color: ColorF) {
+        self.gl.blend_color(color.r, color.g, color.b, color.a);
+        self.gl
+            .blend_func(gl::CONSTANT_COLOR, gl::ONE_MINUS_SRC_COLOR);
```
kvark: we should be setting the blend equation here as well as in `set_blend_mode_subpixel_passX`

mstange:
I think this method can be extended in a fairly straightforward way to cover non-opaque text as well (as long as the alpha is constant):

 - In the shader, use vec4(text.color.a) as the factor
 - Set the blend color to the unpremultiplied text color, i.e. color.r / color.a etc.

Then the resulting equation is

```
result.r = text_color.r / text_color.a * (text_color.a * mask.r) +
           (1 - text_color.a * mask.r) * dest.r
         = text_color.r * mask.r + (1 - text_color.a * mask.r) * dest.r
```

mstange:
> i.e. color.r / color.a etc.

Actually, if the color passed to `set_blend_mode_subpixel_opaque` is already unpremultiplied (which I think it is), then this would just be `self.gl.blend_color(color.r, color.g, color.b, 1.0);`.

webrender/src/device.rs
glennw: Fixed

glennw:
@mstange That's a good point - we can definitely apply that as a follow up optimization, and only use the 2 pass method when a clip mask is present.

@kvark Fixed up that comment. r=you if the try run I'm starting is green?

glennw:
Gecko try looks good:

https://hg.mozilla.org/try/rev/d98097650e8a8168cf2f0736ddd0d2ae4a1947dd
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d98097650e8a8168cf2f0736ddd0d2ae4a1947dd

@bors-servo r=kvark

PR #1999, opened 2017-11-06
Allow RGBA8 + nearest filtered images in shared texture cache.

### LINKED ISSUE #1992: Make nearest filtered images use the texture cache efficiently.

glennw:
This drops the draw call count when drawing the Gecko chrome UI
by a fairly significant amount, where favicons are being drawn
with nearest filtering enabled.

Fixes #1992.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/1999)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -326,19 +326,21 @@ impl FrameProfileCounters {
 
 #[derive(Clone)]
 pub struct TextureCacheProfileCounters {
-    pub pages_a8: ResourceProfileCounter,
-    pub pages_rgb8: ResourceProfileCounter,
-    pub pages_rgba8: ResourceProfileCounter,
-    pub pages_rg8: ResourceProfileCounter,
+    pub pages_a8_linear: ResourceProfileCounter,
+    pub pages_rgb8_linear: ResourceProfileCounter,
+    pub pages_rgba8_linear: ResourceProfileCounter,
+    pub pages_rgba8_nearest: ResourceProfileCounter,
+    pub pages_rg8_linear: ResourceProfileCounter,
 }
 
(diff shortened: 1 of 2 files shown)


glennw:
Yea, it's less than ideal. The reason I did it this way was due to an experience I had on a previous project.

I had found that on certain GPUs (mostly mobile, I think), changing the sampling state on the texture would cause *massive* frame spikes. In the project I was dealing with, switching sampling mode on a texture would often take > 16ms each time we drew something with that updated texture. My guess is that the driver had to rebuild the shader in some fairly significant way when the sampling mode was different.

It may well be that this no longer occurs, but it's definitely something I've run into in the past. I think it's probably reasonable to merge this, even if as an interim fix, what do you think? Perhaps we could  investigate switching the sampler mode on various GPUs and make sure we don't see any spikes as as a follow up?

kvark:
Re-linking the shader for different sampler modes sounds completely bizzare to me. Maybe some ancient hardware? If that was how mobile GPUs work, they'd reflect this somehow in Vulkan, ideally...

Also, as I mentioned, we can bind a different sampler object instead of changing the state of the current texture. This should be more straightforward for the picky driver to handle.

Given the concerns, it does appear reasonable to defer my requested improvements until we investigate the mobile GPU behavior for those cases. Therefore, let's merge this, even if we'll need to scrap/revert the change later.

@bors-servo r+

glennw:
I'm trying to recall the exact hardware - I *think* it might have been iPhone 4 / iPad 2 era GPUs. It was really completely bizarre, I couldn't believe it either :)

Thanks for the review - I'm certainly happy to have this scrapped and a better solution implemented as a follow up!

PR #2007, opened 2017-11-07
Fix rendering of dotted corners when corner radius < border width.

### LINKED ISSUE #1932: Dotted borders render incorrectly
Hi,
I've noticed dotted borders render the corners seemingly incorrectly.

```
---
root: 
  items: 
    - 
      bounds: [5, 5, 490, 490]
      "clip-rect": [5, 5, 490, 490]
      "clip-and-scroll": 0
      "backface-visible": true
      type: border
      width: 25
      "border-type": normal
      color: black
      style: dotted
  id: [0, 0]
pipelines: []
```

nical:
Fixes #1932.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2007)


nical:
Before merging this I'll publish a new version of euclid with the Size2D.abs() implemented to make this code a bit nicer.

r? @glennw

glennw:
It's definitely not specified, but this does look visually wrong to me. Shouldn't we do something like draw a single dot in the corner in this case, and then space the dots along the edge appropriately?

PR #2009, opened 2017-11-07
Use the glBlendColor approach for all subpixel text (except the WithBgColor case)

mstange:
We can use the glBlendColor approach even if we have transparent text and even if we have a clip or a mask. The math works out correctly.

The clip can be viewed as something that gets applied to the text mask before we use the text mask for blending.

r? @kvark or @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2009)


glennw:
If the math for this works out, great (I didn't do the math myself to check)!

There are a few reftest failures though - they might be very small differences that just need a new reference generated?

mstange:
Both reftest failures only have a max difference of 1, so I'll fuzz them.

glennw:
Since they are PNG references, I think it's better to regenerate the images.

mstange:
Oops, you're right, I missed that they were PNG references.

mstange:
Updated the images. r? @glennw

glennw:
@mstange Could you crop the reference images? Sadly, the auto-crop fails here since the PNG name doesn't match the test name (due to the same test being run once in subpx mode and once in alpha mode). Sorry!

glennw:
I should make the auto-crop a bit smarter in that regard...

mstange:
Sure, will crop.

mstange:
Done.

glennw:
Thank you!

@bors-servo r+

PR #2014, opened 2017-11-08
Register thread to WR's profiler and the external profiler.

JerryShih:
The gecko profiler needs the thread to register itself to gecko profiler before collecting the performance data. In this patch, we have a new "thread listener" interface to do somethings when the thread is created and deleted. Then, we could put the registering function inside the listener interface.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2014)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1787,33 +1787,57 @@ impl Renderer {
         let debug_flags = options.debug_flags;
         let payload_tx_for_backend = payload_tx.clone();
         let recorder = options.recorder;
-        let worker_config = ThreadPoolConfig::new()
-            .thread_name(|idx| format!("WebRender:Worker#{}", idx))
-            .start_handler(|idx| {
-                register_thread_with_profiler(format!("WebRender:Worker#{}", idx));
-            });
+        let thread_listener = Arc::new(options.thread_listener);
+        let thread_listener_for_rayon_start = thread_listener.clone();
+        let thread_listener_for_rayon_end = thread_listener.clone();
         let workers = options
             .workers
             .take()
-            .unwrap_or_else(|| Arc::new(ThreadPool::new(worker_config).unwrap()));
+            .unwrap_or_else(|| {
+                let worker_config = ThreadPoolConfig::new()
+                    .thread_name(|idx|{ format!("WRWorker#{}", idx) })
+                    .start_handler(move |idx| {
+                        register_thread_with_profiler(format!("WRWorker#{}", idx));
+                        if let Some(ref thread_listener) = *thread_listener_for_rayon_start {
+                            thread_listener.thread_started(&format!("WRWorker#{}", idx));
+                        }
+                    })
+                    .exit_handler(move |idx| {
+                        if let Some(ref thread_listener) = *thread_listener_for_rayon_end {
+                            thread_listener.thread_stopped(&format!("WRWorker#{}", idx));
+                        }
+                    });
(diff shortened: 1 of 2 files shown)


webrender/src/renderer.rs
```diff
@@ -1787,33 +1787,60 @@ impl Renderer {
         let debug_flags = options.debug_flags;
         let payload_tx_for_backend = payload_tx.clone();
         let recorder = options.recorder;
-        let worker_config = ThreadPoolConfig::new()
```
JerryShih: The worker_config is only used when we don't have the thread pool. So, move the config into the worker's unwrap_or_else() block.

webrender/src/renderer.rs
```diff
@@ -3845,8 +3877,10 @@ pub struct RendererOptions {
     pub workers: Option<Arc<ThreadPool>>,
     pub blob_image_renderer: Option<Box<BlobImageRenderer>>,
     pub recorder: Option<Box<ApiRecordingReceiver>>,
+    pub thread_listener: Option<Box<ThreadListener + Send + Sync>>,
     pub enable_render_on_scroll: bool,
     pub debug_flags: DebugFlags,
+    pub renderer_id: Option<u64>,
```
JerryShih: We might have multiple renderer in the same process. Use this renderer_id for identification.

JerryShih:
Please check:
https://bugzilla.mozilla.org/show_bug.cgi?id=1371838

r? @glennw @kvark

glennw:
Reviewed 2 of 2 files at r1.

---

*[webrender/src/renderer.rs, line 1799 at r1](https://reviewable.io:443/reviews/servo/webrender/2014#-KyUP0ytwbse5Qj4NPMX:-KyUP0ytwbse5Qj4NPMY:bjguotq) ([raw file](https://github.com/servo/webrender/blob/e9f943ccdd9dfaf7bbfb144c3d1a051dd1cec0e1/webrender/src/renderer.rs#L1799)):*
> ```Rust
>                 let worker_config = ThreadPoolConfig::new()
>                     .thread_name(|idx|{ format!("WRWorker#{}", idx) })
>                     .start_handler(move |idx| {
> ```

I think the start and end handlers are only called when the worker thread is created and destroyed. If so, that won't actually be useful for profiling - since we won't know when the thread is doing work or idle?

---

*[webrender/src/renderer.rs, line 1816 at r1](https://reviewable.io:443/reviews/servo/webrender/2014#-KyUOrL0TpYw7kisJQV1:-KyUOrL1F0WDHlVuJD36:b-whhr54) ([raw file](https://github.com/servo/webrender/blob/e9f943ccdd9dfaf7bbfb144c3d1a051dd1cec0e1/webrender/src/renderer.rs#L1816)):*
> ```Rust
>         let blob_image_renderer = options.blob_image_renderer.take();
>         let thread_listener_for_render_backend = thread_listener.clone();
>         let mut thread_name = String::from("WRRenderBackend");
> ```

This could be something like:

let thread_name = format!("WRRenderBackend {}", options.renderer_id.unwrap_or(0));

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2014)*

glennw:
Wow, I had not extensively used the Gecko CPU profiler before. It's really excellent - and with this addition we get perfect callstacks for profiling in the WR backend thread. @mrobinson @kvark heads up - next time you need to profile some CPU code in WR, you should try this out! :)

Thanks @JerryShih and others who worked on the patch.

PR #2015, opened 2017-11-08
Add support for scrollbars per Scroll Frame

nc4rrillo:
Hi,
This PR adds support for specifying scrollbars being enabled per Scroll Frame by adding an additional bool to `define_scroll_frame`.

This is a (albeit tiny) breaking API change, so I've bumped the WR API version. I hope I did that right :).
I also updated the scrolling sample to show off the new functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2015)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -8,7 +8,7 @@ use api::{DeviceUintRect, DeviceUintSize, DisplayItemRef, DocumentLayer, Epoch, 
 use api::{ImageDisplayItem, ItemRange, LayerPoint, LayerPrimitiveInfo, LayerRect};
 use api::{LayerSize, LayerVector2D};
 use api::{LayoutRect, LayoutSize};
-use api::{LocalClip, PipelineId, ScrollClamping, ScrollEventPhase, ScrollLayerState};
+use api::{LocalClip, PipelineId, ScrollClamping, ScrollEventPhase, ScrollLayerState, IframeScrollbars};
 use api::{ScrollLocation, ScrollPolicy, ScrollSensitivity, SpecificDisplayItem, StackingContext};
 use api::{ClipMode, TileOffset, TransformStyle, WorldPoint};
 use clip::ClipRegion;
@@ -28,7 +28,7 @@ use util::ComplexClipRegionHelpers;
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug, Eq, Ord)]
 pub struct FrameId(pub u32);
 
-static DEFAULT_SCROLLBAR_COLOR: ColorF = ColorF {
+pub static DEFAULT_SCROLLBAR_COLOR: ColorF = ColorF {
     r: 0.3,
     g: 0.3,
     b: 0.3,
@@ -196,6 +196,7 @@ impl<'a> FlattenContext<'a> {
         content_rect: &LayerRect,
         clip_region: ClipRegion,
         scroll_sensitivity: ScrollSensitivity,
+        enable_scrollbars: bool
     ) {
         let clip_id = self.clip_scroll_tree.generate_new_clip_id(pipeline_id);
         self.builder.add_clip_node(
@@ -213,6 +214,7 @@ impl<'a> FlattenContext<'a> {
             &frame_rect,
             &content_rect.size,
             scroll_sensitivity,
+            enable_scrollbars,
             self.clip_scroll_tree,
         );
     }
@@ -362,6 +364,7 @@ impl<'a> FlattenContext<'a> {
             &iframe_rect,
             &pipeline.content_size,
             ScrollSensitivity::ScriptAndInputEvents,
+            false,
             self.clip_scroll_tree,
         );
 
@@ -589,6 +592,9 @@ impl<'a> FlattenContext<'a> {
                     .clip_rect()
                     .translate(&reference_frame_relative_offset);
                 let content_rect = item.rect().translate(&reference_frame_relative_offset);
+
+                let enable_scrollbars = IframeScrollbars::Enabled == info.enable_scrollbars;
+
                 self.flatten_scroll_frame(
                     pipeline_id,
                     &clip_and_scroll.scroll_node_id,
@@ -597,6 +603,7 @@ impl<'a> FlattenContext<'a> {
                     &content_rect,
                     clip_region,
                     info.scroll_sensitivity,
+                    enable_scrollbars
                 );
             }
             SpecificDisplayItem::StickyFrame(ref info) => {
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -18,7 +18,7 @@ use clip::{ClipRegion, ClipSource, ClipSources, ClipStore, Contains, MAX_CLIP};
 use clip_scroll_node::{ClipScrollNode, NodeType};
 use clip_scroll_tree::ClipScrollTree;
 use euclid::{SideOffsets2D, vec2};
-use frame::FrameId;
+use frame::{FrameId, DEFAULT_SCROLLBAR_COLOR};
 use glyph_rasterizer::FontInstance;
 use gpu_cache::GpuCache;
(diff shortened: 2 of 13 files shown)


nc4rrillo:
There are two reftest failures I'm looking into related to the root scroll frame scrollbar being offset a bit.

I've tried to revert my change, and instead move the logic to add the scrollbar for other scroll frames into `flatten_scroll_frame` where it seems more appropriate, however its drawing under the content of the scroll frame now.

nc4rrillo:
So I decided to keep the RendererOptions `enable_scrollbars` option around to control whether or not the root scroll frame gets a scrollbar, since it is created implicitly.

webrender/src/frame.rs
```diff
@@ -117,6 +117,7 @@ impl<'a> FlattenContext<'a> {
         self.flatten_items(traversal, pipeline_id, LayerVector2D::zero());
 
         if self.builder.config.enable_scrollbars {
```
nc4rrillo: Maybe the option can be renamed to more accurately reflect what it does if my change is accepted?

webrender/src/frame_builder.rs
```diff
@@ -458,6 +458,7 @@ impl FrameBuilder {
             &viewport_rect,
             content_size,
             ScrollSensitivity::ScriptAndInputEvents,
+            false,
```
nc4rrillo: Previously I set this to true, but it caused reftest failures, I'm wondering if its because of some special handling of the root scroll frame.

webrender/src/frame.rs
```diff
@@ -128,6 +129,23 @@ impl<'a> FlattenContext<'a> {
             );
         }
 
+        // Draw scrollars for other scroll frames
+        for (clip_id, clip_node) in &self.clip_scroll_tree.nodes {
```
nc4rrillo: I'm thinking in `flatten_root` I should filter the clip scroll nodes by the current PipelineId, else we'll run into trouble with a nested Iframe?

glennw:
Apologies for not getting to this - will look at it on Monday if no-one else gets to it before me.

mrobinson:
@nc4rrillo Sorry! I hadn't seen this PR when I made my recent scrollbar change. It looks like the two PRs are compatible, even though there is now a merge conflict.

nc4rrillo:
@mrobinson No worries. I'll rebase my changes and update this PR.

nc4rrillo:
So there is still an issue that's visible in the scrolling example, where if you scroll the outer scroll frame, the inner scrollbar isn't clipped by its parents bounds.

nc4rrillo:
Here's what I mean by the last comment:


It only occurs if I scroll the outer scroll frame

nc4rrillo:
Okay I fixed that issue, I think this is ready for review.

webrender_api/src/display_list.rs
```diff
@@ -1162,6 +1162,7 @@ impl DisplayListBuilder {
         complex_clips: I,
         image_mask: Option<ImageMask>,
         scroll_sensitivity: ScrollSensitivity,
+        enable_scrollbars: bool
```
mrobinson: I think it would be better to use an enum for this argument. Just judging by the changes to the examples, it's already a little hard to tell what this argument does. I suggest something like

```
enum IframeScrollbars {
    Enabled,
    Disabled,
}

File: webrender/src/frame.rs (hunk 3 above)
Comment (mrobinson) [reply webrender/src/frame.rs]: Hrm. It seems like it would be a lot easier just to add the item when processing the scroll frame item during flattening. Otherwise I think that stacked scroll frames will have all their scrollbars on the top of the scene.

File: webrender/src/frame_builder.rs [hunk 5]
```diff
@@ -500,6 +502,7 @@ impl FrameBuilder {
             frame_rect,
             content_size,
             scroll_sensitivity,
+            enable_scrollbars
```
mrobinson: I'm not sure we need to pass this value to the ClipScrollNode. I think you can just add the scrollbar here as a child of the parent ClipScrollNode. You might need to do some special calculation to ensure that the scrollbar stays within the rectangle of the parent node.

webrender/src/frame_builder.rs
mrobinson: Yes. Root scroll frame scrollbars are handled automatically by the ENABLE_SCROLLBARS option. We don't want this turned on in Gecko anyhow.

nc4rrillo:
@mrobinson I've tried to move creating the scrollbars to `add_scroll_frame`, however an issue I'm having is that the scrollbar gets drawn before the content of the scroll frame. 

Here's a screen shot with debug alpha primitives turned on to demonstrate:

nc4rrillo:
I believe this is because `add_scroll_frame` is called, and then the builder gets invoked for the children of the rest of the display items that make up the contents, which draws over what I've pushed into the frame builder. 

One of the reasons why doing it in `flatten_root` worked was because `flatten_items` had already been called at that point.

nc4rrillo:
Okay so my proposed fix is to do the following:
- Introduce a `pending_scroll_frame` stack to `FrameBuilder`
- On every `add_scroll_frame`, check if the stack contains a pending scroll frame, and pop it off and draw scrollbars for it
- Push the current scroll frame to the stack.

This takes care of most cases, I add a final check to `pop_stacking_context` to pop what should be the final remaining pending scroll frame, and draw scrollbars for it.

My thinking is that works around the there not being a specific `PopScrollFrame` display item, so by drawing the _previous_ scroll frame's scrollbars on the _next_ scroll frame's `add_scroll_frame` invocation, the drawing order is preserved and we avoid the stacked scroll frame issue you mentioned.

webrender/src/frame.rs
```diff
@@ -612,6 +594,9 @@ impl<'a> FlattenContext<'a> {
                     .clip_rect()
                     .translate(&reference_frame_relative_offset);
                 let content_rect = item.rect().translate(&reference_frame_relative_offset);
+
+                let enable_scrollbars = IframeScrollbars::Enabled == info.enable_scrollbars;
```
nc4rrillo: I turned the enum into a bool here, but I could just as easily thread it all the way down into FrameBuilder if need be.

webrender/src/frame_builder.rs
```diff
@@ -129,6 +129,9 @@ pub struct FrameBuilder {
     /// parent new scroll nodes.
     reference_frame_stack: Vec<ClipId>,
 
+    parent_scroll_frame_id: Option<ClipId>,
+    pending_scroll_frame_stack: Vec<(ClipId, ClipId, LayerRect)>,
+
```
nc4rrillo: I doubt think this _needs_ to be a Vec, I don't think there will be more than one pending scroll frame that hasn't drawn scrollbars at a time.

nc4rrillo:
I pushed a commit to address your review and implement the proposed change I outlined above.

nc4rrillo:
Test failures are because I need to update wrench

glennw:
Looks like this needs a rebase, sorry!

nc4rrillo:
Rebased and bumped crate versions for wr, wr api, and wrench

mrobinson:
@nc4rrillo Sorry for not giving this as much attention as I should have. I have been rolling this around the back of my head for the past week or so. The approach you take, which is to wait until the display list flattening sees another `add_scroll_frame` is not going to work for the general case and definitely not for Servo (which pushes all ScrollFrames at the start of StackingContexts). 

Instead of doing this now and then having to redo it in the future, I think the right approach here is to let the client control adding scrollbars completely. To me, the most obvious way to do this is to expose a scrollbar display item. This display item would, for the moment, be a rectangle with a height, width, and color. WebRender would be responsible for moving that item around in the ScrolFrame's parent just like for existing scrollbars.

Again, sorry for the delay and perhaps leading you down a dead end here.

nc4rrillo:
No worries. I like your approach much more. I’m going to close this and work on that.

Issue #1384, opened 2017-06-15
Some corners in Servo's transparent rects demo have seams


trishume:
Some of the rectangles in the transparent rects demo with the Servo nightly from 2017/06/14 have seams on the corners on a 2014 rMBP with both the Nvidia GT 750m and the Iris Pro. Which rectangles have seams on the corners does not change between frames. All rectangles also have weird white line artifacts on the corners.

### Nvidia GT 750M
<img width="252" alt="screen shot 2017-06-14 at 10 07 40 pm" src="https://user-images.githubusercontent.com/887610/27166464-a795deb4-514f-11e7-8fa9-1f73bfc6061f.png">

### Nvidia GT 750M after window resize
After resizing the window to full screen, which doesn't actually change the demo, everything starts flickering wildly with black rectangles.
<img width="370" alt="screen shot 2017-06-14 at 10 22 51 pm" src="https://user-images.githubusercontent.com/887610/27166515-1a3d835e-5150-11e7-887a-3d225e641998.png">

### Intel Iris Pro
<img width="510" alt="screen shot 2017-06-14 at 10 21 37 pm" src="https://user-images.githubusercontent.com/887610/27166532-47a53f12-5150-11e7-9d10-04642f16263b.png">

kvark:
cc @glennw 
looks like a border rendering code, perhaps it's transparency is not working correctly with `device_pixel_ratio != 1` case

glennw:
https://github.com/servo/webrender/pull/2016 claims this is fixed.

kvark:
Notice how I changed "fixed XXX" to "related to XXX" in the description. This issue definitely needs to be retested though.

Issue #1297, opened 2017-05-26
Small artifacts in thin white borders


Gankro:
`gecko/layout/reftests/bugs/1156129-1.html` fails because a thin white curved border on a white background anti-aliases to include some non-white pixels.

https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/FkcLuDANTRKocplDYA828g/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1

glennw:
@Gankro It'd be worth checking if this still occurs once https://github.com/servo/webrender/pull/1975 makes its way into a Gecko build.

Gankro:
Looks like we removed the fails-if annotation some time ago

PR #2016, opened 2017-11-08
Smarter opaque/transparent splits and transformed AA

kvark:
Related to #1384, #1797, #1297
~~Also TODO:~~ Gecko try push https://treeherder.mozilla.org/#/jobs?repo=try&revision=e4e8ff793bcf938c2cb9ca414368f4bc808a66ad

~~Currently blocked on the `rotated-image.yaml` breakage, investigation is ongoing~~ - solved!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2016)


glennw:

---

*[webrender/res/prim_shared.glsl, line 772 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTfEf733nOieFs056N:-KyTfEf733nOieFs056O:b-83v4r2) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender/res/prim_shared.glsl#L772)):*
> ```GLSL
> Rectangle fetch_rectangle(int address) {
>     vec4 data[2] = fetch_from_resource_cache_2(address);
>     vec4 mask = vec4((int(data[1].x) & ivec4(1,2,4,8)) != ivec4(0));
> ```

Neat!

---

*[webrender/src/frame.rs, line 46 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTeindnW70UhpFwp_k:-KyTeindnW70UhpFwp_l:bp58l2t) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender/src/frame.rs#L46)):*
> ```Rust
>     pipeline_epochs: Vec<(PipelineId, Epoch)>,
>     replacements: Vec<(ClipId, ClipId)>,
>     opaque_parts: Vec<LayoutRect>,
> ```

Perhaps add a comment here saying that these are part of the context to avoid reallocations?

---

*[webrender/src/frame.rs, line 638 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTeyDD0kXTZm1KNbS2:-KyTeyDD0kXTZm1KNbS3:bsmbhn8) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender/src/frame.rs#L638)):*
> ```Rust
>     /// the rounded parts of the clip. This is pretty simple now, so has a lot of
>     /// potential for further optimizations.
>     fn try_to_add_rectangle_splitting_on_clip(
> ```

It might be worth only calling this function if the size of the rectangle is over some arbitrary threshold (maybe 256?).

---

*[webrender/src/util.rs, line 305 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTdIbMtFFbl1oqxiLz:-KyTdIbMtFFbl1oqxiM-:b-l9d7i7) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender/src/util.rs#L305)):*
> ```Rust
>     }
> 
>     fn split_rectangles(
> ```

This seems like it would still result in t-junctions if the radii are not the same at each corner? I might just be not reading this carefully enough...

---

*[webrender_api/src/display_item.rs, line 42 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTeDyLx6UHR4FW6fP4:-KyTeDyLx6UHR4FW6fP5:b7jos8f) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender_api/src/display_item.rs#L42)):*
> ```Rust
> }
> 
> bitflags! {
> ```

NOTE: Once we make the primitive / clip segments work internally, this shouldn't need to be in the public API, and we can move it from the api crate to the main crate. But we need it here for now since we're actually creating new primitives when splitting via the public API.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2016)*

kvark:
Review points are addressed now, and the try push looks green.

---

Review status: 20 of 21 files reviewed at latest revision, 5 unresolved discussions.

---

*[webrender/src/frame.rs, line 46 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTeindnW70UhpFwp_k:-KyTpge2A0HRKvHC57bb:b-7ql1td) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender/src/frame.rs#L46)):*

Perhaps add a comment here saying that these are part of the context to avoid reallocations?
</blockquote></details>

fixed

---

*[webrender/src/frame.rs, line 638 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTeyDD0kXTZm1KNbS2:-KyTphRudRkOmcHWLlfG:bvkxuh9) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender/src/frame.rs#L638)):*

It might be worth only calling this function if the size of the rectangle is over some arbitrary threshold (maybe 256?).
</blockquote></details>

good idea, added now

---

*[webrender/src/util.rs, line 305 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTdIbMtFFbl1oqxiLz:-KyTpmDIUwR3b59XIV3Q:bh7ow5) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender/src/util.rs#L305)):*

This seems like it would still result in t-junctions if the radii are not the same at each corner? I might just be not reading this carefully enough...
</blockquote></details>

this implementation is sub-optimal (still, even if much better than the old one) for the fact it works off the inner rectangle that is conservative: it takes the biggest of the radii affecting each side of the produced rectangle, so we just waste some space if one of the corners have lesser radii than the others.
My plan was to leave this for further optimization, given that the framework is set (framework of producing the list of opaque/transparent parts).

---

*[webrender_api/src/display_item.rs, line 42 at r1](https://reviewable.io:443/reviews/servo/webrender/2016#-KyTeDyLx6UHR4FW6fP4:-KyTqVVpT2V8iovTcMMt:b-e8513l) ([raw file](https://github.com/servo/webrender/blob/2a1b689242ad3d93cda89e93ed07001e71761153/webrender_api/src/display_item.rs#L42)):*

NOTE: Once we make the primitive / clip segments work internally, this shouldn't need to be in the public API, and we can move it from the api crate to the main crate. But we need it here for now since we're actually creating new primitives when splitting via the public API.
</blockquote></details>

good point!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2016#-:-KyTqqu90ghHZx_vZvXP:bbsv7pl)*

kvark:
Thanks, it is [all green](https://treeherder.mozilla.org/#/jobs?repo=try&revision=e4e8ff793bcf938c2cb9ca414368f4bc808a66ad).
@bors-servo r+

PR #2017, opened 2017-11-09
Fix hue-rotate and saturate filter

demo99:
The result of hue-rotate and saturate was wrong. I also create color matrix in vertex shader for those matrix type filters. 

After the fix, the gecko looks good: https://treeherder.mozilla.org/#/jobs?repo=try&revision=ca4753a3913013cdc80bd5cbba81f2213ce9ae51&selectedJob=143243269

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2017)

### DIFF
--- a/webrender/res/ps_blend.glsl
+++ b/webrender/res/ps_blend.glsl
@@ -8,6 +8,7 @@ varying vec3 vUv;
 flat varying vec4 vUvBounds;
 flat varying float vAmount;
 flat varying int vOp;
+flat varying mat4 vColorMat;
 
 #ifdef WR_VERTEX_SHADER
 void main(void) {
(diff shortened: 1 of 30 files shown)


glennw:
Thanks for the PR!

There seem to be several unexpected fails in filter tests in that try run that look related though?

Also, do the tests check for correctly working with pre-multiplied alpha? We should make sure we have some coverage for those too if not already.

demo99:
Oh, I think there is a bug after I move color matrix to vertex shader. I will fix it and add premultiplied alpha test as well.

glennw:
Looks good! Do we have / need a gecko try run for the updated PR?

demo99:
The try result is same:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=191652942d7af413180593f77738a71239d7d98d&selectedJob=143539747

There are 5 unexpected fails. The max difference is only 1~2. So I will use fuzz for those tests.

staktrace:
@demo99 This seems to have caused shader compilation failures on windows. See https://bugzilla.mozilla.org/show_bug.cgi?id=1415150#c7 - the try push that revealed it is at https://treeherder.mozilla.org/#/jobs?repo=try&revision=ffb402d6fefebf42d40c88889b97a86340fda4fe

mstange:
There's a variable with the name `c` at https://github.com/servo/webrender/blob/master/webrender/res/ps_blend.glsl#L54-L57

Maybe variable declarations aren't allowed within switch case branches without a nested scope, similar to C++?

kvark:
@mstange that's a good candidate! Filed https://github.com/servo/webrender/pull/2024

PR #2018, opened 2017-11-09
Fix some issues with scrollbars

mrobinson:
Scrollbars would scroll horizontally instead of sticking to the edge of
the frame and were also not visible at all on iframes. This change fixes
those issues by adding them to the containing reference frame. As part
of this change, iframe background colors are also attached to the
reference frame and only painted over the frame rect.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2018)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -34,12 +34,14 @@ use render_task::RenderTaskTree;
 use resource_cache::ResourceCache;
 use scene::ScenePipeline;
 use std::{mem, usize, f32, i32};
-use tiling::{CompositeOps, Frame};
-use tiling::{ContextIsolation, RenderTargetKind, StackingContextIndex};
-use tiling::{PrimitiveFlags, PrimitiveRunCmd, RenderPass};
-use tiling::{RenderTargetContext, ScrollbarPrimitive, StackingContext};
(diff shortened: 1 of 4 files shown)


webrender/src/frame_builder.rs
```diff
@@ -34,13 +34,18 @@ use render_task::RenderTaskTree;
...
-use tiling::{PrimitiveFlags, PrimitiveRunCmd, RenderPass};
-use tiling::{RenderTargetContext, ScrollbarPrimitive, StackingContext};
+use tiling::{CompositeOps, ContextIsolation, Frame, PrimitiveRunCmd, RenderPass};
+use tiling::{RenderTargetContext, RenderTargetKind, ScrollbarPrimitive, StackingContext};
+use tiling::StackingContextIndex;
 use util::{self, pack_as_float, RectHelpers, recycle_vec};
 use box_shadow::BLUR_SAMPLE_SCALE;
 
+#[derive(Debug, Copy, Clone)]
+pub enum PrimitiveFlags {
```
kvark: Isn't strictly related to this PR, but I think we should change this to something like `Option<ScrollBar>`

mrobinson: Okay. Sounds reasonable to me! The latest version of the PR has this change.

PR #2020, opened 2017-11-09
Remove ClipInfo

mrobinson:
Since Clip nodes always have clips in their ClipSources, the is_masking
member isn't necessary in ClipInfo. Since only one member remains,
we just store the ClipSourcesHandle and remove ClipInfo entirely. We
also rename ClipSources::is_masking() to ClipSources::has_clips()
because it is possible that a ClipSources may not be rendered to a mask
if the clip is rectangular. This can make the previous name of the
method misleading.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2020)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -245,7 +245,8 @@ impl ClipSources {
         }
     }
 
-    pub fn is_masking(&self) -> bool {
+    /// Whether or not this ClipSources has any clips (does any clipping).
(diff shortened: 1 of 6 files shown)


mrobinson:
@bors-servo retry
```
/Users/travis/.travis/job_stages: line 57: shell_session_update: command not found
```
This looks like an issue with the infrastructure so I'm retry.

PR #2021, opened 2017-11-09
Use premultiplied alpha blending for everything

mstange:
I've split this up into many small commits for easier reviewing. The only part I'm not entirely sure about is the brush primitive. Wrench reftests pass.

I'm going to push to try in a minute.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2021)

PR #2022, opened 2017-11-10
Type-level enforced alpha pre-multiplication

kvark:
Seals the deal with alpha pre-multiplication :)
Based on #2021 (merged now)
~~Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9c222f95e2788cc5444544054539020f1e6f55c2~~

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2022)


webrender/src/prim_store.rs
```diff
@@ -625,10 +625,12 @@ impl TextRunPrimitiveCpu {
...
 
     fn write_gpu_blocks(&self, request: &mut GpuDataRequest) {
-        let mut bg_color = self.font.bg_color;
-        bg_color.a = 0xFF; // not actually affecting the shader
+        let bg_color = ColorF::from(self.font.bg_color);
         request.push(self.font.color);
-        request.push(ColorF::from(bg_color).premultiplied());
+        // this is the only case where we need to provide plain color to GPU
+        request.extend_from_slice(&[
+            GpuBlockData { data: [bg_color.r, bg_color.g, bg_color.b, bg_color.a] }
```
mstange: I think it would be a good idea to use 1.0 or 0.0 instead of `bg_color.a` here, just to make it clear that we don't care about bg_color's alpha at all.

lsalzman:
I am not sure it is safe to store a premultiplied color directly in FontInstance. Whether it is safe with respect to all the font backends would need to be investigated. Plus it increases the memory usage of FontInstance vs. ColorU.

kvark:
@lsalzman I refactored the PR largely, backing off the premultiplication to a later stage, so that `FontInstance` is kept unchanged, and no un-premultiplication is needed. This may not be our end game, but it's a solid incremental (non-breaking!) step in this direction.

mstange:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=47b33ce785750981dc2a104b88ca4f48c2703f67

webrender/src/frame_builder.rs
```diff
@@ -1171,17 +1179,19 @@ impl FrameBuilder {
...
+            match picture.kind {
+                // Only run real blurs here (fast path zero blurs are handled above).
+                PictureKind::TextShadow { blur_radius, .. } if blur_radius > 0.0 => {
+                    picture.add_primitive(
+                        prim_index,
+                        &original_rect,
+                        clip_and_scroll,
+                    );
+                }
+                _ => {}
```
mstange: This is a bit inconsistent with the previous hunk, where you have a `PictureKind::BoxShadow { .. } => unreachable!()` case, but it doesn't really matter.

webrender_api/src/color.rs
```diff
@@ -2,8 +2,32 @@
...
+    pub g: f32,
+    pub b: f32,
+    pub a: f32,
+}
+
+impl PremultipliedColorF {
+    ///
+    pub const BLACK_OPAQUE: Self = PremultipliedColorF { r: 0.0, g: 0.0, b: 0.0, a: 1.0 };
+    ///
+    pub const BLACK_TRANSPARENT: Self = PremultipliedColorF { r: 0.0, g: 0.0, b: 0.0, a: 0.0 };
```
mstange: With premultiplied colors there's no difference between "transparent black" and "transparent anything-else". So I'd rename these two constants to `BLACK` and `TRANSPARENT`.

kvark:
Thanks @mstange ! Review comments are addressed now.
@bors-servo r=mstange

PR #2024, opened 2017-11-10
Work around GLSL issue in ps_blend

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1415150#c7

Relevant wiki entry: https://github.com/servo/webrender/wiki/Driver-issues#1285---declarations-in-switch-cases

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2024)


staktrace:
Confirmed that this fixed the problem, thanks!

PR #2028, opened 2017-11-11
move FontInstance out of webrender_api so it is easier to extend

lsalzman:
It's a bit of a bother to have to update webrender_api every time we do some extending of FontInstance. At this point, all interaction from the API should be done through FontInstanceKey, so the actual internal structure of FontInstance should remain opaque. As such, this just moves FontInstance into webrender, and updates the only actual user of it (get_glyph_dimensions) to use a FontInstanceKey instead.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2028)


glennw:
Reviewed 15 of 15 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2028)*

Issue #1943, opened 2017-10-25
The horizontal blur pass sometimes seems to blur in garbage from the side


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1411693

[Testcase](https://bug1411693.bmoattachments.org/attachment.cgi?id=8922041) (created by Darkspirit)

When hovering over the element on the page, the box-shadow sometimes has extra shadows on the right edges. Here are two screenshots with and without the glitch, and with `gfx.webrender.debug.render-targets` enabled:

good frame:
<img width="1193" alt="box-shadow-good-frame" src="https://user-images.githubusercontent.com/961291/32019087-107010b4-b99a-11e7-8bab-76b8fc234cce.png">

bad frame:
<img width="1193" alt="box-shadow-bad-frame" src="https://user-images.githubusercontent.com/961291/32019096-1585206c-b99a-11e7-9624-9135edf6e153.png">

You can see that in the bad frame, the render target in the middle has a box-shadow whose right edge is a bit brighter than in the good frame. It looks like the right edge got some stuff blurred in from beyond the right edge of the rectangle that is supposed to be sampling from.

glennw:
I can reproduce this with a static box shadow that has a fractional blur radius. There must be somewhere in the code that ends up with a rounding error when a fractional blur radius is supplied. I'll work on a fix today.

glennw:
Note to self: the blur radius is correctly rounded to device pixels, but the CSS space blur radius is retained as a float. Probably related to that.

Darkspirit:
This got better with the latest webrender update. (Thank you!) Those unwanted things on the right and bottom are gone and only a white flickering is left.

Commits between the previous and latest webrender update:
https://github.com/servo/webrender/compare/4b8493d6bdc64d2d83202ac15b06b0d4b14c6e76...c0194de78ce26106a8497484dc8d159069e3a482

Gankro:
@glennw any news on this?

glennw:
@Gankro Not specifically, although the recent work I've done on box-shadows should make this easier to fix / debug. It's also possible that it may have been fixed, but I'm guessing not, from your comment here?

Gankro:
I was just following up on it since I vaguely recalled you mentioning this was going to be fixed soon. Just triaging some fuzziness that has been attributed to this.

Darkspirit:
[...]

Issue #1496, opened 2017-07-18
Isolation failures in gecko reftests


Gankro:
Accidentally noticed that the raw logs of recent gecko try builds include these errors (on a passing test).

Try build: https://treeherder.mozilla.org/#/jobs?repo=try&revision=79738b756f0e08958e493f3020b254817a97b94a&selectedJob=115176289

Relevant raw log output:

```
[task 2017-07-18T10:31:45.277463Z] 10:31:45     INFO - REFTEST TEST-START | file:///home/worker/workspace/build/tests/reftest/tests/layout/reftests/async-scrolling/background-blend-mode-1.html == file:///home/worker/workspace/build/tests/reftest/tests/layout/reftests/async-scrolling/background-blend-mode-1-ref.html
[task 2017-07-18T10:31:45.278803Z] 10:31:45     INFO - REFTEST INFO | RESTORE PREFERENCE pref(apz.disable_for_scroll_linked_effects,false)
[task 2017-07-18T10:31:45.280233Z] 10:31:45     INFO - REFTEST TEST-LOAD | file:///home/worker/workspace/build/tests/reftest/tests/layout/reftests/async-scrolling/background-blend-mode-1.html | 1461 / 1945 (75%)
[task 2017-07-18T10:31:45.346178Z] 10:31:45     INFO - ERROR:webrender::frame_builder: Isolation conflict detected on StackingContextIndex(6): Full
[task 2017-07-18T10:31:45.350652Z] 10:31:45     INFO - ERROR:webrender::frame_builder: Isolation conflict detected on StackingContextIndex(6): Full
[task 2017-07-18T10:31:46.318003Z] 10:31:46     INFO - ERROR:webrender::frame_builder: Isolation conflict detected on StackingContextIndex(6): Full
[task 2017-07-18T10:31:46.318122Z] 10:31:46     INFO - REFTEST TEST-LOAD | file:///home/worker/workspace/build/tests/reftest/tests/layout/reftests/async-scrolling/background-blend-mode-1-ref.html | 1461 / 1945 (75%)
[task 2017-07-18T10:31:46.436053Z] 10:31:46     INFO - ERROR:webrender::frame_builder: Isolation conflict detected on StackingContextIndex(6): Full
[task 2017-07-18T10:31:46.436187Z] 10:31:46     INFO - ERROR:webrender::frame_builder: Isolation conflict detected on StackingContextIndex(6): Full
[...]

Issue #1957, opened 2017-10-27
Image masks don't create an intermediate surface to group the masked contents


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1412375

[Testcase](https://bug1412375.bmoattachments.org/attachment.cgi?id=8922912)

Expected results vs actual results:
<img width="419" alt="screen shot 2017-10-27 at 2 13 04 pm" src="https://user-images.githubusercontent.com/961291/32118795-fdca785e-bb20-11e7-8e4a-e0cd14017982.png">

It might be worth having a different API for grouping masks vs non-grouping clips. For example, a grouping mask could be a property of a stacking context instead of a property of a clip node.

mstange:
> For example, a grouping mask could be a property of a stacking context

It could even be a filter, just like opacity is.

glennw:
This is a timely bug report - as I'm working on optimizing the clip stack bugs / issues in https://github.com/servo/webrender/issues/1817 so I'll try to cover this at the same time, since they are very related.

Gankro:
@glennw in #2443 you said

> R1 failure [2] - I haven't confirmed what is happening in this one, yet. Looking visually, it appears that the image mask is being applied to both the primitive and the resulting stacking context picture. If that is the case, the fix here would be to update Gecko once this lands to supply the image-mask clip chain only to the primitive and/or the stacking context picture, rather than both. As a side-effect of this proposed change, #1957 gets fixed. It may also be related to #2388.

Is this still accurate? Should fixing this bug "just" be a matter of updating gecko now?

Gankro:
Also we don't currently seem to emit a StackingContext in response to a mask; we just PushClip. Are you suggesting we should push an otherwise no-op stacking context and only apply the clip to that, and it will Just Work?

Gankro:
yaml that "should" work (I think):

```yaml
---
root:
  items:
    - type: rect
      bounds: [25, 25, 100, 100]
      color: red
    - type: clip
      bounds: [0, 0, 200, 200]
      id: 0
      image-mask:
          image: "transparent-white.png"
          rect: [0, 0, 200, 200]
          repeat: false
    - type: stacking-context
      bounds: [0, 0, 200, 200]
      clip-and-scroll: 0
      items:
      - type: rect
        bounds: [0, 0, 100, 100]
        color: blue
      - type: rect
        bounds: [50, 50, 100, 100]
        color: green
```

[...]

Issue #1718, opened 2017-09-18
ERROR:webrender::frame_builder: Isolation conflict detected on StackingContextIndex(2): Full


hgallagher1993:
This error printed out literally thousands of times to my terminal after running ``./mach run -r http://www.luzern.co/``. The actual page seemed to load fine except for this printing out pretty much constantly.

glennw:
Fixed by #2031. We can open new bugs for any isolation specific issues.

PR #2031, opened 2017-11-13
Introduce PicturePrimitive::Image.

glennw:
Instead of constructing the render task tree each render (from the
stacking context tree), construct a compositing tree during scene
building. We create a Picture primitive whenever we *may* need to
draw to an intermediate surface (i.e. filters, mix-blend-mode,
3d rendering contexts, frame output).

During culling, we can then make decisions on whether a Picture
does get drawn to an intermediate surface, or drawn directly
into the parent target (e.g. if perspective is present, or
if a child stacking context needs isolation due to mix-blend-mode).

This opens up fixes for a number of issues, such as avoiding
rebuilding the scene in the presence of property bindings, proper
pixel snapping and clip mask handling for composited stacking
contexts, fixing the context isolation bugs when multiple effects
are present in addition to 3D rendering contexts.

A consequence of this change is that plane splitting is handled
later in the frame, during batch generation. This simplfies a lot
of the plane splitting code.

* Remove StackingContext structures after scene is created. Instead, they
  are only used during scene building, and then we drop the concept of a
  stacking context completely and rely on the Picture tree only.

* Move output frame handling from build frame to build scene - so we
  can allocate Pictures based on this.

* Pass clip node ID to push_stacking_context - will be used to handle #1957.

* Move plane splitting to batch generation phase.

* Remove (as no longer used):
    build() step of PicturePrimitive.
    HardwareCompositeOp.
    notify_waiting_for_root_stacking_context().
    RenderTask::inflate (handled implicitly in Picture).
    AlphaRenderItem - We now just walk the primitive runs, saving memory allocations and duplication.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2031)


glennw:
@kvark I am *so sorry* about the size of this patch. I tried to do it incrementally but failed.

If this is too difficult to review, perhaps one way could be to set up a Vidyo call and we could
pair review it together so I can walk you through the changes?

There are a few really messy parts to the code (specifically related to prepare_prim_for_render)
that are mostly hacks around the borrow checker. I need to improve those before this is ready - and then it will probably need a second follow up to restructure the prim store code to be a bit tidier.

All the wrench / servo WPT tests pass.

Remaining work before we can merge this:
- [ ] There are 3 test failures in Gecko - all panics - I think they are all the same bug.
- [ ] Tidy up the prepare_prim_runs / prepare_prim_for_render hacks.

glennw:
Oh, also the code in ```prepare_prim_runs``` is needlessly complicated and messy - that needs to be fixed up too.

webrender/src/frame_builder.rs
```diff
@@ -302,59 +276,283 @@ impl FrameBuilder {
...
-
     pub fn push_stacking_context(
         &mut self,
-        reference_frame_offset: &LayerVector2D,
         pipeline_id: PipelineId,
         composite_ops: CompositeOps,
         transform_style: TransformStyle,
         is_backface_visible: bool,
         is_pipeline_root: bool,
+        clip_and_scroll: ClipAndScrollInfo,
```
mrobinson: So one potential issue with this is that different items in a stacking context can have different clip nodes and even different scroll nodes.

glennw: Yep, that's expected and handled - items in a picture are primitive runs with their own clip and scroll. This parameter just identifies the current clip / scroll nodes when the stacking context is pushed - it's used for plane splitting and a few other things.

glennw: To expand on that, it effectively identifies a root reference frame for that stacking context.

mrobinson:
@glennw I'd love to join the call if possible. This intersects a bit with stuff that I'm doing and understanding it would be very helpful to me.

kvark:
This is incredible work! :tada: :yum: :tada: 
I took a low-level pass through it, like the plane splitting reform so far :)
Going to take a higher-level look now.

---

Review status: all files reviewed at latest revision, 27 unresolved discussions.

---

*[webrender/src/frame.rs, line 1177 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq2DSb8TeqQRSnXHUC:-Kyq2DSb8TeqQRSnXHUD:bxazx2d) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame.rs#L1177)):*
> ```Rust
>                 reference_frame_id,
>                 scroll_frame_id,
>                 output_pipelines,
> ```

can we just make it a part of the `roller`?

---

*[webrender/src/frame_builder.rs, line 48 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq2SmnJUL6NBZAKeIZ:-Kyq2Smo2tv3uAQUNZgr:bbye412) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L48)):*
> ```Rust
> /// during creation of the scene. These structures are
> /// not persisted after the initial scene build.
> pub struct StackingContext {
> ```

given that it became a more localized implementation detail, we should re-evaluate the `pub` qualifiers on it and its members

---

*[webrender/src/frame_builder.rs, line 300 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq7zNQlgbzjDz_Hu9q:-Kyq7zNR-1pddIYoz89j:bl4edv4) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L300)):*
> ```Rust
>         // to correctly handle some CSS cases (see #1957).
>         let max_clip = LayerRect::new(
>             LayerPoint::new(-1000000.0, -1000000.0),
> ```

nit: could re-use the constant defined in `clip.rs`

---

*[webrender/src/frame_builder.rs, line 308 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq8bO6LB7UjJsg9cXN:-Kyq8bO6LB7UjJsg9cXO:bz6j65n) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L308)):*
> ```Rust
>             // Should be no pictures at all if the stack is empty...
>             debug_assert!(self.prim_store.cpu_pictures.is_empty());
>             debug_assert!(transform_style == TransformStyle::Flat);
> ```

nit: use `debug_assert_eq!`

---

*[webrender/src/frame_builder.rs, line 352 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqD7b695Je5bs8LwBU:-KyqD7b695Je5bs8LwBV:beq3kqy) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L352)):*
> ```Rust
>                     // TODO(gw): Handle isolation conflicts here by
>                     //           creating a new picture.
>                     *composite_mode = PictureCompositeMode::Blit;
> ```

did we lose the check for an isolation conflict? i.e. what if `composite_mode` was not trivial to begin with

---

*[webrender/src/frame_builder.rs, line 364 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqDOfQ-EI-HY0Lfg5Y:-KyqDOfQ-EI-HY0Lfg5Z:bggqp95) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L364)):*
> ```Rust
>         // which determines if we *might* need to draw this on
>         // an intermediate surface for plane splitting purposes.
>         let parent_transform_style = match self.sc_stack.last() {
> ```

`sc_stack` is guaranteed to have 3 elements here at least, so no need for a match

---

*[webrender/src/frame_builder.rs, line 387 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqDv35p-X-7IiieCEU:-KyqDv35p-X-7IiieCEV:b7dvpip) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L387)):*
> ```Rust
>             self.sc_stack
>                 .iter()
>                 .rev()
> ```

nit: could be `rfind`

---

*[webrender/src/frame_builder.rs, line 1550 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqMUbC9gb2Z1cnLvwk:-KyqMUbC9gb2Z1cnLvwl:be9iytn) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L1550)):*
> ```Rust
>         );
> 
>         let mut child_tasks = Vec::new();
> ```

I looked into refining `prepare_prim_runs` a bit, especially w.r.t. `child_tasks` propagation and the borrow checker dance... It doesn't appear obvious on how to approach this without changing a lot of code. Minimal we can do is at least concealing the borrowck dance within the `prepare_prim_runs` internals (by passing the `SpecificPrimitiveIndex` to it instead of the runs), but that implementation still involves a bunch of accesses to the `cpu_picture` in various places, and I don't think is worth going for now. Better rethink this part of code as a follow-up.

---

*[webrender/src/frame_builder.rs, line 1615 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqNGSv2905h2eXfdQ4:-KyqNGSv2905h2eXfdQ5:b-klh4r4) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L1615)):*
> ```Rust
> 
>         self.prim_store
>             .cpu_pictures[0]
> ```

haha, nice implementation :) Perhaps, we should rename the function now for clarity?

---

*[webrender/src/picture.rs, line 31 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqYPbqELcVVaAyvi4r:-KyqYPbs2TFkL8RI_h0h:b-8e2kjc) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L31)):*
> ```Rust
> pub enum PictureCompositeMode {
>     /// Don't composite - just draw directly on parent surface.
>     None,
> ```

I think we should instead operate on `Option<PictureCompositeMode>` instead of having `None`, for clarity. The implementation parts using it would be cleaner a bit.

---

*[webrender/src/picture.rs, line 33 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqYqPGHdUuflJaV3p3:-KyqYqPGHdUuflJaV3p4:b-guf1b4) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L33)):*
> ```Rust
>     None,
>     /// Apply CSS mix-blend-mode effect.
>     MixBlendMode(MixBlendMode),
> ```

nit: rename to `MixBlend(MixBlendMode)`?

---

*[webrender/src/picture.rs, line 188 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqZ6Z9xd5FFUIEk3Zx:-KyqZ6Z9xd5FFUIEk3Zy:bovplnt) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L188)):*
> ```Rust
>     pub fn update_local_rect(&mut self,
>         local_rect: LayerRect,
>         local_content_rect: LayerRect,
> ```

we should make these parameters more clear

---

*[webrender/src/picture.rs, line 326 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqZmgqYVQ_5d6__Pmb:-KyqZmgqYVQ_5d6__Pmc:b-q55i3e) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L326)):*
> ```Rust
>                     PictureCompositeMode::None => {
>                         parent_tasks.extend(child_tasks);
>                         self.render_task_id = None;
> ```

the implementation could be simplified a bit if we return here, see https://github.com/kvark/webrender/commit/3369e608ed0506fbbef51ef91a1c95ac5d526c5b

---

*[webrender/src/picture.rs, line 380 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq_EJ664VUG4tZz7C5:-Kyq_EJ664VUG4tZz7C6:b-fl7z63) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L380)):*
> ```Rust
>                 let blur_radius = device_length(blur_radius, prim_context.device_pixel_ratio);
> 
>                 // TODO(gw): Rounding the content rect here to device pixels is not
> ```

any reason we can't just `ceil` here then?

---

*[webrender/src/picture.rs, line 389 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq_MN_nPgVD1aGn4pl:-Kyq_MNagCulQxxcjqqC:b5jqdg) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L389)):*
> ```Rust
>                 let cache_height =
>                     (content_rect.size.height * prim_context.device_pixel_ratio).round() as i32;
>                 let cache_size = DeviceIntSize::new(cache_width, cache_height);
> ```

nit: could use euclid's `TypedSize2D` all the way (it supports rounding, scaling and casts)

---

*[webrender/src/prim_store.rs, line 1070 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqNlyliSgb4h8SpLTJ:-KyqNlyliSgb4h8SpLTK:b-v22nbq) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1070)):*
> ```Rust
>         gpu_cache: &mut GpuCache,
>         render_tasks: &mut RenderTaskTree,
>         child_tasks: Vec<RenderTaskId>,
> ```

this is something I'd like to rethink

---

*[webrender/src/prim_store.rs, line 1365 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqOEEXjmXeng2O-YNB:-KyqOEEXjmXeng2O-YNC:bxpd2b0) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1365)):*
> ```Rust
>                 Some(local_rect) => local_rect,
>                 None => {
>                     if perform_culling {
> ```

nit:
```
None if perform_culling => return None,
None => LayerRect::zero(),
}
```

---

*[webrender/src/prim_store.rs, line 1406 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqOV87G9RaofFgERrO:-KyqOV87G9RaofFgERrP:b-9oky0k) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1406)):*
> ```Rust
>             clip_store,
>             parent_tasks,
>         ) {
> ```

`&& perform_culling`?

---

*[webrender/src/prim_store.rs, line 1440 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqOXq-Anb3KvjsVUym:-KyqOXq-Anb3KvjsVUyn:b1c88eh) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1440)):*
> ```Rust
>         profile_counters: &mut FrameProfileCounters,
>         reference_frame_id: Option<ClipId>,
>     ) -> (LayerRect, LayerRect) {
> ```

The return types are begging to use different type-level spaces. Btw, what is the moving space here exactly? If not types, we should probably have a struct returned here, so that the output fields are named.

---

*[webrender/src/prim_store.rs, line 1491 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqVEJyPMu3VGdG5dAg:-KyqVEJzyZIhW4DPDBgA:b-qvijd) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1491)):*
> ```Rust
>                             let m = m0.pre_mul(&m1);
> 
>                             let vertices = [
> ```

looks like this repeated block could be moved into a helper function

---

*[webrender/src/prim_store.rs, line 1503 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqVWTSyOmVzgknbV4C:-KyqVWTTBopO35DjWeLV:b-i6hnp1) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1503)):*
> ```Rust
>                             let mut y1 = vertices[0].y;
> 
>                             for v in &vertices {
> ```

given that `TypedPoint2D` implements `min/max`, we can operate on it directly.
Also, no need to start with index `0` here

---

*[webrender/src/prim_store.rs, line 1512 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqWMCUqXUyX8Z_FBq6:-KyqWMCUqXUyX8Z_FBq7:bc5wyfj) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1512)):*
> ```Rust
>                             let bounds = LayerRect::new(LayerPoint::new(x0, y0), LayerSize::new(x1 - x0, y1 - y0));
> 
>                             local_rect_in_original_space = local_rect_in_original_space.union(&bounds);
> ```

We need to discuss these rectangles... I'm worried that we can hit the same issues I had with LCCR, given that this projection code ignores produced Z coordinates, so it's not reversible (i.e. transforming this local rectangle back in to the other space would not necessarily give us matching results from what we'd expect).

---

*[webrender/src/render_task.rs, line 156 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqaC3X0SO18CLdcfu-:-KyqaC3X0SO18CLdcfu0:b-n0p1fu) ([raw file](https://github.com/servo/webrender/blob/8a39cf24f493e894a66c2465dd310a2b2923e558/webrender/src/render_task.rs#L156)):*
> ```Rust
> 
> #[derive(Debug)]
> pub enum AlphaRenderItem {
> ```

niice!

---

*[webrender/src/render_task.rs, line 299 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq_qHVMBT-36vYF9QB:-Kyq_qHVMBT-36vYF9QC:b-ctmctl) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/render_task.rs#L299)):*
> ```Rust
>         prim_index: PrimitiveIndex,
>         frame_output_pipeline_id: Option<PipelineId>,
>         children: Vec<RenderTaskId>,
> ```

minor: I don't think we should be passing `children` in here, given that the field has a well-defined default (`Vec::new()`), can be accessed publicly (which can be convenient in places), and the other fields (like `clear_mode`) are not provided right off the bad.

---

*[webrender/src/tiling.rs, line 345 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqaiYaK2Ei3TfG372A:-KyqaiYaK2Ei3TfG372B:b-6bkezu) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/tiling.rs#L345)):*
> ```Rust
> // in that picture are being drawn into the same target.
> fn add_to_batch(
>     clip_id: ClipScrollNodeIndex,
> ```

it sounds like we could have a helper context struct here, similar to `FlattenContext`

---

*[webrender/src/tiling.rs, line 913 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqbbM772vCJLPI3NG3:-KyqbbM772vCJLPI3NG4:brl7slx) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/tiling.rs#L913)):*
> ```Rust
>         let task_address = render_tasks.get_task_address(task_id);
> 
>         // TODO(gw): We could avoid creating the splitter here if we know
> ```

Empty `BspSplitter` doesn't use the heap and is cheap to produce. I think we are safe creating it every time here and avoiding an `Option<>` around it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2031#-:-KyqcZ0F9Olwr1Gw5Ywt:ba94rxv)*

kvark:
It appears to be fairly easy to hide the `StackingContext` from outside of frame builder: https://github.com/kvark/webrender/commit/0cfcfa11156f78802874fa78a7d7e564bd5703a9

glennw:
Review status: 5 of 13 files reviewed at latest revision, 27 unresolved discussions.

---

*[webrender/src/frame.rs, line 1177 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq2DSb8TeqQRSnXHUC:-Kys2MUaSbCKNAwGwYSm:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame.rs#L1177)):*

can we just make it a part of the `roller`?
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 48 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq2SmnJUL6NBZAKeIZ:-Kys2OR_DQXhWjB9qUvP:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L48)):*

given that it became a more localized implementation detail, we should re-evaluate the `pub` qualifiers on it and its members
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 300 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq7zNQlgbzjDz_Hu9q:-Kys2Pf75za9nlhBemcr:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L300)):*

nit: could re-use the constant defined in `clip.rs`
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 308 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq8bO6LB7UjJsg9cXN:-Kys2QvuOFRb7DrARI1K:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L308)):*

nit: use `debug_assert_eq!`
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 352 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqD7b695Je5bs8LwBU:-Kys2StWlq6QDbox0VEX:b-hn69qb) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L352)):*

did we lose the check for an isolation conflict? i.e. what if `composite_mode` was not trivial to begin with
</blockquote></details>

We now only set it to blit if it was previously none. Otherwise, it's already isolated. I think the fix here is why we are seeing some new unexpected PASSes in the Gecko try runs.

---

*[webrender/src/frame_builder.rs, line 364 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqDOfQ-EI-HY0Lfg5Y:-Kys2d--uLNqKpBbNGtw:b-nlyqv3) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L364)):*

`sc_stack` is guaranteed to have 3 elements here at least, so no need for a match
</blockquote></details>

The picture stack is, but not so for the stacking context stack here.

---

*[webrender/src/frame_builder.rs, line 387 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqDv35p-X-7IiieCEU:-Kys2h8Lzbw5Sx3C5L7J:b-9drk48) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L387)):*

nit: could be `rfind`
</blockquote></details>

rfind is not stabilized yet.

---

*[webrender/src/frame_builder.rs, line 1615 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqNGSv2905h2eXfdQ4:-Kys2mKHt8fJ3rvFeY2T:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L1615)):*

haha, nice implementation :) Perhaps, we should rename the function now for clarity?
</blockquote></details>

Done.

---

*[webrender/src/picture.rs, line 31 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqYPbqELcVVaAyvi4r:-Kys2nOZb5fksPDVlgjS:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L31)):*

I think we should instead operate on `Option<PictureCompositeMode>` instead of having `None`, for clarity. The implementation parts using it would be cleaner a bit.
</blockquote></details>

Done.

---

*[webrender/src/picture.rs, line 33 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqYqPGHdUuflJaV3p3:-Kys2oD6UrX0lQ8eEobs:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L33)):*

nit: rename to `MixBlend(MixBlendMode)`?
</blockquote></details>

Done.

---

*[webrender/src/picture.rs, line 380 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyq_EJ664VUG4tZz7C5:-Kys2rgNhrg959NW1aVD:b-gx9d1n) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L380)):*

any reason we can't just `ceil` here then?
</blockquote></details>

It exposes some other fractional rounding bugs in tests when we do that. Once this stuff all lands, and we can sort out the pixel snapping + picture primitive compositing, we should be able to change this to ceil().

---

*[webrender/src/prim_store.rs, line 1365 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqOEEXjmXeng2O-YNB:-Kys31zlQy9e6dVKWn8q:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1365)):*

nit:
```
None if perform_culling => return None,
None => LayerRect::zero(),
}
```
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1406 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqOV87G9RaofFgERrO:-Kys32fxEgZfkPbnhYoi:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1406)):*

`&& perform_culling`?
</blockquote></details>

Done.

---

*[webrender/src/tiling.rs, line 913 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqbbM772vCJLPI3NG3:-Kys384ZJezYeeok72mj:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/tiling.rs#L913)):*

Empty `BspSplitter` doesn't use the heap and is cheap to produce. I think we are safe creating it every time here and avoiding an `Option<>` around it
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2031)*

glennw:
Update:

There are a few more test failures than I realized - I had assumed they were fuzziness errors from other PRs, but they seem related to this.

I've fixed most of the test failures, but there are a couple that are proving difficult to fix (as I can't repro them locally).

Will continue working on those tomorrow, and then fix the remaining review comments (which should just be tidying up the parts of the code mentioned above).

glennw:
Most recent try has:

1x intermittent panic - I have a local repro for this now.
2x unexpected FAIL - fuzzy but they shouldn't be fuzzy - need to investigate (can't repro locally yet).
4x new PASS - from isolation fixes.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=a1fe8327df8ae64dd80965d79c83d49560c204f2

kvark:
Reviewed 1 of 3 files at r2, 6 of 7 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[webrender/src/clip_scroll_node.rs, line 312 at r4](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyv7xR-JI_QbEBNd8SJ:-Kyv7xR0LTj7fPkSVygj:bjhdwtn) ([raw file](https://github.com/servo/webrender/blob/d806b6937a2e6ea9ea9a33dfd489ca3dd1d22c7e/webrender/src/clip_scroll_node.rs#L312)):*
> ```Rust
>             None => {
>                 state.combined_outer_clip_bounds = DeviceIntRect::zero();
>                 self.combined_clip_outer_bounds = DeviceIntRect::zero();
> ```

why is this needed now?

---

*[webrender/src/frame_builder.rs, line 387 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqDv35p-X-7IiieCEU:-Kyv7Uee8wOQcZ3cXMWr:b-ktwvrh) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/frame_builder.rs#L387)):*

rfind is not stabilized yet.
</blockquote></details>

oops, you are right

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2031)*

glennw:
Latest try run here:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=3f0191530af3351480fd261c2f6e88c2b7c8f036

We have:

New FAIL:
```boxshadow-skiprect.html``` - This is caused by https://github.com/servo/webrender/issues/1943. The allocation patterns of the texture cache determine whether that bug is hit or not. I'd like to mark this as an expected FAIL for now, until we resolve #1943 properly.
```same-filter.html``` - One of the tests results in a blob image (blurs done by SVG in Gecko), and one draws with the native WR blur filters. Given that, I think we can mark this as fuzzy.

New PASS:
```mix-blend-mode-and-filter.html```
```outline-and-3d-transform-2.html```
```preserve3d-6a.html```
```backface-visibility-2.html```

The new passes are expected as a result of the changes in this patch.

@staktrace I still need to tidy up a few functions in the code, but are you happy to have this merged once that is done, based on the results above?

glennw:
(All the wrench / servo WPT tests pass).

glennw:
Review status: 12 of 13 files reviewed at latest revision, 15 unresolved discussions.

---

*[webrender/src/clip_scroll_node.rs, line 312 at r4](https://reviewable.io:443/reviews/servo/webrender/2031#-Kyv7xR-JI_QbEBNd8SJ:-KyxBVTUJ3CS7EuT4W3Q:b-dy5rwe) ([raw file](https://github.com/servo/webrender/blob/d806b6937a2e6ea9ea9a33dfd489ca3dd1d22c7e/webrender/src/clip_scroll_node.rs#L312)):*

why is this needed now?
</blockquote></details>

This handles the case where a singular matrix is encountered at the bottom of the clip-scroll tree.

---

*[webrender/src/picture.rs, line 188 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqZ6Z9xd5FFUIEk3Zx:-KyxBhEebLpIjisNOf-m:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/picture.rs#L188)):*

we should make these parameters more clear
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1070 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqNlyliSgb4h8SpLTJ:-KyxBlm5GtnK6hV4GR1f:b-inuzld) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1070)):*

this is something I'd like to rethink
</blockquote></details>

Agreed - I'd to revisit this once the dust settles and think if we can restructure this better.

---

*[webrender/src/prim_store.rs, line 1440 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqOXq-Anb3KvjsVUym:-KyxBrEzYcqswxnI0iio:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1440)):*

The return types are begging to use different type-level spaces. Btw, what is the moving space here exactly? If not types, we should probably have a struct returned here, so that the output fields are named.
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1491 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqVEJyPMu3VGdG5dAg:-KyxBsJDl9_zAHGn_sWt:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1491)):*

looks like this repeated block could be moved into a helper function
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1512 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqWMCUqXUyX8Z_FBq6:-KyxBv-fFmWqM7vQFt69:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/prim_store.rs#L1512)):*

We need to discuss these rectangles... I'm worried that we can hit the same issues I had with LCCR, given that this projection code ignores produced Z coordinates, so it's not reversible (i.e. transforming this local rectangle back in to the other space would not necessarily give us matching results from what we'd expect).
</blockquote></details>

Done.

---

*[webrender/src/tiling.rs, line 345 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqaiYaK2Ei3TfG372A:-KyxC2nG_itQmAad9aHs:busvlot) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/tiling.rs#L345)):*

it sounds like we could have a helper context struct here, similar to `FlattenContext`
</blockquote></details>

Yes - I think that makes sense. Happy if we do that as a follow up?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2031)*

glennw:
@kvark OK, added a couple of new commits that contain tidy ups. I *think* that addresses all your comments (though there is certainly some further clean ups that can be made to the code structure, hopefully as a follow up).

Pending try run here with those tidy ups, which should be the same as the previous result:

https://hg.mozilla.org/try/rev/6e30ae734328485e6f209dcba7b04f3e03f05683
https://treeherder.mozilla.org/#/jobs?repo=try&revision=6e30ae734328485e6f209dcba7b04f3e03f05683

If you're happy with those follow up commits, and @staktrace is happy with the reftest changes mentioned above, this should be ready to go (although I'll squash it before final merge).

glennw:
When landed, this fixes https://github.com/servo/webrender/issues/1496 and https://github.com/servo/webrender/issues/1718.

staktrace:
@glennw I'm happy with the try results and your analysis of it, thanks!

glennw:
Hmmm, just hit an assert testing in Servo that *might* be related to this patch. Investigating now...

glennw:
Ah, sometimes the prepare_prim logic can early out during scrolling, which could result in a primitive being considered visible (if it was the previous frame). For now, just walk the primitive array at the start of the frame and clear the screen rect. It's not ideal, but it should be safe - and we can remove it later on when it's not needed. Just going to do a final try run now to make sure that didn't break anything...

glennw:
OK, I can still hit an assert on one specific website (the w3 conf site) with this patch. I'll look into it first thing in the morning and then this should *finally* be ready to merge.

glennw:
I think I know the cause of it, it will take a little while to verify but the fix should be very simple.

glennw:
OK, I believe that assert is resolved now - the check for clip node being off-screen wasn't respecting the ```perform_culling``` flag.

I've done a fair amount of manual testing in Gecko + Servo now, and can't see or reproduce any more issues.

Pending try run with the most recent assert fix:

https://hg.mozilla.org/try/rev/86fcc58e7103e13c9a05473f4c13f4b6d6f17338

I've squashed all the fixup commits into one, but left the commit from @kvark as a separate commit.

So I *think* this is ready to be merged now, once @kvark is happy with the fixup commit.

kvark:
Thanks @glennw!
I really appreciate the attention to testing you put with this PR, it properly matches the scale of changes here.

---

Reviewed 1 of 1 files at r5, 3 of 3 files at r6.

---

*[webrender/src/tiling.rs, line 345 at r1](https://reviewable.io:443/reviews/servo/webrender/2031#-KyqaiYaK2Ei3TfG372A:-Kz-RoPt8VSEo9PGsFgr:b4vx5rp) ([raw file](https://github.com/servo/webrender/blob/6ca43c34cb60513db0690f64be2527301b442d69/webrender/src/tiling.rs#L345)):*

Yes - I think that makes sense. Happy if we do that as a follow up?
</blockquote></details>

sure :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2031#-:-Kz-Ru61hHsOWdqLMN3U:b88prtg)*

kvark:
🤞 :shipit: 🚀 
@bors-servo r+

PR #2032, opened 2017-11-13
Add program_binary support

### LINKED ISSUE #2005: Add capability to use glProgramBinary()
To speed up shader initialization, it seems good to use glProgramBinary() if it is requested.

This issue is related to [Bug 1391159](https://bugzilla.mozilla.org/show_bug.cgi?id=1391159)

Also related to servo/gleam#132

sotaroikeda:
Fixes #2005, but still WIP. I did implementation, but I am not sure if it is a way to go.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2032)


sotaroikeda:
Cargo.lock has two gl_generator version, since "servo-glutin" still uses "gl_generator 0.5.4.

sotaroikeda:
> Cargo.lock has two gl_generator version, since "servo-glutin" still uses "gl_generator 0.5.4.

servo/glutin#129 is going to handle it.

webrender/src/device.rs
```diff
@@ -1386,59 +1429,109 @@ impl Device {
...
+                    let error_log = self.gl.get_program_info_log(pid);
+                    println!(
+                      "Failed to load a program object with a program binary: {:?} renderer {}\n{}",
+                      base_filename,
+                      self.renderer_name,
+                      error_log
+                    );
+                } else {
+                    // Bind vertex attributes
+                    for (i, attr) in descriptor
```
kvark: this should not be necessary:
>  Additionally, all vertex shader input and fragment shader output assignments that were in effect when the program was linked before saving are restored with glProgramBinary is called

webrender/src/renderer.rs
```diff
@@ -3791,6 +3794,11 @@ impl Renderer {
         self.ps_composite.deinit(&mut self.device);
         self.device.end_frame();
     }
+
+    #[allow(dead_code)]
```
kvark: isn't this public anyway?

sotaroikeda:
Update the pull request based on the comments. But it does not handle "load from disk" nor  "save to disk" yet. But it improves performance of 2nd shader load a lot with Windows + ANGLE.

kvark:
Thanks @sotaroikeda !

> But it does not handle "load from disk" nor "save to disk" yet. But it improves performance of 2nd shader load a lot with Windows + ANGLE.

Yes, I'm not saying that you should implement this within the PR. What I'm saying that the architecture chosen for this PR is not going to scale to what we need eventually, and likely soon. So it would be nice to have the foundation in place that would not need to be rewritten, especially if we have an idea of how it should look like.

sotaroikeda:
@kvark, can you clarify more about what kind of foundation are you expecting?

sotaroikeda:
> @kvark, can you clarify more about what kind of foundation are you expecting?

I am not sure if latest pull request has an foundational problem.

webrender/src/device.rs
```diff
@@ -1086,59 +1148,101 @@ impl Device {
...
-            .vertex_attributes
-            .iter()
-            .chain(descriptor.instance_attributes.iter())
-            .enumerate()
-        {
-            self.gl
-                .bind_attrib_location(pid, i as gl::GLuint, attr.name);
+        let mut loaded = false;
+
+        if self.cached_programs.is_some() {
```
kvark: given that you are accessing it right afterwards, would be simpler to do
```rust
if let Some(cached_programs) = self.cached_programs {
}
```

webrender/src/device.rs
```diff
@@ -1086,59 +1148,101 @@ impl Device {
...
+                    "Failed to link shader program: {:?}\n{}",
+                    base_filename,
+                    error_log
+                );
+                self.gl.delete_program(pid);
+                return Err(ShaderError::Link(base_filename.to_string(), error_log));
+            }
+        }
+
+        if self.cached_programs.is_some() {
```
kvark: similarly, should probably start with `if let Some(cached_programs) = self.cached_programs {...}`

sotaroikeda:
I updated the pull request by applying the comment and am asking @nical a feedback in [Bug 1391159](https://bugzilla.mozilla.org/show_bug.cgi?id=1391159)

sotaroikeda:
In [Bug 1391159](https://bugzilla.mozilla.org/show_bug.cgi?id=1391159), feedback of @nical was feedback+.

kvark:
Ok, great, thanks @sotaroikeda !
@bors-servo r+

glennw:
@kvark  @sotaroikeda  This seems a bit odd - do we really need to introduce a lifetime parameter onto the Renderer object?

glennw:
It seems like this would require the Servo compositor object to introduce a lifetime parameter, which is likely to complicate the code quite a bit. I haven't looked through this PR completely - but could we instead make the program cache a trait object with a callback, or something like that? What's the reason the device holds a mutable reference to the cache?

sotaroikeda:
> but could we instead make the program cache a trait object with a callback, or something like that?

Ok, I am going to work as to use trait object.

> What's the reason the device holds a mutable reference to the cache?

It was requested by a compiler to call cached_programs.insert.

glennw:
OK, sounds good. Thank you!

PR #2033, opened 2017-11-13
Add support for writing out native font handles on linux/mac

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2033)

### DIFF
--- a/wrench/src/yaml_frame_writer.rs
+++ b/wrench/src/yaml_frame_writer.rs
@@ -224,9 +224,14 @@ fn native_font_handle_to_yaml(handle: &NativeFontHandle, parent: &mut yaml_rust:
     u32_node(parent, "stretch", handle.stretch.to_u32());
 }
 
-#[cfg(not(target_os = "windows"))]
+#[cfg(target_os = "macos")]
 fn native_font_handle_to_yaml(_: &NativeFontHandle, _: &mut yaml_rust::yaml::Hash) {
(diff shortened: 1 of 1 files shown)


glennw:
Gecko uses native font handles on (at least) Linux now - so this lets us convert binary recordings to yaml files again.

kvark:
> error[E0609]: no field `pathname` on type `&webrender::<unnamed>::NativeFontHandle`

wrench/src/yaml_frame_writer.rs
```diff
@@ -225,8 +225,8 @@ fn native_font_handle_to_yaml(handle: &NativeFontHandle, parent: &mut yaml_rust:
 }
 
 #[cfg(not(target_os = "windows"))]
-fn native_font_handle_to_yaml(_: &NativeFontHandle, _: &mut yaml_rust::yaml::Hash) {
-    panic!("Can't native_handle_to_yaml on this platform");
+fn native_font_handle_to_yaml(handle: &NativeFontHandle, parent: &mut yaml_rust::yaml::Hash) {
+    str_node(parent, "font", &handle.pathname);
```
jrmuizel: I wonder if it would be more convenient to copy the font from pathname to the res directory. I guess we can see how this is and implement that as needed.

glennw:
Oh, I didn't realize mac has a real native font handle. Will fix up in the morning, or I've enabled edits from maintainers if anyone else wants this in the meantime.

PR #2038, opened 2017-11-14
choose bitmap-compatible rendering modes with FontInstancePlatformOpt…

lsalzman:
…ions::use_embedded_bitmap

We were currently not even reading this platform option at all, and hence we were not managing to actually support embedded bitmaps at all for DWrite. This chooses modes similar to Skia that should resolve it. This is to fix downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1330734

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2038)


glennw:
Looks reasonable.

@bors-servo r+

PR #2041, opened 2017-11-15
GPU queries refactor

### LINKED ISSUE #1994: Please make timer queries a runtime feature
Having the GPU graphs in Firefox can be really useful. Right now we don't have them because Firefox doesn't turn on the required webrender feature. We also don't want the timer query code to be run by default in Firefox because it can have overhead.

So it would be nice if this feature could be enabled at runtime, for example only when the webrender profiler is shown.

kvark:
Removes `query` compile feature. Fixes  #1994
Refactors the GPU queries quite a bit, including moving the logic into a dedicated module.
Ideally should be tested on Android, but we can patch it afterwards anyway.

Note: the API is asking for some type-level protection against:
  - doing anything outside of frame bounds
  - starting to sample queries in a row
But I don't see a way to make it stronger without introducing obscure programming patterns :) So I'm leaving it with just start/finish semantics, plus a bit of RAII.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2041)


glennw:
I didn't get to this today, sorry - will do first thing in the morning. It looks great so far from a quick scan though!

CI failures:

```
$ if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo build --verbose --features profiler,query); fi

error: Package `webrender v0.53.2 (file:///home/travis/build/servo/webrender/webrender)` does not have these features: `query`

The command "if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo build --verbose --features profiler,query); fi" exited with 101.
```

glennw:
(We probably need to update both the travis.yml and the taskcluster config)

glennw:
Reviewed 12 of 14 files at r1.
Review status: 12 of 14 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[webrender/src/profiler.rs, line 882 at r1](https://reviewable.io:443/reviews/servo/webrender/2041#-KyylIDeFDwOP3Rs81P2:-KyylIDeFDwOP3Rs81P3:bohoyrx) ([raw file](https://github.com/servo/webrender/blob/dffb17b17bc90c34a16b114a8e54e0bfaec207b5/webrender/src/profiler.rs#L882)):*
> ```Rust
>         );
> 
>         if false {
> ```

Should we remove this or comment it perhaps?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2041)*

glennw:
(Partial review of the easy bits so far - will cover the remaining files in the morning).

glennw:
Reviewed 2 of 14 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2041)*

glennw:
I noticed that when I toggle the query feature on and off, a small box occurs in the top left of the screen - it looks like perhaps the quad for the profile counters is still being drawn?

Not a big deal anyway - we can fix that as a follow up if you like.

So, r=me - I'll leave it to you to squash / fix that bug if you want to, or merge now :)

kvark:
Thanks @glennw ! Fix appeared to be rather trivial: https://github.com/servo/webrender/pull/2041/commits/19731c1610f0370128f9efe2714720b279bd49e4#diff-59bdb33ab120f3af70ac6ca84fd5451fR882

@bors-servo r=glennw

staktrace:
@kvark This caused webrender to crash on windows, see https://bugzilla.mozilla.org/show_bug.cgi?id=1417062#c5

PR #2043, opened 2017-11-15
Don't rebuild the entire scene when scrolling with property bindings.

### LINKED ISSUE #1921: The scene is rebuilt everytime we scroll
Github issue for https://bugzilla.mozilla.org/show_bug.cgi?id=1408174 since the code it's referencing is in WR proper and not the gecko side of things.

glennw:
Instead of rebuilding the entire scene / frame when scrolling while
property bindings are present, resolve the animated properties as
we prepare primitives and update the clip-scroll tree.

For transforms, reference frames now retain the property binding
information and build the correct transform in update_node(). We
could possibly optimize this in the future to not always rebuild
the reference frame transform, if there is no property binding active
on this reference frame, if it ever shows up in a profile.

For filters, rely on the new Picture tree functionality and resolve
the opacity filter value when preparing the picture primitive for
rendering. If the property resolves to an invisible filter, then
early out from this entire picture primitive.

Also update the animation example to demonstrate animating opacity.

Fixes #1921.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2043)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -11,6 +11,7 @@ use frame_builder::PrimitiveContext;
 use gpu_cache::GpuDataRequest;
 use prim_store::{PrimitiveIndex, PrimitiveRun, PrimitiveRunLocalRect};
 use render_task::{ClearMode, RenderTask, RenderTaskId, RenderTaskTree};
+use scene::{FilterOpHelpers, SceneProperties};
 use tiling::RenderTargetKind;
 
 /*
@@ -112,6 +113,27 @@ impl PicturePrimitive {
             },
             pipeline_id,
             cull_children: false,
+        }
+    }
+
+    pub fn resolve_scene_properties(&mut self, properties: &SceneProperties) -> bool {
+        match self.kind {
+            PictureKind::Image { ref mut composite_mode, .. } => {
+                match composite_mode {
+                    &mut Some(PictureCompositeMode::Filter(ref mut filter)) => {
+                        match filter {
+                            &mut FilterOp::Opacity(ref binding, ref mut value) => {
+                                *value = properties.resolve_float(binding, *value);
+                            }
+                            _ => {}
+                        }
+
+                        filter.is_visible()
+                    }
+                    _ => true,
+                }
+            }
+            _ => true
         }
     }
 
--- a/webrender/examples/animation.rs
+++ b/webrender/examples/animation.rs
@@ -24,7 +24,9 @@ use webrender::api::*;
 
 struct App {
     property_key: PropertyBindingKey<LayoutTransform>,
+    opacity_key: PropertyBindingKey<f32>,
     transform: LayoutTransform,
+    opacity: f32,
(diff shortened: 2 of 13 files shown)


glennw:
r? @mrobinson and/or @kvark

DO NOT MERGE yet - this mostly seems to work but shows up a very strange artifact
in Gecko. I'm going to investigate that today - and post some details of it here
in case anyone on the Gecko team has ideas what may cause it.

@staktrace - heads up that this modifies the public interface. FilterOp::Opacity enum
has an extra value in addition to the property binding key. That second value is the
default / initial opacity value to use (this is used when a filter op has a property
binding key, but the property bindings don't supply a value for that key).

glennw:
prob

This is the artifact I see when running Gecko with this patch. Below the chrome, there is an arbitrary sized rectangle before the main content. It doesn't seem to be overlapping (note that the scrollbar on the content has a rounded edge where it starts). It also causes hit testing to be incorrectly offset by that vertical amount.

It's possible it's a bug in WR, but this patch is quite simple and I'm not sure how it could cause that. I'll do some investigation today - but I'm wondering if anyone on the Gecko team has any thoughts / insight into what might cause this?

glennw:
Ah, as @staktrace and @mstange correctly pointed out, this was just a case of the iframe transform getting applied twice. That's fixed now, so this is ready for review.

Pending gecko try here:

https://hg.mozilla.org/try/rev/e00aabbecfbe81b2525ef23550f806b05888a233
https://treeherder.mozilla.org/#/jobs?repo=try&revision=e00aabbecfbe81b2525ef23550f806b05888a233

r? @mrobinson and/or @kvark

kvark:
New failures?
> REFTEST TEST-UNEXPECTED-FAIL | file:///builds/worker/workspace/build/tests/reftest/tests/layout/reftests/transform/animate-layer-scale-inherit-1.html

glennw:
Yes, they are new failures - well spotted!

In both of those failures, the max difference is 1.0 / 255.0. I think it's probably fine to add slight fuzz on that, since one test invokes an animation and one doesn't - it's quite likely that there are very minor FP accuracy issues between the tests that we now see since we're not rebuilding the scene each time.

Does that sound reasonable @staktrace and @kvark ?

staktrace:
Yeah I'm ok with fuzzing that.

webrender/src/picture.rs
```diff
@@ -115,6 +116,27 @@ impl PicturePrimitive {
         }
     }
 
+    pub fn resolve_scene_properties(&mut self, properties: &SceneProperties) -> bool {
```
kvark: The scope of this method seems a bit inconsistent to me: in includes opacity filters but excludes the transforms. Could we, perhaps, resolve everything in one place (here)?

glennw: I'm not really sure how, since the transforms are stored in the clip-scroll tree, and the filters are stored in the picture tree?

PR #2046, opened 2017-11-16
Add a couple of simple opaque drawing optimizations.

glennw:
1) When detecting if an image is opaque, WR will check if the
   tile_spacing field is non-zero. If so, it assumes that there
   will be some transparent pixels and uses the alpha pass.
   However, Gecko typically supplies the tile_spacing size to be
   equal to the size of the image rect. This case was not being
   detected as opaque compatible. As a result, every image in
   the Gecko integration was being drawn in the alpha pass.

2) When adding primitives to opaque batches, we would previously
   look for any batch in the existing list to add it to. However,
   when a large primitive is encountered, it's typically more
   efficient to break the batch in order to get much better
   z-buffer efficiency. Now, any opaque primitive which is more
   than 1/4 of the screen size will force a batch break.

Example performance difference on my HD4600 on CNN.com:

From:

  GPU time: 5.6ms
  Opaque coverage: 1.72
  Alpha coverage: 1.32

To:

  GPU time: 3.4ms
  Opaque coverage: 1.07
  Alpha coverage: 1.01

So this relatively small optimization can save a lot of memory
bandwidth and pixel overdraw.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2046)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -250,23 +250,43 @@ impl AlphaBatchList {
 }
 
 pub struct OpaqueBatchList {
+    pub pixel_area_threshold_for_new_batch: i32,
     pub batches: Vec<OpaquePrimitiveBatch>,
 }
 
 impl OpaqueBatchList {
-    fn new() -> OpaqueBatchList {
+    fn new(pixel_area_threshold_for_new_batch: i32) -> OpaqueBatchList {
         OpaqueBatchList {
             batches: Vec::new(),
+            pixel_area_threshold_for_new_batch,
         }
     }
 
-    fn get_suitable_batch(&mut self, key: BatchKey) -> &mut Vec<PrimitiveInstance> {
+    fn get_suitable_batch(
+        &mut self,
+        key: BatchKey,
+        item_bounding_rect: &DeviceIntRect
+    ) -> &mut Vec<PrimitiveInstance> {
         let mut selected_batch_index = None;
+        let item_area = item_bounding_rect.size.area();
 
-        for (batch_index, batch) in self.batches.iter().enumerate().rev().take(10) {
-            if batch.key.is_compatible_with(&key) {
-                selected_batch_index = Some(batch_index);
-                break;
+        // If the area of this primitive is larger than the given threshold,
+        // then it is large enough to warrant breaking a batch for. In this
+        // case we just see if it can be added to the existing batch or
+        // create a new one.
+        if item_area > self.pixel_area_threshold_for_new_batch {
+            if let Some(ref batch) = self.batches.last() {
+                if batch.key.is_compatible_with(&key) {
+                    selected_batch_index = Some(self.batches.len() - 1);
+                }
+            }
+        } else {
+            // Otherwise, look back through a reasonable number of batches.
+            for (batch_index, batch) in self.batches.iter().enumerate().rev().take(10) {
+                if batch.key.is_compatible_with(&key) {
+                    selected_batch_index = Some(batch_index);
+                    break;
+                }
             }
         }
 
@@ -300,10 +320,14 @@ pub struct BatchList {
 }
 
 impl BatchList {
-    fn new() -> BatchList {
+    fn new(screen_size: DeviceIntSize) -> BatchList {
+        // The threshold for creating a new batch is
+        // one quarter the screen size.
+        let batch_area_threshold = screen_size.width * screen_size.height / 4;
+
         BatchList {
             alpha_batch_list: AlphaBatchList::new(),
-            opaque_batch_list: OpaqueBatchList::new(),
+            opaque_batch_list: OpaqueBatchList::new(batch_area_threshold),
         }
     }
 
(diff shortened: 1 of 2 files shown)


glennw:
Oh, those GPU numbers were incorrect. It's more like 5.6ms -> 3.4ms when vsync is disabled.

jrmuizel:
It might be easy to fix the tile_spacing on Gecko side. Would that be better?

glennw:
Wow, anecdotally from testing 10-15 of the sites I commonly test in Gecko, this seems to be making a fairly large improvement.

No need to fix the tile spacing on the gecko side - it's just a one-off fix up when building the frame.

glennw:
This is a good indication that doing some of the more involved work to move pixels into the opaque pass (e.g. when there is a clip rect with rounded corners) is going to be a big GPU win.

glennw:
I'll kick off a try run for this shortly...

glennw:
Try run looks good (the orange tests are expected based on other recent PRs).

https://treeherder.mozilla.org/#/jobs?repo=try&revision=4e70108ebb8176ef0682ce2c4836cbbef1f6d690

kvark:
Is this an old one as well?

> REFTEST TEST-UNEXPECTED-FAIL | file:///builds/worker/workspace/build/tests/reftest/tests/layout/reftests/box-shadow/boxshadow-skiprect.html == file:///builds/worker/workspace/build/tests/reftest/tests/layout/reftests/box-shadow/boxshadow-skiprect-ref.html | image comparison, max difference: 137, number of differing pixels: 1048

kvark:
> As a result, every image in the Gecko integration was being drawn in the alpha pass.

😨 glad this is to be fixed :)

> So this relatively small optimization can save a lot of memory
bandwidth and pixel overdraw.

Need more of those :) 👍 

> When detecting if an image is opaque, WR will check if the tile_spacing field is non-zero.

Strictly speaking, we could use `discard` for gap pixels and still place tiled images in the opaque bin.

> However, when a large primitive is encountered, it's typically more efficient to break the batch in order to get much better z-buffer efficiency.

I don't understand this part. Why would a separate draw call have any effect on z-buffer efficiency versus an instance in another draw call?

glennw:
Consider the following scene (which is actually what happens on CNN.com):

Large opaque rect (z=1)
Large opaque image (z=2)
Large opaque rect (z=3)

Previously, we'd then end up with two batches:

2x Rect
1x Image

We then draw the opaque batches in reverse order for (rough) front-to-back ordering for z-buffer efficiency.

In this case, we'll get much more overdraw than if we break the batches here, since it's worth keeping the render order to get best z-buffer efficiency.

glennw:
The discard trick is worth trying - I'd like to do that as a follow up and do some specific perf testing on how discard affects Intel GPUs though.

webrender/src/tiling.rs
```diff
@@ -250,23 +250,45 @@ impl AlphaBatchList {
...
+        key: BatchKey,
+        item_bounding_rect: &DeviceIntRect
+    ) -> &mut Vec<PrimitiveInstance> {
         let mut selected_batch_index = None;
 
-        for (batch_index, batch) in self.batches.iter().enumerate().rev().take(10) {
-            if batch.key.is_compatible_with(&key) {
-                selected_batch_index = Some(batch_index);
-                break;
+        let item_area = item_bounding_rect.size.width *
```
kvark: nit: there is an `area()` method

webrender/src/tiling.rs
```diff
@@ -250,23 +250,45 @@ impl AlphaBatchList {
...
+        // create a new one.
+        if item_area > self.pixel_area_threshold_for_new_batch {
+            if let Some(ref batch) = self.batches.last() {
+                if batch.key.is_compatible_with(&key) {
+                    selected_batch_index = Some(self.batches.len() - 1);
+                }
+            }
+        } else {
+            // Otherwise, look back through a reasonable number of batches.
+            for (batch_index, batch) in self.batches.iter().enumerate().rev().take(10) {
```
kvark: in a more general case, this would be a heuristic `fn(rectangle_relative_area) -> num_of_batches_to_look_back`. Maybe we can write it like that in the code?

kvark:
What's least clear to me is where exactly we should be splitting batches. The PR's code does: if area is large, put this instance into a new batch. But there can be many other ways to split.

For example, we could accumulate the total area covered by a batch (without tracking intersections), and then have a threshold on this area. The could would then look like:
```rust
let max_area_to_skip = screen_area / 4;
let mut skipped_area = 0;
for (batch_index, batch) in self.batches.iter().enumerate().rev() {
  if batch.key.is_compatible_with(&key) {
    selected_batch_index = Some(batch_index);
    break;
  }
  skipped_area += batch.covered_area;
  if skipped_area >= max_area_to_skip {
    break;
  }
}
```

This would be a little more generic, hopefully smarter, and would get rid of the `10` magic constant.

glennw:
@kvark OK, split it into two commits. Interestingly, when I first found the image opaque issue - I fixed that, and it regressed performance. That's why I started then looking into how we were ordering large opaque rectangles. Before the image fix, we weren't really seeing the ordering problem.

Since this appears to be a quite significant win on several real world sites I tested, would you be happy to get this merged, and we can open an issue to do a more detailed analysis of our options for when to break batches (set up a number of complex benchmarks that we can test various schemes against)?

webrender/src/tiling.rs
glennw: Fixed

kvark:
Sure, thank you @glennw !
I'll follow-up with a PR implementing my proposal and compare the numbers.
@bors-servo r+

glennw:
Sounds good, thanks!

PR #2049, opened 2017-11-16
Query GL assertion removal

kvark:
Fixes https://github.com/servo/webrender/pull/2041#issuecomment-344945460
r? anyone

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2049)


kvark:
can't fail an assert

PR #2050, opened 2017-11-16
DocumentAPI support on the Renderer side

kvark:
This is a second batch of changes related to Document API, following #1509.

The idea is to allow Gecko to have separate documents for the chrome UI, page content, and bottom status bar, as opposed to using different pipelines in the same document. This change would allow minimal scene rebuilds per frame when UI is affected.

WIP TODO:
- [x] strict ordering API: https://github.com/servo/webrender/pull/2050/commits/532bff555393533e30e58b5d326b2759a7655aa5
- [x] Servo patch and test runs: https://github.com/kvark/servo/commit/0263fa8919d2737e09091f1ac71a8d517692261a
- [x] Firefox patch and try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=f939fbdab0231e6328061c12903db0c45847dd18
  - ~~`box-shadow/boxshadow-skiprect.html`~~ (from #1943)
  - ~~`css-filter-chains/moz-element.html`~~ (same as in #2053)
  - ~~`transform/animate-layer-scale-inherit-1.html`~~ (from #2043)
- [x] example
- [x] `FrameBuilder` and `RenderedDocument` refactor
https://tools.taskcluster.net/groups/DEvThfCaQWmt5Hp806GDLg/tasks/VLCZI7hVTQCc1xrDR3PgOQ/details#
- [x] review comments

cc  @glennw  @mstange @jrmuizel

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2050)

### DIFF
--- a/webrender_api/src/api.rs
+++ b/webrender_api/src/api.rs
@@ -14,6 +14,7 @@ use std::fmt;
 use std::marker::PhantomData;
 
 pub type TileSize = u16;
+/// Documents are rendered in the ascending order of their associated layer values.
 pub type DocumentLayer = i8;
 
 /// The resource updates for a given transaction (they must be applied in the same frame).
@@ -885,9 +886,8 @@ pub struct DynamicProperties {
 
 pub trait RenderNotifier: Send {
     fn clone(&self) -> Box<RenderNotifier>;
-    fn wakeup(&self);
-    fn new_document_ready(&self, DocumentId);
-    fn new_scroll_document_ready(&self, DocumentId, composite_needed: bool);
+    fn wake_up(&self);
+    fn new_document_ready(&self, DocumentId, scrolled: bool, composite_needed: bool);
     fn external_event(&self, _evt: ExternalEvent) {
         unimplemented!()
     }
--- a/webrender_api/src/api.rs
+++ b/webrender_api/src/api.rs
@@ -14,6 +14,7 @@ use std::fmt;
 use std::marker::PhantomData;
 
 pub type TileSize = u16;
+pub type DocumentLayer = i8;
 
 /// The resource updates for a given transaction (they must be applied in the same frame).
 #[derive(Clone, Deserialize, Serialize)]
@@ -278,7 +279,7 @@ pub enum ApiMsg {
     /// Adds a new document namespace.
     CloneApi(MsgSender<IdNamespace>),
     /// Adds a new document with given initial size.
-    AddDocument(DocumentId, DeviceUintSize),
+    AddDocument(DocumentId, DeviceUintSize, DocumentLayer),
     /// A message targeted at a particular document.
     UpdateDocument(DocumentId, DocumentMsg),
     /// Deletes an existing document.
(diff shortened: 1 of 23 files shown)


kvark:
The performance aspect here is difficult to get right, since it relies on a frame semantics change. Old logic used to clear everything if either requested, or the frame size is less than the viewport. The new logic is:
  - main framebuffer color is cleared if the background color is provided in the options
  - each rendered document unconditionally clears the depth to 1.0 for its affected area
  - each document can also clear the affected color area, if the clear color is provided for it

Thus, we may end up clearing twice with the new logic. I expect the WR client to be careful about the clear colors and don't specify both the framebuffer and document/frame clear color unless necessary.

Aside from this aspect, the PR should be ready to review/merge now.

kvark:
Actually, something wonky is going on at the start of the new `document` example. Investigating...

kvark:
Ok, it turned out to be this: https://github.com/servo/webrender/pull/2050/commits/cc41e042fb26612285a0eb1312cfa487044aa688#diff-f5062b694b9fe53fc1757ed483d577d9R2788

Some squashing is expected, but otherwise should be ready to review now.

kvark:
Fresh gecko try push with all the new code: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9be1f766b547f71074a3e41a9176c2c3584c5241

glennw:
Reviewed 3 of 12 files at r1, 18 of 22 files at r2, 4 of 5 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2050)*

glennw:
Also - before merging, it'd be good to have patches available for Servo and Gecko to deal with the API changes, to simplify the update process.

webrender_api/src/api.rs
```diff
@@ -884,8 +886,9 @@ pub struct DynamicProperties {
 
 pub trait RenderNotifier: Send {
     fn clone(&self) -> Box<RenderNotifier>;
-    fn new_frame_ready(&self);
-    fn new_scroll_frame_ready(&self, composite_needed: bool);
+    fn wakeup(&self);
```
nical: Nittiest of all nits: I'd prefer `wake_up`

webrender_api/src/api.rs
```diff
@@ -884,8 +886,9 @@ pub struct DynamicProperties {
...
+    fn new_document_ready(&self, DocumentId);
```
nical: While you are modifying this interface, maybe it'd be a good time to merge `new_document_ready` and `new_scroll_document_ready` into a single one that has enough parameters (id, did_scroll, composite_needed) to avoid contortions like in #1963 when the scrolled and non-scrolled paths are merged.

kvark: Agreed.

kvark:
Thanks for the reviews! Concerns are addressed now.
Latest try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=1b91c4648c732ad90544488cc992caf0564ebaa9
Gecko changes: everything under `gfx/webrender_bindings` in https://hg.mozilla.org/try/rev/1b91c4648c732ad90544488cc992caf0564ebaa9
Servo changes: https://github.com/kvark/servo/commit/ffbc3a7fa88e47171c4309de3bb560bc9447adc4

glennw:
@kvark Thanks for the update patches! Want to rebase this again and we'll get it merged before it bitrots again? :)

kvark:
Rebased, added an optimization to avoid redundant clears (in a simple case of 1 document), and pushed another try push - https://treeherder.mozilla.org/#/jobs?repo=try&revision=376c737901e3a27405808e5a828126e60bd91cc5

glennw:
@kvark Looks like there are some failures in that try push, but they seem the same (or similar) to ones we've had with recent PRs. I'm not sure if those are relevant or not in this case?

kvark:
@glennw those 10 are expected, same as with your #2052: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e71afa7f78443212f702971f0a4eb05e27bf7b80

PR #2054, opened 2017-11-17
Tiny cleanup for identifying BatchKind GpuProfileTag

mrobinson:
This makes the code a bit easier to read.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2054)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -171,36 +171,65 @@ const GPU_SAMPLER_TAG_TRANSPARENT: GpuProfileTag = GpuProfileTag {
     color: debug_colors::BLACK,
 };
 
-#[cfg(feature = "debugger")]
+impl TransformBatchKind {
+    #[cfg(feature = "debugger")]
+    fn debug_name(&self) -> &'static str {
+        match *self {
+            TransformBatchKind::Rectangle(..) => "Rectangle",
+            TransformBatchKind::TextRun(..) => "TextRun",
+            TransformBatchKind::Image(image_buffer_kind, ..) => match image_buffer_kind {
+                ImageBufferKind::Texture2D => "Image (2D)",
+                ImageBufferKind::TextureRect => "Image (Rect)",
+                ImageBufferKind::TextureExternal => "Image (External)",
+                ImageBufferKind::Texture2DArray => "Image (Array)",
+            },
+            TransformBatchKind::YuvImage(..) => "YuvImage",
+            TransformBatchKind::AlignedGradient => "AlignedGradient",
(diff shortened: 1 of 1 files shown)


webrender/src/renderer.rs
```diff
@@ -171,6 +171,49 @@ const GPU_SAMPLER_TAG_TRANSPARENT: GpuProfileTag = GpuProfileTag {
     color: debug_colors::BLACK,
 };
 
+trait TransformBatchKindMethods {
+    fn debug_name(&self) -> &'static str;
+    fn gpu_sampler_tag(&self) -> GpuProfileTag;
+}
+
+impl TransformBatchKindMethods for TransformBatchKind {
```
kvark: can we just do `impl TransformBatchKind` here instead of adding a trait?

webrender/src/renderer.rs
```diff
@@ -171,6 +171,49 @@ const GPU_SAMPLER_TAG_TRANSPARENT: GpuProfileTag = GpuProfileTag {
...
+            TransformBatchKind::Line => "Line",
+        }
+    }
+
+    fn gpu_sampler_tag(&self) -> GpuProfileTag {
+        match *self {
+            TransformBatchKind::Rectangle(_) => GPU_TAG_PRIM_RECT,
+            TransformBatchKind::Line => GPU_TAG_PRIM_LINE,
+            TransformBatchKind::TextRun(..) =>
+                unreachable!("bug: text runs don't have a GPU sampler tag."),
```
kvark: any reason why we can't have one?

mrobinson: It wouldn't be used because these are special cased, but, you're right, it does make sense to have this return the right thing!

webrender/src/renderer.rs
mrobinson: Thanks for taking a look at this! I used a trait here, because TransformBatchKind is defined in tiling.rs.

kvark: It's still the same crate. Notice how `BatchKind` is also not defined in this module but still implemented.

mrobinson: Ah, you're right. I'll fix this.

glennw:
Looks like there are some CI compiler errors:

```
error[E0599]: no method named `gpu_sampler_tag` found for type `tiling::BatchKind` in the current scope
    --> src\renderer.rs:2743:60
     |
2743 |         let _timer = self.gpu_profile.start_timer(key.kind.gpu_sampler_tag());
```

mrobinson:
@glennw Oops. I think I've taken care of the build issue now, testing both release and debug builds locally. We'll see what the bots say though.

Issue #1808, opened 2017-10-05
Transformed text seems to stretch glyph image border pixels, creating fringes around glyphs


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1405953

[Testcase](https://bug1405953.bmoattachments.org/attachment.cgi?id=8915469)

<img width="302" alt="webrender-text-fringes-2x" src="https://user-images.githubusercontent.com/961291/31213057-4bbf0cf2-a972-11e7-900e-9e48528e5b97.png">

Gankro:
I'm pretty sure this is just subpixels. We currently render text before transforms, and with no regard for their existence.

mstange:
I agree, this depends on #1542 and is just one of the things we should make sure works correctly once that's fixed. There will probably still be cases where we want to rasterize and then transform, and in those cases, this particular rendering error should not occur.

Darkspirit:
Is this also fixed by https://github.com/servo/webrender/pull/1969 ? (They are "See Also"s in Bugzilla)

mstange:
No, it's not. I'll remove it from the see also list.

lsalzman:
This is fixed by PR #2058 with followup in PR #2093.

PR #2058, opened 2017-11-18
support font subpixel AA with content transforms

lsalzman:
This solves issue #1808. It ensures that we render subpixel AA in device space rather than layer local space, so as to side-step the issue with subpixels getting deformed by a layer transform.

The engine of this is FontTransform, which is the upper 2x2 submatrix of the layer transform and which is stored in the FontInstance - automatically and necessarily becoming part of the font key. That transform has to be passed down to the font backend so that rasterization actually uses it.

In the text shader, we can just optionally derive this transform from the layer transform, without having to pass in any other state. We have to make sure we only conditionally use it, as not all glyphs will be pre-transformed. To that end, I made TextShader a replacement for PrimitiveShader (as discussed with Glenn).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2058)


glennw:
Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2058)*

glennw:
This looks great! We'll need a gecko try run before merging though. I'll handle running it through the Servo tests.

glennw:
Oh, one other thing - we should add some wrench reftest cases - rotation, skew etc on both subpx and alpha test runs. Also with some text shadow cases. These can be done as a follow up.

glennw:
There's a few gecko try failures that @lsalzman is working on - we know the cause.

lsalzman:
Fixed issue with computing snap offsets in the text shader.

glennw:
I think we should have at least a couple of reftests (just on one platform is enough) for this stuff - it's so fiddly to get right and easy to regress.

lsalzman:
Fixed merge conflicts.

lsalzman:
Added subpixel rotation, scale, and skew transform reftests.

glennw:
Based on discussions with @lsalzman I think this is blocked until we get some fixes to the mac dependency crates.

lsalzman:
Fixed core-graphics-rs to have the functionality needed and updated this to use it.

glennw:
Reviewed 20 of 20 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2058)*

glennw:
@lsalzman has done a try run and got green results, so this looks good to go. Thanks!

@bors-servo r+

glennw:
Needs a rebase.

lsalzman:
Fixes #1808

PR #2060, opened 2017-11-19
Fix build breakage with current rustc:

glennw:
error: the crate `bitflags` is compiled with the panic strategy `abort` which is incompatible with this crate's strategy of `unwind`

error: the crate `bitflags` is compiled with the panic strategy `abort` which is incompatible with this crate's strategy of `unwind`

error: Could not compile `gleam`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2060)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,4 +7,3 @@ members = [
 
 [profile.release]
 debug = true
-panic = "abort"


Gankro:
@bors-servo r+

do you know what changed?

glennw:
With stable rustc, trying to build release + master of WR gives the compile errors above.

I'm not really sure what changed to cause this - I haven't updated rustc, and I don't think any dependencies should have changed due to the Cargo.lock?

Anyone with knowledge of this stuff explain (a) why this happened now (b) If this is a reasonable workaround for now (I don't think it will break Gecko?) (c) If there is a better fix?

glennw:
This is the last PR to merge, and it contains dependency updates https://github.com/servo/webrender/pull/2032.

So I guess that's what caused the build breakage, but I don't understand how that got through CI if that's the case... :confused:

Gankro:
@bors-servo r-

seeing fishy stuff locally

glennw:
Closing while we try to diagnose this a bit further.

Gankro:
Running `cargo clean` seems to fix it and I have no idea why (but I guess that's why the CI is fine).

glennw:
OK, as @Gankro identified, this seems to be some kind of rustc / cargo bug - running ```cargo clean``` locally and then building completes successfully. :smiley:

PR #2062, opened 2017-11-20
Better inner and outer rect calculation for ClipChains

mrobinson:
We calculate the real inner and outer boundaries of clips in screen
space and use it to discard clips and generate masks that are a little
smaller in some cases. This results in a small performance improvement
on pages with complicated clipping.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2062)


mrobinson:
Here is how this looks on the W3 testcase from https://github.com/servo/webrender/issues/1817.
```
Before                                  After
CPU: 1.54                               CPU: 1.87
Compositor: 15.39                       Compositor: 15.33
GPU: 11.95                              GPU: 10.66
```
I'm still looking at the regression in CPU time, but I'd love to get general comments on this approach.

I did a Gecko try push as well and the only issues I see are some GTest failures, which I think are unrelated.
https://treeherder.mozilla.org/#/jobs?repo=try&revision=3e470a85a5a66cf9e3ae4a97b70e85ba4b35da67

mrobinson:
I updated the reference image for one test that had a small number of subpixel differences.

glennw:
I'll look at this in detail in the morning.

One question though - do you have a sense of what the profile numbers would be like with the improved inner / outer rect calculation and filtering, but without the shader changes?

I wonder if the majority of the wins might be from the CPU-side filtering? (I thought that the clip masks were already intersected with the screen rect of the clip node - since we support the clip mask being a portion of the primitive rect - but maybe that's not working / slightly different).

mrobinson:
@glennw So I took another look at this. I think we need to do this with the change to the shader. The case (and I suspect it might be the important case) is when we can avoid masking at all. This happens when the screen space rectangle of the ClipChain is sufficient to clip the primitive. in this case we can just skip creating the task entirely and rely on the changes to the fragment shader. Maybe there is a better way to do this, but I don't think we can use local geometry in this case.

mrobinson:
I pushed a new version of this PR which moves outer rect calculation to the ClipScrollTree update. This moves a little bit of work from the per-primitive parts to the per-ClipScrollNode part. I'm not seeing a CPU speedup for the W3C case, but I think that this cleaner in general.

I am also thinking about extending this sort of thing to handling some situations where we can generate shorter ClipChains in general.

kvark:
Thanks @mrobinson, impressive work!
I expressed the current questions/concerns. Will want to take another look after the first wave is addressed.

---

Reviewed 22 of 22 files at r1.
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[webrender/res/prim_shared.glsl, line 615 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQhpPajmLaX05F925R:-KzQhpPajmLaX05F925S:b6r5kpb) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/res/prim_shared.glsl#L615)):*
> ```GLSL
>     vScreenClipRect.xy += task_offset;
>     if (task.size == vec2(0.0)) {
>         vScreenClipRect.y = (uBufferHeight - vScreenClipRect.y) - vScreenClipRect.w;
> ```

why are we doing this?

---

*[webrender/res/prim_shared.glsl, line 889 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQh3Ox2fv_Jhr9Zjuh:-KzQh3Ox2fv_Jhr9Zjui:b-92vshj) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/res/prim_shared.glsl#L889)):*
> ```GLSL
> }
> 
> float apply_rectangular_screen_space_clip(vec4 screenSpaceClipRect) {
> ```

any reason not to use `vScreenClipRect` directly here?

---

*[webrender/res/prim_shared.glsl, line 893 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQgakpqlQpKVZAQXKy:-KzQgakqQecNjTDWC33c:b77n3ei) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/res/prim_shared.glsl#L893)):*
> ```GLSL
>     // is inside the screen space clipping box.
>     vec2 topLeft = screenSpaceClipRect.xy;
>     vec2 bottomRight = screenSpaceClipRect.xy + vec2(screenSpaceClipRect.z, screenSpaceClipRect.w);
> ```

nit: could use swizzling `screenSpaceClipRect.zw`

---

*[webrender/src/clip.rs, line 198 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiEpHekwkgttOCrhA:-KzQiEpHekwkgttOCrhB:b-557co0) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip.rs#L198)):*
> ```Rust
>         }
> 
>         let outer = match can_calculate_outer_rect {
> ```

isn't setting this to `true` equivalent to saying `local_outer = None`?

---

*[webrender/src/clip.rs, line 203 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiIb1ROzZoIMPrFIg:-KzQiIb1ROzZoIMPrFIh:b-1ww82u) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip.rs#L203)):*
> ```Rust
>         };
> 
>         let inner = match can_calculate_inner_rect {
> ```

similarly, setting `local_inner` to `zero()` should do the same

---

*[webrender/src/clip.rs, line 216 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiRiSpowHbLrjSQPc:-KzQiRiSpowHbLrjSQPd:b-2e3h1f) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip.rs#L216)):*
> ```Rust
>         resource_cache: &mut ResourceCache,
>     ) {
>         if self.clips.is_empty() {
> ```

I don't think the early out is useful here since all the function does is iterating the list twice

---

*[webrender/src/clip_scroll_node.rs, line 309 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQihKqK_j3EYYoDpeG:-KzQihKqK_j3EYYoDpeH:b-mo4o6z) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip_scroll_node.rs#L309)):*
> ```Rust
>                         self.reference_frame_relative_scroll_offset,
>                     scroll_offset: self.scroll_offset(),
>                     combined_clip_outer_bounds: LayerRect::new(
> ```

could just do `self.combined_clip_outer_bounds.to_f32()`

---

*[webrender/src/clip_scroll_node.rs, line 352 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiugh2TEWQeioBmDK:-KzQiugifIyzHkpRhqNC:btnve69) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip_scroll_node.rs#L352)):*
> ```Rust
>             clip_sources.get_screen_bounds(&self.world_viewport_transform, device_pixel_ratio);
> 
>         let combined_outer_screen_rect = match screen_outer_rect.as_ref() {
> ```

no need for `as_ref()` in a `match`

---

*[webrender/src/clip_scroll_node.rs, line 359 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQjGBquQ_HJfOSFaeq:-KzQjGBr-zSGPRk9XBXm:b3lxzqi) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip_scroll_node.rs#L359)):*
> ```Rust
>             None => state.combined_outer_clip_bounds,
>         };
>         state.combined_outer_clip_bounds = combined_outer_screen_rect;
> ```

hmm, the old code doesn't modify `state.combined_outer_clip_bounds`. Why are we doing it now?

---

*[webrender/src/frame_builder.rs, line 1536 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQjfsVVPNjQvlQLbRb:-KzQjfsW44NriFkyP4UA:byeoudx) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/frame_builder.rs#L1536)):*
> ```Rust
>         device_pixel_ratio: f32,
>         scene_properties: &SceneProperties,
>         screen_rect: &DeviceIntRect,
> ```

there is already `self.screen_rect`, can we use it?

---

*[webrender/src/prim_store.rs, line 1256 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQmecoD16IwR976DmN:-KzQmecqlfvuxR21fH5G:blfgauq) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/prim_store.rs#L1256)):*
> ```Rust
>                     screen_inner_rect,
>                     combined_outer_screen_rect:
>                         combined_outer_rect.unwrap_or_else(DeviceIntRect::zero),
> ```

I don't think `zero()` is appropriate replacement for the lack of the outer bound

---

*[webrender/src/prim_store.rs, line 1266 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQoCZmY2xfXe6VDKu9:-KzQoCZmY2xfXe6VDKuA:b-n6pg4f) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/prim_store.rs#L1266)):*
> ```Rust
>         // If everything is clipped out, then we don't need to render this primitive.
>         let combined_outer_rect = match combined_outer_rect {
>             Some(rect) if !rect.is_empty() => {
> ```

leaving a note for myself that this is related to the `zero()` assignment somewhere above

---

*[webrender/src/prim_store.rs, line 1267 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQkcpJFPaxvuwGAeYH:-KzQkcpJFPaxvuwGAeYI:b-36mm6p) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/prim_store.rs#L1267)):*
> ```Rust
>         let combined_outer_rect = match combined_outer_rect {
>             Some(rect) if !rect.is_empty() => {
>                 metadata.screen_rect = Some(rect);
> ```

nit: let's move this assignment after the `match`, so that the `match` becomes non-mutating

---

*[webrender/src/prim_store.rs, line 1278 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQmwAj76z-2Swv4spT:-KzQmwAlsVctnmuEhcSd:b-yeqoxf) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/prim_store.rs#L1278)):*
> ```Rust
>             .chain(ClipChainNodeIter { current: extra_clip })
>             .filter_map(|node| {
>                 if !node.screen_inner_rect.is_empty() {
> ```

nit: could be `combined_inner_rect = if ... {} else {..};`

---

*[webrender/src/renderer.rs, line 1106 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQlqVtYRjqdoedHm0n:-KzQlqVu-N_ABycBJU2l:bvs0aou) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/renderer.rs#L1106)):*
> ```Rust
>         projection: &Transform3D<f32>,
>         mode: M,
>         buffer_height: u32,
> ```

I'd find it easier to understand if it was named "target_height" instead.

---

*[webrender/src/util.rs, line 442 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQlb3vi24JSJvBLLrJ:-KzQlb3w_uWilZSs4z8d:b-plxcz5) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/util.rs#L442)):*
> ```Rust
>         LayerRect::new(
>             LayerPoint::new(f32::MIN / 2.0, f32::MIN / 2.0),
>             LayerSize::new(f32::MAX, f32::MAX),
> ```

would this guarantee that origin + size is not going to end up at infinity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2062#-:-KzQoP2gKERaBxEE8wfb:b-44v3xs)*

glennw:
OK, I think I have worked out in my head what my concern is with this approach. I get easily confused by the various permutations, so I'll draw up a diagram as reference and try to explain shortly - it's quite likely I am not thinking it all through properly :)

glennw:
OK, my thoughts below. I pre-emptively apologize if the stuff below sounds a bit short - I'm basically just trying to list a series of statements, so that we can easily discuss which bits below are correct and which bits are wrong, or missing some context / information. Let me know if that makes any sense (or completely misses the point!) - we can also set up a Vidyo / Skype call to discuss in detail, if that's easier?

This is a simplified set of examples - but I *think* it extends to cover more complex cases (e.g. when there are multiple clips in a clip chain).

Red blocks are clips. Green blocks are primitives.

untitled diagram


In each of the four scenarios, we have (1) primitive completely inside clip (2) primitive partially clipped (3) primitive not affected by clip.

In (A) both clip and primitive are axis-aligned - the coordinate systems are considered compatible.
In (B) the primitive has a rotation - the coordinate systems are considered *incompatible*.
In (C) the clip has a rotation - the coordinate systems are considered *incompatible*.
In (D) both clip and primitive are rotated by a parent node - the coordinate systems are considered compatible.

I *think* the following statements are true:
	- This change helps with avoiding a clip mask in scenario (A) and (B).
	- (C) still requires a mask, but *only* in the overlapping case. If we detect the primitive is inside/outside the clip region, we can skip the mask. We can do a cheap check first, with AABB checks. If the potential clip mask is large enough to warrant it, and AABB check fails, we can use SAT or similar.
	- (D) doesn't require a mask with or without this change (since we can apply it as a local clip in the vertex shader).
	- In both (A) and (B) we can detect that the primitive is inside or outside the region (using AABB / SAT), we can skip the clip mask for those on CPU.

If the above is correct, what we get from this PR is to skip the clip mask in (A) and (B) when we detect that the clip region of interest intersects with the primitive.

However, in the case of (A), we can apply the same logic as (D) - that is, the coordinate systems are compatible, and therefore we can skip the clip mask and apply this as a local clip in the vertex shader.

The point above is important, for two reasons:
 (1) Scenario (A) is certainly the vast majority of cases we will encounter on real web pages (but it's still important to handle (B) efficiently).
 (2) In terms of GPU time, we are always limited by the number of fragments we process. This solution still invokes the fragment shader for those extra pixels (when drawing the primitive). Although it does save drawing a clip mask, I think we can do better by skipping the clip mask *and* applying a local clip in the vertex shader in these cases - this means we don't invoke the FS at all for the clipped fragments.

So, it does seem like this is a potential win for the intersecting case in scenario (B). My *guess* is that we almost never actually see this in real world, and that the benefits we're seeing are because we're not filtering clip masks out (and into a local clip when compatible) as much as we could be. There's also a slight downside that we have extra instructions running for every fragment shader to apply that screen space clip rect.

Apologies for the wall of text. Thoughts?

mrobinson:
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[webrender/res/prim_shared.glsl, line 615 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQhpPajmLaX05F925R:-KzT1rg8pN0JO0hXP_aH:b-dlj6ft) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/res/prim_shared.glsl#L615)):*

why are we doing this?
</blockquote></details>

This is because we need to flip the y axis when drawing the screen. I've left a comment here, because it is indeed pretty confusing.

---

*[webrender/res/prim_shared.glsl, line 889 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQh3Ox2fv_Jhr9Zjuh:-KzT1xgTxnIKJO2X1wVk:bw10hod) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/res/prim_shared.glsl#L889)):*

any reason not to use `vScreenClipRect` directly here?
</blockquote></details>

Nope. No reason. I'll make that change.

---

*[webrender/res/prim_shared.glsl, line 893 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQgakpqlQpKVZAQXKy:-KzT2-V6XhWvUeFCvhAV:b-5vx5hw) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/res/prim_shared.glsl#L893)):*

nit: could use swizzling `screenSpaceClipRect.zw`
</blockquote></details>

Nice catch. This was due to organic evolution of the code. ☺

---

*[webrender/src/clip.rs, line 198 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiEpHekwkgttOCrhA:-KzT23kA7-DJb0CWlL6r:b-95pc5i) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip.rs#L198)):*

isn't setting this to `true` equivalent to saying `local_outer = None`?
</blockquote></details>

It is equivalent, but I think the extra variables make the code easier to read here.

---

*[webrender/src/clip.rs, line 203 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiIb1ROzZoIMPrFIg:-KzT28T81CnckG50zpGc:bs4q9op) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip.rs#L203)):*

similarly, setting `local_inner` to `zero()` should do the same
</blockquote></details>

Ditto.

---

*[webrender/src/clip.rs, line 216 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiRiSpowHbLrjSQPc:-KzT29g1gsgoltLpYh92:bw5asw8) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip.rs#L216)):*

I don't think the early out is useful here since all the function does is iterating the list twice
</blockquote></details>

Okay. That makes sense. I'll also do all the work in a single loop since I think that's a bit cleaner.

---

*[webrender/src/clip_scroll_node.rs, line 309 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQihKqK_j3EYYoDpeG:-KzT2F0bvXHVMqqmKktN:b-66s3y7) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip_scroll_node.rs#L309)):*

could just do `self.combined_clip_outer_bounds.to_f32()`
</blockquote></details>

Yep! With an addition change which is to make this member a DeviceRect (which it should have been anyway).

---

*[webrender/src/clip_scroll_node.rs, line 352 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQiugh2TEWQeioBmDK:-KzT2K0HLmvSp1u90XsI:b-ucfknh) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip_scroll_node.rs#L352)):*

no need for `as_ref()` in a `match`
</blockquote></details>

Thanks!

---

*[webrender/src/clip_scroll_node.rs, line 359 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQjGBquQ_HJfOSFaeq:-KzT2pAGpfm0VJ_YmLTQ:b-ovzp08) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/clip_scroll_node.rs#L359)):*

hmm, the old code doesn't modify `state.combined_outer_clip_bounds`. Why are we doing it now?
</blockquote></details>

Oh. It is also modified below, but I had forgotten to remove the previous statement. I'll fix that. Nice catch!

---

*[webrender/src/frame_builder.rs, line 1536 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQjfsVVPNjQvlQLbRb:-KzT2tm3e9l10ab97tr4:bkaqkai) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/frame_builder.rs#L1536)):*

there is already `self.screen_rect`, can we use it?
</blockquote></details>

There is self.screen_size, but I'll add a method to generate a screen_rect easily from that. That will allow us to avoid passing the parameter here.

---

*[webrender/src/prim_store.rs, line 1256 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQmecoD16IwR976DmN:-KzT3-5KclWiGe_nwKIQ:b6h7373) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/prim_store.rs#L1256)):*

I don't think `zero()` is appropriate replacement for the lack of the outer bound
</blockquote></details>

So we need to represent two cases:
 1. When the outer bounds are undefined.
 2. When the outer bounds are an empty rect (ie clipped out entirely).

Currently we use None to mean when the outer bounds are undefined. In this case, they cannot be undefined (which is why this parameter is not an Option), so I have used the empty rect here to mean that we are entirely clipped out.

---

*[webrender/src/prim_store.rs, line 1267 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQkcpJFPaxvuwGAeYH:-KzT2zMnG2En-0ysf07r:b-837i3v) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/prim_store.rs#L1267)):*

nit: let's move this assignment after the `match`, so that the `match` becomes non-mutating
</blockquote></details>

Okay.

---

*[webrender/src/prim_store.rs, line 1278 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQmwAj76z-2Swv4spT:-KzT3MUs9Racay66uXzD:bjouagd) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/prim_store.rs#L1278)):*

nit: could be `combined_inner_rect = if ... {} else {..};`
</blockquote></details>

Okay. That's a bit nicer.

---

*[webrender/src/renderer.rs, line 1106 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQlqVtYRjqdoedHm0n:-KzT3ep7fzpyyg3T6RyJ:bdhq1fn) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/renderer.rs#L1106)):*

I'd find it easier to understand if it was named "target_height" instead.
</blockquote></details>

Okay. No problem.

---

*[webrender/src/util.rs, line 442 at r1](https://reviewable.io:443/reviews/servo/webrender/2062#-KzQlb3vi24JSJvBLLrJ:-KzT3hQdzF-QIwHbl7kr:b-8bz2zi) ([raw file](https://github.com/servo/webrender/blob/7b2abb12541fef8919cdf33a378b252519ac7b42/webrender/src/util.rs#L442)):*

would this guarantee that origin + size is not going to end up at infinity?
</blockquote></details>

I think that dividing by two does this, but I have to admit I am not certain. I followed the same strategy used for other implementations of `max_rect()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2062)*

kvark:
@glennw 
Great pictures showing different classes of use-cases for clipping! 👍 

> I think the following statements are true:

Yes, except for the fact A is already classified as D (without the need of this PR), as you mentioned later:
> However, in the case of (A), we can apply the same logic as (D) - that is, the coordinate systems are compatible, and therefore we can skip the clip mask and apply this as a local clip in the vertex shader.

So yes, the PR only addresses use-case (B), and it does add a bit of overhead for the general case. How much though, is still to be figure out.

> (2) In terms of GPU time, we are always limited by the number of fragments we process. This solution still invokes the fragment shader for those extra pixels (when drawing the primitive). Although it does save drawing a clip mask,

I don't think this PR makes us process more fragments than before. Instead of drawing a bunch of stuff into the clip mask, we are now drawing less fragments of the main primitive. So, it appears to me that we'd be processing less fragments for (B) with this PR, and same number of fragments for (A), (C), and (D) with a few added instructions.

> I think we can do better by skipping the clip mask and applying a local clip in the vertex shader in these cases - this means we don't invoke the FS at all for the clipped fragments.

Well, that part needs some elaboration :) Reason we can't do it is because the local clip is in the local space, and the clip we are trying to process here is in screen space. So vertex shader would only be able to help if we had `gl_ClipDistance`, and even then it would likely to turn out into simple `discard()` calls on the modern HW, which we can do already ourselves.

mrobinson:
> Great pictures showing different classes of use-cases for clipping! 👍

Totally agree! Those are really nice. I would love to put together some documentation with these types of diagrams.

> Well, that part needs some elaboration :) Reason we can't do it is because the local clip is in the local space, and the clip we are trying to process here is in screen space. So vertex shader would only be able to help if we had gl_ClipDistance, and even then it would likely to turn out into simple discard() calls on the modern HW, which we can do already ourselves.

This brings up a good point which is to test the impact of using discard instead of 0 alpha.

kvark:
>  So vertex shader would only be able to help if we had gl_ClipDistance

 Another alternative is to have scissor-enabled batches. This would work best in terms of number of fragments and their processing cost, would be fairly straightforward to put in (least intrusive of all proposals) but would introduce batch breaks. If what @glennw was saying is correct ("we almost never actually see this in real world"), then we can totally afford this.

mrobinson:
I just tested this, and interestingly enough using an early branch/`discard` approach seems to erase all performance benefits of this PR on my hardware (Mesa DRI Intel(R) HD Graphics P530 (Skylake GT2)).

glennw:
@mrobinson In terms of GPU timing - could you try with vsync disabled? That will certainly be the cause of the compositor timing being weird (it includes the wait for vsync time), and also has a significant effect on the GPU timer queries (on my machine, anyway).

glennw:
I'll have a proper read through this today (thanks for the detailed replies!) but I do quite like the idea of scissor enabled batches - it'd be interesting to see what effect that has on timing (if it is indeed quick to hack in and test).

glennw:
> I don't think this PR makes us process more fragments than before. Instead of drawing a bunch of stuff into the clip mask, we are now drawing less fragments of the main primitive. So, it appears to me that we'd be processing less fragments for (B) with this PR, and same number of fragments for (A), (C), and (D) with a few added instructions.
> 

That's correct - I didn't mean to imply that we're drawing more fragments than previously. What I meant was that if we're using this technique to mask out fragments in the (A) case, we'd be drawing more fragments than if we were to use the local clip rect in that case. But if I understand the PR, we're not processing *less* fragments either - since we end up still running the FS on all the pixels, just masking out the result at the end of the shader?

> Well, that part needs some elaboration :) Reason we can't do it is because the local clip is in the local space, and the clip we are trying to process here is in screen space. So vertex shader would only be able to help if we had gl_ClipDistance, and even then it would likely to turn out into simple discard() calls on the modern HW, which we can do already ourselves.
> 

Yea, I was discussing that in the context of (A). Certainly if we are encountering a lot of the (B) cases, we either need this technique or clip masks (assuming we don't have gl_ClipDistance etc).

To try and sum that up:
 * I *suspect* that in this test case, we're primarily encountering cases of (A), rather than (B).
 * I *suspect* that we're not doing the optimal thing in the case of (A) - that is, handling these as a local clip rect and avoiding mask generation where possible.
 * If the above is true, I *suspect* this technique would be faster than the current code, due to avoiding mask generation, but that if we were in fact handling the (A) cases with a local clip, we'd be faster again even without this technique.

There's a lot of ifs, buts and maybes there - perhaps you guys know off hand if some of those guesses above are correct or not?

kvark:
@glennw My understanding is that we are already handling (A) optimally, since the local clip does the job. Let's verify!

kvark:
If I understand correctly, our `reftests/mask/aligned-layer-rect.yaml` is case (A). I took a GPU texture and verified that we don't render any clip masks there, as expected. So we are already doing the best job there.

glennw:
@mrobinson I was working on some MotionMark benchmarks today. While looking at #1648 it seemed that we were drawing a huge number of redundant clip masks.

I took your PR here, and removed the parts we're still discussing (the GPU parts). You can see what I ended up with here - https://github.com/glennw/webrender/commit/c9ce44bee70bdb5036377cda85271f9b9c137da3.

The good news is that those improvements to clip chain filtering make a huge improvement to #1648. These changes remove all the redundant clip masks, and take the score on my machine from 165 to 472.

Would you be happy to follow up on that and we could land this PR with the CPU-side improvements now, and then continue investigating the GPU-side improvements as a separate issue?

cc @kvark

mrobinson:
Sure. I'll try to pull apart the CPU and GPU bits and open a new PR to follow  up on the screen space clipping idea.

PR #2063, opened 2017-11-20
Update the whole macOS crate stack

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2063)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,6 +1,6 @@
 [root]
 name = "wrench"
-version = "0.2.3"
+version = "0.2.4"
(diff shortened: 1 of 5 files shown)


staktrace:
@nox This seems to have broken the OS X build of webrender in gecko - try push links are in https://bugzilla.mozilla.org/show_bug.cgi?id=1418315#c3

staktrace:
@nox never mind, mstange pointed out that I need to bump the OS X deps in webrender_bindings too. My script to generate the update didn't handle that.

PR #2067, opened 2017-11-21
Fix line number error reports from shader compiler.

glennw:
Insert a marker at the start of each source file, and also a
line marker after inclusion of a source file with the current
line number.

This means the reported line number for shader compile errors
actually matches the real line numbers in glsl files.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2067)


kvark:
FWIW, I typically just load `apitrace`, extract the source from there, and check the line number.

PR #2069, opened 2017-11-21
Remove manual interpolation of local space in transform shaders.

glennw:
Instead, for vertex transform shaders we now write out:
 (world.xy, z * world.w, world.w)

This gives us constant Z across the polygon for draw ordering.

It also means we can use the hardware interpolators instead of
manually dividing by Z in the fragment shader for local_pos.

This is mostly preparation work for switching shaders over to
be brush primitives, but has the happy side effect of:

 * Fixing a small number of Gecko + Servo reftests.
 * Simpler and faster code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2069)


glennw:
r? @kvark

New passes in Servo:
  ▶ PASS [expected FAIL] /css/css-transforms/css-transforms-3d-on-anonymous-block-001.html
  ▶ PASS [expected FAIL] /_mozilla/css/transform_skew_a.html

New passes in Gecko:
  transform-3d/split-non-ortho1.html

Gecko try is here (looks to be all green except for the PASS above and intermittents):
https://hg.mozilla.org/try/rev/9d2110ecd88f77aec63dd5cfd8c0d58ce08b1128
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9d2110ecd88f77aec63dd5cfd8c0d58ce08b1128

glennw:
There's clearly some more tidy up that can be done as a follow up with this change (e.g. we could unify ```write_vertex``` and ```write_transform_vertex``` quite easily), but I think it's reasonable to do those as follow ups as we switch primitives over to be brush shaders.

glennw:
To save any more rebase hassles for @lsalzman, we should make sure https://github.com/servo/webrender/pull/2058 lands before this (I'll rebase on that if needed, but review of this should still be fine in the meantime).

glennw:
Rebased over the rotated text PR.

kvark:
I really like where this is going! :+1: 
A few concerns need to be cleared out though.

---

---

*[webrender/res/prim_shared.glsl, line 652 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVS5umoGyNl63U2aqR:-KzVS5undT8LziYkpRlJ:b-vlaaxz) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L652)):*
> ```GLSL
>     // Calculate a clip rect from local_rect + local clip + layer clip.
>     RectWithEndpoint local_rect = to_rect_with_endpoint(instance_rect);
>     local_rect.p0 = clamp_rect(local_rect.p0, local_clip_rect);
> ```

can we just clamp to `clip_rect` instead?

---

*[webrender/res/prim_shared.glsl, line 665 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVSlvliEzjjGm9WWwu:-KzVSlvliEzjjGm9WWwv:b-yyqkml) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L665)):*
> ```GLSL
>     // can do some math on the projection matrix to work out a variable
>     // amount to extrude.
>     float extrude_distance = 4.0;
> ```

I'm still a bit confused about `4` here. I'd prefer us either to be precise for the simple case only (of 1:1 scale with the screen, so that the `extrude = 0.5`), or for the general case, but not for an arbitrary point in-between.

---

*[webrender/res/prim_shared.glsl, line 665 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVUYgZoAA5gTQ5eapm:-KzVUYg_Wj0yYzFxwKxy:b-hfgcji) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L665)):*
> ```GLSL
>     // can do some math on the projection matrix to work out a variable
>     // amount to extrude.
>     float extrude_distance = 4.0;
> ```

TODO (possibly, for myself): use `clip_edge_mask` here

---

*[webrender/res/prim_shared.glsl, line 688 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVU5mz4yUyPMtX6Lia:-KzVU5n0IkFopHn7L_df:b-tdsuoa) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L688)):*
> ```GLSL
> 
>     // Apply offsets for the render task to get correct screen location.
>     gl_Position.xy += task.common_data.task_rect.p0 - task.content_origin;
> ```

interesting, so this is changing the semantics: before we had translation to the target space followed by `uTransform`, now we have the opposite order

---

*[webrender/res/prim_shared.glsl, line 704 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVUUAGdL_xgC0Y7CKn:-KzVUUAGdL_xgC0Y7CKo:bbeppw) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L704)):*
> ```GLSL
>         prim.local_rect,
>         prim.local_clip_rect,
>         vec4(1.0),
> ```

why is this changed?

---

*[webrender/res/prim_shared.glsl, line 823 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVUeEFPe_vBr1Y-YtX:-KzVUeEFPe_vBr1Y-YtY:b5mwje1) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L823)):*
> ```GLSL
> }
> 
> float init_transform_fs(vec2 local_pos) {
> ```

sweeet!

---

*[webrender/res/ps_image.glsl, line 87 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVVDrJCKACIyIMfExZ:-KzVVDrK2fn3_VeiRDmu:b-q2bmxo) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/ps_image.glsl#L87)):*
> ```GLSL
>     // We clamp the texture coordinate calculation here to the local rectangle boundaries,
>     // which makes the edge of the texture stretch instead of repeat.
>     vec2 upper_bound_mask = step(vLocalRect.zw, vLocalPos);
> ```

hmm, this would give us a different position, wouldn't it?
previous code used perspective division here in FS, the new code uses non-perspective interpolated XY

---

*[wrench/reftests/transforms/perspective.png, line 0 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVVcjVgbFpfyrEltBH:-KzVVcjWeZllzsUduMjT:bp1j52c) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/wrench/reftests/transforms/perspective.png#L0)):*
so I assume you needed to change this because now it's no longer perspective...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2069#-:-KzVVlKjtJ1Ews3R1_wl:b-qrqwnj)*

glennw:
Review status: 11 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[webrender/res/prim_shared.glsl, line 652 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVS5umoGyNl63U2aqR:-KzVhz-ayfq9d5D0sirL:b-896fix) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L652)):*

can we just clamp to `clip_rect` instead?
</blockquote></details>

Done.

---

*[webrender/res/prim_shared.glsl, line 665 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVSlvliEzjjGm9WWwu:-KzVi-BxE1rh1lu5Yu3W:bqfpxga) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L665)):*

I'm still a bit confused about `4` here. I'd prefer us either to be precise for the simple case only (of 1:1 scale with the screen, so that the `extrude = 0.5`), or for the general case, but not for an arbitrary point in-between.
</blockquote></details>

It's a bit of extra work to determine if we have a super-simple case that needs a minimal extrude (this is only in the transform path, so a simple case with no rotation etc is already going through the path with no-extrude at all). I dropped it to 2 pixels for now, which seems to still give good results on weird perspective transforms, but reduce the extra edge pixels. I think a 2 pixel edge in the case of a rotation / perspective is OK to pay for now, as a conservative approximation, since we don't pay the cost for non-transformed primitives?

---

*[webrender/res/prim_shared.glsl, line 688 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVU5mz4yUyPMtX6Lia:-KzVif8EqND4g43FAUhU:b-p4ybu) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L688)):*

interesting, so this is changing the semantics: before we had translation to the target space followed by `uTransform`, now we have the opposite order
</blockquote></details>

Oops, fixed.

---

*[webrender/res/prim_shared.glsl, line 704 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVUUAGdL_xgC0Y7CKn:-KzVigZJbXP1b2TG_Ziv:b-osvcfi) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L704)):*

why is this changed?
</blockquote></details>

I think this was a bug - it meant that images weren't using the local rect to get AA edges for. This was causing the extra edge drawing on the perspective reftest below. Unless I'm misunderstanding something?

---

*[webrender/res/prim_shared.glsl, line 823 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVUeEFPe_vBr1Y-YtX:-KzVizVVVGGzapNiFb2Z:b-896fix) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L823)):*

sweeet!
</blockquote></details>

Done.

---

*[webrender/res/ps_image.glsl, line 87 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVVDrJCKACIyIMfExZ:-KzVj-WJ6YMcacWSqNG1:b9vf002) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/ps_image.glsl#L87)):*

hmm, this would give us a different position, wouldn't it?
previous code used perspective division here in FS, the new code uses non-perspective interpolated XY
</blockquote></details>

I don't think so - that value should now be perspective interpolated, right? Since we provide the W value from the vertex shader?

---

*[wrench/reftests/transforms/perspective.png, line  at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVVcjVgbFpfyrEltBH:-KzVj9gb2_ndiHsT73Yu:bkw5c03) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/wrench/reftests/transforms/perspective.png#L)):*

so I assume you needed to change this because now it's no longer perspective...
</blockquote></details>

There's a few tiny differences in the pixel output, but mostly this was due to the fix of the default mask value you commented on above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2069)*

glennw:
@kvark Thanks for the review. I've fixed up or answered those items. I'll hold off on a try run until the discussions above are resolved (not sure if those answer your questions or if more changes are needed). Once I get r=you, then I'll do a try run and merge if green.

glennw:
@kvark I did a follow up try with those changes:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=d3e1167029802cc03f329358cabbb671ad2eabd4

All green.

kvark:
:lgtm:

---

Reviewed 4 of 4 files at r3.

---

*[webrender/res/prim_shared.glsl, line 665 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVSlvliEzjjGm9WWwu:-KzWvd30j4oSJSjbTS9C:b-dfo0ud) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L665)):*

It's a bit of extra work to determine if we have a super-simple case that needs a minimal extrude (this is only in the transform path, so a simple case with no rotation etc is already going through the path with no-extrude at all). I dropped it to 2 pixels for now, which seems to still give good results on weird perspective transforms, but reduce the extra edge pixels. I think a 2 pixel edge in the case of a rotation / perspective is OK to pay for now, as a conservative approximation, since we don't pay the cost for non-transformed primitives?
</blockquote></details>

strictly speaking, `sqrt(2)/2` should be enough for rotations only. We'll follow up with possibly better heuristic later, as discussed on IRC

---

*[webrender/res/prim_shared.glsl, line 704 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVUUAGdL_xgC0Y7CKn:-KzWwCDLqvEpcVfGQsjQ:b-nokuen) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/prim_shared.glsl#L704)):*

I think this was a bug - it meant that images weren't using the local rect to get AA edges for. This was causing the extra edge drawing on the perspective reftest below. Unless I'm misunderstanding something?
</blockquote></details>

1.0 means using instance rect for clipping, 0.0 means using the primitive rect
I believe I put 1.0 there for compatibility with the old code, so this is perfectly fine to fix/reconsider here

---

*[webrender/res/ps_image.glsl, line 87 at r1](https://reviewable.io:443/reviews/servo/webrender/2069#-KzVVDrJCKACIyIMfExZ:-KzWy6OOxDzrkCz4FzYp:b-ugoobc) ([raw file](https://github.com/servo/webrender/blob/3f2464a2424c3006faf135502fa43f4124876523/webrender/res/ps_image.glsl#L87)):*

I don't think so - that value should now be perspective interpolated, right? Since we provide the W value from the vertex shader?
</blockquote></details>

I forgot that old `W==1`, thanks for the explanation :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2069#-:-KzWyQyZ-6k-KxE5xPBE:bnfp4nl)*

kvark:
Great stuff, and thanks for addressing the comments!
@bors-servo r+

staktrace:
@glennw This PR caused some regressions on the gecko side, I confirmed the range for https://bugzilla.mozilla.org/show_bug.cgi?id=1420748 and https://bugzilla.mozilla.org/show_bug.cgi?id=1420754 but likely https://bugzilla.mozilla.org/show_bug.cgi?id=1421526 is also the same problem.

kvark:
Thanks @staktrace ! I moved this into a separate issue  #2133

PR #2070, opened 2017-11-21
Add a simple but important filter optimization.

glennw:
Real world pages often contain filters that have no effect,
such as opacity(1) or invert(0).

Make use of the picture tree to detect these after animation
properties have been resolved. If we detect a filter that
has no effect, don't bother creating a render task - just
collapse the primitives runs in this picture into the
current picture surface.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2070)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -356,7 +356,32 @@ impl PicturePrimitive {
 
                         self.render_task_id = Some(render_tasks.add(picture_task));
                     }
-                    Some(PictureCompositeMode::Filter(..)) | Some(PictureCompositeMode::Blit) => {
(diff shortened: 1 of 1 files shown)


glennw:
r? @kvark or @nical 

Pending try run (I haven't checked the results yet):
https://hg.mozilla.org/try/rev/7b2fea57ebd42d67a433705a3148d8ccd3bf6fee
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7b2fea57ebd42d67a433705a3148d8ccd3bf6fee

glennw:
Try looks green.

PR #2072, opened 2017-11-21
Use DebugFlags for time and sample queries.

nical:
And also disabling batching while we are at it.
Debug flags make it easier to interface with gecko (no need to update the bindings or anything, we can even land the gecko glue before updating webrender). `flag_changed` will hopefully make it more convenient to use DebugFlags by default.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2072)


glennw:
Looks good, thanks!

@bors-servo r+

PR #2075, opened 2017-11-21
Main FB clear fix for subtraction overflow

kvark:
An attempt to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1418315#c9

The clear rect may end up negative because the document rectangle is provided by the user and we don't guarantee it being inside the main framebuffer anywhere. Adjusting it for the clear should be safe.

cc @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2075)


kvark:
@mrobinson thanks for taking a look! How about the fixed version?

PR #2076, opened 2017-11-21
Clamp the blur radius

### LINKED ISSUE #2061: Very large box shadows cause rendering bugs
Testcase: `data:text/html,<div style="height:20px;box-shadow:0 0 2000px 2000px rgba(0,0,0,0.4)">`

Loading this page in a WebRender Firefox renders a blank tab and makes text in the tab bar disappear. I think unreasonably large box-shadow blurs should be clamped, or the shadow should be rendered without a blur. In this case, the blur achieves nothing, but I ran into a website that did this anyway.

nc4rrillo:
Fixes #2061 
Still needs a reftest

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2076)

### DIFF
--- a/webrender/src/box_shadow.rs
+++ b/webrender/src/box_shadow.rs
@@ -17,6 +17,10 @@ use render_task::MAX_BLUR_STD_DEVIATION;
 // The blur shader samples BLUR_SAMPLE_SCALE * blur_radius surrounding texels.
 pub const BLUR_SAMPLE_SCALE: f32 = 3.0;
 
+// Maximum blur radius.
+// Taken from https://searchfox.org/mozilla-central/rev/c633ffa4c4611f202ca11270dcddb7b29edddff8/layout/painting/nsCSSRendering.cpp#4412
+pub const MAX_BLUR_RADIUS : f32 = 300.;
+
 // The amount of padding added to the border corner drawn in the box shadow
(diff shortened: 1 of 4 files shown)


nc4rrillo:
I added a reftest, it fails on the OSX build bot with the following:

```
REFTEST TEST-UNEXPECTED-FAIL | reftests/boxshadow/box-shadow-large-border-radius.yaml == reftests/boxshadow/box-shadow-large-border-radius.png | image comparison, max difference: 1, number of differing pixels: 2638
```



Maybe it needs to be fuzzed?

glennw:
Did you generate the reference image with ```./headless.py <yaml>```? That should *in theory* make the reference consistent with the CI machines, which all run tests with OSMesa software rasterizer.

nc4rrillo:
Oh nope, I did it using wrench itself. I'll regenerate.

mstange:
Looks good, but I think the test should be called `box-shadow-large-blur-radius.yaml` instead of `box-shadow-large-border-radius.yaml`. Would it be better to make the reference a YAML file that has a blur radius of 300, rather than a PNG?

nc4rrillo:
I didn't know you could have YAML files on both sides of the ref tests. Cool!
Updated.

r? @mstange

glennw:
Looks good, thank you!

@bors-servo r+

PR #2079, opened 2017-11-22
Fix clip corner optimization.

glennw:
We should only be drawing a full clip mask if we encounter a clip
rect that is *not* in a compatible coordinate system.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2079)

### DIFF
--- a/webrender/src/render_task.rs
+++ b/webrender/src/render_task.rs
@@ -195,7 +195,7 @@ impl ClipWorkItem {
         for &(ref clip, _) in clips {
             match *clip {
                 ClipSource::Rectangle(..) => {
-                    if self.has_compatible_coordinate_system(prim_coordinate_system_id) {
+                    if !self.has_compatible_coordinate_system(prim_coordinate_system_id) {
                         return MaskGeometryKind::Default;
                     }
                 },


glennw:
This is a large performance win in Servo and wrench reftests. It doesn't seem to have much impact on sites in Gecko, which is unexpected.

I suspect this may point to a bug elsewhere in filtering out redundant clips, so I'm going to investigate this in Gecko tomorrow, since clip mask rendering seems to be the worst bottleneck right now on sites with slow-ish GPU time.

Pending gecko try here (I haven't checked results yet):
https://hg.mozilla.org/try/rev/2625c3a5c4b110ee7eee85cae68e94d356663e4b

r? @kvark or @mrobinson

mrobinson:
r=me once the try job is finished.

webrender/src/render_task.rs
```diff
@@ -195,7 +195,7 @@ impl ClipWorkItem {
         for &(ref clip, _) in clips {
             match *clip {
                 ClipSource::Rectangle(..) => {
-                    if self.has_compatible_coordinate_system(prim_coordinate_system_id) {
+                    if !self.has_compatible_coordinate_system(prim_coordinate_system_id) {
```
kvark: nice catch!
btw, why do we still go to `Default` if `rounded_rect_count == 0`?

kvark:
the only failure looks like one fixed by #2075, which I verified the try push [doesn't have](https://hg.mozilla.org/try/file/2625c3a5c4b110ee7eee85cae68e94d356663e4b/gfx/webrender/src/renderer.rs#l2898).
@bors-servo r+

Issue #2001, opened 2017-11-07
Share box shadow intermediate tasks where possible.


glennw:
We should add a cache key for box shadow render tasks, and the associated blurs / downsample tasks. This would save GPU time on pages that use a large number of box shadows with the same parameters.

The cache key for the box shadow task should include the parameters that define the shadow task (e.g. blur radius, border radii, box shadow clip mode).

The cache key for the blur and down-sample tasks could be something like (shadow key, stage index) where stage index is a number from 0..X, incrementing for each stage of the blur / downsample process.

demo99:
I'll work on this.

glennw:
@demo99 Great, thanks! If you set up a benchmark with several identical box shadows as a wrench yaml file, and then hit ```O``` that will show the render target debugger. It should be fairly clear to see from that if the intermediate surfaces are being shared.

glennw:
This happened.

PR #2080, opened 2017-11-22
Add cache key for box shadow.

demo99:
It's for #2001. Try to share the render task for the box shadows which have the same properties.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2080)


demo99:
I will send a gecko try push soon. I'm not very sure if we want to use this way to hash the properties. Maybe you can take a look first.

r? @glennw

demo99:
gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=438e539b0a002fd2e083cec9d52d26b126b4a58f&selectedJob=147006137

Please ignore the debug build assertion. I forgot to remove the assertion in the gecko build. It looks like I need to change a test's fuzz value from 8490 to 8518.

glennw:

---

*[webrender/src/render_task.rs, line 150 at r1](https://reviewable.io:443/reviews/servo/webrender/2080#-KzeyZYnGN0Hc2J9uKOT:-KzeyZYnGN0Hc2J9uKOU:bbj6vci) ([raw file](https://github.com/servo/webrender/blob/b12d918b73217d37ba70b235e9c6560100effeb0/webrender/src/render_task.rs#L150)):*
> ```Rust
>     /// Draw the alpha mask for a shared clip.
>     CacheMask(ClipId),
>     CacheScaling(u64, TypedSize2D<i32, DevicePixel>),
> ```

I don't think we should use a u64 hash here. I think we should define a cache key type for each task type. For example, we would have something like the following for a box shadow:

```
#[derive(Hash, ...)
struct BoxShadowCacheKey {
    blur_radius: Au,
    spread_radius: Au,
    ...
}
```

That gives us a few benefits:
 - We don't need ordered float (since Au can be hashed).
 - Quantizing the radii to Au for the hash key avoids float inaccuracies that may prevent caching for very similar values.
 - Having a specific type here means that rustc will cause a compile error if we accidentally don't fill out a field of the hash key structure.

---

*[webrender/src/render_task.rs, line 151 at r1](https://reviewable.io:443/reviews/servo/webrender/2080#-KzezEChCZtDJBT4R_HO:-KzezEChCZtDJBT4R_HP:b1f4do0) ([raw file](https://github.com/servo/webrender/blob/b12d918b73217d37ba70b235e9c6560100effeb0/webrender/src/render_task.rs#L151)):*
> ```Rust
>     CacheMask(ClipId),
>     CacheScaling(u64, TypedSize2D<i32, DevicePixel>),
>     CacheBlur(u64, i32),
> ```

The i32 here could disappear and be part of a BlurCacheKey as a blur direction (similar to the comment above).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2080)*

webrender/src/box_shadow.rs
```diff
@@ -26,6 +27,17 @@ pub const MAX_BLUR_RADIUS : f32 = 300.;
 // blurred without being affected by the border radius.
 pub const MASK_CORNER_PADDING: f32 = 4.0;
 
+#[derive(Debug, Copy, Clone, Eq, Hash, PartialEq)]
+pub struct BoxShadowCacheKey {
+    pub blur_radius: Au,
+    pub spread_radius: Au,
+    pub offset_x: Au,
+    pub offset_y: Au,
+    pub color: PremultipliedColorF,
```
glennw: I don't think color needs to be part of the cache key - the box shadows are rendered as an alpha mask and the color is applied when drawing the mask to the screen.

webrender_api/src/display_item.rs
```diff
@@ -698,6 +700,22 @@ impl BorderRadius {
     }
 }
 
+impl Hash for BorderRadius {
```
glennw: This shouldn't be needed now.

demo99: If I remove the hash function, I will get the compilation errors:

313 |     pub top_left: LayoutSize,
       |     ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `bitflags::<unnamed>::hash::Hash` is not implemented for `f32`

Or should I pass all radii to BoxShadowCacheKey?

webrender/src/box_shadow.rs
demo99: Okay, I will remove color from the cache key.

glennw:
Looks good! Could we do a final gecko try run?

demo99:
try result: https://treeherder.mozilla.org/#/jobs?repo=try&revision=7aab3682d4c2a34874e7b1a853ebe0e26d8dc96d&selectedJob=148367248

There is an assertion error in R3. I should fix it first. It may mean we should add more attributes to the cache key.
 ERROR - thread 'WRRenderBackend#3' panicked at 'assertion failed: `(left == right)`
  INFO -   left: `116×18`,
  INFO -  right: `136×21`', /builds/worker/workspace/build/src/gfx/webrender/src/tiling.rs:1821:44
  INFO - stack backtrace:
  INFO -    0:     0x7fb1961b4e53 - std::sys::imp::backtrace::tracing::imp::unwind_backtrace::h8ed7485deb8ab958
  INFO -    1:     0x7fb1961af5a0 - std::sys_common::backtrace::_print::h3d4f9ea58578e60f
  INFO -    2:     0x7fb1961c2353 - std::panicking::default_hook::{{closure}}::h0088fe51b67c687c
  INFO -    3:     0x7fb1961c20c2 - std::panicking::default_hook::hf425c768c5ffbbad
  INFO -    4:     0x7fb1961c2856 - std::panicking::rust_panic_with_hook::h25b934bb4484e9e0
  INFO -    5:     0x7fb1961c26e4 - std::panicking::begin_panic::h59483e27e93d7bc6
  INFO -    6:     0x7fb1961c25e9 - std::panicking::begin_panic_fmt::h5f221297e8a3dbdb
  INFO -    7:     0x7fb195cb7ed1 - webrender::frame_builder::FrameBuilder::build::h169e5ac5b450681e
  INFO -    8:     0x7fb195bf8338 - webrender::render_backend::Document::render::hee66196b4647bc7a
  INFO -    9:     0x7fb195c1d5d2 - webrender::renderer::Renderer::new::{{closure}}::he24b38546c967780
  INFO -   10:     0x7fb195bcf9ad - std::sys_common::backtrace::__rust_begin_short_backtrace::h1544667d1e49b657
  INFO -   11:     0x7fb195bc8c6d - <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once::hab66552c3f8db0ba
  INFO -   12:     0x7fb195c8166d - std::panicking::try::do_call::h530c14b1e61be08a
  INFO -   13:     0x7fb1961c712b - __rust_maybe_catch_panic
  INFO - Redirecting call to abort() to mozalloc_abort

demo99:
Ah, I think I should put box shadow's width and height into the cache key. I'll run gecko try again and also update my branch in the meantime.

demo99:
gecko try looks good now: https://treeherder.mozilla.org/#/jobs?repo=try&revision=147d9bc14c17d776cfd395a2e5cd4ce0acde7687

glennw:
Looks good, thanks!

@bors-servo r+

PR #2081, opened 2017-11-22
RenderTargetMode with optional depth renderbuffer

kvark:
This PR figures out automatically if a depth buffer is needed for each target. This is needed in order to provide the following use case for Gecko: having a document covering full screen, with only (transparent, non-intersecting) rounded corners in it. The old code would force the depth clearing, which is expensive.

A nice side effect is that we are no longer allocating the depth buffer for alpha targets and potentially other things, which is saving us VRAM.

Gecko try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=a9d2832688547665dd928f9a513abfb9c76d50bd

cc @mstange 
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2081)

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -128,6 +128,7 @@ pub fn main_wrapper<E: Example>(
     let device_pixel_ratio = window.hidpi_factor();
     println!("Device pixel ratio: {}", device_pixel_ratio);
 
+    println!("Loading shaders...");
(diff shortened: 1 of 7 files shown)


glennw:
Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2081)*

glennw:
Looks good, thanks.

@bors-servo r+

PR #2089, opened 2017-11-23
Add screenshot support to the debugger

### LINKED ISSUE #1672: Add screenshot reading to the debugger.
It would be great to be able to pull screenshots across to the debugger - of the framebuffer, render targets and texture cache.

Basically this would involve handling the message in ```renderer```, calling glReadPixels to get the data, serializing that across the socket (either as binary or base64), and then assembling that as an image in the web debugger (I guess via a data uri).

An initial implementation could just allow viewing the images. Eventually it'd be great to have a image viewer component in the debugger that can pan / zoom on the image - this would make it really easy to debug issues at the pixel level.

nc4rrillo:
Fixes #1672 

A question I have is about getting the width/height for to pass to `glReadPixels`, the window size is on the `Document` object, but there can be multiple documents. Is there a better way of getting the viewport size? 

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2089)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -10,7 +10,7 @@ build = "build.rs"
 default = ["freetype-lib"]
 freetype-lib = ["freetype/servo-freetype-sys"]
 profiler = ["thread_profiler/thread_profiler"]
-debugger = ["ws", "serde_json", "serde", "serde_derive", "image"]
+debugger = ["ws", "serde_json", "serde", "serde_derive", "image", "base64"]
 
 [dependencies]
 app_units = "0.5.6"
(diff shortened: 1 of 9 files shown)


nc4rrillo:
`glReadPixels` reads the framebuffer flipped vertically, maybe I should add support for flipping it back. I also need to make it not show a broken image when the screenshot hasn't yet been fetched.

glennw:
There's a function in wrench that uses the ```image``` crate code to flip the image. You might like to just copy that code.

glennw:
I'm not sure that there's a better way to get the window size right now - that does seem a bit odd for it to be per-document.

nc4rrillo:
I did a `transform: scaleY(-1)` to flip the image in CSS. I also addressed the issue of it showing a broken image by making it conditionally render.

webrender/Cargo.toml
```diff
@@ -32,6 +32,8 @@ ws = { optional = true, version = "0.7.3" }
 serde_json = { optional = true, version = "1.0" }
 serde = { optional = true, version = "1.0" }
 serde_derive = { optional = true, version = "1.0" }
+image = { optional = true, version = "*" }
```
glennw: We should probably specify a version here?

glennw: @nc4rrillo Are you able to specify a version number here? Then we can get this merged :)

nc4rrillo:
Oops didn’t see this! I will update my PR first thing tomorrow 🙂

nc4rrillo:
@glennw this is ready to go

glennw:
Thanks!

@bors-servo r+

PR #2090, opened 2017-11-23
Improve the shader precache code.

glennw:
* In precache mode, also issue a draw call after compiling the shader.
  On some drivers, much of the work is deferred until the first
  draw call is issued.
* Record the time for each precaching compilation and log it.
* Add support for shader precache option to wrench.

These options help with benchmarking, to ensure there are no
stalls during benchmarks caused by shader compilation.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2090)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -924,7 +924,18 @@ impl LazilyCompiledShader {
         };
 
         if precache {
-            try!{ shader.get(device) };
+            let t0 = precise_time_ns();
+            let program = try!{ shader.get(device) };
+            let t1 = precise_time_ns();
+            device.bind_program(program);
+            device.draw_triangles_u16(0, 3);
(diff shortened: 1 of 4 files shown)


webrender/src/renderer.rs
```diff
@@ -924,7 +924,18 @@ impl LazilyCompiledShader {
...
 
         if precache {
-            try!{ shader.get(device) };
+            let t0 = precise_time_ns();
+            let program = try!{ shader.get(device) };
+            let t1 = precise_time_ns();
+            device.bind_program(program);
+            device.draw_triangles_u16(0, 3);
+            let t2 = precise_time_ns();
+            println!("[C: {:.1} ms D: {:.1} ms] Precache {} {:?}",
```
kvark: we are printing this for each and every program? that's quite a spew. Maybe change to `info!`?

PR #2091, opened 2017-11-23
Add drop shadow support.

mephisto41:
The major changes here are using hash map to store render_task_id in PicturePrimitive. In order to support drop shadow, we need add a primitive to multiple pictures. So, if we only stored one render_task_id, the result is wrong. Hence, I use a hash map to store multiple render_task_ids and use parent PrimitiveIndex as key.

Another change is brush_image now has third configuration which is take color target as source, but only use alpha of this target and multiply color as result. This is because drop shadow is similar to box shadow which take a alpha mask and a color, but drop shadow's alpha mask is the alpha channel of color target.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2091)


glennw:
I have some concerns over the hashmap - there's various assumptions built in that a primitive only exists in one picture. Sorry I didn't have a chance to write up more detailed notes yet - I'll write up some more info next week.

glennw:
Reviewed 9 of 13 files at r1.
Review status: 9 of 13 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091)*

glennw:
Instead of having to draw the primitives in a drop shadow twice, how about something like this:

The code for a drop shadow filter is very similar to a normal blur filter - that is, a normal Picture where the render_task_id for the picture points to the render task that contains the blur result, and the blur task points to a task which draws the contents of the Picture we are drop-shadowing.

Then, the tricky part is being able to draw the normal Picture contents as well, since that will be done in a much earlier pass. However, if we extend the functionality of ```SourceTexture::CacheRGBA8``` slightly, we can make it so that it's possible to select an intermediate surface from any previous pass. Then, we can store that contents render task in the Picture as well (we could rename the ```readback_render_task_id``` field to ```secondary_render_task_id``` perhaps). Finally, when we come to add a DropShadow picture to the batches, we first add the main render task (which has the blur result) and then another instance with the secondary render task (which is the main contents).

This would mean that we only ever have to draw the contents of a drop-shadow element once, which I think could be a good performance win in most cases.

Although it might be a little complex to implement the support for ```SourceTexture::CacheRGBA8``` mentioned above, I think it will make the rest of the patch a lot simpler. It's also something that might come in useful in the future for implementing / optimizing some other features.

Does that sound like it would work?

mephisto41:
Sounds like a very good solution. I'll start to implement it! Thanks.

mephisto41:
gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=569e33fdca16263ec09031dfc06ded23afc0cfe0&selectedJob=148673156

The long-chain.html failure looks like a known issue. Even if I remove the drop-shadow effect in long-chain, the result isn't correct. Looks like the result is wrong if filters contain opacity and other effects.

mephisto41:
The long-chain.html should be fixed by #2150

mephisto41:
I think this is ready to review. r? @glennw

webrender/src/internal_types.rs
```diff
@@ -45,6 +45,8 @@ pub enum SourceTexture {
     External(ExternalImageData),
     CacheA8,
     CacheRGBA8,
+    RenderTaskCacheA8(usize),
```
glennw: We could use a newtype here to give a bit of type safety, and more importantly, self-document what these fields are. For example:

```
struct RenderPassIndex(usize);

...
RenderTaskCacheA8(RenderPassIndex),
...
```

Or something like that?

webrender/src/render_task.rs
```diff
@@ -283,6 +283,7 @@ pub struct RenderTask {
     pub children: Vec<RenderTaskId>,
     pub kind: RenderTaskKind,
     pub clear_mode: ClearMode,
+    pub pass_index: usize,
```
glennw: This could use the ```RenderPassIndex``` mentioned above. If we make it ```Option<RenderPassIndex>``` we can ensure that when the later code reads the value, it has been correctly assigned to a pass.

webrender/src/renderer.rs
```diff
@@ -564,6 +564,11 @@ struct SourceTextureResolver {
     /// The current cache textures.
     cache_rgba8_texture: Option<Texture>,
     cache_a8_texture: Option<Texture>,
+
+    pass_rgba8_textures: FastHashMap<usize, usize>,
```
glennw: We could use a stronger type here, or at least a comment to describe what this is mapping between.

webrender/src/renderer.rs
```diff
@@ -595,27 +603,46 @@ impl SourceTextureResolver {
...
         &mut self,
         a8_texture: Option<Texture>,
         rgba8_texture: Option<Texture>,
-        pool: &mut Vec<Texture>,
+        pass_index: usize,
     ) {
         // If we have cache textures from previous pass, return them to the pool.
-        pool.extend(self.cache_rgba8_texture.take());
-        pool.extend(self.cache_a8_texture.take());
+        match self.cache_rgba8_texture {
```
glennw: This can be written as ```if self.cache_rgba8_texture.is_some()```

webrender/src/renderer.rs
```diff
@@ -595,27 +603,46 @@ impl SourceTextureResolver {
...
+            Some(..) => {
+                self.pass_rgba8_textures.insert(pass_index - 1, self.render_target_pool.len());
+                self.render_target_pool.extend(self.cache_rgba8_texture.take());
+            }
+            _ => {}
+        }
+        match self.cache_a8_texture {
```
glennw: Same here

glennw:
Looks like this has a reftest failure on CI too, which will need investigation. It *might* just be a fuzziness issue that requires updating the PNG reference.

mephisto41:
Review status: 4 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[webrender/src/internal_types.rs, line 48 at r2](https://reviewable.io:443/reviews/servo/webrender/2091#-L-5eDNUDgilgqL34YZ6-r2-48:-L-Vmq_66Y9KgFr7OL9o:b-896fix) ([raw file](https://github.com/servo/webrender/blob/f6983f7b3924c04ee8569315029bcf72a73f6cb7/webrender/src/internal_types.rs#L48)):*

We could use a newtype here to give a bit of type safety, and more importantly, self-document what these fields are. For example:

```
struct RenderPassIndex(usize);

...
RenderTaskCacheA8(RenderPassIndex),
...
```

Or something like that?

</blockquote></details>

Done.

---

*[webrender/src/render_task.rs, line 286 at r2](https://reviewable.io:443/reviews/servo/webrender/2091#-L-5eDNUDgilgqL34YZ7-r2-286:-L-VmtGpjI8HuDLBGRli:b-896fix) ([raw file](https://github.com/servo/webrender/blob/f6983f7b3924c04ee8569315029bcf72a73f6cb7/webrender/src/render_task.rs#L286)):*

This could use the ```RenderPassIndex``` mentioned above. If we make it ```Option<RenderPassIndex>``` we can ensure that when the later code reads the value, it has been correctly assigned to a pass.
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 568 at r2](https://reviewable.io:443/reviews/servo/webrender/2091#-Kzc9LktAhTwyuIvn3h5-r2-568:-L-VmubbHW8tBsNJP6WK:b-896fix) ([raw file](https://github.com/servo/webrender/blob/f6983f7b3924c04ee8569315029bcf72a73f6cb7/webrender/src/renderer.rs#L568)):*

We could use a stronger type here, or at least a comment to describe what this is mapping between.
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 632 at r2](https://reviewable.io:443/reviews/servo/webrender/2091#-Kzc9LktAhTwyuIvn3h5-r2-632:-L-VmuualcHmeOH5eori:b-896fix) ([raw file](https://github.com/servo/webrender/blob/f6983f7b3924c04ee8569315029bcf72a73f6cb7/webrender/src/renderer.rs#L632)):*

This can be written as ```if self.cache_rgba8_texture.is_some()```
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 639 at r2](https://reviewable.io:443/reviews/servo/webrender/2091#-Kzc9LktAhTwyuIvn3h5-r2-639:-L-VmvI2OcEZaQdaE-ch:b-896fix) ([raw file](https://github.com/servo/webrender/blob/f6983f7b3924c04ee8569315029bcf72a73f6cb7/webrender/src/renderer.rs#L639)):*

Same here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091)*

kvark:
Reviewed 4 of 13 files at r1, 3 of 11 files at r2, 8 of 9 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[webrender/src/frame_builder.rs, line 1753 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7gQSadBBzWmCYmvM:-L-Y7gQSadBBzWmCYmvN:bgppums) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/frame_builder.rs#L1753)):*
> ```Rust
>                 passes.push(RenderPass::new_off_screen(self.screen_rect.size.to_i32(), RenderPassIndex(idx)));
>             }
>             passes.push(RenderPass::new_main_framebuffer(self.screen_rect.size.to_i32(), RenderPassIndex(required_pass_count)));
> ```

Is this intentionally not `required_pass_count -1` ? If so, needs a comment

---

*[webrender/src/render_task.rs, line 20 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7vuqeySKWj_HMfgU:-L-Y7vuqeySKWj_HMfgV:bl3olq7) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/render_task.rs#L20)):*
> ```Rust
> use tiling::{RenderPass, RenderTargetIndex};
> use tiling::{RenderTargetKind};
> use internal_types::RenderPassIndex;
> ```

nit: let's keep those imports alphabetically ordered

---

*[webrender/src/renderer.rs, line 643 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y8Ml3hfh4a6I0jfjl:-L-Y8Ml3hfh4a6I0jfjm:bqedklz) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L643)):*
> ```Rust
>     ) {
>         // If we have cache textures from previous pass, return them to the pool.
>         if self.cache_rgba8_texture.is_some() {
> ```

```rust
if let Some(texture) = self.cache_rgba8_texture.take() {...}
```

---

*[webrender/src/renderer.rs, line 645 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y8ipd_P42hQRUwRNV:-L-Y8ipeMVzXh9at-TfA:bsrj2jn) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L645)):*
> ```Rust
>         if self.cache_rgba8_texture.is_some() {
>             self.pass_rgba8_textures.insert(
>                 RenderPassIndex(pass_index.0 - 1), RenderTargetPoolId(self.render_target_pool.len()));
> ```

why does this have `-1`?

---

*[webrender/src/renderer.rs, line 689 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6RZvzWOuyWxveUML:-L-Y6RZwg2G2pB_fWttC:b-aquv5s) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L689)):*
> ```Rust
>                 let texture = match self.pass_rgba8_textures.get(&pass_index) {
>                     Some(pool_index) => &self.render_target_pool[pool_index.0],
>                     None => &self.dummy_cache_texture,
> ```

is this legal? I'd expect us to panic here

---

*[webrender/src/renderer.rs, line 4028 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6p5uxIHy1I9oKqMu:-L-Y6p5uxIHy1I9oKqMv:b-z71jd3) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L4028)):*
> ```Rust
>         let mut size = 512;
>         let fb_width = framebuffer_size.width as i32;
>         let num_layers: i32 = self.texture_resolver.render_target_pool
> ```

what is the reasoning behind moving the RT pool under `TextureResolver`?

---

*[webrender/src/tiling.rs, line 1806 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7W5DRgFnFW-ltu5u:-L-Y7W5EeMze7e2GVIaN:b4tkdv9) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/tiling.rs#L1806)):*
> ```Rust
>     tasks: Vec<RenderTaskId>,
>     dynamic_tasks: FastHashMap<RenderTaskKey, DynamicTaskInfo>,
>     pass_index: RenderPassIndex,
> ```

would it be simpler to pass the index into `build` as opposed to storing here?

---

*[webrender/src/tiling.rs, line 1866 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7ENGs-kLlzJGHqWS:-L-Y7ENHUyEB1plgZxTN:b4ydbke) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/tiling.rs#L1866)):*
> ```Rust
>                     {
>                         let task = &mut render_tasks[task_id];
>                         task.pass_index = Some(self.pass_index);
> ```

`render_tasks.get_mut(task_id).unwrap().pass_index = ...`

---

*[wrench/reftests/filters/filter-drop-shadow.yaml, line 6 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y1O5zBrmml810pQmF:-L-Y1O6-5_XegYrGYXxM:b-6qalwt) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/wrench/reftests/filters/filter-drop-shadow.yaml#L6)):*
> ```YAML
>     - type: stacking-context
>       bounds: [100, 100, 400, 400]
>       filters: drop-shadow(73, 73, 20, 1, 0, 0, 1)
> ```

would be cleaner to keep the vector parts vectors, e.g. `drop-shadow([73, 73], 20, [1, 0, 0, 1])`
also, aren't we specifying the color in byte values everywhere else in YAML (except alpha component)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091)*

mephisto41:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[webrender/src/frame_builder.rs, line 1753 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7gQSadBBzWmCYmvM:-L-_mHgnKn2QEKr0mDJ4:bqi5xzd) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/frame_builder.rs#L1753)):*

Is this intentionally not `required_pass_count -1` ? If so, needs a comment
</blockquote></details>

I'll remove this line in next patch.

---

*[webrender/src/render_task.rs, line 20 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7vuqeySKWj_HMfgU:-L-_m7X_OakTGZusWvIg:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/render_task.rs#L20)):*

nit: let's keep those imports alphabetically ordered
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 643 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y8Ml3hfh4a6I0jfjl:-L-_mUTzDkwr68QjqsJ_:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L643)):*

```rust
if let Some(texture) = self.cache_rgba8_texture.take() {...}
```
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 645 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y8ipd_P42hQRUwRNV:-L-_mVJqUhSg4GXIiwVg:b5c0qx6) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L645)):*

why does this have `-1`?
</blockquote></details>

I added a comment. This is because the cache textures that return to pool are last pass's result. So we need ```-1``` for pass_index.

---

*[webrender/src/renderer.rs, line 689 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6RZvzWOuyWxveUML:-L-_mj2dLNHma6Is-Lxj:b-c2v8a4) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L689)):*

is this legal? I'd expect us to panic here
</blockquote></details>

Not every pass generated cache_texture. For example, if a RenderTargetList didn't has any targets, we don't build a texture for them.

---

*[webrender/src/renderer.rs, line 4028 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6p5uxIHy1I9oKqMu:-L-_oLkRyWlFp_POtBqW:b-lxadv8) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L4028)):*

what is the reasoning behind moving the RT pool under `TextureResolver`?
</blockquote></details>

Because we don't need to pass RT poll when calling ```TexutreResolver.bind``` or ```TextureResolver.resolve```.

---

*[webrender/src/tiling.rs, line 1806 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7W5DRgFnFW-ltu5u:-L-_olfcWrS5rJtOe9-Q:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/tiling.rs#L1806)):*

would it be simpler to pass the index into `build` as opposed to storing here?
</blockquote></details>

Done.

---

*[webrender/src/tiling.rs, line 1866 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y7ENGs-kLlzJGHqWS:-L-_olKlK4zdl4184p8Y:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/tiling.rs#L1866)):*

`render_tasks.get_mut(task_id).unwrap().pass_index = ...`
</blockquote></details>

Done.

---

*[wrench/reftests/filters/filter-drop-shadow.yaml, line 6 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y1O5zBrmml810pQmF:-L-_onR66iSlW0rdpWkG:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/wrench/reftests/filters/filter-drop-shadow.yaml#L6)):*

would be cleaner to keep the vector parts vectors, e.g. `drop-shadow([73, 73], 20, [1, 0, 0, 1])`
also, aren't we specifying the color in byte values everywhere else in YAML (except alpha component)?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091)*

kvark:
Reviewed 7 of 7 files at r5.

---

*[webrender/src/render_task.rs, line 139 at r5](https://reviewable.io:443/reviews/servo/webrender/2091#-L-bMy4HVMYEkRBeiGrB:-L-bMy4I5XyNf3l2lmc8:b451ck6) ([raw file](https://github.com/servo/webrender/blob/1a11c5e856ccfde03d267f2d15415e44f90fab07/webrender/src/render_task.rs#L139)):*
> ```Rust
>     }
> 
>     pub fn get_mut(&mut self, id: RenderTaskId) -> &mut RenderTask {
> ```

hmm, why is this needed? the caller should be able to use the array syntax

---

*[webrender/src/renderer.rs, line 689 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6RZvzWOuyWxveUML:-L-bP4wekJD8K9aprLxm:bi19pm4) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L689)):*
<details><summary><i>Previously, mephisto41 (Morris Tseng) wrote…</i></summary><blockquote>

Not every pass generated cache_texture. For example, if a RenderTargetList didn't has any targets, we don't build a texture for them.
</blockquote></details>

Right, but we are also not calling `bind()` (this function) for every pass, we should only be calling it where it makes sense and guaranteed to have the texture

---

*[wrench/src/parse_function.rs, line 21 at r5](https://reviewable.io:443/reviews/servo/webrender/2091#-L-bMFduVRQZ0168Nl6v:-L-bMFdvWB83XT1RYg5-:b-m6tgja) ([raw file](https://github.com/servo/webrender/blob/1a11c5e856ccfde03d267f2d15415e44f90fab07/wrench/src/parse_function.rs#L21)):*
> ```Rust
>     }
>     impl<'a> Parser<'a> {
>         fn skip_whitespace_and_bracket(&mut self) {
> ```

that's not ideal to skip brackets with whitespace... perhaps we could parse the bracket-enclosed value separately as opposed to flattening the list?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091)*

mephisto41:

---

*[webrender/src/render_task.rs, line 139 at r5](https://reviewable.io:443/reviews/servo/webrender/2091#-L-bMy4HVMYEkRBeiGrB:-L-eOa_P62msejtEtCLs:b-896fix) ([raw file](https://github.com/servo/webrender/blob/1a11c5e856ccfde03d267f2d15415e44f90fab07/webrender/src/render_task.rs#L139)):*

hmm, why is this needed? the caller should be able to use the array syntax
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 689 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6RZvzWOuyWxveUML:-L-eOcZg35U95Nt6t9_2:b-4i4rpt) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L689)):*

Right, but we are also not calling `bind()` (this function) for every pass, we should only be calling it where it makes sense and guaranteed to have the texture
</blockquote></details>

Ok, I got it. I added panic for it.

---

*[wrench/src/parse_function.rs, line 21 at r5](https://reviewable.io:443/reviews/servo/webrender/2091#-L-bMFduVRQZ0168Nl6v:-L-eOnfwqBUfKc5w8qF-:b-896fix) ([raw file](https://github.com/servo/webrender/blob/1a11c5e856ccfde03d267f2d15415e44f90fab07/wrench/src/parse_function.rs#L21)):*

that's not ideal to skip brackets with whitespace... perhaps we could parse the bracket-enclosed value separately as opposed to flattening the list?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091)*

kvark:
Thanks!
Please rebase, squash a bit, and answer my last few concerns, and we are good to go ;)

---

Reviewed 7 of 7 files at r6.

---

*[webrender/src/internal_types.rs, line 44 at r6](https://reviewable.io:443/reviews/servo/webrender/2091#-L-gjgZoux6cgM0gcad7:-L-gjgZpxpJpmrN8SMCl:b-92rdk) ([raw file](https://github.com/servo/webrender/blob/1370760c934431026a9829e3937db97f158d1957/webrender/src/internal_types.rs#L44)):*
> ```Rust
> // native texture ID.
> 
> // XXX Remove this once RenderTaskCacheA8 is used.
> ```

can we move it to the specified variant only (as opposed to the whole enum)?

---

*[webrender/src/renderer.rs, line 689 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6RZvzWOuyWxveUML:-L-gjTn3CqT0GXAjeaBM:b-22r3y5) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L689)):*
<details><summary><i>Previously, mephisto41 (Morris Tseng) wrote…</i></summary><blockquote>

Ok, I got it. I added panic for it.
</blockquote></details>

Thanks! nit: the `expect()` is idiomatic for the case where you match `Option` or `Result` expecting it to be `Some`/`Ok` correspondingly.

---

*[webrender/src/tiling.rs, line 1862 at r6](https://reviewable.io:443/reviews/servo/webrender/2091#-L-gjvRZmBLX-q2WZXjB:-L-gjvRZmBLX-q2WZXjC:b-5x830s) ([raw file](https://github.com/servo/webrender/blob/1370760c934431026a9829e3937db97f158d1957/webrender/src/tiling.rs#L1862)):*
> ```Rust
>                 for &task_id in &self.tasks {
>                     assert_eq!(render_tasks[task_id].target_kind(), RenderTargetKind::Color);
>                     render_tasks.tasks.get_mut(task_id.0 as usize).unwrap().pass_index = Some(pass_index);
> ```

does array syntax work here?

---

*[wrench/src/yaml_helper.rs, line 571 at r6](https://reviewable.io:443/reviews/servo/webrender/2091#-L-gk2oE7tEmrSj3wME9:-L-gk2oE7tEmrSj3wMEA:bethdk) ([raw file](https://github.com/servo/webrender/blob/1370760c934431026a9829e3937db97f158d1957/wrench/src/yaml_helper.rs#L571)):*
> ```Rust
>                     let mut yaml_doc = YamlLoader::load_from_str(&str).expect("Failed to parse drop-shadow");
>                     let yaml = yaml_doc.pop().unwrap();
>                     Some(FilterOp::DropShadow(yaml["offset"].as_vector().unwrap(),
> ```

looks much better :+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091#-:-L-gk66Nenm60gUPfUDO:b-r6l63v)*

mephisto41:

---

*[webrender/src/internal_types.rs, line 44 at r6](https://reviewable.io:443/reviews/servo/webrender/2091#-L-gjgZoux6cgM0gcad7:-L-irtLppjDSkeCmlLvU:b-896fix) ([raw file](https://github.com/servo/webrender/blob/1370760c934431026a9829e3937db97f158d1957/webrender/src/internal_types.rs#L44)):*

can we move it to the specified variant only (as opposed to the whole enum)?
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 689 at r4](https://reviewable.io:443/reviews/servo/webrender/2091#-L-Y6RZvzWOuyWxveUML:-L-irvdyLb4HeZo8Di4d:b-7lacan) ([raw file](https://github.com/servo/webrender/blob/e5d178237e4f468841bb9542bc59cb085387565b/webrender/src/renderer.rs#L689)):*

Thanks! nit: the `expect()` is idiomatic for the case where you match `Option` or `Result` expecting it to be `Some`/`Ok` correspondingly.
</blockquote></details>

Thanks for suggestion. I use expect for this.

---

*[webrender/src/tiling.rs, line 1862 at r6](https://reviewable.io:443/reviews/servo/webrender/2091#-L-gjvRZmBLX-q2WZXjB:-L-irzaDyxZD1m8XQw2Q:b2uqlu4) ([raw file](https://github.com/servo/webrender/blob/1370760c934431026a9829e3937db97f158d1957/webrender/src/tiling.rs#L1862)):*

does array syntax work here?
</blockquote></details>

Ah, array syntax works!!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091)*

kvark:
:lgtm:

---

Review status: 14 of 17 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2091#-:-L-izNfh91u-ePzvRoTp:bnfp4nl)*

PR #2092, opened 2017-11-23
Support hit testing for ImageMasks

### LINKED ISSUE #2065: Implement ClipSource::contains for missing ClipSource types
From https://bugzilla.mozilla.org/show_bug.cgi?id=1418830

It looks like some cases (perhaps only with gfx.webrender.hit-test set to true) we are entering ClipSource::contains with a ClipSource type that is not handled, so that function is panicking. We should handle whatever the missing case is.

mrobinson:
Now that Gecko is using WebRender hit testing we need to support this
type of ClipSource. For now we just check whether the point is inside
the masked region.

Fixes #2065.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2092)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -103,15 +103,17 @@ impl From<ClipRegion> for ClipSources {
 
 impl ClipSource {
     pub fn contains(&self, point: &LayerPoint) -> bool {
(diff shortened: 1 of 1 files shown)


mrobinson:
cc @staktrace

Issue #2093, opened 2017-11-23
Unable to find glyph key cache after recent text transform changes


mrobinson:
After #2058, when I try to run the recorded yaml from #1817. I see this error:

```
thread 'WRRenderBackend#0' panicked at 'BUG: Unable to find glyph key cache!', /checkout/src/libcore/option.rs:819:4
```
when I run

```
cargo run --release -- -p 1 show ~/Downloads/w3/w3.yaml
```

mrobinson:
cc @lsalzman

mstange:
I also see this crash when I open that page ([https://www.w3.org/conf/2013sf/](https://www.w3.org/conf/2013sf/)) in Firefox in a Firefox build that includes the [webrender update](https://bugzilla.mozilla.org/show_bug.cgi?id=1419440) that includes #2058.

mstange:
```
* thread #51, name = 'WRRenderBackend#1', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000102685235 libmozglue.dylib`mozalloc_abort(char const*) + 53
    frame #1: 0x0000000102685260 libmozglue.dylib`abort + 16
    frame #2: 0x000000010712e7c9 XUL`rust_panic + 9
    frame #3: 0x000000010712e19e XUL`std::panicking::rust_panic_with_hook::he7ea255c9426c7e7 + 366
    frame #4: 0x000000010712dfdc XUL`std::panicking::begin_panic::hb2bff8bfb74238d7 + 124
    frame #5: 0x000000010712df53 XUL`std::panicking::begin_panic_fmt::h7d0fd00c38594036 + 147
    frame #6: 0x000000010713b4eb XUL`core::panicking::panic_fmt::h584b3a57b69b22f9 + 91
    frame #7: 0x0000000107140c2e XUL`core::option::expect_failed::hc8bc355caf9b01a5 + 94
    frame #8: 0x0000000106db3dad XUL`webrender::glyph_cache::GlyphCache::get_glyph_key_cache_for_font::h2b7a4c398e37666f + 221
    frame #9: 0x0000000106db10e5 XUL`_$LT$webrender..tiling..ColorRenderTarget$u20$as$u20$webrender..tiling..RenderTarget$GT$::add_task::h0427ac6b858f4dbf + 693
    frame #10: 0x0000000106da655c XUL`webrender::render_backend::Document::render::h6c52c7947dada65c + 8972
[...]

PR #2094, opened 2017-11-23
ensure that we request glyphs without a transform for text shadows

lsalzman:
This passes down the rasterization space information into TextRunPrimitiveCpu::prepare_for_render, so that we can ensure glyph requests are kicked off with no transform for text shadows (as discussed with gw). Text shadows do not supply a any transform when calling fetch_glyphs, so if we don't make sure to request the glyphs without a transform, this will cause the crash in issue #2093. Crashtest supplied courtesy of mstange.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2094)


glennw:
Looks great, thanks!

Note to self: when we support selecting whether to rasterize a picture in screen space, we'll want to pull that value from the picture instead of hardcoding it in tiling.rs.

@bors-servo r+

lsalzman:
Fixes #2093

PR #2098, opened 2017-11-24
Flexible render target pool

kvark:
This is a follow-up to #2050, also a more generic solution in place of the fix in #2039.

With multiple documents, we can't rely on stack-based render target re-use, since each document would expect a different stack and will force the wasteful re-allocation of render targets. With this PR, we'll first do a pass to re-use all perfectly matching targets for all documents. Then when each document is rendered, it will grab the remaining targets and re-initialize some parts of them, if needed. This scheme should guarantee efficient re-use of render target space for multiple documents, which are supposed to be used in Gecko for chrome, status bar, main page, and more.

Note: target selection doesn't take the depth surface into account at the moment, so these may sometimes re-allocate inefficiently. This is something to keep an eye on and potentially follow-up.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2098)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -391,7 +391,6 @@ pub struct Texture {
     format: ImageFormat,
     width: u32,
     height: u32,
-
     filter: TextureFilter,
     render_target: Option<RenderTargetInfo>,
     fbo_ids: Vec<FBOId>,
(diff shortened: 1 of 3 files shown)


glennw:
Looks good, thanks!

@bors-servo r+

PR #2101, opened 2017-11-27
Fix the texture cache debug display.

glennw:
For now, we need to create an FBO for the texture cache layers,
since we use glBlitFramebuffer() to display them. We should
fix this up properly, but for now this at least makes the texture
cache debug display work again.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2101)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -10,7 +10,7 @@ use frame::FrameId;
 use freelist::{FreeList, FreeListHandle, UpsertResult, WeakFreeListHandle};
 use gpu_cache::{GpuCache, GpuCacheHandle};
 use internal_types::{CacheTextureId, TextureUpdateList, TextureUpdateSource};
-use internal_types::{SourceTexture, TextureUpdate, TextureUpdateOp};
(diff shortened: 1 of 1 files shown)


webrender/src/texture_cache.rs
```diff
@@ -581,7 +581,14 @@ impl TextureCache {
...
                     filter: texture_array.filter,
-                    render_target: None,
+                    // TODO(gw): Creating a render target here is only used
+                    //           for the texture cache debugger display. In
+                    //           the future, we should change the debug
+                    //           display to use a shader that blits the
+                    //           texture, and then we can remove this
+                    //           memory allocation (same for the other
+                    //           standalone texture below).
+                    render_target: Some(RenderTargetInfo { has_depth: false }),
```
kvark: I like that this is fixed, but I don't quite like the fact that are paying for the texture debug scenario in all the cases where debug is not needed.

glennw: I agree - it's a quick fix for now, but a proper fix would be helpful. I'll add help-wanted tag to the issue @nical opened for it.

PR #2102, opened 2017-11-27
Add support for brushes with transforms plus edge AA.

glennw:
This means that brush items such as box-shadows get correct edge
AA when they are transformed, which was not previously supported.

Also fix up the segment / rect AA support for transformed primitives.
Previously, we were using the clip rect or the local rect. This
works in most cases, but is not correct. Instead, explicitly pass
in the prim rect and the segment rect, and apply the local clip
rect to each of those. In most cases, these are the same, but for
some primitives they differ.

Add a basic test suite to ensure AA is working correctly on border,
box-shadow, rect and image primitives.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2102)


glennw:
r? @kvark 

All Servo / Wrench tests pass.

Gecko try is green (just one unrelated intermittent):
https://hg.mozilla.org/try/rev/81182a405c3edab9fa88e4722a56da1a8367aa90
https://treeherder.mozilla.org/#/jobs?repo=try&revision=81182a405c3edab9fa88e4722a56da1a8367aa90

This is the last bit of prep work before we can start switching old primitive types over to brush shaders. It's a happy side-effect that it also allows us to do proper AA on transformed box-shadows.

kvark:
:lgtm:
special thanks for the new primitive tests! 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2102#-:-Kzz0FhYuRp8zn6Bd_7Z:b-xuy7ee)*

PR #2104, opened 2017-11-27
Better ClipChain optimization

mrobinson:
Better ClipChain optimization

There are a lot of situations where we can omit redundant clips or
combine a series of clips into a single empty mask (which has the
effect of clipping the primitive in screen space). Briefly, some of the
optimizations this change includes:

1. When the inner rect of a clip completely encloses the outer rect
   of a parent, we can ignore it.
2. When the outer rectangle of a clip is completely enclosed by the
   inner rectangle of a parent, we can start a new ClipChan and
   ignore the parents.
3. If the combined inner rectangle of a ClipChain completely
   encloses the intersection of a primitive and the current clips,
   we can stop applying clips from that chain.
4. If a clip's inner rect completely encloses the intersection
   of a primitive and the current clips, we can ignore it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2104)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{BorderRadius, ComplexClipRegion, DeviceIntRect, ImageMask, ImageRendering, LayerPoint};
-use api::{ClipMode, LayerRect, LayerSize};
+use api::{ClipMode, LayerRect};
 use api::{LayerToWorldTransform, LayoutPoint, LayoutVector2D, LocalClip};
 use border::BorderCornerClipSource;
 use ellipse::Ellipse;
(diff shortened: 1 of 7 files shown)


mrobinson:
This is one half of https://github.com/servo/webrender/pull/2062.

glennw:
Awesome! I'll look at this first thing in the morning if no-one else gets to it first. The sooner we can get this part merged the better :)

glennw:
Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2104)*

glennw:
Looks good to me!

I've done a try push that was green yesterday:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d3e515665df920695200feff390c9ec4a52984ac

I don't think this patch has changed since then.

@bors-servo r+

glennw:
@kvark @jrmuizel @staktrace Once this lands, it's probably worth getting it into a Gecko update as soon as possible - it should improve the clip mask profiling on a lot of real world sites (and a few benchmarks too).

glennw:
@kvark @jrmuizel @staktrace Hmm, actually I'm seeing some strange behavior with extra large clip masks that seem redundant - it's probably unrelated to this - but might want to hold off on updating Gecko until I investigate them further.

glennw:
OK, there is a PR https://github.com/servo/webrender/pull/2115 which fixes those issues (pending green try run of course).

PR #2105, opened 2017-11-27
Add support for basic performance annotations to reftests.

glennw:
This basic implementation allows a reftest to optionally specifiy
the expected number of draw calls, alpha targets and color targets
that the test should use.

This can be used for things such as:
 * Ensuring primitives are batching together as expected.
 * Ensuring that tests use the expected number of render targets.
 * Ensuring that invisible filters are removed.
 * Ensuring that certain operations (e.g. opacity) get collapsed
   into a more efficient rendering strategy.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2105)


glennw:
r? @kvark or @nical 

If you have better suggestions for how to specify these in a cleaner way, or the output should be different, let me know.

nical:
Looks good to me! @bors-servo r+

glennw:
@staktrace Heads up that you'll need to change the ```Ok(())``` in bindings.rs to ```Ok(_)```.

PR #2107, opened 2017-11-27
Move profiler ui elements so that they are less likely to end up off screen.

nical:
On some of my screens the profiler ui doesn't fit vertically causing useful elements to be off screen. This PR makes this much less likely to happen.

I also tweaked the offset in a few places so that the ui lines up (some of the ui elements are positioned relative the the text with an inflated rectangle while others are positioned relative to the rectangle with a margin for the text).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2107)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -533,7 +533,9 @@ impl ProfileGraph {
     ) -> Rect<f32> {
         let size = Size2D::new(600.0, 120.0);
         let line_height = debug_renderer.line_height();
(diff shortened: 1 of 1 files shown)


nical:
screenshot from 2017-11-27 14-17-27

r? anyone

nical:
Now that I look at the screen shot I feel like it would be prettier to put the blocky elements like the graphs on the left and the text on the right instead of the opposite, but honestly I would rather spend time polishing this when I am bored on a plane. Right now I just want to see the GPU queries on all of my screens.

PR #2115, opened 2017-11-28
Fix some redundant cases of clip mask generation.

glennw:
There are two fixes here:
 * Image pictures were generating clip masks, but they are never
   actually used. In the future, we *might* want to support this,
   but at the moment it just results in clip masks that are then
   never used.
 * If a primitive is in the root coordinate system, then we know
   that it is axis-aligned, and that there are no coordinate
   system changes in between this primitive and the root. In that
   case, we can infer that the local_clip_rect of the clip scroll
   node will correctly handle the screen space clip rect.

In Gecko with the current WR code, I was seeing a lot of very large
clip masks being generated (4x extra A8 render targets) on HN. The
test case included here is a very reduced test case for these issues.

The test case makes use of the new reftest annotations to assert
that no alpha targets get allocated while rendering the yaml file.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2115)


glennw:
r? @mrobinson 

Do these fixes seem correct to you?

Gecko try run is pending:
https://hg.mozilla.org/try/rev/f17ce9d263711325d1fe2c3e485903d16307e98c
https://treeherder.mozilla.org/#/jobs?repo=try&revision=f17ce9d263711325d1fe2c3e485903d16307e98c

glennw:
PS: In manual testing, this does appear to greatly reduce the amount of redundant clips masks (though there are still quite a few) on many web pages.

glennw:
Even with this change, I'm still seeing heaps of clip masks being drawn when I wouldn't expect there to be. I don't have a good repro yet - I'll try to track this down more tomorrow.

glennw:
Gecko try run above looks green.

webrender/src/prim_store.rs
```diff
@@ -1385,26 +1393,32 @@ impl PrimitiveStore {
...
+                    let (rfid, may_need_clip_mask) = match pic.kind {
+                        PictureKind::Image { reference_frame_id, .. } => {
+                            (Some(reference_frame_id), false)
+                        }
+                        _ => {
+                            (None, true)
+                        }
                     };
-                    (Some((pic.pipeline_id, mem::replace(&mut pic.runs, Vec::new()), rfid)), pic.cull_children)
+                    (Some((pic.pipeline_id, mem::replace(&mut pic.runs, Vec::new()),rfid)),
```
mrobinson: Nit: Missing space in "Vec::new()),rfid"

wrench/reftests/performance/no-clip-mask.yaml
```diff
@@ -0,0 +1,19 @@
+---
+root:
+  items:
+    -
+      type: "stacking-context"
+      items:
+        -
+          bounds: [0, 111, 1887, 1365]
+          "clip-rect": [0, 111, 1887, 1365]
```
mrobinson: It might be worth leaving a comment here explaining why this test doesn't generate any clip masks.

mrobinson:
This does suggest that there is another bug. If the locally clipping of the layer is taking effect, we probably shouldn't see the outer rectangle of the clip shrinking at all. I'll investigate this today.

glennw:
Thanks, fixed up those nits. I do think there's another bug, like you said - since I see a lot of redundant clip masks still after this patch. If you want to hold off on merging this until you investigate and see if the bug fix also covers these issues, that's fine with me. Otherwise, we can merge this and revert if we find a better solution.

mrobinson:
@glennw I think it's better to merge this ASAP, since it's unclear how much work it will be to fix the deeper issue.

webrender/src/prim_store.rs
```diff
@@ -1322,6 +1322,14 @@ impl PrimitiveStore {
             .collect();
 
         if clips.is_empty() {
+            // If this item is in the root coordinate system, then
+            // we know that the local_clip_rect in the clip node
+            // will take care of applying this clip, so no need
+            // for a mask.
+            if prim_coordinate_system_id == CoordinateSystemId::root() {
```
kvark: I don't fully understand why this is needed. Wouldn't the same apply to any axis-aligned coordinate system, which should already be handled?

webrender/src/prim_store.rs
```diff
@@ -1476,7 +1490,7 @@ impl PrimitiveStore {
             (local_rect, xf_rect.bounding_rect)
         };
 
-        if !self.update_clip_task(
+        if perform_culling && may_need_clip_mask && !self.update_clip_task(
```
kvark: It doesn't appear obviously correct to me that we move the `perform_culling` check before the `update_clip_task`. This will result in stale `metadata.screen_rect` and `metadata.clip_task` being present in the rendered primitives.

webrender/src/prim_store.rs
glennw: It doesn't seem to handle it - @mrobinson thinks there might be a deeper issue that he is looking into. If / when that's resolved we could hopefully remove this.

webrender/src/prim_store.rs
glennw: ```perform_culling``` is effectively a constant - it never changes from initial value. Therefore, the clip task will never be set, and the screen rect should only depend on the previous culling values, not part of the clip stack.

PR #2120, opened 2017-11-28
support synthetic bold on Windows using DWrite font simulations

lsalzman:
This is necessary to fix downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1418564

In the process, I cleaned up a lot how certain settings (synthetic bold, embedded bitmaps) were communicated to the font backends. Since these were common to all backends, I made them a cleaner set of platform-independent flags, rather than having a bunch of different bools littered about randomly in FontInstancePlatformOptions.

I also modified dwrote (as evidenced in dwrote 0.4.1) to support a necessary function to clone a font face with the desired simulations for this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2120)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -241,7 +241,7 @@ dependencies = [
 
 [[package]]
 name = "dwrote"
-version = "0.4.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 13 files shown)


glennw:

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2120)*

glennw:
Looks good to me. I'll delegate and leave it to your judgment if we need a gecko try before merging this.

@bors-servo delegate+

PR #2126, opened 2017-11-28
Angle depth clear workaround

kvark:
First commit applies a simple heuristic to see if it's worth clearing the depth of the whole screen once as opposed to each document managing their depth. See: https://bugzilla.mozilla.org/show_bug.cgi?id=1419863

Second commit refactors `clear_rect` to be a little more DRY. Also adds a debug assertion to ensure we never try clearing the depth with disabled depth mask :) (happened to me, but no more!)

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2126)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -1603,29 +1603,11 @@ impl Device {
         self.frame_id.0 += 1;
     }
 
-    pub fn clear_target(&self, color: Option<[f32; 4]>, depth: Option<f32>) {
-        let mut clear_bits = 0;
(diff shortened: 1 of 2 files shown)


kvark:
@glennw good idea! launched https://treeherder.mozilla.org/#/jobs?repo=try&revision=010f493e60db17aa50da5a14a79538bba0a4f432

kvark:
All green!
@bors-servo r=glennw

PR #2130, opened 2017-11-29
Add border edges to the opaque pass where possible.

glennw:
This covers the common use cases of border edges (solid with opaque color).

This helps reduce draw calls significantly on some pages, and also
makes the batching code simpler (since opaque batching is much
faster than alpha batching, due to not needing to check overlap).

It also provides a GPU time win, since we're moving more pixels to
the opaque pass.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2130)

### DIFF
--- a/webrender/src/tiling.rs
+++ b/webrender/src/tiling.rs
@@ -415,17 +415,36 @@ fn add_to_batch(
         PrimitiveKind::Border => {
             let border_cpu =
                 &ctx.prim_store.cpu_borders[prim_metadata.cpu_prim_index.0];
-            // TODO(gw): Select correct blend mode for edges and corners!!
             let corner_kind = BatchKind::Transformable(
                 transform_kind,
                 TransformBatchKind::BorderCorner,
             );
-            let corner_key = BatchKey::new(corner_kind, blend_mode, no_textures);
+            // Selecting alpha blend for all corners is conservative,
+            // but no worse than what we previously had.
+            let corner_key = BatchKey::new(
+                corner_kind,
+                BlendMode::PremultipliedAlpha,
+                no_textures
+            );
             let edge_kind = BatchKind::Transformable(
                 transform_kind,
                 TransformBatchKind::BorderEdge,
             );
-            let edge_key = BatchKey::new(edge_kind, blend_mode, no_textures);
+            // If the edges are simple, then we only need blending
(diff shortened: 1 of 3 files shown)


glennw:
Pending try run here:

https://hg.mozilla.org/try/rev/a135773b7806f4ddb8cf4b7f0d232b2277c64336
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a135773b7806f4ddb8cf4b7f0d232b2277c64336

glennw:
This saves around 10% draw calls and compositor time on a number of web sites I tested. It's a stop-gap fix until we convert the borders over to be brush shaders.

webrender/src/tiling.rs
```diff
@@ -415,17 +415,36 @@ fn add_to_batch(
...
             let border_cpu =
                 &ctx.prim_store.cpu_borders[prim_metadata.cpu_prim_index.0];
-            // TODO(gw): Select correct blend mode for edges and corners!!
             let corner_kind = BatchKind::Transformable(
                 transform_kind,
                 TransformBatchKind::BorderCorner,
             );
-            let corner_key = BatchKey::new(corner_kind, blend_mode, no_textures);
+            // Selecting alpha blend for all corners is conservative,
+            // but no worse than what we previously had.
```
kvark: s/alpha/premultiplied alpha/
also, why is this no worse?

kvark:
Try push shows a few tests with 1-value differences. I guess this can be explained by the different draw order for those simple borders that we see now.

webrender/src/tiling.rs
glennw: Before this patch every border was considered transparent, and so would always end up in the alpha pass for all segments.

glennw:
Hmm, those failures are a bit odd. We could fuzz them, or we could hold off on this - it should be superseded once the border + brush changes I'm working on are ready (maybe 2-3 days work). @jrmuizel Do we want to land this with fuzz for now for a small performance win, or hold off?

jrmuizel:
I don't feel strongly, but it's fine to wait

glennw:
If we're not in a massive rush for this, I'll close for now - I'll re-open and investigate the failures if we need performance wins before the border + brush changes are ready.

PR #2136, opened 2017-11-29
Render task tree view in debug server

kvark:
MOARE debugging!

It would be great to somewhat automate the implementation of Json messages being passed for our structs, but for now the manual code is sufficient.

r? @glennw 

wr-render-tasks

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2136)

### DIFF
--- a/debugger/js/app.js
+++ b/debugger/js/app.js
@@ -2,8 +2,9 @@ var state = {
     connected: false,
     page: "options",
     passes: [],
+    render_tasks: [],
     documents: [],
-    clipscrolltree: [],
+    clip_scroll_tree: [],
     screenshot: []
 }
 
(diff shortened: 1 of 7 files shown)


glennw:
Looks good, but needs a rebase? r=me once that is done.

pcwalton:
This is awesome! I was literally just thinking today that it would be good to have this.

I wonder if we could generate Graphviz files too :) (obviously that shouldn't block this work from landing)

kvark:
@glennw oh, I rebased but forgot to push it afterwards, d'oh :)
Should be good now.

kvark:
btw, I think we should get rid of the `root` in all the trees and instead just list the documents right away

PR #2137, opened 2017-11-29
Tidy up a few parts of the frame builder code.

glennw:
This is just prep work for the upcoming change to switch
rectangle primitives over to be brush shaders. It shouldn't
have any functional effect, but is a bit easier to review
isolated from the rest of the changes.

* Remove EdgeAaSegmentMask from the public API crate, make internal.
* Remove EdgeAaSegmentMask from the display list types.
* Pass EdgeAaSegmentMask explicitly to add_solid_rectangle.
* Add add_clear_rectangle and add_scroll_bar. Make RectangleContent more of an internal detail.
* Remove the serialize attributes from EdgeAaSegmentMask.
* Don't add scrollbars and clear rects to the hit testing list.
* Don't try to split clear rectangles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2137)

### DIFF
--- a/webrender/src/border.rs
+++ b/webrender/src/border.rs
@@ -3,13 +3,14 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{BorderSide, BorderStyle, BorderWidths, ClipAndScrollInfo, ColorF};
-use api::{EdgeAaSegmentMask, LayerPoint, LayerRect};
+use api::{LayerPoint, LayerRect};
 use api::{LayerPrimitiveInfo, LayerSize, NormalBorder, RepeatMode};
 use clip::ClipSource;
 use ellipse::Ellipse;
 use frame_builder::FrameBuilder;
 use gpu_cache::GpuDataRequest;
-use prim_store::{BorderPrimitiveCpu, RectangleContent, PrimitiveContainer, TexelRect};
+use internal_types::EdgeAaSegmentMask;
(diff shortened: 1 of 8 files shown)


mrobinson:
Clear rectangles are added to the hit testing list to allow clients to specify regions that can be hit tested, but not displayed. I think this is necessary for both Gecko and Servo, unfortunately.

glennw:
@mrobinson  I think you're thinking of solid rectangles with alpha == 0? They are still added to the hit testing list. The only ones that aren't are Clear rectangles, which are a special blend mode rectangle that Gecko uses for some mac-specific effects.

mrobinson:
@glennw Oh, sorry. You're correct. I was confused by the term "clear rect." 😄

glennw:
@mrobinson Yea, the terminology here is confusing to say the least :)

mrobinson:
I think in this case they weren't added because they don't have an ItemTag (there's an early return), but I agree that explicit API in frame_buffer.rs makes this much clearer.

kvark:
I see a bag of great incremental changes, ship it!

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2137#-:-L-91EH96H5wdy7BmU_t:b9jci3r)*

kvark:
A try push wouldn't harm here, but the chances of breakage are low. Proceeding...
@bors-servo r+

PR #2141, opened 2017-11-30
Add a debug overlay showing the pipeline epochs of the current frame.

nical:
Handy when debugging APZ jumpiness, otherwise there is no way to correlate logging with a particular frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2141)


nical:
screenshot from 2017-11-30 18-07-40

r? anyone

PR #2142, opened 2017-11-30
First round of ClipScrollTree optiizations

mrobinson:
These optimizations reduce the prominence of `ClipScrollTree::update_tree` in my perf CPU profile.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2142)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -69,7 +69,7 @@ vec4[2] fetch_from_resource_cache_2(int address) {
 
 #ifdef WR_VERTEX_SHADER
 
-#define VECS_PER_LAYER              11
+#define VECS_PER_LAYER              7
 #define VECS_PER_RENDER_TASK        3
 #define VECS_PER_PRIM_HEADER        2
 #define VECS_PER_TEXT_RUN           3
@@ -143,7 +143,6 @@ vec4 fetch_from_resource_cache_1(int address) {
 
 struct ClipScrollNode {
     mat4 transform;
-    mat4 inv_transform;
     vec4 local_clip_rect;
     vec2 reference_frame_relative_scroll_offset;
     vec2 scroll_offset;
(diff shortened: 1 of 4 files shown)


emilio:
(You could use `Rc<RefCell<..>>`, though, I guess... Though it's kinda ugly)

mrobinson:
@emilio: Thanks. I'll just omit that commit for now and try a reference counted type tomorrow. Not sure what I was thinking!

glennw:
This looks good to me. We should do a gecko try run first, then r=me if all green.

mrobinson:
The try job is sitting at 99% with only localization test yet to finish (it's been running all day). There is one failure, but it is a JS test timeout (known intermittent). I think it's okay to hand this one to bors.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=672c6645e527433f72a0939ce4f981cc93ba56ef&group_state=expanded

webrender/res/prim_shared.glsl
```diff
@@ -548,9 +539,11 @@ vec4 get_layer_pos(vec2 pos, Layer layer) {
     // get a point on the layer plane
     vec4 ah = layer.transform * vec4(0.0, 0.0, 0.0, 1.0);
     vec3 a = ah.xyz / ah.w;
+
     // get the normal to the layer plane
-    vec3 n = transpose(mat3(layer.inv_transform)) * vec3(0.0, 0.0, 1.0);
-    return untransform(pos, n, a, layer.inv_transform);
+    mat4 inv_transform = inverse(layer.transform);
```
kvark: I don't think it's a good idea to compute the inverse mat4 per vertex.

mrobinson: So I definitely want to do some more testing here, but I believe this is only called for clips. In any case, it's probably pretty easy to move this back to the CPU only for nodes that need it.

PR #2146, opened 2017-12-01
Convert rectangle from prim to brush shader and support segments.

glennw:
Add support for brushes to be drawn as a list of segments that
are part of a larger primitive. Initially, these segments are
nine-patches only, although they may be expanded in the future
to work with generic segment rects.

This allows large brush primitives to be split into a number
of segments that can be split among the opaque and alpha passes,
which means that large rectangles are often rendered with the
inner rectangle as part of the opaque pass. Additionally, edge
segments are rendered in the opaque pass when the primitive does
not have a rotation or perspective transform applied.

When a rectangle is split, the clip tasks are generated only
where required for individual segments. This can significantly
reduce the amount of allocated area for clip masks, and also
the amount of GPU time spent drawing clip masks.

Previously, rectangle splitting could only work on LocalClip
items specified as part of the item, which meant that rectangles
in Gecko would never be split. Now, we calculate splitting
during prepare_prim_for_render, consulting the current clip chain
for the primitive. This means that we get the benefits of rectangle
splitting in Gecko, which doesn't use LocalClip API.

Having rectangle splitting handled as segments also means that we
don't add extra top-level primitives that take extra time to do
visibility culling on, saving some CPU time.

Future work:
 - Support more cases of primitive splitting.
 - Port images and gradients to be brushes so they automatically get splitting.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2146)


glennw:
For an example of what this changes, consider the rounded-rects reftest in wrench:

rr1

Previously, the clip mask render target looked like this, and each of the primitives was drawn completely in the alpha pass:

rr4

Now, there is a clip mask rendered for each clip region of interest, so we render a lot less clip mask pixels:

rr2

And, when we draw the main scene, the green segments are drawn in the opaque pass, while the red segments are drawn in the alpha pass:

rr3

glennw:
r? @kvark and/or @nical and/or @mrobinson 

(I still need to do a gecko try for this and fix any failures, but this is ready for review). Anecdotally, this makes clip mask generation time negligible on most real sites I tested on (exception where there are large inset box shadows, which still need work).

kvark:
It's been a long way... the ideas we discussed in Hawaii took quite a bit to finally land. Amazing work!
I'm happy with the changes, anticipating to see how Gecko is running blazing fast :)
Left my feedback for your consideration.

---

Reviewed 18 of 18 files at r1, 1 of 1 files at r2, 3 of 7 files at r3.
Review status: 18 of 22 files reviewed at latest revision, 19 unresolved discussions.

---

*[webrender/res/brush.glsl, line 40 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Epr0HHwq_ik1F1SBl:-L-Epr0ICgtjZO5Wxpy8:b72l584) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L40)):*
> ```GLSL
>     int clip_address;
>     int z;
>     int segment_kind;
> ```

the `flags` there were one per edge, given the need to apply AA on some of the edges and not the others. Is this not the case any more?

---

*[webrender/res/brush.glsl, line 93 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Eqbx2imu0_hlAhFoo:-L-Eqbx3qgyDub98jSbl:b-ljpft2) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L93)):*
> ```GLSL
> 
>     vec2 p0 = brush_prim.local_rect.p0;
>     vec2 p1 = brush_prim.local_rect.p0 + brush_prim.offsets.xy;
> ```

need some comments on what `p0-3` are exactly

---

*[webrender/res/brush.glsl, line 102 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Er6C9Bw8MK0pqmkgN:-L-Er6CAKqTioy_JdOGf:bf4asyc) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L102)):*
> ```GLSL
>             break;
> 
>         case SEGMENT_TOP_LEFT:
> ```

I wonder if it would make sense to separate X and Y here, i.e. each can be computed independently and needs 2 bits (0 - whole primitive, 1 - left/top, 2 - right/bottom, 3 - center). They could even share the same helper routine with a `switch` statement.

---

*[webrender/res/brush.glsl, line 139 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-ErVJ0Kkq6foOQDH7H:-L-ErVJ0Kkq6foOQDH7I:b95ux0c) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L139)):*
> ```GLSL
>     switch (brush_prim.aa_kind) {
>         case AA_KIND_SEGMENT:
>             break;
> ```

`edge_aa_segment_mask` seems to be uninitialized by this point?

---

*[webrender/src/border.rs, line 412 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EstSwXMe9qzaQYUka:-L-EstSwXMe9qzaQYUkb:bvixz21) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/border.rs#L412)):*
> ```Rust
>                     &inner_rect,
>                     Some(&[
>                         BrushSegmentKind::MidLeft,
> ```

why are we only including the middle here?

---

*[webrender/src/box_shadow.rs, line 304 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EtKQ-egiqbShNJEzz:-L-EtKQ-egiqbShNJF--:bbta4xv) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/box_shadow.rs#L304)):*
> ```Rust
>                         },
>                         None,
>                         BrushAntiAliasMode::Default,
> ```

I think `Primitive/Segment` distinction would be cleaner for this enum than `Default/Segment`

---

*[webrender/src/prim_store.rs, line 256 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EvK_NphvqTpIA8F0E:-L-EvK_NphvqTpIA8F0F:b-kksoqg) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L256)):*
> ```Rust
>     pub bottom_right_offset: LayerVector2D,
>     pub segments: [BrushSegment; 9],
>     pub enabled_segments: u16,
> ```

I see that you are concerned with this structure size for the fact of wrapping `Option<Box<_>>` to store it and pass it around. I think it's much bigger than it needs to be. Technically we only need a bit mask (`enabled_segments`) and 4 points to reconstruct all the needed segments wherever they are needed. This would be so small it wouldn't require a box.

---

*[webrender/src/prim_store.rs, line 263 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EvpMJ29ZLSidg7Ox2:-L-EvpMJ29ZLSidg7Ox3:brwsr55) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L263)):*
> ```Rust
>         outer_rect: &LayerRect,
>         inner_rect: &LayerRect,
>         valid_segments: Option<&[BrushSegmentKind]>,
> ```

How about having a bitmask of the brush segment kinds, and accept the bitmask here directly instead of the slice?

---

*[webrender/src/prim_store.rs, line 274 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EuwaCFlgX-2siS9_j:-L-EuwaDpKBt3G6hrOck:b-pwihwg) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L274)):*
> ```Rust
>                 let mut enabled_segments = 0;
> 
>                 for segment in valid_segments {
> ```

this is begging for `fold()` :)

---

*[webrender/src/prim_store.rs, line 1487 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Ex-KTxls7YvKGjZM_:-L-Ex-KTxls7YvKGjZMa:b-25u1u7) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1487)):*
> ```Rust
>                 let brush = &mut self.cpu_brushes[metadata.cpu_prim_index.0];
> 
>                 if brush.segments.is_none() && metadata.local_rect.size.area() > 256.0 * 256.0 {
> ```

let's have the constant lower and named somewhere

---

*[webrender/src/prim_store.rs, line 1493 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-ExQcyZ2yGqW9fkNuw:-L-ExQczqBo-UUafbqV0:b-uba2es) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1493)):*
> ```Rust
>                             let local_clips = clip_store.get_opt(&clip.clip_sources).expect("bug");
>                             for &(ref clip, _) in &local_clips.clips {
>                                 if let &ClipSource::RoundedRectangle(ref rect, ref radii, mode) = clip {
> ```

should put the `ClipMode::Clip` right in here

---

*[webrender/src/prim_store.rs, line 1501 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-ExiWNsyW4xCYR3FIH:-L-ExiWNsyW4xCYR3FII:b-bp2j8h) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1501)):*
> ```Rust
>                                         );
>                                         if brush.segments.is_some() {
>                                             break;
> ```

what happens with others in `local_clips.clips`? do we just ignore them here? I imagine we should only be splitting with a single clip

---

*[webrender/src/prim_store.rs, line 1511 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Exuh0-YYkpwcTtLum:-L-Exuh0-YYkpwcTtLun:bjmz2b6) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1511)):*
> ```Rust
> 
>                 match brush.segments {
>                     Some(ref mut segments) => {
> ```

let's rename it to `segment_desc` or something... too many `segments` :)

---

*[webrender/src/prim_store.rs, line 1513 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Ey2toqvgGNDNksgaH:-L-Ey2toqvgGNDNksgaI:b53hkxf) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1513)):*
> ```Rust
>                     Some(ref mut segments) => {
>                         for (i, segment) in segments.segments.iter_mut().enumerate() {
>                             segment.clip_task_id = if i <= BrushSegmentKind::BottomLeft as usize {
> ```

why is the comparison made with `BottomLeft`?

---

*[webrender/src/prim_store.rs, line 1559 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EyKL3N2MzIJ6S6CGv:-L-EyKL3N2MzIJ6S6CGw:b-ia3kd2) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1559)):*
> ```Rust
>             }
>             _ => {
>                 let clip_task = RenderTask::new_mask(
> ```

looks to be the same code as the arm above. Can we merge?

---

*[webrender/src/prim_store.rs, line 1904 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Eysad3tkbzHAz_f59:-L-Eysad3tkbzHAz_f5A:b-dy7yjd) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1904)):*
> ```Rust
>     radii: &BorderRadius
> ) -> Option<Box<BrushSegmentDescriptor>> {
>     let inner = match extract_inner_rect_safe(local_clip_rect, radii) {
> ```

another thing we could possibly do: if we know the primitive is axis-aligned, we should allow T-junctions and thus generate 3 opaque parts instead of 5

---

*[webrender/src/prim_store.rs, line 1905 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Eyj2FWIJoWQlktsaF:-L-Eyj2FWIJoWQlktsaG:b-kvfou0) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1905)):*
> ```Rust
> ) -> Option<Box<BrushSegmentDescriptor>> {
>     let inner = match extract_inner_rect_safe(local_clip_rect, radii) {
>         Some(rect) => rect,
> ```

nit: just map this result instead of the early return

---

*[webrender/src/tiling.rs, line 429 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EzYcXKAD_yR_oNmep:-L-EzYcXKAD_yR_oNmeq:b-3s7t72) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/tiling.rs#L429)):*
> ```Rust
>                             let needs_blending = !prim_metadata.opacity.is_opaque ||
>                                                  segment.clip_task_id.is_some() ||
>                                                  (!is_inner && transform_kind == TransformedRectKind::Complex);
> ```

why do we need blending for non-centric opaque segments?

---

*[webrender/src/tiling.rs, line 2157 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EzkX_lb5vPV1pezZQ:-L-EzkX_lb5vPV1pezZR:by6fkfu) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/tiling.rs#L2157)):*
> ```Rust
>         }
>     }
> }
> ```

nit: newline

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2146#-:-L-EznfKYXIBGdbFEhH7:byk4gff)*

glennw:
Review status: 14 of 22 files reviewed at latest revision, 19 unresolved discussions.

---

*[webrender/res/brush.glsl, line 40 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Epr0HHwq_ik1F1SBl:-L-FlUUAQuHNb1_6lECt:bx8j02q) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L40)):*

the `flags` there were one per edge, given the need to apply AA on some of the edges and not the others. Is this not the case any more?
</blockquote></details>

We can derive the correct flags from the segment kind.

---

*[webrender/res/brush.glsl, line 93 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Eqbx2imu0_hlAhFoo:-L-FladAk6rLo0_qRfgv:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L93)):*

need some comments on what `p0-3` are exactly
</blockquote></details>

Done.

---

*[webrender/res/brush.glsl, line 102 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Er6C9Bw8MK0pqmkgN:-L-Flc_AKjtoUUMyluwE:b-8ww4u2) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L102)):*

I wonder if it would make sense to separate X and Y here, i.e. each can be computed independently and needs 2 bits (0 - whole primitive, 1 - left/top, 2 - right/bottom, 3 - center). They could even share the same helper routine with a `switch` statement.
</blockquote></details>

Yep, that would probably be tidier - OK if we do that as a follow up?

---

*[webrender/res/brush.glsl, line 139 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-ErVJ0Kkq6foOQDH7H:-L-FlkLlT6lEqK-K9FZv:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L139)):*

`edge_aa_segment_mask` seems to be uninitialized by this point?
</blockquote></details>

Done.

---

*[webrender/src/border.rs, line 412 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EstSwXMe9qzaQYUka:-L-FllH4aGqHohDKhg5a:b-xevyui) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/border.rs#L412)):*

why are we only including the middle here?
</blockquote></details>

This matches the previous code - the parts that add the top and bottom rectangles handle adding the corners, so we only need to do that in one place.

---

*[webrender/src/box_shadow.rs, line 304 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EtKQ-egiqbShNJEzz:-L-FltlEvjjhQ8VrkX7q:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/box_shadow.rs#L304)):*

I think `Primitive/Segment` distinction would be cleaner for this enum than `Default/Segment`
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 256 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EvK_NphvqTpIA8F0E:-L-FlvZMc9EaLPmk_7BT:bmnijzw) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L256)):*

I see that you are concerned with this structure size for the fact of wrapping `Option<Box<_>>` to store it and pass it around. I think it's much bigger than it needs to be. Technically we only need a bit mask (`enabled_segments`) and 4 points to reconstruct all the needed segments wherever they are needed. This would be so small it wouldn't require a box.
</blockquote></details>

I did originally implement it this way. I changed to a local_rect representation for a couple of reasons. First, we need to store an individual clip task ID for each segment, so we need an array anyway (although in the future we could perhaps support allocating a range of render task IDs with a guaranteed contiguous range). Second, we need to transform these local rects on every scroll frame (to get the new device rect for the clip task) and so I thought it might be worth calculating the local rects once and storing them. We could definitely profile the difference though - there's pros and cons of each method, I think.

---

*[webrender/src/prim_store.rs, line 263 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EvpMJ29ZLSidg7Ox2:-L-FmR--vLXgSzOzUzI8:bysgxj5) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L263)):*

How about having a bitmask of the brush segment kinds, and accept the bitmask here directly instead of the slice?
</blockquote></details>

I was using bitmasks originally for that - probably a good idea to switch back to them. Are you ok with doing that another time?

---

*[webrender/src/prim_store.rs, line 274 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EuwaCFlgX-2siS9_j:-L-FmiT4vnWqIX28MBgv:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L274)):*

this is begging for `fold()` :)
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1487 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Ex-KTxls7YvKGjZM_:-L-FmjP6sr2UGTfj0ICk:b-vxmflf) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1487)):*

let's have the constant lower and named somewhere
</blockquote></details>

Changed it to 128x128 and made it a constant.

---

*[webrender/src/prim_store.rs, line 1493 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-ExQcyZ2yGqW9fkNuw:-L-Fmn94xlsIByXNTwaC:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1493)):*

should put the `ClipMode::Clip` right in here
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1501 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-ExiWNsyW4xCYR3FIH:-L-Fmo-2bz7Uezbj86Q6:bccxfap) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1501)):*

what happens with others in `local_clips.clips`? do we just ignore them here? I imagine we should only be splitting with a single clip
</blockquote></details>

Yes, you're right - I had originally thought we'd handle this by detecting the case and generating a clip mask for each segment in that case. But for now, I've just done what you suggested and limited it to only one rounded rect clip. We can optimize this in the future by handling more cases.

---

*[webrender/src/prim_store.rs, line 1511 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Exuh0-YYkpwcTtLum:-L-Fn0BuOek8IYtDb2lU:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1511)):*

let's rename it to `segment_desc` or something... too many `segments` :)
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1513 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Ey2toqvgGNDNksgaH:-L-Fn16IsLeAY09bnLfG:b-afxjxr) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1513)):*

why is the comparison made with `BottomLeft`?
</blockquote></details>

Due to the ordering of the enum, and selecting clip masks only for the corners. I added a comment in the code to make this clearer.

---

*[webrender/src/prim_store.rs, line 1559 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EyKL3N2MzIJ6S6CGv:-L-Fn7ZoMxPywUOyCofF:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1559)):*

looks to be the same code as the arm above. Can we merge?
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1904 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Eysad3tkbzHAz_f59:-L-FnDHXOe6R5HzcjAoQ:b-ozdi73) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1904)):*

another thing we could possibly do: if we know the primitive is axis-aligned, we should allow T-junctions and thus generate 3 opaque parts instead of 5
</blockquote></details>

True - although then we will need to rebuild this if the transform changes kind due to property animation. Might be worth considering in the future though.

---

*[webrender/src/prim_store.rs, line 1905 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Eyj2FWIJoWQlktsaF:-L-FnMNUTp7ZqoGPPoPw:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1905)):*

nit: just map this result instead of the early return
</blockquote></details>

Done.

---

*[webrender/src/tiling.rs, line 429 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EzYcXKAD_yR_oNmep:-L-FnN_m8llZ2Ht1WSha:b-yu9u1i) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/tiling.rs#L429)):*

why do we need blending for non-centric opaque segments?
</blockquote></details>

For edge segments that have a transform, we need blending enabled (when not axis-aligned) in order to allow edge AA blending.

---

*[webrender/src/tiling.rs, line 2157 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EzkX_lb5vPV1pezZQ:-L-Fn_8Va0-f4wlKEVVl:b-896fix) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/tiling.rs#L2157)):*

nit: newline
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2146)*

glennw:
@kvark OK, all review comments addressed, I think. Gecko try had a few failures, so investigating those now.

glennw:
Hmm, a lot of failures in the last try run, but they all look like they are probably caused by the same bug. I'll try to set up a repro in wrench and then do another try run after that.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=14f0c4323c9d3b0845dc7db305c9b8b2473c1834

kvark:
:lgtm: (minus the failures)

---

Reviewed 2 of 7 files at r3, 6 of 6 files at r4.

---

*[webrender/res/brush.glsl, line 102 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Er6C9Bw8MK0pqmkgN:-L-HZJlTz8Skjv0qq28_:b-637xzt) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L102)):*

Yep, that would probably be tidier - OK if we do that as a follow up?
</blockquote></details>

Yes, I thought I'd just share the idea while reviewing. Perfect for a follow-up :)
Would be nice to add a TODO/TRY comment near the code to not forget.

---

*[webrender/res/brush.glsl, line 139 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-ErVJ0Kkq6foOQDH7H:-L-HZVwektomybVyyoVN:ban4o1v) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/res/brush.glsl#L139)):*

Done.
</blockquote></details>

Ah, so this is going to be a regression then? Until we implement this TODO

---

*[webrender/src/prim_store.rs, line 256 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EvK_NphvqTpIA8F0E:-L-HcgSbCzfYV8Hs1kSq:b-xtc6ad) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L256)):*

I did originally implement it this way. I changed to a local_rect representation for a couple of reasons. First, we need to store an individual clip task ID for each segment, so we need an array anyway (although in the future we could perhaps support allocating a range of render task IDs with a guaranteed contiguous range). Second, we need to transform these local rects on every scroll frame (to get the new device rect for the clip task) and so I thought it might be worth calculating the local rects once and storing them. We could definitely profile the difference though - there's pros and cons of each method, I think.
</blockquote></details>

Ok, I see. Thanks for the info! We can revisit this later.

---

*[webrender/src/prim_store.rs, line 263 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EvpMJ29ZLSidg7Ox2:-L-HaDiuIhQopjct5OjX:bb7oryy) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L263)):*

I was using bitmasks originally for that - probably a good idea to switch back to them. Are you ok with doing that another time?
</blockquote></details>

sure

---

*[webrender/src/prim_store.rs, line 274 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-EuwaCFlgX-2siS9_j:-L-HaGwbZS01_5VvoPv1:b7cyz14) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L274)):*

Done.
</blockquote></details>

beautiful, thanks!

---

*[webrender/src/prim_store.rs, line 1904 at r1](https://reviewable.io:443/reviews/servo/webrender/2146#-L-Eysad3tkbzHAz_f59:-L-HcMZhlCAQQvWPR9Mb:be9yaaf) ([raw file](https://github.com/servo/webrender/blob/591883a2c1847508a2a2b908ac5d8120458a75a6/webrender/src/prim_store.rs#L1904)):*

True - although then we will need to rebuild this if the transform changes kind due to property animation. Might be worth considering in the future though.
</blockquote></details>

Yes, as with other cases we discussed before - if there is a transform property we should just treat it as transformed all the time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2146#-:-L-HcqMmzjJ5PsVkAszx:b-j5vnp4)*

mrobinson:
This is really cool.

glennw:
OK, rebased and fixed one bug in the Gecko reftests. Doing a gecko try run shortly to see if that solves all the failures or if there's another bug.

kvark:
r=me when grass is green!

glennw:
There is one test failure remaining.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=57423ca51e3c54d8baba954ff5832d48f5b05c26

I'll look into this on Monday and post an updated patch for that.

glennw:
Latest gecko try is all green:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=c18afcae2b73f63391130e4998b9aee3fe1c6d82

The last commit fixes a bug when the selected clip and the primitive are in a different (but still compatible) coordinate space. In this case (which the scroll node indices are different), calculate and apply a relative offset between the spaces. This is safe since we know that the transforms are in a compatible space, so no rotation / skew / perspective is involved.

r? on the last commit @kvark or @mrobinson 

Once I get r= I'll squash these commits before merge.

webrender/src/prim_store.rs
```diff
@@ -1510,10 +1512,28 @@ impl PrimitiveStore {
                                     }
                                 }
                             }
-                            if let Some((rect, radii)) = selected_clip {
+                            if let Some((rect, radii, clip_scroll_node_data_index)) = selected_clip {
+                                // If the scroll node transforms are different between the clip
+                                // node and the primitive, we need to get the clip rect in the
+                                // local space of the primitive, in order to generate correct
+                                // local segments. Since we know that the coordinate systems
+                                // are compatable from above, there's no need to do a full
```
pcwalton: typo: "compatible"

webrender/src/clip_scroll_node.rs
```diff
@@ -520,7 +520,8 @@ impl ClipScrollNode {
                     state.nearest_scrolling_ancestor_viewport
                        .translate(&info.origin_in_parent_reference_frame);
 
-                if !info.resolved_transform.preserves_2d_axis_alignment() {
+                if !info.resolved_transform.preserves_2d_axis_alignment() ||
+                   info.resolved_transform.has_perspective_component() {
```
mrobinson: Hrm...does this work well with 90 degree rotations?

glennw: Ah, good point! I updated the patch to do a full relative transform calculation and transform the rect. That code path almost never gets hit in reality, so the performance difference should be negligible.

kvark:
:lgtm:

---

Reviewed 2 of 4 files at r5, 6 of 7 files at r7, 1 of 1 files at r9.

---

*[wrench/reftests/transforms/segments-bug-ref.yaml, line 1 at r9](https://reviewable.io:443/reviews/servo/webrender/2146#-L-UOXTb7w2P27byFQOa:-L-UOXTcBZxViu-B18oF:b-9vibbi) ([raw file](https://github.com/servo/webrender/blob/e64ff36831bb2f9222f3d99b85842c691e87e6bf/wrench/reftests/transforms/segments-bug-ref.yaml#L1)):*
> ```YAML
> ---
> ```

would be nice to have a more descriptive name of the reftest an/or some comments on what it's testing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2146#-:-L-UNMcsA0u_alqRVDbw:bnfp4nl)*

glennw:
@kvark OK, updated the reftest with some comments describing the test. r=you one squashed?

glennw:
(squashed into one commit - I think this is ready to go now).

kvark:
@glennw thank you!
@bors-servo r+

glennw:
@jrmuizel @staktrace It'd be great to get this into a Gecko update this week, if possible. There's a heap of possible optimizations to come based on this, but this is a start that provides quite a significant win on most real sites already.

jrmuizel:
That's happening in https://bugzilla.mozilla.org/show_bug.cgi?id=1422317

PR #2150, opened 2017-12-01
Reverse apply filters.

mephisto41:
The order of ```composite_ops.filters``` is as same as the order of yaml. But we need reverse apply it to the picture primitive so that the rendering order would match the order of yaml.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2150)


glennw:
Can we do a gecko try for this?

mephisto41:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=b39f556af9961e569d916a46dce08e3d9f285f6d

glennw:
Looks like a few failures (although some of them may be intermittent / unrelated). It'd be good to find out what the cause is though - especially the failures in R7.

mephisto41:
```moz-element.html``` is unxpected-pass. And ```same-filter.html``` need change fuzzy values (only add 1 to the max difference value). Others are intermittent. I think it is good to land.

glennw:
Sounds good, thanks!

@bors-servo r+

PR #2152, opened 2017-12-01
don't override Mac operating system AA settings

lsalzman:
It is possible to disable LCD font smoothing in the operating system settings. By calling CGContext.set_allows_foo() on these we are overriding that, whereas Gecko normally honors them. Just remove these particular AA overrides so that we match Gecko on this. This addresses Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1421041

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2152)


mstange:
r+

lsalzman:
@glennw travis stalled out?

PR #2155, opened 2017-12-01
fake bold for Mac font backend

### LINKED ISSUE #1653: Implement Synthetic Bold Support
Some fonts don't natively support bold, and force the browser to synthesize them. The synthesis is fairly simple: draw each glyph multiple times, with horizontal offset. In fact servo/gecko could just feed us multiple copies of each char to do it for us... *except* that wouldn't produce the correct result for transparent text. Requiring the browser to generate a stacking context with transparency would be wasteful, I think.

"synthetic bold" should be made a font instance key, as introduced in https://github.com/servo/webrender/pull/1602

Sadly, I don't have a good synthetic-bold-needing example yet.

lsalzman:
Mac is the only platform that did not yet support synthetic bold. This solves it by using multi-striking in the same way Gecko does. This resolves #1653 and https://bugzilla.mozilla.org/show_bug.cgi?id=1394570

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2155)

### DIFF
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -82,9 +82,9 @@ impl FontTransform {
     pub fn compute_scale(&self) -> Option<(f64, f64)> {
         let det = self.determinant();
         if det != 0.0 {
-            let major = (self.scale_x as f64).hypot(self.skew_y as f64);
-            let minor = det.abs() / major;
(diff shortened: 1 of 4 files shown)


lsalzman:
Clarified the major/minor scales to be x_scale/y_scale for better readability.

PR #2156, opened 2017-12-02
Next set of ClipScrollTree optimizations

mrobinson:
This seems to cut the prominence of ClipScrollTree updates in my profile a great deal.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2156)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -2,16 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{BorderRadius, ComplexClipRegion, DeviceIntRect, ImageMask, ImageRendering, LayerPoint};
-use api::{ClipMode, LayerRect};
-use api::{LayerToWorldTransform, LayoutPoint, LayoutVector2D, LocalClip};
+use api::{BorderRadius, ClipMode, ComplexClipRegion, DeviceIntRect, ImageMask, ImageRendering};
+use api::{LayerPoint, LayerRect, LayerToWorldTransform, LayoutPoint, LayoutVector2D, LocalClip};
 use border::BorderCornerClipSource;
 use ellipse::Ellipse;
 use freelist::{FreeList, FreeListHandle, WeakFreeListHandle};
 use gpu_cache::{GpuCache, GpuCacheHandle, ToGpuBlocks};
 use prim_store::{ClipData, ImageMaskData};
 use resource_cache::ResourceCache;
-use util::{extract_inner_rect_safe, MaxRect, TransformedRect};
+use util::{MaxRect, calculate_screen_bounding_rect, extract_inner_rect_safe};
 
 pub type ClipStore = FreeList<ClipSources>;
 pub type ClipSourcesHandle = FreeListHandle<ClipSources>;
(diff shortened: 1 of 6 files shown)


mrobinson:
Here is the Gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=51369c294686d137fc80ace6ef8032bd5208ad50&group_state=expanded

97% done and no failures.

glennw:
I'd like to get #2146 landed before we land this, if possible (doing a gecko try now, hopefully it will be ready after that), since they will conflict a bit.

This looks awesome from a quick scan though, it should help with CPU performance a lot!

glennw:
Reviewed 6 of 6 files at r1.

---

*[webrender/src/prim_store.rs, line 1661 at r1](https://reviewable.io:443/reviews/servo/webrender/2156#-L-UuEjPbDMeA3O7Lu0D:-L-UuEjQ8Wbo2SuhyyZ0:b3pj9pp) ([raw file](https://github.com/servo/webrender/blob/3aa471129f5ed3c2a28abd978c60548467e2529b/webrender/src/prim_store.rs#L1661)):*
> ```Rust
> }
> 
> fn get_local_bounding_rect(
> ```

This uses ```transform_point3d``` while the version it's replaced with in ```util.rs``` uses ```transform_point2d```. Perhaps that's not actually relevant here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2156)*

glennw:
Just one question, otherwise this looks good to me (it will need a rebase after #2146 lands too).

mrobinson:

---

*[webrender/src/prim_store.rs, line 1661 at r1](https://reviewable.io:443/reviews/servo/webrender/2156#-L-UuEjPbDMeA3O7Lu0D:-L-VyHK0AMQQyqEED9_b:bmpmi99) ([raw file](https://github.com/servo/webrender/blob/3aa471129f5ed3c2a28abd978c60548467e2529b/webrender/src/prim_store.rs#L1661)):*

This uses ```transform_point3d``` while the version it's replaced with in ```util.rs``` uses ```transform_point2d```. Perhaps that's not actually relevant here?
</blockquote></details>

I think this should be okay. Both methods are using the same code to calculate the x and y coordinates of the final point.

```
    #[inline]
    pub fn transform_point3d(&self, p: &TypedPoint3D<T, Src>) -> TypedPoint3D<T, Dst> {
        let x = p.x * self.m11 + p.y * self.m21 + p.z * self.m31 + self.m41;
        let y = p.x * self.m12 + p.y * self.m22 + p.z * self.m32 + self.m42;
        let z = p.x * self.m13 + p.y * self.m23 + p.z * self.m33 + self.m43;
        let w = p.x * self.m14 + p.y * self.m24 + p.z * self.m34 + self.m44;

        point3(x/w, y/w, z/w)
}
```

```   
#[inline]
    pub fn transform_point2d(&self, p: &TypedPoint2D<T, Src>) -> TypedPoint2D<T, Dst> {
        let x = p.x * self.m11 + p.y * self.m21 + self.m41;
        let y = p.x * self.m12 + p.y * self.m22 + self.m42;

        let w = p.x * self.m14 + p.y * self.m24 + self.m44;

        point2(x/w, y/w)
```

transform_point2d should be *slightly* more efficient.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2156)*

mrobinson:
Okay. I've rebased this branch as well as adding another, very simple, cleanup commit for util.rs.

Issue #2132, opened 2017-11-29
MotionMark bouncing circles is worse the second time it's run


jrmuizel:
On our AMD Window 10 machine the score will go from around 400 to 80. @kvark thinks we might be fragmenting the GPU cache. I suggested it might be useful to expose the size of the gpu cache in the profiler hud to see if that might be the persistent state that's causing the change here.

kvark:
Related - #2110, which was originally filed to address the issue, but it doesn't capture the extent in full.

I believe what's happening here is that GPU cache gets fragmented. Since we are uploading the whole rows, we end up transferring too much data, hitting Angle's soft spots. Steps to take:
  1. confirm the hypothesis by looking at the GPU cache structure. If we don't have a way to inspect it, we should add one
  2. add statistics to tell us how much data is uploaded, and what ratio of it is the valuable payload
  3. consider implementing GPU cache de-fragmentation step

kvark:
We confirmed that the GPU cache texture quickly grows from 1024x512 to 1024x2048, and then the performance still degrade with time. This is supporting the fragmentation hypothesis but not confirming it just yet.

kvark:
I took a look at the GPU cache usage in MM ramping complexity using the new shiny bars of #2139
I can see that 773 rows are allocated, and about 1/4 or 1/3 of them are updated each frame in peak usage. This is waaay more than we need to, and no wonder it's inefficient. The uploaded actual block count is so small it's hard to tell, certainly below 5%.

glennw:
The GPU cache uploads an entire row at a time, as soon as any block within that row is dirty. This can (worst case) mean a 16 kB upload for a row if a single 16-byte block has changed (!). This is one of those things that was meant to be a "come back later an optimize when we have useful data" :)

The problem with just issuing an update for every block-run that has changed is that I was seeing the CPU time get swamped by the number of API calls to update small blocks. That may no longer be the case (we should check), but if it is, we'll need to do something a bit smarter.

A couple ideas we could try:
 * We could try just reducing the width of the GPU cache texture. This will limit the max size of a text run, but we do split text runs for this case now, and even a 256 wide texture would be big enough for the vast majority of text runs to not be split. This is probably the simplest / hackiest solution, but might be worth experimenting with.

[...]

PR #2162, opened 2017-12-04
Scattered GPU cache updates

### LINKED ISSUE #2110: PBO copies are slow on Angle
It appears that Windows/Angle spends tons of time in `update_texture_from_pbo`: https://perfht.ml/2zJnkg4
Angle decided to defer mapping and filling an actual D3D11 staging buffer up until this call. Since the copy itself is done in a draw call, it has to stall the GPU until the copy is done. This is one issue, but the actual current one is different: `Map` itself is waiting. I suppose it's waiting for the GPU considering the texture is still in use, which implies our PBO orphaning doesn't work as expected (see `Device::orphan_pbo`).

cc @jrmuizel @glennw

kvark:
Fixes #2110

This PR introduces the GPU cache uploading via shader writes. It minimizes the amount of data we need to transfer to the GPU in order to reduce the stalls and make us scale better with content (see https://github.com/servo/webrender/issues/2132#issuecomment-348735499).

Note: in terms of efficiency, rasterizing lines would be much better than points. Leaving this for the follow up. Doing so would require us to upload the source data into a texture as opposed to a buffer, and slightly (but non-critically) complicate things.

WIP, because:
  - benchmarking is TODO
  - to be squashed eventually

Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=f583b2deae1c7312437638bdda6ddc65693d53ed

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2162)

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -136,6 +136,7 @@ pub fn main_wrapper<E: Example>(
         precache_shaders: E::PRECACHE_SHADERS,
         device_pixel_ratio,
         clear_color: Some(ColorF::new(0.3, 0.0, 0.0, 1.0)),
+        //scatter_gpu_cache_updates: false,
         ..options.unwrap_or(webrender::RendererOptions::default())
     };
 
--- a/webrender/res/gpu_cache_update.glsl
+++ b/webrender/res/gpu_cache_update.glsl
@@ -2,21 +2,26 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#include base
+
 varying vec4 vData;
 
 #ifdef WR_VERTEX_SHADER
-attribute vec4 aValue;
-attribute vec2 aPosition;
+in vec4 aValue;
+in vec2 aPosition;
 
 void main() {
 	vData = aValue;
 	gl_Position = vec4(aPosition * 2.0 - 1.0, 0.0, 1.0);
+	gl_PointSize = 1.0;
 }
 
 #endif //WR_VERTEX_SHADER
 
 #ifdef WR_FRAGMENT_SHADER
+out vec4 oValue;
+
 void main() {
-	gl_FragColor = vData;
+	oValue = vData;
 }
 #endif //WR_FRAGMENT_SHADER
--- a/webrender/res/gpu_cache_update.glsl
+++ b/webrender/res/gpu_cache_update.glsl
@@ -1,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+varying vec4 vData;
+
+#ifdef WR_VERTEX_SHADER
+attribute vec4 aValue;
+attribute vec2 aPosition;
+
+void main() {
+	vData = aValue;
+	gl_Position = vec4(aPosition, 0.0, 1.0);
+}
+
+#endif //WR_VERTEX_SHADER
+
+#ifdef WR_FRAGMENT_SHADER
+void main() {
+	gl_FragColor = vData;
+}
+#endif //WR_FRAGMENT_SHADER
--- a/webrender/res/gpu_cache_update.glsl
+++ b/webrender/res/gpu_cache_update.glsl
@@ -11,9 +11,9 @@ in vec4 aValue;
 in vec2 aPosition;
 
 void main() {
-	vData = aValue;
-	gl_Position = vec4(aPosition * 2.0 - 1.0, 0.0, 1.0);
-	gl_PointSize = 1.0;
+    vData = aValue;
+    gl_Position = vec4(aPosition * 2.0 - 1.0, 0.0, 1.0);
+    gl_PointSize = 1.0;
 }
 
 #endif //WR_VERTEX_SHADER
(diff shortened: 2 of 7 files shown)


glennw:
I like the general idea of this, but I think that F32 targets are not renderable on ES3 - which would mean this isn't usable on ANGLE?

pcwalton:
Half precision float should be part of GLES3. It's part of WebGL 2. If you can get away with f16 render targets, you should be good.

kvark:
@pcwalton I don't see RGBA16F supported as color-renderable in GLES 3.1. However, it appears that we can use RGBA32UI and cast/reinterpret the floats to/from it, given that we need no filtering.

pcwalton:
Oh, I guess not. Well, all hardware I've tested supports half-float as an extension (the straggler being PowerVR, as it only supports R16F, not R32F).

glennw:
Unfortunately I don't think ANGLE exposes any extensions for that, which is primarily what this is for.

pcwalton:
Really? Pathfinder relies on that extension and works on my Windows laptop in WebGL…

glennw:
Huh, interesting! Which extension is it you're relying on? I'll check on my Windows laptop if it's present with ANGLE.

pcwalton:
@glennw Looks like it's these:

https://developer.mozilla.org/en-US/docs/Web/API/OES_texture_half_float

https://developer.mozilla.org/en-US/docs/Web/API/EXT_color_buffer_half_float

glennw:
OK, I can see that https://chromium.googlesource.com/chromium/src/gpu/+/master/GLES2/extensions/CHROMIUM/CHROMIUM_color_buffer_float_rgba.txt is available on my Windows laptop. @kvark So it looks like we can render to F32 on ANGLE...

kvark:
Here is some good news, bad news, and a somewhat ugly conclusion.

### Good
I managed to fix the correctness/Angle issues, making it run now on all platforms. 🎉 
Angle is quite picky about the shader code, it decided to ignore my scatter draw call because the VS didn't assign anything to `gl_PointSize`...

### Bad
Benchmarking with MotionMark on a Windows/Angle machine didn't show any speedup. In fact, the scores were even ~10 lower, within the margin of measurement error. On the other hand, we don't appear to be nearly bottlenecked by the GPU cache updates. Compositor time I've seen is either mostly waiting, or doing texture uploads (which have been optimized to be Angle-friendly).

Perhaps, we can come up with a stronger test case that stresses the hell out of GPU cache, e.g. by using a thousand of borders?..

### Ugly
The PR brings a bit of a nicer code overall, and it would be nice to reserve this option (to minimize the GPU cache transfers) for the future. So we have a tentative agreement with @glennw to disable it by default and merge.

glennw:
Reviewed 2 of 4 files at r1, 4 of 5 files at r3, 1 of 1 files at r4.

---

*[webrender/res/gpu_cache_update.glsl, line 14 at r4](https://reviewable.io:443/reviews/servo/webrender/2162#-L-rtNhq7UVDTgpyBiw3:-L-rtNhq7UVDTgpyBiw4:b-ggniwg) ([raw file](https://github.com/servo/webrender/blob/8eeeae048910b81e8b60a17c45f519338f1e1a86/webrender/res/gpu_cache_update.glsl#L14)):*
> ```GLSL
> 
> void main() {
> 	vData = aValue;
> ```

nit: tabs here instead of spaces

---

*[webrender/src/device.rs, line 411 at r4](https://reviewable.io:443/reviews/servo/webrender/2162#-L-rtYZB6ds2VLne_2De:-L-rtYZCLzcSPiU_SDHh:b6zzob8) ([raw file](https://github.com/servo/webrender/blob/8eeeae048910b81e8b60a17c45f519338f1e1a86/webrender/src/device.rs#L411)):*
> ```Rust
>     }
> 
>     pub fn streaw_with<'a>(&self, attributes: &'a [VertexAttribute]) -> Stream<'a> {
> ```

nit: spelling

---

*[webrender/src/renderer.rs, line 74 at r4](https://reviewable.io:443/reviews/servo/webrender/2162#-L-ruE-FP0Ce8T3oeV9w:-L-ruE-FP0Ce8T3oeV9x:b-f29qha) ([raw file](https://github.com/servo/webrender/blob/8eeeae048910b81e8b60a17c45f519338f1e1a86/webrender/src/renderer.rs#L74)):*
> ```Rust
> 
> pub const MAX_VERTEX_TEXTURE_WIDTH: usize = 1024;
> const GPU_CACHE_RESIZE_TEST: bool = false;
> ```

Let's add a comment here describing what this is for, if we still need it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2162)*

glennw:
@kvark Looks good! Added a few very minor nits, r=me once those are resolved. I leave it to your judgment if we need a try run here :)

kvark:
Thanks @glennw !
Notes are addressed. The try run has 1 unexpected pass and 1 unexpected failure (in same-filter): https://treeherder.mozilla.org/#/jobs?repo=try&revision=f583b2deae1c7312437638bdda6ddc65693d53ed&selectedJob=150711942
Not sure if those are expected (will check later, unless you know out of hand).

Also, I believe the scatter-enabled Gecko is a little bit snappier at start, for it uses blitting for GPU texture cache resizing, which helps to grow the height from 512 to 2048 without hitches.

glennw:
@kvark Yes, those are expected differences with current WR master.

staktrace:
This PR caused panics on windows reftest runs: https://bugzilla.mozilla.org/show_bug.cgi?id=1424280#c8

staktrace:
I filed #2208 for this regression with a more useful backtrace.

PR #2164, opened 2017-12-05
fix handling of embedded bitmaps in FreeType font backend

lsalzman:
This fixes a small regression I accidentally made with #2120. The prior "(flags & FONT_EMBEDDED_BITMAP) == 0" statement needed to be !contains() rather than just contains().

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2164)


jrmuizel:
Can we get a test?

lsalzman:
@jrmuizel We would need at least a appropriately licensed bitmap font we could include in the tree to test that sort of thing.

PR #2165, opened 2017-12-05
don't redundantly offset bitmap glyph origins in FreeType font backend

lsalzman:
This fixes downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1420956

When outline glyphs are rasterized, FT_Outline_Translate is called on them, which is why we need to offset the glyph origin to account for any leftover translation in the final rasterized bitmap.

However, for glyphs that are actually already bitmaps and are thus not rasterized/translated, we want to pass through the origin untouched. When we erroneously offset them again, this caused problems with glyphs not sitting directly on the baseline.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2165)

### DIFF
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -493,9 +493,10 @@ impl FontContext {
             Some(val) => val,
             None => return None,
         };
+        let GlyphDimensions { mut left, mut top, width, height, .. } = dimensions;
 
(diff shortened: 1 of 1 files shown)


glennw:
@lsalzman has identified a font that seems suitable to add in-tree for reftests. We'll land this now and add the tests as a follow up.

@bors-servo r+

PR #2170, opened 2017-12-06
Clarify a comment in a test

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2170)


mrobinson:
This change is very minor, but the original comment had me confused for a time, so I figured it was worth clarifying things here.

PR #2171, opened 2017-12-06
Don't calculate inner rectangles for non-axis-aligned clips

### LINKED ISSUE #2167: regression: Children of transformed parent elements with overflow:hidden are not displayed correctly
GH issue for https://bugzilla.mozilla.org/show_bug.cgi?id=1423104 - see the testcase on that bug.

This is a regression from #2104.

/cc @mrobinson

mrobinson:
These inner rectangles are going to potentially include regions that
should be clipped away. This may introduce a performance regression
that should be fixed soon when we have better inner rect optimization
for clips that are in the same coordinate system as the primitive that
they clip.

Fixes #2167.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2171)


mrobinson:
cc @staktrace

mrobinson:
Gecko try run here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=dfb5231b718df0d97d38d07b7375f5aeb75a257a

staktrace:
Thanks for the quick fix! I ran a build from your try push and verified it fixes both of the bugs linked #2167.

kvark:
@mrobinson what exactly is wrong with the logic of `calculate_screen_bounding_rect` at the moment?

kvark:
Ok, the logic of using `calculate_screen_bounding_rect` is wrong, supposedly as of #2156.
Lemme produce an alternative fix for this issue.

mrobinson:
@kvark If the old code was calculating an interior rectangle of an on screen polygon, it would just require restoring the old code.

mrobinson:
In this case, probably just for the interior rectangle though.

PR #2172, opened 2017-12-06
Prevent conflicting coordinate system ids

mrobinson:
Since the next coordinate system id was stored on the state, this value
was not being propagated back up to siblings. The TransformUpdateState
is meant to only be updated downward. Pull out the next
CoordinateSystemId as another mutable argument in order to have it be
persistent through the entire tree traversal. There is no test for
this, but the previous design was obviously wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2172)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -329,6 +329,7 @@ impl ClipScrollNode {
     pub fn update(
         &mut self,
         state: &mut TransformUpdateState,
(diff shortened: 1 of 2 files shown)


mrobinson:
Gecko try run here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=3fe9c262e90aa2554f340abe4196516b036d577b

PR #2177, opened 2017-12-06
Fixed inner rectangle computation for transformed clips

### LINKED ISSUE #2167: regression: Children of transformed parent elements with overflow:hidden are not displayed correctly

kvark:
Alternative to #2171, which it steals the test from (thanks @mrobinson !)

Fixes #2167

Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=59424192544b000e9363f382056c68ccc7209dae

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2177)


mrobinson:
The change looks good to me. Thanks for restoring the conservative inner rect calculation. I guess someone else will need to review my test.

mrobinson:
Instead of mentioning my Github name in the test commit message, it would probably make for a cleaner git history to use the `--author=` argument.

kvark:
Thanks @mrobinson , great idea! Fixed now.

glennw:
It would be good to crop the reference image, but that can be done as a follow up.

@bors-servo r+

mrobinson:
It seems that bors has been stuck today, so I've taken the liberty of opening a new PR with the test image cropped.

PR #2184, opened 2017-12-06
Reduce HashMap access in path_for_image

salty-horse:
The old code removed an item from the hashmap, manipulated it, and stored it as a new value.

This patch modifies it in-place.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2184)

### DIFF
--- a/wrench/src/yaml_frame_writer.rs
+++ b/wrench/src/yaml_frame_writer.rs
@@ -539,16 +539,14 @@ impl YamlFrameWriter {
 
 
     fn path_for_image(&mut self, key: ImageKey) -> Option<PathBuf> {
-        if let Some(ref mut data) = self.images.get_mut(&key) {
-            if data.path.is_some() {
-                return data.path.clone();
-            }
-        } else {
-            return None;
+        let data = match self.images.get_mut(&key) {
+            Some(data) => data,
+            None => return None,
         };
 
-        // Remove the data to munge it
-        let mut data = self.images.remove(&key).unwrap();
+        if data.path.is_some() {
+            return data.path.clone();
+        }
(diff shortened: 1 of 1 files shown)


wrench/src/yaml_frame_writer.rs
```diff
@@ -492,16 +492,15 @@ impl YamlFrameWriter {
     }
 
     fn path_for_image(&mut self, key: ImageKey) -> Option<PathBuf> {
-        if let Some(ref mut data) = self.images.get_mut(&key) {
-            if data.path.is_some() {
-                return data.path.clone();
-            }
-        } else {
+        let data = self.images.get_mut(&key);
+        if data.is_none() {
```
glennw: How about something like:

```
match self.images.get_mut(&key) {
    Some(data) => {
       ...
    }
    None => {
      return;
    }
```

salty-horse: I wanted to avoid indenting all the code after the changed lines, and still leave the error handling at the top.

I don't object to changing it if that's what you prefer.

glennw: Yep, let's change it to that style and then this should be good to merge. Thanks!

kvark: Actually, having a `match` there doesn't force indentation:
```rust
let data = match self.images.get_mut(&key) {
  Some(data) => data,
  None => return;
};
```

salty-horse: @kvark Ooh, nice. Will do that.

salty-horse:
Force-pushed a new change based on review comments.

PR #2185, opened 2017-12-06
Avoid creating duplicate copies of image files

salty-horse:
When replaying a recording, and moving back and forth over a frame that introduces a new image, the image was saved repeatedly, each time with a new filename.

This adds a check to `AddImage` in case we remember there's already a filename for that image.

I haven't tested how this interacts with `UpdateImage` and `DeleteImage`, but both commands either delete the path, or remove the image entry altogether, bypassing this check.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2185)


glennw:
Seems reasonable, thanks!

@bors-servo r+

glennw:
Merging manually due to travis / bors issues.

PR #2186, opened 2017-12-07
Add a compact profiler mode.

glennw:
This allows viewing a minimal amount of the most critical data from
the profiler, as shown in the screenshot below.

This is useful for a couple of reasons, namely:
 * Makes it easier to interact with the underlying page.
 * Reduces the amount of GPU time spent drawing the profiler itself.

The profiler code in general is a bit messy - this includes a few
hacks to work around the borrow checker. We should restructure this
a bit in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2186)

### DIFF
--- a/webrender/src/profiler.rs
+++ b/webrender/src/profiler.rs
@@ -37,15 +37,6 @@ trait ProfileCounter {
     fn value(&self) -> String;
 }
 
-impl<'a, T: ProfileCounter> ProfileCounter for &'a T {
-    fn description(&self) -> &'static str {
-        (*self).description()
-    }
-    fn value(&self) -> String {
-        (*self).value()
-    }
-}
-
 #[derive(Clone)]
 pub struct IntProfileCounter {
     description: &'static str,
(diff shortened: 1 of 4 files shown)


glennw:
compact

glennw:
r? @kvark or @nical 

PS: Drawing the profiler view is really expensive, ~2ms on my HD4600 (!).

Showing the compact view takes the text rendering benchmark from 335 FPS up to 440 FPS on my machine ...

webrender/src/profiler.rs
```diff
@@ -947,126 +960,162 @@ impl Profiler {
...
+        self.ipc_time
+            .push(backend_profile.ipc.total_time.nanoseconds);
+        self.gpu_time.push(gpu_time);
+        self.gpu_frames.push(gpu_time, gpu_timers);
 
-        self.draw_gpu_cache_bars(
-            &backend_profile.resources.gpu_cache,
-            debug_renderer,
-        );
+        if compact {
```
kvark: those branches make we want a separate function for the long/short statistics

glennw: Done

kvark:
thanks!
@bors-servo r+

glennw:
Merging manually due to travis / bors issues.

PR #2188, opened 2017-12-07
fix local clipping of transformed glyphs

lsalzman:
This fixes downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1423304

It is inappropriate to transform the device space rect corners back into local space and then clip them against the local clip rects. First we need to transform the entire device rect to a local rect, then select the corner from there, which can then be properly clipped by the local clip rects.

The only problem with that is that the transformed rect may now sample outside the device rect. To resolve this, I replaced the old clamping of texture coordinates (which would produce an undesirable smearing) with clipping, so that we can limit the texture sampling to be within the device rect regardless.

Since this may cause some overdraw sending fragments that are ultimately going to be clipped, this also adds an optimization to check if the inverse transformed local rect will fit within the local clip rects. If it does, we can safely assume that sampling using the device rect corners without any clamping is safe, which avoids the enlarged geometry and overdraw.

Also added a reduced reftest to ensure we don't regress this again.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2188)


glennw:
Can you do a gecko try with this change?

lsalzman:
Gecko try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=0b0e42c0a34870f1d6ca7e47b9c37b5c07984ecc

jrmuizel:
Lee wanted this in sooner rather than latter. @glennw any review comments that you have we can do as a follow up.

@bors-servo r+

glennw:
Manually merging due to bors / travis issues.

PR #2189, opened 2017-12-07
Fixed inner rectangle computation for transformed clips

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2189)


mrobinson:
This is #2177, but with the test reference image properly cropped as requested by @glennw.

staktrace:
@mrobinson This PR caused a talos-g1 failure. Panic stack and details in https://bugzilla.mozilla.org/show_bug.cgi?id=1423203#c10 onwards.

mrobinson:
@kvark Any idea about this one? I'll try to investigate a bit tomorrow. If it seems like the issue is quite tricky we can use my original (super conservative and inefficient) approach until we figure this one out, in order to not block the WR update in Gecko.

kvark:
@mrobinson yeah, I've been meaning to look into it but my Talos doesn't work and I got sucked into other things. Going with your original PR sounds fine to me, for now.

kvark:
Actually, upon another consideration that code is incorrect (non-concervative). We should definitely back it out.

mrobinson:
Okay. I have a workaround PR here #2205.

PR #2196, opened 2017-12-08
Move the python scripts into script folder

kvark:
Officially this is needed for consistency.
In reality though, it's just annoying when bash auto-completion for `cargo run -- show ref...` in `wrench` folder gives you `reftest` instead of `reftests` :D

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2196)


Gankro:
muh muscle memory

PR #2198, opened 2017-12-08
determine whether a font has bitmaps only when fetching glyphs

lsalzman:
This fixes downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1422408

Some TTFs can contain both fixed size bitmap strikes and scalable outlines. The proper selection procedure is to check if there are strikes *exactly* matching the desired size, otherwise use the scalable outlines.

But we violate this currently by always trying to use the bitmap strikes if they are available even when scalable outlines are present. The problem is that we don't know the actual resolved size at the time we decide to mark a FontInstance with FontRenderMode::Bitmap when a FontInstance is newly created, so we never used that to make the decision.

To fix this, we need to delay deciding whether something is a bitmap until we actually go to fetch the glyph from the cache and essentially when we go to rasterize it. We let the font backend communicate that it is actually a bitmap via the new GlyphFormat::Bitmap (and the existing GlyphFormat::ColorBitmap for emoji). 

We then need to make sure this gets down to the ps_text_run shader via TextShaderMode, so we can selectively disable subpixel offsets based on whether something is a bitmap. We can't disable subpixel offsetting earlier since we don't know if the font is a bitmap yet when gpu data is written out, and even then, this will depend on size, so a global decision can't be made for all sizes instances of a given TextRunPrimitiveCpu.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2198)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -995,7 +995,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "webrender"
-version = "0.56.0"
+version = "0.56.1"
 dependencies = [
  "angle 0.5.0 (git+https://github.com/servo/angle?branch=servo)",
  "app_units 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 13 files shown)


lsalzman:
Gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=badf6c95b2088acae29593b075467ad3865e0732

glennw:

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2198)*

PR #2203, opened 2017-12-08
Stop the warning spam about invalid prim rects.

nical:
The first commit makes it so we don't warn in prim_store.rs if the local_rect of a picture primitive is empty. This is because picture primitves are created with an empty rect that is computed later.  It's not entirely clear to me whether we should still early return from `prepare_prim_for_render` in that case so I kept the current behavior.

The second commit filters out empty rects during display list construction(except for stacking contexts). Turns out gecko creates some of those. This is the simplest place to work around it, but if you prefer I can handle this on the gecko side as well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2203)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1754,7 +1754,9 @@ impl PrimitiveStore {
             let metadata = &mut self.cpu_metadata[prim_index.0];
             if metadata.local_rect.size.width <= 0.0 ||
                metadata.local_rect.size.height <= 0.0 {
-                warn!("invalid primitive rect {:?}", metadata.local_rect);
+                if metadata.prim_kind != PrimitiveKind::Picture {
+                    warn!("invalid primitive rect {:?}", metadata.local_rect);
+                }
                 return None;
             }
 
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -771,6 +771,9 @@ impl DisplayListBuilder {
     }
 
     fn push_item(&mut self, item: SpecificDisplayItem, info: &LayoutPrimitiveInfo) {
+        if info.rect.is_empty() {
+            return;
+        }
         serialize_fast(
             &mut self.data,
             &DisplayItem {
@@ -1185,7 +1188,14 @@ impl DisplayListBuilder {
             },
         });
 
-        self.push_item(item, info);
+        serialize_fast(
+            &mut self.data,
+            &DisplayItem {
+                item,
+                clip_and_scroll: *self.clip_stack.last().unwrap(),
+                info: *info,
+            },
+        );
         self.push_iter(&filters);
     }
 


webrender/src/prim_store.rs
```diff
@@ -1754,7 +1754,9 @@ impl PrimitiveStore {
             let metadata = &mut self.cpu_metadata[prim_index.0];
             if metadata.local_rect.size.width <= 0.0 ||
                metadata.local_rect.size.height <= 0.0 {
-                warn!("invalid primitive rect {:?}", metadata.local_rect);
+                if metadata.prim_kind != PrimitiveKind::Picture {
```
glennw: The local rect for a picture is updated in the block just above this, isn't it? Does it happen on all pictures?

webrender_api/src/display_list.rs
```diff
@@ -771,6 +771,9 @@ impl DisplayListBuilder {
     }
 
     fn push_item(&mut self, item: SpecificDisplayItem, info: &LayoutPrimitiveInfo) {
+        if info.rect.is_empty() {
```
glennw: There's a couple of non-visible items that use push_item - e.g. sticky frame and push_shadow. Are those both fine to skip if the rect is empty?

staktrace: I think all the stacking contexts that gecko pushes also have empty rects, so this is probably not safe

nical: stacking context are not affected by this commit but if sticky frames and shadows need to work with empty rects then my naïve stab at this won't do. I can also filter empty rects out on a per item basis (in which case doing this on the wr or gecko side is the same amount of effort so don't hesitate to let me know if you feel like it should be the caller's responsibility.

webrender/src/prim_store.rs
nical: right now I can only say that it happens on a lot of pictures even with the other commit that removes empty rect at during dl building.

glennw: Interesting! I *think* that is unexpected. I wonder if we end up getting a lot of stacking contexts pushed that are empty, resulting in zero-sized pictures or something like that. This is probably worth investigating a bit - it's possible that it's a sign of a deeper issue or something we can catch earlier for a slight performance win. Or it might just be a code that I'm not thinking of right now.

webrender_api/src/display_list.rs
glennw: I *think* that it might be an issue for shadows - like pictures, we calculate the required rect based on the contents placed in the shadow. I'm not sure about sticky frames - @mrobinson may know. I do think it's a good idea to handle this in WR, since it can actually cause visual bugs if we try to process an item with size < 0.

staktrace:
Any update here? This warning is getting quite annoying and if we can't figure out the right tweaks to the condition maybe we should just comment it out for now.

https://bugzilla.mozilla.org/show_bug.cgi?id=1420138
https://bugzilla.mozilla.org/show_bug.cgi?id=1428439

nical:
We turned the warning off in #2267 and filed #2268 to come back and investigate whenever someone has the time.

PR #2204, opened 2017-12-08
Make sure we mulitply by byte_per_pixel when computing byte offsets.

jrmuizel:
This fixes us getting confused by dirty rects when doing image updates.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2204)

### DIFF
--- a/wrench/src/rawtest.rs
+++ b/wrench/src/rawtest.rs
@@ -220,15 +220,16 @@ impl<'a> RawtestHarness<'a> {
             resources,
         );
         self.wrench.api.generate_frame(document_id, None);
+        let pixels_first = self.render_and_get_pixels(window_rect);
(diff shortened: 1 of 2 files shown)


wrench/src/rawtest.rs
```diff
@@ -254,11 +255,46 @@ impl<'a> RawtestHarness<'a> {
...
+            builder.finalize(),
+            false,
+            resources,
+        );
+
+        self.wrench.api.generate_frame(document_id, None);
+
+        let pixels_third = self.render_and_get_pixels(window_rect);
+
+        assert!(pixels_first == pixels_second);
```
kvark: nit: there are `assert_eq!` and `assert_ne!`

PR #2205, opened 2017-12-08
Use a more conservative approach for inner rect calculation

mrobinson:
This is a temporary workaround for a Gecko issue with the conservative
inner rect calculation for clips. In this case we just don't calculate
an inner rect when faced with a non-axis-aligned transformation.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2205)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -10,7 +10,7 @@ use freelist::{FreeList, FreeListHandle, WeakFreeListHandle};
 use gpu_cache::{GpuCache, GpuCacheHandle, ToGpuBlocks};
 use prim_store::{ClipData, ImageMaskData};
 use resource_cache::ResourceCache;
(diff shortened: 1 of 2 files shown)


mrobinson:
Here's the Gecko try run for this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=b370376404d4d6c001a5f92c9a227b8303ac213b

Once we confirm that the Talos test is passing, this should be okay to merge.

mrobinson:
@kvark Unfortunately, it is not exactly the same, because Gecko is still using an older version of Euclid. This required me to use the original version of `calculate_screen_bounding_rect` for the try push. I can update this PR to include that as well.

kvark:
@mrobinson nah, it's fine. We can wait for your new try push's Talos result.

mrobinson:
@kvark Okay. I'm fairly certain the changes you made to `calculate_screen_bounding_rect` did not change behavior.

PR #2207, opened 2017-12-09
pass subpixel dir to shader via glyph instances instead of text run

lsalzman:
This is a follow-up to #2198. I overlooked cs_text_run, which will also need to know about subpx_dir. Since we don't break up batches for these text shadow glyphs, and since subpixel direction now depends on the glyph format which is size/transform dependent, it just made a lot more sense to pass it in via the (now) unused user_data2 in the glyph instance. This also moves some decision logic out of the shader. Since all these parameters get routed out of prim.user_dataN almost straight to fetch_glyph, it makes some sense to put subpx_dir there anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2207)

### DIFF
--- a/webrender/res/cs_text_run.glsl
+++ b/webrender/res/cs_text_run.glsl
@@ -18,10 +18,11 @@ void main(void) {
 
     int glyph_index = prim.user_data0;
     int resource_address = prim.user_data1;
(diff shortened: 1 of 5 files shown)


glennw:
@bors-servo retry
 * travis timeout

PR #2209, opened 2017-12-09
Ignore DeleteDocument during playback of last frame to avoid crashing

salty-horse:
This fixes the crash during "replay" when reaching the last frame.
The render backend gets sent a `DeleteDocument`, so it delete its document, and then wrench asks it to `UpdateDocument`.

If we do let it accept the `DeleteDocument`, and have wrench detect it and do no more updates, then we won't be able to rewind. Ignoring it is the simplest option.

Does this break anything? Can there be recordings with multiple documents that are added and deleted as it goes? From what I've seen, wrench adds its own document when it starts.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2209)

### DIFF
--- a/wrench/src/binary_frame_reader.rs
+++ b/wrench/src/binary_frame_reader.rs
@@ -133,6 +133,7 @@ impl WrenchThing for BinaryFrameReader {
                     let mut buffer = vec![0; len as usize];
                     self.file.read_exact(&mut buffer).unwrap();
                     let msg = deserialize(&buffer).unwrap();
+                    let mut store_message = true;
                     // In order to detect the first valid frame, we
                     // need to find:
                     // (a) SetRootPipeline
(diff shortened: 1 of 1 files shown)


wrench/src/binary_frame_reader.rs
```diff
@@ -152,7 +152,10 @@ impl WrenchThing for BinaryFrameReader {
                         }
                         _ => {}
                     }
-                    self.frame_data.push(Item::Message(msg));
```
glennw: Let's incorporate this logic into the match statement above, to make it a bit tidier. We should also add a brief comment describing why we ignore DeleteDocument messages during replay.

salty-horse:
Force-pushed new patch. I hope I understood your suggestion.

salty-horse:
I should mention that I'm always ignoring DeleteDocument, not just during playback as I mentioned in the title.
Is this not the desired behavior? Is there a practical reason to not ignore it?

glennw:
Looks good to me, thanks!

@bors-servo r+

Issue #2208, opened 2017-12-09
Panic when starting up gecko+webrender on Windows


staktrace:
This is the panic from https://bugzilla.mozilla.org/show_bug.cgi?id=1424280#c8, which is a regression from #2162. I was able to reproduce it on a local windows debug build and got a better stack:

```
thread '<unnamed>' panicked at 'index 5121 out of range for slice of length 5120', src\libcore\slice\mod.rs:746:4
stack backtrace:
   0:     0x7ffd6c9992e1 - std::sys_common::backtrace::_print
                               at C:\projects\rust\src\libstd\sys_common\backtrace.rs:92
   1:     0x7ffd6c9b0d4c - std::panicking::default_hook::{{closure}}
                               at C:\projects\rust\src\libstd\panicking.rs:380
   2:     0x7ffd6c9b08f5 - std::panicking::default_hook
                               at C:\projects\rust\src\libstd\panicking.rs:397
   3:     0x7ffd6c9b1545 - std::panicking::rust_panic_with_hook
                               at C:\projects\rust\src\libstd\panicking.rs:577
   4:     0x7ffd6c9b13b8 - std::panicking::begin_panic<alloc::string::String>
                               at C:\projects\rust\src\libstd\panicking.rs:538
   5:     0x7ffd6c9b1241 - std::panicking::begin_panic_fmt
                               at C:\projects\rust\src\libstd\panicking.rs:522
   6:     0x7ffd6c9b119f - std::panicking::rust_begin_panic
                               at C:\projects\rust\src\libstd\panicking.rs:498
   7:     0x7ffd6d0ace5e - core::panicking::panic_fmt
                               at C:\projects\rust\src\libcore\panicking.rs:71
   8:     0x7ffd6d0acf71 - core::slice::slice_index_len_fail
                               at C:\projects\rust\src\libcore\slice\mod.rs:746
   9:     0x7ffd6d4b514e - core::slice::{{impl}}::index_mut<webrender::gpu_cache::GpuBlockData>
                               at C:\projects\rust\src\libcore\slice\mod.rs:889
  10:     0x7ffd6d493f73 - core::slice::{{impl}}::index_mut<webrender::gpu_cache::GpuBlockData,core::ops::range::Range<usize>>
                               at C:\projects\rust\src\libcore\slice\mod.rs:739
  11:     0x7ffd6d45aff5 - alloc::vec::{{impl}}::index_mut<webrender::gpu_cache::GpuBlockData>
[...]

PR #2212, opened 2017-12-10
Define and enforce BLOCKS_PER_UV_RECT

kvark:
A stronger version of the fix for #2208
Replaces #2211

TODO: try push

Note: I tried going all the way to type-level enforcement of the GPU block count for everything but faced a few run-time driven GPU layouts (clip sources, borders). Perhaps, next time...

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2212)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -720,6 +720,7 @@ struct ImageResource {
 };
 
 ImageResource fetch_image_resource(int address) {
+    //Note: number of blocks has to match `renderer::BLOCKS_PER_UV_RECT`
     vec4 data[2] = fetch_from_resource_cache_2(address);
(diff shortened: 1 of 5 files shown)


kvark:
Funny thing... in an attempt to get a try push with this change I figured I can just fetch the code from failed try pushes of @staktrace , fix it, and push again. What a mistake! Git-cinnabar-powered fetch went OOM on my 8gb RAM laptop after an hour of operation, with all apps except that console window closed...

glennw:
Looks good, r=me. I'll leave it to you if you want to do a try push, but it doesn't seem necessary for this patch, I think.

kvark:
Now that #2211 is merged this PR doesn't change any logic, hence safe to merge.
@bors-servo  r=glennw

PR #2213, opened 2017-12-10
fix glyph clipping when scale is used

lsalzman:
This is to fix downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1424562

It turns out that we're not going to be able to live within the 0.5 to N-0.5 bounds when sampling the texture if there is scaling or some aberrant snapping in play. Imagine a 2 pixel wide bitmap being stretched over 3 device space pixels, the sample taps on the ends will unfortunately be outside these bounds.

We still need to clip, but we need to use a more permissive clip that allows for this sampling to happen anywhere in the 0..N range. To prevent texels bleeding in from outside the glyph, unfortunately we need to restore the original clamping behavior as well. I also ifdef'd the clipping to avoid it costing us in those non-transform cases where we don't actually need it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2213)

### DIFF
--- a/webrender/res/ps_text_run.glsl
+++ b/webrender/res/ps_text_run.glsl
@@ -8,6 +8,10 @@ flat varying vec4 vColor;
 varying vec3 vUv;
 flat varying vec4 vUvBorder;
 
+#ifdef WR_FEATURE_GLYPH_TRANSFORM
+varying vec4 vUvClip;
+#endif
+
 #ifdef WR_VERTEX_SHADER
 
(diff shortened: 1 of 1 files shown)


lsalzman:
Gecko try https://treeherder.mozilla.org/#/jobs?repo=try&revision=4d63e10b6533f2e1c8a8a8c374d995cd52eff5ab

glennw:
Any way we can add a reftest to prevent this regressing?

PR #2214, opened 2017-12-10
Lock serde to 1.0.23 exactly so gankro's replace works even when a ne…

staktrace:
…wer caret-compatible release is available on crates.io

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2214)


staktrace:
If we don't do this running a `cargo update` on webrender will sort-of kind-of try to update serde to 1.0.24 and the [replace] fails and cargo produces an error. This will also prevent gecko from updating past 1.0.23 and getting the same error.

PR #2215, opened 2017-12-10
Use [patch] instead of [replace]

staktrace:
This is the new hotness and is what we need to use in Gecko, so for
consistency let's use it in webrender as well. This requires rust
1.22, so do a version bump to go with. Gecko has already had its minimum
rust version requirement bumped to 1.22.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2215)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: trusty
 language: rust
 rust:
-  - 1.21.0
+  - 1.22.1
   - nightly
 matrix:
   fast_finish: true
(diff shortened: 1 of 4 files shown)


glennw:
CI error on appveyor:

```
error: Package `serde_derive v1.0.23` does not have these features: `deserialize_from`
Command exited with code 101
```

staktrace:
Thanks for the heads up. The error is probably because appveyor.yml is using rust 1.20; I pushed an updated commit that bumps it to 1.22.1. Let's see if that does better.

PR #2217, opened 2017-12-12
Fixed transformed primitives

### LINKED ISSUE #2168: regression: Rotated rect not respecting device pixel ratio
This is a regression from #2069 
```
8253bbb398d03094c704be9333415789cabf6b7c is the first bad commit
commit 8253bbb398d03094c704be9333415789cabf6b7c
Author: Glenn Watson <github@intuitionlibrary.com>
Date:   Tue Nov 21 14:36:11 2017 +1000

    Remove manual interpolation of local space in transform shaders.

    Instead, for vertex transform shaders we now write out:
     (world.xy, z * world.w, world.w)

    This gives us constant Z across the polygon for draw ordering.

    It also means we can use the hardware interpolators instead of
    manually dividing by Z in the fragment shader for local_pos.

    This is mostly preparation work for switching shaders over to
    be brush primitives, but has the happy side effect of:

     * Fixing a small number of Gecko + Servo reftests.
     * Simpler and faster code.
```

### STR:
I first noticed this in the `animations` sample, you can hit `2` to set the device pixel ratio to 2.0 and then hit `,` or `.` to rotate the rect.

You can also run `rotated-clip.yaml` when run with `-p=2` and observe it not respecting the pixel ratio

I didn't see a way to specify a device pixel ratio in `reftest.list` but perhaps it would be a nice enhancement?

kvark:
Fixes #2168

Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e446306f491bf773f335469280f0b3db412ffa9f

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2217)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -670,12 +670,13 @@ VertexInfo write_transform_vertex(RectWithSize local_segment_rect,
 
     // Convert the world positions to device pixel space.
     vec2 device_pos = world_pos.xy / world_pos.w * uDevicePixelRatio;
+    vec2 task_offset = task.common_data.task_rect.p0 - task.content_origin;
 
     // We want the world space coords to be perspective divided by W.
     // We also want that to apply to any interpolators. However, we
     // want a constant Z across the primitive, since we're using it
     // for draw ordering - so scale by the W coord to ensure this.
-    vec4 final_pos = vec4(world_pos.xy + task.common_data.task_rect.p0 - task.content_origin,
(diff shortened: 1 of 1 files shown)


glennw:
Oops! r=me if try is green.

jrmuizel:
Test case?

kvark:
All green.
@bors-servo r+

kvark:
OSX:

@bors-servo retry

kvark:
Hmm, have our OSX build times regressed?
@bors-servo retry

PR #2218, opened 2017-12-12
Add a function to compute various type flags of a matrix.

glennw:
This provides a more unified way to calculate various properties
of a matrix, rather than has_perspective_component etc.

The preserves_2d_alignment method returns true for transforms
with a 90 degree rotation. This function allows us to differentiate
between that and a pure translation matrix, which will be useful
for some of the upcoming clip / segment code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2218)

### DIFF
--- a/webrender/src/util.rs
+++ b/webrender/src/util.rs
@@ -12,6 +12,16 @@ use std::{i32, f32};
 // Matches the definition of SK_ScalarNearlyZero in Skia.
 const NEARLY_ZERO: f32 = 1.0 / 4096.0;
 
+bitflags! {
+    #[derive(Default)]
+    pub struct MatrixFlags: u32 {
+        const TRANSLATION   = 1;
+        const SCALE         = 2;
+        const AFFINE        = 4;
+        const PERSPECTIVE   = 8;
+    }
+}
+
 // TODO: Implement these in euclid!
 pub trait MatrixHelpers<Src, Dst> {
     fn preserves_2d_axis_alignment(&self) -> bool;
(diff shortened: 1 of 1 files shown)


webrender/src/util.rs
```diff
@@ -92,6 +103,40 @@ impl<Src, Dst> MatrixHelpers<Src, Dst> for TypedTransform3D<f32, Src, Dst> {
...
+        let mut mask = MatrixFlags::empty();
+
+        if self.m14 != 0.0 ||
+           self.m24 != 0.0 ||
+           self.m34 != 0.0 ||
+           self.m44 != 1.0 {
+            return MatrixFlags::TRANSLATION |
+                   MatrixFlags::SCALE |
+                   MatrixFlags::AFFINE |
+                   MatrixFlags::PERSPECTIVE;
```
kvark: can we really say the matrix has translation/scale if we only check for the perspective components here?

webrender/src/util.rs
```diff
@@ -92,6 +103,40 @@ impl<Src, Dst> MatrixHelpers<Src, Dst> for TypedTransform3D<f32, Src, Dst> {
...
+        if self.m41 != 0.0 ||
+           self.m42 != 0.0 ||
+           self.m43 != 0.0 {
+            mask |= MatrixFlags::TRANSLATION;
+        }
+
+        if self.m11 != 1.0 ||
+           self.m22 != 1.0 ||
+           self.m33 != 1.0 {
+            mask |= MatrixFlags::SCALE;
```
kvark: could also be pure rotation without scale

glennw:
The flags returned are conservative - this is the comment in the skia source:

```
Returns a bit field describing the transformations the matrix may
        perform. The bit field is computed conservatively, so it may include
        false positives. For example, when kPerspective_Mask is set, all
        other bits are set.
```

kvark:
So what you are trying to encode is 
```rust
enum MatrixKind {
  Simple { scale: bool, translation: bool },
  Affine,
  Perspective,
}
```
Would it be more convenient for the user code to work with flags rather than enums?

glennw:
It's not clear to me at this point if it would be easier for user code to use the flags or the enum - I do like the enum approach in general. We could probably either merge this now, or I can close this and incorporate it as part of a larger PR which will make use of it.

kvark:
@glennw sounds good to me. When you have a use case, you'd have a second look at the enum vs flag and make this call.

glennw:
I realized that what I *actually* want in this use case is going to be covered by the ```TransformOrOffset``` in the PR that @mrobinson is working on. So I'll close that and base my code on that PR.

PR #2221, opened 2017-12-13
Fix inverted translation for reference frames between sticky frames a…

staktrace:
…nd their scrolling ancestor

r? @mrobinson

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2221)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -659,9 +659,10 @@ impl ClipScrollNode {
                 state.parent_reference_frame_transform = self.world_viewport_transform;
                 state.parent_combined_viewport_rect = combined_local_viewport_rect;
                 state.parent_accumulated_scroll_offset = LayerVector2D::zero();
+                let translation = -info.origin_in_parent_reference_frame;
                 state.nearest_scrolling_ancestor_viewport =
                     state.nearest_scrolling_ancestor_viewport
-                       .translate(&info.origin_in_parent_reference_frame);
+                       .translate(&translation);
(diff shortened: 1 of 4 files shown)


glennw:
This looks reasonable to me - but I'll wait for @mrobinson to sign off on it. @staktrace Do we need a try run here?

staktrace:
^ compilation error in osmesa-src, not related to my patch as far as I can tell. can somebody kick a retry?

staktrace:
@glennw Also I'm pretty sure this codepath was basically a no-op in gecko (we would always provide an origin of 0). I've done some try pushes that included this patch, on bug 1424686. There were some failures from my gecko-side changes but nothing from this change. Most recent is at https://treeherder.mozilla.org/#/jobs?repo=try&revision=9ee4096e04d6cb2c1b6b7b131eb248df5b37aaf5

glennw:
I assume @staktrace can raise it with @staktrace if this does cause any try-push issues :)

staktrace:
@staktrace should follow the guidelines like everybody else :)

staktrace:
This is going to fail again, for the same reason. I think there's an actual osmesa-src bustage and we need to bump our osmesa-src version to get around it. We don't have https://github.com/servo/osmesa-src/pull/25 in the version we're using. I'll make a new PR for that.

staktrace:
This should be good to retry now, since #2226 fixed the osmesa build.

glennw:
@bors-servo retry
 * mac builder time seems to have really regressed...

Issue #1959, opened 2017-10-28
fast-shadows of apple emoji aren't colored


Gankro:
Now that we correctly prevent emoji from being effected by colors, the colors we try to apply to them as fast shadows get ignored.

test case:

```
<span style="text-shadow: 2px 2px 0px red">
    Shadow
    <span style="color: blue">🎉</span>
    <span style="color: rgba(0,0,0,0.7)">🎉</span>
    <span style="color: rgba(0,0,255,0.4)">🎉</span>
    <span style="color: transparent">🎉</span>
</span>
```

screen shot (with blurred version for reference)

<img width="266" alt="screen shot 2017-10-27 at 11 29 31 pm" src="https://user-images.githubusercontent.com/1136864/32130885-c0dc7a2a-bb6e-11e7-9181-733f25d29f79.png">

Gankro:
live version is near the top of https://gankro.github.io/blah/webtests/text.html

lsalzman:
Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1425157

glennw:
Fixed in #2223.

Issue #1917, opened 2017-10-23
Incorrect text shadows with 0 blur on color emoji


lsalzman:
In the text shadow with 0 blur case, we use the normal text shader. For color emoji, this shader samples the bitmap color data, and then multiplies by the shadow color, when it should just be simply the shadow color. We probably need to punt back to the text shadow shader somehow or modify the text shader somehow to handle this case.

kvark:
I'm not sure I understand why the shadow application for blur=0 should be different (replace vs multiply) from the more general shadow case.
Also, this needs a corresponding Bugzilla issue.

mstange:
It shouldn't be different. The bug is that it is different at the moment. The blur shader ignores R/G/B, the text shader respects it (which is incorrect for shadows).

mstange:
I suggest closing this issue in favor of #1959 which has a screenshot and is easier to understand.

glennw:
Fixed in #2223.

PR #2223, opened 2017-12-13
ignore glyph color data when doing fast shadows

lsalzman:
This addresses issues #1959 and #1917. We need to swizzle away the glyph's color data for emoji fast shadows. Rather than introducing more shaders/ifdefs to deal with this, I added a reasonably cheap programmable swizzle in the shader to control this. A side benefit of this is that we also get rid of the need for the separate shader with the subpixel-with-bg-color blending passes as they now just use the swizzle mechanism.

In the process of this I noticed that the idea of directly stuffing the shadow color into the font on the cloned text run primitive for the fast shadow has a downside: it will usually cause the font to hash differently and so make duplicate entries in the glyph cache. So I added a shadow_color field back onto TextRunPrimitiveCpu, serving dual purposes that 1) lets the font itself hash the same while also 2) indicating down to the batching code that we have to use shadows in the first place (and thus need to enable the swizzle).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2223)


glennw:
This looks good in general, but there are a couple of test failures on CI. I haven't looked at them - it may just need a reference image update. Also, can we do a gecko try once those are resolved and add a reftest?

lsalzman:
Can we use the Apple Color Emoji font from wrench without having to include it in tree? Does it support using font handles, essentially? Would be needed for a simple testcase, unfortunately.

jrmuizel:
Yeah, I believe we should be able to use it without including it. We
support getting a font by name on OS X.

lsalzman:
Fixed reftest failures

webrender/res/ps_text_run.glsl
```diff
@@ -160,16 +163,13 @@ void main(void) {
 void main(void) {
     vec3 tc = vec3(clamp(vUv.xy, vUvBorder.xy, vUvBorder.zw), vUv.z);
     vec4 mask = texture(sColor0, tc);
+    mask.rgb = mask.rgb * vMaskSwizzle.x + mask.a * vMaskSwizzle.y;
```
kvark: I think `mix(mask.rgb, mask.aaa, vMaskSwizzle)` would look cleaner, and also reflect the fact you only need one value in the `vMaskSwizzle` instead of two.

webrender/src/frame_builder.rs
```diff
@@ -1345,6 +1345,7 @@ impl FrameBuilder {
             glyph_gpu_blocks: Vec::new(),
             glyph_keys: Vec::new(),
             offset: run_offset,
+            shadow_color: ColorU::new(0, 0, 0, 0),
```
kvark: Would be best (but not required) to have the `TextRunPrimitiveCpu` in a separate commit, since it's technically independent of the mask swizzling.

webrender/src/renderer.rs
```diff
@@ -1597,7 +1597,6 @@ pub struct Renderer {
     // output, and the cache_image shader blits the results of
     // a cache shader (e.g. blur) to the screen.
     ps_text_run: TextShader,
-    ps_text_run_subpx_bg_pass1: TextShader,
```
kvark: nice! 👍

webrender/src/tiling.rs
```diff
@@ -581,6 +581,11 @@ fn add_to_batch(
                         ],
                     };
 
+                    // Ignore color and only sample alpha when shadowing.
+                    if text_cpu.shadow_color.a != 0 {
```
kvark: a helper method would be cleaner: `if text_cpu.is_shadow()`

webrender/src/tiling.rs
```diff
@@ -595,10 +600,7 @@ fn add_to_batch(
                                 BlendMode::SubpixelConstantTextColor(text_cpu.font.color.into())
                             }
                         }
-                        GlyphFormat::Alpha |
-                        GlyphFormat::TransformedAlpha |
-                        GlyphFormat::Bitmap |
-                        GlyphFormat::ColorBitmap => BlendMode::PremultipliedAlpha,
+                        _ => BlendMode::PremultipliedAlpha,
```
kvark: any particular reason for this change? In general, providing all the enum values is preferred for issuing an error whenever the enum changes.

webrender/res/ps_text_run.glsl
mstange: This is not true for `MODE_SUBPX_BG_PASS1` which uses `vec2(-1.0, 1.0)` as the swizzle.

lsalzman:
Some changes to address review comments

webrender/res/ps_text_run.glsl
kvark: Good point! It's technically not a swizzle then.

lsalzman:
I managed to find a script that let me create color emoji fonts at https://github.com/behdad/color-emoji

I used this to build a simple single glyph color emoji ttf that gets used for a new testcase.

lsalzman:
Ugh, seems like there might be problems with the generated TTFs in certain versions of FreeType. For now I'll just leave the emoji reftest out of this PR, as it might take a while to figure that issue out and resolve it separately.

wrench/reftests/text/reftest.list
```diff
@@ -21,7 +21,7 @@ fuzzy(1,100) == decorations-suite.yaml decorations-suite.png
 fuzzy(1,735) == shadow-grey.yaml shadow-grey-ref.yaml
 fuzzy(1,614) == shadow-grey-transparent.yaml shadow-grey-ref.yaml
 == subtle-shadow.yaml subtle-shadow-ref.yaml
-== shadow-atomic.yaml shadow-atomic-ref.yaml
+options(disable-aa) == shadow-atomic.yaml shadow-atomic-ref.yaml
```
glennw: Why did this need to change? Could it mean that we're accidentally drawing subpixel in the shadow or something like that?

lsalzman: That is one part of it, yes. Then the further part is that due to differing preblending applied depending on text color, it won't necessarily match either even with grayscale. The only mode that can actually match on all backends for a fast shadow of a given color and normal text of that color, is mono, essentially.

glennw: @Gankro do you see any issues with this change?

glennw:
@lsalzman Do we need a try run here?

lsalzman:
Gecko try https://treeherder.mozilla.org/#/jobs?repo=try&revision=40f08a43b13f9badfabaa2ed0f3aa5c542089847

PR #2225, opened 2017-12-14
Make image mask path optional in YAML reader

kvark:
Otherwise we crash during "show" command on some frames.
Appears to be broken as of #1412

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2225)

### DIFF
--- a/wrench/src/yaml_frame_reader.rs
+++ b/wrench/src/yaml_frame_reader.rs
@@ -464,7 +464,18 @@ impl YamlFrameReader {
             return None;
         }
 
-        let file = rsrc_path(&item["image"], &self.aux_dir);
+        let file = match item["image"].as_str() {
+            Some(filename) => {
+                let mut file = self.aux_dir.clone();
+                file.push(filename);
+                file
+            }
+            None => {
+                warn!("No image provided for the image-mask!");
+                return None;
+            }
+        };
+
         let (image_key, image_dims) =
             self.add_or_get_image(&file, None, wrench);
         let image_rect = item["rect"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,6 +1077,7 @@ dependencies = [
  "gleam 0.4.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "image 0.17.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "osmesa-src 17.4.0-devel (git+https://github.com/servo/osmesa-src)",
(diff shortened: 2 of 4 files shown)


glennw:
Is this what was occurring when converting a binary replay to yaml? I think there might be a deeper bug in the binary writer not correctly storing image mask paths. Is it actually useful to support image masks without a path?

kvark:
> Is this what was occurring when converting a binary replay to yaml?

yes

> I think there might be a deeper bug in the binary writer not correctly storing image mask paths.

would be great to look deeper into that for sure

> Is it actually useful to support image masks without a path?

not much: my patch makes the reader ignore such masks

glennw:
I'm not sure this is super useful - but r=me if it helps in the interim until we fix the bug in the yaml writer. We should probably log a warning that an invalid image mask was found and dropped though?

wrench/src/yaml_frame_reader.rs
```diff
@@ -464,7 +464,19 @@ impl YamlFrameReader {
             return None;
         }
 
-        let file = rsrc_path(&item["image"], &self.aux_dir);
+        //rsrc_path(&item["image"], &self.aux_dir);
```
mrobinson: Looks like there is a bit of old code left over here.

wrench/src/yaml_frame_reader.rs
```diff
@@ -464,7 +464,19 @@ impl YamlFrameReader {
...
+        let file = match item["image"].as_str() {
+            Some(filename) => {
+                let mut file = self.aux_dir.clone();
+                file.push(filename);
+                file
+            }
+            None => {
+                //warn!("No image provided for the image-mask!");
```
mrobinson: This warning looks useful, but if you still think it doesn't belong, might as well remove the line.

wrench/src/yaml_frame_reader.rs
kvark: that was intentional, but I don't mind removing this

wrench/src/yaml_frame_reader.rs
kvark: commented out because wrench isn't hooked up to log. I suppose this fixable :)

mrobinson: Hrm. I notice that when I run the ahem.yaml reftest I get this output:
```
REFTEST reftests/text/ahem.yaml == reftests/text/ahem-ref.yaml
Skipping unknown item type: Hash({String("bounds"): Array([Integer(0), Integer(0), Integer(0), Integer(0)]), String("clip-rect"): Array([Integer(0), Integer(0), Integer(0), Integer(0)]), String("clip-and-scroll"): Integer(0), String("backface-visible"): Boolean(true)})
Skipping unknown item type: Hash({String("bounds"): Array([Integer(0),
```

glennw:
@kvark What's the status of this one?

kvark:
@glennw updated and ready for merging now

PR #2226, opened 2017-12-14
Try bumping osmesa-src to a newer rev

staktrace:
This picks up https://github.com/servo/osmesa-src/pull/25 and should fix the travis error that were showing up in #2221

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2226)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,18 +23,19 @@ notifications:
 addons:
   apt:
     sources:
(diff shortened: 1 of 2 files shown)


staktrace:
Shoot, this didn't work. But it's a different error message this time, so maybe progress. I'll look more.

PR #2228, opened 2017-12-15
Fix overlapping border radii handling.

### LINKED ISSUE #2174: Overlapping opposite rounded corners have drawing errors
https://bugzilla.mozilla.org/show_bug.cgi?id=1423603

[Testcase](https://bug1423603.bmoattachments.org/attachment.cgi?id=8935002)

Actual vs expected results:
screen shot 2017-12-06 at 10 15 29 am

```yaml
---
root:
  items:
    - type: stacking-context
      bounds: [0, 0, 1000, 1000]
      items:
        - type: clip
          bounds: [ 0, 0, 200, 200 ]
          complex:
            - rect: [ 0, 0, 200, 200 ]
              radius:
                top-left: [180, 180]
                top-right: [0, 0]
                bottom-left: [0, 0]
                bottom-right: [180, 180]
          items:
            - type: rect
              bounds: [ 0, 0, 200, 200 ]
              color: blue
```

glennw:
This ensures that the shader takes the max clipped distance
from each active clip region, which fixes this bug.

Additionally, apply the CSS spec radius clamping to clip rects.
This makes the results in the clip/clip-ellipse reftest match
Gecko, so I updated the reference image.

Fixes #2174.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2228)


glennw:
r? @kvark 

Pending try here:
https://hg.mozilla.org/try/rev/f76cba90007c9745674eca3df453b35a892233aa
https://treeherder.mozilla.org/#/jobs?repo=try&revision=f76cba90007c9745674eca3df453b35a892233aa

webrender/res/ellipse.glsl
```diff
@@ -84,9 +84,11 @@ float clip_against_ellipse_if_needed(
...
     }
 
-    return distance_to_ellipse(pos - ellipse_center_radius.xy,
-                               ellipse_center_radius.zw,
-                               aa_range);
+    float distance = distance_to_ellipse(pos - ellipse_center_radius.xy,
+                                         ellipse_center_radius.zw,
+                                         aa_range);
+
+    return max(distance, current_distance);
```
kvark: the semantics of `current_distance` is not clear to me. Shouldn't this be `min(distance, current_distance)`?

glennw: We want to take the maximum distance, so that if any of the active clip regions results in clipping the primitive out, it removes it.

glennw:
Try looks green for this PR.

kvark:
thanks for explanation!
@bors-servo r+

PR #2229, opened 2017-12-15
Fix box shadows when the shadow rect is smaller than the nine-patch.

### LINKED ISSUE #2179: Box-shadows with blur render incorrectly if there are opposite rounded corners whose bounding boxes overlap
https://bugzilla.mozilla.org/show_bug.cgi?id=1423683

[Testcase](https://bug1423683.bmoattachments.org/attachment.cgi?id=8935073)

Actual vs expected results:
screen shot 2017-12-06 at 1 59 59 pm

```yaml
---
root:
  items:
    - type: stacking-context
      bounds: [0, 0, 1000, 1000]
      items:
        - type: box-shadow
          bounds: [ 0, 0, 200, 200 ]
          color: red
          clip-mode: outset
          offset: 200 0
          blur-radius: 1
          spread-radius: 0
          border-radius:
            top-left: [100, 100]
            top-right: [0, 0]
            bottom-left: [0, 0]
            bottom-right: [150, 150]
```

### LINKED ISSUE #2180: Incorrect box-shadow rendering when the blur of adjacent rounded corners overlaps
https://bugzilla.mozilla.org/show_bug.cgi?id=1423692

[Testcase](https://bug1423692.bmoattachments.org/attachment.cgi?id=8935086)

Actual vs expected results:
screen shot 2017-12-06 at 2 23 46 pm

```yaml
---
root:
  items:
    - type: stacking-context
      bounds: [0, 0, 1000, 1000]
      items:
        - type: box-shadow
          bounds: [ 50, 10, 200, 80 ]
          color: black
          clip-mode: outset
          offset: 0 150
          blur-radius: 20
          border-radius:
            top-left: [100, 40]
            top-right: [0, 0]
            bottom-left: [100, 40]
            bottom-right: [0, 0]
```

glennw:
Fixes #2180.
Fixes #2179.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2229)


glennw:
r? @kvark 

Pending try:
https://hg.mozilla.org/try/rev/630569b923bc5b3d3f7d24f390b92711eacab776
https://treeherder.mozilla.org/#/jobs?repo=try&revision=630569b923bc5b3d3f7d24f390b92711eacab776

glennw:
There is one failure here. The reason this particular test keeps flipping between failing and not failing is because it has a fractional height, and is hitting this bug https://github.com/servo/webrender/issues/1943. I think we should probably mark this test as a fail until that is fixed. Try run looks green otherwise.

webrender/src/box_shadow.rs
```diff
@@ -207,6 +208,15 @@ impl FrameBuilder {
                         width = 2.0 * max_width + BLUR_SAMPLE_SCALE * blur_radius;
                         height = 2.0 * max_height + BLUR_SAMPLE_SCALE * blur_radius;
 
+                        // If the width or height ends up being bigger than the original
+                        // primitive shadow rect, just blur the entire rect and draw that
+                        // as a simple blit.
+                        if width > prim_info.rect.size.width || height > prim_info.rect.size.height {
```
kvark: is it excluding `corner_size.is_some()` case by intention?

webrender/src/box_shadow.rs
```diff
@@ -194,7 +195,7 @@ impl FrameBuilder {
                         // case, we ensure the size of each corner is the same,
                         // to simplify the shader logic that stretches the blurred
                         // result across the primitive.
-                        radii_kind = BorderRadiusKind::NonUniform;
+                        image_kind = BrushImageKind::NinePatch;
                         let max_width = shadow_radius.top_left.width
```
kvark: not directly related to the PR, but we could really use some [TypedPoint2D::max](https://docs.rs/euclid/0.16.0/euclid/struct.TypedPoint2D.html#method.max) here

glennw: This is a ```TypedSize``` which doesn't appear to have a ```max``` implementation (we should fix that though in euclid).

webrender/src/box_shadow.rs
glennw: Yep, that's handled in the branch above this one.

Issue #2231, opened 2017-12-18
[meta] Issues of binary recordings


kvark:
Problems I encountered:
  1. Wrench player has to be built from exactly the same WR revision as the recorder, otherwise bad things happen
  2. When replaying, all of the keyboard input, title frame number, and actually displayed frame - appear to update at different moments (de-synchronized). It's difficult to fast-forward in a predictable manner.
  3. Saving frames when running with `--play` doesn't get some of the last frames saved.
  4. Showing individually saved frames doesn't always get the same rendering as the binary replay.

glennw:
@kvark Do we need to keep this open? Should we just remove the old binary recording code completely now?

kvark:
I wouldn't rush with removing it yet. The new capturing is not continuous, let's give it some time to mature before considering to remove the recording. I'd keep this open for as long as it's either fixed, or the whole thing is removed :)

PR #2232, opened 2017-12-18
[RFC, alpha] Capturing infrastructure

kvark:
As we start re-focusing on the correctness issues, it becomes vital to be able to capture a bug and replay it in a reduced standalone environment. Unfortunately, the current binary/player infrastructure does not provide a solid way to achieve that, see #2231.

Changes to existing logic:
  - internal representation of `Document` on the WR backend side is changed to have a new `DocumentView` item.
  - `BuiltDisplayList/DisplayItemRef` serialization (used for JSON/RON) is rewritten to use stronger typing (with `CompletelySpecificDisplayItem` enum) and consistent/robust code. Result also looks a bit better now, since items have less nesting.
  - WR API got "debug-serialization" feature, and WR itself got "capture" feature

New API calls:
  - `save_capture(path)`
  - `load_capture(path)` - still not sure if we might instead just always create a new WR backend instead of trying to overwrite the current state with a capture (comments are welcome!)
  - In the examples, we can now hit 'C' to save a capture or load one, if it's already at "captures/example".

Here is what a capture looks like on disk:
capture-files

This is just the start. What will follow is a series of issues/PRs to extend the function in the following ways:
  1. full external/blob image serialization (this is the biggest piece missing to call this a "beta"). Currently, the following paths are incomplete:
      - #2236: tiled blobs
      - #2237: external GL textures
  2. #2234: storing images as PNG
  3. #2235: supporting other serial formats for the scenes, like JSON, with configurable pretty-printing
  4. wrench integration, removal of the existing JSON/RON writers
  5. #2238, #2239: more exciting stuff ;)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2232)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ script:
   - servo-tidy
   - if [ $BUILD_KIND = DEBUG ]; then (cd webrender_api && cargo test --verbose --features "ipc"); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo build --verbose --no-default-features); fi
-  - if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo build --verbose --features profiler); fi
+  - if [ $BUILD_KIND = DEBUG ]; then (cd webrender && cargo build --verbose --features profiler,capture); fi
   - if [ $BUILD_KIND = DEBUG ]; then (cargo test --all --verbose); fi
   - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && python script/headless.py reftest); fi
   - if [ $BUILD_KIND = RELEASE ]; then (cd wrench && cargo build --release); fi
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1019,8 +1019,8 @@ dependencies = [
  "plane-split 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "rayon 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "ron 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.23 (git+https://github.com/gankro/serde?branch=deserialize_from_enums3)",
- "serde_derive 1.0.23 (git+https://github.com/gankro/serde?branch=deserialize_from_enums3)",
  "serde_json 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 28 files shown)


glennw:

---

*[webrender_api/Cargo.toml, line 11 at r1](https://reviewable.io:443/reviews/servo/webrender/2232#-L0f0uzuHBSW66qPcczV:-L0f0uzvJ9FMLCWL06SU:b-owj33g) ([raw file](https://github.com/servo/webrender/blob/9a61d8cabebca48521806fd1d93932adbbd207c2/webrender_api/Cargo.toml#L11)):*
> ```TOML
> nightly = ["euclid/unstable", "serde/unstable"]
> ipc = ["ipc-channel"]
> serial = []
> ```

I'm not sure about the naming of this - perhaps we can either come up with something a bit more descriptive or add a comment explaining what this feature provides?

---

*[webrender_api/src/display_item.rs, line 58 at r1](https://reviewable.io:443/reviews/servo/webrender/2232#-L0f1mV12c8ZouT07_h2:-L0f1mV12c8ZouT07_h3:beoj6xn) ([raw file](https://github.com/servo/webrender/blob/9a61d8cabebca48521806fd1d93932adbbd207c2/webrender_api/src/display_item.rs#L58)):*
> ```Rust
> #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize)]
> pub struct GenericDisplayItem<T> {
>     pub item: T,
> ```

Do we need @Gankro or @jrmuizel to check that this won't have any bad side effects on serde codegen for (de)serialization times?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2232)*

kvark:
@glennw thanks for the review!
I've implemented a bit of missing critical logic for blobs and external buffers, renamed the wr-api feature to "debug-serialization", and confirmed with Jeff about the `GenericDisplayItem` changes being harmless.

glennw:
@kvark r=me if you want to get this merged now then, or we can revisit if there are still more commits to come first.

kvark:
@glennw thank you!
I believe this is a fairly complete PR that is ready to merge. But I don't want to rush it in (and now would be a perfect moment to do so, given how everyone is still getting relaxed after All Hands). Let's have a few more interested parties taking a look and then proceed.
r? @jrmuizel or @Gankro

jrmuizel:
I've only skimmed it but it seems fine to me.

kvark:
@jrmuizel please take a look at the last commit with an extra raw test, as requested. Afterwards, it should be ready to sail.

jrmuizel:
Looks good to me.

kvark:
Alright-y, thanks everyone who looked at the code!
@bors-servo r=glennw,jrmuizel

PR #2233, opened 2017-12-18
add reftest for color emoji fast shadows

lsalzman:
This is a follow-up to #2223. It ensures that the color bitmap's color does not improperly leak into the shadow. It is Mac only for now due to the relative impossibility of making cross-platform color bitmap fonts.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2233)


glennw:
Thanks!

@bors-servo r+

PR #2242, opened 2017-12-19
Extend segment building to handle clip in/out, multiple clip regions.

glennw:
Introduce segment builder, which is able to handle a much larger
range of clip scenarios than the initial segment implementation.

Instead of only handling the simple case of a single nine-patch
clip, we can now handle any number of local clip regions, and most
importantly, both clip and clip out modes. The support for clip-out
means we can often eliminate large areas of internal pixels for
items such as box-shadows. Note: the box-shadow code is not currently
hooked up to take advantage of this yet.

In the future, the segment builder can quite easily be extended to
handle the remaining clip sources that aren't handled - the image
mask and border corner clip sources.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2242)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -14,21 +14,8 @@ void brush_vs(
 #define RASTERIZATION_MODE_LOCAL_SPACE      0.0
 #define RASTERIZATION_MODE_SCREEN_SPACE     1.0
 
-#define SEGMENT_ALL             0
-#define SEGMENT_TOP_LEFT        1
-#define SEGMENT_TOP_RIGHT       2
-#define SEGMENT_BOTTOM_RIGHT    3
-#define SEGMENT_BOTTOM_LEFT     4
-#define SEGMENT_TOP_MID         5
-#define SEGMENT_MID_RIGHT       6
-#define SEGMENT_BOTTOM_MID      7
-#define SEGMENT_MID_LEFT        8
-#define SEGMENT_CENTER          9
-
-#define AA_KIND_DEFAULT         0
-#define AA_KIND_SEGMENT         1
-
-#define VECS_PER_BRUSH_PRIM                 4
+#define VECS_PER_BRUSH_PRIM                 2
+#define VECS_PER_SEGMENT                    2
 
 struct BrushInstance {
     int picture_address;
@@ -37,7 +24,7 @@ struct BrushInstance {
     int scroll_node_id;
     int clip_address;
     int z;
-    int segment_kind;
+    int segment_index;
     ivec2 user_data;
 };
 
@@ -50,7 +37,7 @@ BrushInstance load_brush() {
     bi.scroll_node_id = aData0.z % 65536;
     bi.clip_address = aData0.w;
     bi.z = aData1.x;
-    bi.segment_kind = aData1.y;
+    bi.segment_index = aData1.y;
     bi.user_data = aData1.zw;
 
     return bi;
@@ -59,18 +46,14 @@ BrushInstance load_brush() {
 struct BrushPrimitive {
     RectWithSize local_rect;
     RectWithSize local_clip_rect;
-    vec4 offsets;
-    int aa_kind;
 };
 
 BrushPrimitive fetch_brush_primitive(int address) {
-    vec4 data[4] = fetch_from_resource_cache_4(address);
+    vec4 data[2] = fetch_from_resource_cache_2(address);
 
     BrushPrimitive prim = BrushPrimitive(
         RectWithSize(data[0].xy, data[0].zw),
-        RectWithSize(data[1].xy, data[1].zw),
-        data[2],
-        int(data[3].x)
+        RectWithSize(data[1].xy, data[1].zw)
     );
 
     return prim;
@@ -86,67 +69,14 @@ void main(void) {
     BrushPrimitive brush_prim = fetch_brush_primitive(brush.prim_address);
 
     // Fetch the segment of this brush primitive we are drawing.
-    RectWithSize local_segment_rect;
-    vec4 edge_aa_segment_mask;
+    int segment_address = brush.prim_address +
+                          VECS_PER_BRUSH_PRIM +
+                          VECS_PER_SPECIFIC_BRUSH +
+                          brush.segment_index * VECS_PER_SEGMENT;
 
-    // p0 = origin of outer rect
-    // p1 = origin of inner rect
-    // p2 = bottom right corner of inner rect
-    // p3 = bottom right corner of outer rect
-    vec2 p0 = brush_prim.local_rect.p0;
-    vec2 p1 = brush_prim.local_rect.p0 + brush_prim.offsets.xy;
-    vec2 p2 = brush_prim.local_rect.p0 + brush_prim.local_rect.size - brush_prim.offsets.zw;
-    vec2 p3 = brush_prim.local_rect.p0 + brush_prim.local_rect.size;
-
-    switch (brush.segment_kind) {
-        case SEGMENT_ALL:
-            local_segment_rect = brush_prim.local_rect;
-            break;
-
-        case SEGMENT_TOP_LEFT:
-            local_segment_rect = RectWithSize(p0, p1 - p0);
-            break;
-        case SEGMENT_TOP_RIGHT:
-            local_segment_rect = RectWithSize(vec2(p2.x, p0.y), vec2(p3.x - p2.x, p1.y - p0.y));
-            break;
-        case SEGMENT_BOTTOM_RIGHT:
-            local_segment_rect = RectWithSize(vec2(p2.x, p2.y), vec2(p3.x - p2.x, p3.y - p2.y));
-            break;
-        case SEGMENT_BOTTOM_LEFT:
-            local_segment_rect = RectWithSize(vec2(p0.x, p2.y), vec2(p1.x - p0.x, p3.y - p2.y));
-            break;
-
-        case SEGMENT_TOP_MID:
-            local_segment_rect = RectWithSize(vec2(p1.x, p0.y), vec2(p2.x - p1.x, p1.y - p0.y));
-            break;
-        case SEGMENT_MID_RIGHT:
-            local_segment_rect = RectWithSize(vec2(p2.x, p1.y), vec2(p3.x - p2.x, p2.y - p1.y));
-            break;
-        case SEGMENT_BOTTOM_MID:
-            local_segment_rect = RectWithSize(vec2(p1.x, p2.y), vec2(p2.x - p1.x, p3.y - p2.y));
-            break;
-        case SEGMENT_MID_LEFT:
-            local_segment_rect = RectWithSize(vec2(p0.x, p1.y), vec2(p1.x - p0.x, p2.y - p1.y));
-            break;
-
-        case SEGMENT_CENTER:
-            local_segment_rect = RectWithSize(p1, p2 - p1);
-            break;
-
-        default:
-            local_segment_rect = RectWithSize(vec2(0.0), vec2(0.0));
-            break;
-    }
-
-    switch (brush_prim.aa_kind) {
-        case AA_KIND_SEGMENT:
-            // TODO: select these correctly based on the segment kind.
-            edge_aa_segment_mask = vec4(1.0);
-            break;
-        case AA_KIND_DEFAULT:
-            edge_aa_segment_mask = vec4(1.0);
-            break;
-    }
+    vec4[2] segment_data = fetch_from_resource_cache_2(segment_address);
+    RectWithSize local_segment_rect = RectWithSize(segment_data[0].xy, segment_data[0].zw);
+    vec4 edge_aa_segment_mask = vec4((int(segment_data[1].x) & ivec4(1,2,4,8)) != ivec4(0));
 
     vec2 device_pos, local_pos;
 
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#define VECS_PER_SPECIFIC_BRUSH 0
+
 #include shared,prim_shared,brush
 
 #ifdef WR_FEATURE_ALPHA_PASS
--- a/webrender/res/brush_mask_corner.glsl
+++ b/webrender/res/brush_mask_corner.glsl
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#define VECS_PER_SPECIFIC_BRUSH 1
+
 #include shared,prim_shared,ellipse,brush
 
 flat varying float vClipMode;
--- a/webrender/res/brush_mask_rounded_rect.glsl
+++ b/webrender/res/brush_mask_rounded_rect.glsl
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#define VECS_PER_SPECIFIC_BRUSH 4
+
 #include shared,prim_shared,ellipse,brush
 
 flat varying float vClipMode;
--- a/webrender/res/brush_solid.glsl
+++ b/webrender/res/brush_solid.glsl
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#define VECS_PER_SPECIFIC_BRUSH 1
+
 #include shared,prim_shared,brush
 
 flat varying vec4 vColor;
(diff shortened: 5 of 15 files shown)


glennw:
r? @kvark

This patch looks really big, but the majority of it (~500 lines) is unit tests. So it shouldn't be too tricky to review.

There is one remaining failure in a gecko try, which I'll be fixing today. But this can be reviewed when you have time.

glennw:
Ah, I forgot to update the unit tests for the auto edge flag calculations - working through those now.

Gecko try run is green though:
https://hg.mozilla.org/try/rev/9b4d2e18cdfaccb5a5f75924f7f6dfd83afc4519
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9b4d2e18cdfaccb5a5f75924f7f6dfd83afc4519

(One unrelated intermittent timeout, and one failure from the previous box-shadow PR in master).

glennw:
OK, all the unit tests, wrench tests are OK, and the gecko try is green.

kvark:
Amazing stuff!

---

Reviewed 13 of 15 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[webrender/src/border.rs, line 440 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0m7UxGoZNPZQgwkbT5:-L0m7UxGoZNPZQgwkbT6:bz838im) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/border.rs#L440)):*
> ```Rust
>                 let descriptor = BrushSegmentDescriptor {
>                     segments: vec![
>                         BrushSegment::new(
> ```

we could have a helper:
```rust
let segment = |x0, y0, x1, y1, mask| BrushSegment::new(
    LayerPoint::new(x0, y0),
    LayerSize::new(x1-x0, y1-y0),
    false,
    mask
};
```

---

*[webrender/src/border.rs, line 487 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0m876Fc9Q7fOhMazs1:-L0m876Fc9Q7fOhMazs2:bfhw5kg) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/border.rs#L487)):*
> ```Rust
>                     &info,
>                     border.left.color,
>                     Some(Box::new(descriptor)),
> ```

let's not box it now, since `BrushSegmentDescriptor` has a heap-allocated `Vec` inside already

---

*[webrender/src/prim_store.rs, line 1362 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mBARWALs-Gjaj1BWS:-L0mBARWALs-Gjaj1BWT:b-oc1vrv) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/prim_store.rs#L1362)):*
> ```Rust
>         // is either because we don't handle segmenting this clip source,
>         // or we have a clip source from a different coordinate system.
>         let mut has_global_mask = false;
> ```

would probably be easier to straight use `BrushClipMaskKind` here

---

*[webrender/src/prim_store.rs, line 1367 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mAWIm7AmLdQGqJ2Uu:-L0mAWIm7AmLdQGqJ2Uv:bt8w73j) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/prim_store.rs#L1367)):*
> ```Rust
>         // that we can.
>         for clip_item in clips {
>             if clip_item.coordinate_system_id == prim_context.scroll_node.coordinate_system_id {
> ```

can we early out of the loop iteration (with `continue`) if this is false? indenting a whole page of code here is not ideal otherwise

---

*[webrender/src/prim_store.rs, line 1425 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mBSwHSq3aCBL7Ift5:-L0mBSwHSq3aCBL7Ift6:b-cw97dp) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/prim_store.rs#L1425)):*
> ```Rust
>         match brush.segment_desc {
>             Some(ref mut segment_desc) => {
>                 segment_desc.clip_mask_kind = clip_mask_kind;
> ```

if this arm is taken, all the segment building is wasted?

---

*[webrender/src/segment.rs, line 42 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mDRbypouWVK-c4gbr:-L0mDRbypouWVK-c4gbs:b-uc6w68) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L42)):*
> ```Rust
>     fn cmp(&self, other: &EventKind) -> cmp::Ordering {
>         match (*self, *other) {
>             (EventKind::Begin, EventKind::Begin) |
> ```

Looks like just a reverse of regular ordering. You could achieve the same by
```rust
#[derive(PartialOrd, Ord)]
enum EventKind {
  End,
  Begin,
}
```

---

*[webrender/src/segment.rs, line 67 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mDP_YIzdODUnHiVoG:-L0mDP_ZFp_Qy8as3OEn:bmyopxm) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L67)):*
> ```Rust
> impl Ord for Event {
>     fn cmp(&self, other: &Event) -> cmp::Ordering {
>         match self.value.cmp(&other.value) {
> ```

should use [then](https://doc.rust-lang.org/stable/std/cmp/enum.Ordering.html#method.then)

---

*[webrender/src/segment.rs, line 269 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mF8kgPRR4egHQJtlR:-L0mF8kh3EH6_Pe30ImI:b-7lzbt7) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L269)):*
> ```Rust
>     // Consume this segment builder and produce a list of segments.
>     pub fn build<F>(self, mut f: F) where F: FnMut(Segment) {
>         if let Some(bounding_rect) = self.bounding_rect {
> ```

similarly, can early out here instead of indenting

---

*[webrender/src/segment.rs, line 285 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mFQGwqpxzO3HGSCmn:-L0mFQGx1sTy2PenmBmg:b-h1x8f2) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L285)):*
> ```Rust
>                 let p1 = item.rect.bottom_right();
> 
>                 x_events.push(Event::begin(p0.x, item_index));
> ```

alternatively, `x_events` and `y_events` could be `BTreeMap<Event, ItemIndex>`, which would avoid the need to have custom `Ord` implementation for `Event`

---

*[webrender/src/segment.rs, line 372 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mGXg6EuHB4db90lcG:-L0mGXg7h5MyajvXuXsa:bvr5bqc) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L372)):*
> ```Rust
>     bounds_p0: &LayerPointAu,
>     bounds_p1: &LayerPointAu,
> ) where F: FnMut(Segment) {
> ```

Could just return `Option<Segment>` for simpler code paths

---

*[webrender/src/segment.rs, line 449 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mGqO0wDXgUEJN3CO7:-L0mGqO0wDXgUEJN3CO8:bdgri4a) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L449)):*
> ```Rust
>         has_mask: bool,
>         edge_flags: Option<EdgeAaSegmentMask>,
>     ) -> Segment {
> ```

A habit of writing `-> Self` would pay for itself, I promise!

---

*[webrender/src/segment.rs, line 465 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mH5BIvc_XkwW3--kN:-L0mH5BIvc_XkwW3--kO:b2z4hg5) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L465)):*
> ```Rust
> 
>         if r0.origin.x != r1.origin.x {
>             r0.origin.x.partial_cmp(&r1.origin.x).unwrap()
> ```

can we just say `((r0.origin.x, r0.origin.y, r0.size.width, r0.size.height)).partial_cmp((r1.origin.x, r1.origin.y, r1.size.width, r1.size.height)).unwrap()` instead?

---

*[webrender/src/tiling.rs, line 409 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mC3W20ldqvgGzVJOh:-L0mC3W3I9iEosK4jsu_:bq98k4r) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/tiling.rs#L409)):*
> ```Rust
> 
>                     for (i, segment) in segment_desc.segments.iter().enumerate() {
>                         let is_inner = segment.edge_flags.is_empty();
> ```

smart :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2242#-:-L0mHe01uEkSxLMgaGEU:b-q1h7m3)*

glennw:
Review status: 11 of 15 files reviewed at latest revision, 13 unresolved discussions.

---

*[webrender/src/border.rs, line 440 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0m7UxGoZNPZQgwkbT5:-L0mVAsdpyjaZQovICuZ:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/border.rs#L440)):*

we could have a helper:
```rust
let segment = |x0, y0, x1, y1, mask| BrushSegment::new(
    LayerPoint::new(x0, y0),
    LayerSize::new(x1-x0, y1-y0),
    false,
    mask
};
```
</blockquote></details>

Done.

---

*[webrender/src/border.rs, line 487 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0m876Fc9Q7fOhMazs1:-L0mVBjnlIChMMAjVfnT:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/border.rs#L487)):*

let's not box it now, since `BrushSegmentDescriptor` has a heap-allocated `Vec` inside already
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1362 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mBARWALs-Gjaj1BWS:-L0mVCdLac0seW7cOTi3:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/prim_store.rs#L1362)):*

would probably be easier to straight use `BrushClipMaskKind` here
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1367 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mAWIm7AmLdQGqJ2Uu:-L0mVDUnOz8w5EPd0evp:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/prim_store.rs#L1367)):*

can we early out of the loop iteration (with `continue`) if this is false? indenting a whole page of code here is not ideal otherwise
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1425 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mBSwHSq3aCBL7Ift5:-L0mVEHIal7VqnaDlYNk:b-9gu88m) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/prim_store.rs#L1425)):*

if this arm is taken, all the segment building is wasted?
</blockquote></details>

The accumulation of the rects is, but the build process itself is skipped. We also only run this the first time through (then the clip mask kind is cached).

---

*[webrender/src/segment.rs, line 42 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mDRbypouWVK-c4gbr:-L0mVRqYMua1fMtLbmEx:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L42)):*

Looks like just a reverse of regular ordering. You could achieve the same by
```rust
#[derive(PartialOrd, Ord)]
enum EventKind {
  End,
  Begin,
}
```
</blockquote></details>

Done.

---

*[webrender/src/segment.rs, line 67 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mDP_YIzdODUnHiVoG:-L0mVSZt3Uwq7F9RL07t:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L67)):*

should use [then](https://doc.rust-lang.org/stable/std/cmp/enum.Ordering.html#method.then)
</blockquote></details>

Done.

---

*[webrender/src/segment.rs, line 269 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mF8kgPRR4egHQJtlR:-L0mVTb6A5YBydIOQ8wB:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L269)):*

similarly, can early out here instead of indenting
</blockquote></details>

Done.

---

*[webrender/src/segment.rs, line 372 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mGXg6EuHB4db90lcG:-L0mVW_OTuTsve3Qyuj4:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L372)):*

Could just return `Option<Segment>` for simpler code paths
</blockquote></details>

Done.

---

*[webrender/src/segment.rs, line 449 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mGqO0wDXgUEJN3CO7:-L0mVXHN9STPK1Ut3Msx:b-e9dfov) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L449)):*

A habit of writing `-> Self` would pay for itself, I promise!
</blockquote></details>

Point taken :) It doesn't work here though, where it's a free function.

---

*[webrender/src/segment.rs, line 465 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mH5BIvc_XkwW3--kN:-L0mVcbBGMF4zpFbYAf1:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L465)):*

can we just say `((r0.origin.x, r0.origin.y, r0.size.width, r0.size.height)).partial_cmp((r1.origin.x, r1.origin.y, r1.size.width, r1.size.height)).unwrap()` instead?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2242)*

glennw:
Thanks for the quick review! All comments addressed, I think. Although I just noticed that I broke a reftest - fixing now.

glennw:
Ah, found the issue. Auto-deriving the ```Ord``` for ```EventKind``` gave the wrong event ordering. I changed it back to an explicit ```Ord``` implementation, since I think it's good to be explicit here that the ordering of events is important. So I think that should cover all the review comments now :)

kvark:
> Auto-deriving the Ord for EventKind gave the wrong event ordering. 

This is *really* strange, isn't it? Are you sure you specified it to have `End` followed by `Begin` and not the other way around?

Anyhow, code is :lgtm_strong: 

---

Reviewed 4 of 4 files at r3.

---

*[webrender/src/segment.rs, line 449 at r2](https://reviewable.io:443/reviews/servo/webrender/2242#-L0mGqO0wDXgUEJN3CO7:-L0oUIvjqeDq5iPTUqLZ:bo4v0vl) ([raw file](https://github.com/servo/webrender/blob/6ef9d0255a428fe7daf8e43f77ca2698be1d6705/webrender/src/segment.rs#L449)):*

Point taken :) It doesn't work here though, where it's a free function.
</blockquote></details>

d'oh, you got me! of all the places where I could say that, I chose the one being an exception :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2242#-:-L0oURgCjsRMQEGX4P4d:b-xplf8e)*

kvark:
those mac bots are lazy again
@bors-servo retry

kvark:
@glennw given that the only threshold here is `Au`, and the asymptotic of the produced segment count is O(input_count ^ 2), wouldn't that result in a huge number of small pieces, potentially?

PR #2243, opened 2017-12-20
Fix drop-shadow render incorrectly on viewport's edge.

### LINKED ISSUE #2197: drop-shadow render incorrectly if the shadow close to the edge of viewport.
testcase

```
---
root:
  items:
    - type: stacking-context
      bounds: [10, 10, 400, 400]
      filters: drop-shadow([0, 0], 20, [255, 0, 0, 1])
      items:
      - type: rect
        bounds: 0 0 256 256
        color: green
```

mephisto41:
Initially, I apply the drop-shadow's offset before render the blur
result. This has problem if the blur is close to edge of
viewport because the blur result might be crop by the viewport.
This patch apply the offset after the blur result generated.

Fix #2197.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2243)


mephisto41:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=ec23089146b038ce93f144037bed510eeacdea3e&selectedJob=152518179

Gecko try here, the long-chain.html get fixed after apply this patch. We can replace the fails-if in long-chain.html with fuzzy value.

webrender/src/picture.rs
```diff
@@ -232,7 +234,7 @@ impl PicturePrimitive {
         let local_content_rect = prim_run_rect.local_rect_in_actual_parent_space;
 
         match self.kind {
-            PictureKind::Image { composite_mode, ref mut real_local_rect, .. } => {
+            PictureKind::Image { composite_mode, ref mut real_local_rect, ref mut content_rect, .. } => {
```
kvark: would it (possibly, in the future) make sense outside of `FilterOp::DropShadow`? If not, we may move it inside the `DropShadow variant` for clarity.

webrender/src/picture.rs
```diff
@@ -350,13 +353,20 @@ impl PicturePrimitive {
                         let blur_render_task_id = render_tasks.add(blur_render_task);
                         self.render_task_id = Some(blur_render_task_id);
                     }
-                    Some(PictureCompositeMode::Filter(FilterOp::DropShadow(offset, blur_radius, color))) => {
+                    Some(PictureCompositeMode::Filter(FilterOp::DropShadow(_, blur_radius, color))) => {
+                        let content_width =
```
kvark: perhaps, we can just do this one liner instead?
``` rust
let rect = (content_rect * TypedScale::<_, DevicePixel>::new(prim_context.device_pixel_ratio).round().to_i32();
```

@glennw btw, I think it would make sense to just store the `device_pixel_ratio` as that `TypedScale` in the first place

webrender/src/picture.rs
```diff
@@ -242,8 +244,8 @@ impl PicturePrimitive {
                     }
                     Some(PictureCompositeMode::Filter(FilterOp::DropShadow(offset, blur_radius, _))) => {
                         let inflate_size = blur_radius * BLUR_SAMPLE_SCALE;
-                        local_content_rect.inflate(inflate_size, inflate_size)
-                                          .translate(&offset)
+                        *content_rect = local_content_rect.inflate(inflate_size, inflate_size);
+                        content_rect.translate(&offset)
```
kvark: Note that technically you can apply the translation later, at the last moment.
This would allow you to merge those `match` arms, since you only need the `blur_radius` out of both to inflate the rect

webrender/src/tiling.rs
```diff
@@ -775,17 +776,19 @@ fn add_to_batch(
                                                 secondary_textures,
                                             );
                                             let batch = batch_list.get_suitable_batch(key, &item_bounding_rect);
-                                            let device_offset_x = device_length(offset.x, ctx.device_pixel_ratio);
-                                            let device_offset_y = device_length(offset.y, ctx.device_pixel_ratio);
+                                            let content_origin_x = device_length(content_rect.origin.x, ctx.device_pixel_ratio);
```
kvark: same oee liner trick can be done here

glennw:
Looks good to me once @kvark 's concerns are addressed.

mephisto41:
Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[webrender/src/picture.rs, line 237 at r1](https://reviewable.io:443/reviews/servo/webrender/2243#-L0lxlF4Itvi2FliJzDT-r1-237:-L0sPUTvcYStIf0MSI45:b-896fix) ([raw file](https://github.com/servo/webrender/blob/17e618e15df34067052645d10839ddb18aa47cc9/webrender/src/picture.rs#L237)):*

would it (possibly, in the future) make sense outside of `FilterOp::DropShadow`? If not, we may move it inside the `DropShadow variant` for clarity.
</blockquote></details>

Done.

---

*[webrender/src/picture.rs, line 248 at r1](https://reviewable.io:443/reviews/servo/webrender/2243#-L0lxlF4Itvi2FliJzDT-r1-248:-L0sPUrsyFW1722tUFV-:b-896fix) ([raw file](https://github.com/servo/webrender/blob/17e618e15df34067052645d10839ddb18aa47cc9/webrender/src/picture.rs#L248)):*

Note that technically you can apply the translation later, at the last moment.
This would allow you to merge those `match` arms, since you only need the `blur_radius` out of both to inflate the rect
</blockquote></details>

Done.

---

*[webrender/src/picture.rs, line 357 at r1](https://reviewable.io:443/reviews/servo/webrender/2243#-L0lxlF4Itvi2FliJzDT-r1-357:-L0sPV_i2TUsT4tMkOhr:b-896fix) ([raw file](https://github.com/servo/webrender/blob/17e618e15df34067052645d10839ddb18aa47cc9/webrender/src/picture.rs#L357)):*

perhaps, we can just do this one liner instead?
``` rust
let rect = (content_rect * TypedScale::<_, DevicePixel>::new(prim_context.device_pixel_ratio).round().to_i32();
```

@glennw btw, I think it would make sense to just store the `device_pixel_ratio` as that `TypedScale` in the first place
</blockquote></details>

Done.

---

*[webrender/src/tiling.rs, line 779 at r1](https://reviewable.io:443/reviews/servo/webrender/2243#-L0lxlF4Itvi2FliJzDU-r1-779:-L0sPY36OgiQacbZuXlj:b-896fix) ([raw file](https://github.com/servo/webrender/blob/17e618e15df34067052645d10839ddb18aa47cc9/webrender/src/tiling.rs#L779)):*

same one liner trick can be done here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2243)*

kvark:
Reviewed 3 of 5 files at r1, 6 of 6 files at r2.

---

*[webrender/src/picture.rs, line 248 at r1](https://reviewable.io:443/reviews/servo/webrender/2243#-L0lxlF4Itvi2FliJzDT-r1-248:-L0t_b0PyatYT1R1G_m7:b7pcdf9) ([raw file](https://github.com/servo/webrender/blob/17e618e15df34067052645d10839ddb18aa47cc9/webrender/src/picture.rs#L248)):*
<details><summary><i>Previously, mephisto41 (Morris Tseng) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Sorry for being unclear, this is not exactly what I meant. By "later" I mean at the time where it's used (in `prepare_for_render`), not right here after the match.

---

*[webrender/src/picture.rs, line 357 at r1](https://reviewable.io:443/reviews/servo/webrender/2243#-L0lxlF4Itvi2FliJzDT-r1-357:-L0taGyDDvBZhaQn_loK:by90lq) ([raw file](https://github.com/servo/webrender/blob/17e618e15df34067052645d10839ddb18aa47cc9/webrender/src/picture.rs#L357)):*
<details><summary><i>Previously, mephisto41 (Morris Tseng) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Please note that I went and refactored most of the device/pixel ratio logic to have some type safety in #2246. Sorry about the need for rebase on that! If this PR was ready, I'd gladly yield and rebase on top.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2243)*

glennw:
I think this is ready to merge once rebased @kvark @mephisto41 ?

mephisto41:
It has 1 nit should be fixed. I'll fix it today.

kvark:
Reviewed 4 of 4 files at r3.

---

*[webrender/src/prim_store.rs, line 1750 at r3](https://reviewable.io:443/reviews/servo/webrender/2243#-L1x5zkk-l9zjaWLIQgL:-L1x5zkleBSsop-2emf0:bnut7gj) ([raw file](https://github.com/servo/webrender/blob/5cc0b8a3474a18efd025da6d3e83f53bd5a92ac1/webrender/src/prim_store.rs#L1750)):*
> ```Rust
>             );
> 
>             match pic.kind {
> ```

can this be done inside `update_local_rect`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2243)*

mephisto41:
I made some significant change here. Since I realized that the ```content_rect``` can be reconstructed by ```prim_metadata.local_rect - drop-shadow-offset```. So I can no longer store content_rect in FilterOp::DropShadow. This change makes the patch cleaner.

kvark:
:lgtm: needs rebase though

---

Reviewed 3 of 7 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2243#-:-L21w0GlpK-BTPexv0vw:bmn0ngl)*

kvark:
Thank you!
@bors-servo r+

kvark:
Ah, since the PR got largely rewritten, could you fire up another try push please?

mephisto41:
Yes, latest try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=99ce5ccd151694845a486b3692dbd8e490eed5a5

kvark:
There is an R7 in `long-chain.html` failing there.

mephisto41:
See the comment 1 for explanation. Before this patch, the long-chain test is completely wrong so we set fails-if for it. After this patch, you can see that I removed the fails-if annotation in the try and the long-chain result is correct although we need tweak some fuzzy value for it. So the long-chain failure is expected. Just some fuzzy value change is fine.

kvark:
Thanks!
@bors-servo r+

glennw:
@bors-servo retry
 *sigh*

Issue #2111, opened 2017-11-27
Upgrade the reftest list format


kvark:
Looking into the growing list of supported prefixes/functions now: https://github.com/servo/webrender/pull/2105/files#diff-f340cf8bfec5772b98e4211803fbe7cdR219

The format chosen ("xxx.list") doesn't seem to scale nicely, given that all the prefixes are expected to come on the same line... so the tests don't align, and the very fact we are using a simple text format falls down here.

I suggest describing the tests in a structured format, e.g. RON:
```rust
[
  (
    op: NonEqual("text.yaml", "blank.yaml"), // or we could have src/dst fields
    fuzz: (diff: 1, count: 100),
    aa: false,
    platforms: [Linux],
    draw_calls: 4,
  ),
  (
    op: Equal("decorations.yaml", "decorations-suite.png"),
    aa: true,
  )
]
```

PR #2244, opened 2017-12-20
Add dual-source blending path for subpixel text when supported.

glennw:
This makes it possible to share batches when the text color is
different between runs. This can result in significantly better
draw call batching.

Add option to wrench to disable dual source blending path.

Add reftest option to control if dual-source blending is used.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2244)

### DIFF
--- a/webrender/res/ps_text_run.glsl
+++ b/webrender/res/ps_text_run.glsl
@@ -22,8 +22,9 @@ varying vec4 vUvClip;
 #define MODE_SUBPX_BG_PASS0     4
 #define MODE_SUBPX_BG_PASS1     5
 #define MODE_SUBPX_BG_PASS2     6
-#define MODE_BITMAP             7
-#define MODE_COLOR_BITMAP       8
+#define MODE_SUBPX_DUAL_SOURCE  7
+#define MODE_BITMAP             8
+#define MODE_COLOR_BITMAP       9
 
 VertexInfo write_text_vertex(vec2 clamped_local_pos,
                              RectWithSize local_clip_rect,
@@ -134,6 +135,7 @@ void main(void) {
             break;
         case MODE_SUBPX_PASS1:
         case MODE_SUBPX_BG_PASS2:
+        case MODE_SUBPX_DUAL_SOURCE:
             vMaskSwizzle = vec2(1.0, 0.0);
             vColor = text.color;
             break;
@@ -170,6 +172,12 @@ void main(void) {
     alpha *= float(all(greaterThanEqual(vUvClip, vec4(0.0))));
 #endif
 
+#ifdef WR_FEATURE_DUAL_SOURCE_BLENDING
+    vec4 alpha_mask = mask * alpha;
+    oFragColor = vColor * alpha_mask;
+    oFragBlend = alpha_mask * vColor.a;
+#else
     oFragColor = vColor * mask * alpha;
+#endif
 }
 #endif
--- a/webrender/res/shared.glsl
+++ b/webrender/res/shared.glsl
@@ -8,6 +8,10 @@
 #extension GL_OES_EGL_image_external_essl3 : require
 #endif
 
+#ifdef WR_FEATURE_DUAL_SOURCE_BLENDING
+#extension GL_ARB_explicit_attrib_location : require
+#endif
+
 #include base
 
 // The textureLod() doesn't support samplerExternalOES for WR_FEATURE_TEXTURE_EXTERNAL.
@@ -46,7 +50,12 @@
     // Uniform inputs
 
     // Fragment shader outputs
-    out vec4 oFragColor;
+    #ifdef WR_FEATURE_DUAL_SOURCE_BLENDING
+        layout(location = 0, index = 0) out vec4 oFragColor;
+        layout(location = 0, index = 1) out vec4 oFragBlend;
+    #else
+        out vec4 oFragColor;
+    #endif
 #endif
 
 //======================================================================================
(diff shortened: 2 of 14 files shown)


glennw:
r? @kvark 

The dual-source integration itself is reasonably simple. The code to weave all the options through wrench etc is a bug ugly - if you have suggestions for another way to achieve that, let me know :)

(Will start a gecko try shortly)

glennw:
Anecdotally this does seem to give a significant performance win on several sites I tested (not all sites - there are a couple of other known causes of bad batch break behaviour we need to fix too).

glennw:
Try looks good, I think.

One fail that's expected from current master, a couple of unrelated intermittents.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=f410cd46b756acd94e0ab1ed5d20f7eb5b306485

kvark:
> the code to weave all the options through wrench etc is a bit ugly - if you have suggestions for another way to achieve that, let me know :)

Yes, this is exactly what I meant in #2111 by saying:
> The format chosen ("xxx.list") doesn't seem to scale nicely

---

---

*[webrender/src/frame_builder.rs, line 77 at r1](https://reviewable.io:443/reviews/servo/webrender/2244#-L0oWlpcok8QGmgy8dKU:-L0oWlpcok8QGmgy8dKV:bp1epft) ([raw file](https://github.com/servo/webrender/blob/5f6b386870f6ff0bae9ba221181b2ab9675591a3/webrender/src/frame_builder.rs#L77)):*
> ```Rust
>     pub default_font_render_mode: FontRenderMode,
>     pub debug: bool,
>     pub dual_source_blending_is_supported: bool,
> ```

this seems a bit out of place, shouldn't it be inside `Device` instead?

---

*[webrender/src/render_backend.rs, line 595 at r1](https://reviewable.io:443/reviews/servo/webrender/2244#-L0oWwt7mN4mAOcRR6o8:-L0oWwt7mN4mAOcRR6o9:bhe0f56) ([raw file](https://github.com/servo/webrender/blob/5f6b386870f6ff0bae9ba221181b2ab9675591a3/webrender/src/render_backend.rs#L595)):*
> ```Rust
>                             // that are created.
>                             self.frame_config
>                                 .dual_source_blending_is_enabled = enable;
> ```

should this check for `_supported` before setting?

---

*[webrender_api/src/api.rs, line 779 at r1](https://reviewable.io:443/reviews/servo/webrender/2244#-L0oXgR5HiphF6ItVmds:-L0oXgR6X0xWEfeCjCzh:b-vzm1h3) ([raw file](https://github.com/servo/webrender/blob/5f6b386870f6ff0bae9ba221181b2ab9675591a3/webrender_api/src/api.rs#L779)):*
> ```Rust
>     pub fn load_capture(&self, path: PathBuf) {
>         let msg = ApiMsg::DebugCommand(DebugCommand::LoadCapture(path));
>         self.send_message(msg);
> ```

jeez, it's almost 2018 and the diff programs are not yet smart enough to detect the new code along the less indented boundary :(

---

*[wrench/reftests/text/reftest.list, line 33 at r1](https://reviewable.io:443/reviews/servo/webrender/2244#-L0oXuR5Ch-pU35OD6Sc:-L0oXuR6Viz4i4RMK3lu:bb7d7pz) ([raw file](https://github.com/servo/webrender/blob/5f6b386870f6ff0bae9ba221181b2ab9675591a3/wrench/reftests/text/reftest.list#L33)):*
> ```
> fuzzy(1,4) platform(linux) options(disable-subpixel) == colors.yaml colors-alpha.png
> # Run without dual-source blending path, batches are broken when text colors change.
> fuzzy(1,6) platform(linux) options(disable-dual-source-blending) draw_calls(5) == colors.yaml colors-subpx.png
> ```

argh

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2244#-:-L0oYaoTEuxZwnihixle:bb4axqq)*

kvark:
I only got minor concerns about this, feel free to r=me please.
And thanks!

glennw:

---

*[webrender/src/frame_builder.rs, line 77 at r1](https://reviewable.io:443/reviews/servo/webrender/2244#-L0oWlpcok8QGmgy8dKU:-L0p_kjD9HLgqpm6oDIa:b-gecf8y) ([raw file](https://github.com/servo/webrender/blob/5f6b386870f6ff0bae9ba221181b2ab9675591a3/webrender/src/frame_builder.rs#L77)):*

this seems a bit out of place, shouldn't it be inside `Device` instead?
</blockquote></details>

This is where it's a bit weird how it threads the information through. The main issue is that the render backend needs to know this, since it affects batching (whether the color of the text is relevant to the batch key). Since the backend doesn't have access to the Device, we need to provide it here.

---

*[webrender/src/render_backend.rs, line 595 at r1](https://reviewable.io:443/reviews/servo/webrender/2244#-L0oWwt7mN4mAOcRR6o8:-L0p_zJZ8ELZq9QXRCMV:big71gz) ([raw file](https://github.com/servo/webrender/blob/5f6b386870f6ff0bae9ba221181b2ab9675591a3/webrender/src/render_backend.rs#L595)):*

should this check for `_supported` before setting?
</blockquote></details>

The enabled flag is really more like "enabled if available" - it's ANDed with the is available flag when passing the details to the batching code on whether to use dual-source blending.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2244)*

glennw:
@kvark Thanks for the review. Added a couple of explanations for your questions. Let me know if there's changes needed based on that, otherwise I think this is ready to go :)

kvark:
Thanks!
@bors-servo r+

PR #2245, opened 2017-12-20
implement synthetic italics with a skew transform on all platforms

lsalzman:
This tries to implement synthetic italics in a way that is compatible with how Gecko does it. That means we no longer use the myriad backend APIs for this, but instead supply skew transforms for the outline that match the transform Gecko is using for metrics.

FreeType requires a bit of extra care in that transforms only apply to outlines and not to bitmaps. So bitmap data must unfortunately be manually transformed to accomplish this.  However, FreeType's own synthetic italics mechanism did not work for bitmaps at all either, so this rewrite was a necessity there anyway. The Windows/Mac backends take care of the bitmap data transforms themselves, so we at least avoid the complexity in those.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2245)

### DIFF
--- a/webrender/src/platform/macos/font.rs
+++ b/webrender/src/platform/macos/font.rs
@@ -22,7 +22,7 @@ use core_text;
 use core_text::font::{CTFont, CTFontRef};
 use core_text::font_descriptor::{kCTFontDefaultOrientation, kCTFontColorGlyphsTrait};
 use gamma_lut::{ColorLut, GammaLut};
-use glyph_rasterizer::{FontInstance, GlyphFormat, RasterizedGlyph};
+use glyph_rasterizer::{FontInstance, FontTransform, GlyphFormat, RasterizedGlyph};
 use internal_types::FastHashMap;
 use std::collections::hash_map::Entry;
 use std::sync::Arc;
@@ -264,6 +264,9 @@ fn is_bitmap_font(ct_font: &CTFont) -> bool {
     (traits & kCTFontColorGlyphsTrait) != 0
 }
 
+// Skew factor matching Gecko/CG.
+const OBLIQUE_SKEW_FACTOR: f32 = 0.25;
+
 impl FontContext {
     pub fn new() -> FontContext {
         debug!("Test for subpixel AA support: {}", supports_subpixel_aa());
@@ -358,7 +361,28 @@ impl FontContext {
                 let glyph = key.index as CGGlyph;
                 let bitmap = is_bitmap_font(ct_font);
                 let (x_offset, y_offset) = if bitmap { (0.0, 0.0) } else { font.get_subpx_offset(key) };
-                let metrics = get_glyph_metrics(ct_font, None, glyph, x_offset, y_offset, 0.0);
+                let transform = if font.flags.contains(FontInstanceFlags::SYNTHETIC_ITALICS) {
+                    let shape = FontTransform::identity().synthesize_italics(OBLIQUE_SKEW_FACTOR);
+                    Some(CGAffineTransform {
+                        a: shape.scale_x as f64,
+                        b: -shape.skew_y as f64,
+                        c: -shape.skew_x as f64,
+                        d: shape.scale_y as f64,
+                        tx: 0.0,
+                        ty: 0.0,
+                    })
+                } else {
+                    None
+                };
+                let extra_strikes = font.get_extra_strikes(1.0);
+                let metrics = get_glyph_metrics(
+                    ct_font,
+                    transform.as_ref(),
+                    glyph,
+                    x_offset,
+                    y_offset,
+                    extra_strikes as f64,
+                );
                 if metrics.rasterized_width == 0 || metrics.rasterized_height == 0 {
                     None
                 } else {
@@ -454,22 +478,28 @@ impl FontContext {
         };
 
         let bitmap = is_bitmap_font(&ct_font);
-        let shape = font.transform.pre_scale(y_scale.recip() as f32, y_scale.recip() as f32);
-        let transform = if bitmap || shape.is_identity() {
-            None
+        let (mut shape, (x_offset, y_offset)) = if bitmap {
+            (FontTransform::identity(), (0.0, 0.0))
         } else {
+            (font.transform.invert_scale(y_scale, y_scale), font.get_subpx_offset(key))
+        };
+        if font.flags.contains(FontInstanceFlags::SYNTHETIC_ITALICS) {
+            shape = shape.synthesize_italics(OBLIQUE_SKEW_FACTOR);
+        }
+        let transform = if !shape.is_identity() {
             Some(CGAffineTransform {
                 a: shape.scale_x as f64,
                 b: -shape.skew_y as f64,
                 c: -shape.skew_x as f64,
                 d: shape.scale_y as f64,
                 tx: 0.0,
-                ty: 0.0
+                ty: 0.0,
             })
+        } else {
+            None
         };
 
         let glyph = key.index as CGGlyph;
-        let (x_offset, y_offset) = if bitmap { (0.0, 0.0) } else { font.get_subpx_offset(key) };
         let (strike_scale, pixel_step) = if bitmap { (y_scale, 1.0) } else { (x_scale, y_scale / x_scale) };
         let extra_strikes = font.get_extra_strikes(strike_scale);
         let metrics = get_glyph_metrics(
--- a/webrender/src/platform/unix/font.rs
+++ b/webrender/src/platform/unix/font.rs
@@ -53,7 +53,56 @@ unsafe impl Send for FontContext {}
 
 extern "C" {
     fn FT_GlyphSlot_Embolden(slot: FT_GlyphSlot);
-    fn FT_GlyphSlot_Oblique(slot: FT_GlyphSlot);
+}
+
+// Skew factor matching Gecko/FreeType.
+const OBLIQUE_SKEW_FACTOR: f32 = 0.2;
+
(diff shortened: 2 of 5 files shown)


glennw:
This generally looks good to me (haven't looked in detail yet). Could we add a reftest for at linux and/or mac? I know this shouldn't affect any reftests but perhaps we should do a gecko try too?

lsalzman:
There's already a synthetic italics reftest. As for Gecko, this is not rigged up in Gecko at all, so a try run isn't going to be able to test anything there yet.

webrender/src/platform/macos/font.rs
```diff
@@ -358,7 +360,31 @@ impl FontContext {
                 let glyph = key.index as CGGlyph;
                 let bitmap = is_bitmap_font(ct_font);
                 let (x_offset, y_offset) = if bitmap { (0.0, 0.0) } else { font.get_subpx_offset(key) };
-                let metrics = get_glyph_metrics(ct_font, None, glyph, x_offset, y_offset, 0.0);
+                let mut shape = FontTransform::identity();
```
glennw: I don't think we need to create ```shape``` here unconditionally. It's probably tidier to make the ```let transform``` line below be based on the ```flags``` rather than checking the shape matrix is identity.

webrender/src/platform/macos/font.rs
```diff
@@ -358,7 +360,31 @@ impl FontContext {
...
+                    Some(CGAffineTransform {
+                        a: shape.scale_x as f64,
+                        b: -shape.skew_y as f64,
+                        c: -shape.skew_x as f64,
+                        d: shape.scale_y as f64,
+                        tx: 0.0,
+                        ty: 0.0,
+                    })
+                };
+                let extra_strikes = font.get_extra_strikes(1.0);
```
glennw: Was this just a pre-existing bug, or is it related to this change?

webrender/src/platform/macos/font.rs
```diff
@@ -454,8 +480,15 @@ impl FontContext {
...
 
         let bitmap = is_bitmap_font(&ct_font);
-        let shape = font.transform.pre_scale(y_scale.recip() as f32, y_scale.recip() as f32);
-        let transform = if bitmap || shape.is_identity() {
+        let (mut shape, (x_offset, y_offset)) = if bitmap {
+            (FontTransform::identity(), (0.0, 0.0))
+        } else {
+            (font.transform.invert_scale(y_scale), font.get_subpx_offset(key))
+        };
+        if font.flags.contains(FontInstanceFlags::SYNTHETIC_ITALICS) {
```
glennw: Same here as comment above.

webrender/src/platform/unix/font.rs
```diff
@@ -53,7 +53,56 @@ unsafe impl Send for FontContext {}
 
 extern "C" {
     fn FT_GlyphSlot_Embolden(slot: FT_GlyphSlot);
-    fn FT_GlyphSlot_Oblique(slot: FT_GlyphSlot);
+}
+
+const OBLIQUE_SKEW_FACTOR: f32 = 0.2;
```
glennw: Why is this value different between mac and linux? We could perhaps add a comment referencing where these numbers are based on.

webrender/src/platform/unix/font.rs
```diff
@@ -53,7 +53,56 @@ unsafe impl Send for FontContext {}
...
+
+fn get_skew_bounds(bottom: i32, top: i32) -> (f32, f32) {
+    let skew_min = ((bottom as f32 + 0.5) * OBLIQUE_SKEW_FACTOR).floor();
+    let skew_max = ((top as f32 - 0.5) * OBLIQUE_SKEW_FACTOR).ceil();
+    (skew_min, skew_max)
+}
+
+fn skew_bitmap(bitmap: &mut Vec<u8>, width: &mut usize, height: usize, left: &mut i32, top: i32) {
```
glennw: Passing in a ```&mut usize``` to modify is a bit odd in rust - I think this would be tidier if we return new bounds, or something like that?

webrender/src/platform/unix/font.rs
```diff
@@ -53,7 +53,56 @@ unsafe impl Send for FontContext {}
...
+    let skew_width = *width + (skew_max - skew_min) as usize;
+    let mut skew_buffer = vec![0u8; skew_width * height * 4];
+    for y in 0 .. height {
+        // Calculate a skew offset at the vertical center of the current row.
+        let offset = (top as f32 - y as f32 - 0.5) * OBLIQUE_SKEW_FACTOR - skew_min;
+        // Get a blend factor in 0..256 constant across all pixels in the row.
+        let blend = (offset.fract() * 256.0) as u32;
+        let mut dest = (y * skew_width + offset.floor() as usize) * 4;
+        let (mut prev_b, mut prev_g, mut prev_r, mut prev_a) = (0u32, 0u32, 0u32, 0u32);
+        for src in bitmap[y * stride .. (y + 1) * stride].chunks(4) {
```
glennw: This could probably be done using ```zip``` and ```iter_mut``` so that it's not necessary to manually increment the dest buffer index?

webrender/src/platform/windows/font.rs
```diff
@@ -239,7 +235,23 @@ impl FontContext {
     ) -> Option<GlyphDimensions> {
         let size = font.size.to_f32_px();
         let bitmaps = is_bitmap_font(font);
-        let analysis = self.create_glyph_analysis(font, key, size, None, bitmaps);
+        let mut shape = FontTransform::identity();
+        if font.flags.contains(FontInstanceFlags::SYNTHETIC_ITALICS) {
+            shape = shape.synthesize_italics(OBLIQUE_SKEW_FACTOR);
+        }
+        let transform = if shape.is_identity() {
```
glennw: Same here with not creating the font transform.

glennw:
Adding delegate since I'll probably be on PTO by the time these comments are addressed. @kvark may also have some comments.

@bors-servo delegate+

webrender/src/platform/unix/font.rs
lsalzman: Gecko uses different skew factors for all 3 platforms to more closely match the native skewing of the various APIs.

webrender/src/platform/unix/font.rs
lsalzman: It seemed a bit obtuse to have to temporarily assign to variables only to then reassign to actual_width/left with those, since tuple assignments are not yet in the language.

webrender/src/platform/macos/font.rs
lsalzman: A pre-existing bug/oversight.

webrender/src/platform/macos/font.rs
lsalzman: Can't do that here due to the fact that the initial transform is not necessarily the identity depending whether or not it is a bitmap. And on top of that, the synthetic italics can apply to possibly either case. So the is_identity() check is necessary here.

lsalzman:
Gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=808e695ab3ca5b90ca4fa1cea767d4c133e5d329

glennw:
Thanks, try looks good, r=me when you're ready to merge this :)

webrender/src/platform/unix/font.rs
```diff
@@ -613,10 +663,17 @@ impl FontContext {
         }
 
         match format {
+            FT_Glyph_Format::FT_GLYPH_FORMAT_BITMAP => {
+                if font.flags.contains(FontInstanceFlags::SYNTHETIC_ITALICS) {
+                    let (skew_width, skew_left) = skew_bitmap(&mut final_buffer, actual_width, actual_height, left, top);
```
kvark: the function returns some stuff and mutates the other. It would be cleaner to just return all the results, as in:
```rust
let (new_buffer, skew_width, skew_left) = skew_bitmap(final_buffer, ...);
```

webrender/src/platform/unix/font.rs
```diff
@@ -53,7 +53,56 @@ unsafe impl Send for FontContext {}
 
 extern "C" {
     fn FT_GlyphSlot_Embolden(slot: FT_GlyphSlot);
-    fn FT_GlyphSlot_Oblique(slot: FT_GlyphSlot);
+}
+
+// Skew factor matching Gecko/FreeType.
+const OBLIQUE_SKEW_FACTOR: f32 = 0.2;
```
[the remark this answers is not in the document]
lsalzman: Yes, they are different in Gecko itself to match the different platforms, and we're trying to match Gecko.

webrender/src/platform/unix/font.rs
```diff
@@ -53,7 +53,56 @@ unsafe impl Send for FontContext {}
...
+
+fn get_skew_bounds(bottom: i32, top: i32) -> (f32, f32) {
+    let skew_min = ((bottom as f32 + 0.5) * OBLIQUE_SKEW_FACTOR).floor();
+    let skew_max = ((top as f32 - 0.5) * OBLIQUE_SKEW_FACTOR).ceil();
+    (skew_min, skew_max)
+}
+
+fn skew_bitmap(bitmap: &mut Vec<u8>, width: usize, height: usize, left: i32, top: i32) -> (usize, i32) {
+    let stride = width * 4;
```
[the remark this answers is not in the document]
lsalzman: This applies to the buffer after it has been converted to the format WR expects, which is always 4 bytes per pixel.

lsalzman:
Fixed as many review comments as I could without turning it into a giant mess.
@bors-servo r+

PR #2247, opened 2017-12-21
Port debugger to a vue + webpack + vuex + single file component setup.

glennw:
* Add a webpack build step, so templates are compiled during build.
  Although this is a slight performance improvement, the main win
  is maintainability.
* Start using VueX for state management (still needs to be tidied up).

This change switches from a hacky vuejs app to a slightly better
structured one, that will be easier to maintain and add new functionality
to in the future.

There's one functionality change as part of this. Previously, when
you disconnected from a WR instance, it was not possible to view the
last state of the render tasks, documents etc. Now, we retain that
state and leave it available for viewing after disconnect. Instead,
when disconnected, buttons for refreshing state and modifying debug
flags are disabled.

Next steps (when we have time):
 * Tidy up some of the duplicated code (e.g. all the tree view pages
   are mostly the same, and could be abstracted).
 * Implement a proper treeview - with controls to expand / collapse
   all nodes etc.
 * Add a richer treeview with delegate style template - this will
   allow us to build a custom display template for each type of
   treeview node, displaying the data in a much more readable way.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2247)


glennw:
r? @kvark or @nical 

This doesn't need super careful review - it just needs a cursory look, I think.

debugger/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "vue": "^2.5.11",
+    "vue-material-design-icons": "^0.8.2",
+    "vuex": "^3.0.1"
+  },
+  "browserslist": [
+    "> 1%",
+    "last 2 versions",
+    "not ie <= 8"
+  ],
+  "devDependencies": {
```
kvark: looks sooo similar to Cargo :)

kvark:
> The job exceeded the maximum time limit for jobs, and has been terminated.

@bors-servo retry

PR #2248, opened 2017-12-26
Document push_(radial_)gradient

### LINKED ISSUE #2222: Document and clarify `push_gradient` behavior
cc @eqrion

How exactly should [DisplayListBuilder::push_gradient](https://doc.servo.org/webrender_api/struct.DisplayListBuilder.html#method.push_gradient) behave?

The signature is:

```
pub fn push_gradient(
    &mut self, 
    info: &LayoutPrimitiveInfo, 
    gradient: Gradient, 
    tile_size: LayoutSize, 
    tile_spacing: LayoutSize
)
```

My understanding is is the following:

* param `info` contains information common to most "elements". Most notably a rectangle describing where on screen the "element" is located. This rectangle is filled with the gradient in this case. What does the `local_clip` attribute do?
* param `gradient`, well contains information about the actual gradient.
* param `tile_size` describes the area on which the gradient should be drawn. If its size differs from the rectangle in `info`multiple "tiles" are drawn next to each other.
* param `tile_spacing` spaces out the tiles specified in `tile_size`. Which CSS features require to set this value to something else then zero?

Firefox:
Firefox
Servo (via webrender):
Servo
(To see this servo rendering you need servo/servo#19554)
Source:
https://gist.github.com/pyfisch/339a41b88e3b794f88be1d9c818ad09e

The Servo rendering looks like the tile is only repeated to the right and bottom but not to the other sides. Is this intentional?
CSS supports repeating only in axis (`repeat-x`, `repeat-y`) or no repeating at all. How should this be expressed?
1. Restrict the rectangle passed inside `info` to the area where the gradient should actually be displayed.
2. Use `std::f64::MAX` or similar values in `tile_spacing` to hide other instances of the tile?

What is preferable performance and semantics wise?

pyfisch:
Closes #2222.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2248)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -1223,6 +1223,20 @@ impl DisplayListBuilder {
         self.push_item(item, info);
     }
 
+    /// Pushes a linear gradient to be displayed.
+    /// 
+    /// The gradient itself is described in the
+    /// `gradient` parameter. It is drawn on
+    /// a "tile" with the dimensions from `tile_size`.
+    /// These tiles are now repeated to the right and
+    /// to the bottom infinitly. If `tile_spacing`
+    /// is not zero spacers with the given dimensions
+    /// are inserted between the tiles as seams.
+    /// 
+    /// The origin of the tiles is given in `info.rect.origin`.
+    /// If the gradient should only be displayed once limit 
+    /// the `info.rect.size` to a single tile.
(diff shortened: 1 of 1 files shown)


pcwalton:
Looks good to me!

@bors-servo: r+

glennw:
@bors-servo retry
 * travis timeout

PR #2249, opened 2018-01-02
Recurse to children of empty ClipScrollNodes

mrobinson:
This is important in order to clean up stale data during scrolling and
fixes a crash in Gecko. There is no test for this, because wrench does
not have the ability to easily test scrolling animation.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2249)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -417,17 +417,18 @@ impl ClipScrollTree {
 
             node.push_gpu_node_data(gpu_node_data);
 
-            if node.children.is_empty() || node.combined_clip_outer_bounds.is_empty() {
-                return
+            if node.children.is_empty() {
+                return;
             }
+
 
             node.prepare_state_for_children(&mut state);
             node.children.clone()
         };
 
-        for child_layer_id in node_children {
+        for child_node_id in node_children {
             self.update_node(
-                child_layer_id,
+                child_node_id,
                 &mut state,
                 next_coordinate_system_id,
                 device_pixel_scale,


mrobinson:
There is no test for this in WebRender, but here is the Gecko try run with the talos test now passing: https://treeherder.mozilla.org/#/jobs?repo=try&revision=58500e8b9a12d65db9100463c3ad81fc99d17645

glennw:
@bors-servo r+

@mrobinson Could you add a GH issue with some (rough) details of what we would need to add to wrench to be able to automate tests for this?

webrender/src/clip_scroll_tree.rs
```diff
@@ -417,17 +417,18 @@ impl ClipScrollTree {
 
             node.push_gpu_node_data(gpu_node_data);
 
-            if node.children.is_empty() || node.combined_clip_outer_bounds.is_empty() {
```
kvark: so why can we not check for the outer bounds here?

mrobinson: We need to keep recursing even in this case in order to properly set the GPU node data for all child nodes (otherwise it might be stale). Additionally, once we support non-hierarchical clip chains, this optimization has to be removed or reworked.

kvark: Aren't child nodes going to be invisible anyway if the outer bounds are empty?

mrobinson: Yes. The issue here is when they were previously marked as visible, we need to now recurse to mark them as invisible. If we do not, the other bits of WebRender will believe they are still visible, but they won't have any corresponding GPU data. 

I think there's still a lot of adjustment we can do here. For instance, as I previously mentioned, we should further decouple the idea of a clip node and a clip chain in order to support complicated clipping situations. I'm working on this now. Additionally, the per-node GPU data is only holding transform information at the moment, so it makes sense that we only push this for nodes that adjust the transform (positioning nodes, but not clipping nodes).

These two things, along with allowing us to fix some Gecko issues, will also allow us to one day separate the clipping and scrolling/positioning trees completely.

PR #2251, opened 2018-01-03
Update serde to 1.0.27, removing some of the patch hacks

Gankro:
1.0.27 includes deserialize_from properly (now deserialize_in_place). serde_derive 1.0.27 also supports deriving it, but we still need to fork it to include my Awful transmute-based impl for enums. That fork is a much smaller patchset now, though.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2251)


Gankro:
cc @staktrace (in case you run into this after vacation)

Issue #2252, opened 2018-01-03
zoomed in blurred text shadows generate garbage bounding box fragments


Gankro:
Try zooming in and out on this page:

```
data:text/html,<meta charset="utf-8"><span style="text-shadow: 2px 2px 1px red"><div>Blur<span style="color: blue">%F0%9F%8E%89</span><span style="color: rgba(0,0,0,0.7)">%F0%9F%8E%89</span><span style="color: rgba(0,0,255,0.4)">%F0%9F%8E%89</span><span style="color: transparent">%F0%9F%8E%89</span><span style="text-shadow: 2px 2px 2px red"></span></div><div>Blur<span style="color: blue">A</span><span style="color: rgba(0,0,0,0.7)">A</span><span style="color: rgba(0,0,255,0.4)">A</span><span style="color: transparent">A</span></div></span>
```

![screen shot 2018-01-03 at 4 10 48 pm](https://user-images.githubusercontent.com/1136864/34540011-e92f1a8e-f0a0-11e7-8133-036958bfa805.png)

Gankro:
Can't repro on windows.

Gankro:
Note the example is extra complicated just to make it easier to see, and to rule out some things. This also demonstrates the issue:

```
<span style="color: transparent; text-shadow: 2px 2px 1px black">A</span>
```

Gankro:
In theory this makes `reftests/text/shadow-single.yaml` sufficient to demonstrate the issue but it seems like zoom is somehow relevant, and wrench can't do it.

Gankro:
Actually hmm I can see this on some unzoomed characters sometimes?

Gankro:
Or not? Seems like reset-zoom was just acting janky?

Gankro:
Ok I spent some more time on this, and I've reproduced it on linux, macos, and windows. I wasn't seeing it initially on windows or wrench because it appears to be based on global effects on the cache: you need to launch a few tabs to social media sites first and then load the example and it will work. I was able to reproduce this in a binary recording in wrench as well, although if you export any one frame as a yaml it won't reproduce the issue!

Gankro:
I have shown @kvark how to reproduce this locally, hopefully they can figure this out >.>

glennw:
@kvark Possibly a good use case for your new capturing WIP ? :)

kvark:
@glennw ^_^ exactly what I was telling @Gankro :)
[...]

PR #2254, opened 2018-01-03
Fixed edge mask decoding in the shader

kvark:
Related to #1797

Apparently, the old GLSL code that I had was all-or-nothing. For some reason, it didn't show the issue before, but it started showing it when the new brush splitting has arrived.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2254)


kvark:
There was already a reftest for this (rotated-clip) but it was so small in size that it went under the segment split threshold and rendered in one go, hence it didn't reveal the problem. This PR includes an updated reftest.

glennw:
Oh, CI is failing on TC too - might be a slight discrepancy in the reftest output.

Gankro:
Could this also be the cause/fix for https://github.com/servo/webrender/issues/2252 ?

kvark:
@glennw agreed, was also thinking about tests with different sizes.
@Gankro the fix affects AA on edges of brush primitives, mostly segments. Not sure if it's related to #2252

kvark:
Try looks green: https://treeherder.mozilla.org/#/jobs?repo=try&revision=68770f9175f0c8e963521ab0a775544564898f5c , if you don't mind 4 unexpected passes.

The new changes from `border.rs` are not new function but rather replicating the exact behavior of the old code for borders. Certainly something to put on the radar for future work on borders.

@bors-servo r=glennw

kvark:
old good OSX timeout...
@bors-servo retry

kvark:
ugh, all sorts of Travis failures... to check out the source, to wait for the completion, etc.

PR #2255, opened 2018-01-04
Make incompatible coordinate system detection more conservative.

glennw:
Only consider references frames with simple translations to be
compatible coordinate systems. This fixes some reftests in servo,
related to rotation transforms that do preserve 2d axis condition,
but cause problems (e.g. a 90 degree rotation).

In the future, we may be able to relax this and handle some more
complex cases, but these are edge cases anyway, so shouldn't have
a significant performance impact.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2255)


glennw:
r? @kvark and/or @mrobinson 

This fixes https://github.com/servo/servo/pull/19614#issuecomment-354899561.

Try run looks green (2 new PASSES, and 1 unrelated intermittent):
https://hg.mozilla.org/try/rev/32df493987d13fe4b3adea5a33404dbb886d3e80
https://treeherder.mozilla.org/#/jobs?repo=try&revision=32df493987d13fe4b3adea5a33404dbb886d3e80

mrobinson:
This seems like a reasonable change to me, but both the test added here and transform_3d_from_outside_viewport.html have a 3d component. Maybe we can make this test less conservative? For instance, perhaps we could create a new coordinate system when the transform has Z translation?

glennw:
@mrobinson I *think* in this case it relies on there being perspective present, since just having Z-translation without perspective won't affect the result.

webrender/src/util.rs
```diff
@@ -99,10 +99,10 @@ impl<Src, Dst> MatrixHelpers<Src, Dst> for TypedTransform3D<f32, Src, Dst> {
         if self.m11 != 1. || self.m22 != 1. || self.m33 != 1. {
             return false;
         }
-        self.m12.abs() < NEARLY_ZERO || self.m13.abs() < NEARLY_ZERO ||
```
kvark: O_O how did this ever work

webrender/src/util.rs
```diff
@@ -99,10 +99,10 @@ impl<Src, Dst> MatrixHelpers<Src, Dst> for TypedTransform3D<f32, Src, Dst> {
...
-            self.m14.abs() < NEARLY_ZERO || self.m21.abs() < NEARLY_ZERO ||
-            self.m23.abs() < NEARLY_ZERO || self.m24.abs() < NEARLY_ZERO ||
-            self.m31.abs() < NEARLY_ZERO || self.m32.abs() < NEARLY_ZERO ||
+        self.m12.abs() < NEARLY_ZERO && self.m13.abs() < NEARLY_ZERO &&
+            self.m14.abs() < NEARLY_ZERO && self.m21.abs() < NEARLY_ZERO &&
+            self.m23.abs() < NEARLY_ZERO && self.m24.abs() < NEARLY_ZERO &&
+            self.m31.abs() < NEARLY_ZERO && self.m32.abs() < NEARLY_ZERO &&
             self.m34.abs() < NEARLY_ZERO
```
kvark: should probably check for `self.m44` as well

PR #2256, opened 2018-01-04
Move batch related code into a separate source file.

glennw:
This removes a large amount of code from tiling.rs. The other
parts of tiling.rs can be moved to more appropriate files as
a follow up.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2256)


glennw:
r? anyone - there's no functional changes here, just moving the batch related code into a separate source file.

kvark:
So, you are aiming to kill `tiling.rs` just as you start doing the actual tiling for WM-specific compositor integration? :D

PR #2257, opened 2018-01-04
Remove legacy shader part from build_shader_strings

zakorgy:
Since legacy shaders are ported to .glsl files, this part is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2257)


nical:
Looks good. @bors-servo r+

PR #2258, opened 2018-01-04
Remove the concept of clip boundaries from ClipScrollNode

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2258)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -169,21 +169,18 @@ impl ClipSources {
         let mut local_outer = Some(LayerRect::max_rect());
         let mut local_inner = local_outer;
         let mut can_calculate_inner_rect = true;
-        let mut can_calculate_outer_rect = true;
+        let mut can_calculate_outer_rect = false;
         for source in clips {
             match *source {
                 ClipSource::Image(ref mask) => {
                     if !mask.repeat {
+                        can_calculate_outer_rect = true;
                         local_outer = local_outer.and_then(|r| r.intersection(&mask.rect));
-                        can_calculate_inner_rect = false;
-                    } else {
-                        can_calculate_inner_rect = false;
-                        can_calculate_outer_rect = false;
-                        break;
                     }
                     local_inner = None;
                 }
                 ClipSource::Rectangle(rect) => {
+                    can_calculate_outer_rect = true;
                     local_outer = local_outer.and_then(|r| r.intersection(&rect));
                     local_inner = local_inner.and_then(|r| r.intersection(&rect));
                 }
@@ -192,10 +189,10 @@ impl ClipSources {
                     // case clip mask size, for now.
                     if mode == ClipMode::ClipOut {
                         can_calculate_inner_rect = false;
-                        can_calculate_outer_rect = false;
                         break;
                     }
 
+                    can_calculate_outer_rect = true;
                     local_outer = local_outer.and_then(|r| r.intersection(rect));
 
                     let inner_rect = extract_inner_rect_safe(rect, radius);
@@ -204,19 +201,18 @@ impl ClipSources {
                 }
                 ClipSource::BorderCorner { .. } => {
                     can_calculate_inner_rect = false;
-                    can_calculate_outer_rect = false;
                     break;
                 }
             }
         }
 
         let outer = match can_calculate_outer_rect {
-            true => local_outer,
+            true => Some(local_outer.unwrap_or_else(LayerRect::zero)),
             false => None,
         };
 
         let inner = match can_calculate_inner_rect {
-            true => local_inner.unwrap_or(LayerRect::zero()),
+            true => local_inner.unwrap_or_else(LayerRect::zero),
             false => LayerRect::zero(),
         };
 
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -17,7 +17,7 @@ use resource_cache::ResourceCache;
 use scene::SceneProperties;
 use spring::{DAMPING, STIFFNESS, Spring};
 use std::rc::Rc;
-use util::{MatrixHelpers, MaxRect, TransformOrOffset, TransformedRectKind};
+use util::{MatrixHelpers, TransformOrOffset, TransformedRectKind};
 
 #[cfg(target_os = "macos")]
 const CAN_OVERSCROLL: bool = true;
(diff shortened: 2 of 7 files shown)


mrobinson:
Here's the Gecko try for this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=71880bdd30ef9fbafde32abe5fa0b829c2a5aa46&selectedJob=154093851

The Windows build failure looks unrelated.

webrender/src/clip.rs
```diff
@@ -169,21 +169,18 @@ impl ClipSources {
         let mut local_outer = Some(LayerRect::max_rect());
         let mut local_inner = local_outer;
         let mut can_calculate_inner_rect = true;
-        let mut can_calculate_outer_rect = true;
+        let mut can_calculate_outer_rect = false;
```
kvark: Shouldn't we also remove the line touching this variable from `ClipSource::BorderCorner { .. } => {}` arm? Otherwise, it looks inconsistent, e.g.:
  - we start with `false`
  - we meet a rect, flip to `true`
  - we meet a border corner, set it to `false`
  - we meet something else, flip it back to `true`

webrender/src/clip_scroll_node.rs
```diff
@@ -377,20 +372,24 @@ impl ClipScrollNode {
...
-                }
-                outer_rect.intersection(&state.combined_outer_clip_bounds)
-                    .unwrap_or_else(DeviceIntRect::zero)
-            }
-            None => state.combined_outer_clip_bounds,
-        };
+        // All clipping ClipScrollNodes should have outer rectangles, because they never
+        // use the BorderCorner clip type and they always have at last one non-ClipOut
+        // Rectangle ClipSource.
+        let screen_outer_rect = screen_outer_rect.expect("Clipping node didn't have outer rect.");
```
kvark: out of curiosity, is this a recommended use of shadowing local variable names, or just your preference?

webrender/src/clip.rs
mrobinson: BorderCorner is currently always the last type of clip added. I think that we can even go a bit further here at some point and allow outer rects with border corners, but perhaps it's better to make this change more cautiously?

webrender/src/clip_scroll_node.rs
mrobinson: I don't know of any official recommendation around variable shadowing, but I know that it isn't totally uncommon for unwrapping Option<...> types. I felt okay using it here for that reason.

mrobinson:
I'll update this PR with the even less conservative change to clips.rs that we have discussed as well as adding another related commit.

mrobinson:
Once the Gecko try job has advanced sufficiently, I'll update this PR.

webrender/src/clip.rs
kvark: It's just inconsistent with the rest of the logic. The core of this local change is go though "outer rect can be computed unless there is a clip that's not compatible" to "outer rect can't be computed unless there is at least one clip that has it". This is a perfectly reasonable thing to change. But that assignment of "can_calculate_outer_rect = false" in the loop stands away from the core logic.

mrobinson:
Whoops. I realized I hadn't posted the Gecko try results for the new version. They are here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=0e5922ab95fe6fc25657609ec92c75bf283802b5

I've added one more related commit and adjusted the PR message as well.

webrender/src/clip.rs
mrobinson: I agree that it's a little bit odd, so I've made the check even less conservative and it looks like things are still working. The latest version of the PR should have this change.

glennw:
Reviewed 2 of 5 files at r1, 1 of 2 files at r2, 4 of 4 files at r3.

---

*[webrender/src/clip.rs, line 172 at r1](https://reviewable.io:443/reviews/servo/webrender/2258#-L20JQyCRf0_MfiFQ2RO-r1-172:-L2HaeqhwDtEr5f5x7u_:bn6qs16) ([raw file](https://github.com/servo/webrender/blob/43c6a902f230751355f7898eda56e4c783a80b33/webrender/src/clip.rs#L172)):*

I agree that it's a little bit odd, so I've made the check even less conservative and it looks like things are still working. The latest version of the PR should have this change.
</blockquote></details>

It does seem like this logic relies on an implementation detail (the order of processing the clip sources)? It seems like it would be simpler to have the initial value be true, and then the BorderCorner case sets it to false and all the other branches don't modify the bool at all?

---

*[webrender/src/clip_scroll_node.rs, line 378 at r1](https://reviewable.io:443/reviews/servo/webrender/2258#-L20JQyCRf0_MfiFQ2RP-r1-378:-L2Hb3FPMJ-HUAoWk-5r:b-7dwy5i) ([raw file](https://github.com/servo/webrender/blob/43c6a902f230751355f7898eda56e4c783a80b33/webrender/src/clip_scroll_node.rs#L378)):*

I don't know of any official recommendation around variable shadowing, but I know that it isn't totally uncommon for unwrapping Option<...> types. I felt okay using it here for that reason.
</blockquote></details>

I use the same style, fwiw - not sure where it came from, but I've seen it used in rust in several projects.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2258)*

mrobinson:
Reviewable seems to be having some issues right now so I'm replying on Github.

> It does seem like this logic relies on an implementation detail (the order of processing the clip sources)? It seems like it would be simpler to have the initial value be true, and then the BorderCorner case sets it to false and all the other branches don't modify the bool at all?

My previous comments about this were really confusing, so I'll try to clarify here.

We want to calculate an outer rectangle for the ClipSource as long as there is at least one RoundedRectangle, Rectangle, or non-repeating Image clip. This is a less conservative condition than we had before where we immediately gave up on outer rect calculation when encountering a BorderCorner ClipSource.

The first version of this PR depended on the order of adding the ClipSources to function properly, because it sometimes set `can_calculate_outer_rect` to true and sometimes to false. Since the current version only ever flips `can_calculate_outer_rect` to true (it's initialized to false), the current version shouldn't depend on the ClipSource order at all. 

I cannot really think of a clean way to do this check without first setting `can_calculate_outer_rect` to false, because I would need another variable to test whether there was at least one outer-rect-defining clip. Given that, is the current approach okay or should I try to look for something else here?

mrobinson:
@bors-servo r=glennw

Glenn confirmed via IRC that the current version of the PR is okay to land.

PR #2259, opened 2018-01-05
Restructure parts of batch.rs, port line shader to brush.

glennw:
This patch has no functional changes - it is necessary preparation
work for some upcoming changes (handling text/box shadow pictures
as part of the normal batching process). It's also a small improvement
on the existing code structure itself.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2259)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -8,11 +8,9 @@ void brush_vs(
     int prim_address,
     vec2 local_pos,
     RectWithSize local_rect,
-    ivec2 user_data
+    ivec2 user_data,
+    PictureTask pic_task
 );
-
-#define RASTERIZATION_MODE_LOCAL_SPACE      0.0
-#define RASTERIZATION_MODE_SCREEN_SPACE     1.0
 
 #define VECS_PER_BRUSH_PRIM                 2
 #define VECS_PER_SEGMENT                    2
@@ -84,8 +82,9 @@ void main(void) {
 
     // Fetch the dynamic picture that we are drawing on.
     PictureTask pic_task = fetch_picture_task(brush.picture_address);
+    ClipArea clip_area = fetch_clip_area(brush.clip_address);
 
-    if (pic_task.rasterization_mode == RASTERIZATION_MODE_LOCAL_SPACE) {
+    if (pic_task.pic_kind_and_raster_mode > 0.0) {
         local_pos = local_segment_rect.p0 + aPosition.xy * local_segment_rect.size;
 
         // Right now - pictures only support local positions. In the future, this
@@ -93,12 +92,18 @@ void main(void) {
         device_pos = pic_task.common_data.task_rect.p0 +
                      uDevicePixelRatio * (local_pos - pic_task.content_origin);
 
+#ifdef WR_FEATURE_ALPHA_PASS
+        write_clip(
+            vec2(0.0),
+            clip_area
+        );
+#endif
+
         // Write the final position transformed by the orthographic device-pixel projection.
         gl_Position = uTransform * vec4(device_pos, 0.0, 1.0);
     } else {
         VertexInfo vi;
         ClipScrollNode scroll_node = fetch_clip_scroll_node(brush.scroll_node_id);
-        ClipArea clip_area = fetch_clip_area(brush.clip_address);
 
         // Write the normal vertex information out.
         if (scroll_node.is_axis_aligned) {
@@ -153,7 +158,8 @@ void main(void) {
         brush.prim_address + VECS_PER_BRUSH_PRIM,
         local_pos,
         brush_prim.local_rect,
-        brush.user_data
+        brush.user_data,
+        pic_task
     );
 }
 #endif
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -29,7 +29,8 @@ void brush_vs(
     int prim_address,
     vec2 local_pos,
     RectWithSize local_rect,
-    ivec2 user_data
+    ivec2 user_data,
+    PictureTask pic_task
 ) {
     // TODO(gw): For now, this brush_image shader is only
     //           being used to draw items from the intermediate
--- a/webrender/res/brush_line.glsl
+++ b/webrender/res/brush_line.glsl
@@ -2,17 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#include shared,prim_shared
+#define VECS_PER_SPECIFIC_BRUSH 2
 
-varying vec4 vColor;
+#include shared,prim_shared,brush
+
+varying vec2 vLocalPos;
+
+flat varying vec4 vColor;
 flat varying int vStyle;
 flat varying float vAxisSelect;
 flat varying vec4 vParams;
 flat varying vec2 vLocalOrigin;
 
-varying vec2 vLocalPos;
-
 #ifdef WR_VERTEX_SHADER
+
 #define LINE_ORIENTATION_VERTICAL       0
 #define LINE_ORIENTATION_HORIZONTAL     1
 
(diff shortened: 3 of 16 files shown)


glennw:
r? @kvark 

I've split this re-structuring patch (which doesn't need super careful review, it's just shuffling code around) from the work I am doing on top of it, to hopefully make the review process a bit simpler.

If I get the follow up work finished today, I'll add them as separate commits on top of this :)

glennw:
There's enough changes in the follow up commits that I should do a gecko try run now, before we merge this. Will kick one off shortly.

glennw:
The follow up commits now include the work to port the line decoration shader to be a brush type.

Try run looks good so far:
https://hg.mozilla.org/try/rev/5773449cd3bf741e68ca31a01d97a8a7416db76e
https://treeherder.mozilla.org/#/jobs?repo=try&revision=5773449cd3bf741e68ca31a01d97a8a7416db76e

The first commit has the majority of the line diffs, and has no functional changes. Only the last 3 commits need proper review. If it's easier to review as separate PRs, let me know and I can split them up easily.

kvark:
Reviewed 1 of 3 files at r2, 4 of 4 files at r3.
Review status: 2 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/batch.rs, line 516 at r3](https://reviewable.io:443/reviews/servo/webrender/2259#-L26dbz0LuihB33T-h5R:-L26dbz1ZpOrV7o59TD9:b-8uch64) ([raw file](https://github.com/servo/webrender/blob/dd59e32f367481bf639c787c795895f3918cbbaf/webrender/src/batch.rs#L516)):*
> ```Rust
>             // picture types.
>             // TODO(gw): Support culling on shadow image types.
>             if pic_type != PictureType::Image || metadata.screen_rect.is_some() {
> ```

the exclusion of `Image` type appears to be a functional change. Simply ignoring it doesn't seem correct. Should we instead turn it into an `assert!` or spew an `error!` at the beginning of the method?

---

*[webrender/src/batch.rs, line 773 at r3](https://reviewable.io:443/reviews/servo/webrender/2259#-L26j9dKYfTVMIRZP_j7:-L26j9dLVzzYT-yq2-GM:b-3xuwv3) ([raw file](https://github.com/servo/webrender/blob/dd59e32f367481bf639c787c795895f3918cbbaf/webrender/src/batch.rs#L773)):*
> ```Rust
>                 let is_shadow = pic_type == PictureType::TextShadow;
> 
>                 let font_transform = if is_shadow {
> ```

is this another functional change? I'd expect this transform be controlled by the `ContentOrigin` enum (`Screen`/`Local`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2259)*

glennw:
Review status: 2 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/batch.rs, line 516 at r3](https://reviewable.io:443/reviews/servo/webrender/2259#-L26dbz0LuihB33T-h5R:-L2761YrbYNtC8ILKwKS:b640qmj) ([raw file](https://github.com/servo/webrender/blob/dd59e32f367481bf639c787c795895f3918cbbaf/webrender/src/batch.rs#L516)):*

the exclusion of `Image` type appears to be a functional change. Simply ignoring it doesn't seem correct. Should we instead turn it into an `assert!` or spew an `error!` at the beginning of the method?
</blockquote></details>

Since there was previously a separate code path for non-Image pictures, they didn't execute any culling step at all. So adding this condition here makes that the same as previous (even though it's a bit inelegant for now).

---

*[webrender/src/batch.rs, line 773 at r3](https://reviewable.io:443/reviews/servo/webrender/2259#-L26j9dKYfTVMIRZP_j7:-L276DrITYed_uUfqFGm:bwmguv2) ([raw file](https://github.com/servo/webrender/blob/dd59e32f367481bf639c787c795895f3918cbbaf/webrender/src/batch.rs#L773)):*

is this another functional change? I'd expect this transform be controlled by the `ContentOrigin` enum (`Screen`/`Local`).
</blockquote></details>

It probably makes sense for it to be controlled by that in the future (once the content origin is actually configurable), but it should match the existing behaviour. In the previous code, the alternative code path selection was based on the type of the picture, so I've matched that here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2259)*

kvark:
Review status: 2 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[webrender/src/batch.rs, line 773 at r3](https://reviewable.io:443/reviews/servo/webrender/2259#-L26j9dKYfTVMIRZP_j7:-L27IO1piRTWeGd0IVXT:b-zib8fc) ([raw file](https://github.com/servo/webrender/blob/dd59e32f367481bf639c787c795895f3918cbbaf/webrender/src/batch.rs#L773)):*

It probably makes sense for it to be controlled by that in the future (once the content origin is actually configurable), but it should match the existing behaviour. In the previous code, the alternative code path selection was based on the type of the picture, so I've matched that here.
</blockquote></details>

Understood. Perhaps, adding a comment here would help our future efforts ;)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2259)*

kvark:
Reviewed 14 of 14 files at r4.
Review status: 15 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/res/prim_shared.glsl, line 246 at r4](https://reviewable.io:443/reviews/servo/webrender/2259#-L27PDmSoSEZGuVoxchc:-L27PDmTloDJoBEXEZhS:b-6qazmu) ([raw file](https://github.com/servo/webrender/blob/ae7f8bb01392a3e59a73898b92df74ee316eadbc/webrender/res/prim_shared.glsl#L246)):*
> ```GLSL
>     RenderTaskCommonData common_data;
>     vec2 content_origin;
>     float pic_kind_and_raster_mode;
> ```

I think it would be more scalable to have `uint flags` here, extracted from the input. The current sign + positive integer scheme is a bit too specific.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2259)*

kvark:
Reviewed 1 of 1 files at r5.

---

*[webrender/src/renderer.rs, line 1615 at r3](https://reviewable.io:443/reviews/servo/webrender/2259#-L27Pxoigy1sMVQGmxtF:-L27Pxoigy1sMVQGmxtG:b5w5fco) ([raw file](https://github.com/servo/webrender/blob/dd59e32f367481bf639c787c795895f3918cbbaf/webrender/src/renderer.rs#L1615)):*
> ```Rust
>     ps_angle_gradient: PrimitiveShader,
>     ps_radial_gradient: PrimitiveShader,
>     ps_line: PrimitiveShader,
> ```

nice :) one up, two down

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2259)*

glennw:
Review status: 15 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/res/prim_shared.glsl, line 246 at r4](https://reviewable.io:443/reviews/servo/webrender/2259#-L27PDmSoSEZGuVoxchc:-L2HctIIfslFMi6BERGK:b-vc1q9e) ([raw file](https://github.com/servo/webrender/blob/ae7f8bb01392a3e59a73898b92df74ee316eadbc/webrender/res/prim_shared.glsl#L246)):*

I think it would be more scalable to have `uint flags` here, extracted from the input. The current sign + positive integer scheme is a bit too specific.
</blockquote></details>

The render task data is a float texture - so do you mean transmute a u32 in that data to an f32, and then use a shader instruction to reinterpret cast from f32 -> u32, or is there an easier way you can think of to do this?

---

*[webrender/src/batch.rs, line 773 at r3](https://reviewable.io:443/reviews/servo/webrender/2259#-L26j9dKYfTVMIRZP_j7:-L2Hd9DM3CUVc-awd9i4:b-896fix) ([raw file](https://github.com/servo/webrender/blob/dd59e32f367481bf639c787c795895f3918cbbaf/webrender/src/batch.rs#L773)):*

Understood. Perhaps, adding a comment here would help our future efforts ;)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2259)*

glennw:
@kvark Thanks for the review! Added a comment to the code, and a question on what you meant by the uint flags.

kvark:
Review status: 15 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/res/prim_shared.glsl, line 246 at r4](https://reviewable.io:443/reviews/servo/webrender/2259#-L27PDmSoSEZGuVoxchc:-L2IaimbdFt3r-Voqs5a:bqae1gv) ([raw file](https://github.com/servo/webrender/blob/ae7f8bb01392a3e59a73898b92df74ee316eadbc/webrender/res/prim_shared.glsl#L246)):*

The render task data is a float texture - so do you mean transmute a u32 in that data to an f32, and then use a shader instruction to reinterpret cast from f32 -> u32, or is there an easier way you can think of to do this?
</blockquote></details>

Even if you just static-cast int to float and back (without re-interpretation), there are 23 mantissa bits available - should be enough for now :)

i.e. Rust side does: `flags as f32`, shader side does `int(flags)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2259)*

kvark:
@glennw please feel free to merge with r=me if you don't feel like playing with flags - it's a minor concern.

glennw:
@kvark Thanks - I have left the flags as is for now to avoid a new try run, but I added a ```TODO``` comment in the code saying we should switch this to ```uint``` flags in the future.

@bors-servo r=kvark

PR #2260, opened 2018-01-05
Remove RG8 and RGB8 support from texture cache.

glennw:
RGB8 support is removed completely from WR. It's not supported
by many modern GPUs as a native format. Instead, expect the client
code to convert it to RGBx first.

RG8 support is removed from the texture cache, but remains as a
supported image format. It's used by some variants of YUV formats.
This allows RG8 images to be used as external images, but not
placed into the texture cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2260)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -458,7 +458,6 @@ impl Texture {
     pub fn get_bpp(&self) -> u32 {
         match self.format {
             ImageFormat::A8 => 1,
-            ImageFormat::RGB8 => 3,
             ImageFormat::BGRA8 => 4,
             ImageFormat::RG8 => 2,
             ImageFormat::RGBAF32 => 16,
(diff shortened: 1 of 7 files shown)


glennw:
r? @kvark or @nical ?

It'd be good for @jrmuizel to check over this too - make sure that my assumptions in the commit message are reasonable.

jrmuizel:
Should be fine.

kvark:
Thanks for chiming in, @jrmuizel !
@bors-servo r+

kvark:
Looks like this broke `yuv` example from running.

glennw:
@kvark Ah, didn't realize that the sample uses CPU-side textures. I think the best option will be to update the sample to use external textures (for RG8) or use a different YUV format that doesn't rely on RG8. Gecko only uses YUV textures via the external image interface.

Issue #2182, opened 2017-12-06
Inset box-shadows render incorrectly in some cases


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1423696

[Testcase](https://bug1423696.bmoattachments.org/attachment.cgi?id=8935091)

Actual vs expected results:
<img width="490" alt="screen shot 2017-12-06 at 2 34 15 pm" src="https://user-images.githubusercontent.com/961291/33681744-7bc3bef0-da93-11e7-8188-72d194b0a2ba.png">

It looks like the element's shape also clips the pre-blur shadow shape, instead of just clipping the blurred result.

```yaml
---
root:
  items:
    - type: stacking-context
      bounds: [0, 0, 1000, 1000]
      items:
        - type: box-shadow
          bounds: [ 22, 22, 200, 80 ]
          color: blue
          clip-mode: inset
          offset: -2 -2
          blur-radius: 20
          spread-radius: 0
          border-radius:
            top-left: [0, 0]
            top-right: [0, 0]
            bottom-left: [0, 0]
            bottom-right: [0, 0]
```

glennw:
This is resolved fixed in bugzilla.

PR #2261, opened 2018-01-05
Fix inflate size for inset box shadow.

demo99:
The inflate size should be multiplied by 2 when downscaling to guarantee the 1 pixel extra edge.  This should fix issue #2182 and [bug 1423696](https://bugzilla.mozilla.org/show_bug.cgi?id=1423696).

gecko try result: https://treeherder.mozilla.org/#/jobs?repo=try&revision=aac72b3488afd74fce9afcd4edf7ce17763bd57e&selectedJob=154326187

There is an unexpected-pass test. I will add a patch to bug 1423696.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2261)


glennw:
This causes a CI failure in the wrench reftests - I haven't looked to check if it looks like a bug or just a slight reference image update is required.

demo99:
The wrench test is about box shadow and the max difference is one. I think I should update the image as well.

glennw:
Thanks!

@bors-servo r+

PR #2265, opened 2018-01-07
Update reftest images for #2262.

glennw:
It's not clear why, but these reftests are failing for me on master
on my local machine. This could either be (a) my machine generating
different images due to a different freetype or (b) something weird
on CI where a test is failing and still being merged.

Opening this as a PR to see what CI makes of it...

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2265)


glennw:
Hmmm, these reference images do fail on CI, but pass on my local machine. I could add fuzziness - but the maximum pixel difference on my machine with the old images is ~65 over ~4000 pixels, which is not ideal. Any ideas what may cause the discrepancy @lsalzman or @jrmuizel ? My local machine is Ubuntu 17.10 running vanilla Gnome 3.

lsalzman:
Could be differences in Mesa or FreeType between your machine and the testers?

lsalzman:
FreeType 2.8, at least, uses the new Harmony LCD filtering algorithm, and ignores any explicit algorithm selection. So if you are somehow using that in your headless run instead of FreeType 2.6, that's one obvious possibility that would automatically cause divergence.

Regenerating subpixel-skew.png and subpixel-rotate.png works fine for me locally, which could be explained by that. What is weird, however, is that alpha-transform.png somehow fails for me locally if I regenerate it.

lsalzman:
Oh, alpha-transform.png works fine too when I regenerate it now. I just forgot to regen with --no-subpixel-aa. 

Other than that, the fact that alpha-transform.png is messed up for you, suggests that it probably isn't LCD filtering that is the issue. Looking in the reftest analyzer, you can see quite clear differences in the hinting/dilation, so for some strange you're getting a different hinting algorithm than I am, and than whatever FreeType is on CI.

This would suggest that FreeType changed how it handles FT_LOAD_NO_HINTING in newer versions somehow.

lsalzman:
Release notes for 2.8 suggest both fixes to the autohinter and the handling of FT_LOAD_NO_HINTING: https://sourceforge.net/projects/freetype/files/freetype2/2.8/

glennw:
Closing this since it breaks on CI. I'm not sure why my local setup is producing different results yet. It appears to be using the bundled FreeType 2.6.5.

PR #2266, opened 2018-01-07
Support page zoom in wrench, and avoid spinning CPU at 100%.

glennw:
This removes the old performance timing code (which is now unused),
and switches the event loop to (a) block until there are events
present and (b) consume all events that have been queued and are
available before drawing a frame.

Before this change, the event queue was getting swamped with
Awaken messages and not processing any scroll events (which our
glutin version only seems to provide if you call poll_events).

This will allow us to add support for zoom to wrench, which we
need for testing a variety of bugs that only occur under certain
zoom and DPI conditions. A side benefit is that wrench no longer
keeps the CPU active 100% of the time.

I've tested that binary replays still work, and that the reftests
still work with this change.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2266)

### DIFF
--- a/wrench/src/main.rs
+++ b/wrench/src/main.rs
@@ -481,6 +481,10 @@ fn main() {
                             DebugFlags::GPU_TIME_QUERIES | DebugFlags::GPU_SAMPLE_QUERIES
                         );
                     }
+                    VirtualKeyCode::R => {
+                        wrench.set_page_zoom(ZoomFactor::new(1.0));
+                        do_frame = true;
+                    }
                     VirtualKeyCode::M => {
                         wrench.api.notify_memory_pressure();
                     }
@@ -502,6 +506,20 @@ fn main() {
                         let file_name = format!("profile-{}.json", cpu_profile_index);
                         wrench.renderer.save_cpu_profile(&file_name);
                         cpu_profile_index += 1;
+                    }
+                    VirtualKeyCode::Up => {
+                        let current_zoom = wrench.get_page_zoom();
+                        let new_zoom_factor = ZoomFactor::new(current_zoom.get() + 0.1);
+
+                        wrench.set_page_zoom(new_zoom_factor);
+                        do_frame = true;
+                    }
+                    VirtualKeyCode::Down => {
+                        let current_zoom = wrench.get_page_zoom();
+                        let new_zoom_factor = ZoomFactor::new((current_zoom.get() - 0.1).max(0.1));
+
+                        wrench.set_page_zoom(new_zoom_factor);
+                        do_frame = true;
                     }
                     _ => (),
                 }
--- a/wrench/src/main.rs
+++ b/wrench/src/main.rs
@@ -62,7 +62,6 @@ use perf::PerfHarness;
 use png::save_flipped;
 use rawtest::RawtestHarness;
 use reftest::{ReftestHarness, ReftestOptions};
-use std::cmp::{max, min};
 #[cfg(feature = "headless")]
 use std::ffi::CString;
 #[cfg(feature = "headless")]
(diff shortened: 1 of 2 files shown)


wrench/src/main.rs
```diff
@@ -442,211 +425,133 @@ fn main() {
...
-                    wrench.show_onscreen_help();
+        for event in events {
+            match event {
+                glutin::Event::Closed => {
+                    break 'outer;
                 }
 
-                wrench.render();
-                window.swap_buffers();
+                glutin::Event::MouseWheel(delta, _, _) => {
```
Gankro: Since this is binary and non-positional, it seems like we should just map to +/- on the keyboard? (@kvark's suggestion)

wrench/src/main.rs
```diff
@@ -442,211 +425,133 @@ fn main() {
...
+                        0.9
+                    };
 
-                min_time = min(min_time, dur);
-                min_min_time = min(min_min_time, dur);
-                max_time = max(max_time, dur);
-                max_max_time = max(max_max_time, dur);
-                sum_time = sum_time + dur;
+                    let current_zoom = wrench.get_page_zoom();
+                    let new_zoom_factor = ZoomFactor::new(current_zoom.get() * delta);
```
Gankro: [thinking out loud] this is a bit weird since (zoom in + zoom out) isn't an identity function? Should we instead just do addition, with a min/max zoom level?

wrench/src/wrench.rs
```diff
@@ -240,6 +242,16 @@ impl Wrench {
...
     }
 
+    pub fn get_page_zoom(&self) -> ZoomFactor {
+        self.page_zoom_factor
+    }
+
+    pub fn set_page_zoom(&mut self, zoom_factor: ZoomFactor) {
+        self.page_zoom_factor = zoom_factor;
+        self.api.set_page_zoom(self.document_id, self.page_zoom_factor);
+        self.set_title("");
```
Gankro: Does setting "extra" to the empty string clobber anything? Or is it basically dead code?

wrench/src/main.rs
```diff
@@ -442,211 +425,133 @@ fn main() {
...
-        wrench.update(dim);
-
-        let frame_num = thing.do_frame(&mut wrench);
-        unsafe {
-            CURRENT_FRAME_NUMBER = frame_num;
-        }
-
-        wrench.render();
-        window.swap_buffers();
-    }
```
Gankro: Can you explain why this was a loop, and now it's not?

glennw: I *think* that when being used for timing / profiling purposes, the idea here was to queue a set of initial frames to avoid initial timing spikes.

wrench/src/main.rs
glennw: I couldn't get the +/- key events to work in our (old fork) of glutin. I think it's probably a bug in our old glutin fork. I could bind it to some other keys if you prefer (cursor up/down)?

wrench/src/main.rs
glennw: Fixed

wrench/src/wrench.rs
glennw: The extra string is only used during binary replay, so although it's not a very tidy API, it shouldn't have any bad side-effects.

wrench/src/main.rs
Gankro: Sure that seems fine.

wrench/src/main.rs
```diff
@@ -457,6 +457,25 @@ fn main() {
...
+                    };
+
+                    let delta = if dy > 0.0 {
+                        0.1
+                    } else {
+                        -0.1
+                    };
+
+                    let current_zoom = wrench.get_page_zoom();
+                    let new_zoom_factor = ZoomFactor::new(current_zoom.get() + delta);
```
Gankro: You should clamp this to 0.1 somewhere

PR #2267, opened 2018-01-08
Comment out warning that is flooding console output

staktrace:
Temporary fix until we figure out what to do in https://github.com/servo/webrender/pull/2203

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2267)


nical:
This warning gets in the way of getting anything done with logs. I'll file an issue to investigate it further when we have time.

@bors-servo r+

PR #2270, opened 2018-01-08
add reftests for synthetic bold

Gankro:
* synthetic bold does a thing
* transparent text is smeared properly (needs to be a tiny bit fuzzy)
* transparent synthetic bold is actually transparent (sanity check)

All pass locally, might need to tweak fuzzyness based on CI results.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2270)


Gankro:
I could also add png reftests; lmk if I should.

jrmuizel:
I think it's fine without the png images (It's nice to not have to update the PNGs if we update the OS). You could always just do the synthetic bold yourself by having another copy of the text shifted slightly and then using a generous amount of fuzz.

Gankro:
linux CI way fuzzier, bumping to `(1,229)`

PR #2271, opened 2018-01-09
Various small steps towards drawing box-shadows as segments.

glennw:
* Add missing items to render pass debugger
* Add a benchmark with many box shadows, based on getpocket.cdn.mozilla.net
* Move code to add a segmented brush to separate function.

This will be needed once we start drawing some pictures (e.g.
box shadows) as segmented primitives.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2271)


glennw:
Shouldn't be any functional changes as a result of this, but it's probably easier to review this now before I add the changes on top to draw box shadows as segments.

r? @kvark or anyone really.

PR #2276, opened 2018-01-10
Switch to using R8 on all GL implementations.

### LINKED ISSUE #788: device.rs code conditional on ARM should probably not be

### LINKED ISSUE #2134: Use R8 format instead of A8
This is a child task of #2110

`A8` is outdated, and Angle in particular doesn't like it (invoking a slow path on uploads).
We should be using `R8` instead and be future-proof.

Note: this doesn't show up in profiles yet, so not high priority.

glennw:
Fixes #2134.
Fixes #788.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2276)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -212,7 +212,7 @@ impl Example for App {
         let image_mask_key = api.generate_image_key();
         resources.add_image(
             image_mask_key,
-            ImageDescriptor::new(2, 2, ImageFormat::A8, true),
+            ImageDescriptor::new(2, 2, ImageFormat::R8, true),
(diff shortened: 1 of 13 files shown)


glennw:
OK, I got the basic WR example running on a Nexus 9 - the sample items show up and the debug profiler displays correctly (which relies on an A8 debug font texture).

@bors-servo r=kvark

glennw:
@bors-servo retry
 (should be queued after the travis wait PR)

PR #2277, opened 2018-01-10
Update clap to a version that takes up less space

jrmuizel:
See https://clap.rs/2018/01/09/new-years-weight-loss/
Less space means faster build times.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2277)


kvark:
👏 👏 👏

PR #2278, opened 2018-01-10
Adjust the clip region for shadow elements.

### LINKED ISSUE #1928: Fast (Unblurred) Shadows Mishandle Clips
Copying from my comment elsewhere:

Basically, gecko likes doing some cute tricks where it draws parts of the text multiple times with different clips to create the illusion of a single glyph/line across multiple runs with possible style changes.

Mostly this works fine, and can be naively lowered with no problem. However this causes a problem for fast shadows, because they inherit the clip from the original element. Therefore fast shadows with an offset will be clipped incorrectly and render broken.

So as a level-1 problem, when we clone the display item to push it in as a fast shadow, we also need to clone its clip and apply the shadow offset. However this creates a level-2 problem: translating the clip isn't always correct to do -- you don't want to translate a clip used for, say, overflow:hidden.

Under the current architecture there's a simple heuristic we can apply: clips that are pushed inside a shadow should be cloned and translated when pushing fast-shadows. These clips are necessarily just "part of drawing that part of the text", and not "real" clips.

Slow shadows appear to be unaffected.

GuanWen-Chen:
This patch is intend to fix #1928.

In webrender, shadow elements inherit the same clip with the original elements which leads the shadow to apply the wrong clip region when doing fast-shadow.
To fix this, when adding a shadow line primitive, I adjust PrimitiveInfo's local_clip_rect to fit the correct clip region and apply the clip outside of the shadow to the shadow line primitive.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2278)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -852,6 +852,8 @@ impl FrameBuilder {
             );
             let mut info = info.clone();
             info.rect = info.rect.translate(&shadow_offset);
+            info.local_clip =
+              LocalClip::from(info.local_clip.clip_rect().translate(&shadow_offset));
             let prim_index = self.create_primitive(
                 &info,
                 Vec::new(),
@@ -1362,6 +1364,8 @@ impl FrameBuilder {
             let rect = info.rect;
             let mut info = info.clone();
             info.rect = rect.translate(&text_prim.offset);
+            info.local_clip =
+              LocalClip::from(info.local_clip.clip_rect().translate(&text_prim.offset));
             let prim_index = self.create_primitive(
                 &info,
                 Vec::new(),
--- a/wrench/reftests/text/reftest.list
+++ b/wrench/reftests/text/reftest.list
@@ -21,6 +21,7 @@ fuzzy(1,735) == shadow-grey.yaml shadow-grey-ref.yaml
 fuzzy(1,614) == shadow-grey-transparent.yaml shadow-grey-ref.yaml
 == subtle-shadow.yaml subtle-shadow-ref.yaml
 options(disable-aa) == shadow-atomic.yaml shadow-atomic-ref.yaml
+options(disable-aa) == shadow-clip-rect.yaml shadow-atomic-ref.yaml
 options(disable-aa) == shadow-ordering.yaml shadow-ordering-ref.yaml
 != synthetic-bold.yaml synthetic-bold-not-ref.yaml
 fuzzy(1,229) options(disable-subpixel) == synthetic-bold-transparent.yaml synthetic-bold-transparent-ref.yaml
--- a/wrench/reftests/text/shadow-clip-rect.yaml
+++ b/wrench/reftests/text/shadow-clip-rect.yaml
@@ -1,0 +1,45 @@
+--- # checks that decorations on "real" content and "shadow" content are on seperate, atomic, layers
+root:
+  items:
+        -
+          type: "shadow"
+          bounds: [14, 18, 205, 35]
+          blur-radius: 0
+          offset: [20, 16]
+          color: green
+        -
+          type: "shadow"
+          bounds: [14, 18, 205, 35]
+          blur-radius: 0
+          offset: [10, 8]
+          color: black
+        -
+          type: line
+          clip-rect: [14, 18, 205, 35]
+          baseline: 45
+          start: 14
+          end: 210
+          width: 3
+          orientation: horizontal
+          color: red
+          style: solid
+        -
+          bounds: [14, 18, 205, 35]
+          clip-rect: [14, 18, 205, 35]
+          glyphs: [55, 75, 76, 86, 3, 76, 86, 3, 87, 75, 72, 3, 69, 72, 86, 87]
+          offsets: [16, 43, 35.533333, 43, 51.533333, 43, 60.4, 43, 72.833336, 43, 80.833336, 43, 89.7, 43, 102.13333, 43, 110.13333, 43, 119, 43, 135, 43, 149.2, 43, 157.2, 43, 173.2, 43, 187.4, 43, 196.26666, 43]
+          size: 18
+          color: red
+          font: "VeraBd.ttf"
+        -
+          type: line
+          clip-rect: [14, 18, 205, 35]
+          baseline: 32
+          start: 14
+          end: 210
+          width: 3
+          orientation: horizontal
+          color: red
+          style: solid
+        -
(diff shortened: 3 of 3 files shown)


glennw:
Does this seem like a reasonable solution @Gankro ?

@GuanWen-Chen Thanks for the patch! Could you also add a wrench reftest to cover this case, and a gecko try run with this patch applied?

Gankro:
I don't understand how the clip scroll tree works well enough anymore to say if this is right, but from what I remember it shouldn't be?

How does this correctly handle:

```
push_clip(A)
  push_shadow
    push_clip(B)
      line
    pop_clip
    text
  pop_shadow
pop_clip
```
(I believe this roughly corresponds to the commands sent for `<div overflow:hidden><span text-shadow; underline>AAAA</span></div>`)

Specifically only B should be offset, while A should still apply unchanged. It seems like the interesection you're doing is potentially applying this logic, but I thought the push_clip commands were too lossy today to reliably show up in the clipscroll tree?

Also whatever the correct logic, it needs to be applied to text run shadows as well, as partial ligatures do the same trick.

glennw:
Sorry, didn't get to this today - will read through it in detail tomorrow.

GuanWen-Chen:
Hello @Gankro 

Yes, I think only B (every clip nodes within shadow) should be offset, and I will apply this logic to text.
I am not sure what do you means by "the push_clip commands were too lossy today to reliably show up in the clipscroll tree". Does it mean that the clip id passed by push_clip command may not show up in clipscroll tree? And how does it effect the result?

However, I found a case this solution couldn't handle.
Consider a more complicate case(Not sure if this case will happened in practice):

push_clip(A)
  push_shadow
    push_clip(B)
      push_clip(C)
        line
      pop_clip
    pop_clip
    text
  pop_shadow
pop_clip

In this case, according to current logic, when handling the line's shadow, we will adjust the line's shadow's local_clip rect to clip C's region and apply clip B; but the clip B is the clip without shadow offset.
To correct this, we should recursively intersect the local_clip_rect to its clip and ascent parents until the clip outside the shadow.
But not sure if it will cause any side effects.

Gankro:
@mrobinson you're still the clip expert, right?

webrender/src/frame_builder.rs
```diff
@@ -843,12 +844,35 @@ impl FrameBuilder {
...
+            // an offset is needed for correct clipping; Therefore, instead of
+            // creating an additional clip node, we adjust the local clip rect
+            // in the LayerPrimitiveInfo to perform the right clipping.
+            let (_, shadow_clip) = self.shadow_prim_stack[idx].1[0];
+            let fast_clip_and_scroll;
+            if clip_and_scroll == shadow_clip {
+                fast_clip_and_scroll = clip_and_scroll;
+            } else {
+                let node = clip_scroll_tree.nodes.get(&clip_and_scroll.clip_node_id()).unwrap();
+                let mut clip_rect = node.local_viewport_rect.translate(&shadow_offset);
```
mrobinson: Hrm. This feels a bit wrong to me. How do we know that the ClipScrollNode has the same coordinate system as the shadow? What about the other clips in the clip hierarchy? I'm not sure this is the correct approach here, but it could be that I simply don't understand the problem. I worry that this is merely hiding the overall issue.

Doesn't it make sense to render

mrobinson:
Sorry for the late reply. Another question I have is how we know which clips should affect the shadow and which ones should not?

Gankro:
@mrobinson the idea is that during DL construction, some clips are pushed while a text-shadow is still pushed, and those should have the shadow offset applied when rendering the shadow. This "inner" clip is used to make wavy/dashed lines from sibling elements "line up", and to draw a glyph where different parts have different styles (by drawing it multiple times with different clips).

Here's some examples of this in action in firefox today:

screen shot 2018-01-11 at 11 45 04 am
screen shot 2018-01-11 at 11 44 51 am

mrobinson:
@Gankro I suppose it isn't possible to modify Gecko to produce another clip with the proper offset and apply it to the shadow? I worry about building assumptions that only hold true for Gecko into WebRender...

glennw:
I discussed this briefly with @Gankro on IRC. Conversation here for posterity:

```
<Gankro> gw: mrobinson and I are gonna have a call about clipping stuff next week if that's the topic
<gw> Gankro: so ... returning to shadows + clips, our favourite topic :)
<gw> Gankro: ah, ok
<gw> Gankro: might be worth referencing https://github.com/servo/webrender/pull/1935#issuecomment-342058148 in your discussion
<Gankro> gw: we're currently hoping we can restrict the transform to the local clip rect only
<gw> Gankro: but doesn't gecko not use local clip at all? (and we were considering removing local clip because of that)?
<Gankro> yes it will need (hopefully) minor gecko changes
<gw> Gankro: ok. that would certainly resolve my concern in the comment referenced above, since there is no clip ID mapping in that case
<gw> Gankro: and, I guess, this use case only needs clip rects (not *rounded* rects), right?
<Gankro> gw: yeah exactly
<gw> Gankro: ok, which would still allow us to remove the complex local clip items from the display list as an optimization...
<Gankro> mhm
<gw> Gankro: great, ok, we can wait on the outcome of that discussion then, thanks!
```

Gankro:
Ok so discussed this @mrobinson and @mstange today, and we think we've worked out what should work here. A key detail here is @demo99 (ethan lin) recently cleaned up the TextDrawTarget clipping, eliminating one of my points of confusion about the different clip sources: https://bugzilla.mozilla.org/show_bug.cgi?id=1424673

The high level idea is that we want shadows to offset *only* the local clip rect, and PushClip/PopClip on TextDrawTarget will manipulate this value.

* Add a stack of Rects to TextDrawTarget which PushClip/PopClip manipulate
* Make all uses of mClipRect instead submit the intersection of this stack (it will likely only ever have 1-2 rects)
* Change this patch to just offset each element's local clip rect

One word of caution is that this introduces a semantic distinction between ClipScrollNodes and local clips; the latter will be offset by shadows. This means it isn't always a correct optimization to squash clips into children. If such an optimization is desirable, we can potentially add machinery to remember if we're currently adding things to a shadow to supress it when necessary. (cc @kats)

Ideally we should add tests to that will break if anyone tries to do this without thinking about it.

Does that make sense to you, @GuanWen-Chen ?

GuanWen-Chen:
Ok, I will modify the code in both gecko and WR sides.

Gankro:
@bors-servo r+

awesome!

PR #2280, opened 2018-01-10
Red flicker fix/workaround

kvark:
Addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1421696
~~TODO:~~ gecko try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=27240e664f521285aa3d3ae76fce18886eba46c1&selectedJob=155374418

PR contains 3 things:
  1. extensive validation of the render target list state with assertions. Note: none of those were triggered by the bug, but it's good to have them anyway for the future.
  2. fixed reset to the previous FBO in `update_texture_storage` (this may fix some other issues, technically)
  3. `WORK_AROUND_TEX_IMAGE ` fix, which resets a render target before trying to re-initialize it

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2280)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -21,6 +21,11 @@ use std::ptr;
 use std::rc::Rc;
 use std::thread;
 
+// Apparently, in some cases calling `glTexImage3D` with
+// similar parameters that the texture already has confuses
+// Angle when running with optimizations.
+const WORK_AROUND_TEX_IMAGE: bool = cfg!(windows);
+
 #[derive(Debug, Copy, Clone, PartialEq, Ord, Eq, PartialOrd)]
 pub struct FrameId(usize);
 
@@ -467,6 +472,10 @@ impl Texture {
 
     pub fn has_depth(&self) -> bool {
         self.depth_rb.is_some()
+    }
+
+    pub fn get_rt_info(&self) -> Option<&RenderTargetInfo> {
+        self.render_target.as_ref()
     }
 }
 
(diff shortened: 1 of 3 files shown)


kvark:
@glennw [try appears green](https://treeherder.mozilla.org/#/jobs?repo=try&revision=27240e664f521285aa3d3ae76fce18886eba46c1&selectedJob=155369513) (minus some sporadic errors unrelated to WR)

glennw:
Wow, that's an unfortunate bug :( Are there any comments you could add to that workaround bool with links to further information (I guess the bugzilla bug, or is there anything known on the angle bug list etc?).

@bors-servo r+

kvark:
There is no simplified use-case yet, and no associated Angle issue. I'll make sure to correct the comment when we know more...

kvark:
OSX Travis is having a moment...
@bors-servo retry

webrender/src/device.rs
```diff
@@ -21,6 +21,11 @@ use std::ptr;
 use std::rc::Rc;
 use std::thread;
 
+// Apparently, in some cases calling `glTexImage3D` with
+// similar parameters that the texture already has confuses
+// Angle when running with optimizations.
+const WORK_AROUND_TEX_IMAGE: bool = cfg!(windows);
```
nical: Oh fun! added https://github.com/servo/webrender/wiki/Driver-issues#2280---glteximage3d-on-angle

PR #2281, opened 2018-01-10
show subpixel AA on fast text shadows

lsalzman:
This lets fast text shadows reuse the subpixel AA masks from the original text. On those platforms that use gamma-lut, this might cause a slight but barely noticeable perceptual difference between if the text was directly rendered in that color vs making a shadow with that shadow color. However, this allows us to keep avoiding double rasterization of the glyphs, allowing fast text shadows to still be fast.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2281)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -806,7 +806,7 @@ impl AlphaBatcher {
                                     } else if ctx.use_dual_source_blending {
                                         BlendMode::SubpixelDualSource
                                     } else {
-                                        BlendMode::SubpixelConstantTextColor(text_cpu.font.color.into())
+                                        BlendMode::SubpixelConstantTextColor(text_cpu.get_color())
                                     }
                                 }
                                 GlyphFormat::Alpha |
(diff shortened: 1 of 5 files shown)


Gankro:
Could you uncomment the `== shadow.yaml shadow-ref.yaml` reftest, and possibly remove the `options(disable-aa)` on `shadow-atomic` and `shadow-ordering`?

lsalzman:
Removing the disable-aa might not be possible due to the fact there can still be luminance/gamma mismatches.

Gankro:
It should at least be only slightly fuzzy?

lsalzman:
I'll try enabling in the PR and see if the failures are fuzzable.

glennw:
Seems reasonable, once we sort out the fuzziness requirements to get CI passing.

lsalzman:
So, shadow-atomic and shadow-ordering still have the stumbling block that luminance/gamma affect even grayscale AA on mac. They do pass on Linux though since we don't disable gamma-lut there.

At least shadow.yaml itself can be re-enabled now, though, with a slight fix to the test itself.

PR #2284, opened 2018-01-11
Add support for caching render tasks across frames and display lists.

glennw:
This patch adds the infrastructure required to allow render tasks
to be rendered into the texture cache, where they can be stored
and referenced across multiple frames.

Since the cache key is based on the contents of the task itself,
the cache items remain valid even when a new display list is provided,
so long as the content of the cached task is the same. When the cache
is determined to be valid, the render task is just drawn as a simple
image, rather than executing the complex render task chain used to
build the output.

Convert box-shadows to make use of this new feature. When a new
box-shadow is required, the render task chain is set up to draw the
minimal box-shadow, and blur it. The blurred result is then stored in
the texture cache, available for use on this and subsequent frames.

Since items are drawn into the texture cache, this will also help
with batching in the future, as we collapse more items to be able to
use a simple shader that draws an image from the texture cache.

Remove blur regions since they are no longer used or required.
Remove old render task sharing within a single frame, it's redundant now.

TODO (as future work):
 * Support caching of RGBA8 render tasks.
 * Convert other tasks (e.g. clips, drop-shadows) to use cacheable tasks.
 * Use this as basis for optimizing rendering of static Pictures (e.g. a picture with a filter chain + image).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2284)


glennw:
r? @kvark 

I still need to do a try run - will kick one off shortly.

As you may expect, this has some quite significant performance and power saving wins. Next step for box shadows is to optimize the clip masks and the application of the box shadows (which are still way slower than they need to be).

glennw:
Try run looks good (just one new PASS):

https://hg.mozilla.org/try/rev/2ea7546e4f797bfd4191921ebdccb428e3ffe27a
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2ea7546e4f797bfd4191921ebdccb428e3ffe27a

kvark:
Implementation looks good to me, minus a few non-blocking concerns. Amazing work!

Concept-wise, this triggers an alarm. Treating some parts of the texture cache as renderable while mostly using it as a texel storage is not going to make driver's life easier for determining the internal representation of the image (tiling mode, color compression, also cache flush points).

I suppose moving the persistent rendered results into separate texture(s) is not a great option in terms of batching and complexity. Perhaps, we'll investigate later if doing a blit from our temporary render target into texture cache is better than rendering into the texture cache directly. Blit requires the TRANSFER_DST flags in Vulkan, which we already need for copying stuff from CPU into the texture cache (with PBOs), so blitting would not incur any additional requirements for the image layout and such.

---

---

*[webrender/res/brush_image.glsl, line 65 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2aMUydF3LO0GTQfQr-:-L2aMUyfBsnWT2pKavbN:b-gav8h9) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/res/brush_image.glsl#L65)):*
> ```GLSL
>     vUv.z = blur_task.common_data.texture_layer_index;
>     vec2 texture_size = vec2(textureSize(sColor0, 0).xy);
>     vColor = blur_task.color;
> ```

could also unconditionally have `vColor` output to merge these paths

---

*[webrender/src/batch.rs, line 472 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2aaQTuJmHZbbsQgGwl:-L2aaQTvyLlVrXtaAsZx:b9toidj) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/batch.rs#L472)):*
> ```Rust
>             let batch = self.batch_list.get_suitable_batch(key, pic_metadata.screen_rect.as_ref().expect("bug"));
> 
>             let render_task_id = match *pic.surface.as_ref().expect("BUG: no surface for splitting") {
> ```

nit: can we match `pic.surface` and move the panic from `expect` here to a match arm?

---

*[webrender/src/render_task.rs, line 660 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2aezbQX4SiHVUZKE2i:-L2aezbR5swurcl4bSIo:b-xc1ojb) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/render_task.rs#L660)):*
> ```Rust
> }
> 
> pub struct RenderTaskCache {
> ```

please add some comments explain what this is

---

*[webrender/src/render_task.rs, line 693 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2ad0XuEN8tLf5D-zMY:-L2ad0XvIfa3wf6IaDpi:b39y07e) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/render_task.rs#L693)):*
> ```Rust
>         // from here so that this hash map doesn't
>         // grow indefinitely!
>         self.entries.retain(|_, value| {
> ```

can we do this by asking the texture cache if each particular entry has been evicted (as opposed to introducing `FRAMES_BEFORE_EVICTION` here)?

---

*[webrender/src/render_task.rs, line 712 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2adOdpuueofS7jL3rb:-L2adOdpuueofS7jL3rc:bjldmcf) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/render_task.rs#L712)):*
> ```Rust
>                               .entry(key)
>                               .or_insert_with(|| {
>                                 RenderTaskCacheEntry {
> ```

nit: `or_insert` would probably be as good here

---

*[webrender/src/render_task.rs, line 723 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2aeHV5sajVhUtMOZdw:-L2aeHV5sajVhUtMOZdx:b-scxhey) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/render_task.rs#L723)):*
> ```Rust
>             // Invoke user closure to get render task chain
>             // to draw this into the texture cache.
>             let (render_task_id, user_data) = f(render_tasks);
> ```

note to self: refactor to avoid passing a closure there, e.g. by having a temporary object/builder

---

*[webrender/src/texture_cache.rs, line 429 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2ahSvocBUXVylF4aGL:-L2ahSvocBUXVylF4aGM:bohxptj) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/texture_cache.rs#L429)):*
> ```Rust
>                  DeviceUintRect::new(origin, entry.size))
>             }
>             None => panic!("BUG: handle not requested earlier in frame"),
> ```

could also go with `as_ref().expect(...)` at the beginning, matching the arm body

---

*[webrender/src/tiling.rs, line 519 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2ahvGjujKvX61AWtTF:-L2ahvGjujKvX61AWtTG:b-zp3f9) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/tiling.rs#L519)):*
> ```Rust
> }
> 
> impl RenderTarget for TextureCacheRenderTarget {
> ```

that's one weird implementation... only having `add_task` doing something meaningful.
Do we have to have this implementation?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2284#-:-L2ak3qjE8rpyoXKIOLx:b-u6ncjg)*

glennw:
Review status: 15 of 21 files reviewed at latest revision, 8 unresolved discussions.

---

*[webrender/res/brush_image.glsl, line 65 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2aMUydF3LO0GTQfQr-:-L2b1G9OJfbu4nrGp-YR:by30jx2) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/res/brush_image.glsl#L65)):*

could also unconditionally have `vColor` output to merge these paths
</blockquote></details>

Merged the two similar branches with some ifdef hackery.

---

*[webrender/src/batch.rs, line 472 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2aaQTuJmHZbbsQgGwl:-L2b1MwfHSYBtVh9ho1j:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/batch.rs#L472)):*

nit: can we match `pic.surface` and move the panic from `expect` here to a match arm?
</blockquote></details>

Done.

---

*[webrender/src/render_task.rs, line 660 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2aezbQX4SiHVUZKE2i:-L2b1Nbj-JArmp1WyQ4G:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/render_task.rs#L660)):*

please add some comments explain what this is
</blockquote></details>

Done.

---

*[webrender/src/render_task.rs, line 693 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2ad0XuEN8tLf5D-zMY:-L2b1OKCXEjD44ackBwi:bfo57b0) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/render_task.rs#L693)):*

can we do this by asking the texture cache if each particular entry has been evicted (as opposed to introducing `FRAMES_BEFORE_EVICTION` here)?
</blockquote></details>

Good idea. I added ```is_allocated``` to the texture cache to check if a handle is valid, and remove items based on that.

---

*[webrender/src/render_task.rs, line 712 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2adOdpuueofS7jL3rb:-L2b1UNOD1KiBxhjQsWk:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/render_task.rs#L712)):*

nit: `or_insert` would probably be as good here
</blockquote></details>

Done.

---

*[webrender/src/texture_cache.rs, line 429 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2ahSvocBUXVylF4aGL:-L2b1Vp7adiGm8CKmUjw:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/texture_cache.rs#L429)):*

could also go with `as_ref().expect(...)` at the beginning, matching the arm body
</blockquote></details>

Done.

---

*[webrender/src/tiling.rs, line 519 at r1](https://reviewable.io:443/reviews/servo/webrender/2284#-L2ahvGjujKvX61AWtTF:-L2b1W_NM6_mCGRPLzgE:b-v7d0th) ([raw file](https://github.com/servo/webrender/blob/6099487bbb0edd6bbeddfaddcf67b7d4487853c7/webrender/src/tiling.rs#L519)):*

that's one weird implementation... only having `add_task` doing something meaningful.
Do we have to have this implementation?
</blockquote></details>

Good point - fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2284)*

glennw:
@kvark Thanks for the review! All comments addressed, I think.

I also wondered about whether it makes sense to blit from the intermediate cache in to the texture cache instead of drawing directly. It should be trivial to make it work like that in the future if profiling shows it to be worthwhile.

kvark:
:lgtm:

---

Reviewed 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2284#-:-L2b8_ejReQlOHfotbP4:bnfp4nl)*

glennw:
gwbors says this is good enough, even with travis failures.

PR #2285, opened 2018-01-11
Correctly handle the inset box shadow offset.

demo99:
If the blur radius is larger than offset size, we should still see the blur on the edge. But currently, the brush will be clipped after moving. So I enlarge the brush rect and picture rect to fix this problem.
It fixes gecko [bug 1429411](https://bugzilla.mozilla.org/show_bug.cgi?id=1429411). 
gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=5291c42ee1c846566c98e6a33b6b9e1270bea3c2&selectedJob=155567986
There is a failure and the fuzz value should be changed to (6,8936) from (6,8490). I think it should be fine.
r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2285)


glennw:
Thanks!

@bors-servo r+

larsbergstrom:
Manually merging to unblock team during testing/ops issues.

PR #2187, opened 2017-12-07
Clamp border corner radius

nc4rrillo:
Fixes #2183 

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2187)

PR #2286, opened 2018-01-11
Fix a bug in the corner radius overlap calculation.

mstange:
This fixes a bug in #2187 which I only noticed after the pull request was merged. I'm also adding a test.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2286)


mstange:
r? anyone

mstange:
Is bors stuck?

PR #2290, opened 2018-01-12
Move edge AA flags from GPU cache data into instance struct.

glennw:
This means that there is now a complete GPU block of unused data
in the GPU cache data for each segment. This will allow us to
store a UV sub-rect in here in the near future.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2290)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -23,6 +23,7 @@ struct BrushInstance {
     int clip_address;
     int z;
     int segment_index;
+    int edge_mask;
     ivec2 user_data;
 };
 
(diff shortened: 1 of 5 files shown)


glennw:
r? @kvark 

This is prep work for some changes to how we draw box shadows. Using segments for box shadow parts will allow us to make the clip-out masks much smaller (and reduce the size of the box-shadow we draw). Once this is done, each segment of the box shadow can just draw a sub-rect of the blurred picture.

larsbergstrom:
@bors-servo retry
- Testing rollout of change to gate homu on AV + TC instead.

PR #2292, opened 2018-01-12
Skip the edge if the border style is none.

demo99:
If we don't skip the BorderStyle::None edges, the shader will render the edges as solid borders. Gecko bug is [bug 1419065](https://bugzilla.mozilla.org/show_bug.cgi?id=1419065).
I add a wrench test. The fuzz is due to the border's corners. I also update an original wrench image because it was wrong.
I just pushed a gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=1e49be4dd74ae96aa2d9fac7d032e841804fb2de

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2292)


demo99:
There are some unexpected failures. I think those are because of the corner. I'll see how to fix them.

demo99:
There are some problems here. One is that we should skip the borderstyle::none. Another one is that the border corner's style should be single style if the other edge is none or hidden. The last one is the double border's width should be at least one pixel or we may render nothing in some cases.

The gecko try:  https://treeherder.mozilla.org/#/jobs?repo=try&revision=81e414007f54f45b94ec820d7225b4e72270c02a&selectedJob=156302140

There are 4 unexpected-pass and 3 unexpected-fail tests. I think the unexpected-fail tests might be AA problem.  It was covered by the borderstyle::none problem since originally we render a solid border on the borderstyle::none edge unconditionally (all collapsed tables have this problem). I want to use fuzz for the 3 unexpected-fail tests. @staktrace, can you check the try link to see if it's okay for you?
r? @glennw

staktrace:
@demo99 I'd rather use fails-if for the 3 new unexpected fails, unless you're sure this is something we're not going to fix. We have a few other tests that are also marked fails-if due to slightly different coloring on table borders and that I suspect that they all have the same root cause.

Regardless I have no objection to merging this given that it fixes 4 tests and fails 3 in what appears to be trivial differences.

glennw:
Looks good, thanks.

@bors-servo r+

PR #2295, opened 2018-01-12
Resurrect non-premultiplied alpha codepath for use in Gecko

staktrace:
I think we need this to fix some canvas reftests that have been failing for a long time. See bug https://bugzilla.mozilla.org/show_bug.cgi?id=1429951 for the gecko side patch, which has the list of failing reftests.

Also I don't really know this code well, most of this patch is just copy-paste, so please let me know if I'm way off base here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2295)

### DIFF
--- a/webrender/examples/blob.rs
+++ b/webrender/examples/blob.rs
@@ -265,6 +265,7 @@ impl Example for App {
             api::LayoutSize::new(500.0, 500.0),
             api::LayoutSize::new(0.0, 0.0),
             api::ImageRendering::Auto,
+            api::AlphaType::PremultipliedAlpha,
             blob_img1,
         );
 
@@ -274,6 +275,7 @@ impl Example for App {
             api::LayoutSize::new(200.0, 200.0),
             api::LayoutSize::new(0.0, 0.0),
             api::ImageRendering::Auto,
+            api::AlphaType::PremultipliedAlpha,
             blob_img2,
         );
 
--- a/webrender/examples/frame_output.rs
+++ b/webrender/examples/frame_output.rs
@@ -164,6 +164,7 @@ impl Example for App {
             info.rect.size,
             LayoutSize::zero(),
             ImageRendering::Auto,
+            AlphaType::PremultipliedAlpha,
             self.external_image_key.unwrap()
         );
 
--- a/webrender/examples/image_resize.rs
+++ b/webrender/examples/image_resize.rs
@@ -59,6 +59,7 @@ impl Example for App {
             image_size,
             LayoutSize::zero(),
             ImageRendering::Auto,
+            AlphaType::PremultipliedAlpha,
             self.image_key,
         );
 
@@ -71,6 +72,7 @@ impl Example for App {
             image_size,
             LayoutSize::zero(),
             ImageRendering::Pixelated,
+            AlphaType::PremultipliedAlpha,
             self.image_key,
         );
 
--- a/webrender/examples/texture_cache_stress.rs
+++ b/webrender/examples/texture_cache_stress.rs
@@ -147,6 +147,7 @@ impl Example for App {
                 image_size,
                 LayoutSize::zero(),
                 ImageRendering::Auto,
+                AlphaType::PremultipliedAlpha,
                 *key,
             );
         }
@@ -162,6 +163,7 @@ impl Example for App {
                 image_size,
                 LayoutSize::zero(),
                 ImageRendering::Auto,
+                AlphaType::PremultipliedAlpha,
                 image_key,
             );
         }
@@ -177,6 +179,7 @@ impl Example for App {
             image_size,
             LayoutSize::zero(),
             ImageRendering::Auto,
+            AlphaType::PremultipliedAlpha,
             swap_key,
         );
         self.swap_index = 1 - self.swap_index;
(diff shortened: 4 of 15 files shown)


glennw:
@mstange @kvark Do you know why we would need this? It seems like we should be able to handle these cases without introducing a non-premultiplied alpha path again, but I haven't looked closely at the tests in question.

mstange:
WebGL lets you specify `premultipliedAlpha: false` in the [WebGLContextAttributes](https://www.khronos.org/registry/webgl/specs/latest/1.0/#WEBGLCONTEXTATTRIBUTES). Then the WebGL canvas needs to be composited using unpremultiplied alpha.

glennw:
Reviewed 7 of 7 files at r1.

---

*[webrender/src/prim_store.rs, line 192 at r1](https://reviewable.io:443/reviews/servo/webrender/2295#-L2rB3CXYQ_lops6-CFb:-L2rB3CXYQ_lops6-CFc:b-8u92nm) ([raw file](https://github.com/servo/webrender/blob/a9dbc7be24367b8e57323cddf2400e42f38538dd/webrender/src/prim_store.rs#L192)):*
> ```Rust
>     pub is_backface_visible: bool,
>     pub screen_rect: Option<DeviceIntRect>,
>     pub non_premultiplied: bool,
> ```

Could we change the logic of this to be ```premultiplied``` and the default to be true? It reads a bit easier to not have a negative here, I think.

---

*[webrender/src/renderer.rs, line 3544 at r1](https://reviewable.io:443/reviews/servo/webrender/2295#-L2rBDwano0JOQTAfDGV:-L2rBDwano0JOQTAfDGW:bw5rb2u) ([raw file](https://github.com/servo/webrender/blob/a9dbc7be24367b8e57323cddf2400e42f38538dd/webrender/src/renderer.rs#L3544)):*
> ```Rust
> 
>                         match batch.key.blend_mode {
>                             BlendMode::Alpha => {
> ```

I think this can probably be a ```panic```. We should never end up in the path where we are compositing text primitives that are not premultiplied.

---

*[webrender_api/src/display_item.rs, line 70 at r1](https://reviewable.io:443/reviews/servo/webrender/2295#-L2rBRIfdvGtCchCW2RO:-L2rBRIfdvGtCchCW2RP:b-5oyhps) ([raw file](https://github.com/servo/webrender/blob/a9dbc7be24367b8e57323cddf2400e42f38538dd/webrender_api/src/display_item.rs#L70)):*
> ```Rust
>     pub local_clip: LocalClip,
>     pub is_backface_visible: bool,
>     pub non_premultiplied: bool,
> ```

Can we make this a property of the ```Image``` display item instead of a generic property that applies to all primitives? That way the comment above (about text prims) becomes irrelevant, we don't pay the memory cost of extra bool for every item. I think that should cover what we need, since this only applies to webgl images?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2295)*

staktrace:
> Can we make this a property of the Image display item instead of a generic property that applies 
> to all primitives? That way the comment above (about text prims) becomes irrelevant, we don't 
> pay the memory cost of extra bool for every item. I think that should cover what we need, since 
> this only applies to webgl images?

We can do this, but it will mean passing around this extra bool through a bunch of functions like create_primitive and add_primitive. That was the main reason I put it into the LayerPrimitiveInfo, so that it would avoid polluting the function signatures.

kvark:
Review status: 0 of 17 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/src/prim_store.rs, line 192 at r1](https://reviewable.io:443/reviews/servo/webrender/2295#-L2rB3CXYQ_lops6-CFb:-L2v09I01T44zFmpU8-4:bn7ng97) ([raw file](https://github.com/servo/webrender/blob/a9dbc7be24367b8e57323cddf2400e42f38538dd/webrender/src/prim_store.rs#L192)):*

Could we change the logic of this to be ```premultiplied``` and the default to be true? It reads a bit easier to not have a negative here, I think.
</blockquote></details>

Perhaps, we should just use an `enum` for that. Passing true/false in many places is not idiomatic, since it's not clear what it means without checking the function signatures.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2295)*

staktrace:
> Perhaps, we should just use an enum for that. Passing true/false in many places is not idiomatic, 
> since it's not clear what it means without checking the function signatures.

Before I go and do this, can you take a look at the second patch I added and let me know if you prefer making that change or not? (Without this second patch, the flag is on PrimitiveInfo, but with the patch it's in ImageDisplayItem and then gets propagated to a bunch of function signatures as an extra parameter).

Also, if I convert it to an enum, is it fine to just one have enum exposed in the display_item.rs file and use that everywhere, or do you want a frontend/backend separation, with another enum defined in webrender/ somewhere?

glennw:
(Discussed on IRC) I think if we store the bool (or enum) in the ```ImagePrimitiveCpu``` struct, we should be able to avoid any changes to fn signatures.

glennw:
Having one enum is fine, I think.

staktrace:
Updated patch to use one AlphaType enum and propagate it via ImagePrimitiveCpu. ^

glennw:
Looks good, thanks! As a follow up, we should add a wrench reftest for this case.

@bors-servo r+

PR #2296, opened 2018-01-12
Add travis_wait to override the 10 minute timeout.

jrmuizel:
We seem to be hitting this on every PR right now. Let's see if
this might fix it.

See documentation:

https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2296)


jrmuizel:
I added an 'unset -f cd' to deal with https://github.com/travis-ci/travis-ci/issues/8703

glennw:
@bors-servo retry

@kvark @jrmuizel  This job failed with exit code 137 - which I *think* typically means killed by OOM. I'm not sure what the best way forward is - maybe running with ```-j 1``` might help, by reducing the number of active processes.

glennw:
I pushed a change here to restrict cargo building to one build job at a time. Let's see if that helps with the OOM issue.

@bors-servo p=1 r=glennw,jrmuizel

glennw:
This doesn't seem to help.

glennw:
I think we have two issues - sometimes 10min timeout related, sometimes OOM related. Updated this patch to just cover the timeout - this may still be helpful to reduce the failure rate.

@bors-servo r=glennw,jrmuizel p=100

jrmuizel:
Oops, I just overrode this by trying something else out.

glennw:
@jrmuizel Want to add back the ```travis_wait``` in this PR and we'll try to merge this next, to see if we get any useful log info?

jrmuizel:
Done.

glennw:
Looks like this has now failed due to exceeding the 4MB log limit :-1:

PR #2298, opened 2018-01-12
Move code to the new transaction API.

nical:
Cleaning up after #1963.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2298)

### DIFF
--- a/webrender/examples/animation.rs
+++ b/webrender/examples/animation.rs
@@ -92,9 +92,9 @@ impl Example for App {
                 let new_transform = self.transform
                     .pre_rotate(0.0, 0.0, 1.0, Angle::radians(angle))
                     .post_translate(LayoutVector3D::new(offset_x, offset_y, 0.0));
-                api.generate_frame(
-                    document_id,
-                    Some(DynamicProperties {
+                let mut txn = Transaction::new();
+                txn.update_dynamic_properties(
+                    DynamicProperties {
                         transforms: vec![
                             PropertyValue {
                                 key: self.property_key,
@@ -107,8 +107,10 @@ impl Example for App {
                                 value: self.opacity,
                             }
                         ],
-                    }),
+                    },
                 );
+                txn.generate_frame();
+                api.send_transaction(document_id, txn);
                 self.transform = new_transform;
             }
             _ => (),
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -355,19 +355,23 @@ impl Example for App {
     }
 
     fn on_event(&mut self, event: glutin::Event, api: &RenderApi, document_id: DocumentId) -> bool {
+        let mut txn = Transaction::new();
         match event {
             glutin::Event::Touch(touch) => match self.touch_state.handle_event(touch) {
                 TouchResult::Pan(pan) => {
-                    api.set_pan(document_id, pan);
-                    api.generate_frame(document_id, None);
+                    txn.set_pan(pan);
                 }
                 TouchResult::Zoom(zoom) => {
-                    api.set_pinch_zoom(document_id, ZoomFactor::new(zoom));
-                    api.generate_frame(document_id, None);
+                    txn.set_pinch_zoom(ZoomFactor::new(zoom));
                 }
                 TouchResult::None => {}
             },
             _ => (),
+        }
+
+        if !txn.is_empty() {
+            txn.generate_frame();
+            api.send_transaction(document_id, txn);
         }
 
         false
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -196,6 +196,7 @@ pub fn main_wrapper<E: Example>(
         events.push(event);
         events.extend(window.poll_events());
 
+        let mut txn = Transaction::new();
         for event in events {
             match event {
                 glutin::Event::Closed |
@@ -249,8 +250,7 @@ pub fn main_wrapper<E: Example>(
                     _,
                     Some(glutin::VirtualKeyCode::Key1),
                 ) => {
-                    api.set_window_parameters(
-                        document_id,
+                    txn.set_window_parameters(
                         framebuffer_size,
                         DeviceUintRect::new(DeviceUintPoint::zero(), framebuffer_size),
                         1.0
@@ -261,8 +261,7 @@ pub fn main_wrapper<E: Example>(
                     _,
                     Some(glutin::VirtualKeyCode::Key2),
                 ) => {
-                    api.set_window_parameters(
-                        document_id,
+                    txn.set_window_parameters(
                         framebuffer_size,
                         DeviceUintRect::new(DeviceUintPoint::zero(), framebuffer_size),
                         2.0
@@ -299,7 +298,6 @@ pub fn main_wrapper<E: Example>(
                         pipeline_id,
                         document_id,
                     );
-                    let mut txn = Transaction::new();
                     txn.set_display_list(
                         epoch,
                         None,
@@ -309,10 +307,10 @@ pub fn main_wrapper<E: Example>(
                     );
                     txn.update_resources(resources);
                     txn.generate_frame();
-                    api.send_transaction(document_id, txn);
                 }
             }
         }
+        api.send_transaction(document_id, txn);
 
         renderer.update();
         renderer.render(framebuffer_size).unwrap();
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -177,17 +177,18 @@ pub fn main_wrapper<E: Example>(
         pipeline_id,
         document_id,
     );
-    api.set_display_list(
-        document_id,
+    let mut txn = Transaction::new();
+    txn.set_display_list(
         epoch,
         None,
         layout_size,
         builder.finalize(),
         true,
-        resources,
     );
-    api.set_root_pipeline(document_id, pipeline_id);
-    api.generate_frame(document_id, None);
+    txn.update_resources(resources);
+    txn.set_root_pipeline(pipeline_id);
+    txn.generate_frame();
+    api.send_transaction(document_id, txn);
 
     println!("Entering event loop");
     'outer: for event in window.wait_events() {
@@ -298,16 +299,17 @@ pub fn main_wrapper<E: Example>(
                         pipeline_id,
                         document_id,
                     );
-                    api.set_display_list(
-                        document_id,
+                    let mut txn = Transaction::new();
+                    txn.set_display_list(
                         epoch,
                         None,
                         layout_size,
                         builder.finalize(),
                         true,
-                        resources,
                     );
-                    api.generate_frame(document_id, None);
+                    txn.update_resources(resources);
+                    txn.generate_frame();
+                    api.send_transaction(document_id, txn);
                 }
             }
         }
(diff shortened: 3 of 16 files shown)


glennw:
Looks like a timeout failure on both the mac and linux TC runs - possibly related to this PR or just a TC issue?

nical:
I bisected it down to the second commit. The timeouts are happening when a transaction both scrolls and builds the frame. The problem is that [wrench does not wake up the event loop when a transaction has scrolling](https://github.com/servo/webrender/blob/d3f72e1f4f13cc1b470fa7518710bd8e845f3e72/wrench/src/wrench.rs#L103) which looks like a mistake.
I believe we should at least wake the event loop up if `composite_needed` is true.

nical:
As a first step I worked around the issue by having wrench send a separate transaction to generate the frame after scrolling. I filed #2303 to figure this out and fix it properly, since I am not sure I know the all of the reasons for skipping scrolling updates in the first place.

nical:
New failures which look also related to scrolling (this time this is the timing of the snapshot that might be different, maybe?).

glennw:
Reviewed 15 of 16 files at r1, 1 of 1 files at r3.

---

*[wrench/src/wrench.rs, line 502 at r3](https://reviewable.io:443/reviews/servo/webrender/2298#-L2xMkwfcwJxNHHmvlQU:-L2xMkwfcwJxNHHmvlQV:b-352cqo) ([raw file](https://github.com/servo/webrender/blob/eabc717666ea39f36e42fdfeea8573f04604df78/wrench/src/wrench.rs#L502)):*
> ```Rust
>         // generate a frame after scrolling, wrench just stops and some tests
>         // will time out.
>         // I suppose this was to avoid taking the screeshot after scrolling if
> ```

nit: screenshot

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

glennw:
@bors-servo delegate+

I'll add delegate here so you can merge whenever ready to. Not sure if @kvark has any ideas about the scrolling + transaction comments.

kvark:
Reviewed 15 of 16 files at r1, 1 of 1 files at r4.

---

*[webrender/examples/document.rs, line 137 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zdctgA3NnpLQFjRUm:-L2zdcthjDn2UBm4SCz8:b-ij9990) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/webrender/examples/document.rs#L137)):*
> ```Rust
>             );
>             txn.generate_frame();
>             api.send_transaction(doc.id, txn);
> ```

does it ever make sense to have `generated_frame` not at the end? In this case, maybe it shouldn't be a dedicated command and a part of the `Transaction` but instead just a parameter to `send_transaction`

---

*[webrender/examples/common/boilerplate.rs, line 184 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zeWCFS5rdAYD4hXoF:-L2zeWCFS5rdAYD4hXoG:b-5dxr5t) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/webrender/examples/common/boilerplate.rs#L184)):*
> ```Rust
>         true,
>     );
>     txn.update_resources(resources);
> ```

Now that we have `Transaction`, does it still make sense to encapsulate a bunch of resource updates into `ResourceUpdates` struct? I'd expect us to just have a resource update be one of the thing a transaction can do.

---

*[wrench/src/wrench.rs, line 488 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zfThnVYRdgY4YTmEI:-L2zfThoiqkWEDsbfLzQ:bg33gze) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/wrench/src/wrench.rs#L488)):*
> ```Rust
>         // operations into separate transactions for mysterious -but probably related
>         // to the other comment below- reasons.
>         self.api.send_transaction(self.document_id, txn);
> ```

hmm, would be great to know what's going on there, since it may affect the transaction stuff in general

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

kvark:
Sorry, going to stop this for the moment we discuss the review points.
@bors-servo r-

Also, could you please reword those commits to something meaningful or squash otherwise?
> Remove some of the old APIs. ef2446f
> Remove more of the old APIs. 2032b54
> Remove more of the old APIs. 181f34c

kvark:
Not sure why bors didn't stop... I cancelled the Travis job on the `auto` branch for now.

nical:
*[webrender/examples/common/boilerplate.rs, line 184 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zeWCFS5rdAYD4hXoF:-L2zokYLp6un_KFhvKAQ:b-jvq5f5) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/webrender/examples/common/boilerplate.rs#L184)):*

Now that we have `Transaction`, does it still make sense to encapsulate a bunch of resource updates into `ResourceUpdates` struct? I'd expect us to just have a resource update be one of the thing a transaction can do.
</blockquote></details>

I don't mind either way (well, maybe a have a slight preference towards merging ResourceUpdates and Transaction). I don't think this should block this PR, though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

kvark:

---

*[webrender/examples/document.rs, line 137 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zdctgA3NnpLQFjRUm:-L2zpMDdOiOq4lnkWmLy:b-6iyg57) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/webrender/examples/document.rs#L137)):*

Indeed generating the frame should always be the last thing to do in a transaction. My plan was to remove this from the `DocumentMsg` enum and store it as a boolean in the transaction without changing the API. I prefer to keep this in the transaction object rather than the `send_transaction` call, because the code that builds the transaction knows whether to generate the frame but not necessarily the piece of code that sends it.
</blockquote></details>

Could you point me to such code for an example?

---

*[webrender/examples/common/boilerplate.rs, line 184 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zeWCFS5rdAYD4hXoF:-L2zpQyvbiZnUPFkuv4L:b-t77xc3) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/webrender/examples/common/boilerplate.rs#L184)):*

I don't mind either way (well, maybe a have a slight preference towards merging ResourceUpdates and Transaction). I don't think this should block this PR, though.
</blockquote></details>

agreed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

kvark:

---

*[wrench/src/wrench.rs, line 488 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zfThnVYRdgY4YTmEI:-L2zrD4UopFJnbiY5-0X:b1wluli) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/wrench/src/wrench.rs#L488)):*

The problem (on the surface) is that wrench hard-codes ignoring `new_document_ready` if the `scrolled` parameter is true. It would be good to figure out why and change that since I believe that's not great. In the mean time this workaround preserves the previous behavior by sending the scrolling updates in their own transaction. I would prefer to investigate this in a followup instead of here since it's not really related to the transaction API.
</blockquote></details>

IIRC, this is because YAML reads the scroll offsets and them needs to apply them all at once. If it sends the scrolls one by one, then the first generated frame would only have one scroll, and the corresponding reftest would fail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

nical:
*[webrender/examples/document.rs, line 137 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zdctgA3NnpLQFjRUm:-L2zqZ7RzW_PJcCj5rYk:b-4x9x68) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/webrender/examples/document.rs#L137)):*

Could you point me to such code for an example?
</blockquote></details>

The idea is to refactor [this code](https://searchfox.org/mozilla-central/rev/41925c0b6c6d58578690121de439d2a8d3d690f3/gfx/layers/wr/AsyncImagePipelineManager.cpp#284) that handles inserting video elements to be able to take a `Transaction` in parameter and append stuff to it instead of always generating separate transactions for videos. This video handling code would not actually be where we send the transaction. We could pass the information about whether to generate the frame to the caller but we might as well store it in the transaction.  The code currently is only superficially adapted to the new API so it doesn't yet take advantage of anything that requires more changes than grouping things that happened in the same scope or function into the same transaction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

nical:
*[wrench/src/wrench.rs, line 488 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zfThnVYRdgY4YTmEI:-L2zrp5xrReJuNe0Yz8u:bxj3wik) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/wrench/src/wrench.rs#L488)):*

IIRC, this is because YAML reads the scroll offsets and them needs to apply them all at once. If it sends the scrolls one by one, then the first generated frame would only have one scroll, and the corresponding reftest would fail.
</blockquote></details>

Ah, interesting. So we should be able to make it so we either:
- don't generate frames on scrolls and rather than ignore scrolls just listen to generated frames,
- or (probably better), group all of the yaml operations in the same transaction and listen to generated frames (by generated frames I meant the composite_needed parameter).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

kvark:

---

*[wrench/src/wrench.rs, line 488 at r4](https://reviewable.io:443/reviews/servo/webrender/2298#-L2zfThnVYRdgY4YTmEI:-L2zv-T4f_GV4fFL6Yf5:bljwycy) ([raw file](https://github.com/servo/webrender/blob/dc1ff7fe046bb74c6ad63c4d5e6c2699c5b41bb4/wrench/src/wrench.rs#L488)):*

Ah, interesting. So we should be able to make it so we either:
- don't generate frames on scrolls and rather than ignore scrolls just listen to generated frames,
- or (probably better), group all of the yaml operations in the same transaction and listen to generated frames (by generated frames I meant the composite_needed parameter).
</blockquote></details>

yep, this is where transaction really helps. If all those scrolls are done at once, then the render notifier can always properly generate frames without ignoring scrolls

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2298)*

kvark:
@nical sorry about the need to rebase!
If I understand correctly, this is now blocked by
> I'm writing a patch for bug 1430660 to use the transaction one instead. If this lands before that's done it'll cause bustage on the gecko side.

staktrace:
Bug 1430660 has landed, so this is good to go. Thanks!

kvark:
@nical do you want to squash or reword those?
> Remove some of the old APIs. b631e6a
> Remove more of the old APIs. 43fe52c
> Remove more of the old APIs. 4e1de36

nical:
Squashed. @bors-servo r+

kvark:
@nical 
> Squashed. bors-servo r+

should have been "r=glennw,kvark"

PR #2300, opened 2018-01-14
Add support for API-defined ClipChains

### LINKED ISSUE #1964: Allow items to be clipped by the intersection of multiple clip chains
I'm not convinced this is entirely covered by #840, so I'm filing a new issue.

The scenario that we want to support is described by Markus at https://bugzilla.mozilla.org/show_bug.cgi?id=1408792#c13 - the problem is that the WR API doesn't allow us to say "this item should be clipped by the intersection of these two clip chains". We can only specify a single clip chain per item, plus a "local clip" on the item, which isn't powerful enough to support the use cases in question.

mrobinson:
API-defined clip chains allow constructing combinations of clips that
select any combination of clips from the ClipScrollTree without regard
to parents of clips. A ClipChain is defined by an optional ClipChain
parent and a vector of clips to use for that chain. A chain's clips are
combined with its parents clips to provide all the clips for display
items that set their clipping ClipId as the clip chain via
ClipId::ClipChain.

Fixes #1964.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2300)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -2,9 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{ClipId, DeviceIntRect, DevicePixelScale, LayerPoint, LayerRect, LayerToWorldTransform};
-use api::{LayerVector2D, LayoutTransform, PipelineId, PropertyBinding, ScrollClamping};
-use api::{ScrollEventPhase, ScrollLayerState, ScrollLocation, WorldPoint};
+use api::{ClipId, ClipChainId, DeviceIntRect, DevicePixelScale, LayerPoint, LayerRect};
+use api::{LayerToWorldTransform, LayerVector2D, LayoutTransform, PipelineId, PropertyBinding};
+use api::{ScrollClamping, ScrollEventPhase, ScrollLayerState, ScrollLocation, WorldPoint};
 use clip::ClipStore;
 use clip_scroll_node::{ClipScrollNode, NodeType, ScrollingState, StickyFrameInfo};
 use gpu_cache::GpuCache;
@@ -40,8 +40,23 @@ impl CoordinateSystemId {
     }
 }
 
+struct ClipChainDescriptor {
+    id: ClipChainId,
+    parent: Option<ClipChainId>,
+    clips: Vec<ClipId>,
+}
+
 pub struct ClipScrollTree {
     pub nodes: FastHashMap<ClipId, ClipScrollNode>,
+
+    /// A Vec of all descriptors that describe ClipChains in the order in which they are
+    /// encountered during display list flattening. ClipChains are expected to never be
+    /// the children of ClipChains later in the list.
+    clip_chains_descriptors: Vec<ClipChainDescriptor>,
+
+    /// A HashMap of built ClipChains that are described by `clip_chains_descriptors`.
+    pub clip_chains: FastHashMap<ClipChainId, ClipChain>,
+
     pub pending_scroll_offsets: FastHashMap<ClipId, (LayerPoint, ScrollClamping)>,
 
     /// The ClipId of the currently scrolling node. Used to allow the same
@@ -94,6 +109,8 @@ impl ClipScrollTree {
         let dummy_pipeline = PipelineId::dummy();
         ClipScrollTree {
             nodes: FastHashMap::default(),
+            clip_chains_descriptors: Vec::new(),
+            clip_chains: FastHashMap::default(),
             pending_scroll_offsets: FastHashMap::default(),
             currently_scrolling_node_id: None,
             root_reference_frame_id: ClipId::root_reference_frame(dummy_pipeline),
@@ -243,6 +260,8 @@ impl ClipScrollTree {
         }
 
         self.pipelines_to_discard.clear();
+        self.clip_chains.clear();
+        self.clip_chains_descriptors.clear();
         scroll_states
     }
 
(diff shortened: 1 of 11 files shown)


mrobinson:
cc @staktrace

mrobinson:
Here is the Gecko try run for this PR: https://treeherder.mozilla.org/#/jobs?repo=try&revision=192ed24cbbacbd7fd2807f3fc06e797377dee5f3

glennw:

---

*[wrench/reftests/clip/custom-clip-chains.yaml, line 1 at r1](https://reviewable.io:443/reviews/servo/webrender/2300#-L2r67ynYhzuve4jRPhc:-L2r67ynYhzuve4jRPhd:b-hnsqn3) ([raw file](https://github.com/servo/webrender/blob/ce16fefa46869e011e9bb97645450dbcfea4f3af/wrench/reftests/clip/custom-clip-chains.yaml#L1)):*
> ```YAML
> root:
> ```

Could we add a comment here describing what this is testing for?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2300)*

glennw:
@staktrace Do you want to take a look at the API here before it gets merged?

mrobinson:
Review status: 10 of 11 files reviewed at latest revision, 1 unresolved discussion.

---

*[wrench/reftests/clip/custom-clip-chains.yaml, line 1 at r1](https://reviewable.io:443/reviews/servo/webrender/2300#-L2r67ynYhzuve4jRPhc:-L2t95ENh6b2KazcscY2:b-896fix) ([raw file](https://github.com/servo/webrender/blob/ce16fefa46869e011e9bb97645450dbcfea4f3af/wrench/reftests/clip/custom-clip-chains.yaml#L1)):*

Could we add a comment here describing what this is testing for?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2300)*

staktrace:
I think this should provide sufficient expressiveness to deal with all the scenarios that we need. Thanks!

glennw:
Needs a rebase.

mrobinson:

---

*[webrender/src/clip_scroll_tree.rs, line 470 at r2](https://reviewable.io:443/reviews/servo/webrender/2300#-L2zjJ7FPckAI-1nPdnF:-L40EhGW-ixdVlx5eT92:b-g8o8r3) ([raw file](https://github.com/servo/webrender/blob/f43028af555016e5d9d24480f2a4e72678472898/webrender/src/clip_scroll_tree.rs#L470)):*

nit: can do a within a parent pattern match
</blockquote></details>

I'm not sure how to do this without running past the end of the line...

---

*[webrender/src/clip_scroll_tree.rs, line 476 at r2](https://reviewable.io:443/reviews/servo/webrender/2300#-L2zibtvzYZPJzbv9Bbq:-L40EqR8sNVBANX94VwZ:b-pxksd4) ([raw file](https://github.com/servo/webrender/blob/f43028af555016e5d9d24480f2a4e72678472898/webrender/src/clip_scroll_tree.rs#L476)):*

I don't clearly understand what is going on here. We take `chain` as the parent clip chain of the `descriptor`. We populate the child nodes from `descriptor.clips`. But then we are inserting it as `descriptor.id`?

Ah, I guess it's actually simple: Current clip chain = Parent clip chain + all the current clip nodes. Makes sense now :) Perhaps, we could have a comment on that function?

Could we re-use the allocations within `self.clip_chains[descriptor.id]` instead of throwing it out?
</blockquote></details>

I'm not sure exactly how to reuse the allocation here, but I will add a comment explaining what is going on. We do need to clone the parent node, but we are just cloning an Rc of the node.

---

*[webrender/src/clip_scroll_tree.rs, line 663 at r2](https://reviewable.io:443/reviews/servo/webrender/2300#-L2zjrb_DUOdTu2QovbL:-L40F9Q4W_uummmPvy4b:bumkzxv) ([raw file](https://github.com/servo/webrender/blob/f43028af555016e5d9d24480f2a4e72678472898/webrender/src/clip_scroll_tree.rs#L663)):*

do `nodes` contain `ClipId::ClipChain` items? If yes, can we just have this path followed here universally (perhaps, with a debug assert that the clip chain matches the one referred by `clip_chain_id`).
</blockquote></details>

Eventually, it might make sense to have all ClipChains represented by ClipChainDescriptors, but for the moment it is hard to unify the two code paths.  `ClipChainNode::update_clip_work_item` does a few extra optimizations that only make sense for ClipScrollNodes.

---

*[webrender_api/src/display_item.rs, line 435 at r2](https://reviewable.io:443/reviews/servo/webrender/2300#-L2zkbJKImOGTtDD9fGr:-L40FYJjhZs8iaJfHnkm:b-nspxji) ([raw file](https://github.com/servo/webrender/blob/f43028af555016e5d9d24480f2a4e72678472898/webrender_api/src/display_item.rs#L435)):*

we can use `stops: PhantomData<Vec<Clipid>>,`
</blockquote></details>

Hrm. Does it make sense to change all of the `// IMPLICIT` lines together in a separate patch?

---

*[webrender_api/src/display_list.rs, line 1350 at r2](https://reviewable.io:443/reviews/servo/webrender/2300#-L2zlFABKMW_heKHM5wX:-L40Ff1YD4ojGpSH2rTQ:b97tuur) ([raw file](https://github.com/servo/webrender/blob/f43028af555016e5d9d24480f2a4e72678472898/webrender_api/src/display_list.rs#L1350)):*

nit: could probably not subtracting 1 here
</blockquote></details>

Is there a good way to do this without keeping an extra variable to record the previous number or the new clip id?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2300)*

mrobinson:
Sorry for taking so long to reply to these comments. I have replied above. I have a patch ready with the new comments which I'll post soon.

kvark:
There is no ability to continue the review here, since the PR is merged and branch is deleted.

> Hrm. Does it make sense to change all of the // IMPLICIT lines together in a separate patch?

Yep.

PR #2302, opened 2018-01-15
Simple example to check how many 1080p alpha passes a GPU can do.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2302)


glennw:
Fixed the CI error.

@bors-servo r=nical

kvark:
From the GPU standpoint, this is analogous to my GL benchmark https://github.com/kvark/gl-bench
We might want to confirm that the numbers are not too far, otherwise WR overhead is too much.

PR #2305, opened 2018-01-15
Capture infrastructure: Part II

### LINKED ISSUE #2238: Capture: post-built frame layer support
Current capturing of #2232 fetches all the internal structures used **before** building a frame. In some cases, the bugs we are seeing may occur due to accumulation effects in the internal re-used structures, such as `TextureCache` or `GpuCache`. It would be helpful to save exactly their state in a capture, and it doesn't seem to difficult.

This serialization would touch way more internal structures, and I suppose needs to go as a separate cargo feature.

kvark:
Fixes #2238

The PR implements support for capturing all the built-frame stage environment:
  - built frames for each document
  - cached fonts, glyphs, images, and render targets
  - gpu cache

How to use:
  - run any example with "--features capture"
  - hit "c" to capture, the data is written into "captures/example"
  - run wrench with "load" command

TODO:
- [x] cleanup
- [x] rebase
- [x] reviews
- [x] proper external textures support (`frame_output` example)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2305)


glennw:
Reviewed 31 of 35 files at r1.
Review status: 31 of 35 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[webrender/src/glyph_cache.rs, line 13 at r1](https://reviewable.io:443/reviews/servo/webrender/2305#-L2wOrJm_eLSgKj_g1bV:-L2wOrJm_eLSgKj_g1bW:bm9ayof) ([raw file](https://github.com/servo/webrender/blob/2640ce88d80308a4b63b494f60ef286b257d1706/webrender/src/glyph_cache.rs#L13)):*
> ```Rust
> 
> #[cfg_attr(feature = "capture", derive(Deserialize, Serialize))]
> pub struct GenericCachedGlyphInfo<D> {
> ```

Maybe it will become clearer later in the review - but why does this need to be made generic? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2305)*

glennw:
Partial review only so far - will be working through the other files today.

kvark:
Review status: 31 of 35 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[webrender/src/glyph_cache.rs, line 13 at r1](https://reviewable.io:443/reviews/servo/webrender/2305#-L2wOrJm_eLSgKj_g1bV:-L2wZEgFKk1-Vq71Ru-R:bljmvvk) ([raw file](https://github.com/servo/webrender/blob/2640ce88d80308a4b63b494f60ef286b257d1706/webrender/src/glyph_cache.rs#L13)):*

Maybe it will become clearer later in the review - but why does this need to be made generic? 
</blockquote></details>

The general approach is to have all the `Arc<Vec<u8>>` things replaced by `String` storing the relative path to a binary file (e.g. `images/10.raw`). In order to avoid code duplication, I made this struct generic over the data.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2305)*

glennw:
Reviewed 4 of 35 files at r1, 1 of 1 files at r2.

---

*[webrender/src/device.rs, line 132 at r2](https://reviewable.io:443/reviews/servo/webrender/2305#-L2xP3z8PP2qrkabFkCv:-L2xP3z8PP2qrkabFkCw:b-yvgm96) ([raw file](https://github.com/servo/webrender/blob/1d397569536ea7a20989e2a442b2dafcb5e6ef00/webrender/src/device.rs#L132)):*
> ```Rust
> 
> #[derive(Clone, Copy, Debug, PartialEq)]
> pub enum ReadPixelsFormat {
> ```

I wonder if we should remove this and just use ImageFormat?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2305)*

kvark:

---

*[webrender/src/device.rs, line 132 at r2](https://reviewable.io:443/reviews/servo/webrender/2305#-L2xP3z8PP2qrkabFkCv:-L2zXcw6vDCMBHfVyL2W:bdsu1gl) ([raw file](https://github.com/servo/webrender/blob/1d397569536ea7a20989e2a442b2dafcb5e6ef00/webrender/src/device.rs#L132)):*

I wonder if we should remove this and just use ImageFormat?
</blockquote></details>

I wish we could. Problem is - the `ImageFormat` doesn't have RGBA versus BGRA...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2305)*

kvark:
@glennw please take a look at the last commit dealing with external images. There is nothing particularly concerning there, but it's new code. `frame_output` example capturing works now 🎉 , and this should be ready to land.

glennw:
The last commit looks good to me. You could perhaps expand on the ```TODO?``` comments with some more details, but otherwise looks fine.

kvark:
Thanks @glennw ! This is done now.
@bors-servo r=glennw

kvark:
@bors-servo r-
Didn't realize `read_pixels_into` is used by Gecko. Need to re-expose it.

kvark:
Should be good now. `read_pixels_into` is back to original form.
@bors-servo r=glennw

PR #2306, opened 2018-01-15
Add support for regions to the segment builder.

glennw:
This allows specifying an inner rect for the segment builder,
which defines the main primitive as a 9-patch. Then, when
segments are created, the region they belong to is passed to
the segment closure.

This allows us to do things like build segments for a box-shadow
primitive, where the box-shadow is made up of sub-regions (i.e.
the corners, edges and inner region are separate regions).

In the case of a fast box shadow path, the region can be used
to identify the UV sub-rect from the blurred image to draw with.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2306)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1385,6 +1385,7 @@ impl PrimitiveStore {
 
         let mut segment_builder = SegmentBuilder::new(
             metadata.local_rect,
+            None,
             metadata.local_clip_rect
         );
 
--- a/webrender/src/segment.rs
+++ b/webrender/src/segment.rs
@@ -5,8 +5,16 @@
 use api::{BorderRadius, ClipMode, LayerPoint, LayerPointAu, LayerRect, LayerSize};
 use app_units::Au;
 use prim_store::EdgeAaSegmentMask;
-use std::cmp;
+use std::{cmp, usize};
 use util::extract_inner_rect_safe;
+
+bitflags! {
+    pub struct ItemFlags: u8 {
+        const X_ACTIVE = 0x1;
+        const Y_ACTIVE = 0x2;
+        const HAS_MASK = 0x4;
+    }
+}
 
 // The segment builder outputs a list of these segments.
 #[derive(Debug, PartialEq)]
@@ -14,6 +22,8 @@ pub struct Segment {
     pub rect: LayerRect,
     pub has_mask: bool,
     pub edge_flags: EdgeAaSegmentMask,
+    pub region_x: usize,
+    pub region_y: usize,
 }
 
 // The segment builder creates a list of x/y axis events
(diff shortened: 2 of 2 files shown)


glennw:
r? @kvark

This contains the logic and unit tests for using regions in the segment builder.

The box-shadow code to make use of this isn't quite ready. I thought I'd open this PR up now
since it might be easier to review as an isolated patch (most of the new lines are tests). If
you'd prefer to wait until the box-shadow patch that uses this functionality is ready, it's fine
to just leave this for now.

mstange:
> This allows specifying an inner rect for the segment builder, which defines the main primitive as a 9-patch.

I haven't read the code, but based on this sentence, I don't think this is sufficient. If you have a large top left corner and a small top right corner, then doesn't this mean that the top right corner's segment size will be unnecessarily large? And if the top left corner and bottom right corner overlap, there's not even a valid inner rect that describes the necessary segmentation.

glennw:
@mstange You're correct - this initial implementation uses a nine-patch, since that will handle the common case (and won't segment at all in the cases you described above). I've done it this way to keep the initial implementation simple, but the segment builder internally can handle arbitrary region definitions, so we can expand it later. Does that sound OK?

mstange:
I see, that makes sense. I just didn't want us to get stuck after going down the wrong path. But it sounds like this is actually just a stop on the right path.

kvark:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[webrender/src/segment.rs, line 43 at r1](https://reviewable.io:443/reviews/servo/webrender/2306#-L2waXiSSS2oUr19Kok5:-L2waXiSSS2oUr19Kok6:bu2oxqp) ([raw file](https://github.com/servo/webrender/blob/b2a91765fd52cc16b4ff79334379b46bc40d4484/webrender/src/segment.rs#L43)):*
> ```Rust
>     Begin,
>     End,
>     BeginRegion,
> ```

could we have some internal docs here explaining the difference between `Begin` and `BeginRegion`?

---

*[webrender/src/segment.rs, line 53 at r1](https://reviewable.io:443/reviews/servo/webrender/2306#-L2waN_onuAVhdKhPlXp:-L2waN_pe-2g8R3dxgK5:b-79slec) ([raw file](https://github.com/servo/webrender/blob/b2a91765fd52cc16b4ff79334379b46bc40d4484/webrender/src/segment.rs#L53)):*
> ```Rust
>     fn cmp(&self, other: &EventKind) -> cmp::Ordering {
>         match (*self, *other) {
>             (EventKind::BeginRegion, EventKind::BeginRegion) => {
> ```

I recall that the previous attempt to get this derived automatically didn't succeed, but we haven't figured out why.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2306#-:-L2wbloJ-NpCendJ5WvH:bnfp4nl)*

glennw:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/segment.rs, line 43 at r1](https://reviewable.io:443/reviews/servo/webrender/2306#-L2waXiSSS2oUr19Kok5:-L2xJugd48cgLPVyRaFr:b-3ya5bf) ([raw file](https://github.com/servo/webrender/blob/b2a91765fd52cc16b4ff79334379b46bc40d4484/webrender/src/segment.rs#L43)):*

could we have some internal docs here explaining the difference between `Begin` and `BeginRegion`?
</blockquote></details>

I added a comment and renamed Begin to BeginClip so it's a bit clearer what they represent.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2306)*

glennw:
Thanks, added a comment and renamed the enum variants to be a bit more descriptive.

PR #2307, opened 2018-01-16
Early return if a box shadow's attributes are zero.

demo99:
It fixes gecko [bug 1419528](https://bugzilla.mozilla.org/show_bug.cgi?id=1419528). If all attributes are zero, the box shadow rect should equal to the prim_info.rect. But after some computations, the clipping results may be slightly different. Then we see some glitches. The early return can fix this problem.

gecko try looks good: https://treeherder.mozilla.org/#/jobs?repo=try&revision=382633bc582b49f213d137c9a267371025c110b9

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2307)


glennw:
Looks good, thanks! There is one CI reftest failure - it might just need a reference image update?

demo99:
I should update the no-blur image.

glennw:
Perfect, thanks!

@bors-servo r+

PR #2308, opened 2018-01-16
Fix incorrect border radius with outset shadows.

### LINKED ISSUE #2287: Box shadows with blur and spread increase the corner radius too much
https://bugzilla.mozilla.org/show_bug.cgi?id=1429793

[Testcase](https://bug1429793.bmoattachments.org/attachment.cgi?id=8941845)

This shadow has a spread radius.

When the shadow has no blur, then the shadow's corner radius is correct. But if it does have a blur, then the corner radius is too large.

Expected vs actual results:

screen shot 2018-01-11 at 4 19 10 pm

```yaml
---
root:
  items:
    - type: stacking-context
      bounds: [0, 0, 1000, 1000]
      items:
        - type: box-shadow
          bounds: [ 0, 0, 200, 200 ]
          color: blue
          clip-mode: outset
          offset: 250 50
          blur-radius: 1
          spread-radius: 40
          border-radius:
            top-left: [0, 0]
            top-right: [120, 120]
            bottom-left: [120, 120]
            bottom-right: [0, 0]
        - type: box-shadow
          bounds: [ 0, 0, 200, 200 ]
          color: blue
          clip-mode: outset
          offset: 50 250
          blur-radius: 0
          spread-radius: 40
          border-radius:
            top-left: [0, 0]
            top-right: [120, 120]
            bottom-left: [120, 120]
            bottom-right: [0, 0]
```

nical:
Fixes #2287.

The problem was that if the blur radius/corners are big enough we blur the entire rounded rectangle. In this case we need to make sure that the blurred rectangle takes the spread amount into account (while this is not necessary when using the nine-patch code path where the rectangular patches can be stretched).
This PR adjusts the size of the rounded rectangle primitive accordingly when it makes sense.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2308)


mstange:
I think you could make a reftest for this by comparing a shadow with spread to one without spread (and a different box size + radius sizes).

mstange:
Thanks!

glennw:
Thanks!

@bors-servo r+

glennw:
@nical Looks like a CI test failure here - I haven't looked further into it.

nical:
Looks like in both cases the reference images should be updated.

PR #2311, opened 2018-01-16
Optimize box-shadow rendering (use segments for pictures).

### LINKED ISSUE #1993: Long GPU times on https://getpocket.cdn.mozilla.net/explore/trending even after #1961
[This page](https://getpocket.cdn.mozilla.net/explore/trending) scrolls very slowly even on a Firefox build that includes the patch for #1961. It has some large blur radius box-shadows.

glennw:
For my HD4600 GPU, this drops the time spent in the
many-boxshadows.html benchmark from 8.3 ms to 1.6ms. It
shows similar wins on the pocket page in Gecko in #1993.

Enable segment rendering for box-shadows (and Pictures in
general). This removes a large number of alpha-blended
pixels from outset box shadows, by cutting out segments
that have ClipOut mode.

Also modify the edge AA flags support, and how segment builder
produces these. The segment builder marks any edge that is on
the primitive edge, or cut out by a ClipOut rect as needing
AA, meaning we get smooth AA on internal (clipped out) edges.

Fixes #1993 (this page runs at ~5ms on my HD4600 now, easily
holding 60 fps).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2311)


glennw:
r? @kvark

The way that the brush / picture code interacts is a bit untidy. However, since this is such a significant performance win, I think it's probably reasonable to review / merge this, and then to the refactoring of the code as a follow up once I think about it a bit more.

I still need to do a gecko try run for this - I wouldn't be surprised if there are a few failures due to changing some of the fiddly AA-related code :)

glennw:
Gecko try seems good, I think (but could do with a sanity check to make sure I didn't miss something):

https://hg.mozilla.org/try/rev/0faba5f8ff906003d40dd7f1b4308ae8ed8897b3
https://treeherder.mozilla.org/#/jobs?repo=try&revision=0faba5f8ff906003d40dd7f1b4308ae8ed8897b3

```
9 - unrelated JS intermittent
gl3 - unexpected PASS (intermittent?)
R2 - 4 new PASS (from #2292)
R4 - 1 new FAIL (from #2285)
R8 - 3 new FAIL (from #2292)
```

glennw:
This also seems to have a good effect of a few of the other existing bugs with slow reported GPU times. But I'll run through those once this lands in Gecko and close as appropriate.

staktrace:
> Gecko try seems good, I think (but could do with a sanity check to make sure I didn't miss something):

Your assessment seems correct - the only one I'm not sure of is the gl3 unexpected PASS. It doesn't look like a pre-existing issue (there is a related bug it apprarently it was fixed already). I've retriggered that job a few more times on your try push to see if it's consistent or intermittent.

glennw:
@staktrace Yea, it does seem unlikely to be an intermittent since it passed on both the debug and release builds. Looking at the test itself, it doesn't look like it actually draws anything - I think it just creates a webgl context and queries an extension. In which case, I don't know what could have changed to make that pass... :man_shrugging:

kvark:
Good stuff!
I can see traces of https://github.com/servo/webrender/pull/2263/commits/8804cab68354572d7711cf2c48999c158d358287 here, and the new edge segmentation logic is precisely what I was missing (so hopefully will come back to that PR soon).

---

---

*[webrender/src/batch.rs, line 1323 at r1](https://reviewable.io:443/reviews/servo/webrender/2311#-L33ah3oGRq24lFuP_Tm:-L33ah3ps6zNhZOHUuiX:b-qxvvvo) ([raw file](https://github.com/servo/webrender/blob/9b939a4061b23181e4672116ecd8f4f098eed413/webrender/src/batch.rs#L1323)):*
> ```Rust
>             }
>             BrushKind::Picture => {
>                 panic!("bug: get_batch_key is handled at higher level for pictures");
> ```

I see what you mean by the ugliness of brush-picture interaction now :)

---

*[webrender/src/box_shadow.rs, line 258 at r1](https://reviewable.io:443/reviews/servo/webrender/2311#-L33anROM4AotmJLFIMK:-L33anRQkKi1Cj_39c7b:bjwa7hh) ([raw file](https://github.com/servo/webrender/blob/ef420106c1d28a7f1b342201b837d4b5f1a53831/webrender/src/box_shadow.rs#L258)):*
> ```Rust
>                     );
> 
>                     // TODO(gw): Right now, we always use a clip out
> ```

:tada: 

---

*[webrender/src/prim_store.rs, line 305 at r1](https://reviewable.io:443/reviews/servo/webrender/2311#-L33b8P-zxxlx7IbVch1:-L33b8P-zxxlx7IbVch2:b-sbt41) ([raw file](https://github.com/servo/webrender/blob/9b939a4061b23181e4672116ecd8f4f098eed413/webrender/src/prim_store.rs#L305)):*
> ```Rust
>         // has to match VECS_PER_SPECIFIC_BRUSH
>         match self.kind {
>             BrushKind::Picture => {
> ```

this doesn't match the comment on top any more
perhaps, we can call the `self.brush.write_gpu_blocks` here?

---

*[webrender/src/segment.rs, line 435 at r1](https://reviewable.io:443/reviews/servo/webrender/2311#-L33cNQd96F4BqdT6VPC:-L33bv1soumCrFqorFHz:bq9xlic) ([raw file](https://github.com/servo/webrender/blob/9b939a4061b23181e4672116ecd8f4f098eed413/webrender/src/segment.rs#L435)):*
> ```Rust
>                 let mut edge_flags = EdgeAaSegmentMask::empty();
> 
>                 if x == 0 || segments[y * x_count + x - 1].is_none() {
> ```

interesting! could probably be refactored quite a bit (bit masks, not bothering computing the `edge_flags` if the current is `None`, etc), but later

---

*[webrender/src/segment.rs, line 585 at r1](https://reviewable.io:443/reviews/servo/webrender/2311#-L33ceijIMT14EnuWsP4:-L33ceijIMT14EnuWsP5:bck8hmz) ([raw file](https://github.com/servo/webrender/blob/9b939a4061b23181e4672116ecd8f4f098eed413/webrender/src/segment.rs#L585)):*
> ```Rust
>         sb.build(|segment| {
>             segments.push(Segment {
>                 ..*segment
> ```

mmm that one seems odd. Couldn't we just do `segments.push(*segment)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2311#-:-L33d5fAUea3BM4x-7hc:bxoc9xx)*

kvark:
Ok, I don't think there is anything in my review that should hold the PR.
@bors-servo r+

staktrace:
The retriggers are consistently showing the same result for gl3. I would try doing another try push based on the latest m-c. Sometimes things like this go away with a new base revision. If it still happens we'll probably need to investigate more.

glennw:
@staktrace OK, I'll do a try run with this today and investigate further if it still passes.

glennw:
New (pending) try run with Gecko + WR master here:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=443d28841b1730fd0f9dec20700bf2ec07c6ec70

glennw:
@staktrace Looks like that fresh try run above is green on all tasks. \o/

PR #2313, opened 2018-01-17
Fix for race condition when caching render tasks.

glennw:
When the renderer receives multiple Frames from the backend
in a single render loop, it drops the old frames and just
draws the most recent frame.

If that old frame contained a render task to write to the
texture cache, then the texture cache will never be updated.

This patches tracks whether a frame has any texture cache
render tasks, and whether it has been drawn before. If a
frame which meets those conditions is being replaced, then
do a render of that frame first, skipping the main framebuffer
pass.

This is not ideal since we may end up drawing offscreen render
tasks that were batched but which aren't strictly necessary
for the texture cache render target. However, it is very rare
that the render backend is (a) producing frames faster than they
are being consumed by the renderer and (b) also contain texture
cache tasks.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2313)

### DIFF
--- a/webrender/src/debug_render.rs
+++ b/webrender/src/debug_render.rs
@@ -261,60 +261,66 @@ impl DebugRenderer {
         self.add_line(p0.x, p1.y, color, p0.x, p0.y, color);
     }
 
-    pub fn render(&mut self, device: &mut Device, viewport_size: &DeviceUintSize) {
-        device.disable_depth();
-        device.set_blend(true);
-        device.set_blend_mode_premultiplied_alpha();
+    pub fn render(
+        &mut self,
+        device: &mut Device,
+        viewport_size: Option<DeviceUintSize>,
+    ) {
+        if let Some(viewport_size) = viewport_size {
+            device.disable_depth();
+            device.set_blend(true);
+            device.set_blend_mode_premultiplied_alpha();
 
-        let projection = Transform3D::ortho(
-            0.0,
-            viewport_size.width as f32,
-            viewport_size.height as f32,
-            0.0,
-            ORTHO_NEAR_PLANE,
-            ORTHO_FAR_PLANE,
-        );
-
-        // Triangles
-        if !self.tri_vertices.is_empty() {
-            device.bind_program(&self.color_program);
-            device.set_uniforms(&self.color_program, &projection, 0);
-            device.bind_vao(&self.tri_vao);
-            device.update_vao_indices(&self.tri_vao, &self.tri_indices, VertexUsageHint::Dynamic);
-            device.update_vao_main_vertices(
-                &self.tri_vao,
-                &self.tri_vertices,
-                VertexUsageHint::Dynamic,
+            let projection = Transform3D::ortho(
+                0.0,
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

This is a rather inelegant fix - if you have a better idea of how to solve it quickly, let me know!

glennw:
@jrmuizel @staktrace I don't think it's likely you'll see this in Gecko, but if you are seeing it, and Gecko has updated past #2284 then this may explain any intermittent failures with box shadows.

webrender/src/debug_render.rs
```diff
@@ -261,60 +261,66 @@ impl DebugRenderer {
...
+                device.update_vao_main_vertices(
+                    &self.line_vao,
+                    &self.line_vertices,
+                    VertexUsageHint::Dynamic,
+                );
+                device.draw_nonindexed_lines(0, self.line_vertices.len() as i32);
+            }
+
+            // Glyph
+            if !self.font_indices.is_empty() {
```
kvark: I assume these lines are just moved

webrender/src/frame_builder.rs
```diff
@@ -1778,6 +1779,10 @@ impl FrameBuilder {
                 &self.clip_store,
                 RenderPassIndex(pass_index),
             );
+
+            if let RenderPassKind::OffScreen { ref texture_cache, .. } = pass.kind {
```
kvark: hmm, so any non-main-FBO tasks are considered for `has_texture_cache_tasks`? If the problem is only about skipping the tasks that are required for future rendering, we need a distinction between re-usable and non-reusable tasks, and only checks for the former here

webrender/src/renderer.rs
```diff
@@ -2377,7 +2377,15 @@ impl Renderer {
                     // Add a new document to the active set, expressed as a `Vec` in order
                     // to re-order based on `DocumentLayer` during rendering.
                     match self.active_documents.iter().position(|&(id, _)| id == document_id) {
-                        Some(pos) => self.active_documents[pos].1 = doc,
+                        Some(pos) => {
+                            // If the document we are replacing must be drawn
+                            // (in order to update the texture cache), issue
+                            // a render just to off-screen targets.
+                            if self.active_documents[pos].1.frame.must_be_drawn() {
+                                self.render_impl(None).ok();
```
kvark: technically, we only need to render the `pos` indexed document, not all of them.
Perhaps, we could accumulate the documents into something like `inactive_documents` vector that is flushed whenever we happen to render stuff?

webrender/src/debug_render.rs
glennw: Yep, just indented.

webrender/src/frame_builder.rs
glennw: It checks for the ```texture_cache``` target being non-empty. So this excludes temporary render tasks.

webrender/src/renderer.rs
glennw: True - would it be OK to do that as a follow up?

PR #2317, opened 2018-01-17
Remove ImageFormat::Invalid

kvark:
Also replaces #2280 

This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2317)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -130,10 +130,8 @@ pub enum UploadMethod {
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum ReadPixelsFormat {
-    R8,
+    Standard(ImageFormat),
     Rgba8,
-    Bgra8,
-    Rgba32F,
 }
 
 pub fn get_gl_format_bgra(gl: &gl::Gl) -> gl::GLuint {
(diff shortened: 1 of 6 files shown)


webrender/src/device.rs
```diff
@@ -1269,7 +1253,7 @@ impl Device {
     pub fn free_texture_storage(&mut self, texture: &mut Texture) {
         debug_assert!(self.inside_frame);
 
-        if texture.format == ImageFormat::Invalid {
+        if texture.width + texture.height == 0 {
```
glennw: This made me do a double take, but I guess it works :)

kvark: As far as I see, the only case it tries to handle is when the texture wasn't ever given contents (leaving the size to 0 and the format to `Invalid`).

kvark:
Let me know if you want the version bumped here. Otherwise this should be ready to go.

kvark:
Now that I think of it, this change is effectively a better solution to the flickering problem than the workaround I put it previously. Angle has problems with changing the format, and this PR makes us not even attempt to change the format ever.

kvark:
@bors-servo try
@glennw so I found the problem, and the solution (see last commit) may be concerning. Please take a look.

kvark:
result of IRC discussion: having a hashmap for the free lists of texture cache is not ideal but not really a performance concern, and we can proceed with it.
@bors-servo r=glennw

PR #2318, opened 2018-01-17
Remove GPU time drawing the profile from the profile time.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2318)


glennw:
r? @kvark 

Drawing the profile is quite expensive (a lot of alpha blended pixels). This was previously being included in the GPU profile totals, skewing the results.

This change makes the reported GPU time the same whether the compact or full profiler view is active.

PR #2319, opened 2018-01-17
Rename brush_image -> brush_picture.

glennw:
In preparation for porting ps_image -> brush_image and starting
to use segments for the image shader.

Eventually, we'll aim to unify brush_image and brush_picture.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2319)


glennw:
r? anyone

Issue #2297, opened 2018-01-12
[RFC] Filters: Add color transform matrix filter


eeejay:
The primary usecase would be accessibility. We are looking for a way to apply a [full-viewport color filter](https://wiki.mozilla.org/Accessibility/ColorFilters) with little performance overhead.

Besides accessibility, this can also be offered as a fast path for SVG color matrix filters (although issue #1880 would address that more holistically).

The public API change would be an extension of the FilterOp type:
```rust
#[derive(Clone, Copy, Debug, PartialEq, Deserialize, Serialize)]
pub enum FilterOp {
    Blur(f32),
    Brightness(f32),
    Contrast(f32),
    Grayscale(f32),
    HueRotate(f32),
    Invert(f32),
    Opacity(PropertyBinding<f32>, f32),
    Saturate(f32),
    Sepia(f32),
    DropShadow(LayoutVector2D, f32, ColorF),
    ColorMatrix([f32; 20]),
}
```

nical:
Sounds good to me although we should box the matrix to prevent it from making all filter ops ~88 bytes.

eeejay:
> we should box the matrix

I don't know how to do that while keeping the FilterOp copyable. Total Rust newb here.

glennw:
This happened.

PR #2323, opened 2018-01-18
Introduce color matrix filter.

eeejay:
This is a patch I came up with for issue #2297. It's not very good! This is my first Rust patch, and I know nothing of graphics or shaders. But it works? And it passes all the ref tests.

For one, I increased the FilterOp enum type to be over 80 bytes.

Second, I badly abused the GPU cache. I'm not sure how to do this. Primitives are too small for storing a whole matrix.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2323)

### DIFF
--- a/webrender/res/ps_blend.glsl
+++ b/webrender/res/ps_blend.glsl
@@ -9,6 +9,7 @@ flat varying vec4 vUvBounds;
 flat varying float vAmount;
 flat varying int vOp;
 flat varying mat4 vColorMat;
+flat varying vec4 vColorOffset;
 
 #ifdef WR_VERTEX_SHADER
 void main(void) {
@@ -50,6 +51,7 @@ void main(void) {
                              vec4(lumG - lumG * oneMinusAmount, lumG + oneMinusLumG * oneMinusAmount, lumG - lumG * oneMinusAmount, 0.0),
                              vec4(lumB - lumB * oneMinusAmount, lumB - lumB * oneMinusAmount, lumB + oneMinusLumB * oneMinusAmount, 0.0),
                              vec4(0.0, 0.0, 0.0, 1.0));
+            vColorOffset = vec4(0.0);
             break;
         }
         case 3: {
@@ -60,6 +62,7 @@ void main(void) {
                             vec4(lumG - lumG * c - lumG * s, lumG + oneMinusLumG * c + 0.140 * s, lumG - lumG * c + lumG * s, 0.0),
                             vec4(lumB - lumB * c + oneMinusLumB * s, lumB - lumB * c - 0.283 * s, lumB + oneMinusLumB * c + lumB * s, 0.0),
                             vec4(0.0, 0.0, 0.0, 1.0));
+            vColorOffset = vec4(0.0);
             break;
         }
         case 5: {
@@ -68,6 +71,7 @@ void main(void) {
                              vec4(oneMinusAmount * lumG, oneMinusAmount * lumG + vAmount, oneMinusAmount * lumG, 0.0),
                              vec4(oneMinusAmount * lumB, oneMinusAmount * lumB, oneMinusAmount * lumB + vAmount, 0.0),
                              vec4(0.0, 0.0, 0.0, 1.0));
+            vColorOffset = vec4(0.0);
             break;
         }
         case 6: {
@@ -76,7 +80,18 @@ void main(void) {
                              vec4(0.769 - 0.769 * oneMinusAmount, 0.686 + 0.314 * oneMinusAmount, 0.534 - 0.534 * oneMinusAmount, 0.0),
                              vec4(0.189 - 0.189 * oneMinusAmount, 0.168 - 0.168 * oneMinusAmount, 0.131 + 0.869 * oneMinusAmount, 0.0),
                              vec4(0.0, 0.0, 0.0, 1.0));
+            vColorOffset = vec4(0.0);
             break;
+        }
+        case 10: {
+          // Color Matrix
+          const int dataOffset = 2; // Offset in GPU cache where matrix starts
+          vec4 data[8] = fetch_from_resource_cache_8(ci.user_data2);
+
+          vColorMat = mat4(data[dataOffset], data[dataOffset + 1],
+                           data[dataOffset + 2], data[dataOffset + 3]);
+          vColorOffset = data[dataOffset + 4];
+          break;
         }
     }
 
@@ -137,7 +152,7 @@ void main(void) {
             oFragColor = Opacity(Cs, vAmount);
             break;
         default:
-            oFragColor = vColorMat * Cs;
+            oFragColor = vColorMat * Cs + vColorOffset;
     }
 
     // Pre-multiply the alpha into the output value.
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -595,9 +595,21 @@ impl PicturePrimitive {
 
     pub fn write_gpu_blocks(&self, request: &mut GpuDataRequest) {
         match self.kind {
-            PictureKind::TextShadow { .. } |
-            PictureKind::Image { .. } => {
-                request.push([0.0; 4]);
+            PictureKind::TextShadow { .. } => {
+                request.push([0.0; 4])
+            }
+            PictureKind::Image { composite_mode, .. } => {
+                match composite_mode {
+                    Some(PictureCompositeMode::Filter(FilterOp::ColorMatrix(m))) => {
+                        // When we start pushing Image pictures through the brush path
+                        // this may need to change as the number of GPU blocks written will
+                        // need to be determinate.
+                        for i in 0..5 {
+                            request.push([m[i], m[i+5], m[i+10], m[i+15]]);
+                        }
+                    },
+                    _ => request.push([0.0; 4]),
+                }
             }
             PictureKind::BoxShadow { color, .. } => {
                 request.push(color.premultiplied());
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1043,6 +1043,7 @@ impl AlphaBatcher {
                                                     FilterOp::Brightness(amount) => (7, amount),
                                                     FilterOp::Opacity(_, amount) => (8, amount),
                                                     FilterOp::DropShadow(..) => unreachable!(),
+                                                    FilterOp::ColorMatrix(_) => (10, 0.0),
                                                 };
 
                                                 let amount = (amount * 65535.0).round() as i32;
@@ -1055,7 +1056,7 @@ impl AlphaBatcher {
                                                     filter_mode,
                                                     amount,
                                                     z,
-                                                    0,
+                                                    prim_cache_address.as_int(),
                                                     0,
                                                 );
 
--- a/webrender/src/scene.rs
+++ b/webrender/src/scene.rs
@@ -171,7 +171,8 @@ impl FilterOpHelpers for FilterOp {
             FilterOp::Invert(..) |
             FilterOp::Saturate(..) |
             FilterOp::Sepia(..) |
-            FilterOp::DropShadow(..) => true,
+            FilterOp::DropShadow(..) |
+            FilterOp::ColorMatrix(..) => true,
             FilterOp::Opacity(_, amount) => {
                 amount > OPACITY_EPSILON
             }
@@ -191,6 +192,12 @@ impl FilterOpHelpers for FilterOp {
             FilterOp::Sepia(amount) => amount == 0.0,
             FilterOp::DropShadow(offset, blur, _) => {
                 offset.x == 0.0 && offset.y == 0.0 && blur == 0.0
+            },
+            FilterOp::ColorMatrix(matrix) => {
+                matrix == [1.0, 0.0, 0.0, 0.0, 0.0,
+                           0.0, 1.0, 0.0, 0.0, 0.0,
+                           0.0, 0.0, 1.0, 0.0, 0.0,
+                           0.0, 0.0, 0.0, 1.0, 0.0]
             }
         }
     }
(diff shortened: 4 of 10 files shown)


glennw:

---

*[webrender/res/ps_blend.glsl, line 24 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BLuD11L533zFElcvA:-L3BLuD11L533zFElcvB:bj57b6u) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender/res/ps_blend.glsl#L24)):*
> ```GLSL
>   ColorTransforms ctransform;
> 
>   ctransform.matrix = mat4(data[3], data[4], data[5], data[6]);
> ```

It wasn't really clear to me why this worked, starting to index from [3]. But I guess it's because the code passes the GPU cache address of the primitive, which already writes [local rect, local clip, color]. You mentioned this in the original commit, but it took me a bit to work out what was going on here :)

---

*[webrender/res/ps_blend.glsl, line 161 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BLKu2BdfNn3eirJBD:-L3BLKu2BdfNn3eirJBE:b-yr262b) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender/res/ps_blend.glsl#L161)):*
> ```GLSL
>             break;
>         case 10:
>             oFragColor = vColorTransforms.matrix * Cs + vColorTransforms.multiplier;
> ```

It seems like the multiplier is really an offset? Perhaps it should be renamed?

---

*[webrender_api/src/display_item.rs, line 505 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BJxlD4aaZ9RFEB3cU:-L3BJxlD4aaZ9RFEB3cV:bka1se8) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender_api/src/display_item.rs#L505)):*
> ```Rust
>     Sepia(f32),
>     DropShadow(LayoutVector2D, f32, ColorF),
>     ColorMatrix([f32; 20]),
> ```

As you mentioned in your commit, increasing the size here isn't ideal. It's probably not that big of a deal - since there aren't that many filter ops on most pages, but it might show up in some of the benchmarks we run that do a lot of opacity ops. I think the "proper" solution is probably to store these in the display list in a separate place, and reference them. Thoughts @Gankro?

---

*[wrench/src/yaml_helper.rs, line 565 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BJfXZJznUl3BYMeHk:-L3BJfXZJznUl3BYMeHl:b-dvclwz) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/wrench/src/yaml_helper.rs#L565)):*
> ```Rust
>                 ("color-matrix", ref args, _) if args.len() == 1 => {
>                     let mut matrix: [f32; 20] = [0.0; 20];
>                     let str = format!("---\nmatrix: {}\n", args[0]);
> ```

It seems a little unfortunate to create a new yaml doc here. Could we modify the format slightly so that we can just read it as an array of floats, perhaps?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2323)*

Gankro:
That's pretty harsh bloat, yeah. We could certainly pull it out as a fixed-size aux-list in the same way we do gradient stops. Otherwise we could add a junk None-like variant to this enum so that my deserialize_in_place optimization is making the overall size of the variant less relevant.

glennw:
Having thought about this over the weekend - I think storing it in the GPU cache as part of the Picture data (as you've done here) is basically correct. It does actually fit in well with how brush primitives work, which the blend shader will eventually be ported to.

I'll add a couple of extra comments related to that. Then, we'll need to work out the display list stuff mentioned above by @Gankro and the other minor comments.

webrender/res/ps_blend.glsl
```diff
@@ -4,13 +4,29 @@
...
 flat varying float vAmount;
 flat varying int vOp;
 flat varying mat4 vColorMat;
 
 #ifdef WR_VERTEX_SHADER
+ColorTransforms fetch_color_transforms(int address) {
+  vec4 data[8] = fetch_from_resource_cache_8(address);
+  ColorTransforms ctransform;
+
+  ctransform.matrix = mat4(data[3], data[4], data[5], data[6]);
```
glennw: I think it'd be clearer what's happening here if we pass in the modified address (offset), and then fetch the data from there.

webrender/res/ps_blend.glsl
```diff
@@ -4,13 +4,29 @@
 
 #include shared,prim_shared
 
+struct ColorTransforms {
+  mat4 matrix;
```
glennw: nit: the rest of the file uses 4-space indentation.

webrender/src/prim_store.rs
```diff
@@ -1276,6 +1276,20 @@ impl PrimitiveStore {
                     let pic = &self.cpu_pictures[metadata.cpu_prim_index.0];
                     pic.write_gpu_blocks(&mut request);
 
+                    match pic.kind {
```
glennw: Can we move this inside the ```write_gpu_blocks``` method of the ```Picture``` type?

eeejay:
Review status: 5 of 10 files reviewed at latest revision, 7 unresolved discussions.

---

*[webrender/res/ps_blend.glsl, line 24 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L39F0hn00S1ss7TmA1n-r1-24:-L3Z-a2H9OP4VAW0BRvw:bfyhowa) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender/res/ps_blend.glsl#L24)):*

I think it'd be clearer what's happening here if we pass in the modified address (offset), and then fetch the data from there.
</blockquote></details>

So have a constant "offset = 3", or pass offset as an argument to fetch_color_transforms?

---

*[webrender/res/ps_blend.glsl, line 24 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BLuD11L533zFElcvA:-L3Z-w1L8v9JcyyBYR2W:b-tdj271) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender/res/ps_blend.glsl#L24)):*

It wasn't really clear to me why this worked, starting to index from [3]. But I guess it's because the code passes the GPU cache address of the primitive, which already writes [local rect, local clip, color]. You mentioned this in the original commit, but it took me a bit to work out what was going on here :)
</blockquote></details>

Interesting! I actually figured out the offset with trial and error.

---

*[webrender/res/ps_blend.glsl, line 161 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BLKu2BdfNn3eirJBD:-L3Z03oQhjTc7tfX4giC:bt26nc3) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender/res/ps_blend.glsl#L161)):*

It seems like the multiplier is really an offset? Perhaps it should be renamed?
</blockquote></details>

That looks like a bad name to me too. I misread an article about feColorMatrix and thought that was the name of that column. I'll rename it to `offset`.

---

*[wrench/src/yaml_helper.rs, line 565 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BJfXZJznUl3BYMeHk:-L3Za4agxEPq2hRHJik2:b9sbpib) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/wrench/src/yaml_helper.rs#L565)):*

It seems a little unfortunate to create a new yaml doc here. Could we modify the format slightly so that we can just read it as an array of floats, perhaps?
</blockquote></details>

Yeah. Made it a function that takes 20 arguments :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2323)*

eeejay:
Review status: 5 of 10 files reviewed at latest revision, 7 unresolved discussions.

---

*[webrender/res/ps_blend.glsl, line 8 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L39F0hn00S1ss7TmA1n-r1-8:-L3_2t-brn1d-e_GUglm:b-896fix) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender/res/ps_blend.glsl#L8)):*

nit: the rest of the file uses 4-space indentation.
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1279 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L39F0hn00S1ss7TmA1p-r1-1279:-L3_2qBcPmebV2FyxnKH:b-896fix) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender/src/prim_store.rs#L1279)):*

Can we move this inside the ```write_gpu_blocks``` method of the ```Picture``` type?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2323)*

eeejay:
Review status: 5 of 10 files reviewed at latest revision, 7 unresolved discussions.

---

*[webrender_api/src/display_item.rs, line 505 at r1](https://reviewable.io:443/reviews/servo/webrender/2323#-L3BJxlD4aaZ9RFEB3cU:-L3_3G_T0yMKBb7ss8K0:b-8ro01v) ([raw file](https://github.com/servo/webrender/blob/793516876fd3528e88777964b82e415a52aed850/webrender_api/src/display_item.rs#L505)):*

As you mentioned in your commit, increasing the size here isn't ideal. It's probably not that big of a deal - since there aren't that many filter ops on most pages, but it might show up in some of the benchmarks we run that do a lot of opacity ops. I think the "proper" solution is probably to store these in the display list in a separate place, and reference them. Thoughts @Gankro?
</blockquote></details>

I'm lost in the display list serialization code, and I'm not sure how to do this...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2323)*

glennw:
@Gankro Can you offer any tips for implementing the display list changes above?

Gankro:
Sorry I was out with flu the last week. I want to spend a bit of today thinking over aux list stuff again since kvark has also been having trouble with this. Hopefully will get back to you tonight.

Gankro:
Ok so having looked at this, I'd like to proceed with the changes to FilterOp as-is, and I'll do a followup PR to optimize our aux-list processing (or not if we measure no impact).

The change I'm picturing is fairly straight-forward for me to implement.

eeejay:
Sounds good! I'm not sure how to use `reviewable.io`. But I think this pull request is ready for another round of reviews. I am PTO starting tomorrow till February 12th, so I'll be slow on followup.

---

Review status: 5 of 10 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2323#-:-L47vG6Q0Ipsu95xMfGi:bjknqgn)*

glennw:
@eeejay Sounds good, thanks! I'll do another review pass over this today.

webrender/res/ps_blend.glsl
```diff
@@ -4,13 +4,31 @@
...
+  vec4 offset;
+};
+
+varying ColorTransforms vColorTransform;
```
glennw: We've had issues in the past with driver bugs when a struct is passed as a varying, unfortunately. See https://github.com/servo/webrender/wiki/Driver-issues#1157---varying-structs for details. The workaround is to only pass GLSL primitive types as varyings.

webrender/src/picture.rs
```diff
@@ -596,9 +596,18 @@ impl PicturePrimitive {
 
     pub fn write_gpu_blocks(&self, request: &mut GpuDataRequest) {
         match self.kind {
-            PictureKind::TextShadow { .. } |
-            PictureKind::Image { .. } => {
-                request.push([0.0; 4]);
+            PictureKind::TextShadow { .. } => {
```
glennw: There are some code paths that rely on knowing the number of GPU blocks that a primitive writes (the brush / segment code). In this case, it's actually fine, since ```PictureKind::Image``` never passes through that path. However, it may do in the future. Could you add a comment here noting that this code will need to change if/when we start pushing Image pictures through the brush path?

eeejay:
Followed up with the nits you mentioned. Hopefully this is mergable, if not I'll be back on the 12th!

---

Review status: 3 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[webrender/res/ps_blend.glsl, line 12 at r2](https://reviewable.io:443/reviews/servo/webrender/2323#-L39F0hn00S1ss7TmA1n-r2-12:-L4CHCsfr8LrOq0KfgrO:b-1efih) ([raw file](https://github.com/servo/webrender/blob/6e02ea899d85a8e73df918d964c2bb0ab1b5f03c/webrender/res/ps_blend.glsl#L12)):*

We've had issues in the past with driver bugs when a struct is passed as a varying, unfortunately. See https://github.com/servo/webrender/wiki/Driver-issues#1157---varying-structs for details. The workaround is to only pass GLSL primitive types as varyings.
 
</blockquote></details>

Removing the structure and function.

---

*[webrender/src/picture.rs, line 599 at r2](https://reviewable.io:443/reviews/servo/webrender/2323#-L3_2-ne_h1hV6Lk-EY1-r2-599:-L4CHOBz8wfLRUDqidKp:beqof5b) ([raw file](https://github.com/servo/webrender/blob/6e02ea899d85a8e73df918d964c2bb0ab1b5f03c/webrender/src/picture.rs#L599)):*

There are some code paths that rely on knowing the number of GPU blocks that a primitive writes (the brush / segment code). In this case, it's actually fine, since ```PictureKind::Image``` never passes through that path. However, it may do in the future. Could you add a comment here noting that this code will need to change if/when we start pushing Image pictures through the brush path?
</blockquote></details>

Cool. I'll add a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2323#-:-L4CHTq33TEZqIs174Q0:b2dnym)*

glennw:
Thank you!

@bors-servo r+

PR #2326, opened 2018-01-19
Capture infrastructure, part 2.5

kvark:
The PR rewrites the way capturing interacts with external images and buffers. For the matter of consistency with the frame-level capturing, and respecting the given UV ranges, the external stuff now goes through an extra level of indirection:
  - Old: "scene-1-0.ron" -> "blobs/4.raw"
  - New: "scene-1-0.ron" -> "externals/4.ron" -> "externals/t2.raw"

This ~~is a WIP for early feedback about the refactor/API changes,~~ comes with absolutely no warranty! Ultimately, the PR ~~will be ready when replaying a capture from Gecko works~~ edit: is ready 🎉 .

Public API changes:
  - `TexelRect` is exposed and used in `ExternalImage`
  - `TextureTarget` is exposed and used in `ExternalImageType`
  - `wrench` compact profiler key is now "S" (same as the examples use), while "C" is reserved for capturing

The PR also contains a bunch of internal refactoring to drop some of the technological debt accumulated, including but not limited to:
  - use of strongly typed `RectWithEndpoint` in `ImageResource` GLSL code
  - private contents of `GpuBlockData`, better conversions to it
  - more use of `Self` in method returns, plus some associated constants
  - simplification of a quite a bit of conversion code thanks to the public API changes ^
  - wrench window size handling is streamlined a bit, it's also getting re-sized on capture load

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2326)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,7 +105,7 @@ name = "cgl"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "gleam 0.4.19 (registry+https://github.com/rust-lang/crates.io-index)",
+ "gleam 0.4.20 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.20 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -371,7 +371,7 @@ dependencies = [
 
 [[package]]
 name = "gleam"
-version = "0.4.19"
+version = "0.4.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "gl_generator 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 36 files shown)


kvark:
The main problem I had with Gecko was caused by the fact I used to issue a capture from the Browser Console, which is in turn a window also using WR, so all the captured data was messed up. When capturing with a key combination (thanks to @mstange for help setting this up), wrench shows the captured FF frame just fine:
ff-wrench

With this in, the PR is ready for review. I'm going to follow up with the FF patch and a gecko try push before this is merged.

glennw:
Reviewed 36 of 36 files at r1.

---

*[webrender/src/capture.rs, line 72 at r1](https://reviewable.io:443/reviews/servo/webrender/2326#-L3U_-98SR5Lm9xeiP_d:-L3U_-99hd17afhoZ_9t:b-tm7n1q) ([raw file](https://github.com/servo/webrender/blob/f1aabc2c2a0aaf3f92ead423886089624514670c/webrender/src/capture.rs#L72)):*
> ```Rust
> 
>         let color_type = match format {
>             ImageFormat::BGRA8 => ColorType::RGBA, //TODO?
> ```

Could you expand on what the TODO here means?

---

*[webrender/src/capture.rs, line 104 at r1](https://reviewable.io:443/reviews/servo/webrender/2326#-L3U_7aQOZ3LRsnSUheD:-L3U_7aQOZ3LRsnSUheE:bcs446c) ([raw file](https://github.com/servo/webrender/blob/f1aabc2c2a0aaf3f92ead423886089624514670c/webrender/src/capture.rs#L104)):*
> ```Rust
>     pub data: String,
>     pub id: ExternalImageId,
>     pub channel_index: u8,
> ```

What does channel_index refer to here?

---

*[webrender/src/renderer.rs, line 494 at r1](https://reviewable.io:443/reviews/servo/webrender/2326#-L3UapKlx0z2LaFmqV3g:-L3UapKm9vLopHg7AY5c:b-qrpxq8) ([raw file](https://github.com/servo/webrender/blob/f1aabc2c2a0aaf3f92ead423886089624514670c/webrender/src/renderer.rs#L494)):*
> ```Rust
> }
> 
> //TODO: those types are the same, can we merge them?
> ```

I think so, yep (doesn't need to be in this PR).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2326)*

glennw:
@kvark Looks good! r=me once those minor comments are addressed, and try is green.

kvark:
@glennw thanks for taking a look! The try push shows some JS errors that may be related to the JS code I put in (unlikely though) but have nothing to do with WR side of things, so I consider it green for the matter of this PR (unless @staktrace says otherwise ;) ).

staktrace:
The try push looks good to me

kvark:
Please not that this is not the end of story. I noticed some external images not showing up in Gecko captures, so I'll follow up with more fixes. This PR main goal is to have the Gecko integration finally set up, and I'm sure it's helpful in the current shape already.

PR #2329, opened 2018-01-22
Don't skip SetGradientStops

### LINKED ISSUE #2328: SetGradientStops missing with debug-serialization
Hi,
I found that serializing a BuiltDisplayList with the "debug-serialization" feature omits the `SetGradientStops` items.

BTW I am curious to hear why the gradient stops are separate items instead of being part of gradient and radial gradient.

* [serialized display list](https://gist.github.com/pyfisch/29af74e0ebffc2d6287ef806143c3cd4)
* Rendering: simple-gradient

cc @kvark

kvark:
Fixes #2328
r? @Gankro

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2329)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -5,7 +5,7 @@
 
 use api::{BuiltDisplayListIter, ClipAndScrollInfo, ClipId, ColorF, ComplexClipRegion};
 use api::{DevicePixelScale, DeviceUintRect, DeviceUintSize};
-use api::{DisplayItemRef, DocumentLayer, Epoch, FilterOp};
+use api::{DisplayItemRef, DocumentLayer, Epoch, FilterOp, GradientStop};
 use api::{ImageDisplayItem, ItemRange, LayerPoint, LayerPrimitiveInfo, LayerRect};
 use api::{LayerSize, LayerVector2D, LayoutSize};
 use api::{LocalClip, PipelineId, ScrollClamping, ScrollEventPhase, ScrollLayerState};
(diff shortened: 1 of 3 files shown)


kvark:
Note: WR frame building is already happy skipping this:
```rust
            // Do nothing; these are dummy items for the display list parser
            SpecificDisplayItem::SetGradientStops => {}
```

pyfisch:
Not sure why but with this PR servo crashes for me because of `Option::unwrap`. https://gist.github.com/pyfisch/8ff4ff4a476152aeb644a340141dc1ca

Gankro:
I'm not familiar with the debug-serializing feature, but one option would be to split `next` into two methods: a "raw" one, and a wrapper that consumes the raw one and behaves like the current next. Then anyone who needs to reproduce the input can just use the raw one?

kvark:
@Gankro how about we don't hack thing as low as traversing items and instead put the gradient stop logic in the context flattening? That would make all the low-level things like debug serialization robust. Please take a look at the code in this PR now.

Gankro:
Presumably the same logic would need to be used in the DL->yaml/json writers? (I don't think we have tests for that)

kvark:
@Gankro good point! I fixed the YAML writer to take those into account and tested it by saving/loading a thing. Appears to work fine.
As to JSON/RON - these folks use debug serialization, and they don't need to be loaded (1) afterwards, so not only they don't require extra fixes, but they are also already improved :)

(1) actually, the capture infrastructure does need RON to be loaded, and with gradient stops we can just have better capturing of gradients on the scene level.

glennw:
@Gankro @kvark Is this one ready to go?

glennw:
ping @Gankro @kvark

kvark:
Marking as blocked, decided to re-think gradient stops more fundamentally with @Gankro .

pyfisch:
Are there now concrete plans for improved gradient stops? If not what are the open questions?

kvark:
@pyfisch last time I heard @Gankro was looking into it

glennw:
@Gankro Should this be kept open?

Gankro:
Let's close this PR in favour of an issue (sorry for letting this slip!)

PR #2330, opened 2018-01-22
Don't re-draw texture cache targets if frame already rendered.

glennw:
This is a minor optimization (no need to draw the texture cache
target tasks if they have already been drawn), but it mostly
makes debugging easier to see where the texture cache update
is occurring.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2330)


glennw:
r? @kvark 

We need a better solution to this long term, that can also skip the sub-passes associated with texture cache targets, but this is an improvement for now.

kvark:
Looks sane
@bors-servo r+

PR #2331, opened 2018-01-23
Handle image sub-rects by pre-rendering to task cache.

glennw:
This removes any shader related code for handling sub-rects
of images. This is a very small optimization by itself - the
primary goal of this PR is to introduce the infrastructure
that can be used for pre-rendering images.

In the future, we'll aim to make use of this infrastructure
for at least two other cases:

 * Pre-render spacing and repeat sequences of images in
   some situations (when the number of instances generated
   would otherwise be large). This will give a significant
   performance improvement to the image shader, which is
   typically the second most expensive shader on most
   pages. It's also another step to porting the image shader
   to a brush shader, and collapsing it into other shaders
   such as brush_image.

 * Pre-render mip-maps of very large images into a smaller
   surface. This will allow fixing some image quality bugs
   where we sample from a very large texture (typically a
   large alpha mask). It is likely to also help with
   performance, since the primitive will then be sampling
   from a much smaller texture surface.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2331)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -727,12 +727,11 @@ TextRun fetch_text_run(int address) {
 struct Image {
     vec4 stretch_size_and_tile_spacing;  // Size of the actual image and amount of space between
                                          //     tiled instances of this image.
-    vec4 sub_rect;                          // If negative, ignored.
 };
 
 Image fetch_image(int address) {
-    vec4 data[2] = fetch_from_resource_cache_2(address);
-    return Image(data[0], data[1]);
+    vec4 data = fetch_from_resource_cache_1(address);
+    return Image(data);
 }
 
 void write_clip(vec2 global_pos, ClipArea area) {
--- a/webrender/res/ps_image.glsl
+++ b/webrender/res/ps_image.glsl
@@ -50,15 +50,8 @@ void main(void) {
     vec2 texture_size_normalization_factor = vec2(textureSize(sColor0, 0));
 #endif
 
-    vec2 uv0, uv1;
-
-    if (image.sub_rect.x < 0.0) {
-        uv0 = res.uv_rect.p0;
-        uv1 = res.uv_rect.p1;
-    } else {
-        uv0 = res.uv_rect.p0 + image.sub_rect.xy;
-        uv1 = res.uv_rect.p0 + image.sub_rect.zw;
-    }
+    vec2 uv0 = res.uv_rect.p0;
+    vec2 uv1 = res.uv_rect.p1;
 
     // vUv will contain how many times this image has wrapped around the image size.
     vec2 st0 = uv0 / texture_size_normalization_factor;
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{AlphaType, DeviceIntRect, DeviceIntSize, ImageKey, LayerToWorldScale};
-use api::{ExternalImageType, FilterOp, ImageRendering, LayerRect};
+use api::{DeviceUintRect, DeviceUintPoint, DeviceUintSize, ExternalImageType, FilterOp, ImageRendering, LayerRect};
 use api::{SubpixelDirection, TileOffset, YuvColorSpace, YuvFormat};
 use api::{LayerToWorldTransform, WorldPixel};
 use border::{BorderCornerInstance, BorderCornerSide, BorderEdgeKind};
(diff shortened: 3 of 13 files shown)


glennw:
r? @kvark 

Still working on getting a try run - will post results here once I get a successfully patched gecko :)

glennw:
Hmm, it's not clear to me how this could cause a failure in ```shadow-ordering``` on OS X, since that doesn't have any images. It might just be a CI weirdness. I'll take a closer look tomorrow.

glennw:
Gecko try:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=44455708bdaa22325eaa9b2f6470a12a2de8e09c

There is one failure, which I suspect is an unrelated intermittent. It doesn't have any images in it, from what I can see. I can re-kick that in the morning if we want to be safe.

kvark:
Reviewed 12 of 12 files at r1.

---

*[webrender/src/prim_store.rs, line 1225 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZWBCrTRIi7y6EZwfI:-L3ZWBCrTRIi7y6EZwfJ:b-an0nj1) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/prim_store.rs#L1225)):*
> ```Rust
>                                     // Cache key to identify this image portion.
>                                     key: ImageCacheKey {
>                                         image_key: image_cpu.image_key,
> ```

there is quite a bit of redundancy now in the image CPU: texel_rect is tightly bound the `ImageSource::Cache` (but not a part of it?), and the latter copies fields over

---

*[webrender/src/prim_store.rs, line 1228 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZW7d2rQQiedIjt1Hs:-L3ZW7d3nGgVwBXtCgcW:b-ushjo0) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/prim_store.rs#L1228)):*
> ```Rust
>                                         image_rendering: image_cpu.image_rendering,
>                                         tile_offset: image_cpu.tile_offset,
>                                         texel_rect: image_cpu.texel_rect.map(|texel_rect| {
> ```

at this point we already know `texel_rect`

---

*[webrender/src/prim_store.rs, line 1250 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZWjXabx-gaRSHeESS:-L3ZWjXabx-gaRSHeEST:b7fi6pr) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/prim_store.rs#L1250)):*
> ```Rust
>                     // TODO(gw): Don't actually need this in cached source mode if
>                     //           the cache item is still valid...
>                     resource_cache.request_image(
> ```

if we move this up a little, we'd be able to merge the two code blocks working with `ImageSource::Cache`

---

*[webrender/src/renderer.rs, line 83 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZV4_Y73GdETwPFthr:-L3ZV4_Y73GdETwPFths:b-fjej7r) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/renderer.rs#L83)):*
> ```Rust
> 
> /// Number of GPU blocks per UV rectangle provided for an image.
> pub const BLOCKS_PER_UV_RECT: usize = 2;
> ```

This should be changed to `1` now

---

*[webrender/src/renderer.rs, line 3339 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZXSYUolsH3vyyOADu:-L3ZXSYUolsH3vyyOADv:byezb4f) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/renderer.rs#L3339)):*
> ```Rust
>         render_tasks: &RenderTaskTree,
>     ) {
>         if !blits.is_empty() {
> ```

let's early return instead of intending the whole function?

---

*[webrender/src/tiling.rs, line 381 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZYn9JiZSs3i2HQxn6:-L3ZYn9JiZSs3i2HQxn7:b-96t3x) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/tiling.rs#L381)):*
> ```Rust
>                     BlitSource::Image { image_key, image_rendering, tile_offset, sub_rect } => {
>                         // Get the cache item for the source texture.
>                         let cache_item = resolve_image(
> ```

since we've already resolved it during the `prepare_prim_for_render`, could we store the "cache_item" right away instead of resolving again here during the task population? The need to have `deferred_resolves` and mutable `gpu_cache` feels a bit out of place here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2331)*

glennw:
Review status: 11 of 12 files reviewed at latest revision, 6 unresolved discussions.

---

*[webrender/src/renderer.rs, line 83 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZV4_Y73GdETwPFthr:-L3Zja7HhWdNQH1AiYiM:b-e6qbzh) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/renderer.rs#L83)):*

This should be changed to `1` now
</blockquote></details>

I don't think this should be changed - this is the blocks for a ```uv_rect_handle``` which is still the same (an ImageResource in shader terms)

---

*[webrender/src/renderer.rs, line 3339 at r1](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZXSYUolsH3vyyOADu:-L3Zjnz4e0SQf-LVOnGR:b-896fix) ([raw file](https://github.com/servo/webrender/blob/f37baeb8b8d78c4709d3f0391dccf612988c7423/webrender/src/renderer.rs#L3339)):*

let's early return instead of intending the whole function?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2331)*

kvark:
Thanks for addressing the concerns!
:lgtm: 

---

Reviewed 1 of 1 files at r2, 5 of 5 files at r3.

---

*[webrender/src/frame_builder.rs, line 1447 at r3](https://reviewable.io:443/reviews/servo/webrender/2331#-L3Zs35wJ0ywRgzg_xqa:-L3Zs35wJ0ywRgzg_xqb:b-7m74ox) ([raw file](https://github.com/servo/webrender/blob/a13a0a19d47d86402a9582f88a49c747cbff9e0e/webrender/src/frame_builder.rs#L1447)):*
> ```Rust
>             current_epoch: Epoch::invalid(),
>             source: ImageSource::Default,
>             key: ImageCacheKey {
> ```

nice!

---

*[webrender/src/render_task.rs, line 270 at r3](https://reviewable.io:443/reviews/servo/webrender/2331#-L3ZsBHCLm2GWssPntLQ:-L3ZsBHCLm2GWssPntLR:b-boe2u) ([raw file](https://github.com/servo/webrender/blob/a13a0a19d47d86402a9582f88a49c747cbff9e0e/webrender/src/render_task.rs#L270)):*
> ```Rust
> #[cfg_attr(feature = "capture", derive(Deserialize, Serialize))]
> pub enum BlitSource {
>     Image {
> ```

nit: now that the variants are trivial, we don't need the named fields

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2331#-:-L3ZsJj9Oah4rU9Y3LKR:b5y5xc)*

glennw:
@bors-servo r=kvark

If there's any issues with disabling that shadow ordering test, we can sort that out as a follow up.

Issue #2099, opened 2017-11-24
Make hit-testing more robust to timing irregularities


staktrace:
I've been working on a gecko mochitest to exercise the WR hit-testing API, and I've been getting intermittent failures. I'm not 100% sure about this yet but it seems like these failures happen when the hit-test message arrives at the render_backend between scene being built and the clip_scroll_tree being rebuilt. Conceptually this makes sense because it means WR might be in an inconsistent state as it has updated half of its data structures but not the other half. I'll add some more logging to confirm but assuming this is the case, does anybody have any thoughts on how to avoid this problem?

glennw:
From a quick look at the source code, we call ```clip_scroll_tree.update_tree``` when building a frame for render, but not when doing a hit test. This would suggest your analysis is right - that it can be out of sync when a new scene arrives but before a render.

If that's the case, calling ```clip_scroll_tree.update_tree``` at the start of hit test - and probably adding a dirty tracker to ```update_tree``` so we don't update it again when we then go to render if not dirty might fix it?

Thoughts @mrobinson ?

mrobinson:
Hrm. I wonder if the right thing to do here is to wait to update the hit test item list until we call `clip_scroll_tree.update_tree`. I will take a look at this tomorrow!

mrobinson:
This is actually a bit tricky! Perhaps there's a clever way to handle this though. Ideally, hit testing should be done against the currently rendered frame. As far as I can tell though, the ClipScrollTree and the ClipStore are not retained after a `FrameBuilder::build(...)` is called. The ClipScrollTree can almost certainly be replaced by the vector of ClipScrollNodeData, but I'm still not sure how to handle new data in the ClipStore.

staktrace:
@nical Does your transaction stuff affect this at all? I thought we had a design that would allow hit-testing to work against currently rendered frames while we are in the process of building the next frame. (I have some vague recollection of this from that google doc you made)

nical:
I think it does. With the transactions, anything that is part of a transaction will be applied atomically in the sense that we don't go back to the render backend event loop in between the things of a given transaction.
I'll finish the transaction work in January since it's not in the P2/P1 list at the moment.

staktrace:
Bugzilla bug to go with this: https://bugzilla.mozilla.org/show_bug.cgi?id=1425451

staktrace:
[...]

PR #2333, opened 2018-01-23
Make WR hit-testing produce consistent results.

staktrace:
Ideally we would want the hit-test to operate against the previously
rendered frame. However, once a SetDisplayList call happens, the clip
scroll tree for the previouly rendered frame is clobbered, and so the
hit-test operates on inconsistent data and therefore produces
inconsistent results. This patch is a temporary workaround that triggers
a new render if needed in this scenario, so that at least the hit-test
produces consistent results.

Workaround for #2099

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2333)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -77,6 +77,11 @@ struct Document {
     // the first frame would produce inconsistent rendering results, because
     // scroll events are not necessarily received in deterministic order.
     render_on_scroll: Option<bool>,
+    // A helper flag to prevent any hit-tests from happening between calls
+    // to build_scene and rendering the document. In between these two calls,
+    // hit-tests produce inconsistent results because the clip_scroll_tree
+    // is out of sync with the display list.
+    render_on_hittest: bool,
 }
 
 impl Document {
@@ -105,8 +110,9 @@ impl Document {
             },
             frame_ctx: FrameContext::new(config),
             frame_builder: Some(FrameBuilder::empty()),
-            render_on_scroll,
             output_pipelines: FastHashSet::default(),
+            render_on_scroll,
+            render_on_hittest: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


webrender/src/render_backend.rs
```diff
@@ -77,6 +77,11 @@ struct Document {
     // the first frame would produce inconsistent rendering results, because
     // scroll events are not necessarily received in deterministic order.
     render_on_scroll: Option<bool>,
+    // A helper flag to prevent any hit-tests from happening between calls
+    // to build_scene and rendering the document. In between these two calls,
+    // hit-tests produce inconsistent results because the clip_scroll_tree
+    // is out of sync with the display list.
```
kvark: this is a little bit worrying, could we get a little more details on what's going on there and why?

staktrace: Does the comment at https://github.com/servo/webrender/issues/2099#issuecomment-347161775 help? I'm not sure what kind of details you're looking for exactly.

webrender/src/render_backend.rs
```diff
@@ -403,6 +409,15 @@ impl RenderBackend {
             }
             DocumentMsg::HitTest(pipeline_id, point, flags, tx) => {
                 profile_scope!("HitTest");
+                if doc.render_on_hittest {
```
kvark: there are two problems with this bit:
  1. the result is not going anywhere (i.e. `Renderer` doesn't get it), which is untypical and may go across some of the implicit assumptions we have in the code
  2. nothing should be rendered inside the `DocumentMsg` arms. Instead, we return the proper `DocumentOps` struct.

staktrace: 2. is incompatible with the intent of the code, which requires that the render be done prior to the hit_test. If we do the render via the DocumentOps return, that will happen after the hit test is carried out. Any suggestions on how to accomplish this?

PR #2343, opened 2018-01-25
Move hit-test commands to be processed after document rendering.

staktrace:
This adds a new field to DocumentOp, which is a "query" that
update_document is meant to run after doing updates to the document.
HitTest commands now just trigger a render (if needed) but defer the
real work into a query. This ensures the hit-test is still run after the
document is rendered, but is more in line with the structure of the
current code, and does a better job of not violating implicit
assumptions about document rendering/publishing.

This fixes bustage from #2333.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2343)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -6,9 +6,9 @@ use api::{ApiMsg, BlobImageRenderer, BuiltDisplayList, DebugCommand, DeviceIntPo
 #[cfg(feature = "debugger")]
 use api::{BuiltDisplayListIter, SpecificDisplayItem};
 use api::{DevicePixelScale, DeviceUintPoint, DeviceUintRect, DeviceUintSize};
-use api::{DocumentId, DocumentLayer, DocumentMsg};
-use api::{IdNamespace, PipelineId, RenderNotifier};
-use api::channel::{MsgReceiver, PayloadReceiver, PayloadReceiverHelperMethods};
+use api::{DocumentId, DocumentLayer, DocumentMsg, HitTestFlags, HitTestResult};
+use api::{IdNamespace, PipelineId, RenderNotifier, WorldPoint};
+use api::channel::{MsgReceiver, MsgSender, PayloadReceiver, PayloadReceiverHelperMethods};
 use api::channel::{PayloadSender, PayloadSenderHelperMethods};
 #[cfg(feature = "capture")]
 use api::CapturedDocument;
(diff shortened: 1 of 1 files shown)


mstange:
So this fixes the flickering? Do we know why, other than "some assumptions were violated"?

staktrace:
It seems to fix the flickering for me. I don't really know which assumptions were being violated, maybe kvark has a better idea?

webrender/src/render_backend.rs
```diff
@@ -402,29 +405,21 @@ impl RenderBackend {
...
                     render: should_render,
+                    ..DocumentOps::nop()
                 }
             }
             DocumentMsg::HitTest(pipeline_id, point, flags, tx) => {
-                profile_scope!("HitTest");
-
+                let mut op = DocumentOps {
+                    queries: vec![DocumentMsg::HitTest(pipeline_id, point, flags, tx)],
+                    ..DocumentOps::nop()
```
kvark: nit: could have `render: doc.render_on_hittest`

webrender/src/render_backend.rs
```diff
@@ -484,6 +479,27 @@ impl RenderBackend {
...
 
+    fn query_document(
+        &mut self,
+        document_id: DocumentId,
+        message: DocumentMsg,
+    ) {
+        let doc = self.documents.get_mut(&document_id).unwrap();
+
+        match message {
+            DocumentMsg::HitTest(pipeline_id, point, flags, tx) => {
```
kvark: let's store the fields of this in a tuple instead of matching the enum again

webrender/src/render_backend.rs
```diff
@@ -484,6 +479,27 @@ impl RenderBackend {
         }
     }
 
+    fn query_document(
+        &mut self,
+        document_id: DocumentId,
+        message: DocumentMsg,
+    ) {
+        let doc = self.documents.get_mut(&document_id).unwrap();
```
kvark: I think we'd end up with cleaner code if we fully inline this function (no need for separate function here)

kvark:
> It seems to fix the flickering for me. I don't really know which assumptions were being violated, maybe kvark has a better idea?

I have an idea. `tiling::Frame`, being a product of scene building carries an important field:
```rust
    // List of updates that need to be pushed to the
    // gpu resource cache.
    pub gpu_cache_updates: Option<GpuCacheUpdateList>,
```
Applying these cache updates is basically synchronizing the RB and Renderer with regards to the GPU cache layout. Dropping those means the Renderer gets de-synchronized, and we end up using older/trashy GPU blocks for actual rendering. For some primitives, that results in geometry changes, for others - texture coordinates, properties, etc.

kvark:
Let's get this fix in (after a short cleanup), and I'll think about better ways to enforce the synchronization.

staktrace:
Patch updated per review comments, let me know if I did it right. I wasn't quite sure I understood correctly what you meant when you said to use a tuple instead. Is it just for performance reasons?

kvark:
> Is it just for performance reasons?

No, just to avoid a redundant match.

@bors-servo r+

PR #2344, opened 2018-01-26
Track GPU cache frame ID

kvark:
The PR separates frame rendering from GPU cache updates and enforces their order at run time.
It's supposedly also fixing #2333, but in a different (complimentary!) way from #2343.
cc @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2344)


staktrace:
I verified that this patch also seems to fix the flickering issue.

PR #2346, opened 2018-01-26
Remove old yaml files

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2346)


staktrace:
For context, this is so that we can start including wrench source code in m-c (so we can easily build a version of wrench matching a version of firefox) without unnecessary repo bloat. We'll remove the reftests/, benchmarks/, and examples/ folders when copying wrench into m-c.

kvark:
Justification: these haven't been CI tested, and we already have an extensive use-case coverage with reftests.
@bors-servo r+

PR #2347, opened 2018-01-26
Track removed pipelines and expose them in the renderer.

### LINKED ISSUE #2345: Get notifications when pipelines are removed.
We use epochs (see `Renderer::flush_rendered_epochs`) to track where the renderer is at and manage the lifetime of resources that are used by the renderer but owned by another thread. This works well except when we remove a pipeline, because we never get the notification for the last epoch of that pipeline and this results in leaks in gecko as we keep waiting for a rendered epoch notification that never comes.
Having special code to handle this in gecko is tricky because knowing that a pipeline is removed at the api level doesn't say anything about when it will not be used by the renderer anymore.

We need an equivalent of `flush_rendered_epochs` that signals all of the removed pipelines.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1432708

nical:
This information is useful to manage the lifetime of resources that are owned by the embedder and used by the renderer such as external images.

Fixes #2345.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2347)

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -314,6 +314,8 @@ pub fn main_wrapper<E: Example>(
 
         renderer.update();
         renderer.render(framebuffer_size).unwrap();
+        let _ = renderer.flush_rendered_epochs();
+        let _ = renderer.flush_removed_pipelines();
         example.draw_custom(&*gl);
         window.swap_buffers().ok();
     }
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -314,8 +314,7 @@ pub fn main_wrapper<E: Example>(
 
         renderer.update();
         renderer.render(framebuffer_size).unwrap();
-        let _ = renderer.flush_rendered_epochs();
-        let _ = renderer.flush_removed_pipelines();
+        let _ = renderer.flush_pipeline_info();
         example.draw_custom(&*gl);
         window.swap_buffers().ok();
     }
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -22,6 +22,7 @@ use profiler::{GpuCacheProfileCounters, TextureCacheProfileCounters};
 use resource_cache::{FontInstanceMap,ResourceCache, TiledImageMap};
 use scene::{Scene, StackingContextHelpers, ScenePipeline, SceneProperties};
 use tiling::{CompositeOps, Frame};
+use renderer::PipelineInfo;
 use std::mem::replace;
 
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug, Eq, Ord)]
@@ -1117,8 +1118,10 @@ impl FrameContext {
     pub fn make_rendered_document(&mut self, frame: Frame) -> RenderedDocument {
         let nodes_bouncing_back = self.clip_scroll_tree.collect_nodes_bouncing_back();
         RenderedDocument::new(
-            self.pipeline_epoch_map.clone(),
-            replace(&mut self.removed_pipelines, Vec::new()),
+            PipelineInfo {
+                epochs: self.pipeline_epoch_map.clone(),
+                removed_pipelines: replace(&mut self.removed_pipelines, Vec::new()),
+            },
             nodes_bouncing_back,
             frame
         )
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -22,6 +22,7 @@ use profiler::{GpuCacheProfileCounters, TextureCacheProfileCounters};
 use resource_cache::{FontInstanceMap,ResourceCache, TiledImageMap};
 use scene::{Scene, StackingContextHelpers, ScenePipeline, SceneProperties};
 use tiling::{CompositeOps, Frame};
+use std::mem::replace;
 
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug, Eq, Ord)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
@@ -958,6 +959,7 @@ pub struct FrameContext {
     window_size: DeviceUintSize,
     clip_scroll_tree: ClipScrollTree,
     pipeline_epoch_map: FastHashMap<PipelineId, Epoch>,
+    removed_pipelines: Vec<PipelineId>,
     id: FrameId,
     pub frame_builder_config: FrameBuilderConfig,
 }
@@ -967,6 +969,7 @@ impl FrameContext {
         FrameContext {
             window_size: DeviceUintSize::zero(),
             pipeline_epoch_map: FastHashMap::default(),
+            removed_pipelines: Vec::new(),
             clip_scroll_tree: ClipScrollTree::new(),
             id: FrameId(0),
             frame_builder_config: config,
@@ -1023,7 +1026,7 @@ impl FrameContext {
     pub fn create_frame_builder(
         &mut self,
         old_builder: FrameBuilder,
-        scene: &Scene,
+        scene: &mut Scene,
         resource_cache: &mut ResourceCache,
         window_size: DeviceUintSize,
         inner_rect: DeviceUintRect,
@@ -1101,6 +1104,9 @@ impl FrameContext {
 
         self.clip_scroll_tree
             .finalize_and_apply_pending_scroll_offsets(old_scrolling_states);
+
+        self.removed_pipelines.extend(scene.removed_pipelines.drain(..));
+
         frame_builder
     }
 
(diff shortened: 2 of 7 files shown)


kvark:
Could we perhaps add a special value `Epoch::FINAL`, which is getting pushed to the `pipeline_epoch_map` and then handled after the client calls `flush_rendered_epochs`?

glennw:
Looks like a compile error on CI - I haven't looked any deeper than that.

nical:
> Could we perhaps add a special value Epoch::FINAL, which is getting pushed to the pipeline_epoch_map and then handled after the client calls flush_rendered_epochs?

We can if you feel strongly about it. Personally I prefer keeping the epoch map (which represent the continuous state of epochs whether they change or not, and that you look at when needed) and the removed pipelines (which are more like one-off events, and missing any of them can be a leak) separated because they represent information somewhat differently and the consumer of the API presumably doesn't do exactly the same things with them.

The build error is odd, this code compiles locally and rustc is complaing that the document is not mutable while it is declared mut a few lines above. I wonder if the recent switch to rustc 1.23 fixed it.

kvark:
I just find this to be too verbose, especially considering that both flushes are very similar (and can be combined as I suggested). Leaving for @glennw to look over.

nical:
We can address the verbosity of the API by storing the pipeline epoch map and the removed pipeline list in a `PipelineInfo` struct and return it in a single `flush_pipeline_info` method (that's actually what the gecko side does to pass them around). On the render backend side, maintaining the two structures is really a different logic, so the amount of code wouldn't change anyway, but we can still store and pass the two data structures around in a `PipelineInfo` struct as well to have a single parameter in a few functions where they take two.

glennw:
I do think it's probably a good idea to keep them as separate concepts. It makes sense to me that the user would potentially care about removed pipelines, but not really care about rendered pipelines, in some cases.

It does seem unfortunate that we effectively leak memory (in tiny amounts) if the user never calls those flush APIs, but that's maybe not a huge deal.

Perhaps we could consider a single flush function that returns the two vectors to reduce the API surface? Or would it make sense for one or both of those vectors to be returned from either ```update()``` or ```render()```? That would solve any leaks, since they'll just be freed if the user doesn't look at them. I'm not sure what the semantics would be in the case of ```render``` returning an error result though.

Either way, I don't feel strongly about any of the above - we can always iterate and improve on this later as required. So I'll delegate r+ to @nical and leave to work out with @kvark if he feels further changes are needed :)

@bors-servo delegate+

nical:
I added a commit that does what I suggested in my previous comment. @kvark let me know if you feel better about this.

> It does seem unfortunate that we effectively leak memory (in tiny amounts) if the user never calls those flush APIs, but that's maybe not a huge deal.

Agreed. I didn't feel bad enough about it that I'd add debug logic to track whether we render N frames without ever flushing the removed pipelines, but that would be easy to add. I think that it'd be error prone to just flush every frame unless we return the PipelineInfo in the render function. Whoever feels strongly about this can fix it as a followup, though. There are many more important things that we could be spending our time on.

webrender/src/frame.rs
```diff
@@ -1090,16 +1094,26 @@ impl FrameContext {
...
     }
 
-    pub fn make_rendered_document(&self, frame: Frame) -> RenderedDocument {
+    pub fn make_rendered_document(&mut self, frame: Frame) -> RenderedDocument {
         let nodes_bouncing_back = self.clip_scroll_tree.collect_nodes_bouncing_back();
-        RenderedDocument::new(self.pipeline_epoch_map.clone(), nodes_bouncing_back, frame)
+        RenderedDocument::new(
+            PipelineInfo {
+                epochs: self.pipeline_epoch_map.clone(),
+                removed_pipelines: replace(&mut self.removed_pipelines, Vec::new()),
```
kvark: wouldn't this suffer from the same problem we have with GPU cache, where the consistency of the system depends on strictly producing and consuming rendered documents in order? Instead, I'd like to see the backend being able to produce any number of rendered documents, and Renderer being able to skip any of them, and still be correct. At least the current GPU cache updates are structured this way.

webrender/examples/common/boilerplate.rs
```diff
@@ -314,6 +314,7 @@ pub fn main_wrapper<E: Example>(
 
         renderer.update();
         renderer.render(framebuffer_size).unwrap();
+        let _ = renderer.flush_pipeline_info();
```
kvark: this is better 👍

webrender/src/render_backend.rs
```diff
@@ -120,7 +120,7 @@ impl Document {
         // this code is why we have `Option`, which is never `None`
         let frame_builder = self.frame_ctx.create(
             self.frame_builder.take().unwrap(),
-            &self.scene,
+            &mut self.scene,
```
kvark: this is worrying

webrender/src/renderer.rs
```diff
@@ -4925,6 +4924,20 @@ impl OutputImageHandler for () {
     }
 }
 
+pub struct PipelineInfo {
+    pub epochs: FastHashMap<PipelineId, Epoch>,
+    pub removed_pipelines: Vec<PipelineId>,
+}
+
+impl PipelineInfo {
+    pub fn new() -> Self {
```
kvark: `derive(Default)` ftw

webrender/src/frame.rs
nical: On the render backend side I believe I made sure that ordering is preserved and notifications are not lost no matter which order we build the scenes and frames in (that's the reason why the removed pipelines vector goes through all of the various pieces of the render backend and not just the scene and/or frame context).
On the renderer side, I think that it works as long as all documents render to the same swap chain, in the sense that the goal is to notify removed pipelines at least one frame after the last time it was rendered, so even if we skip rendering a document but notify its removed pipelines, said pipelines have already been removed, won't be around next time we do render the document and can be safely removed, I think.

As far as ordering go when sending to the renderer, I think that there shouldn't be any mismatch because the information is passed along with the frame itself (isn't in an separate channel or something that would need extra synchronization).

So I believe that this should work as intended even if we don't render some documents as long as it's in one swap chain (when we support direct composition and friends we'll need to track this information per swap chain). Unless by skipping rendering a document you mean receive a document in the renderer but keep sending drawing the commands of its previous version, in which case we need to add some extra logic to manage that, probably by tracking this information per document. in the renderer.

webrender/src/render_backend.rs
nical: it's unfortunately needed to flush the removed pipelines out of the scene. Storing them in the scene itself is part of what helps making sure we don't mess up ordering if we modify a scene and render frames from previous versions. I don't think there is much we can do about the mutability here other than things like Cell which I find more worrying.

webrender/src/renderer.rs
nical: Good point, fixed.

nical:
@staktrace heads up this is an API breaking change. `Renderer::flush_rendered_epochs` is replaced with `Renderer::flush_pipeline_info` and the return type is slightly different. The patch in https://bugzilla.mozilla.org/show_bug.cgi?id=1432708 will have the corresponding gecko change.

staktrace:
@nical please have that patch reviewed by somebody prior to merging this PR.

webrender/src/frame.rs
kvark: My desire here is to have rendering independent of modelling. The backend should be able to spew as many frames as it needs, and the renderer should be able to consume any subset of those. This is reflected in the type system by the fact we don't mutate most of the things when producing a `RenderedDocument`.

webrender/src/render_backend.rs
kvark: Perhaps we could keep the list of removed pipelines in the `Document` itself as opposed to the `Scene`? There are two spots where `remove_pipelines` is called, so we aren't going to be doing much extra work, but the semantics of `Scene` would be better then.

webrender/src/frame.rs
nical: I agree with this desire, I don't think this interferes.

Since removed pipelines aren't *states* that define the scene but really *events* associated to it that must be sent once, there is no getting around moving them through the pipeline which requires the events to be mutable.

In addition, we need these events to be ordered with frames that are sent to the renderer. By far the easiest way to preserve this ordering is to send the information along with the frames and require that the renderer (whether or not it ends up rendering the frames) at least collects the events and makes them available to the embedder.

To preserve the non-mutable ref to the scene we would need to track the removed pipelines in parallel to the scene instead of inside the scene, make sure that we associate the right list of removed pipelines with the right scene and not forget to send the removed pipelines of a given scene when we produce a frame out of it. That sounds a lot more error prone. If you care a lot about the scene being passed immutably, then I wouldn't be opposed to using a Cell for the removed pipelines vector.

kvark: > In addition, we need these events to be ordered with frames that are sent to the renderer.

They don't have to be 1:1. We only have to guarantee that a pipeline is not used after the removed pipeline event arrives. I see the situation exactly matching GPU cache updates, and I want the handling to be consistent. Can we have a single mechanism/entity responsible for those one-time updates? Perhaps, by extending and renaming `UpdateGpuCache` event or something. Just handle it all in the same way.

> That sounds a lot more error prone.

This is much easier than you imply. We already have association of the Document - Scene.

PR #2348, opened 2018-01-26
Hotfix for the GPU cache frame check

kvark:
Follow up to #2344
Handles the case where we had no GPU updates in this frame, also keeps track of the last received updates persistently.
r? @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2348)


staktrace:
Try push with this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=ffb13b51ea3e81a8c064e44c1f0e64b0260c8758

mstange:
r=me, kvark explained this to me in person and I'm fairly convinced that it doesn't make things worse.

PR #2349, opened 2018-01-26
Capture font portability

### LINKED ISSUE #2342: Capture portability: fonts
Font structures are platform specific, they prevent us from replaying a capture on a different machine. Technically, this should only affect `SCENE`-level captures, but we currently don't have an ability to only save the `FRAME` level without also saving/loading the basics.

kvark:
Fixes #2342
Also wraps the `UpdateResources` into device frame begin/end, since we may need to access GL context there (and I'm not sure how it works otherwise - maybe it's not used by another other than capturing?).

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2349)


kvark:
Note: also includes "Proggy.fft" statically, which weights ~5Kb and has the following license:
> Files downloaded from this site come with no guarantees or warranties and may not be suitable for any purpose whatsoever.

glennw:
I'm not sure that would be considered a safe license to include in-tree, unfortunately :(

leeoniya:
Proggy does appear to vetted as MIT by multiple sources:

https://geoff.greer.fm/2013/12/24/programming-fonts/

"The author of the these fonts is Tristan Grimmer. He was kind enough to release his work under the MIT License."

https://packages.gentoo.org/packages/media-fonts/proggy-fonts

glennw:
Ah, great, that should be fine then. Thanks!

PR #2352, opened 2018-01-28
Make FontContext::new fallible.

### LINKED ISSUE #1622: FontContext::new can panic
https://bugzilla.mozilla.org/show_bug.cgi?id=1394266

Not sure if this is some sort of corruption or what, but we're seeing a user for whom FontContext::new is panicking, because FT_Init_FreeType is returning -1819741792. It doesn't look like there's much we can do here to handle this better but if there is we might want to try.

glennw:
Fixes #1622.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2352)


glennw:
r? @kvark or anyone.

I know this isn't super high-priority, but it was easy enough to do in between rustc compiles of more substantial bugs :)

kvark:
Looks like an improvement!
@bors-servo r+

PR #2353, opened 2018-01-29
Fix a panic that can occur when a cache texture was being deleted before the render task targeting it was executed.

glennw:
This was reported in https://github.com/servo/servo/issues/19888.

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -2380,10 +2380,6 @@ impl Renderer {
                     texture_update_list,
                     profile_counters,
                 ) => {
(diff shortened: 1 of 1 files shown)


glennw:
r? @kvark 

I think this will fix the intermittent panic we have been seeing in Gecko too @staktrace

glennw:
@kvark Semi-related to this - it occurred to me that a fix for a completely unrelated bugzilla issue I'm working on will allow us to completely isolate texture cache render tasks from temporary render tasks. That should allow us to tidy up this part of the code quite a bit, and avoid re-rendering texture cache render task chains in edge case conditions etc.

PR #2360, opened 2018-01-30
Start to restructure / tidy up some parts of the frame builder traversal.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2360)


glennw:
r? @kvark or anyone

I've been stuck on fixing a few (related) bugzilla bugs for a couple of days now. I can now see the issue and know what needs to be changed, but fitting that in to the current (complex, messy) frame traversal is proving difficult.

My intent here is to introduce a number of context / state structures to the traversal - at least Frame, Picture and PrimitiveRun context / state structures. This should make it easier to introduce the state I need to move parts of the plane splitting code from batching to this traversal, which is what's required to fix this bug.

As a happy side effect, moving the plane splitting here will allow us to skip plane-splitting and split composites in a lot of cases where it is currently being done redundantly.

Regardless of the above, the commits here tidy up the frame traversal a bit, and I'll expand on this with follow up commits.

kvark:
Reviewed 1 of 3 files at r1, 5 of 5 files at r2.
Review status: 3 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/src/frame_builder.rs, line 137 at r2](https://reviewable.io:443/reviews/servo/webrender/2360#-L44DVbTo91dKW_KjIaz:-L44DVbUegfrXmLgYtfp:bmyj8be) ([raw file](https://github.com/servo/webrender/blob/e62eced9e3fb54d2d33a670149cec67294f4ae0c/webrender/src/frame_builder.rs#L137)):*
> ```Rust
> 
> impl<'a> FrameContext<'a> {
>     pub fn new(
> ```

we don't actually need this constructor

---

*[webrender/src/frame_builder.rs, line 166 at r2](https://reviewable.io:443/reviews/servo/webrender/2360#-L44Ea3vvVNHIlEgHryr:-L44Ea3wKCnaZWTxNiZX:b-qtjjcb) ([raw file](https://github.com/servo/webrender/blob/e62eced9e3fb54d2d33a670149cec67294f4ae0c/webrender/src/frame_builder.rs#L166)):*
> ```Rust
>         scroll_node: &'a ClipScrollNode,
>     ) -> Self {
>         PrimitiveRunContext {
> ```

similarly, I don't think this constructor does anything good for us.
There is no logic in here, and both this and `FrameContext` are created in this module so they can just instantiate directly.

---

*[webrender/src/frame_builder.rs, line 1831 at r2](https://reviewable.io:443/reviews/servo/webrender/2360#-L44DrSjbhxhpyP4A7ho:-L44DrSjbhxhpyP4A7hp:b8dkwn7) ([raw file](https://github.com/servo/webrender/blob/e62eced9e3fb54d2d33a670149cec67294f4ae0c/webrender/src/frame_builder.rs#L1831)):*
> ```Rust
>             profile_counters,
>             passes,
>             node_data: clip_scroll_tree.take_node_data(),
> ```

It's not a strong feeling, but this change doesn't look like an architectural improvement to me. Node data used to live locally in this function (created, passed by mutable/immutable references, then consumed here at the end), and now it's a mutable state of the tree that gets flushed every time. I wish we had this scheme for other things that we currently flush as well (pending resource updates, gpu cache updates, etc).

Is this required for your plan, or just a collateral change?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2360)*

glennw:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/src/frame_builder.rs, line 137 at r2](https://reviewable.io:443/reviews/servo/webrender/2360#-L44DVbTo91dKW_KjIaz:-L44Hz97vLUfG0RSJfgR:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e62eced9e3fb54d2d33a670149cec67294f4ae0c/webrender/src/frame_builder.rs#L137)):*

we don't actually need this constructor
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 166 at r2](https://reviewable.io:443/reviews/servo/webrender/2360#-L44Ea3vvVNHIlEgHryr:-L44I-xl_5IXTmv2GKgW:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e62eced9e3fb54d2d33a670149cec67294f4ae0c/webrender/src/frame_builder.rs#L166)):*

similarly, I don't think this constructor does anything good for us.
There is no logic in here, and both this and `FrameContext` are created in this module so they can just instantiate directly.
</blockquote></details>

Done.

---

*[webrender/src/frame_builder.rs, line 1831 at r2](https://reviewable.io:443/reviews/servo/webrender/2360#-L44DrSjbhxhpyP4A7ho:-L44I0ggLKN5cx11N7JC:b-r5169m) ([raw file](https://github.com/servo/webrender/blob/e62eced9e3fb54d2d33a670149cec67294f4ae0c/webrender/src/frame_builder.rs#L1831)):*

It's not a strong feeling, but this change doesn't look like an architectural improvement to me. Node data used to live locally in this function (created, passed by mutable/immutable references, then consumed here at the end), and now it's a mutable state of the tree that gets flushed every time. I wish we had this scheme for other things that we currently flush as well (pending resource updates, gpu cache updates, etc).

Is this required for your plan, or just a collateral change?
</blockquote></details>

I restored the node data as a separate field - now stored in FrameContext.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2360)*

kvark:
Thank you!
@bors-servo r+

PR #2362, opened 2018-01-30
Async scene building.

nical:
## TODO

- [x] plumbing to have a dedicated thread for scene building,
- [x] remove all resource cache access from scene building,
- [x] remove access to other render-backend-only objects,
- [x] reorganize transactions to know when we need to build a scene and when we don't,
- [x] move scene building to its own thread,
- [ ] enable by default.
- [x] Clean up.

## Done 

List of (subtle) things that I am changing here (I probably won't remember all of them by the time this is ready):

 - Arc the ScenePipelines so that Scene can be somewhat cheaply cloned and sent to the building thread.
 - window_size moved from FrameContext to FrameBuilder (resizing the window always require re-layout, so the DL and window size need to be kept in sync).
 - The clip scroll tree is built in the scene builder but finalized when received in the render backend thread (needs the old scroll state which isn't available in the scene building thread).
 - Arc the clip chain nodes to make the clip scroll tree sendable (because of the item above).
 - Added ApiMsg::WakeUp so that the scene builder thread can wake the render backend up after sending the built scene from another channel (because ipc channel).
 - Making scene building stateless. Currently FrameContext is responsible for two things: managing scrolling and scene building. Now that these two happen on different threads they have to be separated. As part of that some recycling of various data structure is lost (for now).
 - Split transaction into the things that need to be applied before and after scene building, also separate resource updates which will need special treatment in the near future. Right now I kept a single DocumentMsg type for both the prefix and postfix ops which untidy but is a best effort to not touch every single line in render_backend.rs and make rebasing easier, this should change.
 - Move information not used for scene building out of the Scene struct (dynamic properties and removed pipelines moved to Document).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2362)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -1207,6 +1207,7 @@ impl FrameContext {
                 pipeline_epochs: Vec::new(),
                 replacements: Vec::new(),
                 output_pipelines: &request.output_pipelines,
+                id_to_index_mapper: ClipIdToIndexMapper::new(),
             };
 
             roller.builder.push_root(
@@ -1214,6 +1215,7 @@ impl FrameContext {
                 &root_pipeline.viewport_size,
                 &root_pipeline.content_size,
                 roller.clip_scroll_tree,
+                &mut roller.id_to_index_mapper,
             );
 
             roller.builder.setup_viewport_offset(
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -25,6 +25,8 @@ use scene::{Scene, StackingContextHelpers, ScenePipeline, SceneProperties};
 use tiling::{CompositeOps, Frame};
 use renderer::PipelineInfo;
 
+use std::sync::Arc;
+
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug, Eq, Ord)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
@@ -423,7 +425,8 @@ impl<'a> FlattenContext<'a> {
             &mut self.id_to_index_mapper,
         );
 
-        self.pipeline_epochs.push((iframe_pipeline_id, pipeline.epoch));
+        let epoch = *self.scene.pipeline_epochs.get(&iframe_pipeline_id).unwrap();
+        self.pipeline_epochs.push((iframe_pipeline_id, epoch));
 
         let bounds = item.rect();
         let iframe_rect = LayerRect::new(LayerPoint::zero(), bounds.size);
@@ -1100,8 +1103,9 @@ impl FrameContext {
 
         let old_scrolling_states = self.reset();
 
-        self.pipeline_epoch_map
-            .insert(root_pipeline_id, root_pipeline.epoch);
+        let root_epoch = *scene.pipeline_epochs.get(&root_pipeline_id).unwrap();
+        self.pipeline_epoch_map.insert(root_pipeline_id, root_epoch);
+
 
         let background_color = root_pipeline
             .background_color
@@ -1179,7 +1183,7 @@ impl FrameContext {
         frame_builder: &mut FrameBuilder,
         resource_cache: &mut ResourceCache,
         gpu_cache: &mut GpuCache,
-        pipelines: &FastHashMap<PipelineId, ScenePipeline>,
+        pipelines: &FastHashMap<PipelineId, Arc<ScenePipeline>>,
         device_pixel_scale: DevicePixelScale,
         layer: DocumentLayer,
         pan: WorldPoint,
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -24,6 +24,7 @@ use resource_cache::{FontInstanceMap,ResourceCache, TiledImageMap};
 use scene::{Scene, StackingContextHelpers, ScenePipeline, SceneProperties};
 use tiling::{CompositeOps, Frame};
 use renderer::PipelineInfo;
+use render_backend::SceneRequest;
 
 use std::sync::Arc;
 
@@ -1158,6 +1159,86 @@ impl FrameContext {
             .finalize_and_apply_pending_scroll_offsets(old_scrolling_states);
 
         frame_builder
+    }
+
+    // WIP: this will likely not end up in FrameContext but for now I am mostly mimicking
+    // the current code.
+    pub fn create_async(
+        config: &FrameBuilderConfig,
+        request: SceneRequest
+    ) -> Option<FrameBuilder> {
+        let scene = &request.scene;
+        let inner_rect = request.view.inner_rect;
+        let window_size = request.view.window_size;
+        let device_pixel_scale = request.view.accumulated_scale_factor();
+
+        let root_pipeline_id = match scene.root_pipeline_id {
+            Some(root_pipeline_id) => root_pipeline_id,
+            None => return None,
+        };
+
+        let root_pipeline = match scene.pipelines.get(&root_pipeline_id) {
+            Some(root_pipeline) => root_pipeline,
+            None => return None,
+        };
+
+        if window_size.width == 0 || window_size.height == 0 {
+            error!("ERROR: Invalid window dimensions! Please call api.set_window_size()");
+        }
+
+        let background_color = root_pipeline
+            .background_color
+            .and_then(|color| if color.a > 0.0 { Some(color) } else { None });
+
+        // WIP: we need the real onw here.
+        let mut clip_scroll_tree = ClipScrollTree::new();
+
+        let frame_builder = {
+            let mut roller = FlattenContext {
+                scene,
+                builder: FrameBuilder::empty().recycle(
+                    inner_rect,
+                    background_color,
+                    *config,
+                ),
+                clip_scroll_tree: &mut clip_scroll_tree,
+                font_instances: request.font_instances,
+                tiled_image_map: request.tiled_image_map,
+                pipeline_epochs: Vec::new(),
+                replacements: Vec::new(),
+                output_pipelines: &request.output_pipelines,
+            };
+
+            roller.builder.push_root(
+                root_pipeline_id,
+                &root_pipeline.viewport_size,
+                &root_pipeline.content_size,
+                roller.clip_scroll_tree,
+            );
+
+            roller.builder.setup_viewport_offset(
+                inner_rect,
+                device_pixel_scale,
+                roller.clip_scroll_tree,
+            );
+
+            roller.flatten_root(
+                &mut root_pipeline.display_list.iter(),
+                root_pipeline_id,
+                &root_pipeline.viewport_size,
+            );
+
+            debug_assert!(roller.builder.picture_stack.is_empty());
+
+            // WIP: move this to the render backend thread I think
+            //self.pipeline_epoch_map.extend(roller.pipeline_epochs.drain(..));
+            roller.builder
+        };
+
+        // WIP
+        //clip_scroll_tree.finalize_and_apply_pending_scroll_offsets(old_scrolling_states);
+
+        Some(frame_builder)
     }
 
     pub fn update_epoch(&mut self, pipeline_id: PipelineId, epoch: Epoch) {
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -27,6 +27,7 @@ use renderer::PipelineInfo;
 use render_backend::{SceneRequest, BuiltScene};
 
 use std::sync::Arc;
+use std::mem::replace;
 
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug, Eq, Ord)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
@@ -1175,9 +1176,9 @@ impl FrameContext {
     // This will soon replace the method above and move somewhere else.
     pub fn create_frame_builder_async(
         config: &FrameBuilderConfig,
-        request: SceneRequest,
+        mut request: SceneRequest,
     ) -> BuiltScene {
-        let scene = &request.scene;
+        let scene = &mut request.scene;
         let inner_rect = request.view.inner_rect;
         let window_size = request.view.window_size;
         let device_pixel_scale = request.view.accumulated_scale_factor();
@@ -1246,6 +1247,7 @@ impl FrameContext {
             frame_builder,
             clip_scroll_tree,
             pipeline_epoch_map,
+            removed_pipelines: replace(&mut scene.removed_pipelines, Vec::new()),
             document_id: request.document_id,
             render: request.render,
         }
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -1012,7 +1012,6 @@ impl<'a> FlattenContext<'a> {
 /// Frame context contains the information required to update
 /// (e.g. scroll) a renderer frame builder (`FrameBuilder`).
 pub struct FrameContext {
-    window_size: DeviceUintSize,
     clip_scroll_tree: ClipScrollTree,
     pipeline_epoch_map: FastHashMap<PipelineId, Epoch>,
     id: FrameId,
@@ -1022,7 +1021,6 @@ pub struct FrameContext {
 impl FrameContext {
     pub fn new(config: FrameBuilderConfig) -> Self {
         FrameContext {
-            window_size: DeviceUintSize::zero(),
             pipeline_epoch_map: FastHashMap::default(),
             clip_scroll_tree: ClipScrollTree::new(),
             id: FrameId(0),
@@ -1037,6 +1035,12 @@ impl FrameContext {
         self.id.0 += 1;
 
         self.clip_scroll_tree.drain()
+    }
+
+    pub fn set_clip_scroll_tree(&mut self, tree: ClipScrollTree) {
+        let old_scrolling_states = self.reset();
+        self.clip_scroll_tree = tree;
+        self.clip_scroll_tree.finalize_and_apply_pending_scroll_offsets(old_scrolling_states);
     }
 
     #[cfg(feature = "debugger")]
@@ -1100,7 +1104,6 @@ impl FrameContext {
         if window_size.width == 0 || window_size.height == 0 {
             error!("ERROR: Invalid window dimensions! Please call api.set_window_size()");
         }
-        self.window_size = window_size;
 
         let old_scrolling_states = self.reset();
 
@@ -1118,6 +1121,7 @@ impl FrameContext {
                 builder: old_builder.recycle(
                     inner_rect,
                     background_color,
+                    window_size,
                     self.frame_builder_config,
                 ),
                 clip_scroll_tree: &mut self.clip_scroll_tree,
@@ -1165,7 +1169,8 @@ impl FrameContext {
     // the current code.
     pub fn create_async(
         config: &FrameBuilderConfig,
-        request: SceneRequest
+        request: SceneRequest,
+        clip_scroll_tree: &mut ClipScrollTree,
     ) -> Option<FrameBuilder> {
         let scene = &request.scene;
         let inner_rect = request.view.inner_rect;
@@ -1190,18 +1195,17 @@ impl FrameContext {
             .background_color
             .and_then(|color| if color.a > 0.0 { Some(color) } else { None });
 
-        // WIP: we need the real onw here.
-        let mut clip_scroll_tree = ClipScrollTree::new();
-
         let frame_builder = {
             let mut roller = FlattenContext {
                 scene,
+                // WIP, we're not really recycling anything here, clean this up.
                 builder: FrameBuilder::empty().recycle(
                     inner_rect,
                     background_color,
+                    window_size,
                     *config,
                 ),
-                clip_scroll_tree: &mut clip_scroll_tree,
+                clip_scroll_tree,
                 font_instances: request.font_instances,
                 tiled_image_map: request.tiled_image_map,
                 pipeline_epochs: Vec::new(),
@@ -1281,7 +1285,6 @@ impl FrameContext {
             self.id,
             &mut self.clip_scroll_tree,
             pipelines,
-            self.window_size,
             device_pixel_scale,
             layer,
             pan,
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -1248,8 +1248,6 @@ impl FrameContext {
             clip_scroll_tree,
             pipeline_epoch_map,
             removed_pipelines: replace(&mut scene.removed_pipelines, Vec::new()),
-            document_id: request.document_id,
-            render: request.render,
         }
     }
 
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -427,7 +427,7 @@ impl<'a> FlattenContext<'a> {
             &mut self.id_to_index_mapper,
         );
 
-        let epoch = *self.scene.pipeline_epochs.get(&iframe_pipeline_id).unwrap();
+        let epoch = self.scene.pipeline_epochs[&iframe_pipeline_id];
         self.pipeline_epochs.push((iframe_pipeline_id, epoch));
 
         let bounds = item.rect();
@@ -1113,7 +1113,7 @@ impl FrameContext {
 
         let old_scrolling_states = self.reset();
 
-        let root_epoch = *scene.pipeline_epochs.get(&root_pipeline_id).unwrap();
+        let root_epoch = scene.pipeline_epochs[&root_pipeline_id];
         self.pipeline_epoch_map.insert(root_pipeline_id, root_epoch);
 
 
@@ -1183,7 +1183,7 @@ impl FrameContext {
         let window_size = request.view.window_size;
         let device_pixel_scale = request.view.accumulated_scale_factor();
 
-        // We checked that the root pipeline is available on therender backend.
+        // We checked that the root pipeline is available on the render backend.
         let root_pipeline_id = scene.root_pipeline_id.unwrap();
         let root_pipeline = scene.pipelines.get(&root_pipeline_id).unwrap();
 
@@ -1194,7 +1194,7 @@ impl FrameContext {
         let mut pipeline_epoch_map = FastHashMap::default();
         let mut clip_scroll_tree = ClipScrollTree::new();
 
-        let root_epoch = *scene.pipeline_epochs.get(&root_pipeline_id).unwrap();
+        let root_epoch = scene.pipeline_epochs[&root_pipeline_id];
         pipeline_epoch_map.insert(root_pipeline_id, root_epoch);
 
         let frame_builder = {
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -22,9 +22,9 @@ use prim_store::ScrollNodeAndClipChain;
 use profiler::{GpuCacheProfileCounters, TextureCacheProfileCounters};
 use resource_cache::{FontInstanceMap,ResourceCache, TiledImageMap};
 use scene::{Scene, StackingContextHelpers, ScenePipeline, SceneProperties};
+use scene_builder::{SceneRequest, BuiltScene};
 use tiling::{CompositeOps, Frame};
 use renderer::PipelineInfo;
-use render_backend::{SceneRequest, BuiltScene};
 
 use std::sync::Arc;
 
@@ -1086,6 +1086,8 @@ impl FrameContext {
             .discard_frame_state_for_pipeline(pipeline_id);
     }
 
+    // When changing this, please make the same modification to build_scene,
+    // which will soon replace this method completely.
     pub fn create_frame_builder(
         &mut self,
         old_builder: FrameBuilder,
@@ -1170,88 +1172,6 @@ impl FrameContext {
         frame_builder
     }
 
-    // WIP: There is no reason for this to be in FrameContext other than making it easier
-    // to keep in sync with changes with the create_frame_builder method above.
-    // This will soon replace the method above and move somewhere else.
-    pub fn create_frame_builder_async(
-        config: &FrameBuilderConfig,
-        mut request: SceneRequest,
-    ) -> BuiltScene {
-        let inner_rect = request.view.inner_rect;
-        let window_size = request.view.window_size;
-        let device_pixel_scale = request.view.accumulated_scale_factor();
-
-        let mut pipeline_epoch_map = FastHashMap::default();
-        let mut clip_scroll_tree = ClipScrollTree::new();
-
-        let frame_builder = {
-            // Creating the borrow here brings happiness to the borrow checker.
-            let scene = &mut request.scene;
-            // We checked that the root pipeline is available on the render backend.
-            let root_pipeline_id = scene.root_pipeline_id.unwrap();
-            let root_pipeline = scene.pipelines.get(&root_pipeline_id).unwrap();
-
-            let background_color = root_pipeline
-                .background_color
-                .and_then(|color| if color.a > 0.0 { Some(color) } else { None });
-
-            let root_epoch = scene.pipeline_epochs[&root_pipeline_id];
-            pipeline_epoch_map.insert(root_pipeline_id, root_epoch);
-
-            let mut roller = FlattenContext {
-                scene,
-                // WIP, we're not really recycling anything here, clean this up.
-                builder: FrameBuilder::empty().recycle(
-                    inner_rect,
-                    background_color,
-                    window_size,
-                    *config,
-                ),
-                clip_scroll_tree: &mut clip_scroll_tree,
-                font_instances: request.font_instances,
-                tiled_image_map: request.tiled_image_map,
-                pipeline_epochs: Vec::new(),
-                replacements: Vec::new(),
-                output_pipelines: &request.output_pipelines,
-                id_to_index_mapper: ClipIdToIndexMapper::new(),
-            };
-
-            roller.builder.push_root(
-                root_pipeline_id,
-                &root_pipeline.viewport_size,
-                &root_pipeline.content_size,
-                roller.clip_scroll_tree,
-                &mut roller.id_to_index_mapper,
-            );
-
-            roller.builder.setup_viewport_offset(
-                inner_rect,
-                device_pixel_scale,
-                roller.clip_scroll_tree,
-            );
-
-            roller.flatten_root(
-                &mut root_pipeline.display_list.iter(),
-                root_pipeline_id,
-                &root_pipeline.viewport_size,
-            );
-
-            debug_assert!(roller.builder.picture_stack.is_empty());
-
-            pipeline_epoch_map.extend(roller.pipeline_epochs.drain(..));
-
-            roller.builder
-        };
-
-        BuiltScene {
-            scene: request.scene,
-            frame_builder,
-            clip_scroll_tree,
-            pipeline_epoch_map,
-            removed_pipelines: request.removed_pipelines,
-        }
-    }
-
     pub fn update_epoch(&mut self, pipeline_id: PipelineId, epoch: Epoch) {
         self.pipeline_epoch_map.insert(pipeline_id, epoch);
     }
@@ -1301,5 +1221,85 @@ impl FrameContext {
         let hit_tester = frame_builder.create_hit_tester(&self.clip_scroll_tree);
 
         (hit_tester, self.make_rendered_document(frame, removed_pipelines))
+    }
+}
+
+
+pub fn build_scene(
+    config: &FrameBuilderConfig,
+    mut request: SceneRequest,
+) -> BuiltScene {
+    let inner_rect = request.view.inner_rect;
+    let window_size = request.view.window_size;
+    let device_pixel_scale = request.view.accumulated_scale_factor();
+
+    let mut pipeline_epoch_map = FastHashMap::default();
+    let mut clip_scroll_tree = ClipScrollTree::new();
+
+    let frame_builder = {
+        // Creating the borrow here brings happiness to the borrow checker.
+        let scene = &mut request.scene;
+        // We checked that the root pipeline is available on the render backend.
+        let root_pipeline_id = scene.root_pipeline_id.unwrap();
+        let root_pipeline = scene.pipelines.get(&root_pipeline_id).unwrap();
+
+        let background_color = root_pipeline
+            .background_color
+            .and_then(|color| if color.a > 0.0 { Some(color) } else { None });
+
+        let root_epoch = scene.pipeline_epochs[&root_pipeline_id];
+        pipeline_epoch_map.insert(root_pipeline_id, root_epoch);
+
+        let mut roller = FlattenContext {
+            scene,
+            // WIP, we're not really recycling anything here, clean this up.
+            builder: FrameBuilder::empty().recycle(
+                inner_rect,
+                background_color,
+                window_size,
+                *config,
+            ),
+            clip_scroll_tree: &mut clip_scroll_tree,
+            font_instances: request.font_instances,
+            tiled_image_map: request.tiled_image_map,
+            pipeline_epochs: Vec::new(),
+            replacements: Vec::new(),
+            output_pipelines: &request.output_pipelines,
+            id_to_index_mapper: ClipIdToIndexMapper::new(),
+        };
+
+        roller.builder.push_root(
+            root_pipeline_id,
+            &root_pipeline.viewport_size,
+            &root_pipeline.content_size,
+            roller.clip_scroll_tree,
+            &mut roller.id_to_index_mapper,
+        );
+
+        roller.builder.setup_viewport_offset(
+            inner_rect,
+            device_pixel_scale,
+            roller.clip_scroll_tree,
+        );
+
+        roller.flatten_root(
+            &mut root_pipeline.display_list.iter(),
+            root_pipeline_id,
+            &root_pipeline.viewport_size,
+        );
+
+        debug_assert!(roller.builder.picture_stack.is_empty());
+
+        pipeline_epoch_map.extend(roller.pipeline_epochs.drain(..));
+
+        roller.builder
+    };
+
+    BuiltScene {
+        scene: request.scene,
+        frame_builder,
+        clip_scroll_tree,
+        pipeline_epoch_map,
+        removed_pipelines: request.removed_pipelines,
     }
 }
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -24,7 +24,7 @@ use resource_cache::{FontInstanceMap,ResourceCache, TiledImageMap};
 use scene::{Scene, StackingContextHelpers, ScenePipeline, SceneProperties};
 use tiling::{CompositeOps, Frame};
 use renderer::PipelineInfo;
-use render_backend::SceneRequest;
+use render_backend::{SceneRequest, BuiltScene};
 
 use std::sync::Arc;
 
@@ -1037,10 +1037,15 @@ impl FrameContext {
         self.clip_scroll_tree.drain()
     }
 
-    pub fn set_clip_scroll_tree(&mut self, tree: ClipScrollTree) {
+    pub fn new_async_scene_ready(
+        &mut self,
+        clip_scroll_tree: ClipScrollTree,
+        pipeline_epoch_map: FastHashMap<PipelineId, Epoch>,
+    ) {
         let old_scrolling_states = self.reset();
-        self.clip_scroll_tree = tree;
+        self.clip_scroll_tree = clip_scroll_tree;
         self.clip_scroll_tree.finalize_and_apply_pending_scroll_offsets(old_scrolling_states);
+        self.pipeline_epoch_map = pipeline_epoch_map;
     }
 
     #[cfg(feature = "debugger")]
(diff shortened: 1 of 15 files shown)


glennw:
@nical Did you want this reviewed (and merged?) as is, or is there more coming before reviewing this bit?

nical:
There is a ton more stuff to do before this is worth merging. I do want to have the work in progress visible here so that interested parties can chime in and have a chance to influence the work before I've written everything.

nical:
I'm going to keep a list of (subtle) things that I am changing here (I probably won't remember all of them by the time this is ready):

- `window_size` moved from FrameContext to FrameBuilder (resizing the window always require re-layout, so the DL and window size need to be kept in sync).
- The clip_scroll_tree is build in the scene builder but finalized when received in the render backend thread (needs the old scroll state which isn't available in the scene building thread).

glennw:
@nical I'm working on switching images to brush shaders. As part of that, if all works out as I'm planning, we'll be able to completely remove the image decompose and ```tiled_image_map``` code from the resource cache.

I'm not sure if this is relevant to your work here, but it might simplify things a bit, as it will mean less shared state that is required during scene building.

Hoping to have it done by end of the week / early next week.

nical:
This PR is already way too big, it would be great to start the review process now.

At this point most of the async scene building is in place although there are a few things to resolve and bugs to fix before it can be enabled (the code on the branch maintains both async and sync scene building code paths and defaults to non-async).
Once I can get rid of the sync scene building there is a lot of clean up that can be done.

We can go about this by:

 - Review/land this ASAP and clean things after the fact when the feature is complete and I can remove the old code path.
 - Wait for the thing to be finished, but if so:
    - please refrain from touching render_backend.rs, if you really need to let me know. Rebasing this PR is hard and I'm likely to mess something up each time I have to do it.
    - the PR to review will be cleaner but bigger.

kvark:
I believe we want this feature anyway, so we should land the intermediate result as long as it's not breaking anything and is going in the right direction. Will review later today.

glennw:
I agree - I'm fine with landing the intermediate work given the pre-conditions mentioned above by @kvark

kvark:
looks like a good start :+1: 

---

Reviewed 5 of 8 files at r2, 8 of 8 files at r3.

---

*[webrender/src/frame.rs, line 363 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sm3KDd6tN5U0RsTqY:-L4sm3KDd6tN5U0RsTqZ:bottkhn) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/frame.rs#L363)):*
> ```Rust
>         );
> 
>         let epoch = *self.scene.pipeline_epochs.get(&pipeline_id).unwrap();
> ```

nit: can just do `[&pipeline_id]`

---

*[webrender/src/frame.rs, line 1127 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4smp5IQWEMBX0NohAD:-L4smp5IQWEMBX0NohAE:b-p0mj58) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/frame.rs#L1127)):*
> ```Rust
>     // to keep in sync with changes with the create_frame_builder method above.
>     // This will soon replace the method above and move somewhere else.
>     pub fn create_frame_builder_async(
> ```

there is also a mismatch with return type now, which implies the function to be rather `build_scene_async`

---

*[webrender/src/frame.rs, line 1136 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sn1BE1N-ze4q77hnd:-L4sn1BE1N-ze4q77hne:b-o0ma4o) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/frame.rs#L1136)):*
> ```Rust
>         let device_pixel_scale = request.view.accumulated_scale_factor();
> 
>         // We checked that the root pipeline is available on therender backend.
> ```

"therender"

---

*[webrender/src/render_backend.rs, line 761 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sr2F3EL3If35__J59:-L4sr2F4YltMZ_ml0LR_:b-dxq39y) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/render_backend.rs#L761)):*
> ```Rust
>             }
> 
>         return true;
> ```

nit: `return`

---

*[webrender/src/render_backend.rs, line 776 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4srEppYubsNkW9ulMa:-L4srEpqM6D2ad9207AP:b-vdvf0a) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/render_backend.rs#L776)):*
> ```Rust
>         // will most likely change soon.
>         let prefix_ops = replace(&mut txn.prefix_ops, Vec::new());
>         for doc_msg in prefix_ops {
> ```

I wonder if it would make sense to separate the ops enum into prefix/postfix at the type level

---

*[webrender/src/render_backend.rs, line 1184 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4srX9V1EyF5v9XZuvo:-L4srX9Wi5YnPc0KD6ty:b7dl806) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/render_backend.rs#L1184)):*
> ```Rust
> pub enum SceneBuilderMsg {
>     Transaction {
>         document_id: DocumentId,
> ```

looks like we need a (generic) struct here

---

*[webrender_api/src/api.rs, line 133 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sosz900B2dHNxya-a:-L4sosz900B2dHNxya-b:b-1kj7u3) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L133)):*
> ```Rust
> ///  - no redundant work is performed if two commands in the same transaction cause the scene or
> ///    the frame to be rebuilt.
> pub struct Transaction {
> ```

It appears to be specifically tailored towards the scene building now (one transaction - one scene built), where it used to be more generic.
For example, why can't the scene building just be a `DocumentMsg` in the list of messages?

---

*[webrender_api/src/api.rs, line 135 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4spEr7OTJZumNyk3fq:-L4spEr7OTJZumNyk3fr:b-3rk8zh) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L135)):*
> ```Rust
> pub struct Transaction {
>     // Operations to apply before scene building.
>     prefix_ops: Vec<DocumentMsg>,
> ```

also, we could use `TransactionMsg` in the implementation

---

*[webrender_api/src/api.rs, line 147 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4soYTwSPmA-XIJUTqX:-L4soYTwSPmA-XIJUTqY:b-ao38pl) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L147)):*
> ```Rust
>     // If true the transaction is piped through the scene building thread, if false
>     // it will be applied directly on the render backend.
>     use_scene_builder_thread: bool,
> ```

seems like an implementation detail on who is processing the transactions

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362#-:-L4ss1zm7js_Vo5Y2Pp9:b-5binfo)*

kvark:
I'm going to take another (higher-level) look at it tomorrow.

nical:
Thanks a lot for starting the review process. Let me know if you want to schedule a vidyo meeting or something for me to assist you in getting around this wall of code. 

> seems like an implementation detail on who is processing the transactions

If we know that scrolling is always able to be out of sync with scenes we can detect it and figure it out on the render backend (although I'd feel better about letting the user of the api decide, because it's trivial to do there and more flexible).
We will have to let the sender of the transaction opt into skipping the scene building thread for async video playback, though. The rest of the web is forced to respect the ordering of transactions and there currently is no way for webrender to know that an image resource is a video that can be updated out of sync with the scenes.

nical:

---

*[webrender/src/render_backend.rs, line 776 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4srEppYubsNkW9ulMa:-L4tx_6bXGysYWNH7TT7:byoebfk) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/render_backend.rs#L776)):*

I wonder if it would make sense to separate the ops enum into prefix/postfix at the type level
</blockquote></details>

It would totally make sense, the reason I haven't done this (yet) is to keep the diff smaller at least until we get close to landing.

---

*[webrender_api/src/api.rs, line 133 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sosz900B2dHNxya-a:-L4tv5LDcVllDQmCe319:b-uwila0) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L133)):*

It appears to be specifically tailored towards the scene building now (one transaction - one scene built), where it used to be more generic.
For example, why can't the scene building just be a `DocumentMsg` in the list of messages?
</blockquote></details>

I am not sure what you mean. Transactions don't necessarily result in scenes being built (only if they contain SetDisplayList ops). The information stored is the same but is represented in a way that is easier for the render backend and scene builder to consume. Otherwise we'd have to do a pass over all the ops and separate them on the render backend to figure out what to do.
We could definitely have a document message to build the scene (and SetDL would not automatically build the scene), but that's mostly independent of this work I think.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362)*

mrobinson:
Sorry for the conflict. The resolution is pretty simple and I have done it here: https://github.com/mrobinson/webrender/tree/pr/2362

nical:
Stuff that isn't done yet but might help with understanding some of the awkwardness in this PR:

The render backend will hold two scenes. The "next" scene is the scene on which we apply the prefix document ops before sending to the scene builder. The idea is that it represents the the most up to date scene with the latest transactions applied, but we aren't yet able to render it.
The "current" scene is what we receive from the scene builder thread, and it is ready to be rendered.

Note that by the time I am done with this, the Scene struct will probably only contain a hash map of `Arc<ScenePipeline>`, and the root pipeline id, so it should be cheap to clone and throw away (the expensive stuff being reference counted).
The awkward prefix/postfix terminology will probably go away and there should be a `SceneMsg` enum corresponding to the `DocumentMsg` variants that the PR currently applies in the "prefix" pass (what needs to be applied before building the scene is mostly modifications to the scene).

nical:
Review status: 0 of 15 files reviewed at latest revision, 9 unresolved discussions.

---

*[webrender/src/frame.rs, line 363 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sm3KDd6tN5U0RsTqY:-L5FLLASiAotnIvadC5T:b-896fix) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/frame.rs#L363)):*

nit: can just do `[&pipeline_id]`
</blockquote></details>

Done.

---

*[webrender/src/frame.rs, line 1127 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4smp5IQWEMBX0NohAD:-L5FLPr9-zKb3VytcBde:b-896fix) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/frame.rs#L1127)):*

there is also a mismatch with return type now, which implies the function to be rather `build_scene_async`
</blockquote></details>

Done.

---

*[webrender/src/frame.rs, line 1136 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sn1BE1N-ze4q77hnd:-L5FLQxoPx2woxF5P-qe:b-896fix) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/frame.rs#L1136)):*

"therender"
</blockquote></details>

Done.

---

*[webrender/src/render_backend.rs, line 761 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sr2F3EL3If35__J59:-L5FLTImKIsao8mUMI_k:b-896fix) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender/src/render_backend.rs#L761)):*

nit: `return`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362)*

nical:
Building scenes asynchronously now passes the tests. However, enabling it doesn't provide any advantages unless we actually use the ability to let some transactions skip the scene building thread (It's opt-in and exposed in the `Transaction` struct).

To recap, the flow looks like this:

rect6475

Where render means generate a frame and send it to the renderer.

One thing that I realized recently is that building the scene resets the scrolling state. This isn't great because we want to be scrolling while we build the scene (that's the whole point), and as soon as a scene is built, it should pick up what it can of the scrolling information from the previous scene, otherwise building a scene jumps back to where things were at the time set_display_list was called.

Are the ids of scrolling elements stable between display lists ? Is there some fixing up that needs to be done to pick up the previous scrolling information (I'm guessing that as we scroll we generate new display lists with a new origin for the scrolled element and so we might need to remember the offset between the origins of a scrollable element between two DL)?

cc @staktrace

staktrace:
I think that instead of trying to preserve the scrolling state from the previous scene we should just get it from APZ. Keep in mind that we also need to move the APZ update from where it's happening now to the point of the "Set scene" in your picture.

In other words, right now when the compositor gets a new display list from the main thread, it [immediately calls UpdateAPZ](https://searchfox.org/mozilla-central/rev/d03ad8843e3bf2e856126bc53b0475c595e5183b/gfx/layers/wr/WebRenderBridgeParent.cpp#628-629) to apply the changes to APZ. However if that display list is going to be built in WR async, then we don't want to apply the APZ changes right away either. Instead we should leave APZ with the previous state until the scene is built and we're ready to do the scene swap. At that point we should do the APZ update and also ask APZ for the latest scroll data (along with any other async transforms, OMTA, etc.) which currently happens [here](https://searchfox.org/mozilla-central/rev/d03ad8843e3bf2e856126bc53b0475c595e5183b/gfx/layers/wr/WebRenderBridgeParent.cpp#1231-1254).

This will also solve the problems in bug 1436200 and bug 1434682 which are happening now. The only tricky part IMO is storing the APZ data for a particular display list until the scene swap happens. Instead of putting it in WebRenderBridgeParent::mScrollData we'll need to stash it somewhere associated with a transaction id or epoch or something, such that we can retrieve it when the scene swap happens and apply it. If we build multiple scenes before we finally do a swap, then we should store all of the associated APZ datas and apply them all at swap time. FWIW this behaviour seems very similar to what we're doing with the HoldPendingTransactionId/FlushTransactionIdsForEpoch so maybe we can add the scroll data into the PendingTransactionId struct and just have it ride along there.

nical:
@kvark @glennw review ping.
Async scene building is disabled by default in this PR and it should be safe to land this in its current state. The majority of the work in place is independent of the APZ discussion (although we'll have to figure the latter out before we can really enable async scene building).

staktrace:
@nical I left a comment at https://bugzilla.mozilla.org/show_bug.cgi?id=1391318#c2 about APZ. In retrospect might have been better to leave here, but I think this PR should merge anyway so it doesn't bitrot while we figure out how to fit APZ into the picture.

webrender/src/frame.rs
```diff
@@ -21,9 +21,12 @@ use internal_types::{FastHashMap, FastHashSet, RenderedDocument};
 use profiler::{GpuCacheProfileCounters, TextureCacheProfileCounters};
 use resource_cache::{FontInstanceMap,ResourceCache, TiledImageMap};
 use scene::{Scene, StackingContextHelpers, ScenePipeline, SceneProperties};
+use scene_builder::{SceneRequest, BuiltScene};
 use tiling::{CompositeOps, Frame};
 use renderer::PipelineInfo;
 
+use std::sync::Arc;
```
mrobinson: This should probably go with the rest of the `use` statements.

webrender/src/render_backend.rs
```diff
@@ -510,20 +549,82 @@ impl RenderBackend {
...
+                                doc.current.removed_pipelines.extend(built_scene.removed_pipelines.drain(..));
+                                doc.frame_ctx.new_async_scene_ready(
+                                    built_scene.clip_scroll_tree,
+                                    built_scene.pipeline_epoch_map,
+                                );
+                                doc.render_on_hittest = true;
+                            }
+                        } else {
+                            // The document was removed while we were building it, skip it.
+                            // TODO: we might want to just ensure that removed documents are
```
mrobinson: It seems like there might always be the possibility for a race condition here even if the removed documents are forwarded to the scene building thread?

webrender/src/render_backend.rs
```diff
@@ -510,20 +549,82 @@ impl RenderBackend {
...
+                            // always forwarded to the scene builder thread to avoid this case.
+                            continue;
+                        }
+
+                        let txn = TransactionMsg {
```
mrobinson: I think that `transaction_msg` might be a better name here.

webrender/src/render_backend.rs
```diff
@@ -510,20 +549,82 @@ impl RenderBackend {
...
+        }
+    }
+
+    fn process_api_msg(
+        &mut self,
+        msg: ApiMsg,
+        profile_counters: &mut BackendProfileCounters,
+        frame_counter: &mut u32,
+    ) -> bool {
+        // WIP: reeindent that when the work is closer to land (otherwise rebasing is
```
mrobinson: I guess you'll need to do this before landing?

webrender/src/render_backend.rs
```diff
@@ -623,85 +724,136 @@ impl RenderBackend {
...
-        }
+
+        true
     }
 
     fn update_document(
         &mut self,
         document_id: DocumentId,
-        doc_msgs: Vec<DocumentMsg>,
+        mut txn: TransactionMsg,
```
mrobinson: `transaction_msg` again, I think.

webrender_api/src/api.rs
```diff
@@ -358,9 +430,15 @@ pub struct AddFontInstance {
     pub variations: Vec<FontVariation>,
 }
 
+// TODO: Split this in two emums.
+// I am postponing this because while it is a trivial change, it is a hard one to rebase.
```
mrobinson: Will you do this now? If not, probably better to remove this comment and open a bug tracking this.

webrender_api/src/api.rs
```diff
@@ -814,7 +887,10 @@ impl RenderApi {
                     flags: HitTestFlags)
                     -> HitTestResult {
         let (tx, rx) = channel::msg_channel().unwrap();
-        self.send(document_id, DocumentMsg::HitTest(pipeline_id, point, flags, tx));
```
mrobinson: Is this line longer than 100 characters? If not, I think this rewrapping is unnecessary.

wrench/src/yaml_frame_writer.rs
```diff
@@ -1153,12 +1151,18 @@ impl webrender::ApiRecordingReceiver for YamlFrameWriterReceiver {
...
+                }
+                for doc_msg in &txn.frame_ops {
+                    match *doc_msg {
                         DocumentMsg::UpdateDynamicProperties(ref properties) => {
                             self.scene.properties.set_properties(properties);
                         }
                         _ => {}
                     }
                 }
+
```
mrobinson: This extra line seems to be added on accident.

kvark:
Reviewed 3 of 8 files at r4, 11 of 12 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[webrender/src/frame.rs, line 1137 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oS5Rt3reiR1Nk9Gda:-L5oS5Rt3reiR1Nk9Gdb:bsujite) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/frame.rs#L1137)):*
> ```Rust
>         texture_cache_profile: &mut TextureCacheProfileCounters,
>         gpu_cache_profile: &mut GpuCacheProfileCounters,
> 		scene_properties: &SceneProperties,
> ```

tabs detected!

---

*[webrender/src/record.rs, line 79 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ogXdEsy2GxDW5NX1o:-L5ogXdGVOdboBkuWFG1:b-m8xzm1) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/record.rs#L79)):*
> ```Rust
>             for msg in &msgs.frame_ops {
>                 match *msg {
>                     DocumentMsg::GetScrollNodeState(..) |
> ```

why are those types of messages applicable to both before/after building a scene?

---

*[webrender/src/render_backend.rs, line 74 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oU1VX3nkryw3OTzbH:-L5oU1VX3nkryw3OTzbI:b13d2ke) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L74)):*
> ```Rust
> struct SceneData {
>     scene: Scene,
>     removed_pipelines: Vec<PipelineId>,
> ```

eh, the removed pipelines logic doesn't fit too nice here

---

*[webrender/src/render_backend.rs, line 167 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oUqTF-v3_tRcZON0p:-L5oUqTF-v3_tRcZON0q:b-77cfe1) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L167)):*
> ```Rust
>             &self.output_pipelines,
>         );
>         self.current.removed_pipelines.extend(self.pending.removed_pipelines.drain(..));
> ```

shouldn't we clear this list first?

---

*[webrender/src/render_backend.rs, line 783 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oVWQsSvuAHe1-zVmt:-L5oVWQsSvuAHe1-zVmu:b3iypvt) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L783)):*
> ```Rust
>         let mut op = DocumentOps::nop();
> 
>         for doc_msg in replace(&mut txn.scene_ops, Vec::new()) {
> ```

do we still need this `replace` here?

---

*[webrender/src/scene.rs, line 119 at r2](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZIYg0I3tbV9mzzs2:-L5oZIYifXWVy6ZR3n1I:brp01ui) ([raw file](https://github.com/servo/webrender/blob/4cfdc94c247a987c41043e5617fd44b7ce304eb5/webrender/src/scene.rs#L119)):*
> ```Rust
>             root_pipeline_id: None,
>             pipelines: FastHashMap::default(),
>             removed_pipelines: Vec::new(),
> ```

hmm, so why do we need to separate `removed_pipeline`s from `pipeline_epochs` now?

---

*[webrender/src/scene_builder.rs, line 20 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ohVZWkkgtaI4Hd4LS:-L5ohVZWkkgtaI4Hd4LT:b-boj0fo) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L20)):*
> ```Rust
>     Transaction {
>         document_id: DocumentId,
>         scene: Option<SceneRequest>,
> ```

In what case would a scene be `None` here?

---

*[webrender/src/scene_builder.rs, line 22 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZUVkOTByKN_onGsQ:-L5oZUVkOTByKN_onGsR:bng38u4) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L22)):*
> ```Rust
>         scene: Option<SceneRequest>,
>         resource_updates: ResourceUpdates,
>         frame_ops: Vec<DocumentMsg>,
> ```

`frame operations: Vec<document message>` does't sound like a good match :)

---

*[webrender/src/scene_builder.rs, line 29 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZokMzbNPCx4l4vUG:-L5oZokMzbNPCx4l4vUH:b-6otfsp) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L29)):*
> ```Rust
> 
> // Message from scene builder to render backend.
> pub enum SceneBuilderMsg {
> ```

both this and `SceneBuilderRequest` are message, but only this one ends with `Msg`. Perhaps, we can call it `SceneBuilderResult` instead?

---

*[webrender/src/scene_builder.rs, line 31 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZfNMMCDDUvpN6xxN:-L5oZfNMMCDDUvpN6xxO:bycv3tm) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L31)):*
> ```Rust
> pub enum SceneBuilderMsg {
>     Transaction {
>         document_id: DocumentId,
> ```

any reason not to share this struct with `SceneBuilderRequest`?

---

*[webrender/src/scene_builder.rs, line 123 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ogxpY_CgGVa171VEC:-L5ogxpY_CgGVa171VED:b-ftju7l) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L123)):*
> ```Rust
>                 let _ = self.api_tx.send(ApiMsg::WakeUp);
>             }
>             SceneBuilderRequest::Stop => { return false; }
> ```

can we instead just treat a closed channel on the other side as `Stop`?

---

*[webrender_api/src/api.rs, line 133 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sosz900B2dHNxya-a:-L5ofD2hl6rs06o_2gWi:b-3547ua) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L133)):*
Well, now that you removed the commits it's no longer obvious, but previously we had
> Operations to apply before scene building.
> Operations to apply after scene building.

Which pretty much implies 1 transaction = 1 scene building operation

---

*[webrender_api/src/api.rs, line 434 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L4l54jdopml1VR0JeX0-r7-434:-L5ofqMvI2gydMJxZzPK:bv7nsx7) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender_api/src/api.rs#L434)):*

Will you do this now? If not, probably better to remove this comment and open a bug tracking this.
</blockquote></details>

Also, either specify who ("I am") or rephrase to something like "Note: this is a trivial change but it makes difficult to rebase"

---

*[webrender_api/src/api.rs, line 457 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ogCy88qPQV5a9ukJz:-L5ogCyAHH5yadlWtTrc:b-hh2ulw) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender_api/src/api.rs#L457)):*
> ```Rust
>     },
> 
>     // Frame messages (apply after building the scene).
> ```

our `Frame` term usage is quite overloaded

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362)*

nical:
Review status: all files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[webrender/src/render_backend.rs, line 167 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oUqTF-v3_tRcZON0p:-L5tGRNUuxmBg_PTn-EW:b-en8fwm) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L167)):*

shouldn't we clear this list first?
</blockquote></details>

I don't think so. If there are still removed pipelines in this list it means they haven't been signaled and clearing the list would lose them. It's unlikely because we tend to render right after building scenes and that case would happen if we build the scene twice without rendering, but I'd rather be safe than sorry.

---

*[webrender/src/render_backend.rs, line 574 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L46KWoyuSQPxqRgV_bB-r7-574:-L5tH6GOfsE_jD1zDois:b-ycjros) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L574)):*

It seems like there might always be the possibility for a race condition here even if the removed documents are forwarded to the scene building thread?
</blockquote></details>

What do you have in mind (in the case where we forward all remove pipelines messages through the scene building thread)?

---

*[webrender/src/render_backend.rs, line 622 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L46KWoyuSQPxqRgV_bB-r7-622:-L5tHI8EqYdMl0kSoVvl:b-bbkcwj) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L622)):*

I guess you'll need to do this before landing?
</blockquote></details>

Yeah, as soon as I know that I can land it right after to avoid making rebases trickier than they are already.

---

*[webrender/src/render_backend.rs, line 783 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oVWQsSvuAHe1-zVmt:-L5tHXgSXzXqZmcFEnkn:b-qouv86) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L783)):*

do we still need this `replace` here?
</blockquote></details>

Yeah unfortunately, because the tcn object used further down (it is passed to forward_transaction_to_scene_builder) so we can't move out part of it without upsetting the borrowck.

---

*[webrender/src/scene.rs, line 119 at r2](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZIYg0I3tbV9mzzs2:-L5tIGMrZPag_Txe7go0:bss2fz5) ([raw file](https://github.com/servo/webrender/blob/4cfdc94c247a987c41043e5617fd44b7ce304eb5/webrender/src/scene.rs#L119)):*

hmm, so why do we need to separate `removed_pipeline`s from `pipeline_epochs` now?
</blockquote></details>

They look similar because pipeline is in both names but they aren't really handled the same way and aren't used by the same things.
The pipeline_epochs are used during scene building at the same time as the scene, so keeping it in there makes sense to me. More importantly, when we clone the scene to send it to the scene builder, we need to clone the pipeline epoch map along with it, bit we need to drain the removed pipelines rather than cloning them (cloning would mean they are signaled twice) so it is less error prone to move the removed pipelines out of the scene.

---

*[webrender/src/scene_builder.rs, line 20 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ohVZWkkgtaI4Hd4LS:-L5tK-bOVgvuat7SEQ2K:b-gny8rc) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L20)):*

In what case would a scene be `None` here?
</blockquote></details>

It can be none when we have to forward a transaction that doesn't involve building a scene for the sake of preserving the ordering of the transactions. Only specific things can avoid this ordering constraint, like video frames and scrolling.

---

*[webrender/src/scene_builder.rs, line 22 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZUVkOTByKN_onGsQ:-L5tKOt6FmeiiJ0wHrt5:bxq60r2) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L22)):*

`frame operations: Vec<document message>` does't sound like a good match :)
</blockquote></details>

Yeah, I mentioned that elsewhere but the messages will be split into at least two enums, one for scene operations, and one for "frame operations". I am doing it as soon as I get the green light for merging, since splitting the enum will mean reorder a lot of code in render_backend.rs and I am trying to avoid making rebasing hard. 

---

*[webrender/src/scene_builder.rs, line 29 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZokMzbNPCx4l4vUG:-L5tKzdKgcEyH4p5ynbF:b-80tlqk) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L29)):*

both this and `SceneBuilderRequest` are message, but only this one ends with `Msg`. Perhaps, we can call it `SceneBuilderResult` instead?
</blockquote></details>

Sure.

---

*[webrender/src/scene_builder.rs, line 31 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZfNMMCDDUvpN6xxN:-L5tL20Qo7CrcjFkpUro:bmd94g0) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L31)):*

any reason not to share this struct with `SceneBuilderRequest`?
</blockquote></details>

Mostly because I think that the fields in both structures are likely to change as we polish this stuff and while aggressively sharing the structures is good when things settle I think it'd get in the way in the mean time.

---

*[webrender/src/scene_builder.rs, line 123 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ogxpY_CgGVa171VEC:-L5tLo6Cjab-qPquuCRH:b-kgc156) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L123)):*

can we instead just treat a closed channel on the other side as `Stop`?
</blockquote></details>

Maybe? I'll have a look.

---

*[webrender_api/src/api.rs, line 133 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sosz900B2dHNxya-a:-L5tM2LAaOJqCeb0E4F9:b-cckk4c) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L133)):*

Well, now that you removed the commits it's no longer obvious, but previously we had
> Operations to apply before scene building.
> Operations to apply after scene building.

Which pretty much implies 1 transaction = 1 scene building operation
</blockquote></details>

It was rather things to apply before and after if there is scene building, but anyway, I hope that the way the code looks now alleviates this concern.

---

*[webrender_api/src/api.rs, line 434 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L4l54jdopml1VR0JeX0-r7-434:-L5tMIYlcQbQVF6fDw1c:b82ho8d) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender_api/src/api.rs#L434)):*

Also, either specify who ("I am") or rephrase to something like "Note: this is a trivial change but it makes difficult to rebase"
</blockquote></details>

The idea is that this is a note to the reviewers and won't land as is since I would make the change before landing.

---

*[webrender_api/src/api.rs, line 457 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ogCy88qPQV5a9ukJz:-L5tMYchdHxvLY6kvN6O:bfy5v3b) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender_api/src/api.rs#L457)):*

our `Frame` term usage is quite overloaded
</blockquote></details>

Agreed, although in general frame is used for variations of things that are built each frame that we send to the renderer. So in my head in this case this resonates with what they all have in common (but maybe I am missing other overloads of the term fram).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362)*

kvark:
Review status: all files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[webrender/src/render_backend.rs, line 167 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oUqTF-v3_tRcZON0p:-L5tQjY5krwgYjGP9JaM:bauaqvh) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L167)):*

I don't think so. If there are still removed pipelines in this list it means they haven't been signaled and clearing the list would lose them. It's unlikely because we tend to render right after building scenes and that case would happen if we build the scene twice without rendering, but I'd rather be safe than sorry.
</blockquote></details>

I see. We might do something like
```rust
if !self.current.removed_pipelines.is_empty() {
    warn!("Stale removed pipelines detected: {:?}", self.current.removed_pipelines);
}
```

---

*[webrender/src/render_backend.rs, line 783 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oVWQsSvuAHe1-zVmt:-L5tR3k56A60A8WwodVq:b3fklcj) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L783)):*

Yeah unfortunately, because the tcn object used further down (it is passed to forward_transaction_to_scene_builder) so we can't move out part of it without upsetting the borrowck.
</blockquote></details>

Hmm, I see. The alternative could be destructuring `txn` before working with the parts of it.

---

*[webrender/src/scene.rs, line 119 at r2](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZIYg0I3tbV9mzzs2:-L5tRBWubgW_3dv5J2D6:b-hznfom) ([raw file](https://github.com/servo/webrender/blob/4cfdc94c247a987c41043e5617fd44b7ce304eb5/webrender/src/scene.rs#L119)):*

They look similar because pipeline is in both names but they aren't really handled the same way and aren't used by the same things.
The pipeline_epochs are used during scene building at the same time as the scene, so keeping it in there makes sense to me. More importantly, when we clone the scene to send it to the scene builder, we need to clone the pipeline epoch map along with it, bit we need to drain the removed pipelines rather than cloning them (cloning would mean they are signaled twice) so it is less error prone to move the removed pipelines out of the scene.
</blockquote></details>

and those limitations are a consequence of having unique semantics for removed pipelines in the first place. If it was just `LAST_EPOCH` in the epoch map, we'd have less headache managing them.

---

*[webrender/src/scene_builder.rs, line 20 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ohVZWkkgtaI4Hd4LS:-L5tRWaBdvwAzkCDQ1T6:bqvcm0l) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L20)):*

It can be none when we have to forward a transaction that doesn't involve building a scene for the sake of preserving the ordering of the transactions. Only specific things can avoid this ordering constraint, like video frames and scrolling.
</blockquote></details>

Sounds like it needs a separate variant then in `SceneBuilderRequest`

---

*[webrender/src/scene_builder.rs, line 22 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZUVkOTByKN_onGsQ:-L5tRbAHE7sqPJ0bhbbJ:b-9ehkt8) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L22)):*

Yeah, I mentioned that elsewhere but the messages will be split into at least two enums, one for scene operations, and one for "frame operations". I am doing it as soon as I get the green light for merging, since splitting the enum will mean reorder a lot of code in render_backend.rs and I am trying to avoid making rebasing hard. 
</blockquote></details>

ok, let's put a todo above then, even if temporary

---

*[webrender_api/src/api.rs, line 133 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sosz900B2dHNxya-a:-L5tRhZZPnCNvObF0uJS:b-782t59) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L133)):*
correction:
> now that you removed the commits it's no longer obvious

Was meant to be
> now that you removed the **comments** it's no longer obvious

---

*[webrender_api/src/api.rs, line 457 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ogCy88qPQV5a9ukJz:-L5tRuu39_oAuqmq2Yod:b-igf2qv) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender_api/src/api.rs#L457)):*
> in general frame is used for variations of things that are built each frame

gotcha :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362)*

nical:
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[webrender/src/scene.rs, line 119 at r2](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oZIYg0I3tbV9mzzs2:-L5tT0eom-eJjLTDwfxJ:bob98ck) ([raw file](https://github.com/servo/webrender/blob/4cfdc94c247a987c41043e5617fd44b7ce304eb5/webrender/src/scene.rs#L119)):*

and those limitations are a consequence of having unique semantics for removed pipelines in the first place. If it was just `LAST_EPOCH` in the epoch map, we'd have less headache managing them.
</blockquote></details>

Interestingly, what you see as a limitation here, I see as something straightforward working the way it is supposed to. I remain absolutely certain that the last epoch thing would be a lot harder to manage correctly, because of the very different nature of the two things.

---

*[webrender_api/src/api.rs, line 133 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sosz900B2dHNxya-a:-L5tTmUwwgI6MRlDOm26:b63ized) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L133)):*

correction:
> now that you removed the commits it's no longer obvious

Was meant to be
> now that you removed the **comments** it's no longer obvious
</blockquote></details>

Ah, I'll add comments there if it helps.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362)*

nical:
Review status: 4 of 15 files reviewed at latest revision, 20 unresolved discussions.

---

*[webrender/src/frame.rs, line 28 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L4HMddH0iljmSCCKVK7-r7-28:-L5wuRD5PinY5raWoa6-:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/frame.rs#L28)):*

This should probably go with the rest of the `use` statements.
</blockquote></details>

Done.

---

*[webrender/src/frame.rs, line 1137 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oS5Rt3reiR1Nk9Gda:-L5wuS83KdyB46HdhW1_:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/frame.rs#L1137)):*

tabs detected!
</blockquote></details>

Done.

---

*[webrender/src/record.rs, line 79 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ogXdEsy2GxDW5NX1o:-L5wuXg16w-ewXbbzbw_:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/record.rs#L79)):*

why are those types of messages applicable to both before/after building a scene?
</blockquote></details>

Done.

---

*[webrender/src/render_backend.rs, line 46 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L46KWoyuSQPxqRgV_bB-r7-46:-L5wuZaioVY8cPqBEjAF:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L46)):*

Might be good to leave your name here just to make it a bit easier to know who is speaking (ie "WIP(nical)" or something like that).
</blockquote></details>

Done.

---

*[webrender/src/render_backend.rs, line 167 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oUqTF-v3_tRcZON0p:-L5wuhmJPhOrb4ouA5ke:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L167)):*

I see. We might do something like
```rust
if !self.current.removed_pipelines.is_empty() {
    warn!("Stale removed pipelines detected: {:?}", self.current.removed_pipelines);
}
```
</blockquote></details>

Done.

---

*[webrender/src/render_backend.rs, line 579 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L46KWoyuSQPxqRgV_bB-r7-579:-L5wumjw09w3NtWIsxLE:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L579)):*

I think that `transaction_msg` might be a better name here.
</blockquote></details>

Done.

---

*[webrender/src/render_backend.rs, line 783 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5oVWQsSvuAHe1-zVmt:-L5wuvL7pEMVxDuXHCXm:b-hbg0d8) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/render_backend.rs#L783)):*

Hmm, I see. The alternative could be destructuring `txn` before working with the parts of it.
</blockquote></details>

That would quite verbose, but actually, `Vec::drain` does what we need without using replace.

---

*[webrender/src/scene_builder.rs, line 20 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L5ohVZWkkgtaI4Hd4LS:-L5wwSkTphVW8SD5og-o:b-1vbufa) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender/src/scene_builder.rs#L20)):*

Sounds like it needs a separate variant then in `SceneBuilderRequest`
</blockquote></details>

Let's iterate on that in the followup APZ work. This code may look very different depending on what happens in that discussion.

---

*[webrender_api/src/api.rs, line 133 at r3](https://reviewable.io:443/reviews/servo/webrender/2362#-L4sosz900B2dHNxya-a:-L5wwhOB_yRe6XH2Qd3n:b-896fix) ([raw file](https://github.com/servo/webrender/blob/754895bd4d9be8fed4e4d488df1aa7d6101bfa65/webrender_api/src/api.rs#L133)):*

Ah, I'll add comments there if it helps.
</blockquote></details>

Done.

---

*[webrender_api/src/api.rs, line 434 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L4l54jdopml1VR0JeX0-r7-434:-L5wwmEVbHKxxPbLzocX:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/webrender_api/src/api.rs#L434)):*

The idea is that this is a note to the reviewers and won't land as is since I would make the change before landing.
</blockquote></details>

Done.

---

*[wrench/src/yaml_frame_writer.rs, line 1165 at r7](https://reviewable.io:443/reviews/servo/webrender/2362#-L4qY_-YrV9TtDTpmRum-r7-1165:-L5wwpWoVmpiwaNM1RRi:b-896fix) ([raw file](https://github.com/servo/webrender/blob/12ce449299da5a794cacc695bb68d7159b108c9b/wrench/src/yaml_frame_writer.rs#L1165)):*

This extra line seems to be added on accident.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2362)*

nical:
Let's land this today.
Unless there are changes left that are important enough that they should block this, please delegate the landing privileges to me and I'll do the remaining re-indentation and enum spitting right before asking bors to merge.

glennw:
I'm a bit out of the loop on this PR (sorry) but it sounds like it's basically ready to go and comments have been addressed.

@bors-servo delegate+

staktrace:
@nical This PR has made things pretty orange on my WR update try push. See https://bugzilla.mozilla.org/show_bug.cgi?id=1439565#c8 for links

PR #2365, opened 2018-01-31
Debug commands and render task capturing

kvark:
When we hook it up to Gecko, playing with those debug commands would allow us to quickly identify the failing area.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2365)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -2684,6 +2684,9 @@ impl Renderer {
             DebugCommand::EnableGpuSampleQueries(enable) => {
                 self.set_debug_flag(DebugFlags::GPU_SAMPLE_QUERIES, enable);
             }
+            DebugCommand::EnableDualSourceBlending(_) => {
+                panic!("Should be handled by render backend");
+            }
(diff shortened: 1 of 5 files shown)


webrender/src/renderer.rs
```diff
@@ -2861,7 +2875,7 @@ impl Renderer {
 
             for &mut (_, RenderedDocument { ref mut frame, .. }) in &mut active_documents {
                 self.prepare_gpu_cache(frame);
-                assert!(frame.gpu_cache_frame_id <= self.gpu_cache_frame_id);
+                //assert!(frame.gpu_cache_frame_id <= self.gpu_cache_frame_id);
```
glennw: I guess this is no longer valid? Should it be removed?

kvark: Ah, you caught me!
Problem is that this assertion would false-fully trigger when replaying any capture made between the introduction of the assert and #2355. So I temporarily (!) commented it out when I was looking at a problem, and forgot about that when preparing the PR.

kvark:
Should be fixed now :)

PR #2366, opened 2018-01-31
A couple more small tidy ups in the prim prepare code.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2366)


webrender/src/frame_builder.rs
```diff
@@ -144,12 +144,14 @@ pub struct FrameState<'a> {
...
 
-pub struct PictureContext {
+pub struct PictureContext<'a> {
     pub pipeline_id: PipelineId,
-    pub pic_index: SpecificPrimitiveIndex,
     pub perform_culling: bool,
     pub prim_runs: Vec<PrimitiveRun>,
     pub original_reference_frame_id: Option<ClipId>,
+    pub display_list: &'a BuiltDisplayList,
+    pub draw_text_transformed: bool,
```
kvark: is it always equal to `self.original_reference_frame_id.is_some()`?

glennw: For now, yes. That won't be true in the future though (once we expose to the API whether a stacking context should be drawn in screen or local space).

PR #2367, opened 2018-01-31
Make hit testing consistent between frames

### LINKED ISSUE #2099: Make hit-testing more robust to timing irregularities

mrobinson:
Move hit testing data structures to the Document. This makes hit testing
results consistent with the currently rendered frame. To do this we are
required to copy important information from the ClipScrollTree during
the final step of producing a RenderedDocument, so this implies a
certain per-frame overhead. In the future we can look to make this less
expensive.

Fixes #2099.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2367)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -2,9 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{BorderRadius, ClipMode, ComplexClipRegion, DeviceIntRect, ImageMask, ImageRendering};
-use api::{LayerPoint, LayerRect, LayoutPoint, LayoutVector2D, LocalClip};
-use api::{DevicePixelScale, LayerToWorldTransform};
+use api::{BorderRadius, ClipMode, ComplexClipRegion, DeviceIntRect, DevicePixelScale, ImageMask};
+use api::{ImageRendering, LayerRect, LayerToWorldTransform, LayoutPoint, LayoutVector2D};
+use api::LocalClip;
 use border::{BorderCornerClipSource, ensure_no_corner_overlap};
 use ellipse::Ellipse;
 use freelist::{FreeList, FreeListHandle, WeakFreeListHandle};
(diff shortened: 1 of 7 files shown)


glennw:
@mrobinson Did you want this reviewed now, or not until the [DO NOT MERGE] is removed?

mrobinson:
@glennw Just looking for general feedback about my approach now, but I don't feel comfortable landing this until I've done more thorough testing with Gecko.

mrobinson:
It seems that this change exposes a shortcoming in the Gecko test harness. Since this fixes one test and has the potential to remove a stall from hit testing, @staktrace suggested that we land it and just skip `dom/events/test/test_bug946632.html`. @glennw, this PR should be ready to review and land.

glennw:
Reviewed 6 of 7 files at r1.
Review status: 6 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/hit_test.rs, line 23 at r1](https://reviewable.io:443/reviews/servo/webrender/2367#-L4Ib9f5AllEirmbYmfH:-L4Ib9f6MRKrs5Dq8Q1n:btvbbwe) ([raw file](https://github.com/servo/webrender/blob/ff703196fe23bf3be36f88d8cbdea0a2fe71ae71/webrender/src/hit_test.rs#L23)):*
> ```Rust
>     /// A particular point must be inside all of these regions to be considered clipped in
>     /// for the purposes of a hit test.
>     regions: Option<Vec<HitTestRegion>>,
> ```

Is there a benefit to having this be an Option, as opposed to just an empty Vec if no regions present?

---

*[webrender/src/hit_test.rs, line 99 at r1](https://reviewable.io:443/reviews/servo/webrender/2367#-L4IbLrbcdcDAgv_mSz0:-L4IbLrbcdcDAgv_mSz1:bye4yas) ([raw file](https://github.com/servo/webrender/blob/ff703196fe23bf3be36f88d8cbdea0a2fe71ae71/webrender/src/hit_test.rs#L99)):*
> ```Rust
> 
>         for (id, node) in &clip_scroll_tree.nodes {
>             // XXX: Need to read ClipSources here and generated regions.
> ```

Could you expand on this - when / what is that needed for?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2367)*

glennw:
Reviewed 1 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2367)*

mrobinson:

---

*[webrender/src/hit_test.rs, line 23 at r1](https://reviewable.io:443/reviews/servo/webrender/2367#-L4Ib9f5AllEirmbYmfH:-L4LH3JKYR-DeIrKC5xS:bbk65zz) ([raw file](https://github.com/servo/webrender/blob/ff703196fe23bf3be36f88d8cbdea0a2fe71ae71/webrender/src/hit_test.rs#L23)):*

Is there a benefit to having this be an Option, as opposed to just an empty Vec if no regions present?
</blockquote></details>

I don't think there is a big win here. The distinction between empty Clip nodes and other types of nodes isn't as important during hit testing. I'll just make this a regular vector.

---

*[webrender/src/hit_test.rs, line 99 at r1](https://reviewable.io:443/reviews/servo/webrender/2367#-L4IbLrbcdcDAgv_mSz0:-L4LHBwpmYrw7uSlSAuQ:bkft0bv) ([raw file](https://github.com/servo/webrender/blob/ff703196fe23bf3be36f88d8cbdea0a2fe71ae71/webrender/src/hit_test.rs#L99)):*

Could you expand on this - when / what is that needed for?
</blockquote></details>

Ack. This was a comment to myself that I forgot to remove. I'll remove it now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2367)*

PR #2368, opened 2018-01-31
Add a super-simple multi-window example.

glennw:
This just draws the same rect and text string in two windows for now.

We can expand on this to try and reproduce some of the multi-window
text corruption bugs we see intermittently in Gecko.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2368)

### DIFF
--- a/webrender/examples/multiwindow.rs
+++ b/webrender/examples/multiwindow.rs
@@ -1,0 +1,282 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+extern crate app_units;
+extern crate euclid;
+extern crate gleam;
+extern crate glutin;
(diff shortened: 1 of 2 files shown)


webrender/examples/multiwindow.rs
```diff
@@ -0,0 +1,282 @@
...
+            None,
+            MixBlendMode::Normal,
+            Vec::new(),
+        );
+
+        let info = LayoutPrimitiveInfo::new(LayoutRect::new(LayoutPoint::new(100.0, 100.0), LayoutSize::new(100.0, 200.0)));
+        builder.push_rect(&info, ColorF::new(0.0, 1.0, 0.0, 1.0));
+
+        let text_bounds = LayoutRect::new(LayoutPoint::new(100.0, 50.0), LayoutSize::new(700.0, 200.0));
+        let glyphs = vec![
```
kvark: let's remove the text from our basic example to avoid duplication?

glennw: Done

Issue #1611, opened 2017-08-24
panic in request_image()


jrmuizel:
This is with tiling turned on in Gecko:
https://github.com/servo/webrender/blob/1a5129396ad2fe9e34ff9ebae203d0119558b261/webrender/src/resource_cache.rs#L439

I'll try to investigate more into what's causing it.

glennw:
It sounds like tiling was enabled when the image template was created, but then an image was requested without any tile offset supplied.

jrmuizel:
It looks like this is caused by asking for tiling on an image that used for as a mask. I don't believe we support tiled masks yet so we end up panic. It probably wouldn't hurt to panic earlier or implement tiled masks, but in the mean time it's easy for Gecko to work around this.

Gankro:
bugzilled https://bugzilla.mozilla.org/show_bug.cgi?id=1435896

nical:
IIRC the associated gecko bug is closed (Bugzilla down right now). There is also an issue specifically for tiled masks.

PR #2375, opened 2018-02-02
[FIXME] rawtest for blob-image masks

Gankro:
Getting this rawtest to pass is hopefully the last blocker for turning blob-images on by default in WR. Currently it crashes:

```
thread 'WRRenderBackend#0' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
 [ED: removed noise]
  10: <core::option::Option<T>>::unwrap
             at /Users/travis/build/rust-lang/rust/src/libcore/macros.rs:20
  11: webrender::resource_cache::ResourceCache::request_image
             at webrender/src/resource_cache.rs:613
  12: webrender::clip::ClipSources::update
             at webrender/src/clip.rs:249
  13: webrender::clip_scroll_node::ClipScrollNode::update_clip_work_item
             at webrender/src/clip_scroll_node.rs:355
  14: webrender::clip_scroll_node::ClipScrollNode::update
             at webrender/src/clip_scroll_node.rs:328
  15: webrender::clip_scroll_tree::ClipScrollTree::update_node
             at webrender/src/clip_scroll_tree.rs:428
  16: webrender::clip_scroll_tree::ClipScrollTree::update_node
             at webrender/src/clip_scroll_tree.rs:449
  17: webrender::clip_scroll_tree::ClipScrollTree::update_node
             at webrender/src/clip_scroll_tree.rs:449
  18: webrender::clip_scroll_tree::ClipScrollTree::update_tree
             at webrender/src/clip_scroll_tree.rs:389
  19: webrender::frame_builder::FrameBuilder::build
             at webrender/src/frame_builder.rs:1765
  20: webrender::frame::FrameContext::build_rendered_document
             at webrender/src/frame.rs:1124
  21: webrender::render_backend::Document::render
             at webrender/src/render_backend.rs:141
  22: webrender::render_backend::RenderBackend::update_document
             at webrender/src/render_backend.rs:700
  23: webrender::render_backend::RenderBackend::run
             at webrender/src/render_backend.rs:648
  24: webrender::renderer::Renderer::new::{{closure}}
             at webrender/src/renderer.rs:2247
``` 

The root of the issue seems to be that [ClipSources::update](https://searchfox.org/mozilla-central/source/gfx/webrender/src/clip.rs#249) is passing [request_image](https://searchfox.org/mozilla-central/source/gfx/webrender/src/resource_cache.rs#613) a tile_offset of None, but the image template thinks we're tiled so we unwrap the tile_offset and die.

I'm not certain what the "correct" fix is here.

I also can't reliably trigger this crash in gecko, even though the rawtest seems to always fail.

### DIFF
--- a/wrench/src/rawtest.rs
+++ b/wrench/src/rawtest.rs
@@ -39,6 +39,7 @@ impl<'a> RawtestHarness<'a> {
     }
 
     pub fn run(mut self) {
+        self.test_tiled_blob_masks();
         self.test_retained_blob_images_test();
         self.test_blob_update_test();
         self.test_blob_update_epoch_test();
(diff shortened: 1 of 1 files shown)


Gankro:
Looks like this is #1611, and the discussion seems to suggest we should just try harder to prevent gecko from emitting this?

glennw:
We discussed this briefly in IRC. It seems that in the test case above, there's no need for tiling to be enabled (this should only occur when the texture size is larger than the supported hardware texture size). So, it may just be a bug in this instance (perhaps an interaction between blobs and tiled images?). We will want to support tiled image masks in the long term, but it may not be the underlying issue here.

staktrace:
Isn't this issue the corresponding github issue for https://bugzilla.mozilla.org/show_bug.cgi?id=1435896 ? Or are they different? If this is, then I'd like to leave it open since it blocks enabling the crashtest on the gecko side.

Gankro:
Use #1611 for that

PR #2382, opened 2018-02-06
Support partial rendering of off-screen pictures / render tasks.

glennw:
This allows drawing a render task where the allocated space in
the surface cache is smaller than the required size if we were
to draw the complete contents of a render task.

The primary motivation of this task is to fix some existing bugs.
Specifically, if an off-screen render task is intersecting with
the main screen rect, we would only allocate enough space in the
surface cache texture for the visible portion. Then, the render
task would be drawn, and in some cases could extend outside the
bounds of the allocated rect for the task. This can result in
image corruption - for instance, a long scrolling region which
is drawn to an off-screen target (due to a filter) could draw
outside its region, affecting other tasks in the same surface
cache texture layer.

A second motivation is as a performance and power saving optimization.
In the future, we may want to render and cache "tiles" that can be
cached and provided to the OS compositor interfaces, to avoid work
during scrolling.

Finally, the implementation also contains the infrastructure necessary
to enable us to completely skip texture-cache sub-render tasks when
a frame is re-rendered, which is a performance gain in some cases.

To achieve this:
 * Instead of one alpha-batcher per target, there is one per render task.
 * Each alpha-batcher tracks the task-relative, screen-space bounding
   rect of primitives added to it.
 * Once each alpha-batcher is complete, we check if the allocated rect
   for this task completely contains the combined bounding rect of the
   primitives inside it.
 * If it *does* contain it (common case) then we run a batch merging
   step where this set of batches is merged with any other alpha batch
   tasks that are also self-contained. This merging step can be very
   simple, since we know there is no overlap ordering constraints.
 * If it *doesn't* contain it, then we will submit this set of batches
   separately to the merged batch list, applying a scissor rect set to
   the size of the render task allocation.

The batching results should be as good or better than previously, except
in the genuine case where this fixes a bug, in which case there may be
a small number of extra draw calls.

Since the batching of each render task is independent, and also only
references read-only shared state, we can easily run each batch creation
task on a worker thread, if profiling shows any benefit to that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2382)

### DIFF
--- a/webrender/examples/common/boilerplate.rs
+++ b/webrender/examples/common/boilerplate.rs
@@ -226,13 +226,6 @@ pub fn main_wrapper<E: Example>(
                 glutin::Event::KeyboardInput(
                     glutin::ElementState::Pressed,
                     _,
-                    Some(glutin::VirtualKeyCode::B),
-                ) => {
-                    renderer.toggle_debug_flags(webrender::DebugFlags::ALPHA_PRIM_DBG);
-                }
-                glutin::Event::KeyboardInput(
-                    glutin::ElementState::Pressed,
-                    _,
                     Some(glutin::VirtualKeyCode::S),
                 ) => {
                     renderer.toggle_debug_flags(webrender::DebugFlags::COMPACT_PROFILER);
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -4,7 +4,7 @@
 
 use api::{AlphaType, DeviceIntRect, DeviceIntSize, LayerToWorldScale};
 use api::{DeviceUintRect, DeviceUintPoint, DeviceUintSize, ExternalImageType, FilterOp, ImageRendering, LayerRect};
-use api::{SubpixelDirection, YuvColorSpace, YuvFormat};
+use api::{DeviceIntPoint, LayerPoint, SubpixelDirection, YuvColorSpace, YuvFormat};
 use api::{LayerToWorldTransform, WorldPixel};
 use border::{BorderCornerInstance, BorderCornerSide, BorderEdgeKind};
 use clip::{ClipSource, ClipStore};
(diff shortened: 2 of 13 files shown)


glennw:
This has some outstanding Gecko try failures - I need to fix those before this is ready for review / merge.

This also removes the alpha primitive debug option - the information is no longer available in the renderer, and we can always view this in RenderDoc, Frame Analyzer etc now.

When the remaining issues are fixed, https://bugzilla.mozilla.org/show_bug.cgi?id=1429537 will be resolved.

glennw:
Gecko try:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=dcb7f90ba3475e3a8216cb58cf72b51e2c44e355

There's a couple of orange results, but they appear to be unrelated intermittents (although @staktrace may like to verify).

r? @kvark

staktrace:
Yup, the oranges on the try push are unrelated intermittents.

kvark:
The reasoning sounds good to me.
A few notes here since Reviewable appears to bug out and block me from adding some comments:
  - why is `AlphaBatchList` no longer serializable?
  - could we have a `Vec<Rect>` per batch (or just one rect?) as opposed to having a separate `item_rects: Vec<Vec<Rect>>`

---

Reviewed 12 of 13 files at r1.
Review status: 12 of 13 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[webrender/examples/common/boilerplate.rs, line 231 at r1](https://beta.reviewable.io/reviews/servo/webrender/2382#-L4lbzRjAtSydJOhrLet:-L4lbzRjAtSydJOhrLeu:b-2jlv2m) ([raw file](https://github.com/servo/webrender/blob/a4202af2130e999ead44e10d8eb4f323c0ddb582/webrender/examples/common/boilerplate.rs#L231)):*
> ```Rust
>                     Some(glutin::VirtualKeyCode::B),
>                 ) => {
>                     renderer.toggle_debug_flags(webrender::DebugFlags::ALPHA_PRIM_DBG);
> ```

why are we removing this debug toggle?

---

*[webrender/src/picture.rs, line 549 at r1](https://beta.reviewable.io/reviews/servo/webrender/2382#-L4lf8ww9Q5rbfUYPjJ6:-L4lf8wxEjh38aQb07aE:bmf6zqo) ([raw file](https://github.com/servo/webrender/blob/49df33244e6c7a7751f7b48f5870f0c3ba7de401/webrender/src/picture.rs#L549)):*
> ```Rust
> 
>                         let picture_task = RenderTask::new_picture(
>                             RenderTaskLocation::Dynamic(None, cache_size),
> ```

back to the future? :) I recall us changing from the `RenderTaskLocation`, and now it's back

---

*[webrender/src/renderer.rs, line 3851 at r1](https://beta.reviewable.io/reviews/servo/webrender/2382#-L4lgAiy5JD7UmQX7XAR:-L4lgAiy5JD7UmQX7XAS:bjfnl4s) ([raw file](https://github.com/servo/webrender/blob/49df33244e6c7a7751f7b48f5870f0c3ba7de401/webrender/src/renderer.rs#L3851)):*
> ```Rust
>             }
> 
>             if let Some(..) = alpha_batch_container.target_rect {
> ```

nit: use `is_some()`?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2382#-:-L4lgpt74Bpd5vpo83Fa:bf9vc8d)*

glennw:
@kvark Thanks for the review, answers below:

```AlphaBatchList``` is no longer serializable since it's never stored - it's only used during batch building. So I just made ```AlphaBatchContainer``` serializable, since that's what is built and stored in a frame.

The ```Vec<Vec<Rect>>``` code isn't ideal. I thought it might be a lesser evil than having a ```Vec<Rect>``` in the alpha batch, since that would make the merge step more complex / expensive. But it's definitely worth doing some experiments here. Having one rect altogether which is a union may well provide just as good batching as we get now, with less cost. We need to do some testing of this though.

The debug toggle is removed because we no longer have the screen-space bounding rects of alpha items. Once batches are merged, we no longer have the offsets to get the true bounding rect of each item. I figured we probably don't need this anymore, since we can use RenderDoc and friends, and since it would be more CPU cost / memory to re-calculate this information.

I updated the nit to use ```is_some()```.

Thanks!

kvark:
Ok, sounds good.
I think the alpha debug toggle was quite useful though in some cases, but I agree it's not required.
@bors-servo r+

PR #2390, opened 2018-02-07
Allow generate_checkerboard to create non-square images.

glennw:
This is a small change that will be useful for reftests with the
upcoming changes to how image tiling, repeat and spacing works.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2390)


glennw:
r? anyone

PR #2391, opened 2018-02-07
Don't request source image from texture cache, if pre-rendered cache is valid.

glennw:
When using pre-rendered image sources, there's no need to request the source image each frame. It just wastes room in the texture cache having the source image referenced if it's not being used. This is a minor win - but will become more important once we start handling image tiling internally within each image primitive.

Also tidied up the image request APIs slightly to pass a struct instead of various fields that make up a request.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2391)


glennw:
r? anyone

PR #2392, opened 2018-02-07
Add hit testing support for ClipChains from the API

### LINKED ISSUE #2383: hit-testing code doesn't handle ClipChainId properly
I mentioned this to @mrobinson on IRC yesterday but filing it here so we can track it better.

```
08:26:16 kats: mrobinson: it looks like the is_point_clipped_in_for_node function doesn't handle being passed a ClipChainId for the node_id parameter, is that right? It tries to look it up in `self.nodes` which afaik will never contain a ClipChainId
09:07:08 mrobinson: kats: Oh, I think you are right!
09:07:22 mrobinson: kats: I can write a fix for that pretty easily.
09:07:33 kats: mrobinson: great, thanks :)
```

mrobinson:
Hit testing now has support for tests against items that were clipped by
API-defined ClipChains. Both clipping ClipScrollNodes nodes and
ClipChains are converted to pairs of ClipScrollNodes and ClipChains for
purposes of the HitTester. This more-or-less matches the logical next
step in ClipScrollTree design and makes the logic for HitTesting
simpler.

Fixes #2383.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2392)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -43,10 +43,10 @@ impl CoordinateSystemId {
     }
 }
 
-struct ClipChainDescriptor {
-    id: ClipChainId,
-    parent: Option<ClipChainId>,
-    clips: Vec<ClipId>,
+pub struct ClipChainDescriptor {
+    pub id: ClipChainId,
+    pub parent: Option<ClipChainId>,
+    pub clips: Vec<ClipId>,
 }
 
 pub struct ClipScrollTree {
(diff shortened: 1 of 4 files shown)


mrobinson:
This change also adds a rawtest which tests hit testing. The tests aren't totally complete yet, but we can add more as we go along.

glennw:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2392)*

Issue #2384, opened 2018-02-06
fast text-shadows of non-black text have unreasonably incorrect subpixels


Gankro:
I suspect this is the result of the gamma fudging @lsalzman and I agreed on. I don't think we can ship this.

```yaml
---
root:
  items:
        -
          type: "shadow"
          bounds: [10, 14, 220, 200]
          blur-radius: 0
          offset: [0, 20]
          color: black
        -
          text: "wow this is really wrong"
          origin: [10, 44]
          size: 18
          color: red
          font: "VeraBd.ttf"
        -
          type: "pop-all-shadows"
```

![screen shot 2018-02-06 at 11 21 46 am](https://user-images.githubusercontent.com/1136864/35870569-fa295c92-0b2f-11e8-8861-6275406c637a.png)

glennw:
Fixed by #2393.

PR #2393, opened 2018-02-07
re-rasterize fast text shadows directly with the given shadow color

lsalzman:
It seems that in the wild people are expecting unblurred text shadows to be precisely the same as if the text is drawn with the shadow color. So long as we have to deal with subpixel anti-aliasing and preblending, it is not really possible to reuse the subpixel AA masks and satisfy those constraints. Thus, we have to go back to the old method of just re-rasterizing the text shadow AA masks with the font color being exactly the shadow color, or the preblending just gets messed up.

This fixes issue #2384 and also the downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1434029

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2393)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -773,7 +773,7 @@ impl AlphaBatcher {
                         debug_assert_ne!(texture_id, SourceTexture::Invalid);
 
                         // Ignore color and only sample alpha when shadowing.
-                        if text_cpu.is_shadow() {
+                        if text_cpu.shadow {
                             glyph_format = glyph_format.ignore_color();
                         }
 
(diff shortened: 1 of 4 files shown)


Gankro:
Just checking, this manages to do something special to color glyphs? (Emoji)

Gankro:
Also could you make [shadow-atomic](https://github.com/servo/webrender/blob/master/wrench/reftests/text/reftest.list#L23) strict again?

lsalzman:
This is why we still need to keep the shadow flag around on TextRunPrimitiveCpu, essentially to disambiguate that one remaining case. Otherwise, we wouldn't need it.

glennw:
@lsalzman  ```shadow-ordering``` is failing on mac only, for some reason. It was previously running on linux only, so it's probably fine to re-enable that as a platform specific test.

lsalzman:
Okay, back to linux-only for now then.

PR #2394, opened 2018-02-08
update to bincode 1.0.0

TyOverby:
This is a PR showing how to update to bincode 1.0.0

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2394)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,7 +63,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bincode"
-version = "0.9.0"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "byteorder 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -91,7 +91,7 @@ version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "byteorder 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "iovec 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "iovec 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -402,6 +402,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "fuchsia-zircon"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bitflags 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fuchsia-zircon-sys 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "fuchsia-zircon-sys"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "fxhash"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -537,7 +551,7 @@ dependencies = [
 
 [[package]]
 name = "iovec"
-version = "0.1.0"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "libc 0.2.20 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


Eijebong:
This should also update ipc-channel to 0.10 to avoid having two versions of bincode in servo

Cargo.lock
```diff
@@ -391,6 +391,20 @@ dependencies = [
  "servo-freetype-sys 4.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
+name = "fuchsia-zircon"
```
kvark: WR is going to ship parts of Fuchsia?

TyOverby: Looks like it's a new dependency of mio?

TyOverby: From Mio Cargo.toml

```toml
[target.'cfg(target_os = "fuchsia")'.dependencies]
fuchsia-zircon = "0.3.2"
fuchsia-zircon-sys = "0.3.2"
```

TyOverby:
@kvark, @Eijebong Everything look good?

Eijebong:
@TyOverby For me everything is looking good :+1: (can't r+) I'll do the necessary servo work once this lands.

Cargo.lock
kvark: It has `BSD-3-Clause` license, which is atypical for Rust ecosystem. Gotta check with Gecko folks before merging, since all dependencies are vendored for Gecko, and licenses matter.

Cargo.lock
```diff
@@ -830,6 +846,16 @@ dependencies = [
  "libc 0.2.20 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
```
kvark: why do we need `rand` now?

Eijebong: rand is already in the dependency tree, this is just a dupe because some dependencies haven't updated yet (I've started doing what's needed for that)

Cargo.lock
mbrubeck: These `fuchsia-zircon` crates are already in the Gecko source tree because the latest version of `rand` depends on them too.

Eijebong:
Anything I could do to get this moving again ?

TyOverby:
can I regenerate the Cargo.lock file by just running `cargo update` or is that a bad idea?

Eijebong:
Running a full cargo update is a pretty bad idea, you'll end up with a lot of unwanted stuff updated.
I usually reset the Cargo.lock to current master state and rerun my cargo update -p whatever command

TyOverby:
How do I know which packages to `cargo update -p`?

Eijebong:
Well you need to update `ipc-channel` and that's probably it (unless it won't update it in which case you'll have to add every ipc-channel dependency until it unlocks, it happens sometimes).

TyOverby:
```
Tys-MacBook-Pro-2:webrender tyoverby$ cargo update -p ipc-channel
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: failed to select a version for `mio` (required by `ipc-channel`):
all possible versions conflict with previously selected versions of `mio`
  version 0.6.7 in use by mio v0.6.7
  possible versions to select: 0.6.14, 0.6.13, 0.6.12, 0.6.11
Tys-MacBook-Pro-2:webrender tyoverby$ 
```

Eijebong:
Yeah, add mio to the updated crates list `cargo update -p ipc-channel -p mio`

TyOverby:
@Eijebong I did that, and it got down to `iovec` or something, and then adding `-p iovec` just gave me the same iovec error again

Eijebong:
@TyOverby Mhh going from master,

```
cargo update -p iovec -p mio -p ipc-channel -p bincode -p net2
cargo update -p mio --precise 0.6.12
```

I downgraded mio to avoid duping slab (I'll have a look at it later)

Eijebong:
Also it would probably be better if those commits were all squashed :)

glennw:
Is this ready for merge now? Do we need any sign-off to make sure that both Gecko and Servo are ready to get this merged?

staktrace:
Try push to make sure the gecko side is good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d7f41b5ad16de72866a76a08bbedabe0f7fcbc33

staktrace:
Looks like there's some crashes but it's from the `audioipc` update which I'll be pushing along with the WR update since they both bump bincode. So I have no objections to merging this PR from the gecko side, although I'll need to get help on the audioipc crashes before we can actually push the WR update into gecko (or just do WR alone which will end up with two vendored bincode versions).

Eijebong:
@staktrace https://github.com/djg/audioipc-2/pull/35/files#diff-f7e555878138f2fef856422bafa6439aR157 is the only thing that could crash in audioipc I think and I'm pretty sure that this unwrap is safe.
Do we have a backtrace or something like that ?

staktrace:
@Eijebong https://treeherder.mozilla.org/logviewer.html#?job_id=167943511&repo=try&lineNumber=9622 is the stacktrace. Note that the audioipc update includes other changes too, everything in the range 7e866e5ba304b271ab4c504a354ed7aa628343b8..b93386611d7d9689c4f0177a4704f0adc16bc2d1 is fair game.

Eijebong:
Oh... This is way too domain specific for me to help :(

staktrace:
Oh, the crash is probably from https://github.com/djg/audioipc-2/commit/ead320ae24fdf8540efa909f35bec9b78925186a which needs a gecko-side change. But that landed after the bincode update so I can just update audioipc to the bincode update and skip the audioipc_client_init change if needed.

glennw:
Could we get this squashed into one commit (makes it easier to revert if we should ever need to), and then this is ready to merge.

TyOverby:
@glennw: doesn't github have a "merge and squash" feature?

Eijebong:
@TyOverby It does but they're using bors here to merge PRs instead of the github interface

TyOverby:
If I squash `git rebase master -i`, I have to go back and redo all my conflict resolutions.  Is there a way to use git to avoid this?

TyOverby:
apparently `git reset --soft master && git commit` will do a squash without any weird resolution issues.

glennw:
Thanks!

@bors-servo r+

glennw:
Blocked by #2522.

Eijebong:
Need to add rand to the list of duped dependencies

staktrace:
I rebased this patch, it's up on #2547 now.

glennw:
Closing in favor of #2547.

PR #2397, opened 2018-02-08
Rename point and line in GLSL

kvark:
... to make our shaders compatible with HLSL
Also see https://github.com/zakorgy/webrender/commit/58f0a283b44922f3d69aee87bcb2595e20c096e8
cc @zakorgy

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2397)

### DIFF
--- a/webrender/res/brush_line.glsl
+++ b/webrender/res/brush_line.glsl
@@ -40,20 +40,21 @@ void brush_vs(
 ) {
     vLocalPos = local_pos;
 
-    Line line = fetch_line(prim_address);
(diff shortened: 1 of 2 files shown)


zakorgy:
@kvark I think the `point` variable should be updated [here](https://github.com/kvark/webrender/blob/5cf3fa4d11fdc9ba050766bc325c82fb272c3325/webrender/res/prim_shared.glsl#L484) as well.

kvark:
@zakorgy thanks for the heads up! I had the change locally, but slipped on including it in the commit :)

PR #2398, opened 2018-02-08
Fix perspective-origin reftest.

### LINKED ISSUE #2379: Rendering issue with perspective + images.
In the ```perspective-origin.yaml``` test case from https://github.com/servo/webrender/pull/2378, the displayed image should contain the red border around the edges. However, it extends past that by a few pixels.

Note to self: this might be fixed by the planned work to pre-render image repeats and spacing. That will remove that code from the image shader, which I think is wrong and can produce artifacts like this.

glennw:
I was misinterpreting the output of this test - I hadn't realized
that in wrench the stretch-size defaults to the image dimensions,
rather than the local rect dimensions. I modified the test to
take account of this and it draws correctly now.

Fixes #2379.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2398)


glennw:
r? @kvark or anyone

I temporarily removed the fuzziness from this test, to see what CI fails with. It passes locally on my machine with no fuzziness now. We might still need a fuzz setting on CI, but it will hopefully be less.

PR #2401, opened 2018-02-09
Port ps_blend to be a brush shader.

glennw:
Previously, ps_blend applied the blend operation
in 2D screen-space. Now, blends are performed as
a normal brush shader. This provides several benefits:

* For a transformed primitive, we blend fewer pixels,
  since we only draw the composite step as a transformed
  primitive, rather than a 2D screen aligned rect.

* We can optionally apply all the usual features
  that are part of the brush shader, specifically
  clip masks, edge AA and pixel snapping.

* This lays the groundwork for supporting picture
  targets that are selectable to be either rasterized
  in local-space or screen-space, depending on what
  the caller requests as a performance / quality
  trade-off.

Next, we'll port hw_composite and composite shaders
to use the same brush format, and add support for
applying the source texture based on whether it was
rasterized in local or screen space.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2401)


glennw:
This is not quite ready yet:
 * There is a wrench reftest that fails with a pixel difference of 17 in places, when it was pixel-perfect previously. Possibly an accuracy bug somewhere.
 * Still needs a gecko try once that bug is fixed.

But I don't expect those to change the PR massively, so it can probably be reviewed when someone has time.

kvark:
:lgtm: 

---

---

*[webrender/src/picture.rs, line 589 at r1](https://reviewable.io:443/reviews/servo/webrender/2401#-L520wLPfypXx_zAfIiZ:-L520wLQxvSu2cYg9q2Z:b-9oogrf) ([raw file](https://github.com/servo/webrender/blob/0fd3fc9209c2584465d3ad15ad2c87395ad5eb92/webrender/src/picture.rs#L589)):*
> ```Rust
> 
>     pub fn write_gpu_blocks(&self, request: &mut GpuDataRequest) {
>         // TODO(gw): It's unfortunate that we pay a fixed cost
> ```

we could have an extra GPU block range for the color matrix specifically

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2401#-:-L521CU_A-Kci2nMdvEj:bg7igkv)*

glennw:
Gecko try looks good for this patch:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=c0b581040687afa53d6214015adc1b7786c71c0b

(apart from some unrelated intermittent failures).

I changed the failing wrench reftest to have a much larger fuzziness value. The issue here is that the text run caching shader doesn't apply any form of snapping, which can result in issues like this. We can fix this separately.

glennw:
Hmmm, a couple of mac-specific failures. I'll investigate those further - maybe there is an accuracy issue somewhere that's not being picked up by the gecko tests.

glennw:
Rebased and added a commit that (I think) fixes the accuracy issues. It drops the max difference in the wrench tests from 17 down to 1, which is what I'd expect. Still need to do a gecko try to see if it breaks anything there.

glennw:
@kvark OK, I think this is finally ready to go (after review of the most recent commit).

The gecko try looks good now:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=5c9acdce54c6958eabea3777b288c4d55f9788e7

There is one failure (expected due to https://github.com/servo/webrender/pull/2408) and one unrelated intermittent failure.

PR #2403, opened 2018-02-09
Tidy more files

mrobinson:
Expand servo-tidy to include webrender_api/ and webrender/ (except the
main source directory). The ordering rules for use statements seem a
little capricious, but at least these will be in line with how Servo is
formatted.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2403)


webrender/examples/blob.rs
```diff
@@ -11,14 +11,12 @@ extern crate webrender;
 mod boilerplate;
 
 use boilerplate::{Example, HandyDandyRectBuilder};
-use rayon::Configuration as ThreadPoolConfig;
-use rayon::ThreadPool;
+use rayon::{Configuration as ThreadPoolConfig, ThreadPool};
```
kvark: we really need to ensure that rustfmt is in agreement with servo tidy here, could you verify?

mrobinson: I ran rustfmt nightly on blob.rs and it make some changes. Those changes still ran cleanly through servo-tidy, so I think it's just that servo-tidy is more stringent than rustfmt currently.

kvark:
@nical are you worried about collisions with #2362 here? Would you prefer this to be delayed then? I assume re-running servo-tidy is simpler than resolving the merge conflicts.

mrobinson:
@kvark I'm happy to wait until #2362 has landed, I just confirmed that it rebases cleanly on top of this change.

kvark:
ok, thanks @mrobinson !
@bors-servo r+

mrobinson:
@kvark Thanks for the review!

PR #2404, opened 2018-02-09
Remove an empty api.rs file

mrobinson:
This was probably checked in by mistake.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2404)


kvark:
@bors-servo r+
the less APIs, the better

PR #2405, opened 2018-02-09
Actually apply the local clip_rect to box_shadows 

Gankro:
See commits for details

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2405)

### DIFF
--- a/webrender/src/box_shadow.rs
+++ b/webrender/src/box_shadow.rs
@@ -88,7 +88,8 @@ impl FrameBuilder {
             if box_offset.x == 0.0 && box_offset.y == 0.0 && spread_amount == 0.0 {
                 return;
             }
-            let mut clips = Vec::new();
+            let mut clips = Vec::with_capacity(2);
+            clips.push(ClipSource::Rectangle(*prim_info.local_clip.clip_rect()));
 
             let fast_info = match clip_mode {
                 BoxShadowClipMode::Outset => {
@@ -264,8 +265,10 @@ impl FrameBuilder {
                         border_radius,
                         ClipMode::ClipOut,
                     ));
-
-                    let pic_info = LayerPrimitiveInfo::new(pic_rect);
+                    let pic_info = LayerPrimitiveInfo::with_clip_rect(
+                        pic_rect,
+                        *prim_info.local_clip.clip_rect()
+                    );
                     self.add_primitive(
                         clip_and_scroll,
                         &pic_info,
@@ -332,6 +335,12 @@ impl FrameBuilder {
                         clip_and_scroll
                     );
 
+                    let clip_rect = prim_info.local_clip.clip_rect();
+                    let clip_rect = match prim_info.rect.intersection(clip_rect) {
+                        Some(clip_rect) => clip_rect,
+                        None => return,
+                    };
+
                     // Draw the picture one pixel outside the original
                     // rect to account for the inflate above. This
                     // extra edge will be clipped by the local clip
@@ -339,7 +348,7 @@ impl FrameBuilder {
                     let pic_rect = prim_info.rect.inflate(inflate_size + box_offset.x.abs(), inflate_size + box_offset.y.abs());
                     let pic_info = LayerPrimitiveInfo::with_clip_rect(
                         pic_rect,
-                        prim_info.rect
+                        clip_rect
                     );
 
                     // Add a normal clip to ensure nothing gets drawn
--- a/wrench/src/yaml_frame_reader.rs
+++ b/wrench/src/yaml_frame_reader.rs
@@ -1216,7 +1216,12 @@ impl YamlFrameReader {
         wrench: &mut Wrench,
         yaml: &Yaml,
     ) {
-        let full_clip = LayoutRect::new(LayoutPoint::zero(), wrench.window_size_f32());
+        // A very large number (but safely far away from finite limits of f32)
+        let big_number = 1.0e30;
+        // A rect that should in practical terms serve as a no-op for clipping
+        let full_clip = LayoutRect::new(
+            LayoutPoint::new(-big_number / 2.0, -big_number / 2.0),
+            LayoutSize::new(big_number, big_number));
 
         for item in yaml.as_vec().unwrap() {
             // an explicit type can be skipped with some shorthand
--- a/wrench/reftests/boxshadow/box-shadow-clip-ref.yaml
+++ b/wrench/reftests/boxshadow/box-shadow-clip-ref.yaml
@@ -1,0 +1,40 @@
+---
+root:
+  items:
+    - type: stacking-context
+      bounds: [0, 0, 1000, 1000]
+      items:
+        - type: clip
+          bounds: [0, 0, 100, 200]
+          items:
+            - type: box-shadow
+              bounds: [ 20, 20, 200, 80 ]
+              color: blue
+              offset: [10, 20]
+              clip-mode: outset
+        - type: clip
+          bounds: [200, 0, 100, 200]
+          items:
+            - type: box-shadow
+              bounds: [ 220, 20, 200, 80 ]
+              color: red
+              offset: [10, 20]
+              blur-radius: 10
+              clip-mode: outset
+        - type: clip
+          bounds: [0, 200, 100, 200]
+          items:
+            - type: box-shadow
+              bounds: [ 20, 220, 200, 80 ]
+              color: green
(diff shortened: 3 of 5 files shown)


Gankro:
(should probably do a try build for this...)

Gankro:
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=23edb90fccda624aa766f45e58f2e7c9c7aaa32a

Gankro:
Try build looks good, trivial fuzzyness changes (cc @staktrace here's your warning if you see those)

wrench/src/yaml_frame_reader.rs
```diff
@@ -1216,7 +1216,9 @@ impl YamlFrameReader {
         wrench: &mut Wrench,
         yaml: &Yaml,
     ) {
-        let full_clip = LayoutRect::new(LayoutPoint::zero(), wrench.window_size_f32());
+        // The window plus some fudging for nesting in stacking contexts
+        // FIXME: compute this properly?
+        let full_clip = LayoutRect::new(LayoutPoint::zero(), wrench.window_size_f32()).inflate(100.0, 100.0);
```
kvark: would be great to resolve this FIXME

webrender/src/box_shadow.rs
```diff
@@ -88,7 +88,8 @@ impl FrameBuilder {
             if box_offset.x == 0.0 && box_offset.y == 0.0 && spread_amount == 0.0 {
                 return;
             }
-            let mut clips = Vec::new();
+            let mut clips = Vec::with_capacity(2);
+            clips.push(ClipSource::Rectangle(*prim_info.local_clip.clip_rect()));
```
kvark: Not sure I understand this part. It looks like we are both adding the local clip rect to the clip list as well as the picture info and layer primitive info for box shadows. Sounds redundant. Please clarify.

Gankro: This is in a completely separate branch where using ClipSources *does* work for whatever reason.

wrench/src/yaml_frame_reader.rs
Gankro: This is a bit tricky, since aiui we can't just keep a running sum of the parent offsets. If a parent has a transform like a rotation we need to, like, compute rotated bounding boxes, I think? Maybe it's simpler than I think though.

glennw:
Reviewed 5 of 5 files at r1.

---

*[webrender/src/box_shadow.rs, line 270 at r1](https://reviewable.io:443/reviews/servo/webrender/2405#-L56aoEApS4kQi_DJptz:-L56aoEApS4kQi_DJpu-:b-8hcgf1) ([raw file](https://github.com/servo/webrender/blob/e69bb38ed3b52191b0a20595e760d1bcb3550cf2/webrender/src/box_shadow.rs#L270)):*
> ```Rust
>                     let pic_info = LayerPrimitiveInfo::with_clip_rect(
>                         pic_rect,
>                         *prim_info.local_clip.clip_rect()
> ```

Is there a reason we do the clip rect intersection below for inset, but not for here? Not a big deal either way, just seems inconsistent?

---

*[webrender/src/box_shadow.rs, line 338 at r1](https://reviewable.io:443/reviews/servo/webrender/2405#-L56aTvuQwhPBycRWW9K:-L56aTvuQwhPBycRWW9L:b-xvcb58) ([raw file](https://github.com/servo/webrender/blob/e69bb38ed3b52191b0a20595e760d1bcb3550cf2/webrender/src/box_shadow.rs#L338)):*
> ```Rust
>                     );
> 
>                     let clip_rect = prim_info.rect.intersection(prim_info.local_clip.clip_rect());
> ```

Is:

```let clip_rect = match intersection() { Some(clip_rect) => clip_rect, None => return, }```

tidier, to avoid the extra unwrap() call below?
```


---

*[wrench/src/yaml_frame_reader.rs, line 1221 at r1](https://reviewable.io:443/reviews/servo/webrender/2405#-L4vzwSaGxNK4RuSS7Qg-r1-1221:-L56ZfWf9n1PKDRDxVrI:b-i5up91) ([raw file](https://github.com/servo/webrender/blob/e69bb38ed3b52191b0a20595e760d1bcb3550cf2/wrench/src/yaml_frame_reader.rs#L1221)):*

This is a bit tricky, since aiui we can't just keep a running sum of the parent offsets. If a parent has a transform like a rotation we need to, like, compute rotated bounding boxes, I think? Maybe it's simpler than I think though.
</blockquote></details>

I think we could basically make this ```rect::max()```. The goal is really to say that we don't want to supply a clip for this display item, since none is provided. WR internally already clips / culls to the visible viewport each frame, so there's no performance implications to doing that when you don't supply a clip.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2405)*
<!-- Sent from Reviewable.io -->

glennw:
Wow, not sure what got messed up with the reviewable output above - the bit at the bottom is a comment about the change in yaml reader, which says:

I think we could basically make this ```rect::max()```. The goal is really to say that we don't want to supply a clip for this display item, since none is provided. WR internally already clips / culls to the visible viewport each frame, so there's no performance implications to doing that when you don't supply a clip.

Gankro:

---

*[webrender/src/box_shadow.rs, line 270 at r1](https://reviewable.io:443/reviews/servo/webrender/2405#-L56aoEApS4kQi_DJptz:-L5AlgeJito3Rs8WoKkf:b-m74qiv) ([raw file](https://github.com/servo/webrender/blob/e69bb38ed3b52191b0a20595e760d1bcb3550cf2/webrender/src/box_shadow.rs#L270)):*

Is there a reason we do the clip rect intersection below for inset, but not for here? Not a big deal either way, just seems inconsistent?
</blockquote></details>

Because in this branch we were already using a clip_rect, but the other branch wasn't. I cannot say why the existing code was the way it is, but I'm just preserving the existing semantics as much as possible.

---

*[webrender/src/box_shadow.rs, line 338 at r1](https://reviewable.io:443/reviews/servo/webrender/2405#-L56aTvuQwhPBycRWW9K:-L5Alz-ZMeUaQcN_7mr2:brgdc3s) ([raw file](https://github.com/servo/webrender/blob/e69bb38ed3b52191b0a20595e760d1bcb3550cf2/webrender/src/box_shadow.rs#L338)):*

Is:

```let clip_rect = match intersection() { Some(clip_rect) => clip_rect, None => return, }```

tidier, to avoid the extra unwrap() call below?

</blockquote></details>

I don't have a strong opinion, happy to change if you or kvark does.

---

*[wrench/src/yaml_frame_reader.rs, line 1221 at r1](https://reviewable.io:443/reviews/servo/webrender/2405#-L4vzwSaGxNK4RuSS7Qg-r1-1221:-L5Am6zgyyKFWji47xa5:b-a7qbsw) ([raw file](https://github.com/servo/webrender/blob/e69bb38ed3b52191b0a20595e760d1bcb3550cf2/wrench/src/yaml_frame_reader.rs#L1221)):*

I think we could basically make this ```rect::max()```. The goal is really to say that we don't want to supply a clip for this display item, since none is provided. WR internally already clips / culls to the visible viewport each frame, so there's no performance implications to doing that when you don't supply a clip.
</blockquote></details>

I was a bit worried about a max rectangle causing overflow bugs (might get expand()ed?), but I can try if you think it'll be fine...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2405)*

glennw:
No need to make it absolute max representable rect - it can just be a really big rect, with a comment explaining that it's effectively a nop?

kvark:
Reviewed 4 of 5 files at r1.

---

*[webrender/src/box_shadow.rs, line 338 at r1](https://reviewable.io:443/reviews/servo/webrender/2405#-L56aTvuQwhPBycRWW9K:-L5FV-dczcpWbkSTavgu:b8dzsuy) ([raw file](https://github.com/servo/webrender/blob/e69bb38ed3b52191b0a20595e760d1bcb3550cf2/webrender/src/box_shadow.rs#L338)):*

I don't have a strong opinion, happy to change if you or kvark does.
</blockquote></details>

I agree with Glenn, although it's not a blocker :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2405)*

Gankro:
Comments addressed (unwrap replaced with match, wrench changed to use massive rect).

glennw:
Looks good, thanks!

@bors-servo r+ :rocket:

PR #2408, opened 2018-02-12
Use mipmaps and trilinear filtering for large, standalone textures.

### LINKED ISSUE #2023: Heavily-downscaled images look bad
https://bugzilla.mozilla.org/show_bug.cgi?id=1405469

This can be seen with the header image at [https://www.reddit.com/r/firefox/](https://www.reddit.com/r/firefox/): The header [image file](https://b.thumbs.redditmedia.com/yRlXvjV2ZFDLGkklcn78T8KgfHS65c4X0bNID4p68UU.png) is 3,242px × 833px big and gets displayed at 175px x 45px, which is a downscale factor of 18.5 (or 9.25 on a HiDPI@2x display).

The downscaled image looks very pixellated with WebRender.

Maybe WebRender should be generating different mip map levels as required, using subsequent 0.5 scaling so that the pixellation doesn't occur?

glennw:
Images that are large (currently this means one dimension > 512) are
placed into standalone textures rather than allocated into the
shared texture cache pages. Enabling mipmaps and trilinear filtering
on this subset of images is very simple, and also fixes the majority
of images where lack of mipmaps are a noticeable quality issue.

In the future, we can expand this to support mipmaps within the
shared cache pages, if desired, by manually drawing the mips with
a shader and making use of the mip-bias parameter to disable mip
selection for shared cache regions without mips.

Fixes #2023.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2408)


glennw:
@staktrace The xul failure should be fixed by https://github.com/servo/webrender/pull/2409. I think we can fuzz the other failure.

PR #2409, opened 2018-02-13
Only apply mip-maps if both image dimensions are > 512.

glennw:
This is a hack / workaround for some test failures in gecko.

There are some test images with resolution 2048x32. These were
having mips enabled, but then failing reftests since the tests
were expecting bilinear filtering only. When these images were
being drawn with a size of 4x4 pixels, the smallest mip was
being selected, which is not what the test expects.

This workaround is a reasonable tradeoff for now - we get mip
mapping on large images, while allowing the existing tests to
pass.

In the future, we could consider some alternatives (such as a better
heuristic for enabling mips, or a different mip algorithm).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2409)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -904,7 +904,9 @@ impl ResourceCache {
                     // that is > 512 in either dimension, so it should cover
                     // the most important use cases. We may want to support
                     // mip-maps on shared cache items in the future.
-                    if !self.texture_cache.is_allowed_in_shared_cache(
+                    if descriptor.width > 512 &&
+                       descriptor.height > 512 &&
+                       !self.texture_cache.is_allowed_in_shared_cache(
                         TextureFilter::Linear,
                         &descriptor,
                     ) {


glennw:
r? @staktrace 

(This should fix the test issue - just building it locally in Gecko to confirm).

glennw:
Confirmed that the test in question passes in gecko with this patch.

mstange:
In this case, with a 1/128 x 1/2 scale, ideally you'd want horizontal mip mapping but no vertical mip mapping.

glennw:
@mstange Yup - the initial implementation just allows the driver to build the mips. We can certainly add manual mip creation in the future though.

staktrace:
Try push with this PR looks good, the R3 failure is gone. I'll fuzz the R8.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=2425a4a7e5f6ba268c41c5eee3eca310ce603d5a

r+ from me, but I'm not qualified to actually review the code change.

PR #2410, opened 2018-02-13
Remove wrench's scrolling workaround.

nical:
This PR makes wrench build a single transaction with the display list and scrolling information, and remove the workaround that makes wrench ignore transactions containing scrolling.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2410)

### DIFF
--- a/wrench/src/main.rs
+++ b/wrench/src/main.rs
@@ -349,8 +349,8 @@ impl RenderNotifier for Notifier {
         self.tx.send(NotifierEvent::ShutDown).unwrap();
     }
 
-    fn new_document_ready(&self, _: DocumentId, scrolled: bool, _composite_needed: bool) {
-        if !scrolled {
+    fn new_document_ready(&self, _: DocumentId, _scrolled: bool, composite_needed: bool) {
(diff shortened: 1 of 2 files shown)


nical:
Ah! these two scrolling reftests are still failing because of how transactions are processed in the render backend: we go through all DocumentMsg's and wait until the end to decide whether we want to build the scene. It's great because it avoids building the scene multiple times in the same transaction, but the problem is that scrolling messages which are processed before mutate the clip scroll tree, which is then over-written when we build the scene. oops.

This is actually fixed in the async building PR (It was obvious that it had to be addressed with the async stuff but it didn't occur to me that we had the problem with non-async scenes too). I can try to come up with a fix that won't bitrot the whole async scene PR, but I'm rather tempted to just wait until it lands.

glennw:
@nical What's the status of this PR? Should we keep it open?

glennw:
@nical ping about the status of this PR

nical:
Woops! I forgot this hadn't landed yet. I just rebased the PR, it should be good to go now.

PR #2412, opened 2018-02-13
Support all image sampler kinds with brush_image implementation.

glennw:
As a follow up, we need to add better test coverage of these other
image types via a rawtest.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2412)

### DIFF
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -22,7 +22,13 @@ void brush_vs(
     ivec2 user_data,
     PictureTask pic_task
 ) {
-    vec2 texture_size = vec2(textureSize(sColor0, 0).xy);
+    // If this is in WR_FEATURE_TEXTURE_RECT mode, the rect and size use
+    // non-normalized texture coordinates.
+#ifdef WR_FEATURE_TEXTURE_RECT
+    vec2 texture_size = vec2(1, 1);
+#else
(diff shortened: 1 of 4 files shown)


glennw:
I haven't tested this in a try build yet - will be able to kick one off in an hour or so.

r? @kvark cc @staktrace

webrender/res/brush_image.glsl
```diff
@@ -46,7 +52,7 @@ void brush_vs(
 vec4 brush_fs() {
     vec2 uv = clamp(vUv.xy, vUvBounds.xy, vUvBounds.zw);
 
-    vec4 color = texture(sColor0, vec3(uv, vUv.z));
+    vec4 color = TEX_SAMPLE(sColor0, vec3(uv, vUv.z));
```
kvark: Actually, this would be incorrect, given that we want to use the mipmaps, and this macro always picks lod of 0.0

glennw: Good catch, fixed.

kvark:
Thanks Glenn! I'd like to see a try push, unless @kats wants to get this in quickly and bear with consequences :)

staktrace:
At the moment there's a big backlog of windows jobs in gecko CI, so I'm ok with landing this without a try push. If anything turns up I'll let you know.

PR #2413, opened 2018-02-13
Aggressively re-use render targets in the pool

kvark:
Note: this is not tested on Servo/Gecko yet, but conceptually should be safe. Main problem about this improvement is coming up with a good test case / metric that would show it's efficiency.

Basically, instead of pre-allocating all the targets at the start of the frame, we only allocated them at rendering a pass. Since we return the targets back to the pool at the end of the pass, what we want to see is some of the targets being re-used multiple times per frame (in different passes).
This could be improved if, instead of finding an exact match by size, we could fall-back to finding a target that may be slightly larger.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2413)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -15,7 +15,7 @@ use gpu_cache::{GpuCache, GpuCacheAddress};
 use gpu_types::{BrushImageKind, BrushInstance, ClipChainRectIndex};
 use gpu_types::{ClipMaskInstance, ClipScrollNodeIndex};
 use gpu_types::{CompositePrimitiveInstance, PrimitiveInstance, SimplePrimitiveInstance};
-use internal_types::{FastHashMap, SourceTexture};
+use internal_types::{FastHashMap, SavedTargetIndex, SourceTexture};
 use picture::{ContentOrigin, PictureCompositeMode, PictureKind, PicturePrimitive, PictureSurface};
 use plane_split::{BspSplitter, Polygon, Splitter};
 use prim_store::{ImageSource, PrimitiveIndex, PrimitiveKind, PrimitiveMetadata, PrimitiveStore};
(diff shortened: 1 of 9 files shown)


glennw:
Looks like a few CI test failures in wrench here.

kvark:
Heh, I believe this doesn't work (yet) because we have a bit of logic that assumes all rendered targets keep their contents till the end of the frame. Namely, `pass_rgba8_textures` and `pass_a8_textures` lists. I think we can make this nicer.

kvark:
In this PR, instead of assuming all passes are permanent and juggling with indices into the RT pool, we only index and keep specific target lists that have tasks marked for saving with `RenderTask::saved_index`. All the other ("regular") target lists can be instantly re-used by subsequent passes and/or frames.

kvark:
A small follow-up is needed to prevent target re-allocation within the frame between multiple documents. I think we can just cache the frame ID in the render target list and force an allocation of new targets instead of changing the old ones if they were used successfully this frame.

kvark:
Alright, the followup is here. @glennw should be ready for review, but not urgent :)

glennw:
@kvark I'll try to take a look through this today :) Semi-related - it'd be great sometime to have a debug output displaying how many render targets are being allocated per-frame, or something similar. I'm not really sure how or where would be best to display this though...

glennw:
Reviewed 6 of 7 files at r2, 2 of 2 files at r3.

---

*[webrender/src/tiling.rs, line 805 at r3](https://reviewable.io:443/reviews/servo/webrender/2413#-L5M3WTZ3FWk_jn7cYgi:-L5M3WTZ3FWk_jn7cYgj:bbu2q7v) ([raw file](https://github.com/servo/webrender/blob/62df3b40c2483e11354e6b901d1a7f29b0b9b5bd/webrender/src/tiling.rs#L805)):*
> ```Rust
>             }
>             RenderPassKind::OffScreen { ref mut color, ref mut alpha, ref mut texture_cache } => {
>                 let is_shared_alpha = self.tasks.iter().any(|&task_id| {
> ```

Instead of multiple iterations of the task list, can we accumulate those flags below as we add tasks to batches?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2413)*

kvark:
Pending try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=b7962c5f79f7ba7d734ce63d895c9b2cf9d578f1

kvark:
The try push looks green-ish, although it wouldn't harm to confirm (@staktrace ).
I've added the RT changes info to profiling dashboard and confirmed it behaves as designed in Gecko. Also ported the visible primitive counts to use the bar. That last commit 16c3aa212c3df24e10060af9b543197aa5a91e9c needs to be reviewed before merging (r? anyone).

staktrace:
Try push LGTM

kvark:
as per IRC:
@bors-servo r=glennw

PR #2414, opened 2018-02-13
Default case for shader matches

kvark:
Addresses https://github.com/servo/webrender/wiki/Driver-issues#lack-of-default-case-in-a-switch
It's not something we **have to** do. An alternative would be to try providing upstream fixes to SPIRV-Cross, but I'm worried about the quality of my sleep given that our shaders expose UB whenever an unexpected switch value comes in.
cc @zakorgy

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2414)


glennw:
Seems reasonable.

@bors-servo r+

PR #2415, opened 2018-02-14
Pack extra user data value into BrushInstance structure.

glennw:
In order to port the YUV image shader, we will need three
user data fields (one per channel).

To enable this work, we now pack the two render task
addresses for the picture and clip task as 2x u16 into
a single u32. This is all a bit messy still, but once
we have everything converted over to use the brush vertex
format, we can define it correctly as u16 attributes and
avoid all the packing / unpacking.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2415)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -8,7 +8,7 @@ void brush_vs(
     VertexInfo vi,
     int prim_address,
     RectWithSize local_rect,
-    ivec2 user_data,
+    ivec3 user_data,
     PictureTask pic_task
 );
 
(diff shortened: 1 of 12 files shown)


glennw:
r? @kvark or anyone

Pending try here:
https://hg.mozilla.org/try/rev/98ac813b4cbbcfec0649b6709e70e719cb33401d
https://treeherder.mozilla.org/#/jobs?repo=try&revision=98ac813b4cbbcfec0649b6709e70e719cb33401d

glennw:
Try looks good - the R8 failure is expected due to a previous PR.

kvark:
Messy indeed, but I see nothing wrong with the logic :)
@bors-servo r+

kvark:

@bors-servo retry

PR #2416, opened 2018-02-14
Update YUV example to work again.

glennw:
Previously, the example created textures that were placed in
the texture cache. However, RG8 textures are no longer supported
in the texture cache.

Instead, the example now behaves more like a real video decoding
application. The YUV planes are created as native textures and
passed to WR via the external image callback.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2416)

### DIFF
--- a/webrender/examples/yuv.rs
+++ b/webrender/examples/yuv.rs
@@ -10,158 +10,67 @@ extern crate webrender;
 mod boilerplate;
 
 use boilerplate::Example;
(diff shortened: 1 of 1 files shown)


glennw:
r? anyone

I also removed the duplicated touch state code from other examples, since it's not relevant here. I'll be working on porting the YUV shader to a brush shader this week, so it makes sense to get this working again now.

PR #2422, opened 2018-02-14
Port ps_composite shader -> brush_mix_blend.

glennw:
This uses the same technique as the blend shader - reduces the
number of pixels touched for transformed elements, and enables
all the usual brush features (edge AA, clips, segments) for
mix-blend-mode operations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2422)

### DIFF
--- a/webrender/res/brush_mix_blend.glsl
+++ b/webrender/res/brush_mix_blend.glsl
@@ -2,40 +2,41 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#include shared,prim_shared
+#define VECS_PER_SPECIFIC_BRUSH 5
 
-varying vec3 vUv0;
-varying vec3 vUv1;
+#include shared,prim_shared,brush
+
+varying vec3 vUv;
+varying float vW;
+
+varying vec3 vSrcUv;
+varying vec3 vBackdropUv;
 flat varying int vOp;
 
 #ifdef WR_VERTEX_SHADER
-void main(void) {
-    CompositeInstance ci = fetch_composite_instance();
-    PictureTask dest_task = fetch_picture_task(ci.render_task_index);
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

Try looks green (just the R8 expected failure from the mip-map patch):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d9c5c73afceedcac2be04dfb1d1a131ffe31d189

webrender/src/batch.rs
```diff
@@ -75,17 +75,17 @@ pub enum BrushBatchKind {
...
+        source_id: RenderTaskId,
+        backdrop_id: RenderTaskId,
+    },
 }
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
 pub enum BatchKind {
-    Composite {
```
kvark: 🎉

webrender/res/brush_mix_blend.glsl
```diff
@@ -203,17 +204,17 @@ const int MixBlendMode_Saturation  = 13;
 const int MixBlendMode_Color       = 14;
 const int MixBlendMode_Luminosity  = 15;
 
-void main(void) {
-    vec4 Cb = texture(sCacheRGBA8, vUv0);
-    vec4 Cs = texture(sCacheRGBA8, vUv1);
+vec4 brush_fs() {
+    vec2 uv = vUv.xy / vW;
```
kvark: this can be improved, as discussed on IRC. I'll follow-up with fixes

glennw: Yep, sounds good - so the basic idea will be to set W=1 in the brush shader code when we want to emulate ```noperspective```, since that isn't available in ES3.0. This can be applied to both ```brush_blend``` and ```brush_mix_blend```.

Issue #2421, opened 2018-02-14
Border radius clip problem with perspective transforms


pcwalton:
Failing test case:

https://gist.github.com/pcwalton/07bd4ae28e953553c20fda0dfa257334

Screen shot:

https://i.imgur.com/PFOPzWx.png

pcwalton:
Still has problems—corners become square with heavy perspective distortion—but less severe now.

<img width="405" alt="screen shot 2018-09-17 at 1 35 09 pm" src="https://user-images.githubusercontent.com/157897/45648892-a31c8100-ba7e-11e8-98bc-b79e26e8a643.png">

PR #2423, opened 2018-02-14
Use one iteration of Newton's method to approximate distance to an ellipse rather than computing the exact distance.

pcwalton:
This is a well-known trick. It's described in Loop-Blinn and is
thoroughly detailed in G. Taubin, "Distance Approximations for
Rasterizing Implicit Curves".

I've verified that this looks good for circular borders, elliptical borders, and both under rotation. Perspective is broken, but in the same way it's already broken in master (see #2421).

Let me know if there are any reftests/etc. you'd like me to add.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2423)


glennw:
We have a fair number of reftests with ellipse borders, so if this passes those, it should be fine.

We'll need a gecko try though as well.

glennw:
@pcwalton OK, the reftests are failing - but it might just be subpixel differences from the reference image. If you run ```./script/headless.py reftest >ref.log``` you'll be able to view the differences in the reftest analyzer (which is in-tree). If you need to re-generate reference images, you can use ```./script/headless.py png <yaml file>```.

pcwalton:
@glennw OK, I fixed failures relating to zero-width borders and updated the various reftests.

pcwalton:
Cropped the reference images. r? @glennw

glennw:
Rebased and pushed updated Linux-specific reftest images. Let's see if CI is happy now.

glennw:
Pending gecko try:

https://hg.mozilla.org/try/rev/c52c5828ac2504c928093c29557217be72497915
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c52c5828ac2504c928093c29557217be72497915

glennw:
There are some failures in the gecko try. The R8 and R4 failures are known (https://github.com/servo/webrender/pull/2405). From a quick look at the R3 failures, they seem to be subpixel differences between the SVG rendered blob and the WR rendered version. I think they should all be fine to fuzz - but I'll get @staktrace to confirm before we merge.

glennw:
r=me if @staktrace is happy with the try, btw.

staktrace:
Sorry for the delay. The R3 failures look fine - those tests are all fuzzy already and the increase in fuzz is just by one in most cases, so I'm fine with that. The R4 one was not fuzzy before as far as I know (I don't see anything about it in #2405?) but the changes are minor and I'm ok with fuzzing it now. R8 failure is known, as @glennw said. So I'm good with this, thanks!

staktrace:
> The R4 one was not fuzzy before as far as I know 

Ignore this bit, #2405 introduced nondeterministic fuzziness, so that confused me for a bit.

PR #2425, opened 2018-02-15
No perspective interpolation for blend brushes

kvark:
A follow-up to #2422

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2425)


glennw:
Paraphrasing IRC conversation - for later reference:

Once we support rasterizing pictures in screen or local space, what we'll want to do is:
 * Remove the ```FORCE_NO_PERSPECTIVE``` define.
 * Have ```brush.glsl``` select whether to force ```w=1``` based on the raster mode of the picture in question.

glennw:
Yup, very slight difference in one of the reftests which will need an updated reference image.

kvark:
Reftest is updated. Also, pending try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=1f1d9f2620554952095d20d77edd1b3a1bb78301

kvark:
@staktrace 
> builds/worker/workspace/build/tests/reftest/tests/dom/html/reftests/bug917595-pixel-rotated.jpg

Looks like an intermittent bug that is also affected by the way we interpolate things.
Is this the case where Gecko updates a reftest, or the PR needs to be fixed (I don't particularly see anything wrong with it, for the record)?

staktrace:
@kvark you can ignore that one, it was introduced by #2408 and I already have a fuzzing patch in my queue for it.

kvark:
Thanks!
@bors-servo r=glennw

PR #2426, opened 2018-02-15
Remove discard from radial gradient, apply tile/repeat on CPU.

glennw:
Introduce a new method to decompose a PrimitiveInfo struct info
a (bounded) maximum number of instances based on a tile size and
repeat structure.

This allows us to remove the discard instruction from the radial
gradient shader, which means it doesn't disable early-Z.

This is a start point - in the future we'll want to expand this
to work on other gradients and images, and do a proper job
of handling edge cases with huge numbers of repetitions.

For now, this improves the common case of radial gradients, and
makes it easier to convert it to a brush shader.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2426)

### DIFF
--- a/webrender_api/src/display_item.rs
+++ b/webrender_api/src/display_item.rs
@@ -649,6 +649,22 @@ impl LocalClip {
             ),
         }
     }
+
+    pub fn clip_by(&self, rect: &LayoutRect) -> LocalClip {
+        match *self {
+            LocalClip::Rect(clip_rect) => {
+                LocalClip::Rect(
+                    clip_rect.intersection(rect).unwrap_or(LayoutRect::zero())
+                )
+            }
+            LocalClip::RoundedRect(clip_rect, complex) => {
+                LocalClip::RoundedRect(
+                    clip_rect.intersection(rect).unwrap_or(LayoutRect::zero()),
+                    complex,
+                )
+            }
+        }
+    }
 }
 
 #[repr(C)]
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -75,6 +75,7 @@ vec4[2] fetch_from_resource_cache_2(int address) {
 #define VECS_PER_PRIM_HEADER        2
 #define VECS_PER_TEXT_RUN           3
 #define VECS_PER_GRADIENT           3
+#define VECS_PER_RADIAL_GRADIENT    2
 #define VECS_PER_GRADIENT_STOP      2
 
 uniform HIGHP_SAMPLER_FLOAT sampler2D sClipScrollNodes;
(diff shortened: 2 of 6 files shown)


glennw:
r? @kvark or @nical 

Try looks good:
https://hg.mozilla.org/try/rev/52c50d90136b21e12b57a66d8b81ca89b4ddad93
https://treeherder.mozilla.org/#/jobs?repo=try&revision=52c50d90136b21e12b57a66d8b81ca89b4ddad93&selectedJob=162346549

One expected fail from a previous PR, and 2 new PASS for radial gradient tests.

webrender_api/src/display_item.rs
```diff
@@ -89,6 +89,58 @@ impl LayerPrimitiveInfo {
...
         }
     }
+
+    pub fn decompose(
+        &self,
+        tile_size: LayerSize,
+        tile_spacing: LayerSize,
+        max_prims: usize,
+    ) -> Vec<LayerPrimitiveInfo> {
+        let mut prims = Vec::new();
```
nical: nit: it'd be nice to not have to allocate a vector at all, but we can at least pre-compute its capacity.

webrender_api/src/display_item.rs
```diff
@@ -89,6 +89,58 @@ impl LayerPrimitiveInfo {
...
+        if tile_repeat.width <= 0.0 ||
+           tile_repeat.height <= 0.0 {
+            return prims;
+        }
+
+        if tile_repeat.width < self.rect.size.width ||
+           tile_repeat.height < self.rect.size.height {
+            let local_clip = self.local_clip.clip_by(&self.rect);
+            let rect_p0 = self.rect.origin;
+            let rect_p1 = self.rect.origin + self.rect.size;
```
nical: nit: `self.rect.bottom_right()`

webrender_api/src/display_item.rs
```diff
@@ -89,6 +89,58 @@ impl LayerPrimitiveInfo {
...
+
+            let mut y0 = rect_p0.y;
+            while y0 < rect_p1.y {
+                let mut x0 = rect_p0.x;
+
+                while x0 < rect_p1.x {
+                    prims.push(PrimitiveInfo {
```
nical: Should we skip emitting primitives that are completely clipped out by `local_clip` or is this already taken care of elsewhere?

webrender_api/src/display_item.rs
```diff
@@ -89,6 +89,58 @@ impl LayerPrimitiveInfo {
...
+                        ),
+                        local_clip,
+                        is_backface_visible: self.is_backface_visible,
+                        tag: self.tag,
+                    });
+
+                    // Mostly a safety against a crazy number of primitives
+                    // being generated. If we exceed that amount, just bail
+                    // out and only draw the maximum amount.
+                    if prims.len() > max_prims {
```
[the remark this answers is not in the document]
glennw: warn! isn't available in the api crate, so I moved the tiling function to a trait / implementation in the WR crate, and added a warning here.

webrender_api/src/display_item.rs
glennw: Vec::new() is guaranteed to not allocate on the heap until something is pushed. So I did it this way since the common case will be no tiling involved. We could pre-calculate a capacity inside the code branch once we determine we *do* need tiling though.

webrender_api/src/display_item.rs
glennw: Fixed

webrender_api/src/display_item.rs
glennw: They are handled elsewhere - we could potentially save a bit of memory allocation by doing it early here though.

glennw:
@nical Thanks for the review, comments addressed.

PR #2431, opened 2018-02-15
Bump rayon to 1.0

Eijebong:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2431)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,6 +40,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayvec"
+version = "0.4.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "nodrop 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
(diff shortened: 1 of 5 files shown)


glennw:
From CI it looks like some of the unit tests need to be updated too. You can check by running ```cargo test``` from the root workspace directory.

Eijebong:
Oh woups, my bad for not running that :p

staktrace:
This breaks the gecko build, if you can provide a patch to fix it that would be much appreciated.

https://treeherder.mozilla.org/logviewer.html#?job_id=162576263&repo=try

glennw:
I'll work on a patch now.

glennw:
@staktrace Attached two patches to https://bugzilla.mozilla.org/show_bug.cgi?id=1437572 - one for the webrender_bindings/ directory, and one for toolkit/ (which I assume your build script generates automatically, but included just in case). Let me know if that covers what you need.

staktrace:
Thanks for the patches!

PR #2432, opened 2018-02-16
Add brush flags and allow selection of perspective interpolation per brush instance.

glennw:
This is a step towards removing ps_hardware_composite shader
completely, which will allow fixing some existing drop-shadow bugs.

By selecting whether the source image needs perspective
interpolation or not, we can collapse ps_hardware_composite into
the brush_image shader.

This will then mean that hardware composites get the same
benefits as brush_blend and brush_mix_blend (reduced pixel
counts for transformed primitives, opt-in clip masks, edge AA).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2432)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -624,7 +624,8 @@ VertexInfo write_transform_vertex(RectWithSize local_segment_rect,
                                   vec4 clip_edge_mask,
                                   float z,
                                   ClipScrollNode scroll_node,
-                                  PictureTask task) {
+                                  PictureTask task,
+                                  bool do_perspective_interpolation) {
     // Calculate a clip rect from local_rect + local clip
     RectWithEndpoint clip_rect = to_rect_with_endpoint(local_clip_rect);
     RectWithEndpoint segment_rect = to_rect_with_endpoint(local_segment_rect);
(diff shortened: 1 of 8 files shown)


glennw:
r? @kvark 

Try looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c499c85fb73b24e69330ea08875e21e15cdf2579

The R8 failure is the mip-map patch, and the R4 failures are from https://github.com/servo/webrender/pull/2405/.

glennw:
@kvark It occurred to me that fixing this will help fixing a couple of bugzilla issues related to drop shadows, so I figured I would do it now.

webrender/res/prim_shared.glsl
```diff
@@ -661,9 +662,10 @@ VertexInfo write_transform_vertex(RectWithSize local_segment_rect,
     vec2 device_pos = world_pos.xy / world_pos.w * uDevicePixelRatio;
     vec2 task_offset = task.common_data.task_rect.p0 - task.content_origin;
 
-#ifdef FORCE_NO_PERSPECTIVE
-    world_pos.w = 1.0;
-#endif
+    // Force w = 1, if we don't want perspective interpolation (for
+    // example, drawing a screen-space quad on an element with a
+    // perspective transform).
+    world_pos.w = mix(1.0, world_pos.w, do_perspective_interpolation);
```
kvark: wouldn't it be more straightforward to just do `if !do_perspective_interpolation {..}`?

kvark:
no blockers discovered, shipit!
@bors-servo r+

webrender/res/prim_shared.glsl
glennw: Force of habit to avoid things that look like ```if``` statements even though I'm sure all shader compilers would probably find the ```if``` easier to convert to an optimal implementation. I'll change that as a follow up since this is already r+'ed.

PR #2424, opened 2018-02-15
Port YUV primitive to be a brush.

glennw:
Also refactor and optimize the YUV shader implementation to do
less per-pixel work.

Support segment building for YUV primitive - this is a large
performance win for any YUV images with a complex clip on them.

This is part of the prep work to allow selection of whether
a stacking context is rasterized in local or screen-space.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2424)

PR #2433, opened 2018-02-16
Port radial gradient shader to be a brush primitive.

glennw:
This is mostly a straightforward conversion, with a couple of
interesting notes:

* The previous radial gradient shader was missing a transform
  implementation so would never have worked properly with
  rotation and/or perspective.
* We didn't (and still don't) detect if a radial gradient is
  opaque - it's always assumed transparent. So even though
  it can now opt in to segments (and save clip mask generation
  time), the inner segments will still be in the alpha pass.
  We should fix this!
* Previously, we used to append the gradient stops directly
  after the primitive information in the main GPU cache handle.
  Now, there is a separate GPU cache handle for the stops. This
  is quite a bit more efficient in the GPU cache, since the
  size of the stops array is large, and a power of 2, it now
  fits exactly into a GPU cache slab, and allows the main
  primitive data to be allocated in a smaller row.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2433)


glennw:
r? @kvark or @nical 

I will kick off a gecko try shortly.

glennw:
Gecko try:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=af47d705000970aa0504145c2bbdc6314b8c1a61&selectedJob=162598036

Despite the number of orange, I think this is actually good. I think the Windows gpu, mda, 2 and C items are unrelated intermittents. The fails in R4 and R8 are from previous PRs (the mipmap and local box-shadow clip ones).

The other two unexpected results in R4 are new radial gradient PASS results.

Of course, it'd be good for @staktrace or someone else to verify the above before merging.

staktrace:
The windows failures in the try push don't look like intermittents to me. They seem to be all crashes of some sort, and I haven't seen that kind of failure before. I can do a try push for this change using my script and retrigger the jobs a few times to confirm one way or another.

staktrace:
Try push is in progress at https://treeherder.mozilla.org/#/jobs?repo=try&revision=d7e716bb5071bcb271c8a14a4c0101bc5a36b15c (I just included the windows jobs)

staktrace:
@glennw Looks like there's shader compilation failures. See output at https://treeherder.mozilla.org/logviewer.html#?job_id=162670092&repo=try&lineNumber=1904

staktrace:
I think the shader compilation failures are actually coming from #2424 which merged already. Although this PR might make it worse since the try push for this has mda job failures as well which I didn't see on the try push including just #2424.

kvark:
:lgtm:

---

---

*[webrender/src/prim_store.rs, line 208 at r1](https://reviewable.io:443/reviews/servo/webrender/2433#-L5VFG2Wb16VVvCKzzRi:-L5VFG2Wb16VVvCKzzRj:b-1i6oel) ([raw file](https://github.com/servo/webrender/blob/33b8ec9cd75f55d99a2ab6607c9103d8fd9ce715/webrender/src/prim_store.rs#L208)):*
> ```Rust
>         start_center: LayerPoint,
>         end_center: LayerPoint,
>         start_radius: f32,
> ```

may consider using `ops::Range` for some of the start/stop pairs

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2433#-:-L5VGQ2dqf-jxfJqs8zU:bnfp4nl)*

glennw:
Try run after rebase:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=00bec7660cc074f579ba22eaa221c7b598fa9350&selectedJob=162923263

We have:
 * A few fails in border radii cases which are going to be fuzzed as a result of the ellipse distance AA change.
 * A few fails in the input tests using box-shadows related to the box-shadow local clip changes.
 * Three new passes in the radial gradient tests.

So, this should be OK to merge.

@bors-servo r=kvark

PR #2437, opened 2018-02-16
Fix brush YUV shader

kvark:
Fixes https://github.com/servo/webrender/pull/2424#issuecomment-366298946
r? @glennw , cc @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2437)


kvark:
As per IRC discussion:
@bors-servo r=staktrace

PR #2441, opened 2018-02-19
Remove aligned gradient shader, port to brush.

glennw:
Instead, push aligned gradients through the linear angle gradient
shader (for now).

This will simplify converting linear gradient code over to be
a brush shader. The optimization here was always a bit dubious
(it made more sense when we first started WR, when the angle
gradient shader was more expensive than it is now).

Instead, what we can do in the future if gradients ever show
up in the GPU profile is support vertex colors for the rectangle
shader and decompose simple aligned gradients into a series
of rectangle segments with different vertex colors.

The benefit of this is reduced shader changes and simpler batching.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2441)


glennw:
r? @kvark or anyone really.

Pending try here:
https://hg.mozilla.org/try/rev/490eda306772c66d84b1c3531264977f829f708d
https://treeherder.mozilla.org/#/jobs?repo=try&revision=490eda306772c66d84b1c3531264977f829f708d

glennw:
Added a follow up commit that switches the tiling / repeat of linear gradients to be done on CPU rather than in the shader (exactly the same implementation details as how this was done for radial gradients previously).

glennw:
Added another commit that ports the linear gradient primitive to be a brush shader.

glennw:
The overall summary of this PR:

* Remove aligned gradient shader and primitive (there's better ways to do fast path gradients, if we need to, in the future).
* Rename angle gradient -> linear gradient, which is more accurate.
* Switch linear gradient tile/repeat to be done on CPU, exactly as radial gradients now work.
* Port angle gradient shader to brush primitive, allowing segments for linear gradients.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4a4739d322e1588e02f396e7b3f6db23a00d8063
https://hg.mozilla.org/try/rev/4a4739d322e1588e02f396e7b3f6db23a00d8063

kvark:
Looks sane to me!
One thing I'd like to be changed (possibly in a follow-up), and also there appear to be failures in the last try push, related to border shaders and input-radio-XXX.

---

Reviewed 2 of 7 files at r1, 1 of 5 files at r2, 8 of 8 files at r3.

---

*[webrender/tests/angle_shader_validation.rs, line 104 at r3](https://reviewable.io:443/reviews/servo/webrender/2441#-L5nYIh3InP6u_ztfF57:-L5nYIh3InP6u_ztfF58:bveog7r) ([raw file](https://github.com/servo/webrender/blob/1da09d227bcc284cce23dba3935582fceb688ef7/webrender/tests/angle_shader_validation.rs#L104)):*
> ```Rust
>     Shader {
>         name: "brush_radial_gradient",
>         features: &[],
> ```

let's have at least one of them with dithering feature

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2441#-:-L5nYYJ9G48fT5OIdNhn:b-j1zwjp)*

glennw:
@kvark Thanks for the review! I updated the validation to include the dither feature. The reftest failures are from previous PRs (the input/radio are from the local-clip-rect + box-shadow PR, and the border-radii are from the ellipse distance optimization PR). Both are fuzzed in the latest WR update.

glennw:
Rebased.

@bors-servo r=kvark

glennw:
@staktrace Ugh, there are some gradient reftest failures I missed seeing in the try push above - these are going to cause failures in the wr-future bug. I'll look into these as soon as possible (they weren't there initially - possibly a rebasing mistake I made or something like that).

glennw:
Interestingly, they only appear in a debug build - which might be how I managed to miss them. I also can't reproduce locally with a release build. Compiling a debug build now.

glennw:
Can't reproduce locally with rustc debug enabled. Trying another build with full debug enabled.

staktrace:
I'll do a local debug build as well to see if I can repro.

staktrace:
I've been able to reproduce a failure in layout/reftests/css-gradients/height-dependence-2.html but very sporadically. I tried getting WR captures of it but they came out blank, so I don't know how useful that will be. The failure as reported by the reftest harness shows the "pad pad pad" text not running all the way to the bottom of the rect in the test image (there's only 16 lines of it vs. 32 in the reference image). I did, however, get a rr recording of the failure (since I reproduced it with rr's chaos mode). If you have any suggestions on how to debug it I'm all ears.

staktrace:
[height-dependence-2-captures.zip](https://github.com/servo/webrender/files/1745590/height-dependence-2-captures.zip)

^ the captures along with the reftest harness output (for use in reftest analyzer) if they're of any use to anybody.

kvark:
Are you sure the breakage is caused by this PR though? Gradients don't appear to be rendering any differently between the captures. And the PR doesn't seem to touch anything on either the clip chain side or glyph positioning.

glennw:
We're not sure it's this PR - we are bisecting a couple of patches to find the cause.

staktrace:
Pretty sure it's from this PR, I retriggered the R4 jobs a few times to make sure:

at 3797d9c9c28f93738bb037cddadee6c1e72e85e0 - R4 is green
https://treeherder.mozilla.org/#/jobs?repo=try&revision=ddc8f978d3bfb1b75a40631486130ac10ba0ef38

at b2acbc94943810b8a93e5bb9a3f4c0ccc446379e - R4 is green (but retriggers are still going)
https://treeherder.mozilla.org/#/jobs?repo=try&revision=e34663d5ac99e273b2c5fd56a2cd0de3ebe5b836

at 2d2d7b7f293eaac2b463fe0e61a7face9253053e - R4 is orange
https://treeherder.mozilla.org/#/jobs?repo=try&revision=3c1d576ae6c701b62efc56619583b175c3898c84

glennw:
Yup, that seems fairly conclusive. There's 3 commits in that patch, so I'll try to narrow down the cause there.

glennw:
Haven't managed to reproduce locally yet. Here are pending try pushes with each of the 3 commits in this PR reverted though, which should help narrow it down:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=4f9443cde2d85641c2510ecf02b2be704cffda36

https://treeherder.mozilla.org/#/jobs?repo=try&revision=c62948592adb70ed9534648cdeeeac0611b117ef

https://treeherder.mozilla.org/#/jobs?repo=try&revision=1cb15ddf82a8ce99c59f0d21688d73b7be82bd48

https://treeherder.mozilla.org/#/jobs?repo=try&revision=928c958792ddab98bc73ba934cdec41df307a91c

Although, TC seems to be having network issues right now :sob:

glennw:
The ```height-dependence-2``` test has a gradient that is 1px wide and gets repeated across the screen, resulting in 1280 primitives being generated. This results in a large number of gradients being added to the GPU cache. My current theory is that this is exposing a bug in the GPU cache, perhaps related to resizing the GPU cache texture.

glennw:
Oh, I see what the problem is. Working on a fix now.

kvark:
Speaking of GPU cache issues, we are lacking the appropriate toolset - #2453
WR capturing helps, but text format is not the best for inspecting GPU cache data.

glennw:
Potential fix here - https://github.com/servo/webrender/pull/2454 - try run in progress to see if it helps.

kvark:
I confirm this is a problem with exceeding maximum texture size of 8K.
Also see #2459 to have run-time checks for it.

PR #2442, opened 2018-02-20
Increase the area threshold for segmenting to 256 x 256 pixels.

glennw:
This seems like a more reasonable tradeoff. Based on a very simple
and unscientific study of a dozen or so websites, having the area
threshold at 128x128 was resulting in a lot of render tasks being
created where it doesn't provide any benefit (e.g. the home page
was ending up with ~600 render tasks of sizes ~10x10 pixels).

We can probably come up with some better heuristics in the future
but for now this is a better default and a simple change.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2442)


glennw:
r? anyone

kvark:
We need to make tuning knobs like this exposed, perhaps via `RendererOptions` or something else.

Issue #2388, opened 2018-02-06
ClipChain with two clips descending from mask has mask applied twice


staktrace:
I was reworking the gecko clipping glue to use the ClipChain APIs and ran into a problem. (I had asked @mrobinson about it on IRC, but upon further investigation it turns out the scenario is a little different from what I described to him).

Not entirely sure if this behaviour is expected or not, but the scenario is as follows:
- there is a clip with a mask image that has a 0.5 opacity (id 6)
- there are two clips that are descendant clips of that one (id 7 and id 10)
- there is a ClipChain containing those two clips (id 11)
- there is a rect that uses the ClipChain

In this scenario the mask image gets applied to the rect twice so the rect ends up with opacity 0.25. Whereas I'm expecting it to have an opacity of 0.5 (because the mask would be "deduplicated" from the set of clips that apply to the rect, and only get applied once).

Here's the yaml and mask image:
[yaml_frames.zip](https://github.com/servo/webrender/files/1700823/yaml_frames.zip)

The zip contains both my reduced version (frame-4.yaml) and the original unreduced version which was generated from a recording of `layout/reftests/w3c-css/submitted/masking/mask-opacity-1a.html` in mozilla-central.

**Note**: One thing that might be relevant is that in the original unreduced version, clips "7" and "10" were not in the same ClipChain, but in two chains linked by a parent reference. That is, chain id=11 just contained clip 10 but pointed to a parent chain id=9 which contained clip 7.

mrobinson:
Odd! In this case, I would expect that the mask does not get applied at all. ClipChains should only apply the clips specified by the clips themselves and not any of their parents.

mrobinson:
I will try to take a look at this issue next.

PR #2443, opened 2018-02-20
Allow clip masks for picture tasks.

glennw:
Now that we draw pictures by using their local rect as a normal
primitive, we also need to apply clip masks to them, when they
are required.

For example, a picture may have a transform on it, and when it
gets drawn it needs to sample from a clip mask in this case since
the coordinate systems are not equal and the local clip rect
can not be applied in this case.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1439059.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2443)


glennw:
There is a try run from this change here:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=d5168fd21fbdd1ab172bf44dbd9d0f3d97ba6b9f

This does cause two new failures. However, I *think* that it's probably worth getting this merged, marking those as failures and following them up as individual bugs. The reasoning for this is (a) I believe the failures are unrelated bugs exposed by this (b) the bug this fixes is a major artifact on a lot of real world pages, and it'd be good to get it fixed quickly. Discussion of the failures below:

R6 failure [1] - in this case the screen bounding rect becomes zero, and the primitive gets clipped out early. This is an existing bug with calculating the screen bounding rect, somehow caused by the level of nesting of preserve-3d stacking contexts. Before this patch, it gets drawn anyway, since we don't run the clip code that culls out the zero sized bounding rect.

R1 failure [2] - I haven't confirmed what is happening in this one, yet. Looking visually, it appears that the image mask is being applied to both the primitive and the resulting stacking context picture. If that *is* the case, the fix here would be to update Gecko once this lands to supply the image-mask clip chain only to the primitive and/or the stacking context picture, rather than both. As a side-effect of this proposed change, https://github.com/servo/webrender/issues/1957 gets fixed. It may also be related to https://github.com/servo/webrender/issues/2388.

[1] https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/aaADyPUUTUam-EV8ktkkyw/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1

[2] https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/Vp6J628QRYacCrUl5yPqpA/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1

@staktrace I'm going to keep looking at those two failures tomorrow. Assuming what I've written above is correct, does it seem reasonable to get this merged and mark those two as FAILs for now?

r? @kvark (let's hold off on merging until the above is resolved though).

glennw:
cc @mrobinson as well for comments about whether this makes sense, or if it has any effect on things you're working on.

glennw:
(This is slightly conservative by the way - we'll end up drawing some potentially redundant clip masks temporarily. But I did a significant amount of manual testing in Gecko with this change and didn't find any noticeable performance regressions).

mrobinson:
@glennw This is pretty unrelated on what I'm working on right now and the change looks sensible to me.

staktrace:
The R1 failure does look exactly like what I was seeing when I tried updating Gecko to use the ClipChain API, and I filed #2388 for exactly that. I don't know if we can fix this on the Gecko side without updating to the ClipChain API, and we need #2388 fixed for that. @mrobinson do you have an ETA for tackling that issue?

But yeah I guess we can merge this PR but I'd like to try hard to get the followups resolved as soon as we can.

glennw:
@staktrace Thanks. I'll spend some time today continuing to look into those and see I can at least confirm the exact issue, if not resolve them, before we merge this.

glennw:
Rebased and testing with a slight modification to the logic. Pending try:

https://hg.mozilla.org/try/rev/2e0875e2010714797c60557eb8c6428f48390389
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2e0875e2010714797c60557eb8c6428f48390389

glennw:
@staktrace The latest patch fixes the 3d transform failure that was previously in R6. The only remaining failure is the double image-mask clip in R1. Are you happy to land it with that and mark that test as failing? It fixes a number of other glitch bugs we have in current Gecko nightly.

kvark:
sweet!
@bors-servo r+

PR #2444, opened 2018-02-20
Use FastTransform almost everywhere

mrobinson:
This helps to avoid the cost of expensive matrix math and allows us to
more liberally call things like `inverse()` on our transformations,
since the worst case is that we will be retrieving the already-cached
inverted matrix. This decreases the amount of time that
FrameBuilder::build takes in my local CPU profiles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2444)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -811,7 +811,7 @@ impl AlphaBatchBuilder {
                 let font_transform = if is_shadow {
                     None
                 } else {
-                    Some(&scroll_node.transform)
+                    Some(scroll_node.transform)
                 };
 
                 let font = text_cpu.get_font(
@@ -977,8 +977,10 @@ impl AlphaBatchBuilder {
                                 if is_in_3d_context {
                                     // Push into parent plane splitter.
 
-                                    let real_xf = &ctx.clip_scroll_tree.nodes[&reference_frame_id].world_content_transform;
-
+                                    let real_xf = &ctx.clip_scroll_tree
+                                        .nodes[&reference_frame_id]
+                                        .world_content_transform
+                                        .into();
                                     let polygon = make_polygon(
                                         real_local_rect,
                                         &real_xf,
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -3,8 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{BorderRadius, ClipMode, ComplexClipRegion, DeviceIntRect, DevicePixelScale, ImageMask};
-use api::{ImageRendering, LayerRect, LayerToWorldTransform, LayoutPoint, LayoutVector2D};
-use api::LocalClip;
+use api::{ImageRendering, LayerRect, LayoutPoint, LayoutVector2D, LocalClip};
 use border::{BorderCornerClipSource, ensure_no_corner_overlap};
 use clip_scroll_tree::{ClipChainIndex, CoordinateSystemId};
 use ellipse::Ellipse;
@@ -13,7 +12,8 @@ use gpu_cache::{GpuCache, GpuCacheHandle, ToGpuBlocks};
 use gpu_types::ClipScrollNodeIndex;
 use prim_store::{ClipData, ImageMaskData};
 use resource_cache::{ImageRequest, ResourceCache};
-use util::{MaxRect, MatrixHelpers, calculate_screen_bounding_rect, extract_inner_rect_safe};
+use util::{LayerToWorldFastTransform, MaxRect, calculate_screen_bounding_rect};
+use util::extract_inner_rect_safe;
 use std::rc::Rc;
 
 pub type ClipStore = FreeList<ClipSources>;
@@ -252,7 +252,7 @@ impl ClipSources {
 
     pub fn get_screen_bounds(
         &self,
-        transform: &LayerToWorldTransform,
+        transform: &LayerToWorldFastTransform,
         device_pixel_scale: DevicePixelScale,
     ) -> (DeviceIntRect, Option<DeviceIntRect>) {
         // If this translation isn't axis aligned or has a perspective component, don't try to
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -3,9 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{ClipId, DevicePixelScale, ExternalScrollId, LayerPixel, LayerPoint, LayerRect};
-use api::{LayerSize, LayerToWorldTransform, LayerTransform, LayerVector2D, LayoutTransform};
-use api::{LayoutVector2D, PipelineId, PropertyBinding, ScrollClamping, ScrollEventPhase};
-use api::{ScrollLocation, ScrollSensitivity, StickyOffsetBounds, WorldPoint};
+use api::{LayerSize, LayerVector2D, LayoutTransform, LayoutVector2D, PipelineId, PropertyBinding};
+use api::{ScrollClamping, ScrollEventPhase, ScrollLocation, ScrollSensitivity, StickyOffsetBounds};
+use api::WorldPoint;
 use clip::{ClipChain, ClipSourcesHandle, ClipStore, ClipWorkItem};
 use clip_scroll_tree::{ClipChainIndex, CoordinateSystemId, TransformUpdateState};
 use euclid::SideOffsets2D;
@@ -15,7 +15,8 @@ use gpu_types::{ClipScrollNodeIndex, ClipScrollNodeData};
 use resource_cache::ResourceCache;
 use scene::SceneProperties;
 use spring::{DAMPING, STIFFNESS, Spring};
-use util::{MatrixHelpers, TransformOrOffset, TransformedRectKind};
+use util::{LayerToWorldFastTransform, LayerFastTransform, LayoutFastTransform};
+use util::{TransformedRectKind};
 
 #[cfg(target_os = "macos")]
 const CAN_OVERSCROLL: bool = true;
@@ -91,10 +92,10 @@ pub struct ClipScrollNode {
     /// between our reference frame and this node. For reference frames, we also include
     /// whatever local transformation this reference frame provides. This can be combined
     /// with the local_viewport_rect to get its position in world space.
-    pub world_viewport_transform: LayerToWorldTransform,
+    pub world_viewport_transform: LayerToWorldFastTransform,
 
     /// World transform for content transformed by this node.
-    pub world_content_transform: LayerToWorldTransform,
+    pub world_content_transform: LayerToWorldFastTransform,
 
     /// Pipeline that this layer belongs to
     pub pipeline_id: PipelineId,
@@ -119,7 +120,7 @@ pub struct ClipScrollNode {
     /// The transformation from the coordinate system which established our compatible coordinate
     /// system (same coordinate system id) and us. This can change via scroll offsets and via new
     /// reference frame transforms.
-    pub coordinate_system_relative_transform: TransformOrOffset,
+    pub coordinate_system_relative_transform: LayerFastTransform,
 
     /// A linear ID / index of this clip-scroll node. Used as a reference to
     /// pass to shaders, to allow them to fetch a given clip-scroll node.
@@ -135,15 +136,15 @@ impl ClipScrollNode {
     ) -> Self {
         ClipScrollNode {
             local_viewport_rect: *rect,
-            world_viewport_transform: LayerToWorldTransform::identity(),
-            world_content_transform: LayerToWorldTransform::identity(),
+            world_viewport_transform: LayerToWorldFastTransform::identity(),
+            world_content_transform: LayerToWorldFastTransform::identity(),
             parent: parent_id,
             children: Vec::new(),
             pipeline_id,
             node_type: node_type,
             invertible: true,
             coordinate_system_id: CoordinateSystemId(0),
-            coordinate_system_relative_transform: TransformOrOffset::zero(),
+            coordinate_system_relative_transform: LayerFastTransform::identity(),
             node_data_index: ClipScrollNodeIndex(0),
         }
     }
@@ -177,10 +178,12 @@ impl ClipScrollNode {
         pipeline_id: PipelineId,
     ) -> Self {
         let identity = LayoutTransform::identity();
+        let source_perspective = source_perspective.map_or_else(
+            LayoutFastTransform::identity, |perspective| perspective.into());
         let info = ReferenceFrameInfo {
-            resolved_transform: LayerTransform::identity(),
+            resolved_transform: LayerFastTransform::identity(),
             source_transform: source_transform.unwrap_or(PropertyBinding::Value(identity)),
-            source_perspective: source_perspective.unwrap_or(identity),
+            source_perspective: source_perspective,
             origin_in_parent_reference_frame,
             invertible: true,
         };
@@ -258,8 +261,8 @@ impl ClipScrollNode {
 
     pub fn mark_uninvertible(&mut self) {
         self.invertible = false;
-        self.world_content_transform = LayerToWorldTransform::identity();
-        self.world_viewport_transform = LayerToWorldTransform::identity();
+        self.world_content_transform = LayerToWorldFastTransform::identity();
+        self.world_viewport_transform = LayerToWorldFastTransform::identity();
     }
 
     pub fn push_gpu_node_data(&mut self, node_data: &mut Vec<ClipScrollNodeData>) {
@@ -274,7 +277,7 @@ impl ClipScrollNode {
             TransformedRectKind::Complex
         };
         let data = ClipScrollNodeData {
-            transform: self.world_content_transform,
+            transform: self.world_content_transform.into(),
             transform_kind: transform_kind as u32 as f32,
             padding: [0.0; 3],
         };
@@ -303,9 +306,9 @@ impl ClipScrollNode {
 
         self.update_transform(state, next_coordinate_system_id, scene_properties);
 
-        // If this node is a reference frame, we check if the determinant is 0, which means it
-        // has a non-invertible matrix. For non-reference-frames we assume that they will
-        // produce only additional translations which should be invertible.
+        // If this node is a reference frame, we check if it has a non-invertible matrix.
+        // For non-reference-frames we assume that they will produce only additional
+        // translations which should be invertible.
         match self.node_type {
             NodeType::ReferenceFrame(info) if !info.invertible => {
                 self.mark_uninvertible();
@@ -362,7 +365,7 @@ impl ClipScrollNode {
 
         let mut clip_chain = clip_chains[state.parent_clip_chain_index.0].new_with_added_node(
             work_item,
-            self.coordinate_system_relative_transform.apply(&local_outer_rect),
+            self.coordinate_system_relative_transform.transform_rect(&local_outer_rect),
             screen_outer_rect,
             screen_inner_rect,
         );
@@ -398,7 +401,7 @@ impl ClipScrollNode {
         // provided by our own sticky positioning.
         let accumulated_offset = state.parent_accumulated_scroll_offset + sticky_offset;
         self.world_viewport_transform = if accumulated_offset != LayerVector2D::zero() {
-            state.parent_reference_frame_transform.pre_translate(accumulated_offset.to_3d())
+            state.parent_reference_frame_transform.pre_translate(&accumulated_offset)
         } else {
             state.parent_reference_frame_transform
         };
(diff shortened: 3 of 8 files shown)


mrobinson:
Gecko try job for this change is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2f8bc89e36afa986e5f3ea8011caeec007b66cc9

The build failures seem to be unrelated infrastructure issues.

kvark:
Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[webrender/src/clip_scroll_node.rs, line 183 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oL38eJyIvd7t6mBsl:-L5oL38gGQAZmeNQJu6-:b1obrjh) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/clip_scroll_node.rs#L183)):*
> ```Rust
>         let source_perspective = source_perspective.map_or_else(
>             LayoutTransformOrOffset::identity,
>             |perspective| LayoutTransformOrOffset::new_transform(perspective)
> ```

nit: could just be `LayoutTransformOrOffset::new_transform`

---

*[webrender/src/prim_store.rs, line 1475 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oMS5ksP-ZbpTJL9FA:-L5oMS5ksP-ZbpTJL9FB:bv6o7tm) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/prim_store.rs#L1475)):*
> ```Rust
>                     local_clip_rect
>                 } else {
>                     let clip_transform = LayerToWorldTransformOrOffset::new_transform(
> ```

nit: it could also just implement `From<Transform>` and `From<Vector>` instead of constructors

---

*[webrender/src/util.rs, line 157 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oMuQPgim-C8SBHvEP:-L5oMuQPgim-C8SBHvEQ:b-mj215q) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L157)):*
> ```Rust
> 
> pub fn calculate_screen_bounding_rect(
>     transform: &TransformOrOffset<LayerPixel, WorldPixel>,
> ```

nit: use the new alias here

---

*[webrender/src/util.rs, line 347 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oN9JgOs8OMT4eDcIO:-L5oN9JgOs8OMT4eDcIP:b3n5mbj) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L347)):*
> ```Rust
> /// when possible when dealing with non-perspective axis-aligned transformations.
> #[derive(Debug, Clone, Copy)]
> pub enum TransformOrOffset<Src, D> {
> ```

we should be consistent in naming of those generic arguments

---

*[webrender/src/util.rs, line 347 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNPaoKtDAuNSlBDR0:-L5oNPaoKtDAuNSlBDR1:b-dfws8p) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L347)):*
> ```Rust
> /// when possible when dealing with non-perspective axis-aligned transformations.
> #[derive(Debug, Clone, Copy)]
> pub enum TransformOrOffset<Src, D> {
> ```

I don't think name like `XOrY` is elegant. It exposes the internal detail. What is important here is that this is a transform, it quacks like one, maybe not exactly a Matrix4, but can be converted into it. So maybe `FastTransform`? More opinions are welcome :)

---

*[webrender/src/util.rs, line 349 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNDZm3icP9aloHwts:-L5oNDZm3icP9aloHwtt:b-hwmrst) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L349)):*
> ```Rust
> pub enum TransformOrOffset<Src, D> {
>     /// A simple offset, which can be used without doing any matrix math.
>     Offset(TypedVector2D<f32, Src>),
> ```

I think this should use the destination space - makes it more convenient for the using code

---

*[webrender/src/util.rs, line 358 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oN-qPaXKs1Seu5f59:-L5oN-qPaXKs1Seu5f5A:b-99fn9k) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L358)):*
> ```Rust
>     },
> 
>     //Uninvertible,
> ```

is this comment needed?

---

*[webrender/src/util.rs, line 371 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNn-mDmIgWsVkJCwu:-L5oNn-mDmIgWsVkJCwv:b-tevdck) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L371)):*
> ```Rust
> 
>     #[inline(always)]
>     pub fn new_transform(transform: TypedTransform3D<f32, Src, D>) -> Self {
> ```

doesn't look to be consistent with `with_offset` constructor

---

*[webrender/src/util.rs, line 380 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNub5aQvY-33WEU4_:-L5oNub5aQvY-33WEU4a:b2awdh2) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L380)):*
> ```Rust
>     }
> 
>     pub fn to_transform(&self) -> TypedTransform3D<f32, Src, D> {
> ```

could probably just implement `Into<>`

---

*[webrender/src/util.rs, line 391 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oO2Qaz1LJtcNdtSyz:-L5oO2Qaz1LJtcNdtSz-:bvofiqr) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L391)):*
> ```Rust
>         match *self {
>             TransformOrOffset::Offset(..) => true,
>             TransformOrOffset::Transform { inverse, .. } => inverse.is_some(),
> ```

`ref inverse` should avoid copies

---

*[webrender/src/util.rs, line 402 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oOIMPP2UjKF3j0X5j:-L5oOIMPP2UjKF3j0X5k:b-944k3j) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L402)):*
> ```Rust
>         match (self, other) {
>             (&TransformOrOffset::Offset(ref offset), &TransformOrOffset::Offset(ref other_offset)) => {
>                 let offset = TypedVector2D::from_untyped(&offset.to_untyped());
> ```

nit: could also do `offset * TypedScale::<_, NewSrc, D>::new(1.0)`

---

*[webrender/src/util.rs, line 407 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oPUDgt-j-9N3aHacD:-L5oPUDiJ8CsIhU2OC55:bssppwp) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L407)):*
> ```Rust
>             _ => {
>                 let new_transform = self.to_transform().pre_mul(&other.to_transform());
>                 TransformOrOffset::new_transform(new_transform)
> ```

is it desirable to re-compute the inverse during each step of transformations like this one?

---

*[webrender/src/util.rs, line 417 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oOXIwzCLReB7Czu3T:-L5oOXIwzCLReB7Czu3U:b-czk3l7) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L417)):*
> ```Rust
>             &TransformOrOffset::Offset(ref offset) =>
>                 return TransformOrOffset::Offset(*offset + *other_offset),
>             &TransformOrOffset::Transform { transform, .. } =>
> ```

similarly, we probably want `ref transform` here and below

---

*[webrender/src/util.rs, line 442 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oOzxaAdJPMGefp8Qb:-L5oOzxaAdJPMGefp8Qc:bln1n4l) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L442)):*
> ```Rust
>     pub fn is_backface_visible(&self) -> bool {
>         match *self {
>             TransformOrOffset::Offset(..) => true,
> ```

I think this should be `false`

---

*[webrender/src/util.rs, line 483 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oPczyOY6eFWnj0EtU:-L5oPczyOY6eFWnj0EtV:b729x15) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L483)):*
> ```Rust
>                 Some(inverse.transform_rect(&rect)),
>             TransformOrOffset::Transform { transform, is_2d: false, .. } =>
>                 Some(transform.inverse_rect_footprint(rect)),
> ```

not required for this PR, but I think `inverse_rect_footprint` should return an `Option` itself

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2444)*

mrobinson:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[webrender/src/clip_scroll_node.rs, line 183 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oL38eJyIvd7t6mBsl:-L5ock-m2cnsUNV0wiMR:bol2uh5) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/clip_scroll_node.rs#L183)):*

nit: could just be `LayoutTransformOrOffset::new_transform`
</blockquote></details>

I think  LayoutTransformOrOffset::identity is a little more efficient here.

---

*[webrender/src/prim_store.rs, line 1475 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oMS5ksP-ZbpTJL9FA:-L5ocwyHeg1W25TurkFZ:b-ovwawk) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/prim_store.rs#L1475)):*

nit: it could also just implement `From<Transform>` and `From<Vector>` instead of constructors
</blockquote></details>

Good idea!

---

*[webrender/src/util.rs, line 157 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oMuQPgim-C8SBHvEP:-L5ocz6zvSdYhg-ablCy:b-73m6vj) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L157)):*

nit: use the new alias here
</blockquote></details>

Nice catch.

---

*[webrender/src/util.rs, line 347 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oN9JgOs8OMT4eDcIO:-L5od-IO7abSrTM_TOUY:bf9zscq) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L347)):*

we should be consistent in naming of those generic arguments
</blockquote></details>

Also here. I had planned to fix this, but it got lost during the day.

---

*[webrender/src/util.rs, line 347 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNPaoKtDAuNSlBDR0:-L5od3Hh5B9p8ONdb_YK:bqz77n8) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L347)):*

I don't think name like `XOrY` is elegant. It exposes the internal detail. What is important here is that this is a transform, it quacks like one, maybe not exactly a Matrix4, but can be converted into it. So maybe `FastTransform`? More opinions are welcome :)
</blockquote></details>

I'm happy to rename this to FastTransform.

---

*[webrender/src/util.rs, line 349 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNDZm3icP9aloHwts:-L5od6NrUr43wJKs-VYJ:buiathf) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L349)):*

I think this should use the destination space - makes it more convenient for the using code
</blockquote></details>

When it's an offset the `Src` and `Dst` are the same unit. I'm not sure how to express this in the type system. I tested changing this to `Dst` and it seems that it complicates a lot of callsites.  Choosing one or the other will have convenience tradeoffs. Using Src keeps almost all of the `to_untyped``/ `from_untyped` stuff inside the struct implementation, which I think is a good indication that `Src` is the correct choice here. 

---

*[webrender/src/util.rs, line 358 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oN-qPaXKs1Seu5f59:-L5odBRxcJNkHRk1YHRC:b-7kyd0i) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L358)):*

is this comment needed?
</blockquote></details>

This was left over. I'll remove it.

---

*[webrender/src/util.rs, line 371 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNn-mDmIgWsVkJCwu:-L5odDWY2qcuad8_ct4-:bt2vr1b) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L371)):*

doesn't look to be consistent with `with_offset` constructor
</blockquote></details>

Fixed.

---

*[webrender/src/util.rs, line 380 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNub5aQvY-33WEU4_:-L5odEkmm9ehxryIxZII:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L380)):*

could probably just implement `Into<>`
</blockquote></details>

Done.

---

*[webrender/src/util.rs, line 391 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oO2Qaz1LJtcNdtSyz:-L5odFOiYxnKlZwmSxbY:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L391)):*

`ref inverse` should avoid copies
</blockquote></details>

Done.

---

*[webrender/src/util.rs, line 402 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oOIMPP2UjKF3j0X5j:-L5odGdeKth-BPJjQWvx:b-7wu5d9) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L402)):*

nit: could also do `offset * TypedScale::<_, NewSrc, D>::new(1.0)`
</blockquote></details>

It's hard to say which version is more readable, but I think it's a bit easier to read without the angle brackets.

---

*[webrender/src/util.rs, line 407 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oPUDgt-j-9N3aHacD:-L5odQ2NeBW3hwOv_k2e:bybf6wm) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L407)):*

is it desirable to re-compute the inverse during each step of transformations like this one?
</blockquote></details>

It's not necessary and we could specialize this even further, but I wanted to avoid premature optimization here. 

---

*[webrender/src/util.rs, line 417 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oOXIwzCLReB7Czu3T:-L5odWkM5lWRnLbvKiG5:b-896fix) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L417)):*

similarly, we probably want `ref transform` here and below
</blockquote></details>

Done.

---

*[webrender/src/util.rs, line 442 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oOzxaAdJPMGefp8Qb:-L5odXesFCLf7IElIwPr:b-a6taqh) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L442)):*

I think this should be `false`
</blockquote></details>

Thanks! Fixed.

---

*[webrender/src/util.rs, line 483 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oPczyOY6eFWnj0EtU:-L5odYb4vEv08x4ChJLl:bfcnpp3) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L483)):*

not required for this PR, but I think `inverse_rect_footprint` should return an `Option` itself
</blockquote></details>

Okay. That's a good change for the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2444)*

mrobinson:
@kvark Thanks for the excellent review. I think I have answered almost all of your concerns, but let me know if I missed anything.

kvark:
Thanks for addressing my concerns! We are almost there :)

---

Reviewed 8 of 8 files at r2.

---

*[webrender/src/clip_scroll_node.rs, line 183 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oL38eJyIvd7t6mBsl:-L5omOCW64cPX1SWJah6:b-tllysw) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/clip_scroll_node.rs#L183)):*

I think  LayoutTransformOrOffset::identity is a little more efficient here.
</blockquote></details>

this comment was about the line you changed to `|perspective| perspective.into()`, which btw could also be written as `From::from`

---

*[webrender/src/util.rs, line 371 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNn-mDmIgWsVkJCwu:-L5onFodWjOrWdI5Q-Dk:b5q4lp8) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L371)):*

Fixed.
</blockquote></details>

do we still need the constructors though?..

---

*[webrender/src/util.rs, line 402 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oOIMPP2UjKF3j0X5j:-L5onMahJT0OiDk1WsfL:bvdn52l) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L402)):*

It's hard to say which version is more readable, but I think it's a bit easier to read without the angle brackets.
</blockquote></details>

Sure, I don't hold strong feelings about this. Just seeing `untyped` makes me cringe :)

---

*[webrender/src/util.rs, line 115 at r2](https://reviewable.io:443/reviews/servo/webrender/2444#-L5omuMOm7er9r4eQYYp:-L5omuMOm7er9r4eQYYq:b-53y8h2) ([raw file](https://github.com/servo/webrender/blob/8fe094ebc72aa2cde5e1d500d3b2b8e94dba61a3/webrender/src/util.rs#L115)):*
> ```Rust
>         }
> 
>         self.m43 < NEARLY_ZERO
> ```

I think this should come with `abs()`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2444#-:-L5onj68soe5whVuqRPR:b-6jn6o4)*

mrobinson:

---

*[webrender/src/util.rs, line 371 at r1](https://reviewable.io:443/reviews/servo/webrender/2444#-L5oNn-mDmIgWsVkJCwu:-L5p-itwuPyH2dASLlWr:bbl4th3) ([raw file](https://github.com/servo/webrender/blob/e5d48642dcb481fab473cfe0fd187f981d712b9e/webrender/src/util.rs#L371)):*

do we still need the constructors though?..
</blockquote></details>

The constructors are still used in a couple places to avoid having to use explicitly specify the types in order to satisfy the type inference. I think it's better to be explicit in those cases rather than rely on the inference + writing the type.

---

*[webrender/src/util.rs, line 115 at r2](https://reviewable.io:443/reviews/servo/webrender/2444#-L5omuMOm7er9r4eQYYp:-L5p-gSmKJpRf9cAub4x:bu2f2h1) ([raw file](https://github.com/servo/webrender/blob/8fe094ebc72aa2cde5e1d500d3b2b8e94dba61a3/webrender/src/util.rs#L115)):*

I think this should come with `abs()`
</blockquote></details>

Okay. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2444)*

PR #2445, opened 2018-02-20
Use Pathfinder's more accurate area approximation for distance AA.

pcwalton:
Smoothstep is not a very good approximation because it has zero
derivatives at 0.0 and 1.0. It makes edges too sharp (i.e. more
aliased).

From the comments in `prim_shared.glsl`:

This cubic polynomial approximates the area of a 1x1 pixel square under a
line, given the signed Euclidean distance from the center of the square to
that line. Calculating the *exact* area would require taking into account
not only this distance but also the angle of the line. However, in
practice, this complexity is not required, as the area is roughly the same
regardless of the angle.

The coefficients of this polynomial were determined through least-squares
regression and are accurate to within 2.16% of the total area of the pixel
square 95% of the time, with a maximum error of 3.53%.

Happy to answer any questions you might have.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2445)


pcwalton:
Here's an image showing the difference. "New" is with this patch, while "old" is the master branch.

border-radii-difference

Notice that the pixels inside the red box have a smoother look with this patch, especially the top row.

pcwalton:
Fixed the merge conflict and rebased.

glennw:
Pending Gecko try:

https://hg.mozilla.org/try/rev/cb882891f8f7a85aac19a10da1b10bf7aa63d0cc
https://treeherder.mozilla.org/#/jobs?repo=try&revision=cb882891f8f7a85aac19a10da1b10bf7aa63d0cc

webrender/res/prim_shared.glsl
```diff
@@ -747,22 +749,41 @@ float compute_aa_range(vec2 position) {
...
+/// The coefficients of this polynomial were determined through least-squares
+/// regression and are accurate to within 2.16% of the total area of the pixel
+/// square 95% of the time, with a maximum error of 3.53%.
+///
+/// See the comments in `compute_aa_range()` for more information on the
+/// cutoff values of -0.5 and 0.5.
 float distance_aa(float aa_range, float signed_distance) {
-    return 1.0 - smoothstep(-aa_range, aa_range, signed_distance);
+    float dist = 0.5 * signed_distance / aa_range;
+    if (dist <= -0.5 + EPSILON)
```
kvark: why do we need to take `EPSILON` into account here?

pcwalton: If the `EPSILON` fudge factor isn't taken into account, `reftests/clip/custom-clip-chains.yaml` has some minor 1 pixel differences on the sides due to what I presume to be floating point error.

glennw:
I think this is fine, but I'm going to wait until https://github.com/servo/webrender/pull/2454 lands to fix the intermittents, and then do another try run with this to be safe.

glennw:
Rebased on the intermittents fix in master. Pending try:

https://hg.mozilla.org/try/rev/cf4d827c6ae079c708e188b96e8d6ae2296c3293
https://treeherder.mozilla.org/#/jobs?repo=try&revision=cf4d827c6ae079c708e188b96e8d6ae2296c3293

glennw:
I checked each of the failures in the reftest analyzer, and they are all subpixel AA differences. So we can land this and fuzz those, since this is clearly an improvement in quality.

@bors-servo r+

staktrace:
@pcwalton @glennw  Any theories as to why the box-decoration-break-with-inset-box-shadow-1.html reftest (in R4) is behaving nondeterministically? In opt it seems to always have 8924 differing pixels but on debug occasionally there's only 8772 differing pixels. This is the case in both the try push from https://github.com/servo/webrender/pull/2445#issuecomment-367574374 and the one at https://treeherder.mozilla.org/#/jobs?repo=try&revision=f2c0d91c2fef35bea888505733b98c824c92110c (I did some retriggers on both).

pcwalton:
@staktrace Very strange. This is a GPU-code-only change, which shouldn't be affected by debug vs. release…

I would guess the issue is likely some sort of race in the CPU-side code that debug vs. release triggers.

staktrace:
It's likely not specific to debug vs release, but just a race somewhere that happens to get hit on debug more frequently.

glennw:
@staktrace I think this is *probably* yet another instance of https://github.com/servo/webrender/issues/1943 occurring, since box shadows are involved. I've assigned that bug to me, since it seems to be the main cause of non-determinism we have. I'll see if I can get it fixed next week.

PR #2446, opened 2018-02-21
Update x11-dl version.

glennw:
As of this morning, I am seeing XCB SIGILL crashes whenever I
try to load wrench in release mode.

Updating x11-dl appears to fix this. The commit notes between
the version we're using and the current version include a few
references to undefined behaviour, and incorrect bindings, so
it seems reasonable that this fixes the crash.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2446)


glennw:
r? anyone

PR #2447, opened 2018-02-21
Fix external texture UVs that are inverted with image brush shader.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2447)


glennw:
r? @kvark or anyone

We can use a rawtest to test this. However, raw tests don't currently support external images, so for now I think we should get this merged to unblock updating WR in Servo.

kvark:
oops, wrong bors :sweat_smile: 
@bors-servo r+

PR #2454, opened 2018-02-22
Share gradient GPU cache entries for repeated gradient primitives.

glennw:
This is a potential fix for the intermittent failures here:
https://github.com/servo/webrender/pull/2441#issuecomment-367211596

I still can't reproduce them locally. However, I could see that the
GPU cache size in that test was getting dangerously close to the
maximum size of the texture supported by the GL implementation.

That would fit with the issues we're seeing on CI, where there
seems to be items not being drawn, and they are affected even
though the test in question doesn't draw any gradients.

This is not a very elegant fix - however it will hopefully fix
the issues we are seeing on CI for now. We can revisit this
later to consider a better fix.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2454)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -71,6 +71,21 @@ impl PrimitiveOpacity {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+pub struct CachedGradientIndex(pub usize);
+
+pub struct CachedGradient {
+    pub handle: GpuCacheHandle,
+}
+
+impl CachedGradient {
+    pub fn new() -> CachedGradient {
+        CachedGradient {
+            handle: GpuCacheHandle::new(),
+        }
+    }
+}
+
 // Represents the local space rect of a list of
 // primitive runs. For most primitive runs, the
 // primitive runs are attached to the parent they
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark 

Pending try:
https://hg.mozilla.org/try/rev/62d31feeb7f89ccca95018f8213c54e60c051eb5
https://treeherder.mozilla.org/#/jobs?repo=try&revision=62d31feeb7f89ccca95018f8213c54e60c051eb5

glennw:
Try looks green with this patch @staktrace

webrender/src/prim_store.rs
```diff
@@ -71,6 +71,21 @@ impl PrimitiveOpacity {
...
 }
 
+#[derive(Debug, Copy, Clone)]
+pub struct CachedGradientIndex(pub usize);
+
+pub struct CachedGradient {
+    pub handle: GpuCacheHandle,
+}
+
+impl CachedGradient {
```
kvark: we should probably just `derive(Default)` for cases like this

staktrace:
Fix looks good, thanks!

PR #2455, opened 2018-02-22
Properly handle optimized ClipChain in custom ClipChain nodes

### LINKED ISSUE #2388: ClipChain with two clips descending from mask has mask applied twice

mrobinson:
When a node included in a ClipScrollNode ClipChain is optimized away, we need to
ensure that custom ClipChains that rely on that node are using the
original node and not one of the parents. In order to achieve this, we
store a copy of the original ClipChainNode in the ClipScrollNode and use
it when building custom ClipChains.

Fixes #2388.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2455)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -382,29 +382,15 @@ impl ClipChain {
         }
     }
 
-    pub fn new_with_added_node(
-        &self,
-        work_item: ClipWorkItem,
-        local_clip_rect: LayerRect,
-        screen_outer_rect: DeviceIntRect,
-        screen_inner_rect: DeviceIntRect,
-    ) -> ClipChain {
+    pub fn new_with_added_node(&self, new_node: &ClipChainNode) -> ClipChain {
         // If the new node's inner rectangle completely surrounds our outer rectangle,
         // we can discard the new node entirely since it isn't going to affect anything.
-        if screen_inner_rect.contains_rect(&self.combined_outer_screen_rect) {
+        if new_node.screen_inner_rect.contains_rect(&self.combined_outer_screen_rect) {
             return self.clone();
         }
 
-        let new_node = ClipChainNode {
-            work_item,
-            local_clip_rect,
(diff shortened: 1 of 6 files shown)


mrobinson:
cc @staktrace

glennw:
The change seems fine - let's do a try run though before merging.

mrobinson:
@glennw Thanks! I kicked off a Gecko try job here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=321e364a093eff86a5a73d5b722806e06bd8d220

kvark:
Looks fairly green to me, thank you!
@bors-servo r=glennw

Issue #2434, opened 2018-02-16
Eliminate as many calls to Transform3D::inverse() as possible


mrobinson:
This operation consistently appears in CPU profiles, but most web pages are not using 3D transforms. For example see the profiles generated in #2363. A general plan for this would be to extend TransformOrOffset to handle all the operations we need from Transform3D. For instance, `inverse()` would often turn into the creation of a new TransformOrOffset with negative offsets. We would then replace every Transform3D with TransformOrOffset. I have a feeling this would make `inverse()` disappear from the profile for most pages.

glennw:
Another thing I noticed recently is that the ```is_2d``` method that determines if we have a transform or offset is very conservative. It doesn't use an epsilon so any floating point error that accumulates and results in 1.0000001, for example, makes it treat it as a transform. If we make this method apply an epsilon, we may end up with a lot more transforms that can be treated as offsets.

mrobinson:
This was fixed by #2444 and the suggestion from @glennw is tracked in #2457.

PR #2457, opened 2018-02-22
Be less strict when detecting 2D translations

mrobinson:
We can avoid comparing directly against 1 here and use an epsilon to
account for floating point error.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2457)


mrobinson:
This is a suggestion from #2434. The Gecko try job for this is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=0fd18cae59eff15f670587663adaa43223b6793a

PR #2458, opened 2018-02-22
Add an example app using the DirectComposition API

SimonSapin:
~This PR is not intended to be merged, only to have a place to discuss while showing some code.~

This is a standalone demo app that uses the `winit` crate to create a window, then uses COM APIs through the `winapi` crate to set up DirectComposition in this window with two "visuals", each a rectangle of a given size with its own Direct3D swap chain, render target, and context. One of the two visuals scrolls on `MouseWheel` events, without either of them being re-rendered.

For now `ClearRenderTargetView` is used to fill each rectangle with a solid RGBA color. This is intended to be replaced with rendering arbitrary Direct3D content.

~A next step is probably integrating ANGLE so that separate OpenGL contexts can be created to render into respective visuals. I’m not sure where to start for this, though. ANGLE’s own documentation seems to be more about building than APIs.~

Looking up `directcomposition angle` specifically does yield some interesting things. It looks like ANGLE itself has some explicit support for DirectComposition:

* https://bugs.chromium.org/p/chromium/issues/detail?id=524838
* https://bugs.chromium.org/p/angleproject/issues/detail?id=1178
* https://github.com/google/angle/blob/master/extensions/EGL_ANGLE_direct_composition.txt

CC @jrmuizel @pcwalton @glennw

----

**Update:** the demo now renders with WebRender and ANGLE:

capture2

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2458)


directcomposition/src/com.rs
```diff
@@ -0,0 +1,134 @@
...
+impl<T> ops::Deref for ComPtr<T> where T: Interface {
+    type Target = T;
+    fn deref(&self) -> &T {
+        unsafe { &*self.0 }
+    }
+}
+
+impl<T> ops::DerefMut for ComPtr<T> where T: Interface {
+    fn deref_mut(&mut self) -> &mut T {
+        unsafe { &mut *self.0 }
```
emilio: This is not sound, right? `ComPtr` is shared, so I can effectively do:

```rust
let mut a: ComPtr<T> = something_that_returns_a_comptr();
let mut b = a.clone();

let mut a: &mut T = &mut *a;
let mut b: &mut T = &mut *b;
```

And get mutable aliasing.

SimonSapin: Oh right. I added this hoping to use deref coercions to get `*mut IUnknown`, ended up doing something else (see the `as_ptr` function), and forgot to remove it. retep998 had already pointed out that even `mem::replace` would be enough to make Bad Things happen if `DerefMut` exists.

jrmuizel:
The relevant code for hooking up ANGLE to DirectComposite exists in Chrome and Gecko in the following places:

https://cs.chromium.org/chromium/src/gpu/ipc/service/direct_composition_child_surface_win.cc?l=248&rcl=0a4de6f832840ebcd642c30b9e028d9853811af3

https://searchfox.org/mozilla-central/rev/9a8d3f8191b15cdca89a7ce044c7bea2dd0462dc/gfx/webrender_bindings/RenderCompositorANGLE.cpp#259

sotaroikeda:
Bug 1191971 has wip patches to add DirectComposition support to RenderCompositorANGLE.cpp of gecko. They just enables DirectComposition and use DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL swap chain. Just using DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL improved performance than DXGI_SWAP_EFFECT_SEQUENTIAL.

sotaroikeda:
ANGLE's support for DirectComposition seemed to be added originally for chromium's video overlay support. But current chromiun seemed not use it now. Instead, DirectCompositionSurfaceWin and DirectCompositionChildSurfaceWin are used.

One tricky thing about video overlay is that we need to swith between video overlay rendering and normal WebRender video rendering depends on scene situation.

In chromium, overlay usage decision is done aroundd DirectRenderer, OverlayProcessor and CompositorOverlayCandidateValidatorWin.
https://cs.chromium.org/chromium/src/components/viz/service/display/direct_renderer.cc
https://cs.chromium.org/chromium/src/components/viz/service/display/overlay_processor.cc
https://cs.chromium.org/chromium/src/components/viz/service/display_embedder/compositor_overlay_candidate_validator_win.cc

Actual DirectComposition layer is managed by DCLayerTree
https://cs.chromium.org/chromium/src/gpu/ipc/service/direct_composition_surface_win.cc?l=167

SimonSapin:
So. I was hoping to use https://github.com/servo/angle so that just adding a Cargo dependency would build ANGLE, but that crate only compiles just enough of ANGLE to support shader compilation.

The official build instructions are somewhat involved (I think it downloaded a copy of clang pre-compiled by Google?), so I did it manually and for now this demo assumes that `libEGL.dll` and `libGLESv2.dll` are available *somehow*.

I wrote some code similar to Gecko’s to create an EGL device, display, config, context, and surface, then call EGL’s `MakeCurrent` and GLES’s `ClearColor` instead of D3D’s `ClearRenderTargetView`. Everything runs without returning an error and Windows’s DebugView is silent, but the window stays blank. I’m not sure what to try next.

directcomposition/src/com.rs
```diff
@@ -1,5 +1,3 @@
-//! Similar to https://github.com/retep998/wio-rs/blob/44093f7db8/src/com.rs , but can be null
```
kvark: I'd prefer the module name to have `_` between words

pcwalton:
@SimonSapin Have you tried [RenderDoc]? As I understand it, it's the go-to tool for debugging graphics code on Windows.

Thanks for your work on this!

[RenderDoc]: https://github.com/baldurk/renderdoc

jrmuizel:
FWIW, I've started trying to get a appveyor config that can build ANGLE in CI.
https://github.com/jrmuizel/angle/blob/master/appveyor.yml

SimonSapin:
@kvark Done.

@pcwalton Thanks for the tip, but it looks like RenderDoc captures *frames* (and analyses how they’re rendered) but my non-functional EGL usage doesn’t manage to render a frame at all to begin with, so RenderDoc isn’t helpful in this case.

@sotaroikeda You’ve worked on similar code in Gecko, do you have a suggestion? An idea why my EGL surface doesn’t seem correctly "connected" to my DXGI swap chain?

sotaroikeda:
I seemed to reproduce the problem. I take a look.

sotaroikeda:
"visual.gleam.clear(gleam::gl::COLOR_BUFFER_BIT)" seems missing in render_plain_rgba_frame(). When I changed the function like the following. I saw the 2 squares in Window.

-----------------------------------------------------
fn render_plain_rgba_frame(visual: &D3DVisual, rgba: &[f32; 4]) {
    visual.make_current();
    visual.gleam.clear_color(rgba[0], rgba[1], rgba[2], rgba[3]);
    assert_eq!(visual.gleam.get_error(), 0);
    visual.gleam.clear(gleam::gl::COLOR_BUFFER_BIT); // Added line ***********
    visual.present();
}

SimonSapin:
That was it, thanks a lot!

SimonSapin:
Alright, now that we have multiple ANGLE contexts composited with DirectComposition, let’s put WebRender in them.

… and `webrender::Renderer::new` returns `Err(RendererError::MaxTextureSize)`. Indeed, `gl.get_integer_v(gl::MAX_TEXTURE_SIZE)` returns zero. This is with a gleam context created with ANGLE’s `eglGetProcAddress`. I’m hitting https://github.com/servo/webrender/issues/1260, but the only fix there was returning an error in this situation instead of panicking later.

That zero in ANGLE comes from: https://chromium.googlesource.com/angle/angle/+/a8e868f2450d3386ca8cf6dde51d7474ea41c067/src/libANGLE/renderer/d3d/d3d11/renderer11_utils.cpp#409

```cpp
size_t GetMaximum2DTextureSize(D3D_FEATURE_LEVEL featureLevel)
{
    switch (featureLevel)
    {
        case D3D_FEATURE_LEVEL_11_1:
        case D3D_FEATURE_LEVEL_11_0:
            return D3D11_REQ_TEXTURE2D_U_OR_V_DIMENSION;

        case D3D_FEATURE_LEVEL_10_1:
        case D3D_FEATURE_LEVEL_10_0:
            return D3D10_REQ_TEXTURE2D_U_OR_V_DIMENSION;

        case D3D_FEATURE_LEVEL_9_3:
            return D3D_FL9_3_REQ_TEXTURE2D_U_OR_V_DIMENSION;
        case D3D_FEATURE_LEVEL_9_2:
        case D3D_FEATURE_LEVEL_9_1:
            return D3D_FL9_1_REQ_TEXTURE2D_U_OR_V_DIMENSION;

        default:
            UNREACHABLE();
            return 0;
    }
}
```

It looks like `featureLevel` might be incorrect and I’m reaching the "unreachable" code. This, in turn, comes from `ID3D11Device::GetFeatureLevel`. But calling that method in my code returns `D3D_FEATURE_LEVEL_11_0 = 0xb000` which is one of the values that makes ANGLE return a non-zero max texture size.

SimonSapin:
This demo now uses WebRender to paint two rounded rectangles. (Each with a separate `webrender::Renderer` and a separate DXGI swap chain.)

However there is one bug left that I don’t understand: what shows up on screen appears to be one frame late. This did not happen in the previous commit that uses OpenGL directly.

SimonSapin:
> one frame late

This is fixed (by blocking on a channel until we’re notified that the pipeline is ready before rendering, like Wrench does). Thanks @jrmuizel for the suggestion.

With this, I think this demo is pretty much done?

jrmuizel:
Yep. We should try to get this mergeable.

After that we'll need to look at what's needed to have a single WebRender Renderer painting into multiple directcomposition visuals. (Probably with different documents going into different visuals). Then we can look at hooking up scrolling and video somehow.

SimonSapin:
@jrmuizel In what form would you want to merge this? I intended this code as a stand-alone, it’s only in a branch of this repo because I didn’t bother creating a separate repo. If this is to become a library, what should be in scope for it? (For example, creating a D3D device? Creating a window?)

At first I tried having a single renderer with multiple documents, but as far as I can tell both documents ended up rendered in both visuals. I think that what would be needed is some API to tell WR to only render one document.

SimonSapin:
Also if we want this to be more than a demo we probably need to get full ANGLE packaged in a new crate, or expand https://github.com/servo/angle.

SimonSapin:
> get full ANGLE packaged in a new crate

I did this at https://github.com/servo/mozangle, by reading importing `gfx/angle` from mozilla-central and reading the generated `moz.build` files.

This demo app should now build with just `cargo build`, no need to build ANGLE separately.

CC @jdashg @jrmuizel @emilio

SimonSapin:
This hits https://github.com/rust-lang/cargo/issues/4866#issuecomment-362930403 when `.travis.yml` runs `cargo test --all`, the Windows-only `mozangle/egl` feature ends up being selected even on non-Windows…

Perhaps I should make that Cargo feature be a no-op rather than a compile error, on platforms where it’s not supported?

SimonSapin:
I’ve also made AppVeyor (try to) build this app, but:

```
gfx/angle/checkout/src/common/platform.h:65:10: fatal error: d3d11_3.h: No such file or directory
```

SimonSapin:
Oh, `appveyor.yml` uses the `x86_64-pc-windows-gnu` toolchain, whereas I’m using MSVC on my machine.

jrmuizel:
We should switch appveyor to to the msvc toolchain instead of gnu one. We don't use the gnu one in production

PR #2459, opened 2018-02-22
Check for texture size

kvark:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2459)


kvark:
It's not entirely clear how to gracefully handle this case, especially for GPU cache being rather sensitive. Some checks are still better than no checks though :)

(as per IRC)
@bors-servo r=glennw

PR #2461, opened 2018-02-23
Track pipeline epochs in two places rather than three.

nical:
Also make sure that both the current and pending scenes get updated at the right time:
 - The pending scene should have all epoch information as soon as the transaction is received
 - The current scene should receive epoch updates at the end of the transaction (if scenes are built asynchronously it means when receiving the transaction from the scene builder thread).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2461)

### DIFF
--- a/webrender/src/frame.rs
+++ b/webrender/src/frame.rs
@@ -1116,6 +1116,8 @@ impl<'a> FlattenContext<'a> {
 }
 
 pub fn build_scene(config: &FrameBuilderConfig, request: SceneRequest) -> BuiltScene {
+    // TODO: mutably pass the scene and update its own pipeline epoch map instead of
+    // creating a new one here.
     let mut pipeline_epoch_map = FastHashMap::default();
     let mut clip_scroll_tree = ClipScrollTree::new();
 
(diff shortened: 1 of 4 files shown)


webrender/src/frame.rs
```diff
@@ -1116,6 +1116,8 @@ impl<'a> FlattenContext<'a> {
 }
 
 pub fn build_scene(config: &FrameBuilderConfig, request: SceneRequest) -> BuiltScene {
+    // TODO: mutably pass the scene and update its own pipeline epoch map instead of
```
leeoniya: probably `TODO(nical):`

webrender/src/frame.rs
```diff
@@ -1131,11 +1133,13 @@ pub fn build_scene(config: &FrameBuilderConfig, request: SceneRequest) -> BuiltS
         &mut pipeline_epoch_map
```
kvark: couldn't we just pass `&mut request.scene.pipeline_epochs` here?

mrobinson:
This looks good to me as well.

webrender/src/frame.rs
nical: The scene flattening reads from a pipeline epoch map, and builds a new one (which has the effect of filtering out pipeline epochs that are not in the scene), so there has to be two epoch maps at some point. That said I agree that the way it is presented here isn't very obvious, I'll look into making this a bit nicer.

glennw:
Is this ready to merge @nical ? Is this a fix for some of the orange-ness on try or something else?

nical:
@glennw it is a fix for (some of) the gecko fallouts from async scene building. It's ready to merge.

PR #2462, opened 2018-02-23
Properly free resources on wrench reftests

kvark:
some graphics debuggers are not happy to capture stuff that doesn't have resources properly cleaned up
r? anyone

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2462)


kvark:
Thanks for taking a look!
@bors-servo r=mrobinson

PR #2463, opened 2018-02-23
Add debug output to identify the cause of frequent CI failures.

jdm:
This will help narrow down the actual cause of https://github.com/servo/servo/issues/13480.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2463)


webrender_api/src/api.rs
```diff
@@ -714,10 +714,21 @@ impl RenderApiSender {
             channel::msg_channel().expect("Failed to create channel");
         let msg = ApiMsg::CloneApi(sync_tx);
         self.api_sender.send(msg).expect("Failed to send CloneApi message");
+        let namespace_id = match sync_rx.recv() {
```
jrmuizel: Maybe add a comment about how this code is for debugging an issue, so we can discard it in the future if we want.

PR #2466, opened 2018-02-26
Port drop-shadow and blur filter to use brush_image.

glennw:
There are a couple of goals that will come out of this work:

 * We want to remove brush_picture shader altogether, collapsing
   it into brush_image. This is the first part of that. The final
   part is modifying box-shadows to draw with segments so that we
   don't have to tweak the UVs in the fragment shader. Fixing this
   will also simplify a number of other tasks related to existing
   box-shadow bugs.
 * Output descriptors of render tasks are written into the GPU cache
   as a normal ImageResource structure, the same format as items in
   the texture cache. The idea here is to allow the image drawing
   shader to not care whether the source is a render task or a
   normal static image. This simplifies the image drawing shader,
   and also helps get the code into a better position to cache
   the outputs of more render task types in the texture cache.

Adding a color to every texture cache item isn't ideal. I did some
profiling and it didn't make any noticeable performance impact.
In the future, if we notice a cost to it, we could add a separate
shader feature for A8 vs RGBA8 images, and have a different
texture cache item format for each of those shaders.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2466)

### DIFF
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#define VECS_PER_SPECIFIC_BRUSH 0
+#define VECS_PER_SPECIFIC_BRUSH 1
 
 #include shared,prim_shared,brush
 
@@ -12,8 +12,13 @@ varying vec2 vLocalPos;
 
 varying vec3 vUv;
 flat varying vec4 vUvBounds;
-
+flat varying vec4 vColor;
+flat varying vec2 vSelect;
 #ifdef WR_VERTEX_SHADER
+
+#define IMAGE_SOURCE_COLOR              0
+#define IMAGE_SOURCE_ALPHA              1
+#define IMAGE_SOURCE_MASK_FROM_COLOR    2
 
 void brush_vs(
     VertexInfo vi,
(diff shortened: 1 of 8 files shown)


glennw:
r? @kvark 

Try looks green:
https://hg.mozilla.org/try/rev/bc15c65726a3c00c77dffc10145dc6428ee10f2a
https://treeherder.mozilla.org/#/jobs?repo=try&revision=bc15c65726a3c00c77dffc10145dc6428ee10f2a

kvark:
:lgtm:

---

Reviewed 8 of 8 files at r1.

---

*[webrender/res/brush_image.glsl, line 17 at r1](https://reviewable.io:443/reviews/servo/webrender/2466#-L6HXe5g6XquU7ZrcpP7:-L6HXe5g6XquU7ZrcpP8:b-nzrmzo) ([raw file](https://github.com/servo/webrender/blob/84a7058c83b033f014821786e9323112bcb26c32/webrender/res/brush_image.glsl#L17)):*
> ```GLSL
> flat varying vec4 vColor;
> flat varying vec2 vSelect;
> #ifdef WR_VERTEX_SHADER
> ```

let's not mix this with the varyings declaration (add a new line)

---

*[webrender/src/gpu_types.rs, line 276 at r1](https://reviewable.io:443/reviews/servo/webrender/2466#-L6HXUme1Z8AiMVT0euM:-L6HXUme1Z8AiMVT0euN:by6fkfu) ([raw file](https://github.com/servo/webrender/blob/84a7058c83b033f014821786e9323112bcb26c32/webrender/src/gpu_types.rs#L276)):*
> ```Rust
>         ]);
>         request.push(self.color);
>     }
> ```

nit: newline

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2466#-:-L6HYMimWMJy1zvIumIq:bnfp4nl)*

kvark:
Minor things, can be followed up with.
Thank you!
@bors-servo r+

PR #2468, opened 2018-02-26
Create DisplayListFlattener and move all flattening code there

mrobinson:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2468)


mrobinson:
This is a reduction in the amount of code, but the line count is almost neutral to the additional documentation.

nical:
It's too bad git couldn't figure out that frame.rs was renamed and destroyed/recreated all of its code (making digging in the history harder).

mrobinson:
@nical This is two commits and it seems if you look just at the first commit the rename is detected properly: https://github.com/servo/webrender/pull/2468/commits/559f8c334cc8a6a4a5715fe2dbc103e7a42e1bac

PR #2471, opened 2018-02-27
Remove the single corner optimization for box-shadows.

glennw:
This removes an optimization where we draw only a single corner
of a box-shadow with uniform radii, and mirror that across the
primitive.

There's a number of reasons for removing this:
 * It's no longer a very important optimization. We now ensure
   that we cache box-shadow render tasks (across frames), and draw
   the smallest possible nine-patch and optimized blur possible.
   Instead, this will allow us to simplify and optimize the shader
   that *applies* the box-shadow result on screen, which is typically
   run much more often and on many more pixels.
 * It simplifies the box-shadow code, which has a few accuracy
   bugs related to rounding. This reduces the number of code paths
   while we further optimize / fix these.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2471)


glennw:
r? @kvark or anyone

Pending try:
https://hg.mozilla.org/try/rev/4a4db45338f9268dbe91e78eefb212910ece2d4c
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4a4db45338f9268dbe91e78eefb212910ece2d4c

glennw:
Added a follow up commit that removes some more unused shaders and code paths. Pending try with that included:

https://hg.mozilla.org/try/rev/3044e2e1a555330cae10ddae11ceaa5a586455b1
https://treeherder.mozilla.org/#/jobs?repo=try&revision=3044e2e1a555330cae10ddae11ceaa5a586455b1

glennw:
Try looks good - just one new PASS in R6.

kvark:
:lgtm:

---

Reviewed 7 of 10 files at r1, 7 of 7 files at r2.

---

*[webrender/src/batch.rs, line 50 at r2](https://reviewable.io:443/reviews/servo/webrender/2471#-L6KEJ5DYDUuAmQY8isq:-L6KEJ5F5EcfjNhcRajv:b-yfzsro) ([raw file](https://github.com/servo/webrender/blob/3b72e0f96860c7d625bb2609020137ffb5dd0c46/webrender/src/batch.rs#L50)):*
> ```Rust
> pub enum BrushImageSourceKind {
>     Color = 0,
>     //Alpha = 1,            // Unused for now, but left here as shaders need to match.
> ```

we could call it `_Alpha` instead of commenting out

---

*[webrender/src/tiling.rs, line 484 at r2](https://reviewable.io:443/reviews/servo/webrender/2471#-L6KFm29hG_Z0V8Xmchq:-L6KFm29hG_Z0V8Xmchr:b-fdznap) ([raw file](https://github.com/servo/webrender/blob/4e374f0c5139ca81ee8de99afe2fc3bc53e3632e/webrender/src/tiling.rs#L484)):*
> ```Rust
>     pub clip_batcher: ClipBatcher,
>     pub brush_mask_corners: Vec<PrimitiveInstance>,
>     pub brush_mask_rounded_rects: Vec<PrimitiveInstance>,
> ```

should this one be renamed now for clarity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2471#-:-L6KFr_L2VAHeqb_wyVw:bnfp4nl)*

glennw:
Thanks for the quick review. If all goes according to plan with the current patch I'm working on, we'll start using the ```Alpha``` enum field then, and ```brush_mask_rounded_rects``` will also disappear. So I think it's probably OK to merge this now.

PR #2474, opened 2018-02-27
Simplify the way pipeline epochs are handled after building the scene.

nical:
This makes `DisplayListFlattener::create_frame_builder` update the current scene instead of just updating the pipeline epoch map, avoiding the mistake of updating the pipeline epoch map and then overwriting it by copying the pending scene into the current one.

The PR also merges render_bcakend.rs's create_frame_builder and build_scene, which makes things a little clearer.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2474)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -216,14 +216,13 @@ impl<'a> DisplayListFlattener<'a> {
         view: &DocumentView,
         output_pipelines: &FastHashSet<PipelineId>,
         frame_builder_config: &FrameBuilderConfig,
-        pipeline_epochs: &mut FastHashMap<PipelineId, Epoch>,
+        new_scene: &mut Scene,
     ) -> FrameBuilder {
         // We checked that the root pipeline is available on the render backend.
         let root_pipeline_id = scene.root_pipeline_id.unwrap();
         let root_pipeline = scene.pipelines.get(&root_pipeline_id).unwrap();
 
         let root_epoch = scene.pipeline_epochs[&root_pipeline_id];
-        pipeline_epochs.insert(root_pipeline_id, root_epoch);
 
         let background_color = root_pipeline
             .background_color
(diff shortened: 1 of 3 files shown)


nical:
Even though this fixes a bug I don't think it'll actually fix the oranges on try, but I am having a hard time checking because cargo-vendor fails to build on my laptop right now and I need to run it to do the wr update and push to try.

r? anyone

glennw:
I attempted to set up a try run for this, but failed - https://treeherder.mozilla.org/#/jobs?repo=try&revision=5798c3683571044cee84805f14a6270bf931e863.

With a local build of the changes in the try patch above, it just crashes immediately with:

```
[Child 21019, Chrome_ChildThread] WARNING: pipe error (22): Connection reset by peer: file ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 353
```

I haven't looked into it further to see if I did something wrong when getting Gecko compiling with the latest WR (probably) or if it's something else.

staktrace:
I just kicked off a try push at this revision: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9a8a6b307ab785057825bd572156a8dab3a95e77

staktrace:
Sadly the try push still doesn't look great.

glennw:
Do we want to merge this now or wait until the try push is looking better?

nical:
I'd merge it since it simplifies things a bit and fixes a leak.

staktrace:
@nical This PR might have made the windows webgl mochitests (which I only turned on recently) more orange. They were green initially, then something in the WR update made them intermittently orange, and with this PR they're a lot more orange. Hard to say for sure though. https://bugzilla.mozilla.org/show_bug.cgi?id=1442422#c1

staktrace:
The webgl mochitest issues are likely the same root cause as https://bugzilla.mozilla.org/show_bug.cgi?id=1442608. Try push is in progress with a workaround for that bug and I'll see if it fixes the mochitests as well.

staktrace:
@nical This PR did have some sort of effect on the windows mochitests, it looks like it basically negated the workaround from bug 1442608 or something. Not sure exactly what but if your fix for bug 1442748 doesn't make stuff green again it will require investigation.

staktrace:
https://bugzilla.mozilla.org/show_bug.cgi?id=1442422#c4 has links to try pushes.

Issue #2438, opened 2018-02-17
"Port ps_blend to be a brush shader" regressions


Darkspirit:
Regression range: https://github.com/servo/webrender/compare/38df48c62c929a3603f3c6e6393c2be8f1c88903...d69bd28a055e2afb2ad234eea297dd45f9b4b4e5 ([webrender update](https://bugzilla.mozilla.org/show_bug.cgi?id=1437572))

- [x] [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1439059) | tab loading animation is broken ([Screencast](https://bug1439059.bmoattachments.org/attachment.cgi?id=8951824))

**Edit:**
Also affected:
- [x] [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1439083) | https://vuetifyjs.com/en/components/buttons (click on a button, "become a patron" on the top left and on a menu item on the left)
    - [ ] Reopened: Black glitch ([screencast](https://bugzilla.mozilla.org/attachment.cgi?id=8955567), [about:support](https://bugzilla.mozilla.org/show_bug.cgi?id=1439083#c5))
- [x] [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1440087) | http://www.senecacollege.ca/campuses/janelocation.html (click plus/minus to zoom the map)
    - [ ] Reopened: Black boxes ([screencast](https://bug1440087.bmoattachments.org/attachment.cgi?id=8955573))
- [x] [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1439257) | https://www.facebook.com/search/str/Firefox/keywords_search ([example screenshot](https://bugzilla.mozilla.org/attachment.cgi?id=8952016))
  - [ ] Reopened: Black boxes ([screencast](https://bug1439257.bmoattachments.org/attachment.cgi?id=8955578))
- [x] https://web.whatsapp.com (file sending menu; not 100% sure)
- [x] [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1440158) | Google Calendar animations show junk data (untested; not sure about the regression range, but it sounds similar to the other regressions)
[...]

PR #2477, opened 2018-02-28
Correctly handle blend primitives, where the contents are clipped.

glennw:
In some situations, the items within a filter stacking context will
have clips applied. In this case, we minimize the allocated size
of the intermediate surface, to avoid drawing redundant information
that won't be seen.

However, if the stacking context itself does not have the same clip
applied, then we need to ensure that when we draw the blend primitive,
we don't sample outside the smaller region of the intermediate
surface that was rendered.

Ideally, the stacking context should have a clip applied to it so
that we don't waste time drawing transparent pixels. Adding a clip
to the stacking context here would provide an optimization win in
many cases. However, even without one present, we should still
ensure correct rendering, thus this fix is required.

Also fix a case where brush primitives with segments were not setting
the correct bounding rect for a render task. This wasn't causing a
problem here, but may have caused other issues.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2477)


glennw:
This fixes at least one of the bugs listed in #2438. I think it will probably fix most or all of them, but I haven't confirmed that.

Pending try run to see if this breaks any existing reftests:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=b6987fdf2f5cad510d74179bb7196829fec1e00e

r? @kvark

glennw:
@staktrace Does the above bit seem reasonable? (specifically, does it seem reasonable from the Gecko side to assign the clip-chain ancestor to the stacking context as well as the contents of the stacking context?). This won't be necessary for correctness, but would be a performance optimization in some cases.

glennw:
Try looks good, I think.

There are a couple of failures - one is a WR backend panic (Vector image error Unknown error) - which I assume is unrelated to this change.

glennw:
The other failure looks like an unrelated intermittent.

kvark:
Great stuff! But one (non-minor) concern

---

Reviewed 6 of 6 files at r1.

---

*[webrender/res/brush_blend.glsl, line 152 at r1](https://reviewable.io:443/reviews/servo/webrender/2477#-L6SiVWVUeL6Ov0V9Oor:-L6SiVWXdslipUIwEbmM:bxxvjzc) ([raw file](https://github.com/servo/webrender/blob/4c73f896e31983d79ee1176a652a28148b5261a2/webrender/res/brush_blend.glsl#L152)):*
> ```GLSL
>     // Fail-safe to ensure that we don't sample outside the rendered
>     // portion of a blend source.
>     color.a *= point_inside_rect(vUv.xy, vUvBounds.xy, vUvBounds.zw);
> ```

I don't think this is correct. The `vUv.xy` **is** allowed to be outside `vUvBounds`, hence why we need those bounds :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2477#-:-L6SisEaltlvpRuKCL86:byufi08)*

staktrace:
> @staktrace Does the above bit seem reasonable? (specifically, does it seem reasonable from the Gecko side to assign the clip-chain ancestor to the stacking context as well as the contents of the stacking context?). This won't be necessary for correctness, but would be a performance optimization in some cases.

Just to be clear - you're asking for a change on the gecko side here, for better performance in WR, correct? Assuming that's correct - I think what you're proposing might be a little tricky to do, because we process the nsDisplayFilter item that generates the stacking context before we ever recurse into the contents, which is where the clips are (on the child items of the nsDisplayFilter). So if I understand correctly, we'd have to walk through the child items in nsDisplayFilter::CreateWebRenderCommands (just before we create the stacking context), and check to see if all the items have the same clip, and if so, apply it to the stacking context as well.

Rather than doing it this way, it might be easier to do it at a higher level, when we first create the nsDisplayFilter item. I'm not sure of the details of how the clips are assigned to that item but it might be easier at that point to ensure the item has the same clip chain as the items to which the filter applies. Either way, it might be worth filing a bug about it in bugzilla and after the Final Clipping Rewrite(TM) lands we can revisit this and see how hard it would be.

glennw:
@staktrace That's correct, but having thought about it a bit more, I think we can do this optimization entirely within WR itself. I'll add an issue for it.

glennw:
@mstange Yup, agreed.

PR #2480, opened 2018-02-28
Use a dedicated thread to buffer payloads, so avoiding deadlock

### LINKED ISSUE #2478: Deadlock when sending a transaction
This is (I think) the cause of https://github.com/servo/servo/issues/20127

The problem is that WR serializes transactions by first serializing the payload. then the transaction https://github.com/servo/webrender/blob/d7735c7da9844775d4a3e6ec8b6cea25fb0b0486/webrender_api/src/api.rs#L897-L903

Since IPC send is blocking if the buffer is full, sending the payloads can block before the transaction is sent, causing deadlock.

cc @glennw @antrik @jdm @pcwalton

asajeffrey:
Fixes #2478 

The problem is that sending a transaction first sends the payloads, then the transaction. This causes deadlock if sending the payload blocks, since the transaction is never sent, so the payloads are never removed from the buffer.

This PR fixes that, by adding a dedicated thread that routes from the payload IPC channel to a payload mpsc channel. It also removes a hot loop in the case of missing payloads.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2480)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -9,8 +9,7 @@ use api::{DeviceIntPoint, DevicePixelScale, DeviceUintPoint, DeviceUintRect, Dev
 use api::{DocumentId, DocumentLayer, ExternalScrollId, FrameMsg, HitTestResult};
 use api::{IdNamespace, LayerPoint, PipelineId, RenderNotifier, SceneMsg, ScrollClamping};
 use api::{ScrollEventPhase, ScrollLocation, ScrollNodeState, TransactionMsg, WorldPoint};
-use api::channel::{MsgReceiver, PayloadReceiver, PayloadReceiverHelperMethods};
-use api::channel::{PayloadSender, PayloadSenderHelperMethods};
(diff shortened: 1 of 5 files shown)


glennw:
Some CI tidy failures:

```
./webrender_api/src/channel_ipc.rs:9: use statement is not in alphabetical order
	expected: std::{error, io}
	found: std::sync::mpsc
```

asajeffrey:
@glennw fixed

glennw:
Looks good! Let's squash the commits and then this is ready to go.

kvark:
I wonder why we couldn't send the payload *after* transaction, which would also resolve the deadlock.

asajeffrey:
@kvark that would probably fix the problem, but a) it puts the WR thread into a hot loop until the payload arrives, and b) it tightly couples the sender and receiver which then have to agree about which is the current channel, which is rather brittle code. In the long run I think the plan is to move WR over to just using mpsc channels, and move any ipc over to the application.

PR #2482, opened 2018-03-01
Remove no-op "debug" config.

SimonSapin:
It was probably useful once upon a time, but it doesn’t do anything anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2482)


SimonSapin:

@bors-servo retry

SimonSapin:
Same again. The previous line in the log is 
> Running `rustc --crate-name webrender […] -C opt-level=3 […]`

kvark:
What's so special about this change? Other PRs aren't going instantly but still going through less iterations...
@bors-servo retry

SimonSapin:
My guess is that this PR is not special, just "unlucky": compiling the `webrender` crate in release mode on Travis takes close to 10 minutes, but with variability it ends up sometimes slightly under or slightly over the time out.

We can try using `travis_wait`: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

glennw:
@bors-servo 
 * 7th time lucky?

kvark:
Ok, so this little unused boolean is what's been secretly keeping our build times in check? :D

SimonSapin:
If other PRs are landing without the same difficulty I must say that my theory is becoming less and less statistically likely, but I still don’t have a better one :/

staktrace:
Have you tried reproducing the build time difference locally?

glennw:
@bors-servo retry
 * Trying again on the off-chance that recent code changes might prevent the timeout.

PR #2484, opened 2018-03-01
Fix missing break statement from b4e030dc

staktrace:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2484)


kvark:
LGTM
@bors-servo r+

PR #2485, opened 2018-03-01
Synchronously shut the render backend down after reftests and rawtests.

nical:
To avoid going through the debacle of the infinite render backend loop after the async scene building PR, let's make sure we do get at least a test that checks that the render backend does not spin forever.

The first commit rearranges the the condition that breaks out of the render backend loop in a way that is slightly less error prone.

The second commit makes sure we wait for the shut down notification to have been propagated to the render notifier before exiting reftests and rawtests.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2485)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -695,8 +695,9 @@ impl RenderBackend {
 
     pub fn run(&mut self, mut profile_counters: BackendProfileCounters) {
         let mut frame_counter: u32 = 0;
+        let mut keep_going = true;
 
-        loop {
+        while keep_going {
             profile_scope!("handle_msg");
 
             while let Ok(msg) = self.scene_rx.try_recv() {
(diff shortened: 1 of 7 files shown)


glennw:
CI errors:

```
./wrench/src/rawtest.rs:6: use statement is not in alphabetical order
	expected: {WindowWrapper, NotifierEvent}
	found: blob
  Progress: 20% (137/681)
./wrench/src/wrench.rs:19: use statement is not in alphabetical order
	expected: std::sync::{Arc, Mutex}
	found: std::sync::mpsc::Receiver
  Progress: 20% (138/681)
  Progress: 20% (139/681)
  Progress: 20% (140/681)
  Progress: 20% (141/681)
./wrench/src/reftest.rs:6: use statement is not in alphabetical order
	expected: {WindowWrapper, NotifierEvent}
	found: base64
  Progress: 20% (142/681)
./wrench/src/png.rs:6: use statement is not in alphabetical order
	expected: {WindowWrapper, NotifierEvent}
	found: image::png::PNGEncoder
```

mrobinson:
For this lint to pass, capitalized struct names from within the crate go last instead of first.

glennw:
@nical Since we're still waiting on the PR to remove the CI error here, perhaps it's easiest to fix this PR up and get it landed in the meantime?

PR #2487, opened 2018-03-05
Switch box-shadows to be a clip source instead of a picture.

glennw:
In hindsight, drawing box-shadows as a picture was not a good
abstraction. Instead, we now treat them as a simple clip source
that writes to the primitive clip mask, and is applied to the
box-shadow primitive which is just a solid rectangle.

This has a number of benefits:
 * The code for inset box-shadows is significantly simpler.
 * We remove two shaders (brush_picture and brush_mask) completely.
 * We get better batching, since box-shadows are just a solid rect + clip mask.
 * When we start to cache clip masks, box-shadows will gain this automatically.
 * Both outset and inset box-shadows now correctly use segments, which makes inset shadows significantly faster than previously.
 * Since the code is abstracted from Picture, fixing up how we handle sub-pixel blur radii etc is easier.
  * This is not done in this PR, but is much simpler to do than it would have been previously.

Overview of specific changes:
 * Remove PictureKind::BoxShadow.
 * Add ClipMaskSource::BoxShadow.
 * Use segment rendering to minimize clip mask allocated size and pixels for all box-shadows.
 * Remove brush_picture and brush_mask_rounded_rect shader.
 * Remove scale_factor from blur task - no longer needed.
  * This makes possible some future work to fix up existing drop-shadow bugs.
 * Add cs_clip_box_shadow shader. Can probably be unified with cs_clip_image quite easily.
 * (Initial) support for local-space clip mask generation. Needed for future tasks related to rasterizing animated stacking contexts.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2487)


glennw:
Apologies for the huge size of this commit. If it's too difficult to review in one chunk, let me know and I'll try to split it up.

Try run here - https://treeherder.mozilla.org/#/jobs?repo=try&revision=ce6605ff176b62aa4895471fba9f0ce12f6b6d85&selectedJob=165917107

There's a lot of orange - but I believe the Windows test failures are unrelated (they seem to be present in wr-future as well). 

There are a few genuine failures in R2, R6, R7 related to this change. I'm investigating these now to find the root cause.

r? @kvark

glennw:
I think most (perhaps all) of the failures are related to some inaccuracies in our clip-mask code (we don't appear to correctly apply snapping in clip-mask generation). I'm hoping that fixing those will not only fix the failures above, but also some other clip-related bugs we have right now.

glennw:
Interestingly, all the failures in R2, R6 and R7 are already fuzzy, on WR and other backends. So it may actually be viable to increase the fuzziness on these tests. But I'll do some more investigation tomorrow to see if we can find the root cause.

glennw:
Rebased. I wonder if switching the box-shadow code to use the ```to_cache_size``` will be enough to work around those reftest failures for now. Will try a gecko run shortly.

glennw:
Here is the latest try run:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=68c511b449b2a0f52c2615c42852985c4bcdecc3&selectedJob=166093364

Of the relevant differences:

In R4:
 ```boxshadow-large-border-radius.html``` and ```boxshadow-rotated.html``` - already fuzzed on WR, D2D and OSX. I think we could probably increase the fuzziness on these.

In R6:
 ```box-decoration-break-with-inset-box-shadow-1.html``` - new PASS.

In R7:
 ```input-radio-customerror.html``` and ```input-radio-nogroup-required-invalid.html``` - already fuzzed on WR and Skia. I think we could increase the fuzziness here slightly.

Given the above (and assuming the other orange items are irrelevant here, which I think is true), then I think this is fine to get reviewed and merged.

I still intend to investigate the fuzziness, but I am fairly confident that the problem is either (a) related to lack of clip mask snapping in WR or (b) Gecko related since other backends need fuzziness on these tests too. I did verify that in at least a couple of the tests, the display lists coming from Gecko are not quite the same (e.g. one had a position of 49.025 compared to 49.0, which would show up our clip mask snapping issue).

What do you think @staktrace @kvark ?

kvark:
:lgtm: awesome work!

---

Reviewed 35 of 35 files at r1.

---

*[webrender/src/batch.rs, line 524 at r1](https://reviewable.io:443/reviews/servo/webrender/2487#-L6tb0BGmoWTxUmkhdi5:-L6tb0BI0_IiJSL_-Yon:b-7m74ox) ([raw file](https://github.com/servo/webrender/blob/6196843c24bd63acb086c76655094e2b6627cb9f/webrender/src/batch.rs#L524)):*
> ```Rust
>             let batch = self.batch_list.get_suitable_batch(key, &pic_metadata.screen_rect.as_ref().expect("bug").clipped);
> 
>             let render_task_id = pic.surface.expect("BUG: unexpected surface in splitting");
> ```

nice!

---

*[webrender/src/clip.rs, line 158 at r1](https://reviewable.io:443/reviews/servo/webrender/2487#-L6tf298iJ3WRPCRCtLD:-L6tf298iJ3WRPCRCtLE:b-wvo6xv) ([raw file](https://github.com/servo/webrender/blob/6196843c24bd63acb086c76655094e2b6627cb9f/webrender/src/clip.rs#L158)):*
> ```Rust
>         // as a simple blit. This is necessary for correctness, since the
>         // blur of one corner may affect the blur in another corner.
>         if actual_shadow_rect_size.width > shadow_rect.size.width ||
> ```

what happens if the shadow is super slim but wide? Is it desirable to still cover the whole primitive?

---

*[webrender/src/picture.rs, line 30 at r1](https://reviewable.io:443/reviews/servo/webrender/2487#-L6tg7K338-C-_sk4BcT:-L6tg7K338-C-_sk4BcU:b-ur6if9) ([raw file](https://github.com/servo/webrender/blob/3f1ff076a03c89f3d5009f462d853bd9010244dd/webrender/src/picture.rs#L30)):*
> ```Rust
> // Note: zero-square tasks are prohibited in WR task tree, so
> // we ensure each dimension to be at least the length of 1 after rounding.
> fn to_cache_size(size: DeviceSize) -> DeviceIntSize {
> ```

what makes the new code free of this problem?

---

*[webrender/src/segment.rs, line 213 at r1](https://reviewable.io:443/reviews/servo/webrender/2487#-L6thDsVHWvtmK7C2nW8:-L6thDsVHWvtmK7C2nW9:bkt5txq) ([raw file](https://github.com/servo/webrender/blob/6196843c24bd63acb086c76655094e2b6627cb9f/webrender/src/segment.rs#L213)):*
> ```Rust
>         inner_clip_mode: Option<ClipMode>,
>     ) {
>         let p0 = outer_rect.origin;
> ```

might need a debug assert that inner rect is inside the outer rect?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2487#-:-L6thypyRQjozZiOg7TI:bqdl9us)*

glennw:
Review status: 34 of 35 files reviewed at latest revision, 4 unresolved discussions.

---

*[webrender/src/clip.rs, line 158 at r1](https://reviewable.io:443/reviews/servo/webrender/2487#-L6tf298iJ3WRPCRCtLD:-L6tl9el6Yx9T0WkT3D0:b-8butqm) ([raw file](https://github.com/servo/webrender/blob/6196843c24bd63acb086c76655094e2b6627cb9f/webrender/src/clip.rs#L158)):*

what happens if the shadow is super slim but wide? Is it desirable to still cover the whole primitive?
</blockquote></details>

It's not perfectly optimal in that case - but it adds a bit of complexity to handle that case when we apply the stretching in the cs_clip_box_shadow shader, so I left it for now.

---

*[webrender/src/picture.rs, line 30 at r1](https://reviewable.io:443/reviews/servo/webrender/2487#-L6tg7K338-C-_sk4BcT:-L6tlMatq9yoW9JwvY4B:b-360m4o) ([raw file](https://github.com/servo/webrender/blob/3f1ff076a03c89f3d5009f462d853bd9010244dd/webrender/src/picture.rs#L30)):*

what makes the new code free of this problem?
</blockquote></details>

It's not - I moved the function to util.rs, since it's used by the box shadow clip source code too.

---

*[webrender/src/segment.rs, line 213 at r1](https://reviewable.io:443/reviews/servo/webrender/2487#-L6thDsVHWvtmK7C2nW8:-L6tlVhhj45WVBjGopyy:b-896fix) ([raw file](https://github.com/servo/webrender/blob/6196843c24bd63acb086c76655094e2b6627cb9f/webrender/src/segment.rs#L213)):*

might need a debug assert that inner rect is inside the outer rect?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2487)*

staktrace:
> Here is the latest try run:
> 
> https://treeherder.mozilla.org/#/jobs?repo=try&revision=68c511b449b2a0f52c2615c42852985c4bcdecc3&selectedJob=166093364

boxshadow-large-border-radius.html - this goes up from (4, 360) to (27, 1680) which is a pretty big jump. Visually it's noticeably different. I looked at one of the existing win10 runs on m-c, for which the annotation allows (16, 999) of fuzziness - in that case the actual fuzziness is only (3, 908). So it seems to me like jumping to (27, 1680) warrants some investigation.

boxshadow-rotated.html - This one goes up from (43, 200) to (22, 3258) so it covers a much larger area but the pixel component differences are smaller. Visually also I can't really tell the difference so I'd be ok with adjusting the fuzz to cover this.

For the R7, note that it's not just input-radio-customerror.html and input-radio-nogroup-required-invalid.html. The tests in this folder are already nondeterministic with WR - sometimes they pass and sometimes they need fuzzing. If you compare the R7 debug vs opt jobs you'll see already there are different failures. I've retriggered these jobs to get a better sample of what's happening. But it looks like the fuzzy range will need to go up from fuzzy-if(webrender,0-14,0-8) to fuzzy-if(webrender,0-42,0-3) which I'm ok with. Would really like that nondeterministic behaviour squashed though since it makes it hard to tell when things are failing and which tests need annotating.

glennw:
@staktrace Yea, I definitely want to fix that non-determinism. I intend to do some work on this next week (I believe it's a general clip-mask issue). I'll take a closer look at ```boxshadow-large-border-radius.html``` today and see if I can make that any better in this PR without fixing the general clip-mask issue that I suspect to be the main cause. Thanks!

glennw:
@staktrace OK, with the last commit, we now have:

In R4:
```boxshadow-large-border-radius.html``` -> now becomes a PASS.
```boxshadow-rotated.html``` - agreed to fuzz this in comment above.

In R6:
```box-decoration-break-with-inset-box-shadow-1.html``` - FAIL. Was fuzzed ```11-11,8772-8924```. Max difference is increased to 28. Differing pixels is reduced to 5270.

In R7:
Various known intermittents as discussed above.

So I can take a detailed look at the R6 failure tomorrow, and see if there's anything we can do about it in this PR. Or we can increase the fuzz max difference on that test. Apart from that, everything else looks OK I think.

glennw:
Err, and the actual link to the try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=d797f83e8971f9766c5722cc33943802353ec1c4&selectedJob=166409271

staktrace:
@glennw LGTM, thanks!

PR #2488, opened 2018-03-05
Move ClipScrollNode inverse calculation back to CPU

mrobinson:
We have the ability to avoid expensive matrix math in most cases and,
where we cannot, we calculate this value on the CPU anyway, so avoid
calculating it on the GPU as well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2488)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -71,7 +71,7 @@ vec4[2] fetch_from_resource_cache_2(int address) {
 
 #ifdef WR_VERTEX_SHADER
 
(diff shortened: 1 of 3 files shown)


mrobinson:
Gecko try job for this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=48503e86d96397aa5e6d9d4843e37fb158517698

(still in progress)

glennw:
Looks good, thanks!

@bors-servo r+

glennw:
@bors-servo retry
 * network error

PR #2490, opened 2018-03-05
Ensure non-zero picture task size

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1442718

~~Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=dc0d35306d006a09ddb0cf7db453a97fa9bf8dca~~

Another try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=11b5d65d0e682c277992c1cea5d4048227184118

r? @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2490)


staktrace:
According to the comment the rounding was intentional to pass Gecko reftests. So presumably there will be reftest failures from this change? Let's wait and see what the try push results look like.

kvark:
@staktrace PTAL, another try push in flight

webrender/src/picture.rs
```diff
@@ -18,6 +19,19 @@ use resource_cache::CacheItem;
 use scene::{FilterOpHelpers, SceneProperties};
 use tiling::RenderTargetKind;
 
+
+// TODO(gw): Rounding the content rect here to device pixels is not
+// technically correct. Ideally we should ceil() here, and ensure that
+// the extra part pixel in the case of fractional sizes is correctly
+// handled. For now, just use rounding which passes the existing
+// Gecko tests.
```
staktrace: The comment should be updated to indicate that width/height of 0 causes badness, hence the max'ing with 1. But other than that, LGTM

kvark:
Fixed now, thanks for the review!
@bors-servo r=staktrace

kvark:

here we go again...
@bors-servo retry

glennw:
@kvark @staktrace I think it's fine to merge this as a bandaid fix, but I don't think it's the "correct" fix.

It seems like we should detect a zero-sized picture, and not be trying to add it to the texture cache at all?

That would probably be more efficient since we'd then skip building that picture, adding batches for it, allocating target space etc.

kvark:
@glennw I don't think we should ignore it here. It's not zero-sized, it's just less than 0.5 sized.
As your comments states, the proper use would be `ceil()`, and that's what we do here just for a very specific near-zero case instead of the general one.

glennw:
@kvark Ah, I see - that makes sense. Thanks!

PR #2491, opened 2018-03-06
textureSize() workaround for Angle/NV

kvark:
Temporary workaround for flickering issues of https://bugzilla.mozilla.org/show_bug.cgi?id=1431560

It is not covering all the cases, but on the other hand is rather harmless. Should ease the pain of Windows/Nvidia users while the proper cure (my Angle fix) is in flight. Check out [this comment](https://chromium-review.googlesource.com/c/angle/angle/+/949412#message-fe4f62522349e7754d545ffc8e96c999d3192d26) for nasty technical details.

WR driver bug entry: https://github.com/servo/webrender/wiki/Driver-issues#texturesize-in-vertex-shaders

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2491)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -86,6 +86,16 @@ uniform HIGHP_SAMPLER_FLOAT sampler2D sRenderTasks;
 in ivec4 aData0;
 in ivec4 aData1;
 
+// Work around Angle bug that forgets to update sampler metadata,
+// by making the use of those samplers uniform across programs.
+// https://github.com/servo/webrender/wiki/Driver-issues#texturesize-in-vertex-shaders
+void markCacheTexturesUsed() {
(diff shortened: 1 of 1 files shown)


glennw:
Wow, nice detective work.

Should we add an issue to remove this once the ANGLE fixes land in Gecko?

@bors-servo r+

kvark:
Here is the gist of the situation:
  - Angle has a specific workaround for Nvidia that uses the `baseLevel` of a sampler for backing `textureSize` calls instead of the parameter we give in the shader (just `0`).
  - Angle stores the `baseLevel` in the special driver constants buffer. It's updated with `MAP_WRITE_DISCARD` mapping, then filling in exactly the data needed by the current program, and unmapping it. Consequently, everything above the number of samplers used by the program is turned into trash by the driver.
  - The buffer is invalidated based on a number of triggers, but unfortunately the shader program change is not one of them. Thus, once we switch to a bigger shader program, we sometimes get trash results in `textureSize` calls (and possibly other things), visible as flickering.

PR #2496, opened 2018-03-07
Do division and mod on unsigned integers

jrmuizel:
The hlsl compiler was complaining about doing division on signed
integers being slow so we'll cast do our div and cast back.

This fixes bug 1443747

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2496)

### DIFF
--- a/webrender/res/prim_shared.glsl
+++ b/webrender/res/prim_shared.glsl
@@ -100,7 +100,8 @@ void markCacheTexturesUsed() {
 // TODO: convert back to a function once the driver issues are resolved, if ever.
 // https://github.com/servo/webrender/pull/623
 // https://github.com/servo/servo/issues/13953
-#define get_fetch_uv(i, vpi)  ivec2(vpi * (i % (WR_MAX_VERTEX_TEXTURE_WIDTH/vpi)), i / (WR_MAX_VERTEX_TEXTURE_WIDTH/vpi))
+// Do the division with unsigned ints because that's more efficient with D3D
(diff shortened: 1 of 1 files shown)


webrender/res/prim_shared.glsl
```diff
@@ -100,7 +100,8 @@ void markCacheTexturesUsed() {
 // TODO: convert back to a function once the driver issues are resolved, if ever.
 // https://github.com/servo/webrender/pull/623
 // https://github.com/servo/servo/issues/13953
-#define get_fetch_uv(i, vpi)  ivec2(vpi * (i % (WR_MAX_VERTEX_TEXTURE_WIDTH/vpi)), i / (WR_MAX_VERTEX_TEXTURE_WIDTH/vpi))
+// Do the division with unsigned ints because that's more efficient with D3D
+#define get_fetch_uv(i, vpi)  ivec2(int(uint(vpi) * (uint(i) % uint(WR_MAX_VERTEX_TEXTURE_WIDTH/vpi))), int(uint(i) / uint(WR_MAX_VERTEX_TEXTURE_WIDTH/vpi)))
```
kvark: I'm not certain this is correct. `WR_MAX_VERTEX_TEXTURE_WIDTH` is defined as a simple integer constant, so `WR_MAX_VERTEX_TEXTURE_WIDTH / vpi` is still integer division. Also, I don't think GLSL requires you to do `uint(int(x), int(y))` as opposed to just `uint(x, y)`.

jrmuizel: WR_MAX_VERTEX_TEXTURE_WIDTH / vpi gets constant folded so using integer division there shouldn't matter.

jrmuizel: Also uint(x, y) doesn't seem to work:
'constructor' : too many arguments

kvark: Makes sense, thanks!

PR #2497, opened 2018-03-08
Small optimization for the brush_image shader.

glennw:
The mask select and swizzle are only needed for images running
in the alpha pass. Remove these FS instructions when drawing
image segments in the opaque pass.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2497)


glennw:
Pending try here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4be5f6d5691ae5825f7023e47358565f4f9faf08

r? @kvark or anyone

glennw:
Try run looks good - the reftest failures here are due to the box-shadow PR landing and needing new annotations.

PR #2499, opened 2018-03-08
Revert "textureSize() workaround for Angle/NV"

kvark:
This reverts #2491 for apparently causing https://bugzilla.mozilla.org/show_bug.cgi?id=1444076
Must be some real confusion by the driver, since the code changes are semantically harmless.
I'll come up with a better fix.
r? @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2499)


cpdean:
would this be related to this bug i'm seeing in servo? https://github.com/servo/servo/issues/20239

PR #2500, opened 2018-03-08
Clip blurred and shadowed text with local clip rect

mrobinson:
Take into account the local clipping rectangle in the brush and caching
text shaders. This allows the blurred shadow of text to be properly
clipped.

Co-authored-by: Alexis Beingessner <a.beingessner@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2500)


glennw:
Seems to fail reftests on CI. The change generally looks good to me, though we'll need a try run too once the CI issue is resolved.

mrobinson:
Whoops. Here is the try job for this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=17f13c2ccb9742d5360e7457b9ccbbcc94b0e1bb

It seems that the new test is failing on the bots. It could be that the test needs to be fuzzed since it uses a blurred shadow. I know that other similar tests have a fuzz defined

kvark:
@mrobinson FYI, Alexis is @Gankro on github

mrobinson:
@kvark Is there something different I could have done with the Co-author tag using @Gankro's GitHub nick?

glennw:
```
REFTEST TEST-UNEXPECTED-FAIL | reftests/text/blurred-shadow-local-clip-rect.yaml == reftests/text/blurred-shadow-local-clip-rect.png | image comparison, max difference: 181, number of differing pixels: 4357
```

The max difference in this is 181, so it seems to be something other than a fuzziness issue.

mrobinson:
It seems the issue here is that the rendered image is different on the bots. I have this issue with three others tests, also related to text rendering.

mrobinson:
The test is passing now after:

1. Getting the output image from the build bot.
2. Running this test only on Linux.

This seems to be the norm for many of the text tests, so it seems their output is very fragile. I wonder how we can improve this situation.

kvark:
@mrobinson oh, I didn't realize this is a github hook. TIL!

Gankro:
This should fix https://bugzilla.mozilla.org/show_bug.cgi?id=1401653 and unblock https://bugzilla.mozilla.org/show_bug.cgi?id=1403261

glennw:
@mrobinson Ah, I also have that problem with generating text reference images on my machine. I haven't had a chance to delve into the cause either.

kvark:

@bors-servo retry

mrobinson:
Thanks for the reviews!

Issue #2436, opened 2018-02-16
Share the shader/feature list between Renderer and the tests


kvark:
Letting them out of sync ends up with issues like https://github.com/servo/webrender/pull/2424#issuecomment-366298946

PR #2505, opened 2018-03-10
Nicer shader management

kvark:
@glennw 
This is a big (but shallow) refactor of the way we manage shaders:
  - move all the shader related logic into a separate module. Our `renderer` is still too big
  - separate getting a lazy shader pointer from its binding
  - introduce special handling of the `mode` uniform
  - modernize the shape a bit, including the use of `?` macro

The last bit, implemented in 0d6b922b43b4549172607a01e8bbbfd5af3f3682, is not essential to the PR but I believe it compliments it somewhat nicely. This commit can be seen as controversial because it makes program binding a little less functional. I'd argue that the device is already stateful (e.g. one is expected to bind a program first, then draw, not the other way around), but can remove the commit if there are objections.

The PR reduces the amount of code, separates the concerns better, and makes things nicer to work with overall. The desire to clean it up came to me after messing with the flickering workarounds...

I'm deliberately dropping this PR bomb during the weekend for a chance to avoid mergocalypse later on :)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2505)

### DIFF
--- a/webrender/src/debug_render.rs
+++ b/webrender/src/debug_render.rs
@@ -282,7 +282,7 @@ impl DebugRenderer {
             // Triangles
             if !self.tri_vertices.is_empty() {
                 device.bind_program(&self.color_program);
-                device.set_uniforms(&self.color_program, &projection, 0);
+                device.set_uniforms(&self.color_program, &projection);
                 device.bind_vao(&self.tri_vao);
                 device.update_vao_indices(&self.tri_vao, &self.tri_indices, VertexUsageHint::Dynamic);
                 device.update_vao_main_vertices(
@@ -296,7 +296,7 @@ impl DebugRenderer {
             // Lines
             if !self.line_vertices.is_empty() {
                 device.bind_program(&self.color_program);
-                device.set_uniforms(&self.color_program, &projection, 0);
+                device.set_uniforms(&self.color_program, &projection);
                 device.bind_vao(&self.line_vao);
                 device.update_vao_main_vertices(
                     &self.line_vao,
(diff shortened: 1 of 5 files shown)


kvark:
Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=02b1c1e9084a9208f925f4c288b69a6576dd4ed4

Also, isolation of the shaders could help with #2436 in the future.

glennw:
A lot of orange on the try run. I think the R reftest results are expected, not sure about those Windows failures though?

kvark:
All windows failures look like WebGL timeouts, which is hardly related. @staktrace any clues?

staktrace:
@kvark I don't know which base revision your try push was made from, but the webgl stuff is green on central. So either you're using an older base rev where stuff was broken, or your patch breaks those tests. I'd try rebasing to latest central to make sure it's not the former; the failures persist then it must be something in your patch.

kvark:
@staktrace: a try push after rebasing on latest central - https://treeherder.mozilla.org/#/jobs?repo=try&revision=47b5eac4436ebd319bdc0c5f4b7b5dfeb8cc3f90
I find it weird that it hasn't finished still (15 pending jobs?).

glennw:
Latest try looks good, thanks!

@bors-servo r+

staktrace:
> I find it weird that it hasn't finished still (15 pending jobs?).

@kvark Seems to be done now. Looking at the timestamps I guess the pending jobs were the talos jobs which run on different hardware. Sometimes there's a backlog.

PR #2507, opened 2018-03-12
Move the image decomposition code out of DisplayListFlattener.

nical:
This is a preamble for the task of doing lazy image decomposition during frame building. This PR should not change the current behavior. Decoupling the decomposition code from DisplayListFlattener will make it easier to move it to frame building as a followup.

There is a comment in prim_store.rs about creating `image.rs` and moving image sepcific logic there so I went ahead and created the file, which now holds the decomposition code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2507)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -5,10 +5,10 @@
 
 use api::{AlphaType, BorderDetails, BorderDisplayItem, BuiltDisplayListIter};
 use api::{ClipAndScrollInfo, ClipId, ColorF, ComplexClipRegion, DeviceIntPoint, DeviceIntRect};
-use api::{DeviceIntSize, DevicePixelScale, DeviceUintRect, DeviceUintSize};
+use api::{DeviceIntSize, DevicePixelScale, DeviceUintRect};
 use api::{DisplayItemRef, Epoch, ExtendMode, ExternalScrollId, FilterOp};
 use api::{FontInstanceKey, FontRenderMode, GlyphInstance, GlyphOptions, GradientStop};
-use api::{IframeDisplayItem, ImageDisplayItem, ImageKey, ImageRendering, ItemRange, LayerPoint};
+use api::{IframeDisplayItem, ImageKey, ImageRendering, ItemRange, LayerPoint};
 use api::{LayerPrimitiveInfo, LayerRect, LayerSize, LayerVector2D, LayoutSize, LayoutTransform};
 use api::{LayoutVector2D, LineOrientation, LineStyle, LocalClip, PipelineId};
 use api::{PropertyBinding, RepeatMode, ScrollFrameDisplayItem, ScrollPolicy, ScrollSensitivity};
@@ -19,10 +19,11 @@ use border::ImageBorderSegment;
 use clip::{ClipRegion, ClipSource, ClipSources, ClipStore};
 use clip_scroll_node::{ClipScrollNode, NodeType, StickyFrameInfo};
 use clip_scroll_tree::{ClipChainIndex, ClipScrollNodeIndex, ClipScrollTree};
-use euclid::{SideOffsets2D, rect, vec2};
+use euclid::{SideOffsets2D, vec2};
 use frame_builder::{FrameBuilder, FrameBuilderConfig};
 use glyph_rasterizer::FontInstance;
 use hit_test::{HitTestingItem, HitTestingRun};
+use image::{decompose_image, TiledImageInfo};
 use internal_types::{FastHashMap, FastHashSet};
 use picture::{PictureCompositeMode, PictureKind, PicturePrimitive};
 use prim_store::{BrushKind, BrushPrimitive, BrushSegmentDescriptor, CachedGradient};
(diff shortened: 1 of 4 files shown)


webrender/src/display_list_flattener.rs
```diff
@@ -638,12 +639,29 @@ impl<'a> DisplayListFlattener<'a> {
...
-                            tiling.tile_size as u32,
+                        decompose_image(
+                            &TiledImageInfo {
+                                rect: prim_info.rect,
+                                tile_spacing: info.tile_spacing,
+                                stretch_size: info.stretch_size,
+                                device_image_size: tiling.image_size,
+                                device_tile_size: tiling.tile_size as u32,
+                            },
+                            &mut|tile| {
```
mrobinson: Instead of passing a callback here, would it be possible to make the new function part of a DisplayListFlattener impl in the `image.rs` file? This is how `box_shadow.rs` and `border.rs` are structured.

nical: The goal is to decouple this code entirely from `DisplayListFlattener` so that in the next step I move it to the frame building phase (and make it operate only on the visible area).

kvark:
Reviewed 4 of 4 files at r1.

---

*[webrender/src/display_list_flattener.rs, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QAyU924BqNYAvH37_:-L7QAyU924BqNYAvH37a:b-1bqzcb) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/display_list_flattener.rs#L26)):*
> ```Rust
> use glyph_rasterizer::FontInstance;
> use hit_test::{HitTestingItem, HitTestingRun};
> use image::*;
> ```

do we have to globally import here?

---

*[webrender/src/image.rs, line 43 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QC4oNPGZ1r9DGYLZU:-L7QC4oNPGZ1r9DGYLZV:b-avnrc1) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/image.rs#L43)):*
> ```Rust
> /// in layout space, and the tiling that we do in the texture cache (to avoid hitting texture
> /// size limits). The latter is referred to as "device" tiling here to disambiguate.
> pub fn decompose_image(info: &TiledImageInfo, callback: &mut FnMut(&DecomposedTile)) {
> ```

I'm not certain that callbacks are really needed here. Wouldn't it be easier to just fill/return a list of `DecomposedTile` things?

---

*[webrender/src/image.rs, line 274 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QBxwNQycrhKtfPhHo:-L7QBxwNQycrhKtfPhHp:b-kejqgd) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/image.rs#L274)):*
> ```Rust
>     // Fix up the primitive's rect if it overflows the original item rect.
>     if let Some(rect) = prim_rect.intersection(&item_rect) {
>         callback(&DecomposedTile {
> ```

no need to pass it by reference here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2507)*

nical:

---

*[webrender/src/display_list_flattener.rs, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QAyU924BqNYAvH37_:-L7QFPTm_HPg6srI4yYh:babss3c) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/display_list_flattener.rs#L26)):*

do we have to globally import here?
</blockquote></details>

We don't need to but it doesn't matter since this is disappearing in a followup PR where tile decomposition moves out of the scene flattening entirely. Also we do use glob imports in various places in WebRender already.

---

*[webrender/src/image.rs, line 43 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QC4oNPGZ1r9DGYLZU:-L7QFzIq3uXDN-xKdOJK:b-izv4v1) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/image.rs#L43)):*

I'm not certain that callbacks are really needed here. Wouldn't it be easier to just fill/return a list of `DecomposedTile` things?
</blockquote></details>

The first version of this PR did exactly that, but wasn't any easier. You'd exchange a `callback(tile)` call with a `tiles.push(tile)`, and then a loop through the vector that does exactly what the callback body does here. Moving to the callback even made the code marginally more concise, but structurally it's really the same, without the (sometimes numerous) unnecessary allocations.

---

*[webrender/src/image.rs, line 274 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QBxwNQycrhKtfPhHo:-L7QH_J-dSud_ai7EOvH:btl0mxh) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/image.rs#L274)):*

no need to pass it by reference here
</blockquote></details>

Sure, no need to pass by value either, though. FWIW, some of the clang-based static analysis that we run on gecko's C++ code warns when we pass structs by value as soon as it gets bigger than 64 bits. Granted it's not the same language but it's the same underlying codegen framework that decided that passing by ref is preferable in most cases. Personally I wouldn't mind getting into the habit of passing 128bit+ structures by reference by default rather than the opposite.
I don't actually care very much in this case (just for the sake of the discussion).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2507)*

kvark:

---

*[webrender/src/display_list_flattener.rs, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QAyU924BqNYAvH37_:-L7QLOOQ52xDk3gClKai:b-kf18c6) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/display_list_flattener.rs#L26)):*
> it doesn't matter since this is disappearing in a followup PR

Well, as long as you do want to merge it as opposed to keep WIP until those follow-ups arrive, I'd like it to look landable.

> we do use glob imports in various places in WebRender already

There is a very limited number of cases where glob imports are considered idiomatic in Rust, namely: prelude and tests (sometimes, examples, as an extension of tests).
Scanning our modules in "webrender" project only shows `use rayon::prelude::*;`, which makes sense. Thus, I don't think a global import of `image` fits here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2507)*

nical:

---

*[webrender/src/display_list_flattener.rs, line 26 at r1](https://reviewable.io:443/reviews/servo/webrender/2507#-L7QAyU924BqNYAvH37_:-L7QT6IaaDUWvWm2-ykV:b682hf5) ([raw file](https://github.com/servo/webrender/blob/02ecd102ebf91a7c1f308b75c9e30cedace212ce/webrender/src/display_list_flattener.rs#L26)):*
> I'd like it to look landable.

Well that's a bit strong. It does look quite landable to me although I'll add the explicit imports if it makes you happy (even if I do see a few other glob imports that aren't related to rayon's prelude).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2507)*

kvark:
Looks like everything is addressed now, thank you!
@bors-servo r+

PR #2509, opened 2018-03-13
Use brush_image for blurs, instead of 2d hardware composite.

glennw:
Once this lands, we can fix drop-shadows by making use of this
functionality, and port the other remaining uses of ps_hardware_composite
to use brush_image. Then, ps_hardware_composite can be removed,
and everything except 3d plane-splits will then be drawn as
normal primitives.

Incidentally, this fixes the way we handle blurs when the source
picture is clipped by either the screen bounds or a clip-chain.
The code will now inflate the rendered portion of the source
render task correctly, so include enough pixels to get a correct
blur result at the edges.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2509)


glennw:
r? @kvark 

Pending try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7ab1171bbe9359d45612b7590a1e90b40c599a30

webrender/res/brush_image.glsl
```diff
@@ -26,6 +27,16 @@ flat varying vec2 vSelect;
...
 #endif
 
+struct ImageBrush {
+    RectWithSize rendered_task_rect;
+};
+
+ImageBrush fetch_image_primitive(int address) {
+    vec4 data = fetch_from_resource_cache_1(address);
+    RectWithSize rendered_task_rect = RectWithSize(data.xy, data.zw);
+    return ImageBrush(rendered_task_rect);
```
kvark: let's avoid returning with a struct constructor, no need to bring up those GLSL driver bugs from the dead :)

webrender/res/brush_image.glsl
```diff
@@ -48,17 +59,51 @@ void brush_vs(
...
+            ) / texture_size.xyxy;
+            break;
+        }
+        case RASTER_LOCAL:
+        default: {
+            f = (vi.local_pos - local_rect.p0) / local_rect.size;
+
+            // Set the clip bounds to a value that won't have any
+            // effect for local space images.
+            vUvClipBounds = vec4(-1.0, -1.0, 2.0, 2.0);
```
kvark: why -1.0 to 2.0 as opposed to 0.0 to 1.0 ?

webrender/res/brush_image.glsl
glennw: Fixed

glennw:
I fixed up those comments, and pushed a new try run that I hope will fix the one failure in the try above (the UV generated should be using the snapped device pos, not the unsnapped pos).

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=edd58cb4436687a19fdbeca8a3b4963accbcbd04

glennw:
Most recent try, with fast path fix to skip blurs when radius is zero.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=a90a105bc5d21d8939a702f953dbd7c4642e548e

Looks green so far, will just wait for the last few tests to complete.

glennw:
There's a few windows tests that didn't seem to kick off correctly, but I don't think that could be caused by this change - they also passed once they were automatically re-triggered.

@bors-servo r=kvark

PR #2512, opened 2018-03-13
No longer allow clipping items by complex clips

### LINKED ISSUE #1742: [RFC] Change public item APIs to support a local clip rect only.
We change the public display item APIs to support a local clip rect only. Any clips in addition to that (e.g. rounded rects, image masks etc) must be defined via a clip item and the clip ID for the item.

This seems like it would cover the common cases easily, without having to create a heap of clip items in the DL. It's also really easy for WR to handle the simple local clip rect case.

This could potentially allow us to both simplify and improve the performance of several areas of WR - DL serialization time, clip stack generation etc.

Thoughts @mrobinson @jrmuizel @staktrace @kvark @nical ?

mrobinson:
This is part of an API simplification which should hopefully lead to
less code complexity in the future.

Fixes #1742.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2512)

### DIFF
--- a/direct-composition/src/main_windows.rs
+++ b/direct-composition/src/main_windows.rs
@@ -131,15 +131,18 @@ impl Rectangle {
         let mut builder = api::DisplayListBuilder::new(pipeline_id, layout_size);
 
         let rect = euclid::TypedRect::new(euclid::TypedPoint2D::zero(), layout_size);
-        builder.push_rect(
-            &api::PrimitiveInfo::with_clip(
-                rect,
-                api::LocalClip::RoundedRect(rect, api::ComplexClipRegion::new(
-                    rect, api::BorderRadius::uniform(20.), api::ClipMode::Clip,
-                ))
-            ),
-            self.color,
+
+        let region = api::ComplexClipRegion::new(
+            rect,
+            api::BorderRadius::uniform(20.),
+            api::ClipMode::Clip
         );
+        let clip_id = builder.define_clip(rect, vec![region], None);
+        builder.push_clip_id(clip_id);
+
+        builder.push_rect(&api::PrimitiveInfo::new(rect), self.color);
+
+        builder.pop_clip_id();
 
         let mut transaction = api::Transaction::new();
         transaction.set_display_list(
--- a/webrender/examples/alpha_perf.rs
+++ b/webrender/examples/alpha_perf.rs
@@ -29,10 +29,7 @@ impl Example for App {
         _document_id: DocumentId,
     ) {
         let bounds = (0, 0).to(1920, 1080);
-        let info = LayoutPrimitiveInfo {
-            local_clip: LocalClip::Rect(bounds),
-            .. LayoutPrimitiveInfo::new(bounds)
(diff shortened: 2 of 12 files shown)


mrobinson:
Gecko try job is here: https://github.com/servo/webrender/pull/2512

staktrace:
Presumably you meant https://treeherder.mozilla.org/#/jobs?repo=try&revision=ad251afbee824f371097a3ea2c378faa31043be7

mrobinson:
@staktrace Thanks for provided the correct link. I was in a bit of a rush this morning.

The branch should be rebased and ready to review now.

glennw:
Is there a Gecko / Servo patch you could post here for the API update, that we can use while doing try runs in the interim?

mrobinson:

---

*[direct-composition/src/main_windows.rs, line 135 at r1](https://beta.reviewable.io/reviews/servo/webrender/2512#-L7W8NmDNb0E5-9aVA9B:-L7YooLX9Da4DkxSeWNX:b-iikw9w) ([raw file](https://github.com/servo/webrender/blob/1de1c0c8efda98fd3450e74c915ea392a9996eaf/direct-composition/src/main_windows.rs#L135)):*

any reason we aren't doing `push_clip_id` with this complex clip?
</blockquote></details>

It looked like the rounded rectangle here wasn't important for the example, but I can add it back. I didn't want to make too many changes here because I don't have a Windows environment ready for testing. I can spend a little time getting the rendered content back to its original state though.

---

*[wrench/src/rawtest.rs, line 606 at r1](https://beta.reviewable.io/reviews/servo/webrender/2512#-L7W9V1NmDp8NvCvB3Qb:-L7Yp22X4Kqisfc4tTAu:b-jk10zs) ([raw file](https://github.com/servo/webrender/blob/1de1c0c8efda98fd3450e74c915ea392a9996eaf/wrench/src/rawtest.rs#L606)):*

any reason this is removed?
</blockquote></details>

I added this one to explicitly test items with their own rounded rectangle local clips. Since I removed that feature, I also removed the test for it. Directly below is the same test, but using separate clipping nodes.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2512)*

mrobinson:
@glennw I don't think there are any code changes necessary for Gecko (it always uses the PrimitiveInfo constructors) and here is a branch with the changes necessary for Servo: https://github.com/mrobinson/servo/tree/changes-for-localclip-changes. I haven't tested this one, but I'm fairly certain there is no change in behavior.

kvark:
@mrobinson looks like this is ready to go?

mrobinson:
@kvark Yep. Is it okay to land it?

kvark:
@mrobinson well, the try looks good, code is approved by me, and Gecko doesn't (seemingly?) appear to need any changes. Shipit!

mrobinson:
@kvark Thanks!

@bors-servo r=kvark

glennw:
Blocked by #2522.

PR #2516, opened 2018-03-14
delete platform fonts, font instances, and cached glyphs in ResourceCache::clear_namespace

lsalzman:
In Gecko, when a tab is destroyed, we rely on clear_namespace() to tear down all fonts for that tab, as opposed to individually calling delete_font_template() for every available font from the Gecko side. However, clear_namespace() has a bug in that unlike delete_font_template() it doesn't ensure glyph_rasterizer.delete_font() is called, nor does it clear out font_instances (which corresponds to Gecko ScaledFonts). This fixes that, in addition to a related problem in delete_font_template() where cached_glyphs wasn't getting cleared out either.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2516)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -392,6 +392,8 @@ impl ResourceCache {
     pub fn delete_font_template(&mut self, font_key: FontKey) {
         self.glyph_rasterizer.delete_font(font_key);
         self.resources.font_templates.remove(&font_key);
+        self.cached_glyphs
+            .clear_fonts(|font| font.font_key == font_key);
         if let Some(ref mut r) = self.blob_image_renderer {
             r.delete_font(font_key);
         }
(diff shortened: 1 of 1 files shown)


glennw:
I think this looks fine - @nical could you check over it too?

webrender/src/resource_cache.rs
```diff
@@ -956,13 +958,19 @@ impl ResourceCache {
...
             .images
             .retain(|key, _| key.0 != namespace);
+        self.cached_images
+            .clear_keys(|request| request.key.0 == namespace);
 
+        self.resources.font_instances
+            .write()
+            .unwrap()
+            .retain(|key, _| key.0 == namespace);
+        for &key in self.resources.font_templates.keys().filter(|key| key.0 == namespace) {
```
kvark: This reads incorrect to me: first we retain "namespace"  only keys (already wrong - we should retain everything *but* "namespace" keys), and then we go through them, filtering on the same criteria again.

lsalzman: Accidental off-by-one-character bug. This should be fixed in PR #2520.

PR #2517, opened 2018-03-14
Refactor the way picture and brush primitives are stored.

glennw:
Previously, Picture was a separate primitive type, and it contained
a BrushPrimitive structure. Now there is a separate array of Picture
structures, and there is a BrushPrimitive::Picture type that refers
to a Picture by index.

This PR doesn't contain any functional changes - I think it will
be easier to get this reviewed and landed, and then work on the
functional changes as a separate patch.

The idea is to separate the concept of a Picture (a collection of
primitives, with children, optionally drawn into an intermediate
surface) from how that picture is drawn into the parent picture
(via BrushPrimitive::Picture).

What this means in the future is that a Picture can be referenced
by multiple Brush::Picture primitives and composited in a different
way. The primary use for this is for drop shadows - where we need
to draw the picture both as normal content, and also as an alpha
mask in a separate primitive, that has an offset / spread radius.

To achieve this, we'll move some fields currently on Picture (such
as filters, frame output) to the Brush::Picture primitive instance.

The benefits of such a change are:
 * We can reference a picture multiple times in the scene on a
   different primitive, which will allow us to fix outstanding
   drop-shadow bugs.
 * Since the Picture elements are now stored outside the primitive
   tree, we can start caching these across frames. This will form
   the basis of caching text-shadows, opacity filters etc.
 * We create fewer Picture instances (rather than one per filter
   stage as we do now). This helps debugging primarily.
 * Allowing a Picture to be referenced multiple times is more
   applicable to how SVG filter graphs work (when we implement those).

Since this is an intermediate patch, it introduces a bit of extra
indirection and duplication in places. I expect these to be resolved
by the follow up changes referenced above.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2517)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -19,7 +19,7 @@ use internal_types::{FastHashMap, SavedTargetIndex, SourceTexture};
 use picture::{ContentOrigin, PictureCompositeMode, PictureKind, PicturePrimitive};
 use plane_split::{BspSplitter, Polygon, Splitter};
 use prim_store::{CachedGradient, ImageSource, PrimitiveIndex, PrimitiveKind, PrimitiveMetadata, PrimitiveStore};
-use prim_store::{BrushPrimitive, BrushKind, DeferredResolve, EdgeAaSegmentMask, PrimitiveRun};
+use prim_store::{BrushPrimitive, BrushKind, DeferredResolve, EdgeAaSegmentMask, PictureIndex, PrimitiveRun};
 use render_task::{RenderTaskAddress, RenderTaskId, RenderTaskKind, RenderTaskTree};
 use renderer::{BlendMode, ImageBufferKind};
 use renderer::BLOCKS_PER_UV_RECT;
@@ -518,7 +518,8 @@ impl AlphaBatchBuilder {
                 BatchTextures::no_texture(),
             );
             let pic_metadata = &ctx.prim_store.cpu_metadata[prim_index.0];
-            let pic = &ctx.prim_store.cpu_pictures[pic_metadata.cpu_prim_index.0];
+            let brush = &ctx.prim_store.cpu_brushes[pic_metadata.cpu_prim_index.0];
+            let pic = &ctx.prim_store.pictures[brush.get_picture_index().0];
             let batch = self.batch_list.get_suitable_batch(key, &pic_metadata.screen_rect.as_ref().expect("bug").clipped);
 
             let render_task_id = pic.surface.expect("BUG: unexpected surface in splitting");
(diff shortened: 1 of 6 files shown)


glennw:
r? @kvark 

This patch is basically just moving code around.

Try run is pending, but thus far the orange items appear to be unrelated intermittents:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=b7f722d5097f33346804f215d82e3b16c4a88f51&selectedJob=167855614

kvark:
Some legitimate error here?
> GECKO(4041) | thread 'WRRenderBackend#1' panicked at 'Vector image error Unknown error', gfx/webrender/src/resource_cache.rs:854:29

kvark:
:lgtm:

---

Reviewed 6 of 6 files at r1.

---

*[webrender/src/display_list_flattener.rs, line 1110 at r1](https://reviewable.io:443/reviews/servo/webrender/2517#-L7Zm9KcBE6JwuvD9k48:-L7Zm9KeEbzLydoD_KOT:b-np8g2t) ([raw file](https://github.com/servo/webrender/blob/7243555047e1bec26cf2db90c7314c633370ca58/webrender/src/display_list_flattener.rs#L1110)):*
> ```Rust
>             );
> 
>             let src_prim = BrushPrimitive::new(
> ```

judging by how often this is repeated, we may add `BrushPrimitive::new_picture(index)`

---

*[webrender/src/prim_store.rs, line 921 at r1](https://reviewable.io:443/reviews/servo/webrender/2517#-L7ZmXJPkTeYGKzOhR7f:-L7ZmXJPkTeYGKzOhR7g:b-ib86vt) ([raw file](https://github.com/servo/webrender/blob/7243555047e1bec26cf2db90c7314c633370ca58/webrender/src/prim_store.rs#L921)):*
> ```Rust
>     pub cpu_borders: Vec<BorderPrimitiveCpu>,
> 
>     pub pictures: Vec<PicturePrimitive>,
> ```

any reason this is not adhering to the same naming convention as the other fields? e.g. `cpu_pictures`, `PicturePrimitiveCpu`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2517#-:-L7Zn-7LYJ9f05dp3CqV:bnfp4nl)*

glennw:
Thanks for the review!

The error in the try run is a known intermittent, unrelated to this PR (see https://bugzilla.mozilla.org/show_bug.cgi?id=1440054).

I added a ```new_picture``` constructor.

Previously, ```prim_store``` had ```cpu_``` and ```gpu_``` fields. This was before the GPU cache was implemented. Now having the ```cpu_``` prefix seems redundant. Actually, I mostly renamed it because it makes refactoring via compiler error easier, but it seems reasonable to keep it that way :)

kvark:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2517)*

kvark:
Thanks!
@bors-servo r+

glennw:
Blocked by #2522.

PR #2520, opened 2018-03-14
fix ResourceCache::clear_namespace handling of font instances

lsalzman:
PR #2516 accidentally introduced a small bug where the retain condition for font_instances was inverted. This fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2520)


glennw:
Blocked on https://github.com/servo/webrender/pull/2522.

PR #2521, opened 2018-03-15
Color field is no longer needed by the blur render tasks.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2521)


glennw:
r? anyone

glennw:
(This is a left over from when the box shadows were switched to be clip sources).

glennw:
[Not ready for review / merge yet]

PR #2523, opened 2018-03-15
Fix artifacts on box shadows that don't use nine-patch stretching.

### LINKED ISSUE #2467: Incorrect box-shadow rendering with large border radii and negative spread radius
http://tests.themasta.com/box-shadow-playground/#blue_red_outset_64_77_200_200_140x140_60x60_60x60_140x140_275_81_3_-46

glennw:
Fixes #2467.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2523)


glennw:
Try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=defcfa9b14cb578332570a45e4d4653a021695c6&selectedJob=168128618

There's two new failures here:
```box-decoration-break-with-inset-box-shadow-1.html``` - fuzzy test (28, 5270) <= (23, 14016) <= (28, 5270)``` - the max difference is lower, but far more pixels are different.

```boxshadow-large-border-radius.html``` - fuzzy test (1, 238) <= (5, 1440) <= (1, 238) - both the max difference and pixel count have increased.

Visually, I can't see any differences between the two images. I'm tempted to get this merged, since it fixes an obvious artifact (#2467). But we can wait for an opinion from @staktrace - the PR queue is backed up, so I could investigate this further tomorrow if you'd prefer.

r? @kvark

**Comment by Darkspirit** [comment]
mozregression --repo try --launch defcfa9b14cb578332570a45e4d4653a021695c6 --pref gfx.webrender.all:true

I just opened some new tabs, wikipedia, reloaded, close some tabs. Then all type of text disappeared.

**Comment by glennw** [comment]
That's caused by a bug in https://github.com/servo/webrender/pull/2516 - there is a fix in the queue - https://github.com/servo/webrender/pull/2520. We're having some CI merging issues today though :sob:

**Comment by Darkspirit** [comment]
The shape seems to be fine now. But it's too sharp at 30% zoom.
screenshot_20180315_053150

**Comment by glennw** [comment]
I think we can probably open that as a separate issue from #2467.

**Comment by staktrace** [comment]
I think the reftest differences in https://github.com/servo/webrender/pull/2523#issuecomment-373254992 are acceptable, so feel free to merge.

File: webrender/res/cs_clip_box_shadow.glsl [hunk 1]
```diff
@@ -60,11 +77,22 @@ void main(void) {
 #ifdef WR_FRAGMENT_SHADER
 void main(void) {
     vec2 local_pos = vPos.xy / vPos.z;
+    vec2 uv;
+
+    switch (vStretchMode) {
+        case MODE_STRETCH: {
+            uv = clamp(vUv.xy, vec2(0.0), vEdge.xy);
+            uv += max(vec2(0.0), vUv.xy - vEdge.zw);
+            uv = mix(vUvBounds_NoClamp.xy, vUvBounds_NoClamp.zw, uv);
```
kvark: nit: could go straight to `vUvBounds` and move the follow up clamping into the `MODE_SIMPLE` case

webrender/src/clip.rs
```diff
@@ -181,8 +181,13 @@ impl ClipSource {
         // primitive shadow rect, just blur the entire rect and draw that
         // as a simple blit. This is necessary for correctness, since the
         // blur of one corner may affect the blur in another corner.
-        minimal_shadow_rect.size.width = minimal_shadow_rect.size.width.min(shadow_rect.size.width);
-        minimal_shadow_rect.size.height = minimal_shadow_rect.size.height.min(shadow_rect.size.height);
+        let mut stretch_mode = BoxShadowStretchMode::Stretch;
+        if shadow_rect.size.width < minimal_shadow_rect.size.width ||
+           shadow_rect.size.height < minimal_shadow_rect.size.height {
+            minimal_shadow_rect.size.width = shadow_rect.size.width;
```
kvark: why not just `minimal_shadow_rect.size = shadow_rect.size`?

glennw:
Fixed the second comment. Could you expand on the first comment, I'm not quite sure what you mean?

glennw:
Oh wait, I see what you're saying - will fix now.

webrender/res/cs_clip_box_shadow.glsl
glennw: It's not quite clear to me what you mean here, since that statement is a ```mix```. Could you expand on the above?

kvark: on a second thought, ignore that :)

kvark:
please r=me once CI is happy

glennw:
I think the CI issue is remnants of the servo-tidy issue, hopefully it'll pass now.

@bors-servo r=kvark

PR #2525, opened 2018-03-15
Attempt to configure servo-tidy in a way that doesn't mess everything up.

nical:
Turn off alphabetical checks (\o/) and  whitelist duplicated packages. As far as I can see they come from dev-dependencies (mostly glutin). Ideally we would not have too many duplicated deps in our dev dependencies because it adds to the build times when building wrench and the examples, but we can live with them for now and they should not interfere with gecko.

updating glutin remvoes a lot of dupes, but also requires an update of winapi in a lot of packages which I suspect is going to be a pain. 

It would be nice if servo-tidy could be set to ignore dev-dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2525)

### DIFF
--- a/servo-tidy.toml
+++ b/servo-tidy.toml
@@ -1,10 +1,11 @@
 [configs]
 skip-check-length = false
 skip-check-licenses = false
+check-alphabetical-order = false
 
 [ignore]
 # Ignored packages with duplicated versions
-packages = []
+packages = ["gdi32-sys", "winapi", "gl_generator", "log", "core-foundation", "core-foundation-sys", "yaml-rust", "user32-sys", "lazy_static", "core-graphics"]
 
 # Files that are ignored for all tidy and lint checks.
 files = [ ]
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ before_install:
   - python --version
   - pip install mako voluptuous PyYAML servo-tidy
 script:
-#  - servo-tidy
+  - servo-tidy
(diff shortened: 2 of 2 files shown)


servo-tidy.toml
```diff
@@ -1,10 +1,11 @@
 [configs]
 skip-check-length = false
 skip-check-licenses = false
+check-alphabetical-order = false
 
 [ignore]
 # Ignored packages with duplicated versions
-packages = []
+packages = ["gdi32-sys", "winapi", "gl_generator", "log", "core-foundation", "core-foundation-sys", "yaml-rust", "user32-sys", "lazy_static", "core-graphics"]
```
Eijebong: Would it be possible to put them one by line like in servo so it's easier to read ? :)

staktrace:
Do we also want to remove webrender/src from the list of ignored folders? Are there a lot of issues coming from that code?

staktrace:
Looks like this is also the part where we (1) disable running servo-tidy in .taskcluster.yml, (2) update servo-tidy on the taskcluster images/machines, and (3) re-enable running servo-tidy in .taskcluster.yml as part of this PR.

nical:
> Would it be possible to put them one by line like in servo so it's easier to read ? :)

> Do we also want to remove webrender/src from the list of ignored folders? Are there a lot of issues coming from that code?

Sure, one thing at a time, though. Let's first get something to work at all, and then we can make it nice and all.

staktrace:
This should be closed in favor of #2527.

PR #2526, opened 2018-03-15
Disable servo-tidy on taskcluster CI as well

staktrace:
In order to get #2525 passing taskcluster CI we need to update the version of servo-tidy on the taskcluster machines/images. But if we do that without #2525 we're going to run into failures. So we need to disable servo-tidy temporarily while we do the update, and then re-enable it as part of #2525.

servo-tidy is currently disabled on travis CI as well so it's not like taskcluster is "special" in that sense.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2526)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -63,7 +63,6 @@ tasks:
           curl https://sh.rustup.rs -sSf | sh -s -- -y &&
           git clone {{event.head.repo.url}} webrender && cd webrender &&
           git checkout {{event.head.sha}} &&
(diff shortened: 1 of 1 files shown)


staktrace:
Also as a heads up, any PRs inflight that are based on a revision prior to this one will start failing taskcluster CI when the servo-tidy version is updated. So it's best to *not* bump the priority on this PR and allow as many inflight things to merge as possible before landing this.

PR #2527, opened 2018-03-15
Re-enable servo-tidy with a passing configuration

staktrace:
This takes @nical's patch from servo/webrender#2525, fixes an additional
lint error in wrench, and updates the .taskcluster.yml to re-enable the
updated servo-tidy.

r? @nical

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2527)


staktrace:
I think the TC linux job failed because it didn't pick up the new docker image. I might have pushed this PR too quickly for the change to propagate fully.

staktrace:
Re-push looks good.

Issue #741, opened 2017-01-18
Run the es3 code path on travis / CI.


glennw:

jrmuizel:
It looks OSMesa doesn't support ES so this might be a bit more tricky.

jrmuizel:
This might be sufficient to make it work:
```patch
diff --git a/include/GL/osmesa.h b/include/GL/osmesa.h
index 39cd54e..2acfe57 100644
--- a/include/GL/osmesa.h
+++ b/include/GL/osmesa.h
@@ -101,16 +101,17 @@ extern "C" {
 #define OSMESA_DEPTH_BITS            0x30
 #define OSMESA_STENCIL_BITS          0x31
 #define OSMESA_ACCUM_BITS            0x32
 #define OSMESA_PROFILE               0x33
 #define OSMESA_CORE_PROFILE          0x34
 #define OSMESA_COMPAT_PROFILE        0x35
 #define OSMESA_CONTEXT_MAJOR_VERSION 0x36
 #define OSMESA_CONTEXT_MINOR_VERSION 0x37
+#define OSMESA_ES2_PROFILE           0x38
 
 
 typedef struct osmesa_context *OSMesaContext;
 
 
 /*
  * Create an Off-Screen Mesa rendering context.  The only attribute needed is
  * an RGBA vs Color-Index mode flag.
diff --git a/src/mesa/drivers/osmesa/osmesa.c b/src/mesa/drivers/osmesa/osmesa.c
index 356ba71..4487bda 100644
--- a/src/mesa/drivers/osmesa/osmesa.c
+++ b/src/mesa/drivers/osmesa/osmesa.c
@@ -713,16 +713,18 @@ OSMesaCreateContextAttribs(const int *attribList, OSMesaContext sharelist)
             return NULL;
          break;
       case OSMESA_PROFILE:
          profile = attribList[i+1];
          if (profile == OSMESA_COMPAT_PROFILE)
             api_profile = API_OPENGL_COMPAT;
          else if (profile == OSMESA_CORE_PROFILE)
             api_profile = API_OPENGL_CORE;
+         else if (profile == OSMESA_ES2_PROFILE)
+            api_profile = API_OPENGLES2;
          else
             return NULL;
          break;
       case OSMESA_CONTEXT_MAJOR_VERSION:
          version_major = attribList[i+1];
          if (version_major < 1)
             return NULL;
          break;
```

jrmuizel:
Here's a patch that actually works:

```patch
[...]

PR #2529, opened 2018-03-15
Add ANGLE rendering in Wrench

SimonSapin:
The `--angle` command-line parameter enables it on Windows, when `--headless` is not also used. It panics on non-Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2529)

### DIFF
--- a/wrench/src/angle.rs
+++ b/wrench/src/angle.rs
@@ -1,0 +1,72 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use glutin;
+use glutin::{WindowBuilder, ContextBuilder, EventsLoop, Window, CreationError};
+
+#[cfg(not(windows))]
+pub enum Context {}
+
+#[cfg(windows)]
+pub use ::egl::Context;
+
+impl Context {
+    #[cfg(not(windows))]
+    pub fn with_window(
+        _: WindowBuilder,
+        _: ContextBuilder,
+        _: &EventsLoop,
+    ) -> Result<(Window, Self), CreationError> {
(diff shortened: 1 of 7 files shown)


wrench/src/angle.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[cfg(windows)]
+    pub fn with_window(
+        window_builder: WindowBuilder,
+        context_builder: ContextBuilder,
+        events_loop: &EventsLoop,
+    ) -> Result<(Window, Self), CreationError> {
+        use glutin::os::windows::WindowExt;
+
+        // FIXME: &context_builder.pf_reqs  https://github.com/tomaka/glutin/pull/1002
+        let pf_reqs = &glutin::PixelFormatRequirements::default();
```
SimonSapin: At the moment Wrench does not call any `ContextBuilder` method that touches the pixel format requirements, so using `default()` doesn’t make a difference.

glennw:
This should cover https://github.com/servo/webrender/issues/741.

wrench/src/egl.rs
```diff
@@ -0,0 +1,614 @@
...
+    let mut num_configs = mem::uninitialized();
+    if egl::ChooseConfig(display, descriptor.as_ptr(), &mut config_id, 1, &mut num_configs) == 0 {
+        return Err(CreationError::OsError(format!("eglChooseConfig failed")));
+    }
+    if num_configs == 0 {
+        return Err(CreationError::NoAvailablePixelFormat);
+    }
+
+    // analyzing each config
+    macro_rules! attrib {
```
nical: nit: could you make this a simple function instead of a macro?

SimonSapin: This is code copy/pasted from Glutin. I’d rather limit changes made to it to removing stuff, in order to make things easier in case we want to backport future changes from glutin.

nical: Fair enough.

SimonSapin:
@bors-servo r=nical

Tidy

PR #2530, opened 2018-03-15
Fix brush_blend on SandyBridge iGPU

### LINKED ISSUE #2088: ps_blend.glsl could be using a mat3 instead of a mat4
All the matrices generated in [ps_blend.glsl](https://github.com/servo/webrender/blob/master/webrender/res/ps_blend.glsl) only make use of the top left 3x3 square of the matrix. I think this matrix could use the `mat3` type instead, and the fragment shader could be doing `oFragColor = vec4(vColorMat * Cs.rgb, Cs.a)`, which should be a bit cheaper.

kvark:
Dividing by zero is not a good idea in GLSL.
Also fixes #2088 since we are at it.
r? @jrmuizel

Edit: wiki entry - https://github.com/servo/webrender/wiki/Driver-issues#division-by-zero-on-sandybridge-in-the-fragment-shader

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2530)

### DIFF
--- a/webrender/res/brush_blend.glsl
+++ b/webrender/res/brush_blend.glsl
@@ -11,8 +11,8 @@ varying vec3 vUv;
 
 flat varying float vAmount;
 flat varying int vOp;
-flat varying mat4 vColorMat;
-flat varying vec4 vColorOffset;
+flat varying mat3 vColorMat;
+flat varying vec3 vColorOffset;
 flat varying vec4 vUvClipBounds;
 
 #ifdef WR_VERTEX_SHADER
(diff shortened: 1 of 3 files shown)


kvark:
hold on, I missed one thing
@bors-servo r-

kvark:
(was missing the filter functions changed from `vec4` to `vec3`)
@bors-servo r=jrmuizel

kvark:
Upon checking with @mstange , it appears that having the color matrix affecting alpha is only needed for SVG color filters, and we don't quite support it in WR anyway, and we'll need way more logic to support SVG filters (not covered by brush_blend). Thus, we figured the test case that modifies the alpha via color matrix needs to be removed.
@bors-servo r=jrmuizel,mstange

jrmuizel:
This still needs the test case removed.

kvark:
Should be good now.
@bors-servo r=jrmuizel,mstange

PR #2539, opened 2018-03-18
Update base64 and image deps.

### LINKED ISSUE #2538: Update dependency versions of `base64` and `image` to be inline with servo.
It looks like the dependencies for the following should be updated to match servo's versions of the dependencies.

https://github.com/servo/servo/pull/20328

* base64 `0.3.1` -> `0.6.0`
* image `0.17.0` -> `0.18.0`

glennw:
Fixes #2538.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2539)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,10 +50,11 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.3.0"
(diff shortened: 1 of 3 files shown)


glennw:
r? anyone

PR #2540, opened 2018-03-19
Draw text shadows in screen space.

glennw:
With box shadows drawn via a clip source, the only picture type
that was drawing in local space was text shadows. Switching them
to draw in screen-space simplifies a lot of code, and also gives
a better quality result.

Specifically:
 * Remove ContentOrigin enum.
 * Remove cs_text_run shader (use standard text shader).
 * Remove cache text run hash maps.
 * Remove local-space raster mode from brush primitive.
 * Don't fetch color from picture for text runs.
 * Allow primitive culling on all Picture types.
 * Optimize allocated region for text-shadows.

There's one performance regression with this change - text shadows
which have large rotations will blur more pixels than previously.
This can be fixed in a follow up, by drawing the blur as a normal
primitive instead of a 2D rect. This will also optimize other blurs,
such as CSS filters.

This is preparation work for caching text shadows, and also some
other optimizations (such as drawing text decorations as clip
sources).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2540)


glennw:
This is not quite ready for merge yet - it can be reviewed if you have time, or can wait until the remaining work below is complete:

- [ ] One failure in try run to fix (https://treeherder.mozilla.org/#/jobs?repo=try&revision=fa2a7084b1a32acb6edaff20eb8f19531d007022).
- [ ] Update wrench reftest images once the try failure above is fixed.
- [ ] Add a couple of wrench reftests with rotated and perspective text-shadow elements.

r? @kvark

kvark:
Holy crab, this is indeed simplifying quite a bit of code!
I do worry though, why did you decide to spin 180 w.r.t. local-space render tasks? Didn't you put a ton of effort in a series of PRs during the past 2-3 months to support those properly? My understanding was that we want to have an ability to do this trade-off, based on how frequent are the local versus global space changes.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2540#-:-L7zefbbBqMdjTH2f0fM:b3q2tsc)*

glennw:
I fixed the failure in R1 (disabling the local clip rect inside the text-shadow picture context).

Latest try is here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2c46f56f830590a4e87bdc608138dcaee883333b&selectedJob=169088199

There's a couple of intermittent failures I haven't seen before, but I don't think any of them are related to this change. It'd be good to get @staktrace to confirm though. The Windows tests are still pending.

Apart from that, this should be ready for review / merge now, r? @kvark

glennw:
I re-triggered each of the orange results, and they passed - so they do appear to be intermittent failures. I'm fairly confident they are all unrelated to this PR. The Windows tests also look green. :tada:

staktrace:
Try push looks good to me!

webrender/src/display_list_flattener.rs
```diff
@@ -1552,13 +1559,13 @@ impl<'a> DisplayListFlattener<'a> {
             self.shadow_prim_stack[idx].1.push((prim_index, clip_and_scroll));
         }
 
-        let prim_index = self.create_primitive(
```
mrobinson: Hrm. Out of curiosity, does this change allow for making this optimization or is it an unrelated change?

glennw: This change allows it - previously we would reference the same primitive in the slow shadow path below. Now we build a separate primitive there. There's a slight memory overhead to this for slow text shadows, which we might be able to improve on in the future, but it should be minor.

PR #2543, opened 2018-03-19
evict items from GlyphCache when evicted from TextureCache

lsalzman:
This is a follow-up to #2516 to make the GlyphCache itself much leaner to further reduce the amount of memory leaking/bloat we're seeing in Gecko by limiting the size of the GlyphCache.

Per discussion with Glenn, since the TextureCache is itself limited in size, and GlyphCache entries are ultimately upload to it, it makes sense to just use that to effectively limit the GlyphCache by purging its entries when they are removed from the TextureCache. This has many nice downwind effects in that GlyphCache entries themselves are now much smaller.

To link them together, I needed to introduce the notion of an EvictionNotice which the TextureCache uses to signal that an entry has been evicted, so that the GlyphKeyCaches can know when to actually check for purge entries, and avoids the expense of regularly having to scan through all glyphs.

I also optimized the amount of FontInstances we need to copy around to plumb the GlyphCache, which will further reduce memory thrashing.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2543)

### DIFF
--- a/webrender/src/glyph_cache.rs
+++ b/webrender/src/glyph_cache.rs
@@ -2,38 +2,43 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{DevicePoint, DeviceUintSize, GlyphKey};
+use api::GlyphKey;
 use glyph_rasterizer::{FontInstance, GlyphFormat};
 use internal_types::FastHashMap;
 use resource_cache::ResourceClassCache;
-use std::sync::Arc;
-use texture_cache::TextureCacheHandle;
+use texture_cache::{TextureCache, TextureCacheHandle, EvictionNotice};
 
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
-pub struct GenericCachedGlyphInfo<D> {
+pub struct CachedGlyphInfo {
     pub texture_cache_handle: TextureCacheHandle,
-    pub glyph_bytes: D,
-    pub size: DeviceUintSize,
-    pub offset: DevicePoint,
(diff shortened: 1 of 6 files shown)


glennw:
Reviewed 5 of 5 files at r1.

---

*[webrender/src/texture_cache.rs, line 201 at r1](https://reviewable.io:443/reviews/servo/webrender/2543#-L80O4LI0RV55HY8vLZK:-L80O4LI0RV55HY8vLZL:b-k8s5e8) ([raw file](https://github.com/servo/webrender/blob/6418f2340b6714f9beecb479acd0ba06b65416bd/webrender/src/texture_cache.rs#L201)):*
> ```Rust
> 
> // Only checked and updated on the main thread.
> unsafe impl Send for EvictionNotice {}
> ```

I think this is actually safe with the current way that the async scene building works, but perhaps not if frame building is moved to that thread? We discussed on IRC we could use an Arc/Atomic to be on the safe side. Others may have a better idea for a simpler solution, so we can check with them too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2543)*

glennw:
It's great to not have to serialize all the bytes of the cached glyph entries for capturing, that simplifies things quite a lot!

Given that this also touches things in the capture and resource cache code, it'd be good for @nical and @kvark to take a look over this too.

glennw:
https://github.com/glennw/webrender/commit/66b9fb8bc27ddaf51f322eb63b3e57983bce3859

This commit moves the creation of the resource and texture cache inside the render backend thread. This means rustc knows that the ```EvictionNotice``` type is never sent across threads, so we can remove the ```unsafe impl```.

glennw:
@bors-servo r+

Merging now, since it's quite a large change. It'd be great if @nical and @kvark could take a look over this though, and we'll fix any comments as a follow up.

PR #2548, opened 2018-03-20
Remove complex radial gradient API

### LINKED ISSUE #2439: [RFC] Remove "complex" radial gradient support?
In ```display_list.rs``` there are two methods for creating radial gradients.

```create_radial_gradient``` takes a center point, an (elliptical) radii, and a series of stops.

```create_complex_radial_gradient``` takes several more parameters (a start and end center, start and end radii, ratio_xy, stops etc).

As far as I can see, neither Gecko nor Servo ever use the complex radial gradient API. And from reading the CSS specification, I can't see the usefulness of the complex radial gradient API - it seems like the simple API should be able to express all radial gradients that CSS is able to.

So, I'd like to remove the complex radial gradient API - it will allow some quite significant optimizations to the radial gradient shader (and gradients in general).

Does this sound reasonable? Is anyone aware of a use for the complex radial gradient API? Is the Gecko + WR radial gradient integration complete?

mrobinson:
Instead of allowing complex radial gradients, denormalize the stops
array when serializing display lists to YAML. We also store start and
end offset instead of start and end radius. This allows us to
denormalize the stops array in a more straight-forward way. This opens
up more opportunities for optimizing the radial gradient shader.

Fixes #2439.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2548)


mrobinson:
The Gecko try job for this change is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9f9f6111454442c4d0a381ad6d8d508833f6e725

PR #2550, opened 2018-03-20
Add a callback for RB thread creation.

staktrace:
We'll want this for APZ, because APZ will need to know which thread is
the render backend thread corresponding to a particular window.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2550)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1487,6 +1487,7 @@ impl Renderer {
         let blob_image_renderer = options.blob_image_renderer.take();
         let thread_listener_for_render_backend = thread_listener.clone();
         let thread_listener_for_scene_builder = thread_listener.clone();
+        let renderer_id_for_render_backend = options.renderer_id.clone();
         let rb_thread_name = format!("WRRenderBackend#{}", options.renderer_id.unwrap_or(0));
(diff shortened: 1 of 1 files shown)


webrender/src/renderer.rs
```diff
@@ -3828,6 +3830,7 @@ pub trait OutputImageHandler {
 pub trait ThreadListener {
     fn thread_started(&self, thread_name: &str);
     fn thread_stopped(&self, thread_name: &str);
+    fn new_render_backend_thread(&self, renderer_id: Option<u64>);
```
nical: How are you going to store the state related to the APZ logic on the render backend? wouldn't it make sense to pass an object in the renderer options with an init method that runs on the render backend before starting the loop and that is called by the render backend at the right times?

staktrace: I don't think we'll actually need to store any state in the render backend. We'll just need this call plus one other at the time of the scene swap. I'm doing this incrementally so I just wanted to land this for now as I'll need it for bug 1447299, and then when I need the other callback we can introduce a new object at that point if it makes sense.

glennw:
@bors-servo r+

(assuming that the comment above addresses @nical 's question).

PR #2552, opened 2018-03-21
Remove brush_line, convert line decorations to a clip source.

glennw:
This removes another shader, by making line decorations a clip
mask source, similar to how box-shadows are drawn.

The line decorations are now drawn as a brush_solid primitive,
with the decoration drawn into the clip mask, as appropriate.

For solid line decorations, no clip mask is required. This helps
batching significantly on sites that use a lot of underlines and
links, such as nytimes.com. Additionally, solid line decorations
that are opaque are drawn in the opaque pass now, which provides
more z-buffer gains and fewer transparent pixels to draw.

Previously, we used the primitive index as the z-buffer value.
However, this is not actually correct, since in text shadows the
ordering of primitives to be drawn does not always match that of
the primitive indices. This was never an issue previously, as
text decorations were never included in the opaque render pass.
To handle this, introduce ZBufferId type which is an incrementing
counter that is generated as each primitive it added to a batch.

Finally, remove the color from the Picture render task data since
it is no longer required. This also makes it possible to reduce
the size of the RenderTaskData struct - this is an optimization
that means vertex shaders fetch one less texel for each picture
task.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2552)


glennw:
This is actually a relative small PR, but it looks big because it includes https://github.com/servo/webrender/pull/2540 which is pending review / merge.

Pending try here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=45f001dbd9201a3915f40397a35ca11832662280

r? @mrobinson or @kvark 

Feel free to leave this until #2540 lands, if that will make it easier to review this one.

glennw:
Try looks good, just one unrelated intermittent failure.

PR #2554, opened 2018-03-21
Clear the texture cache (and resource cache) before running each reftest in wrench.

glennw:
The OSMesa rasterizer can give slightly different results for tests,
depending on where an image is located in the texture cache. This
attempts to mitigate that, by ensuring the texture cache is empty
at the start of each test.

This is an attempt to prevent the addition of a new reftest causing
unrelated test failures in later tests during a reftest run.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2554)


glennw:
r? anyone

jrmuizel:
@bors-servo r+

This seems like it could be useful as debugging tool too.

glennw:
Updated the reftest annotations for the previous PR that landed before this one.

@bors-servo r=jrmuizel

Issue #2519, opened 2018-03-14
Tab switching to Julia Evans' blog is slow


mstange:
Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1444767

Here's a profile of a tab switch to https://jvns.ca/blog/2018/01/26/mac-memory-maps/ : https://perfht.ml/2twMA6Q

There's a composite that takes 132ms.
 - 33ms glBufferData
 - 35ms glTexSubImage3D
 - 37ms glGenerateMipmap
 - 12ms memmove inside glDrawElements

jrmuizel:
So for the glGenerateMipmap, I'm wondering if we actually need to do this. High quality downscaling should actually be being handled on the Gecko side. @glennw was this not happening for some reason?

glennw:
The mip map generation code was added for this bug (https://github.com/servo/webrender/issues/2023) and the bugzilla bug referenced in that. We can definitely consider removing it or adding a flag on each image to give Gecko explicit control over whether to generate mips, if that would help?

mstange:
Here's another profile of slow `glGenerateMipmap`: https://perfht.ml/2Gv42LJ
Particularly interesting is the part where `_gleMakeMipmapDataNPOT` is doing the downscaling on the CPU.

I encountered this one after during the animation that plays after clicking the "Begin expedition" button on https://developer.microsoft.com/en-us/microsoft-edge/testdrive/demos/variable-fonts/ .

glennw:
I'm guessing from the name of that function your GPU / driver falls back to CPU mipmap generation if the size is non power of two?

mstange:
Yes, that's what it looks like.

The image that's causing the slowness is the fallback item for a border image: the diagonal stripes on the side which cover the whole length of the page. Is this blob image rasterized in its entirety, or in tiles? Do we mipmap the entire thing or do we mipmap the tiles individually?

mstange:
This is fixed now that mipmaps are disabled by default.

PR #2555, opened 2018-03-22
Make automatic mipmap generation opt-in.

glennw:
Gecko downscales images before passing them to WR, so typically
doesn't want automatic mipmap generation enabled, which can be
slow on some platforms / drivers.

Mipmaps can still be enabled on a per-image basis, by setting the
allow_mipmaps field in the ImageDescriptor.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2555)

### DIFF
--- a/webrender/examples/basic.rs
+++ b/webrender/examples/basic.rs
@@ -202,7 +202,7 @@ impl Example for App {
         let image_mask_key = api.generate_image_key();
         resources.add_image(
             image_mask_key,
-            ImageDescriptor::new(2, 2, ImageFormat::R8, true),
+            ImageDescriptor::new(2, 2, ImageFormat::R8, true, false),
             ImageData::new(vec![0, 80, 180, 255]),
             None,
         );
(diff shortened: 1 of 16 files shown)


glennw:
r? @jrmuizel 

This should help with https://github.com/servo/webrender/issues/2519.

glennw:
@staktrace Heads up that this adds a new ```bool``` to the ```ImageDescriptor``` public struct, which will affect Gecko bindings.

staktrace:
@glennw Thanks for the heads-up. It looks like I need to set allow_mipmaps to true on the Gecko side to preserve current behaviour in the WR update, and then we can make additional changes on the Gecko side if we want to switch it to false for certain cases or all the time.

staktrace:
@glennw It looks like turning off mipmaps also reduced memory usage quite a bit, see https://bugzilla.mozilla.org/show_bug.cgi?id=1448006#c6 - if that's not expected it might indicate a leak in the mipmap code.

glennw:
It would be expected to reduce (GPU-side) texture memory by ~33% in cases where mipmaps are being enabled. I'm not sure if that would be counted by the memory reporting system though. I guess it's possible that the driver was maintaining the mipmaps internally in CPU memory too.

PR #2556, opened 2018-03-22
Unify text-shadows with the blur filter.

glennw:
This removes any special code for text shadows. A text shadow is
now simply a Picture with text runs and/or line decorations that
has a blur filter applied to it.

This is possible due to the recent work related to removing the line
decoration and cached text run primitives.

This greatly simplifies the shadow code, and a lot of the functionality
of the text shadows (e.g. fast path shadows) just fall out of the
existing Picture functionality for blur filters.

Next step is to expand the primitive types that can run through the
shadow paths, which will then allow us to unify the drop-shadow
filter to be a simple shadow + blur filter.

NOTE: The PictureKind enum now has only a single kind (Image). I could
      have removed that in this PR, and added the fields to Picture
      directly. However, that will result in a lot of indentation
      changes. So I'll do that as a follow up PR to make this one
      easier to review.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2556)


glennw:
This PR also includes #2552. It may be easier to review this PR after that lands.

The try run has several orange items, but I think they are all OK:

* Wr5 is 2x new PASS.
* R3 is 1x new PASS.
* R4 is 2x new PASS.
* rust item looks to be an unrelated build failure.
* mda2 looks like an unrelated intermittent.
* wpt2 looks like an unrelated intermittent.

@staktrace, does that look correct to you?

r? @mrobinson or @kvark

glennw:
With try run link:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=2b31e1906983c87cc73b8b9496d1de0036650eb9

staktrace:
@glennw I agree with your try push assessment, I think this is fine to land.

staktrace:
Surprise! The unexpected-pass results are all actually from turning off mipmaps, not from this PR. https://treeherder.mozilla.org/#/jobs?repo=try&revision=b114b7b9e9565b2f03470931c6574040e354aa10

glennw:
@staktrace I thought that might actually be the case here :)

glennw:
I've rebased this on top of master, now down to one commit with +269 / -410 lines :)

wrench/reftests/text/shadow-clip.yaml
```diff
@@ -12,7 +12,7 @@ root:
         -
           type: "shadow"
           bounds: [0, 0, 200, 200]
-          blur-radius: 1
```
Gankro: why was this necessary?

glennw:

---

*[webrender/src/display_list_flattener.rs, line 941 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YIXFykHqFgWgRx7Ct:-L8ZA99H0NFBLhBVjTPS:bbrp1og) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/display_list_flattener.rs#L941)):*

I don't recall us handling local clip sources around here before, why now?
</blockquote></details>

Because we now draw text decorations as a clip source - this has only landed recently.

---

*[webrender/src/display_list_flattener.rs, line 957 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YJVvLsdspGX65os82:-L8ZAX4XneqlSu1olbsQ:bcs04fq) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/display_list_flattener.rs#L957)):*

Hmm, is it actually possible for a primitive to be in a non-visible container while the shadow isn't? (is a shadow ever in a different container from its contents?)

PrimitiveContainers are a new notion to me so idk what they represent (kvark says they're the new Picture, but I still see Pictures are alive and well?)

Is it accurate to say they're just temporary metadata used during flattening? so e.g. the fact that we do tons of non-recycled calls to create_shadow is "fine"? If so I'd prefer to call create_shadow something that indicates it's uh, softer, than it sounds. with_shadow?
</blockquote></details>

A PrimitiveContainer is just an enum used to handle pushing all of the primitive types. In this case, the is_visible just checks if alpha > 0, which is what we used to do manually in the add_text and add_line functions.

---

*[webrender/src/display_list_flattener.rs, line 1419 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YM797-BIOQghBahgN:-L8ZAkqq4I3KOaKTFzuq:b-896fix) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/display_list_flattener.rs#L1419)):*

do -> to
</blockquote></details>

Done.

---

*[webrender/src/display_list_flattener.rs, line 1449 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YLaWNO4vyY6vmsa-K:-L8ZApOTqlQiEjlT5Mlt:b-l9u3m3) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/display_list_flattener.rs#L1449)):*

😭
</blockquote></details>

Is that a good reaction, or an issue?

---

*[webrender/src/picture.rs, line 193 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YZgpLy_LlP4XZaMMb:-L8ZAx9L2tBO23emmP30:b-vvdvja) ([raw file](https://github.com/servo/webrender/blob/c32585c1ed94d6cad938ecd0ef7db566a367a00d/webrender/src/picture.rs#L193)):*

Why is FilterOp::Blur handling deferred to prepare_prim_for_render, but not e.g. DropShadow?
</blockquote></details>

Because the DropShadow code is much broken. I'm working on fixing the drop shadow code as a follow up patch.

---

*[webrender/src/tiling.rs, line 409 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YFlkD7sigICPCkaFO:-L8ZB6O1rirRpjjnZwWQ:bb1h3wy) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/tiling.rs#L409)):*

Why isn't this just `if let PictureKind::Image { frame_output_pipeline_id: Some(pipeline_id), .. } = pic.kind { self.outputs.push(...) }`?
</blockquote></details>

I got a rustc compile error, since there's just the one PictureKind now (I'm intending to remove that enum as a follow up, but left it here for now to avoid a heap of indentation changes).

---

*[wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png, line  at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YEmvB5vOyR1-fGHm8:-L8ZBKFXkYyWINnYB9vK:b6s3akd) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png#L)):*

All the content seems to be shifting down a ~pixel here. That seems suspicious?
</blockquote></details>

I didn't look into that diff too closely, since the try run passed. I can try to find the exact cause though, if you like?

---

*[wrench/reftests/text/shadow-clip.yaml, line 15 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YEfKoelzV7IIvjiQj:-L8ZBWO5rdcOR0v5UpVK:blp49ur) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/wrench/reftests/text/shadow-clip.yaml#L15)):*

why was this necessary?
</blockquote></details>

A blur radius of 1 was producing a fractional offset in the primitive, due to the halving of the blur radius, and then scaling by 3 for the blur range. This was producing pixel perfect results on all the hardware I tested, but OSMesa seemed to be rounding it incorrectly, causing the test to fail. Since it only affected OSMesa, and the blur radius being 2 doesn't affect what this test is checking, I used a whole number to work around that for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2556)*

Gankro:
Review status: 11 of 12 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/src/display_list_flattener.rs, line 1449 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YLaWNO4vyY6vmsa-K:-L8ZZ78AMBB4VScvvzBH:b99sa82) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/display_list_flattener.rs#L1449)):*

Is that a good reaction, or an issue?
</blockquote></details>

tears of joy!

---

*[webrender/src/tiling.rs, line 409 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YFlkD7sigICPCkaFO:-L8ZYyhIgRB69dDHPW5B:bvxyozc) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/tiling.rs#L409)):*

I got a rustc compile error, since there's just the one PictureKind now (I'm intending to remove that enum as a follow up, but left it here for now to avoid a heap of indentation changes).
</blockquote></details>

Yes that's why I moved the `Some()` match inline, which makes the whole pattern fallible.

---

*[wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png, line  at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YEmvB5vOyR1-fGHm8:-L8ZZJEUswATXRA6krSE:bdawlrb) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png#L)):*

I didn't look into that diff too closely, since the try run passed. I can try to find the exact cause though, if you like?
</blockquote></details>

It's also possible github is just being dumb here. But wait, how did you get a passing try *and* a diff?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2556)*

glennw:
Review status: 10 of 12 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/tiling.rs, line 409 at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YFlkD7sigICPCkaFO:-L8Z_yM31ca0LxzSp-hC:b-bon5f2) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/webrender/src/tiling.rs#L409)):*

Yes that's why I moved the `Some()` match inline, which makes the whole pattern fallible.
</blockquote></details>

Ah, I misread your comment. Fixed now.

---

*[wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png, line  at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YEmvB5vOyR1-fGHm8:-L8Za6Apt3zG4eN5y8fs:bmtvbsy) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png#L)):*

It's also possible github is just being dumb here. But wait, how did you get a passing try *and* a diff?
</blockquote></details>

Oh, I meant I hadn't looked too closely at the wrench reference image difference, since the gecko try had passed. I can take a closer look at this difference before we merge though, if you prefer, to see if it's actually a bug with the previous image or the new one?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2556)*

Gankro:
Review status: 10 of 12 files reviewed at latest revision, 1 unresolved discussion.

---

*[wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png, line  at r1](https://reviewable.io:443/reviews/servo/webrender/2556#-L8YEmvB5vOyR1-fGHm8:-L8Zc2uK7oNfsys14Ubk:b-97932c) ([raw file](https://github.com/servo/webrender/blob/5b8a81ebf1794e2f4b9bad1ede0eff3182987f9d/wrench/reftests/text/blurred-shadow-local-clip-rect-ref.png#L)):*

Oh, I meant I hadn't looked too closely at the wrench reference image difference, since the gecko try had passed. I can take a closer look at this difference before we merge though, if you prefer, to see if it's actually a bug with the previous image or the new one?
</blockquote></details>

I'd appreciate it, but not a blocker if you think you have better stuff to do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/webrender/2556)*

glennw:
I'll take a look today, and see what causes the difference. r=you if it turns out to be expected? (otherwise if there's a bug I'll put up an extra commit + gecko try).

Gankro:
yep, r=me with the image investigate

glennw:
@Gankro You're quite right - that reference image was incorrectly clipped. I guess we don't have Gecko test coverage for that. Pushed a commit that restores the local rect inflate for blurs, will kick off a gecko try shortly.

glennw:
Try run here https://treeherder.mozilla.org/#/jobs?repo=try&revision=e5cabd0b6e5a2c58b75f4fd20d55a0c782166b97 has one new PASS and one new FAIL.

I'll investigate the failure today.

glennw:
Hmmm, seems to not reproduce locally so far :disappointed:

glennw:
Latest try run (not quite complete yet):

https://treeherder.mozilla.org/#/jobs?repo=try&revision=a51d30f477a9adbf31ba1e3053275b5177a0e06e

There is a new PASS in R7. The other orange results look unrelated, so I think this is ready to merge now.

glennw:
Try run looks good now.

@bors-servo r=Gankro :rocket:

PR #2559, opened 2018-03-22
Add hit testing support in wrench

mrobinson:
Process hit testing tags when writing and reading wrench YAML. Also, add
a keybinding which performs a hit test. This is useful for debugging hit
testing issues.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2559)


glennw:
Nice!

@bors-servo r+

PR #2563, opened 2018-03-23
Handle box-shadow stretch mode per-axis.

### LINKED ISSUE #2561: Tall shadow exceeds texture size limit, crashing renderer
With some help from @glennw I've found a probable cause for the crashes tracked in Firefox [bug 1442921](https://bugzilla.mozilla.org/show_bug.cgi?id=1442921).

WR tries to draw box shadows by stretching a 9-patch shadow mask across the entire shadow. For this it needs to calculate the minimum size of such a mask, so that blurring the corners doesn't affect the other patches. For this it only needs the border radii and the blur radius: https://github.com/servo/webrender/blob/277e35f7d6a7b39c59d02313aed5badbfb44e5e2/webrender/src/clip.rs#L166-L191

However, in the case that this 9-patch drawing is not applicable, WR falls back to drawing the entire shadow mask: https://github.com/servo/webrender/blob/277e35f7d6a7b39c59d02313aed5badbfb44e5e2/webrender/src/clip.rs#L205-L214 Note that if *either* `width` *or* `height` is smaller than the minimum, `Simple` mode is chosen.

The issue appears when one dimension is smaller than the minimum, but the other is large enough to exceed texture size limits. In my case I happened upon a shadow that was 400×25743 px in size, with 200 px corner radii and a 60 px blur. The latter result in a minimum size of 580×580 px.

WR will attempt to allocate an alpha texture for the entire shadow, which is clipped to the maximum 16384 px tall. Afterwards it runs into an assertion due to the size mismatch.

glennw:
This is a crash fix for box shadows that have one axis which is
very long, and one axis that is small.

It's also an optimization for box-shadows that have one long and
one short axis, since it allows drawing a minimal shadow rect in
these cases. This is handled by having a stretch mode per-axis
rather than one for the entire shadow primitive.

Fixes #2561.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2563)


glennw:
r? @mrobinson or anyone

Pending try run here (have not checked results yet):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c9bbe08556ee1d784cbeab6f345adde782cfac72

glennw:
Try run looks good. There are a couple of seemingly unrelated intermittents, but all the box-shadow reftests appear the same.

kvark:
:lgtm:

---

Reviewed 9 of 9 files at r1.

---

*[webrender/res/cs_clip_box_shadow.glsl, line 23 at r1](https://beta.reviewable.io/reviews/servo/webrender/2563#-L8_LQuD8e6moTzv0Gic:-L8_LQuD8e6moTzv0Gid:bmc0mt3) ([raw file](https://github.com/servo/webrender/blob/b0f31e803ae3e92d0ab3385c3ecdf438747fa858/webrender/res/cs_clip_box_shadow.glsl#L23)):*
> ```GLSL
>     vec2 src_rect_size;
>     float clip_mode;
>     int stretch_mode_x;
> ```

nit: could be `ivec2 stretch_mode;`

---

*[webrender/res/cs_clip_box_shadow.glsl, line 77 at r1](https://beta.reviewable.io/reviews/servo/webrender/2563#-L8_LyAR4eCOWXcqsFr0:-L8_LyAR4eCOWXcqsFr1:b-7eehsh) ([raw file](https://github.com/servo/webrender/blob/b0f31e803ae3e92d0ab3385c3ecdf438747fa858/webrender/res/cs_clip_box_shadow.glsl#L77)):*
> ```GLSL
>     }
> 
>     switch (bs_data.stretch_mode_y) {
> ```

somewhat unfortunate code duplication, can be addressed later

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2563#-:-L8_NHMH9G0M-O-IEqGY:bnfp4nl)*

PR #2564, opened 2018-03-23
Prevent segmented primitives from reading outdated clip tasks.

### LINKED ISSUE #2531: Large element with 2d transform disappears sometimes
When scrolling up and down on [this testcase](https://bug1446129.bmoattachments.org/attachment.cgi?id=8959296), the blue box sometimes disappears.

Here's a recording of a frame where the element is missing: http://tests.themasta.com/webrender-transformed-large-box-disappears-capture.zip

nical:
Fixes #2531.

Brush segments store handles to their own clip tasks. If segmentation heuristics change we need to either clear the segment information or update it to ensure the clip tasks they point to are requested (this commit does the latter).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2564)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1484,14 +1484,8 @@ impl PrimitiveStore {
         frame_state: &mut FrameBuildingState,
     ) -> bool {
         let metadata = &self.cpu_metadata[prim_index.0];
-        let brush = match metadata.prim_kind {
-            PrimitiveKind::Brush => {
-                &mut self.cpu_brushes[metadata.cpu_prim_index.0]
-            }
-            _ => {
-                return false;
-            }
-        };
+        let brush = &mut self.cpu_brushes[metadata.cpu_prim_index.0];
+        debug_assert!(metadata.prim_kind == PrimitiveKind::Brush);
 
         PrimitiveStore::write_brush_segment_description(
             brush,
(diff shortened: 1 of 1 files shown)


nical:
r? @glennw 

This is I think the simplest way to fix #2531, although I suspect that we might want to instead un-segment primitives in some case to allow the most optimized path when segmentiation heuristics change between frames, or maybe it is enough/correct to go through the segments and clear the clip task ids.
Un-segmenting means we need to differentiate between segments created automatically during frame building and segments created explicitly when flattening the scene, plus probably a bunch of other edge cases I don't know about.
Perhaps we can land the simple fix first and then look for the best solution?

nical:
Yeah, I know this isn't optimal since we hang on to a heuristic that was good for a previous frame but not the current one. I was unsure whether clearing the render task ids would be enough, since the new frame might not even create segments (in fact in this case I am pretty sure it wouldn't) and so we'd still be in a half-unexpected configuration, but if you think it will do, sounds good to me. I'll give this a go on Monday.

glennw:
The segments should always be constant for a given scene, since they rely only on clips in the same reference frame (local space) as the primitive, and therefore shouldn't change with scrolling. So I think it should be fine to just clear the task IDs in this case. Thanks!

nical:
Closing in favor of #2568.

PR #2566, opened 2018-03-23
Render glyphs with Pathfinder if the `pathfinder` feature flag is enabled.

pcwalton:
This is very preliminary. Notable problems:

* Intermediate textures are not cached.

* Font appearance resembles that of macOS regardless of platform.

* Subpixel AA is always used, even if disabled.

* The `SourceTexture` extension is hacky.

* Emoji and other color glyphs will not render properly. We should fall
back to the CPU for those.

* There is no hinting or pixel snapping yet.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2566)


glennw:
Awesome! Will review first thing Monday :)

glennw:
Windows CI failure:

```
error[E0061]: this function takes 8 parameters but 4 parameters were supplied
   --> src\glyph_rasterizer.rs:973:9
    |
624 | /     pub fn resolve_glyphs(
625 | |         &mut self,
626 | |         glyph_cache: &mut GlyphCache,
627 | |         texture_cache: &mut TextureCache,
...   |
720 | |         }
721 | |     }
    | |_____- defined here
...
973 |           &mut glyph_cache,
    |           ^^^^^^^^^^^^^^^^ expected 8 parameters
```

fschutt:
Question, would this change make freetype (possibly) obsolete in the future?

pcwalton:
No, because FreeType is still used to load the outlines and hint the
glyphs. Besides, HarfBuzz has to use it during layout, I believe.

glennw:

---

*[webrender/Cargo.toml, line 13 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V89Qb7UWRjuvRbECE:-L8V89Qb7UWRjuvRbECF:bqxpcfl) ([raw file](https://github.com/servo/webrender/blob/92c2cb48d234ae06db3771e9cddfaf8639b81e48/webrender/Cargo.toml#L13)):*
> ```TOML
> freetype-lib = ["freetype/servo-freetype-sys"]
> profiler = ["thread_profiler/thread_profiler"]
> debugger = ["ws", "serde_json", "serde", "image", "base64"]
> ```

Why is image no longer an optional dependency?

---

*[webrender/res/ps_text_run.glsl, line 177 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V9Ckz9qgKX6uvF7Og:-L8V9Ckz9qgKX6uvF7Oh:bl9kank) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/res/ps_text_run.glsl#L177)):*
> ```GLSL
>     vec3 tc = vec3(clamp(vUv.xy, vUvBorder.xy, vUvBorder.zw), vUv.z);
>     vec4 mask = texture(sColor0, tc);
>     if (vMaskSwizzle.z > 0.0)
> ```

Is this necessary? For the normal (CPU rendered) alpha mode, we store the channels correctly in the RGBA output during rasterization, so there's no runtime swizzle needed in this case.

---

*[webrender/src/device.rs, line 441 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V9uOE0Enra6MrzZyX:-L8V9uOE0Enra6MrzZyY:bcx2frb) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/src/device.rs#L441)):*
> ```Rust
> pub struct Texture {
>     id: gl::GLuint,
>     target: TextureTarget,
> ```

Why do we need to store this as TextureTarget and calculate the GL target each time it's used?

---

*[webrender/src/glyph_rasterizer.rs, line 530 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8VFPDQ9Of0VvKWrRln:-L8VFPDQ9Of0VvKWrRlo:b-7ieyax) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/src/glyph_rasterizer.rs#L530)):*
> ```Rust
>                                                                      false)) {
>                 (Ok(outline), Ok(dimensions)) => {
>                     // TODO(pcwalton)
> ```

Could you expand on this TODO?

---

*[webrender/src/texture_cache.rs, line 174 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8VATS23fieEDwgJv-H:-L8VATS23fieEDwgJv-I:b-bx4d1r) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/src/texture_cache.rs#L174)):*
> ```Rust
> // In this case, the cache handle needs to re-upload this item
> // to the texture cache (see request() below).
> #[derive(Clone, Debug)]
> ```

Texture cache handles shouldn't be cloned in normal use - if one is updated then the other won't have the correct location. Is this still relied on?

---

*[webrender/src/tiling.rs, line 977 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8VBwII9GYXl_EwRhkl:-L8VBwII9GYXl_EwRhkm:b-uz9vry) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/src/tiling.rs#L977)):*
> ```Rust
>         let instance = BlurInstance {
>             task_address: task_address,
>             src_task_address: task_address,
> ```

This doesn't seem right. The two addresses are the render task being drawn to (provides target rect etc) and the source render task we are going to blur.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2566)*

glennw:
The main bits I don't fully understand yet are:

* Why we need a new variant in the ```SourceTexture``` enum.
* The code that unconditionally creates an extra alpha/color pass for glyphs.

I *think* the second one is probably because we don't have proper support for multiple top-level render task trees yet, and the pathfinder stuff doesn't really fit into the hierarchical nature of the existing render task tree? I've been planning to support multiple roots in the task tree at some point, which I think will tidy this stuff up.

I'll take another look at it this afternoon regarding the first question above.

We'll also need to do a gecko try (instructions here https://wiki.mozilla.org/Platform/GFX/Quantum_Render#Try_syntax) and it'd be good to add an entry to the TaskCluster config file in the repo that builds the Pathfinder feature, so we at least don't regress on compilation errors there.

pcwalton:

---

*[webrender/Cargo.toml, line 13 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V89Qb7UWRjuvRbECE:-L8Z3UdwDfnOxgpY3VJS:b-t5sirk) ([raw file](https://github.com/servo/webrender/blob/92c2cb48d234ae06db3771e9cddfaf8639b81e48/webrender/Cargo.toml#L13)):*

Why is image no longer an optional dependency?
</blockquote></details>

Because it's needed to load the area LUT, which is stored in PNG format.

---

*[webrender/res/ps_text_run.glsl, line 177 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V9Ckz9qgKX6uvF7Og:-L8dUT3uEKpVs4xdfjrM:b-x2x2y8) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/res/ps_text_run.glsl#L177)):*

Is this necessary? For the normal (CPU rendered) alpha mode, we store the channels correctly in the RGBA output during rasterization, so there's no runtime swizzle needed in this case.
</blockquote></details>

Yes, it's needed to grab the single R8 texture channel properly from a `MODE_ALPHA` texture. Here's what it looks like without that line (notice the text on the top left):


---

*[webrender/src/device.rs, line 441 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V9uOE0Enra6MrzZyX:-L8dUvbwCleVxPoNDZBA:b-s6m8we) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/src/device.rs#L441)):*

Why do we need to store this as TextureTarget and calculate the GL target each time it's used?
</blockquote></details>

To make the `to_external()` method I added work. `ExternalTexture` requires a `TextureTarget`, not a `GLuint`.

---

*[webrender/src/glyph_rasterizer.rs, line 530 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8VFPDQ9Of0VvKWrRln:-L8dVHkwD4wAil_Qk59t:b-sje1ot) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/src/glyph_rasterizer.rs#L530)):*

Could you expand on this TODO?
</blockquote></details>

Oops, that was a comment I forgot to remove. Deleted it.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2566)*

glennw:

---

*[webrender/Cargo.toml, line 13 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V89Qb7UWRjuvRbECE:-L8dguwT6ILBND9o0dIQ:b-xl1aqm) ([raw file](https://github.com/servo/webrender/blob/92c2cb48d234ae06db3771e9cddfaf8639b81e48/webrender/Cargo.toml#L13)):*
<details><summary><i>Previously, pcwalton (Patrick Walton) wrote…</i></summary><blockquote>

Because it's needed to load the area LUT, which is stored in PNG format.
</blockquote></details>

Ah, that makes sense. Should we enable it only when pathfinder is enabled, so that Gecko doesn't pull in a new image dependency then? Perhaps in the future we could consider just embedding it in the binary as raw image data.

---

*[webrender/res/ps_text_run.glsl, line 177 at r1](https://beta.reviewable.io/reviews/servo/webrender/2566#-L8V9Ckz9qgKX6uvF7Og:-L8dhOcX59jEnIQNv0Ib:b-ymwz2q) ([raw file](https://github.com/servo/webrender/blob/a0011e952afc16503d26b3d52c11b3780a6f04f4/webrender/res/ps_text_run.glsl#L177)):*
<details><summary><i>Previously, pcwalton (Patrick Walton) wrote…</i></summary><blockquote>

Yes, it's needed to grab the single R8 texture channel properly from a `MODE_ALPHA` texture. Here's what it looks like without that line (notice the text on the top left):


</blockquote></details>

Instead of writing to an R8 texture though, what if we stored the output as an RGBA texture? That's what we currently do when rasterizing glyphs, even in alpha mode.

The benefits of this are that we don't need an extra swizzle here (pay the cost during glyph rasterizing once rather than an extra cost during rendering the glyph each frame), and batching is better since glyphs will end up in the same texture atlas as almost everything else.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2566)*

pcwalton:
Updated per feedback.

r? @glennw

glennw:
CI failures from ```servo-tidy```:

```
Checking files for tidiness...
  Progress: 0% (2/708)
./Cargo.lock:1: duplicate versions for package `env_logger`
	The following packages depend on version 0.4.3 from 'crates.io':
		pathfinder_partitioner
	The following packages depend on version 0.5.3 from 'crates.io':
		webrender
		wrench
./Cargo.lock:1: duplicate versions for package `num-traits`
	The following packages depend on version 0.1.43 from 'crates.io':
		app_units
		enum_primitive
		euclid
		image
		lyon_geom
		num
		num-integer
		num-iter
		num-rational
		plane-split
		serde_json
		webrender
	The following packages depend on version 0.2.2 from 'crates.io':
		num-traits
```

Are those ones easy to fix up? If not, we can possibly whitelist them as allowed duplicates, but that's not ideal.

glennw:
Reviewed 4 of 25 files at r1, 19 of 23 files at r2.

---

*[webrender/src/device.rs, line 47 at r2](https://beta.reviewable.io/reviews/servo/webrender/2566#-L97qwu9-adeNVfOdFEs:-L97qwu9-adeNVfOdFEt:b-qdi1sv) ([raw file](https://github.com/servo/webrender/blob/9a7610d6384e50d0cfc55616ada6e3248f2c0612/webrender/src/device.rs#L47)):*
> ```Rust
> 
> #[derive(Debug, Copy, Clone, PartialEq, Ord, Eq, PartialOrd)]
> pub struct TextureId(gl::GLuint);
> ```

Is this still needed?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2566)*

glennw:
Looks good! Added a short list of the remaining things to get this merged:

- [ ] Fix up the CI tidy errors above (or whitelist duplicated dependencies if we have to).
- [ ] Do one more pass over glyph_rasterizer and/or get @nical to take a pass over it.
- [ ] Do a gecko try run - see https://wiki.mozilla.org/Platform/GFX/Quantum_Render#Try_syntax.
- [ ] Fix / answer the one remaining reviewable question above.

pcwalton:
I don't know why `num-traits` 1.4.43 depends on `num-traits` 0.2.2. That doesn't make sense to me…

glennw:
That num-traits error is weird - maybe the cargo log is wrong and there are other crates that depend on each of them?

Also CI compile error on Windows (this might be on all platforms via ```cargo test```):

```
error[E0061]: this function takes 8 parameters but 5 parameters were supplied
   --> src\glyph_rasterizer.rs:983:13
    |
613 | /     pub fn request_glyphs(
614 | |         &mut self,
615 | |         glyph_cache: &mut GlyphCache,
616 | |         font: FontInstance,
...   |
669 | |         self.request_glyphs_from_backend(font, new_glyphs);
670 | |     }
    | |_____- defined here
...
983 |               &mut glyph_cache,
    |               ^^^^^^^^^^^^^^^^ expected 8 parameters
error[E0061]: this function takes 6 parameters but 4 parameters were supplied
   --> src\glyph_rasterizer.rs:994:9
    |
738 | /     pub fn resolve_glyphs(
739 | |         &mut self,
740 | |         glyph_cache: &mut GlyphCache,
741 | |         texture_cache: &mut TextureCache,
...   |
809 | |         self.remove_dead_fonts();
810 | |     }
    | |_____- defined here
...
994 |           &mut glyph_cache,
    |           ^^^^^^^^^^^^^^^^ expected 6 parameters
Caused by:
```

Emerentius:
`num-traits` `0.1.43` re-exports some traits from `0.2`. It's the [semver trick](https://github.com/dtolnay/semver-trick/blob/master/README.md).

Eijebong:
Yeah the `num-traits` update is going to take some time, you can probably dupe it in the meantime (we'll probably have to do it in servo too :/)
For now we're blocked mainly on https://github.com/andersk/enum_primitive-rs/pull/16
Once this one is out the way, I think the rest is pretty straightforward.

pcwalton:
@glennw Updated to fix test build failure.

glennw:
@Emerentius Ah, thanks for the info. So, this means we should, for now, white-list having duplicate versions of that crate, right? @pcwalton that can be done in https://github.com/servo/webrender/blob/master/servo-tidy.toml, if that's the correct thing to do here.

Eijebong:
@glennw It is the right thing to do

pcwalton:
@glennw Done.

glennw:
I have pushed a (pending) try build to Gecko with this patch, just to make sure it doesn't break anything in the default code path:

https://hg.mozilla.org/try/rev/20feada4fb73b5467cd45bb274093a78674838d8
https://treeherder.mozilla.org/#/jobs?repo=try&revision=20feada4fb73b5467cd45bb274093a78674838d8

The WR CI build is failing on one of the configs that denies warnings:

```
error(B: unused import: `glyph_cache::CachedGlyphData`(B
  (B--> (Bwebrender/src/resource_cache.rs:24:5(B
   (B|(B
24(B (B| (Buse glyph_cache::CachedGlyphData;(B
   (B| (B    (B^^^^^^^^^^^^^^^^^^^^^^^^^^^^(B
   (B|(B
   (B= (Bnote(B: `-D unused-imports` implied by `-D warnings`(B
error(B: aborting due to previous error(B
```

pcwalton:
Pushed a new change that should eliminate that warning.

glennw:
The failing tests on try appear to be unrelated intermittent failures.

@bors-servo r+ :rocket:

ebraminio:
https://github.com/servo/webrender/pull/2566#issuecomment-375913352
@pcwalton: Thanks for the work, HarfBuzz can work without freetype and working without others callbacks is suggested now actually, the only issue we don't have glyph outline fetch https://github.com/harfbuzz/harfbuzz/issues/1686 yet and having that hopefully makes servo able to do better concurrent rendering in future

PR #2568, opened 2018-03-26
Reset segment clip tasks each frame.

### LINKED ISSUE #2531: Large element with 2d transform disappears sometimes

nical:
Clipping can change each frame. Clearing the segment clip tasks avoids invalidated state from previous frames from leaking into new ones. Fixes #2531.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2568)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1543,6 +1543,18 @@ impl PrimitiveStore {
         true
     }
 
+    fn reset_clip_task(&mut self, prim_index: PrimitiveIndex) {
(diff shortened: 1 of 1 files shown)


nical:
This PR replaces #2564.
r? @glennw

PR #2572, opened 2018-03-26
Decompose repeated images and gradients into brush segments in most cases.

nical:
Moves decomposition of images and gradients to frame building.

- [x] Clip against the viewport to avoid generating segments that are off-screen.
- [x] Implement tiled image decomposition.
- [x] Properly handle the segment parameters (aa flags, etc).
- [x] Use this for gradients.

Support for large amounts of repetitions will come in a followup PR.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2572)


nical:
This PR is (or is close to be) in a landable state now. It does not have a good solution for decomposing image with many repetitions, so we fall back to the non-brush shader in this case (during scene building). I'd prefer tackling this problem as a followup since it's not trivial.
A few notable changes in this PR:
 - Make the local clip rect as tight as possible (intersecting it with the brush's local rect) to handle the repetitions that are only partially inside the local rect.
 - Pass a source uv rect in the per-segment gpu data (existing TODOs in the code seemed to indicate this was the plan). This rect in normalized (0..1) coordinates indicates which portion of the source image to sample from (I mean source image as in item in a texture so in practice it is the sub rect of a sub rect when several images are packed into a single texture).
 - Compute the segment source uv rects on the CPU (it used to be implicit in the shader).

It's probable that the source uv rect now allows us to use brushes for images with a sub-rect (I haven't looked into that yet).

webrender/src/prim_store.rs
```diff
@@ -298,6 +302,7 @@ pub enum BrushClipMaskKind {
 pub struct BrushSegmentDescriptor {
     pub segments: Vec<BrushSegment>,
     pub clip_mask_kind: BrushClipMaskKind,
+    pub source_per_segment: bool,
```
nical: I am not sure what to call this (I am only sure that I a don't like the current name I gave it). Right now this defines how to compute the UVs for each segment (either have all segments sample the entire source image or have segments sample parts of the source image).
Suggestions are very much welcome.

Maybe an enum `SegmentUv` with variants `Auto` and `Full` would be better than a boolean.

kvark:
Reviewed 11 of 13 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.

---

*[webrender/res/brush_linear_gradient.glsl, line 43 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8deGmA19TWVUQmtg0f:-L8deGmA19TWVUQmtg0g:b-gzzvd7) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/res/brush_linear_gradient.glsl#L43)):*
> ```GLSL
>     Gradient gradient = fetch_gradient(prim_address);
> 
>     vPos = segment_src_rect.p0 * segment_local_rect.size + (vi.local_pos - segment_local_rect.p0) * segment_src_rect.size;
> ```

please explain this change, maybe split it into multiple sentences

---

*[webrender/res/brush_mix_blend.glsl, line 18 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8dePPU2LErgkGoI6O9:-L8dePPY3NoaGw3OMU7V:b-a0e3pf) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/res/brush_mix_blend.glsl#L18)):*
> ```GLSL
>     VertexInfo vi,
>     int prim_address,
>     RectWithSize segment_local_rect,
> ```

not sure if we really need the `segment_` prefix for this and UV rect

---

*[webrender/src/image.rs, line 297 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e4a7G-Qgnwx9qmxMZ:-L8e4a7G-Qgnwx9qmxM_:b-xgp99q) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/image.rs#L297)):*
> ```Rust
>     }
> 
>     let clip = clip.unwrap();
> ```

```rust
let clip = match clip {
   Some(c) => c,
   None => return,
};
```

---

*[webrender/src/image.rs, line 300 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e59iSBfPbhWbgvAbI:-L8e59iSBfPbhWbgvAbJ:bj0h8mm) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/image.rs#L300)):*
> ```Rust
> 
>     let nx = if clip.origin.x > prim_rect.origin.x {
>         f32::floor((visible_rect.origin.x - prim_rect.origin.x) / stride.width)
> ```

any reason to use global functions here instead of method-style invocation (`xxx.floor()`)?

---

*[webrender/src/image.rs, line 306 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e5Y8p0qZ1d8_h8D5O:-L8e5Y8p0qZ1d8_h8D5P:b-c9sxp1) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/image.rs#L306)):*
> ```Rust
> 
>     let ny = if clip.origin.y > prim_rect.origin.y {
>         f32::floor((visible_rect.origin.y - prim_rect.origin.y) / stride.height)
> ```

it would be better to use `clip` versus `visible_rect` consistently here (both in condition and the body, not different like they are now)

---

*[webrender/src/prim_store.rs, line 1360 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e82FI5zyWSc9tTqvE:-L8e82FI5zyWSc9tTqvF:b-m7qb4x) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/prim_store.rs#L1360)):*
> ```Rust
>                             if tile_spacing.height > 0.0 {
>                                 base_edge_flags |= EdgeAaSegmentMask::TOP | EdgeAaSegmentMask::BOTTOM;
> ```

could you explain a bit on why `base_edge_flags` is needed

---

*[webrender/src/util.rs, line 159 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e8sdC5ab8yX2p29kH:-L8e8sdC5ab8yX2p29kI:bcv2f94) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/util.rs#L159)):*
> ```Rust
> }
> 
> pub fn screen_to_layer_bounding_rect(
> ```

we can't do it like this. Proper way would be to use `inverse_rect_footprint` instead, which was introduced for this task.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2572)*

glennw:
I just completely confused myself - I started looking at this thinking it was attempting to handle tiling in the frame builder, then I started writing up the reply below, and then I realized that it's not what this PR is attempting to do at all.

I'll paste it below anyway, since I think it might be relevant to what we do here. Specifically, we might be able to use something like the idea below for both tiling and large repetitions (via pre-caching small images with large repeats to a tile size image). If it works out as I think it would, it has the advantage of always generating a fairly minimal number of segments, regardless of tiling / repetition / size of the source texture.

---

I'm not sure that we need the segment_uv_rect / segment_src_rect to handle this. I think, if I'm correct about that, the SegmentSrc enum could be removed too (well, made specific to brush_image user data, rather than used by all brush types).

I sketched out below roughly what I had in my head. I'll apologize in advance for the hand-waviness, I haven't actually dug into this, so I could easily be missing some critical details here.

We start with no segment information for a tiled image - the brush segment descriptor is None. In prepare_prim_for_render, we do the following (for tiled images only):

* Set a local rect to the entire local rect of the image primitive (rounded up to tile size), and transform it to screen-space.
* While that screen rect is visible, recursively (or iteratively perhaps):
  * Split that local rect into 4 sub-quads, and transform each of those, checking visibility in screen-space.
  * This is basically a simple quad-tree that gets evaluated to find which tiles are visible.
* Use the results of that recursive subdivision to build the segment descriptor.
  * This means we only generate approximately the right number of segments for what is visible.
  * We can cache the results of this in the image primitive brush, so we only evaluate the visibility on frames where scrolling has occurred.

* For each segment that we found to be visible, we:
  * Request that image tile from the texture cache.
  * Set the user data field of the segment instance to be the texture cache item for that tile.

* In brush_image.glsl, we pass both the local primitive rect and the local segment rect.
  * We pass a user data field that describes for the brush_image shader whether the texture is for this segment or the whole primitive.
  * This determines whether we calculate our 'f' factor for UV generation from the segment or complete primitive rect.

glennw:
@nical I know we have timezone issues, but we could try to set up a vidyo call to try and run through this in more detail? I can try and do either an early morning or evening call, if that would be helpful?

glennw:
So, the basic idea would be that to do repetition, we do a similar thing - if the image primitive is large enough (relative to the size of the source image / number of repetitions), we divide that into a small number of segments, ~512px each. And then, for each of those, we pre-render (via the render task cache) the right image.

For tiling with no repeats, this is just requesting the tile. For large repetitions of non-tiled images, this is doing a series of blits to a texture cache item to pre-repeat up to a 512px tile, and then drawing those as a small number of segments. For images with both tiling and repeat, it gets a bit trickier, but I think is still doable. For spacing, we can include that in the pre-rendering step.

I'm sorry if this is all just a wall of words, it's mostly a dump of what I've been pondering for the "long term" solution to this stuff.

glennw:
(The pre-render step would use ```request_render_task```, which caches the output in the texture cache across multiple frames / display lists, so long as the cache key matches).

glennw:
All of the above is just rough ideas by the way, it's not something I'm super convinced is the *right* way to do it - it's quite likely there's flaws in it, or a better way to achieve what we want to support.

nical:
@glennw So this PR is really just about handling repetitions and tile spacing. The tiling would be implemented on top of that.
There a re a few points I did not completely understand (but maybe that's because some of them are more specific to how to do tiling rather than repetitions):

> I'm not sure that we need the segment_uv_rect / segment_src_rect to handle this. I think, if I'm correct about that, the SegmentSrc enum could be removed too (well, made specific to brush_image user data, rather than used by all brush types).

> We pass a user data field that describes for the brush_image shader whether the texture is for this segment or the whole primitive.

I almost did that but ended up specifying a rectangle per segment because of [a comment](https://github.com/servo/webrender/blob/22493328352ba432a7cd89491d81bfaa19bc1bce/webrender/src/prim_store.rs#L2124) in prim_store that seems to say that it was the plan. This rect makes it easy to implement [image primitives with a sub-rect](https://github.com/servo/webrender/blob/22493328352ba432a7cd89491d81bfaa19bc1bce/webrender/src/display_list_flattener.rs#L2127) for the brush image shader, and there are a few other tricks we can pull if we have full control over how we sample from the source texture (for example nine-patch/stretching type stuff comes for free).
If instead we pass a flag through user data, we save an unused gpu cache slot, but trading a multiply and an add with a branch and a move is not a clear win on the vertex shader side (although the code would be simpler).

> While that screen rect is visible, recursively [..] Split that local rect into 4 sub-quads, and transform each of those, checking visibility in screen-space.

I need some more details about what the recursive subdivision brings as opposed to just iterating over the visible rect like the PR currently does. I can make this iteration happen in screen space instead of layer-space if you prefer.
I think that the part about pre-rendering a bunch of the repetitions should be orthogonal with how repetitions themselves are rendered (be it the screen or the cache), my plan is to implement it next.

> For spacing, we can include that in the pre-rendering step.

Tile spacing is the easy part. Whether we render the spacing in the pre-rendering step or directly to the screen makes no difference since either way we need to be able to render spacing, and that's just a matter of how the segment rects are built. I mean we don't need pre-rendering for tile spacing or worry about tile spacing at all.

> We can cache the results of this in the image primitive brush, so we only evaluate the visibility on frames where scrolling has occurred.

I don't think it's worth optimizing for non-scrolling frames. Scrolling is the biggest reason for new frames to be generated and it's also the most important type of frames to keep smooth so I think it should be what we optimize and not the other way around.
What I had in mind was to first decide whether it is worth pre-rendering chunks of repetitions into the cache, and then we only have to compute the visibility of the repeated chunks by iterating the visible rect (and there's only going to be a few chunks visible at any given time).

nical:
> @nical I know we have timezone issues, but we could try to set up a vidyo call to try and run through this in more detail? I can try and do either an early morning or evening call, if that would be helpful?

That would be helpful, I can stay late this thursday night (friday morning for you), or come in early this thursday or friday however you prefer.

nical:
@glennw and I talked about this, and I am going to switch the segment rect logic to just passing a flag in the user data of image and gradient shaders and select the rect in the vertex shader instead of computing a sub-rect computed on the CPU. We can solve the image primitive with sub-rect case using other tricks. The culling phase will remain based on a layer-space iteration over the visible rect for now.

nical:
Added a commit to switch the logic to select whether the source should be sampled at the primitive or segment level to the vertex shader.
Since all `user_data` fields are already taken in the brush_image shader I ended up packing it with the image source which adds some bit fiddling but oh well.

I also moved linear gradients over to this to see how the thing generalizes. There is a bit of code duplication in prim_store, but I want to first figure out tiled images before addressing that (as factoring out the duplicated code won't be as easy when we have special logic for the tiled image case).

nical:
*[webrender/res/brush_mix_blend.glsl, line 18 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8dePPU2LErgkGoI6O9:-L8q5_Y6AFwfuak78PyX:b-7io6tj) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/res/brush_mix_blend.glsl#L18)):*

not sure if we really need the `segment_` prefix for this and UV rect
</blockquote></details>

I am pretty sure (I did mix up segment and primitive local rects a bunch in the making of this PR).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2572)*

nical:
Review status: 2 of 14 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[webrender/res/brush_linear_gradient.glsl, line 43 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8deGmA19TWVUQmtg0f:-L8q5Py5Ds__CHiB-mrz:b4oq3xe) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/res/brush_linear_gradient.glsl#L43)):*

please explain this change, maybe split it into multiple sentences
</blockquote></details>

Done. (The part this comment was referring to rolled back to be close to the original code)

---

*[webrender/src/image.rs, line 297 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e4a7G-Qgnwx9qmxMZ:-L8q5rE8-Jy1x8xrP8DI:b-896fix) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/image.rs#L297)):*

```rust
let clip = match clip {
   Some(c) => c,
   None => return,
};
```
</blockquote></details>

Done.

---

*[webrender/src/image.rs, line 300 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e59iSBfPbhWbgvAbI:-L8q5rr13AFMnJ1mGLXi:bilvgeu) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/image.rs#L300)):*

any reason to use global functions here instead of method-style invocation (`xxx.floor()`)?
</blockquote></details>

Yeah, I passionately dislike the method style invocation for these math functions.

---

*[webrender/src/image.rs, line 306 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e5Y8p0qZ1d8_h8D5O:-L8q69GLBomFkvXPLXAy:b-896fix) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/image.rs#L306)):*

it would be better to use `clip` versus `visible_rect` consistently here (both in condition and the body, not different like they are now)
</blockquote></details>

Done.

---

*[webrender/src/prim_store.rs, line 1360 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e82FI5zyWSc9tTqvE:-L8q6BmQ83d5ZRgWKZWf:b-896fix) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/prim_store.rs#L1360)):*

could you explain a bit on why `base_edge_flags` is needed
</blockquote></details>

Done.

---

*[webrender/src/util.rs, line 159 at r2](https://beta.reviewable.io/reviews/servo/webrender/2572#-L8e8sdC5ab8yX2p29kH:-L8q6DO85QKjS8IlOmSt:b-896fix) ([raw file](https://github.com/servo/webrender/blob/50d3cdd8b4f60875f0984d0f8ccb1dc51b24b338/webrender/src/util.rs#L159)):*

we can't do it like this. Proper way would be to use `inverse_rect_footprint` instead, which was introduced for this task.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2572)*

nical:
3 failing tests, two of which are aa differences which need to be fuzzed, while the tile-spacing one is definitely a bug (looking into it).

@glennw It would be good to start the review now rather than wait for the last piece (pre-rendering for many repetitions) to be implemented as there is a lot here already.

nical:
Fixed the 3 failing tests, then rebased, and apparently lost the rebasing game (new failures).
Looking into it.

nical:
The gecko try push has a lot of failures due to hitting the maximum number of repetitions, so I guess this can't land without the extra work of pre-rendering repetitions when there are too many of them.

In retrospect, while I think that it's good to use the brush shader, I don't think segments are a good fit for handling repetitions. My concerns are:
- We lose the ability to further segment them for clipping, without having to rethink the way we deal with the source uv rects.
- We quickly run into the limit of the number of segments per primitive.
- The extra flexibility we need (segments sampling from different image requests, etc) adds complexity to the segments code that doesn't exist if each tile is "primitive-like" rather than a segment.
- It's strictly more complicated than the approach I took in my attempt without using segments, and I don't any real benefit.

I think that we would be better off adding support for repetition in the image and gradient brush shaders (but not spacing), and pre-render spacing (I am not sure but it might be something @glennw proposed a little while back I forgot about it during the making of this).
For tiling, I think that the simplest would be to handle this in the image brush primitive with an array of visible tiles without segments in a way closer to what I did in the first PR. I think it would be simpler but we can also stick to segments since the code is already written (and hope that we never run into the max number segments).

My next step is to work on pre-rendering because it will work the more or less the same way whether we chose to stick to what this PR does, or move to what I just proposed.

nical:
Closing in favor of #2695.

PR #2574, opened 2018-03-26
Make DebugRenderer optional to save startup time

fschutt:
This change makes the `DebugRenderer` optional, thereby saving the compilation time of two shaders ("debug_font" and "debug_color") and improving the startup time of webrender. If you don't need to debug webrender, there is no point in having a `DebugRenderer` and no point in compiling the shaders. 

The relevant feature is called `debug_renderer`. It is enabled by default for the features `debugger` and `profiler` (mostly because `wrench` needs them). Otherwise, this feature is not enabled by default, which also saves roughly 2.5 MB of RAM.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2574)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1290,6 +1290,7 @@ dependencies = [
  "bincode 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitflags 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-foundation 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-text 9.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -9,7 +9,7 @@ build = "build.rs"
 [features]
 default = ["freetype-lib"]
 freetype-lib = ["freetype/servo-freetype-sys"]
-profiler = ["thread_profiler/thread_profiler"]
+profiler = ["thread_profiler/thread_profiler", "debug_renderer"]
 debugger = ["ws", "serde_json", "serde", "image", "base64", "debug_renderer"]
 capture = ["webrender_api/serialize", "ron", "serde"]
 replay = ["webrender_api/deserialize", "ron", "serde"]
@@ -39,6 +39,7 @@ serde = { optional = true, version = "1.0", features = ["serde_derive"] }
 image = { optional = true, version = "0.18" }
 base64 = { optional = true, version = "0.6" }
 ron = { optional = true, version = "0.1.7" }
+cfg-if = "0.1.2"
 
 [dev-dependencies]
 mozangle = "0.1"
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -3,9 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use super::shader_source;
-use api::{ColorF, ImageDescriptor, ImageFormat};
+use api::{ColorF, ImageFormat};
 use api::{DeviceIntPoint, DeviceIntRect, DeviceUintRect, DeviceUintSize};
 use api::TextureTarget;
+#[cfg(feature = "debug_renderer")]
+use api::ImageDescriptor;
 use euclid::Transform3D;
 use gleam::gl;
 use internal_types::{FastHashMap, RenderTargetInfo};
@@ -20,7 +22,6 @@ use std::path::PathBuf;
 use std::ptr;
 use std::rc::Rc;
 use std::thread;
-
 
 #[derive(Debug, Copy, Clone, PartialEq, Ord, Eq, PartialOrd)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
@@ -60,6 +61,7 @@ const DEFAULT_TEXTURE: TextureSlot = TextureSlot(0);
 
 #[repr(u32)]
 pub enum DepthFunction {
+    #[cfg(feature = "debug_renderer")]
     Less = gl::LESS,
     LessEqual = gl::LEQUAL,
 }
@@ -76,6 +78,7 @@ pub enum TextureFilter {
 #[derive(Debug)]
 pub enum VertexAttributeKind {
     F32,
+    #[cfg(feature = "debug_renderer")]
     U8Norm,
     U16Norm,
     I32,
@@ -230,6 +233,7 @@ impl VertexAttributeKind {
     fn size_in_bytes(&self) -> u32 {
         match *self {
             VertexAttributeKind::F32 => 4,
+            #[cfg(feature = "debug_renderer")]
             VertexAttributeKind::U8Norm => 1,
             VertexAttributeKind::U16Norm => 2,
             VertexAttributeKind::I32 => 4,
@@ -265,6 +269,7 @@ impl VertexAttribute {
                     offset,
                 );
             }
+            #[cfg(feature = "debug_renderer")]
             VertexAttributeKind::U8Norm => {
                 gl.vertex_attrib_pointer(
                     attr_index,
@@ -458,10 +463,12 @@ impl Texture {
         self.format
     }
 
+    #[cfg(feature = "debug_renderer")]
     pub fn get_filter(&self) -> TextureFilter {
         self.filter
     }
 
+    #[cfg(feature = "debug_renderer")]
     pub fn get_render_target(&self) -> Option<RenderTargetInfo> {
         self.render_target.clone()
     }
@@ -636,6 +643,7 @@ impl UniformLocation {
     pub const INVALID: Self = UniformLocation(-1);
 }
 
+#[cfg(feature = "debug_renderer")]
 pub struct Capabilities {
     pub supports_multisampling: bool,
 }
@@ -662,6 +670,7 @@ pub struct Device {
     upload_method: UploadMethod,
 
     // HW or API capabilties
+    #[cfg(feature = "debug_renderer")]
     capabilities: Capabilities,
 
     // debug
@@ -708,6 +717,7 @@ impl Device {
             upload_method,
             inside_frame: false,
 
+            #[cfg(feature = "debug_renderer")]
             capabilities: Capabilities {
                 supports_multisampling: false, //TODO
             },
@@ -749,6 +759,7 @@ impl Device {
         self.max_texture_size
     }
 
+    #[cfg(feature = "debug_renderer")]
     pub fn get_capabilities(&self) -> &Capabilities {
         &self.capabilities
     }
@@ -1451,6 +1462,7 @@ impl Device {
         }
     }
 
+    #[cfg(feature = "debug_renderer")]
     pub fn get_uniform_location(&self, program: &Program, name: &str) -> UniformLocation {
         UniformLocation(self.gl.get_uniform_location(program.id, name))
     }
@@ -1518,6 +1530,7 @@ impl Device {
         }
     }
 
+    #[cfg(feature = "debug_renderer")]
     pub fn read_pixels(&mut self, img_desc: &ImageDescriptor) -> Vec<u8> {
         let desc = gl_describe_format(self.gl(), img_desc.format);
         self.gl.read_pixels(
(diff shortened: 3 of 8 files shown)


fschutt:
Yeah, of course, the tests need to be fixed, sorry ...

nical:
Having these debug utilities in release mode is useful, but it'd be great to initialize it lazily to avoid the startup cost.

fschutt:
@nical As far as I can see, servo doesn't even use the `DebugRenderer` in any way. The only tool using the `DebugRenderer` is wrench, which is the debugger for webrender. So it would make more sense to make this a compile-time feature, wouldn't it?

nical:
I use it in Firefox quite a bit, and it'd be good to not have keep Firefox startup times as fast as possible.

nical:
Also a some of the information provided by the debug renderer, like performance counters, only makes sense in optimized builds.

fschutt:
@nical I think I'll do both, lazy initialization and a compile-time feature. For people who don't need the `DebugRenderer` at all, it is disabled by default. I've also noticed that getting rid of the `DebugRenderer` also elleviates the need to create a `Profiler` in the `Renderer` struct (https://github.com/servo/webrender/blob/master/webrender/src/renderer.rs#L1152) - since the `DebugRenderer` is the only struct that uses the `Profiler` (as far as I can see). If I disable the `Profiler`, I can remove the whole module `debug_profiler.rs`, which saves compilation time (of course, only if you don't need the `DebugRenderer` or the `Profiler`).

So I think, moving the `DebugRenderer` and the `Profiler` to an optional feature is a good idea - for shipping Firefox, you could have the option to include the `DebugRenderer`, but people who don't need it (like me) can disable it to save compilation time, startup time and memory.

nical:
I doubt it will make a significant difference for servo build times, but as long as the debug render is available by default in wrench and Firefox it's fine by me. You should check with @glennw about servo, I think that he uses it in this configuration regularly.

fschutt:
Ok now I've created a feature `debug_renderer` in webrender, it's enabled by default if the `debugger` feature is enabled (which is true for wrench, but may be false for servo). So, no change for wrench. Lazy initialization is also done, although there's a slight bit of code duplication. If you want to enable the debugger in servo, you'd need to enable `debug_renderer`.

fschutt:
I removed any code that wasn't needed (see commit message). I added the crate [`alexcrichton/cfg_if`](https://github.com/alexcrichton/cfg-if) because it has a macro to disable multiple imports in one go, making the imports a bit more readable. Not sure if this is allowed, I didn't want to write:

```rs
#[cfg(feature = "debug_renderer")]
use foo::Foo;
#[cfg(feature = "debug_renderer")]
use bar::Bar;
#[cfg(feature = "debug_renderer")]
use baz::Baz;
```

but rather 

```rs
cfg_if! {
    if #[cfg(feature = "debug_renderer")]  {
        use foo::Foo;
        use bar::Bar;
        use baz::Baz;
    }
}
```

It's a bit more readable. I'll remove it if it isn't allowed to add crates without permission.

fschutt:
The `debug_renderer` feature is now only enabled for the `debugger` and `profiler` features.

glennw:
We should add a config to the taskcluster file in the repo that compiles with / without this feature, so we don't regress compilation in either mode.

I think having cfg_if is fine, but it would probably be better just to exclude entire sources files if this is disabled - for example, the entire profiler.rs file is not useful (I think?) if there's no debug output enabled to render it.

fschutt:
Hm, isn't the with / without already taken care of by compiling webrender / wrench? webrender has the feature disabled, while wrench has it enabled. So yes, you could make a seperate pass, but it would test the same things as we do now - unless webrender would be tested completely seperately from wrench, which it isn't. 

As for the profiler.rs - I did try to disable the whole module, but some functions / structs are still used in other modules. I could move that to a different module, though and refactor the profiler.rs into multiple modules.

glennw:
Reviewed 6 of 7 files at r3.
Review status: 6 of 7 files reviewed at latest revision, all discussions resolved.

---

*[webrender/src/renderer.rs, line 80 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eScQ109S_fA4rTFg4:-L8eScQ109S_fA4rTFg5:b-2hj2cp) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L80)):*
> ```Rust
> cfg_if! {
>     if #[cfg(feature = "debug_renderer")] {
>         use api::ColorU;    
> ```

nit: trailing spaces

---

*[webrender/src/renderer.rs, line 2182 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSoIy8Sjj0ECLCE9E:-L8eSoIy8Sjj0ECLCE9F:b-2hj2cp) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L2182)):*
> ```Rust
>             self.gpu_profile.end_frame();
>             #[cfg(feature = "debug_renderer")]
>             {        
> ```

nit: trailing spaces

---

*[webrender/src/renderer.rs, line 2184 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eTLHyATE_s6xpbta9:-L8eTLHyATE_s6xpbtaA:b-su0fpc) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L2184)):*
> ```Rust
>             {        
>                 let device = &mut self.device;
>                 self.debug.get_or_insert_with(|| DebugRenderer::new(device))
> ```

Instead of having an Option and doing this wherever it's accessed, how about we always create a DebugRenderer (even if it's an empty struct when feature is disabled), which internally lazily initializes when it's used?

---

*[webrender/src/renderer.rs, line 3687 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSr6e48UR94vXeZeI:-L8eSr6e48UR94vXeZeJ:b-2hj2cp) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L3687)):*
> ```Rust
>     #[cfg(feature = "debug_renderer")]
>     fn draw_epoch_debug(&mut self) {
>         if !self.debug_flags.contains(DebugFlags::EPOCHS) || 
> ```

nit: trailing spaces

---

*[webrender/src/renderer.rs, line 3762 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSs2103D-mbHjS9xG:-L8eSs2103D-mbHjS9xH:b-2hj2cp) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L3762)):*
> ```Rust
>         self.device.delete_vao(self.clip_vao);
>         self.device.delete_vao(self.blur_vao);
>         
> ```

nit: trailing spaces

---

*[webrender/src/renderer.rs, line 3764 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSsfe2n99izHy_H51:-L8eSsfe2n99izHy_H52:b-2hj2cp) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L3764)):*
> ```Rust
>         
>         #[cfg(feature = "debug_renderer")]
>         {         
> ```

nit: trailing spaces

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2574)*

fschutt:
The `DebugRenderer` needs a `&mut Device` for intialization and deinitialization. So every call to `LazyInitializedDebugRenderer::get_mut()` needs to pass a `&mut self.device` along, in case the `DebugRenderer` hasn't been initialized up until this point. 

Trailing whitespace should be fixed now.

fschutt:
Review status: 6 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[webrender/src/renderer.rs, line 80 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eScQ109S_fA4rTFg4:-L8hHdHKCzBYRI_Zx-U6:b-896fix) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L80)):*

nit: trailing spaces
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 2182 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSoIy8Sjj0ECLCE9E:-L8hHFlKEo6k6m3Y479f:b-896fix) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L2182)):*

nit: trailing spaces
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 2184 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eTLHyATE_s6xpbta9:-L8h-FkWEfTsKy_ZBK_S:b-29g398) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L2184)):*

Instead of having an Option and doing this wherever it's accessed, how about we always create a DebugRenderer (even if it's an empty struct when feature is disabled), which internally lazily initializes when it's used?
</blockquote></details>

Done. See the LazyInitializedDebugRenderer struct.

---

*[webrender/src/renderer.rs, line 3687 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSr6e48UR94vXeZeI:-L8hHBdICDbWqcVVIghb:b-896fix) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L3687)):*

nit: trailing spaces
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 3762 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSs2103D-mbHjS9xG:-L8hGvIdE_t6cRTFsBH5:b-896fix) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L3762)):*

nit: trailing spaces
</blockquote></details>

Done.

---

*[webrender/src/renderer.rs, line 3764 at r3](https://beta.reviewable.io/reviews/servo/webrender/2574#-L8eSsfe2n99izHy_H51:-L8hH9td4v2Qt5IoHqs3:b-896fix) ([raw file](https://github.com/servo/webrender/blob/93563784839b6d6334d0e29aa06ec06127142123/webrender/src/renderer.rs#L3764)):*

nit: trailing spaces
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2574)*

fschutt:
@glennw fixed everything, ready to merge. I am not sure how to close discussions on Reviewable so that the the "6 discussions left" goes away. But if there are no further problems, I'd merge it before this branch inevitably gets merge conflicts. Sorry for being so pushy, but I hate stale PRs where I can't do anything to advance them.

glennw:
Looks good, thanks!

@bors-servo r+

staktrace:
This seems to have broken the gecko build. See https://treeherder.mozilla.org/logviewer.html#?job_id=171025276&repo=try&lineNumber=13033 for the failure log. https://bugzilla.mozilla.org/show_bug.cgi?id=1449562#c2 has the regression range.

staktrace:
@fschutt It looks like we're building webrender with just the capture feature, see https://hg.mozilla.org/try/file/c405075b734f1ae4a50d7bfae00e34e5c7449aa0/gfx/webrender_bindings/Cargo.toml#l19 - does this feature need to include the new debug_renderer feature too?

fschutt:
@staktrace Is this a debug or release build? If it's a debug build, just add the feature `debug_renderer` and it should build. If this is a release build, I probably have to go back and split the `debug_renderer` feature into two parts. Because in release mode, it's supposed to work with the feature disabled. Otherwise, this PR would be pretty pointless.

staktrace:
@fschutt We do both debug and release. And both are failing. I can reproduce just by running `cargo build --no-default-features --features capture` in a standalone webrender checkout. Same happens if I add `--debug`.

staktrace:
As far as I can tell the DebugRenderer feature is not used by Gecko, unless the capture feature internally requires it.

I've put up a patch on #2591 to fix this for now the "obvious" way, feel free to write another patch that adds the extra annotations. I thought that in general every combination of features is supposed to be valid in rust, since features are additive.

fschutt:
@staktrace The `capture` feature doesn't seem to need access to the `DebugRenderer`. I just disabled a few too many functions on the `Device` because rustc complained they were never used. I probably shouldn't have done that. I'll add these few functions to the `capture` feature and then remove the `debug_renderer` from the `capture` feature. That should be the "proper" fix.

kvark:
@fschutt are you intending to do this now, or want the quick fix merged first?

fschutt:
I've already fixed it, just added `#[cfg(any(feature = "debug_renderer", feature="capture"))]` to some functions in `device.rs`. I'll do a seperate PR, do the quick fix for now.

fschutt:
See https://github.com/servo/webrender/pull/2593

PR #2575, opened 2018-03-27
Remove PictureKind enum.

glennw:
Previously, we had separate Picture kinds for box shadows and
text shadows.

Now, box-shadows are drawn as clip sources, and text shadows are
drawn as a regular picture with a blur filter set as the picture
composite mode.

This PR should not contain any functional changes - it's just
re-arranging the code to no longer use PictureKind.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2575)


glennw:
r? anyone

PR #2578, opened 2018-03-27
Allow borders to be clipped by clip nodes

mrobinson:
For some reason borders that are segmented are not clippable by clip
scroll nodes. This change fixes that issue and adds a test for that
behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2578)


mrobinson:
@glennw This change is so tiny, I feel like I must be misunderstanding something. Do you mind taking a look here and setting me straight if this is incorrect? Thanks!

Gankro:
r=me, but feel free to not merge this if you want glenn's feedback first

glennw:
This does seem correct to me, I don't think you're missing anything. We should probably do a try run first though, just to be safe. I can kick one off for this patch this morning, unless you already have one pending?

glennw:
Pending try run:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=358f71697a04fd75c2d40948ebd41b8bd54f9f40

glennw:
Orange items appear unrelated.

@bors-servo r+

PR #2585, opened 2018-03-28
Handle backface-visibility:hidden in hit testing (Gecko bug 1447131)

theres-waldo:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2585)

### DIFF
--- a/webrender/src/hit_test.rs
+++ b/webrender/src/hit_test.rs
@@ -57,6 +57,7 @@ pub struct HitTestingItem {
     rect: LayerRect,
     clip_rect: LayerRect,
     tag: ItemTag,
+    is_backface_visible: bool,
 }
 
 impl HitTestingItem {
@@ -65,6 +66,7 @@ impl HitTestingItem {
             rect: info.rect,
             clip_rect: info.clip_rect,
             tag: tag,
+            is_backface_visible: info.is_backface_visible,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


theres-waldo:
r? @mrobinson 

I wonder if it's ok to increase the size of every HitTestingItem like this, or if we can somehow pack that bit of info somewhere more efficiently, or even avoid constructing the relevant hit test items altogether?

webrender/src/hit_test.rs
```diff
@@ -265,6 +267,11 @@ impl HitTester {
                     None => continue,
                 };
 
+                // Don't hit items with backface-visibility:hidden if they are facing the back.
+                if !item.is_backface_visible && transform.is_backface_visible() {
```
mrobinson: We should cache the value of `transform.is_backface_visible()` above since it calls `Transform3D::determinant()` which is a fairly expensive operation.

theres-waldo: Fixed. I also took care not to compute `transform.is_backface_visible()` eagerly in the case where none of the items are backface-visibility:hidden (which is the common case).

webrender/src/hit_test.rs
```diff
@@ -265,6 +268,17 @@ impl HitTester {
                     None => continue,
                 };
 
+                // Don't hit items with backface-visibility:hidden if they are facing the back.
+                if !item.is_backface_visible {
+                    if facing_backwards.is_none() {
```
mrobinson: Thanks for fixing this. Can you use `Option::get_or_insert_with` here? I think this could become a one-liner with that. I'm thinking something like this:

```rust
if *facing_backward.get_or_insert_with(|| transform.is_backface_visible()) {
    continue;
}
```

theres-waldo: Will do, thanks for the suggestion!

PR #2589, opened 2018-03-29
Fix up some shader validations, remove unused shader.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2589)


glennw:
r? anyone

PR #2597, opened 2018-03-31
Added DPI to GlyphOptions to indicate the DPI for glyph rasterization

### LINKED ISSUE #2596: Font sizes don't respect the monitor DPI on Linux
Whenever I create a `FontInstanceKey` with, let's say, 10px, the actual font that ends up on the screen is 6px or 7px large. 

For reference, I set the font in the title bar of the window to the same font at 10px:

Webly Sleeky UI:

azul_webrender_bug

Droid Sans:

pic_azul_wrong_droidsans

Both times, I set the font size to exactly `10 px * AU_PER_PX`. The problem is that I don't know what kind of scale factor I have to apply to get the actual, correct 10px font size on the screen. I originally noticed this as a bug in limn [here](https://github.com/christolliday/limn/issues/22#issuecomment-348175574) (last bullet point), so it's not only my code. 

What kind of scaling does the `add_font_instance()` apply and how do I get the font to be actually 10 pixels high? **To be clear, the HighDPI factor is set to 1 in both examples** - the high_dpi factor reports "1.0", and my screen is set to 96 DPI.

fschutt:
Fixes #2596 - the PR is a bit all over the place, not sure if the DPI is needed in all structs. Also not sure if Windows or Mac have the same problems. The DPI still has to be set by the user, but at least there is an option to set it, instead of having none.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2597)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,7 +395,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "freetype"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "libc 0.2.20 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -908,7 +908,7 @@ dependencies = [
  "lazy_static 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.20 (registry+https://github.com/rust-lang/crates.io-index)",
  "num_cpus 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "rand 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


glennw:
I'm not sure why this doesn't affect Gecko, if it's an issue? Comments / thoughts @lsalzman or @jrmuizel ?

jrmuizel:
@lsalzman might recall more specifically, but I believe we do the 97/72 scaling someplace in Gecko. 

However, I don't think it makes sense to set a dpi per font. If we do anything this should just be in RendererOptions.

fschutt:
Yeah, I think I'll need to fix this PR up somehow else, especially since the horizontal DPI and the vertical DPI don't need to be the same. However, it doesn't make sense to hard-code some DPI values, if Gecko does that. For example, my system settings allow me to set the system font to 120 DPI. 96 DPI would then look out of place. 

I've created https://github.com/tomaka/winit/pull/442 for an example of how the DPI could be detected from the monitor size - at least so that the font isn't to small or too big. The next step would then be to respect the system DPI instead of calculating the DPI from the monitor dimensions.

pcwalton:
Can we hold off on merging this until Pathfinder is merged? This will almost certainly break it…

fschutt:
@pcwalton I had some problems with using your pathfinder PR and couldn't test that your PR fixes the issue because of that - see https://github.com/servo/webrender/issues/2617

fschutt:
https://github.com/pcwalton/pathfinder/blob/6e1fe50202586ea3f0400299407fabd2447c7fae/font-renderer/src/freetype/mod.rs#L45

Well, that's one way to do it ... #2596 is still unsolved (and gives the same result as before). However, since pathfinder is the only way forward, I am closing this in favor of a better PR. At least I can now give pathfinder a scaling factor by which to do the 96 / 72 conversion, so thanks for that.

PR #2598, opened 2018-04-01
Use serde_bytes crate

jonleighton:
This speeds up serializing and deserializing of binary data.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,7 +395,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "freetype"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "libc 0.2.20 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -965,6 +965,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_bytes"
+version = "0.10.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "serde 1.0.35 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "serde_derive"
 version = "1.0.35"
 source = "git+https://github.com/servo/serde?branch=deserialize_from_enums5#de4534b21f263752ed3b641c3c07e012574985bf"
@@ -1304,7 +1312,7 @@ dependencies = [
  "dwrote 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "euclid 0.17.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "freetype 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "freetype 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "fxhash 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "gleam 0.4.23 (registry+https://github.com/rust-lang/crates.io-index)",
  "glutin 0.12.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -395,7 +395,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "freetype"
-version = "0.3.0"
+version = "0.4.0"
```
jonleighton: This update happened automatically, I believe it is due to https://github.com/servo/webrender/commit/d16294236e0849ba0337815944628df78ec4741f

webrender_api/src/image.rs
```diff
@@ -110,11 +112,26 @@ impl ImageDescriptor {
...
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum ImageData {
-    Raw(Arc<Vec<u8>>),
-    Blob(BlobImageData),
+    Raw(#[serde(with = "serde_image_data_raw")] Arc<Vec<u8>>),
+    Blob(#[serde(with = "serde_bytes")] BlobImageData),
     External(ExternalImageData),
 }
 
+mod serde_image_data_raw {
```
jonleighton: This is necessary because we are dealing with an `Arc`. However, maybe it would be better to patch `serde_bytes` to support this?

Cargo.lock
Eijebong: Doesn't webrender have a CI check for that ?

jonleighton: I guess not? Though I don't know, I haven't contributed to webrender before. I've opened a separate PR to fix this: https://github.com/servo/webrender/pull/2599

webrender_api/src/image.rs
nox: Cc @dtolnay

dtolnay: Could you propose a more general signature for serde_bytes that could work here?

webrender_api/src/image.rs
```diff
@@ -110,11 +112,26 @@ impl ImageDescriptor {
...
+    extern crate serde_bytes;
+
+    use std::sync::Arc;
+    use serde::{Deserializer, Serializer};
+
+    pub fn serialize<'a, S: Serializer>(bytes: &'a Arc<Vec<u8>>, serializer: S) -> Result<S::Ok, S::Error> {
+        serde_bytes::serialize(&bytes.as_slice(), serializer)
```
dtolnay: There was a missing `T: ?Sized` in serde_bytes::serialize. I fixed it in 0.10.4 so the `&` on this line should no longer be necessary.

webrender_api/src/image.rs
jonleighton: Perhaps not, though I haven't looked into it very closely. Assuming a more general signature is not workable, maybe `serde_bytes` could also ship `serde_bytes_rc`/`serde_bytes_arc` modules though? (Essentially what I've done here, but inside the `serde_bytes` crate.)

jrmuizel:
It seems like this should only be needed when using ipc-channel. i.e. not
with Gecko

glennw:
@jrmuizel It looks like this would be used even in Gecko (due to the changes to FontData and the Blob structs), wouldn't it?

Is there any issue with having this enabled for Gecko?

Do we need to do anything further here (benchmarks?) or is this ready to merge?

glennw:
@jonleighton Thanks for the info! What does this mean in terms of the Gecko integration which doesn't use the ```ipc-channel``` feature in WR?

Does it have any effect at all? Should we be only enabling it in WR when the ```ipc-channel``` feature is enabled (e.g. does it increase the binary size for no performance win in that use case)?

jonleighton:
If the Gecko integration never serializes/deserializes, then yeah, I guess this code has no effect there.

However, wouldn't the compiler be able to figure out that the code is not used and exclude it from the binary? I'm a bit out of my depth here so I don't know if that is really the case, just struck me as something that might happen.

If that isn't the case, then should WebRender exclude the `derive(Deserialize, Serialize)` attributes when the `ipc` feature is not enabled?

jonleighton:
> If that isn't the case, then should WebRender exclude the derive(Deserialize, Serialize) attributes when the ipc feature is not enabled?

I.e. don't they also bloat the binary?

glennw:
Well, Gecko does serialize / deserialize things (such as the display list) - but they are sent over the Gecko IPC mechanism, rather than the rust ```ipc-channel``` crate.

So, I guess the question is does this change provide a benefit in that situation? I would think yes, but I just want to make sure that's the case, since @jrmuizel seems to suggest it's only useful if using ipc-channel, per the comment above.

jonleighton:
Right, so I also think the answer is yes since `serde_bytes` speeds up the serializing and deserializing of the data, irrespective of how the serialized data is then moved around.

glennw:
That makes sense to me. @jrmuizel Any issues with getting this merged?

jrmuizel:
Gecko doesn't serialize those structs in normal usage, only the display list. That being said, we probably do serialize them when doing a binary recording (we should probably only support this on Nightly)

I guess it's probably ok to take the extra dependency for now and have a feature flag to disable recording support in a follow up.

glennw:
Sounds good, thanks!

@bors-servo r+

PR #2588, opened 2018-03-29
Fix drop-shadows and remove last use of hardware_composite shader.

glennw:
This commit reworks how drop-shadows are handled. We now add two
primitives, each one referencing the same Picture.

The first primitive draws the drop shadow by blurring the alpha
channel of the picture, and generating screen-space UVs after
applying a local offset.

The second primitive draws the main image with a normal
brush_image shader blit.

There are a couple of hacky bits of code to make this work, but
it's such an improvement on the current state of drop-shadows
that I think it's worth getting this merged and tidying those
up later (mentioned as TODO comments in the patch).

Also, finally remove the hardware composite shader, since this
removes the last use of it.

Fixes #2584.
Fixes #2374.

PR #2600, opened 2018-04-02
Make stacking context clipping explicit

mrobinson:
Instead of relying on the combination of a stacking contexts which
render to intermediate surfaces and the PushStackingContext's clip node,
add an explicit clipping node parameter for stacking contexts. This is
necessary, because currently stacking context clipping is done by always
rendering to an intermediate surface. Once we support merging stacking
context and contained item clip chains we can remove this and the extra
parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2600)


mrobinson:
The Gecko try job for this change is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=653eda9276eb3ecaede897f2f98e6263a69ca4b5&selectedJob=171426186

Looks like there are some intermittent failures and one newly passing test.

glennw:

---

*[webrender/src/batch.rs, line 680 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97MC0N9Q_wvrEiexiI:-L97MC0N9Q_wvrEiexiJ:b-al01h0) ([raw file](https://github.com/servo/webrender/blob/bba5602b59e6c0a7a911779c5aa42c8ae69e66b9/webrender/src/batch.rs#L680)):*
> ```Rust
>                                 let source_id = cache_task_id;
> 
>                                 match picture.composite_mode.expect("bug: only composites here") {
> ```

This change will mean the catch-all _ at the bottom of this match executes if a bug occurs and we get here without a composite mode. Is there a specific reason to remove the expect here?

---

*[webrender/src/display_list_flattener.rs, line 1201 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97N7WB2lvxLSzZySEf:-L97N7WB2lvxLSzZySEg:b-ix8l1c) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/display_list_flattener.rs#L1201)):*
> ```Rust
>             frame_output_pipeline_id,
>             true,
>             clipping_node.is_some(),
> ```

We should perhaps test here (well, later probably) if the clipping node actually applies any useful clips? This would avoid doing redundant render work if the client supplies a clip node that has no effect.

If we did something like that, we'd probably check during the picture prepare each frame if the clip node applies to this Picture, and select at that point whether to use an intermediate surface.

We can do that as a follow up optimization later though - this will definitely let us do some testing in Gecko and make sure it works as we want the API to.

---

*[webrender/src/picture.rs, line 204 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97ORf73LoajuE4kvIu:-L97ORf73LoajuE4kvIv:bstaf0k) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/picture.rs#L204)):*
> ```Rust
> 
>         if self.can_draw_directly_to_parent_surface() {
>             if !self.force_intermediate_surface {
> ```

Should this bool check just be added to the can_draw_directly_to_parent_surface method?

---

*[webrender/src/picture.rs, line 373 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97OLUoE2hP6PqCF4Rz:-L97OLUoE2hP6PqCF4S-:b-iackb8) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/picture.rs#L373)):*
> ```Rust
>                 prim_screen_rect.clipped
>             }
>             None =>unreachable!("None case not handled by can_draw_directly_to_parent_surface."),
> ```

nit: space

---

*[webrender/src/prim_store.rs, line 1050 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97MXuw4-Vn7ZK-D_Jz:-L97MXuw4-Vn7ZK-D_K-:btykaq) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/prim_store.rs#L1050)):*
> ```Rust
>             apply_local_clip_rect,
>         );
>         picture.force_intermediate_surface = force_intermediate_surface;
> ```

Should we just push this through the new_image method rather than making it a mut variable here and setting it?

---

*[wrench/reftests/filters/blend-clipped.png, line 0 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97LhVBBwO8oreGNVTf:-L97LhVBBwO8oreGNVTg:b-g1xsyy) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/wrench/reftests/filters/blend-clipped.png#L0)):*
Is this just a change to the crop dimensions or was the content affected too?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2600)*

mrobinson:
Thanks for the review! I've answered you comments below and updated the PR.

---

---

*[webrender/src/batch.rs, line 680 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97MC0N9Q_wvrEiexiI:-L9ASzx92tBV3EPuz6wP:b3mmlv0) ([raw file](https://github.com/servo/webrender/blob/bba5602b59e6c0a7a911779c5aa42c8ae69e66b9/webrender/src/batch.rs#L680)):*

This change will mean the catch-all _ at the bottom of this match executes if a bug occurs and we get here without a composite mode. Is there a specific reason to remove the expect here?
</blockquote></details>

The issue is that now even Pictures without a composition mode might have a surface assigned. This code was expecting that to never happen. Unless there is a better way to handle this, I can maintain the runtime check is by adding an assertion like:

```rust
                                assert!(
                                    picture.composite_mode.is_some() ||
                                    picture.force_intermediate_surface
                                );

```

---

*[webrender/src/display_list_flattener.rs, line 1201 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97N7WB2lvxLSzZySEf:-L9AMeDk4THcBmNA2pQc:b21qrg5) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/display_list_flattener.rs#L1201)):*

We should perhaps test here (well, later probably) if the clipping node actually applies any useful clips? This would avoid doing redundant render work if the client supplies a clip node that has no effect.

If we did something like that, we'd probably check during the picture prepare each frame if the clip node applies to this Picture, and select at that point whether to use an intermediate surface.

We can do that as a follow up optimization later though - this will definitely let us do some testing in Gecko and make sure it works as we want the API to.


</blockquote></details>

That's correct. Right now the clips are applied twice, which in our rendering model could even affect the output. Eventually I think we will need to append the two clip chains on all items in the stacking context (while perhaps removing duplicate clips). In the end, this should remove the need to force the use of an intermediate surface.

 This may require us to revisit how clips are applied in WebRender, in general.

---

*[webrender/src/picture.rs, line 204 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97ORf73LoajuE4kvIu:-L9AQn---V8ZobT2UsSD:b-s45tni) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/picture.rs#L204)):*

Should this bool check just be added to the can_draw_directly_to_parent_surface method?
</blockquote></details>

Yes, this made the code quite a bit simpler!

---

*[webrender/src/picture.rs, line 373 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97OLUoE2hP6PqCF4Rz:-L9AQrNLCqgjMuSHaQC0:b-3n43fh) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/picture.rs#L373)):*

nit: space
</blockquote></details>

This should be fixed with the previous change,

---

*[webrender/src/prim_store.rs, line 1050 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97MXuw4-Vn7ZK-D_Jz:-L9ARwWR5iJFMk2QV_3b:b-896fix) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/prim_store.rs#L1050)):*

Should we just push this through the new_image method rather than making it a mut variable here and setting it?
</blockquote></details>

Done.

---

*[wrench/reftests/filters/blend-clipped.png, line  at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97LhVBBwO8oreGNVTf:-L9ARyfFBk0Amw-7LpiT:bs7bpag) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/wrench/reftests/filters/blend-clipped.png#L)):*

Is this just a change to the crop dimensions or was the content affected too?
</blockquote></details>

The content was affected here as well.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2600#-:-L9AYKDi3TMb950lyJmG:b4nk62c)*

glennw:
Review status: 20 of 24 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/batch.rs, line 680 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97MC0N9Q_wvrEiexiI:-L9CJUtm0Q2WKSrqv3Ea:btzjism) ([raw file](https://github.com/servo/webrender/blob/bba5602b59e6c0a7a911779c5aa42c8ae69e66b9/webrender/src/batch.rs#L680)):*

The issue is that now even Pictures without a composition mode might have a surface assigned. This code was expecting that to never happen. Unless there is a better way to handle this, I can maintain the runtime check is by adding an assertion like:

```rust
                                assert!(
                                    picture.composite_mode.is_some() ||
                                    picture.force_intermediate_surface
                                );

```
</blockquote></details>

Ah, I see. I inverted that logic a bit in https://github.com/servo/webrender/pull/2588/commits/6b63b208f5cdc03e0e97e2d38d195ab614029fd3, which is merging now. That will unfortunately conflict with your change, but perhaps the logic there might work a bit better with your change now (it inverts the composite mode / surface check). Or does that complicate things further?

---

*[webrender/src/display_list_flattener.rs, line 1201 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97N7WB2lvxLSzZySEf:-L9CJy8i3X4SqXKBCXeY:b-r7mxzb) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/display_list_flattener.rs#L1201)):*

That's correct. Right now the clips are applied twice, which in our rendering model could even affect the output. Eventually I think we will need to append the two clip chains on all items in the stacking context (while perhaps removing duplicate clips). In the end, this should remove the need to force the use of an intermediate surface.

 This may require us to revisit how clips are applied in WebRender, in general.
</blockquote></details>

In theory, it's possible to avoid having clips applied twice, by having a different clip chain applied to the stacking context from the one that is set on primitives within the stacking context, isn't it? I thought the idea was that the caller could then choose whether to specify a clip-chain for the stacking context AND/OR the primitive as they want to. Or is my conceptual model of how this works not right?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2600)*

glennw:
r=me btw, once rebased and the questions above are covered.

mrobinson:
Review status: 17 of 24 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/src/batch.rs, line 680 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97MC0N9Q_wvrEiexiI:-L9FzIsL9IZ9rWVjG1qY:b71bj2a) ([raw file](https://github.com/servo/webrender/blob/bba5602b59e6c0a7a911779c5aa42c8ae69e66b9/webrender/src/batch.rs#L680)):*

Ah, I see. I inverted that logic a bit in https://github.com/servo/webrender/pull/2588/commits/6b63b208f5cdc03e0e97e2d38d195ab614029fd3, which is merging now. That will unfortunately conflict with your change, but perhaps the logic there might work a bit better with your change now (it inverts the composite mode / surface check). Or does that complicate things further?
</blockquote></details>

I think it works a bit better with your change from https://github.com/servo/webrender/commit/6b63b208f5cdc03e0e97e2d38d195ab614029fd3. Hopefully we can remove this completely in the future

---

*[webrender/src/display_list_flattener.rs, line 1201 at r1](https://beta.reviewable.io/reviews/servo/webrender/2600#-L97N7WB2lvxLSzZySEf:-L9FzZFy6bRoaSFzfGrk:bep7o13) ([raw file](https://github.com/servo/webrender/blob/8e980971db204b471b634d152f88348bc693ce02/webrender/src/display_list_flattener.rs#L1201)):*

In theory, it's possible to avoid having clips applied twice, by having a different clip chain applied to the stacking context from the one that is set on primitives within the stacking context, isn't it? I thought the idea was that the caller could then choose whether to specify a clip-chain for the stacking context AND/OR the primitive as they want to. Or is my conceptual model of how this works not right?
</blockquote></details>

Oh I definitely want to preserve the functionality where when the stacking context and primitives are assigned different clip chains they both clip the primitive. This is how things work after this change. 

What I mean by avoiding double clipping is to handle the situation where, for instance, the stacking context clip chain is a subset of the primitive clip chain. In this example, the stacking context clip chain would be ignored completely. It would be great to have this de-duplication done on a per-clip basis while processing primitive runs.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2600)*

glennw:
@staktrace Heads up for public API change. Passing ```None``` for the new parameter to ```push_stacking_context``` will retain current behaviour.

PR #2602, opened 2018-04-03
Disable segment path for Pictures, we don't support it yet.

glennw:
We need a bit of extra work in batch.rs before pictures will be
correctly added when they have a valid segmentation. For now,
just disable segment building for pictures until we do that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2602)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -237,11 +237,14 @@ impl BrushKind {
     fn supports_segments(&self) -> bool {
         match *self {
             BrushKind::Solid { .. } |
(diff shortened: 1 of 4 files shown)


glennw:
r? @kvark or anyone

Pending try here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=edf3fa081a23ae84eb92977c6fdf6b7f5884a185

glennw:
This is a fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1445723.

glennw:
Try run looks good.

PR #2604, opened 2018-04-04
Fix AA on ellipse clip code under some edge cases.

### LINKED ISSUE #2576: Incorrect rendering in some cases with new AA method.
In some cases, the new AA method from https://github.com/servo/webrender/pull/2445 is producing incorrect rendering.

For example, compare these two images:

bad
good

The first image is using the new technique - the values in the center of the clip mask are 243 / 255, which is why they are considered partially transparent and (incorrectly) blend with the box shadow below.

I locally reverted #2445 to generate the second image - in this case, it produces an opaque clip region in the center of the primitive, so it draws correctly without any translucency.

These images were generated from loading the following test case in wrench:

```
---
root:
  items:
    - type: stacking-context
      transform: rotate-z(45) rotate-x(60)
      transform-origin: 300 300
      items:
        - type: box-shadow
          bounds: [0, 0, 80, 80]
          color: [100, 100, 100]
          offset: [30, 30]
          border-radius: 8

        - type: clip
          bounds: [0, 0, 80, 80]
          complex:
            - rect: [0, 0, 80, 80]
              radius: 8
          items:
            - type: rect
              color: [100, 150, 200]
              bounds: [0, 0, 80, 80]

        - type: clip
          bounds: [90, 0, 80, 80]
          complex:
            - rect: [90, 0, 80, 80]
              radius: 8
          items:
          - type: rect
            color: [100, 150, 200]
            border-radius: 8
            bounds: [90, 0, 80, 80]
```

glennw:
If the AA range is > 1.0, due to a specific transform, the inner
pixels of a primitive that are not affected by the elliptical
radii in a clip mask may end up with a signed distance of 1.0,
which causes blending artifacts.

Instead, initialize the default value for signed distance to
the calculated AA range, which ensures we get a blend factor
of 1.0 in such cases.

Fixes #2576.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2604)

### DIFF
--- a/webrender/res/ellipse.glsl
+++ b/webrender/res/ellipse.glsl
@@ -58,7 +58,7 @@ float rounded_rect(vec2 pos,
     // Start with a negative value (means "inside") for all fragments that are not
     // in a corner. If the fragment is in a corner, one of the clip_against_ellipse_if_needed
     // calls below will update it.
-    float current_distance = -1.0;
+    float current_distance = -aa_range;
 
     // Clip against each ellipse.
     current_distance = clip_against_ellipse_if_needed(pos,
--- a/wrench/reftests/aa/aa-dist-bug-ref.yaml
+++ b/wrench/reftests/aa/aa-dist-bug-ref.yaml
@@ -1,0 +1,6 @@
+---
+root:
+  items:
+    - type: rect
+      color: red
+      bounds: [340, 184, 50, 20]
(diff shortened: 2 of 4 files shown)


glennw:
r? @kvark or anyone.

I'll kick off a gecko try shortly.

glennw:
Pending Gecko try:

https://hg.mozilla.org/try/rev/83ce3728be2447721d56a53829de1ad0089987c7
https://treeherder.mozilla.org/#/jobs?repo=try&revision=83ce3728be2447721d56a53829de1ad0089987c7

glennw:
Try run looks good.

PR #2605, opened 2018-04-04
Update glutin and winit.

waywardmonkeys:
This removes a number of dependencies that had 2 versions being
built, like core-foundation and core-graphics.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2605)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,7 +13,7 @@ dependencies = [
 
 [[package]]
 name = "android_glue"
-version = "0.2.1"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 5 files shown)


glennw:
Thanks!

Looks like ```servo-tidy.toml``` needs to be updated:

```
./Cargo.lock:1: duplicates for `gdi32-sys` are allowed, but only single version found
	The following packages depend on version 0.2.0 from 'crates.io':
		dwrote
		font-loader
./Cargo.lock:1: duplicates for `gl_generator` are allowed, but only single version found
	The following packages depend on version 0.9.0 from 'crates.io':
		gleam
		glutin
		mozangle
./Cargo.lock:1: duplicates for `core-foundation` are allowed, but only single version found
	The following packages depend on version 0.5.1 from 'crates.io':
		core-graphics
		core-text
		font-loader
		glutin
		webrender
		webrender_api
		winit
		wrench
./Cargo.lock:1: duplicates for `core-foundation-sys` are allowed, but only single version found
	The following packages depend on version 0.5.1 from 'crates.io':
```

waywardmonkeys:
@glennw Updated the PR ... waiting on the checks now. :)

glennw:
CI looks good, thanks!

@bors-servo r+

Issue #2501, opened 2018-03-09
Clip masks + snapping


glennw:
When primitives are drawn, we snap them to the nearest device pixel. However, we don't do the same with clip masks.

Consider the following reftest:

```
---
root:
  items:
    - type: clip
      bounds: [0, 0, 1000, 1000]
      complex:
        - rect: [50.3, 50.3, 100, 100]
          radius: 16
      items:
      - type: rect
        bounds: 50.3 50.3 100 100
        color: red

    - type: rect
      bounds: 200.3 50.3 100 100
      color: green
```

And the reference:

```
---
root:
  items:
    - type: clip
      bounds: [0, 0, 1000, 1000]
      complex:
        - rect: [50, 50, 100, 100]
          radius: 16
      items:
      - type: rect
        bounds: 50 50 100 100
        color: red

    - type: rect
      bounds: 200 50 100 100
      color: green
```

In this case, the green rect is exactly the same in both cases, because the ```0.3``` gets snapped to a device pixel. However, the red rect is different - the primitive gets snapped to a device pixel, but the clip mask doesn't.

![diff](https://user-images.githubusercontent.com/2350882/37192615-70bbe028-235e-11e8-8479-5dd4926fe0f2.png)

glennw:
It seems like we need to make the clip mask generation take into account clipping of the primitive vertices as we draw the mask. @kvark Does that sound right to you, and does that seem fairly simple to do (given then we draw the clip masks with a screen-aligned quad)?

kvark:
@glennw yes, that appears to be a good course of action

glennw:
@kvark Do you have any thoughts on the best way to implement this, given that for clip masks the vertices are the screen space bounds of the task rect?

glennw:
Prototype PR that fixes this is here https://github.com/servo/webrender/pull/2608#issuecomment-378518162.

Issue #2595, opened 2018-03-30
dark outline around links on cnet.com


nical:
On this page: https://www.cnet.com/news/mozilla-open-source-firefox-move-helped-rewrite-tech-rules-anniversary/ the links have some sort of thin dark outline around them which looks like a sampling artifact. It was reproduced on linux and mac in the latest firefox nightly.

glennw:
Here is a minimal repro:

```
---
root:
  items:
    - type: box-shadow
      bounds: [ 56.4833374023438, 50, 81, 25 ]
      color: black
      clip-mode: inset
      offset: 0 -4
```

I haven't confirmed, but I believe this will be an instance of this bug https://github.com/servo/webrender/issues/2501.

The bounds of the shadow are on a sub-pixel boundary, which is getting snapped to a device pixel. However, as per the bug above, the clip mask generation doesn't correctly deal with device snapping in this case.

Note to self: if this was being segmented, the problem would disappear - the segment builder logic eliminates the need for a clip mask, resulting in correct rendering. In this case we don't select segment building, since the area of the primitive is too small - we might want to revisit this heuristic, since in this case segmenting is probably worth it, to save a clip mask being generated. We still need to fix the clip snapping logic above though.

glennw:
This will be fixed when #2608 lands.

PR #2608, opened 2018-04-04
Apply device pixel snapping to axis-aligned clip masks.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2608)


glennw:
This is a quick hack that attempts to apply device pixel snapping to clip masks. It makes the reftest in https://github.com/servo/webrender/issues/2501 pass.

The try run with this change is:
https://hg.mozilla.org/try/rev/4185c56c80f49bbae569c3e0d07a284640755e8b
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4185c56c80f49bbae569c3e0d07a284640755e8b&selectedJob=171784344

Of the relevant reftests, there are two new PASS results, and one FAIL. It may be reasonable to fuzz the failure (max difference is 7, looks visually the same).

@kvark @staktrace @mrobinson  I wanted to get some feedback on (a) if this approach seems correct (b) if there's a more obvious / efficient way to achieve this and (c) if fuzzing that test seems reasonable.

If we *do* think this is the right approach, I'll tidy up the shader code in the morning, and then this would be ready to go.

webrender/res/clip_shared.glsl
```diff
@@ -51,7 +51,23 @@ RectWithSize intersect_rect(RectWithSize a, RectWithSize b) {
...
+    vec2 device_pos = area.screen_origin + aPosition.xy * area.common_data.task_rect.size;
+    vec2 actual_pos;
+
+    // =============================
+    if (scroll_node.is_axis_aligned) {
+        vec4 world_p0 = scroll_node.transform * vec4(local_clip_rect.p0, 0.0, 1.0);
+        vec4 world_p1 = scroll_node.transform * vec4(local_clip_rect.p0 + local_clip_rect.size, 0.0, 1.0);
+        vec4 world_pos = uDevicePixelRatio * vec4(world_p0.xy, world_p1.xy) /
+                                             vec4(world_p0.ww, world_p1.ww);
+        vec4 snap_offsets = floor(world_pos + 0.5) - world_pos;
```
kvark: could we move this out into a helper method, shared with other write_vertex functions?

webrender/res/clip_shared.glsl
```diff
@@ -51,7 +51,23 @@ RectWithSize intersect_rect(RectWithSize a, RectWithSize b) {
...
+        vec2 normalized_snap_pos = (device_pos - world_pos.xy) / (world_pos.zw - world_pos.xy);
+        vec2 ref_pos = mix(snap_offsets.xy, snap_offsets.zw, normalized_snap_pos);
```
kvark: there appears to be no guarantee that normalized_snap_pos is between 0 and 1, which would lead to extra big snap offsets

webrender/res/clip_shared.glsl
```diff
@@ -51,7 +51,23 @@ RectWithSize intersect_rect(RectWithSize a, RectWithSize b) {
 ClipVertexInfo write_clip_tile_vertex(RectWithSize local_clip_rect,
                                       ClipScrollNode scroll_node,
                                       ClipArea area) {
-    vec2 actual_pos = area.screen_origin + aPosition.xy * area.common_data.task_rect.size;
+    vec2 device_pos = area.screen_origin + aPosition.xy * area.common_data.task_rect.size;
+    vec2 actual_pos;
+
+    // =============================
```
kvark: is this decoration intended?

webrender/res/clip_shared.glsl
glennw: I split the existing ```compute_snap_offset``` function into helper functions, and used those here.

webrender/res/clip_shared.glsl
glennw: That's actually by design. Since we are providing values at the vertex corners of the clip region, if they go outside that value, they should (I *think!*) be the correct values once interpolated to get the right snap for a local rect inside that region.

webrender/res/clip_shared.glsl
glennw: Nope, was just part of a WIP hack :)

glennw:
@kvark OK, this is ready for re-review. I'll do another try run before merging, just to make sure it's not broken somehow. Also still need to check with @staktrace if fuzzing that test is OK.

glennw:
Pending try with updated PR:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=737bb8e3340887ea2e1731e5e7e8a4ab11c5969c

glennw:
This fixes #2501 and also #2595.

glennw:
Rebased

glennw:
Latest try run appears to still be good (same results).

glennw:
Closing in favor of #2618 which includes this commit.

PR #2613, opened 2018-04-04
Add hooks for APZ to use the scene builder thread

staktrace:
r? @nical

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2613)


staktrace:
In poke() we basically process IPC messages that came from content-side gecko but that we need to handle on this thread. For example if we get a focus state update from content-side gecko, we dispatch it to the updater thread at [1] and then it gets run during the poke at [2], assuming that the scene it depends on has been built.

[1] https://hg.mozilla.org/try/file/0dfe1478092b32844b9d6ed5010bf31bf3008702/gfx/layers/apz/src/APZUpdater.cpp#l146
[2] https://hg.mozilla.org/try/file/0dfe1478092b32844b9d6ed5010bf31bf3008702/gfx/layers/apz/src/APZUpdater.cpp#l114

glennw:
@bors-servo r=nical

@nical fyi, bors doesn't look at notifications written in the GH review box.

PR #2616, opened 2018-04-05
Remove force_intermediate_surface.

glennw:
We can slightly simplify the picture code by forcing the composite
mode to Blit when there is a clip node for stacking contexts.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2616)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -664,12 +664,6 @@ impl AlphaBatchBuilder {
                             return;
                         }
 
-                        //assert!(
-                        //    picture.composite_mode.is_some() ||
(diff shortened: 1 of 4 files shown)


webrender/src/batch.rs
```diff
@@ -664,12 +664,6 @@ impl AlphaBatchBuilder {
                             return;
                         }
 
-                        //assert!(
-                        //    picture.composite_mode.is_some() ||
-                        //    picture.force_intermediate_surface
-                        //);
```
mrobinson: Ack. Sorry about this.

PR #2618, opened 2018-04-05
Support snapping on clip masks, and fast path rectangle + clip out mode.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2618)


glennw:
This supersedes #2608 (it includes that commit).

This PR achieves two things:
 * Support snapping of axis-aligned clip masks (mostly reviewed in #2608).
 * Native Rectangle + ClipOut mode (an optimization, and accuracy win).

This allows primitives that only contain clip rects to avoid generating
a clip mask, and also fixes a number of reftests.

Fixes #2595.
Fixes #2501.

Try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=89310dbe1725653477a524a69804cd34dcb084e1

Of the relevant reftest differences:

R4 - 1 new PASS
R6 - 1 new PASS
R7 - 1 new PASS, 1 new FAIL (agreed to fuzz https://github.com/servo/webrender/pull/2608#issuecomment-378748930)

r? @mrobinson or @kvark

glennw:

---

*[webrender/res/clip_shared.glsl, line 85 at r1](https://beta.reviewable.io/reviews/servo/webrender/2618#-L9M6Gz77ZfAjuTCQp7M:-L9MYbCP9NSxWkC0B9oX:b5qlt3p) ([raw file](https://github.com/servo/webrender/blob/400f99eaecdec4107058b4ad3fe45aebb3ade4a1/webrender/res/clip_shared.glsl#L85)):*

I don't understand this part. When we snap, we want the final VS results to be snapped. But this code uses the pre-snapped device pos
</blockquote></details>

Because in the clip shaders, the vertices are the flat coordinates of the corners of the 2d bounding rect of the clip task. So I don't see how / why we'd be able to apply snapping here? What this PR does is keep the original vertex positions the same as the task rect, but adjust the interpolants for the local position at each corner vertex such that the snapping is accounted for. At least, that's the intent :) Does that sound reasonable?

---

*[webrender/src/prim_store.rs, line 1576 at r1](https://beta.reviewable.io/reviews/servo/webrender/2618#-L9M7FukEhp3nEZr1jDJ:-L9MYxGu4hLtwb3Yv6uc:b41n6ih) ([raw file](https://github.com/servo/webrender/blob/400f99eaecdec4107058b4ad3fe45aebb3ade4a1/webrender/src/prim_store.rs#L1576)):*

why are we segmenting a small but rect-only primitive?
</blockquote></details>

Because rect only primitives can be efficiently segmented into CPU regions and thus completely avoid any clip-mask at all. For example, I intend to draw simple borders as a rectangle with a clip + clip-out rectangle (got a local patch nearly ready for this). This means that for the most common border types we don't invoke the border shader at all, and don't allocate / render any clip masks in the case of no rounded corners.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2618)*

glennw:
Rebased.

kvark:
:lgtm:

---

Reviewed 4 of 4 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2618#-:-L9NuauW3ZbGnYBVcpCZ:bnfp4nl)*

kvark:
Thank you!
@bors-servo r+

PR #2619, opened 2018-04-05
Support sampling from the render task cache in brush images.

nical:
This PR adds support for sampling the render task cache in brush images, and test it out adds support for image sub rects in brush_image.

The sub rects implementation is duplicated between the brush and non-brush primitives, but I think it's fine to live with this temporarily because if things go well the non-brush code path will go away soon and the way I am planning to support pre-rendering tile spacing will make the two code paths diverge.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2619)


nical:
This is related to the work in #2572 but since I am under the impression that the latter will take forever to land, I'm trying to separate whatever I can into independent pull requests.

PR #2622, opened 2018-04-05
Readable shader error messages

kvark:
Working with shaders, I find `#line` navigation experience to be rather poor:
  1. first, I take an `apitrace` capture
  2. load the capture, go to the last `glShaderSource`
  3. copy out the input into a new file
  4. for each `#line` directive, try to match the reported error number ...

This PR removes line overrides. Instead, it prints the relevant piece to the console, e.g.:
```
Loading shaders...
Failed to compile shader: cs_clip_image
0:1152(17): error: syntax error, unexpected NEW_IDENTIFIER, expecting ',' or ';'

|	    ClipScrollNode scroll_node = fetch_clip_scroll_node(cmi.scroll_node_id);
>	    DImageMaskData mask = fetch_mask_data(cmi.clip_data_address);
|	    RectWithSize local_rect = mask.local_rect;
```

This information instantly gives me an idea of what's wrong (note: unlike the original driver message), and makes all the listed steps unnecessary for 99% of cases. And when the steps are necessary, we don't have to check for each `#line` occurrence since the driver-reported line number is no longer ambiguous.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2622)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -53,7 +53,6 @@ const SHADER_VERSION_GLES: &str = "#version 300 es\n";
 const SHADER_KIND_VERTEX: &str = "#define WR_VERTEX_SHADER\n";
 const SHADER_KIND_FRAGMENT: &str = "#define WR_FRAGMENT_SHADER\n";
 const SHADER_IMPORT: &str = "#include ";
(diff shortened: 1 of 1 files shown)


glennw:
Nice!

@bors-servo r+

PR #2623, opened 2018-04-05
Separate primitive shaders from clip shaders

kvark:
Includes #2622

Note: this is not done because I'm bored.
The problem Szeged team has bumped into is that `aDataX` attributes end up being used in the shaders they don't relate to (e.g. clip shaders). This isn't affecting GLSL but it does affect GLSL -> SPIRV -> XXX shader pipelines.

The solution I came up with involves moving parts of `prim_shared` into separate shared modules, and turning `clip_shared` into a catch-all header of clip shades.

Downsides: more shader modules (but not programs) to navigate between.
Any suggestions on alternative solutions are welcome ;)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2623)

### DIFF
--- a/webrender/res/base.glsl
+++ b/webrender/res/base.glsl
@@ -33,6 +33,5 @@
 
 #ifdef WR_FRAGMENT_SHADER
     precision highp float;
-
     #define varying in
 #endif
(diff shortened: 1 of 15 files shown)


nical:
Sounds good to me, could you document the constraints on the driver issues wiki page? It's not really a driver issue per se but it's in the bag if platform/backend/configuration specific issues that we don't necessarily catch locally.

PR #2626, opened 2018-04-06
Optimization: draw solid, single color borders as rect + clip.

glennw:
Due to various recent optimizations and features added for the
segment logic, it's now feasible to start drawing borders as
a solid rect + clip mask.

For now, we start with only the simplest and most common case,
borders with style: solid and a single color.

This optimization applies in a couple of ways:
 * Fewer shader switches and better batching. Specifically, we
   can often batch these simple borders into an existing solid
   rect batch. Additionally, the edge segments of borders are
   typically considered opaque segments, which are more efficient
   to batch, and help with early z-reject of other fragments.

 * Once we start to cache clip masks between elements, and also
   across frames / display lists, we'll automatically get some
   wins here by only rendering a border corner mask once where
   applicable. Many real world sites use a common border radius
   on all buttons, for example, so this can be a significant
   optimization in these cases.


glennw:
r? @kvark 

This includes the 2 commits from #2618, it might be easier to review this after that's landed - only the last commit here is new.

Try run is here:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=2d090150fea8692f35a4cbc961076d9a801f8930&selectedJob=172218279

Same results as in #2618 plus 1 new PASS in addition to those.

glennw:
Rebased on master so there is just the one commit now.

kvark:
> Once we start to cache clip masks between elements

I have a strong dejavu feeling about this :)

:lgtm_strong: 

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2626#-:-L9Qkraj-rLr4yejkSFe:b-54clxn)*

PR #2629, opened 2018-04-06
Remove overscroll code and api.

### LINKED ISSUE #1314: Make it possible to configure overscroll behavior
*This issue is part of an effort to improve Servo's embedding story. See https://github.com/paulrouget/servoshell/projects/2*

From the embedder perspective, we want to be able to configure and change the overscroll behavior.

Maybe something like this:

```rust
pub struct ViewOverscrollBehavior {
    top: OverscrollBehavior,
    right: OverscrollBehavior,
    bottom: OverscrollBehavior,
    left: OverscrollBehavior,
}

pub enum OverscrollBehavior {
    Enabled(OverscrollParameters),
    Disabled,
}

pub struct OverscrollParameters {
  minPanDistanceRatio: f32,
  springFiction: f32,
  springStiffness: f32,
  stopDistanceThreshold: f32,
  stopVelocityThreshold: f32,
}
```

### LINKED ISSUE #1762: Make overscrolling a runtime option and add tests for it
This will allow us to test overscrolling, which has no tests currently.

### LINKED ISSUE #2545: Remove the overscroll code
By overscroll, I mean the bouncing effect when scrolling beyond the borders of the page in https://github.com/servo/webrender/blob/master/webrender/src/clip_scroll_node.rs

We are implementing this at the embedder level. For example, on Android, we will be using an OverScroller and draw the overscroll effect directly on top of Servo.

glennw:
Fixes #2545.
Fixes #1762.
Fixes #1314.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2629)


glennw:
This is better handled at a higher level by the embedder / client, as per #2545.

I'm not sure if I removed too much - I wasn't quite sure if we still need the ScrollEventPhase?

Long term, we should probably switch Servo to use the API that Gecko uses to scroll by a specific scroll node, and remove the other scroll API, I think?

r? @mrobinson

mrobinson:
I wonder if we can also remove `DisplayListFlattener::setup_viewport_offset` and related API as well.

PR #2630, opened 2018-04-06
Border edge checks

kvark:
This change prevents zero-sized edges from being produced by the frame builder and rendered.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2630)


glennw:
I guess we'll need a try run for this.

PR #2631, opened 2018-04-07
appveyor: download latest stable rust with rustup

UK992:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2631)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,10 @@
 environment:
-  PATH: 'C:\msys64\mingw64\bin;C:\msys64\usr\bin;%PATH%;C:\Rust\bin'
+  PATH: '%PATH%;C:\Users\appveyor\.cargo\bin'
   RUST_BACKTRACE: 1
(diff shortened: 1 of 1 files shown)


glennw:
I'm not sure this is what we want. We want to specifically track the most recent rustc that is allowed for Gecko builds, which is our minimal target. This seems like it would get rustc versions ahead of that?

staktrace:
Note that the taskcluster Linux CI also uses the most recent stable rustc.
The taskcluster mac CI uses whatever we installed - rustc has to be
manually updated there. While using gecko's minimal rustc would be ideal I
don't know of an API to get that info, but I haven't looked either.

glennw:
In the past we've just been updating the travis.yml manually with the current Gecko rustc version each time it changes (6 weeks, I think?).

glennw:
Closing this, since I think we'll want to retain the ability to pin to a specific rustc version. If our current rustc version is behind the gecko minimum rustc, we could update that though.

PR #2633, opened 2018-04-09
Remove unused glyph count field.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2633)


Gankro:
Oh nice, that field's existence sucked! (how are we reserving buffer sizes now..?)

PR #2634, opened 2018-04-09
Bump serde to 1.0.37

Eijebong:


staktrace:
This is going to add a third copy of syn into mozilla-central. We already have syn 0.11 and syn 0.12 and this adds 0.13 as well. :(

Eijebong:
@staktrace The goal of this is to remove syn 0.11/0.12 in servo

staktrace:
@Eijebong yeah but it would be good to coordinate these changes with gecko as well. For example if we bumped xpcom-macros from syn 0.11 to syn 0.13 (per https://bugzilla.mozilla.org/show_bug.cgi?id=1429816#c8) before landing this it would have made things a little smoother by avoiding the intermediate state with three copies of syn. We can still do it but it means now the next WR update into gecko is blocked until that happens.

staktrace:
@Eijebong I don't think anybody is actively working on it, I don't think @mystor would mind if you stole it.

PR #2635, opened 2018-04-09
Fix and run the rawtests on MacOS

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2635)


jrmuizel:
It looks like macOS has the same display problem as Linux

glennw:
@jrmuizel What's the status of this one?

jrmuizel:
I'm going to close this. We'll figure something out once the Mac CI problems are understood.

PR #2639, opened 2018-04-09
Switch to a serial osmesa build to fix mac CI issues

jrmuizel:
Something like this worked for me on my last update to #2635

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2639)

### DIFF
--- a/wrench/Cargo.toml
+++ b/wrench/Cargo.toml
@@ -24,7 +24,7 @@ ron = "0.1.5"
 time = "0.1"
 crossbeam = "0.2"
 osmesa-sys = { version = "0.1.2", optional = true }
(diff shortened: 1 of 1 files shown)


glennw:
What does a serial build change / mean?

jrmuizel:
We build osmesa with -j1. (I really could have been clearer :))

jrmuizel:
This seems to have worked on this PR as well. I suggest we take this for now and try to figure out why it doesn't work anymore after.

glennw:
@jrmuizel Sounds good, thanks!

I'm still not entirely clear on what has changed recently that could cause this?

@bors-servo r+

PR #2640, opened 2018-04-10
Set the dirty rect to the full image when we create it.

jrmuizel:
Previously in the following situation:
 add_image
 update_image(small dirty rect)
 draw_image()

We'd resolve the blob image with small dirty rect without having
done it for the whole area first.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2640)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -4,7 +4,7 @@
 
 use api::{AddFont, BlobImageData, BlobImageResources, ResourceUpdate, ResourceUpdates};
 use api::{BlobImageDescriptor, BlobImageError, BlobImageRenderer, BlobImageRequest};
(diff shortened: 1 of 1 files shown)


glennw:
CI compile error:

```
error[E0433]: failed to resolve. Use of undeclared type or module `DeviceUintPoint`
   --> src\resource_cache.rs:485:50
    |
485 |             dirty_rect: Some(DeviceUintRect::new(DeviceUintPoint::zero(),
    |                                                  ^^^^^^^^^^^^^^^ Use of undeclared type or module `DeviceUintPoint`
```

staktrace:
@jrmuizel This caused a crashtest failure in gecko CI. https://bugzilla.mozilla.org/show_bug.cgi?id=1452603#c4

PR #2641, opened 2018-04-11
Use a freelist to store render task cache entries, make cache entry public.

glennw:
This is a small patch in preparation for upcoming changes that are
related to caching render tasks. Storing these as a freelist allows
handing out weak handles, rather than always fetching by hash key.

We can provide these to the caller of request_render_task() to allow
fetching the CacheItem during batching, without another hash map
lookup. This is more in line with how the PathFinder integration
works, and also useful for some upcoming patches which will cache
other render task kinds. Specifically, in addition to the CacheItem
for a cached render task, the batching code will be able to retrieve
other details about a cached render task, such as required blend mode.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2641)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -798,7 +798,7 @@ dependencies = [
 [[package]]
 name = "osmesa-src"
 version = "17.4.0-devel"
-source = "git+https://github.com/servo/osmesa-src#1cf8c2f21c7e9047c3bd5a2f1accbc08e5636cf7"
+source = "git+https://github.com/jrmuizel/osmesa-src?branch=serialize#0f81d082b4cca16ea4d344bf6758b5a2f8d98277"
 
 [[package]]
 name = "osmesa-sys"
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -798,7 +798,7 @@ dependencies = [
 [[package]]
 name = "osmesa-src"
 version = "17.4.0-devel"
-source = "git+https://github.com/servo/osmesa-src#1cf8c2f21c7e9047c3bd5a2f1accbc08e5636cf7"
+source = "git+https://github.com/jrmuizel/osmesa-src?branch=serialize#0f81d082b4cca16ea4d344bf6758b5a2f8d98277"
```
kvark: why are we changing the osmesa path here?

glennw: It's from https://github.com/servo/webrender/pull/2639 - some CI weirdness that @jrmuizel needed temporarily.

glennw:
Yes, handing out weak handles is a follow up to this :)

kvark:
r=me after the CI is green:
> error(B: type alias is never used: `RenderTaskCacheEntryHandle`(B

glennw:
Fixed the warning, thanks.

@bors-servo r=kvark

PR #2642, opened 2018-04-11
Improve border dot handling for small radii

mrobinson:
When rendering dotted borders with small radii, put a single dot in the
corner and render the edge based on that. This fixes an issue where dots
smoosh into each other with small radii.

The resulting rendering isn't perfectly aligned for the moment, but this
can be improved as we handle more complicated situations such as
combinations of borders of different widths.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2642)


mrobinson:
Here is the try job for this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=3694b2bf17f2768425d50363665bfe8d4f3fb0f4

kvark:
wow, quite a few failures, it seems?

mrobinson:
@kvark The odd thing is that the failures are all on Windows and many in tests that are not related at all to borders (for instance in SVG only tests). I'm wondering if there is something funky with that revision of Gecko. I will kick off a new job with a newer checkout.

mrobinson:
I had made some fixes for ANGLE validation errors and had neglected to integrate them into my Gecko branch. This try job looks a lot better: https://treeherder.mozilla.org/#/jobs?repo=try&revision=c5c38aeb0f49b4d4db24d832f82091c4e1d08e36&selectedJob=173228446

glennw:
Yup, try run looks good now.

@bors-servo r+ :rocket:

PR #2615, opened 2018-04-05
[WIP] Round up the stretch size

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1401665
Pending try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e138e9e5aa30bd5f1c40501c990a73dc056a60fe

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2615)

PR #2644, opened 2018-04-11
Use the brush shaders for repeated images/gradient in some cases.

nical:
Built on top of #2621, this PR makes use of the repetition support added in the brush shaders in the simple case (no tile spacing).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2644)


kvark:
:lgtm: minus a few small concerns

---

---

*[webrender/res/brush_image.glsl, line 150 at r1](https://beta.reviewable.io/reviews/servo/webrender/2644#-L9paGqX16twQONda7Uf:-L9paGqX16twQONda7Ug:b-624uao) ([raw file](https://github.com/servo/webrender/blob/df9ef5e1e46d18662636214509cbb2f29c455e46/webrender/res/brush_image.glsl#L150)):*
> ```GLSL
>     vUv.xy = mix(uv0, uv1, f) - min_uv;
>     vUv.xy /= texture_size;
>     vUv.xy *= repeat.xy;
> ```

shouldn't this be something like `vUv.xy += repeat.xy * (max_uv-min_uv) / texture_size`?

---

*[webrender/res/brush_radial_gradient.glsl, line 61 at r1](https://beta.reviewable.io/reviews/servo/webrender/2644#-L9pdNr_1QepPFYFc-vp:-L9pdNr_1QepPFYFc-vq:bu7exjd) ([raw file](https://github.com/servo/webrender/blob/df9ef5e1e46d18662636214509cbb2f29c455e46/webrender/res/brush_radial_gradient.glsl#L61)):*
> ```GLSL
>     vRepeatedSize.y *=  ratio_xy;
> 
>     vPos;
> ```

is that an accidental leftover?

---

*[webrender/res/brush_radial_gradient.glsl, line 80 at r1](https://beta.reviewable.io/reviews/servo/webrender/2644#-L9pdSIG5eEaz9RQhftt:-L9pdSIG5eEaz9RQhftu:b-mydqic) ([raw file](https://github.com/servo/webrender/blob/df9ef5e1e46d18662636214509cbb2f29c455e46/webrender/res/brush_radial_gradient.glsl#L80)):*
> ```GLSL
> #ifdef WR_FEATURE_ALPHA_PASS
>     // Handle top and left inflated edges (see brush_image).
>     vec2 local_pos = max(vPos, vec2(0.0));
> ```

This piece is sort of repeated 3 times now. Any plans to unify it? At the very least, could put it into an included module

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2644#-:-L9peWZk2DRwo3HtiDvR:b-6bki0e)*

nical:

---

*[webrender/res/brush_image.glsl, line 150 at r1](https://beta.reviewable.io/reviews/servo/webrender/2644#-L9paGqX16twQONda7Uf:-L9pmp9PDhrB_oE2Nt9t:b-k38wzs) ([raw file](https://github.com/servo/webrender/blob/df9ef5e1e46d18662636214509cbb2f29c455e46/webrender/res/brush_image.glsl#L150)):*

shouldn't this be something like `vUv.xy += repeat.xy * (max_uv-min_uv) / texture_size`?
</blockquote></details>

it's really a multiplication. Basically multiplying by repeat stretches the rectange, which is then chopped off by the modulo, whereas an addition would translate it.
Sorry I am not super good at explaining this type of thing without waving hands and drawing on pieces of papers, but we can chat about it on vidyo or something if you want to go into the details.

---

*[webrender/res/brush_radial_gradient.glsl, line 61 at r1](https://beta.reviewable.io/reviews/servo/webrender/2644#-L9pdNr_1QepPFYFc-vp:-L9pn0d_Cb_RzwsGM3w1:b-834n6v) ([raw file](https://github.com/servo/webrender/blob/df9ef5e1e46d18662636214509cbb2f29c455e46/webrender/res/brush_radial_gradient.glsl#L61)):*

is that an accidental leftover?
</blockquote></details>

Oops!

---

*[webrender/res/brush_radial_gradient.glsl, line 80 at r1](https://beta.reviewable.io/reviews/servo/webrender/2644#-L9pdSIG5eEaz9RQhftt:-L9pn5D16CxmDMuXFAau:b5zk0vn) ([raw file](https://github.com/servo/webrender/blob/df9ef5e1e46d18662636214509cbb2f29c455e46/webrender/res/brush_radial_gradient.glsl#L80)):*

This piece is sort of repeated 3 times now. Any plans to unify it? At the very least, could put it into an included module
</blockquote></details>

Sorta. I don't think that we can use the same code for the image and gradient shaders, but we could look into factoring the code for the two gradient shaders (which are almost the same) when the dust settles.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2644)*

kvark:

---

*[webrender/res/brush_image.glsl, line 150 at r1](https://beta.reviewable.io/reviews/servo/webrender/2644#-L9paGqX16twQONda7Uf:-L9poapSB1gFISGedSAO:b-9cs2bb) ([raw file](https://github.com/servo/webrender/blob/df9ef5e1e46d18662636214509cbb2f29c455e46/webrender/res/brush_image.glsl#L150)):*

it's really a multiplication. Basically multiplying by repeat stretches the rectange, which is then chopped off by the modulo, whereas an addition would translate it.
Sorry I am not super good at explaining this type of thing without waving hands and drawing on pieces of papers, but we can chat about it on vidyo or something if you want to go into the details.
</blockquote></details>

Ok, I think I understand now.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2644)*

kvark:
:+1: Needs a try push and CI fixed.

nical:
The pixel-snapping issue on the try run is related to how gecko renders some types of gradients with borders:

gradient-snapping

In the image above, the transparent red shape is a border. There's a gradient rect in the middle and css mandates that we repeat the gradient under the border, so we setup a gradient rect with extra repetitions (it starts one repetition before the original rect in css and ends one repetition after), and set the local clip rect to clip out the gradient repetitions beyond the border (in this drawing the clip rect isn't represented but it would basically hide the areas not covered by the red border or the rectangle inside of it.

This works with gecko's regular painting path because each gradient repetition is pixel snapped individually, and somehow the pixel snapping is expected to match the way the border itself is snapped.
If we do gradient repetitions in the shader, then we have a problem because pixel snapping is applied at the level of the whole gradient item, and then the way the shader renders repetitions acts like a sort of pixel snapping (there is no blurriness between repetitions), but this repetition snapping doesn't necessarily line up with gecko's snapping applied to the border.

Bummer.

This whole pixel-snapping business is unfortunately affecting a lot of different things and not just this PR so it'd be great if we could come up with a coherent solution (that doesn't involve snapping each repetition on the CPU).

For the specific case of borders and gradients/images, I think that it can be fixed by having gecko manually generate nine gradient primitives so that we can snap them in a way that matches the border, Not sure how complicated this would be.

nical:
We talked about this during the gfx meeting and decided that it was okay to just snap the stretch size in gecko, which means the error potentially accumulates up to large values when there are a lot of repetitions, but it seems that Firefox is the only browser trying to do this properly right now, so the accumulated error issue would end up matching what Chrome and Safari do.

kvark:
>  okay to just snap the stretch size in gecko

That would also fix https://bugzilla.mozilla.org/show_bug.cgi?id=1401665 (relevant - #2615)

mstange:
> We talked about this during the gfx meeting and decided that it was okay to just snap the stretch size in gecko

We also decided that Gecko needs to tell WebRender the location of the most important repetition of the image, so that snapping can snap where it actually matters. Otherwise, as in the case shown above, WebRender will snap the top left repetition and end up with a misalignment for the center repetition, even though the center repetition is the most important one.

As for the API, we should probably replace the "stretch size" parameter with a "dest rect" parameter. That's the rectangle that will be snapped and which all repetitions will start from and be aligned with.

nical:
The change I mentioned in my previous comment is quite impactful, and can't make it before the end of Firefox's soft freeze in may 7th. I propose to do the following in the next few days:
 - land #2621 
 - land this
 - in the next webrender update, disable the reftest I am referring to in my earlier comment

If the repetition+border issue is too noticeable I can make a quick wberender-only workaround on the gecko side (just round both the prim rect and stretch size when building the webrender commands).

I'll look into changing the repetition behavior in gecko globally, wait until after the end of the soft freeze to land it, and after that reenable the reftest.

@staktrace how does this sound to you?

staktrace:
The soft freeze doesn't start until April 26 - will it take longer than that to make the change? If so then I guess your proposal seems reasonable enough.

nical:
Oh I misread the email. I'm still keen on breaking the dependency chain because I don't know how long it will take me to figure out the snapping change for all of gecko and I am fed up with rebasing pull requests and landing a gazillion things at once when things finally line up.

staktrace:
Ok, fair enough. I'm all in favour of landing stuff incrementally where possible, as you might have noticed with the trickle of async-scene-building stuff :)

glennw:
Reviewed 11 of 11 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2644)*

nical:
I updated the reference for the blend-clipped test. The difference is in the bottom row of pixels on the shape which is now slightly lighter. This is most likely due to the change in the anti-aliasing (it's the reason for updating the other reftest references).
I'm not quite sure why there would be aa on a primitive that looks axis-aligned, but the aa was already there before. I also rebased the PR and I think something related to pixel-snapping clips landed recently so I might have to update the reference again, we'll see.

Just to be sure, what kind of breakage happened on this test and were symptomatic of many artifacts in gecko?

nical:
As far as I am concerned this is ready to land. Any objection?

glennw:
I'm fine with landing this - although it probably makes sense to do a final try run first?

nical:
Did another try push and the result haven't changed. @bors-servo r=glennw

staktrace:
@nical @glennw this PR seems to have introduced some reftest failures in Gecko. See https://bugzilla.mozilla.org/show_bug.cgi?id=1457241#c3 - some of them might be fuzzable but /layout/reftests/bugs/1291528.html at least seems to have a visible difference.

PR #2648, opened 2018-04-11
stop using bare println

Gankro:
println writes to stdout unconditionally, which can cause release builds of
firefox to crash (panic) if stdout becomes unavailable. We should feed all
prints (that aren't explicitly for print-debugging) through our logging
framework so that stdout disappearing can be correctly handled by our
logging crate.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2648)


Gankro:
see: https://bugzilla.mozilla.org/show_bug.cgi?id=1434630
and: https://bugzilla.mozilla.org/show_bug.cgi?id=1449306

PR #2651, opened 2018-04-12
Add basic infrastructure for caching Pictures, enable for some blurs.

glennw:
Adds the basic infrastructure for caching the outputs of a
rasterized Picture in the texture cache.

For now, this only applies if the following conditions are met:
 * The picture has a blur filter on it (includes text-shadows).
 * The contents of the picture and clips are in the root
   coordinate system (no rotations / perspective etc).

This covers most real-world use cases of blurs and text-shadows.

Future work:
 * Expand to cache other Picture types (e.g. drop-shadows, other filters).
 * Support caching of complex coordinate system pictures which
   are not animating (and thus worth caching).
 * Support deep comparison of picture elements between display lists.
   (caching only applies during scrolling in this first pass).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2651)


webrender/src/picture.rs
```diff
@@ -229,26 +304,92 @@ impl PicturePrimitive {
...
+                                pic_relative_device_rect,
+                            }),
+                        },
+                        frame_state.gpu_cache,
+                        frame_state.render_tasks,
+                        None,
+                        |render_tasks| {
+                            let child_tasks = mem::replace(&mut pic_state_for_children.tasks, Vec::new());
+
+                            let picture_task = RenderTask::new_picture(
```
mrobinson: This is an almost duplicated version of the code above. Is there anyway to only write this one time?

webrender/src/batch.rs
```diff
@@ -1442,6 +1452,45 @@ impl AlphaBatchHelpers for PrimitiveStore {
...
+            }
+        }
+    }
+
+    // Retrieve the render task id for a picture surface. Should only
+    // be used where it's known that this picture surface will never
+    // be persisted in the texture cache.
+    fn resolve_render_task_id(
+        &self,
+    ) -> RenderTaskId {
```
mrobinson: Looks like this fits on one line?

kvark:
Good stuff! I got a few concerns, don't appear to be blockers.

---

---

*[webrender/src/picture.rs, line 98 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vO17ICvHzTTEDbool:-L9vO17ICvHzTTEDboom:b-ey41w1) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L98)):*
> ```Rust
>     //           happen as a cached picture scrolls off-screen,
>     //           we could round up the size of the off-screen
>     //           targets we draw (e.g. 512 pixels). This may
> ```

great idea

---

*[webrender/src/picture.rs, line 102 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vNx4ACnhxl5VBq1xG:-L9vNx4BEm61GlYSYWZR:b-xrku75) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L102)):*
> ```Rust
>     //           deal with clipped/unclipped rects, such as
>     //           the code to inflate the device rect for blurs.
>     pic_relative_device_rect: DeviceIntRect,
> ```

sounds like it needs a `PictureIntRect` :)

---

*[webrender/src/picture.rs, line 108 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vO8rR646nYlTL6s8b:-L9vO8rR646nYlTL6s8c:byi8oi5) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L108)):*
> ```Rust
>     // happen, for example, during zooming or changes
>     // in device-pixel-ratio.
>     unclipped_size: DeviceIntSize,
> ```

how is this different from `pic_relative_device_rect.size`?

---

*[webrender/src/picture.rs, line 385 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vUHLY3YXeU9XSDUEZ:-L9vUHLY3YXeU9XSDUE_:b-r6trgy) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L385)):*
> ```Rust
>                             pic_state.tasks.push(render_task_id);
> 
>                             (render_task_id, false)
> ```

we should really replace this boolean with an enum value (not needed in this PR)

---

*[webrender/src/picture.rs, line 392 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vQ7gi2aPrsg-y2eN2:-L9vQ7gj-oMFFoqjBpFB:b-plbzvb) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L392)):*
> ```Rust
>                 };
> 
>                 self.surface = Some(surface);
> ```

I'm concerned about what's going to happen with the old `surface` we had, and if we are leaking anything here.
While `GpuCacheHandle` will expire, the `CacheTextureId` associated with the `SourceTexture` may need manual freeing (depending on who owns it).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2651#-:-L9vUvF-0693psd43lNk:b-v1i47k)*

glennw:
OK, I finally see what's causing the test failure on try. Ugh, why is it that drop-shadows break every single assumption in WR :sob: I'll need to work on a couple of changes to the drop shadow implementation as a separate PR, so this is blocked until that work is done.

webrender/src/picture.rs
glennw: Not easily, unfortunately. The APIs rely on the new Picture tasks being added to the task tree to get the ID passed to the following constructor, so they can't be done outside the loop (and often won't be needed at all). We can probably improve this in the future though.

webrender/src/batch.rs
glennw: Fixed

glennw:
Review status: 0 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[webrender/src/picture.rs, line 102 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vNx4ACnhxl5VBq1xG:-L9xADW_7c4_S2zJdP8W:b-896fix) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L102)):*

sounds like it needs a `PictureIntRect` :)
</blockquote></details>

Done.

---

*[webrender/src/picture.rs, line 108 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vO8rR646nYlTL6s8b:-L9xAEK8AOM8yR9lDZC8:bmztgg5) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L108)):*

how is this different from `pic_relative_device_rect.size`?
</blockquote></details>

The pic relative size is the size of the clipped rect within the total prim screen rect.

The unclipped size is the total size of the primitive rect in screen space.

Although it would probably never happen in a real-world situation, it's technically possible that the zoom / pan of a scene could change in such a way that the unclipped size changes, but the relative clipped rect stays the same. Comparing both here ensures that the cache would be invalidated when that occurs.

---

*[webrender/src/picture.rs, line 385 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vUHLY3YXeU9XSDUEZ:-L9xBA0SDUP6YnXYRoDa:bs3c4pe) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L385)):*

we should really replace this boolean with an enum value (not needed in this PR)
</blockquote></details>

Yep, agreed. Instead of being a bool for opacity, we will probably end up making it be a BlendMode instead, which can express a superset of what's currently possible.

---

*[webrender/src/picture.rs, line 392 at r1](https://beta.reviewable.io/reviews/servo/webrender/2651#-L9vQ7gi2aPrsg-y2eN2:-L9xBNet7mpTcLXxFg1i:b-khypx7) ([raw file](https://github.com/servo/webrender/blob/7ca20a9a8203543c3a38f3c36a7037b0000397a4/webrender/src/picture.rs#L392)):*

I'm concerned about what's going to happen with the old `surface` we had, and if we are leaking anything here.
While `GpuCacheHandle` will expire, the `CacheTextureId` associated with the `SourceTexture` may need manual freeing (depending on who owns it).
</blockquote></details>

This is handled by the render task cache. In https://github.com/servo/webrender/blob/1b1fa6acd4975570ad96d27d04652663d4bb6d0d/webrender/src/render_task.rs#L896, we scan the render task cache for any cache items where the backing texture has been evicted, and clean up those entries there.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2651)*

glennw:
OK, this patch now includes #2656 (which can be reviewed separately) which fixes the test failure in the try run. I've also addressed each of the review comments. I'll kick off a new try run shortly to make sure that these changes don't break any other tests.

glennw:
Try run for this patch looks good now:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=3928cd6a91763adc7672f5a53c8ee003629bc536&selectedJob=173423585

Just one failure in R3, which is a max difference of 1 for 59 pixels, so it should be fine to fuzz that.

re-r? @kvark and @mrobinson

kvark:
:lgtm:

---

Reviewed 2 of 13 files at r2, 14 of 14 files at r3.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2651#-:-L9zn62o79NlqAKlkpBd:bnfp4nl)*

Issue #2469, opened 2018-02-26
Things with massive sizes break webrender (in various ways)


mateon1:
This was found while fuzzing servo. Things with massive widths/heights break WebRender, causing panics and freezes.

@emilio tested these on Firefox, the freeze testcase also freezes Firefox with WR.

Normal:
```html
<span style="border-radius:50%;border:dashed;display:inline-block;width:18353px">xxx</span>
```
Panics:
```html
<span style="border-radius:50%;border:dashed;display:inline-block;width:18354px">xxx</span>

<span style="border-radius:50%;border:dashed;display:inline-block;width:115000px">xxx</span>
```
Freeze:
```html
<span style="border-radius:50%;border:dashed;display:inline-block;width:115001px">xxx</span>
```

Servo panic:
```rust
assertion failed: block_count <= self.max_block_count (thread WRRenderBackend#0, at /home/mateon/.cargo/git/checkouts/webrender-c3596abe1cf4f320/7ae075a/webrender/src/gpu_cache.rs:509)
stack backtrace:
   0:     0x5603bcfb4704 - backtrace::backtrace::libunwind::trace::h79984cabe80721dc
                        at /home/mateon/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.2/src/backtrace/libunwind.rs:53
                         - backtrace::backtrace::trace::h10c59b076f73358b
                        at /home/mateon/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.2/src/backtrace/mod.rs:42
   1:     0x5603bcfabe8c - backtrace::capture::Backtrace::new::haef9cefb7535bc7c
                        at /home/mateon/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.2/src/capture.rs:64
   2:     0x5603b6c34a16 - servo::main::{{closure}}::hf0fa8cccddc0cbc3
                        at ports/servo/main.rs:146
   3:     0x5603bcfc7b45 - std::panicking::rust_panic_with_hook::haa00f5a9417cd684
                        at libstd/panicking.rs:577
   4:     0x5603bb671d07 - std::panicking::begin_panic::h84df6a344622a174
                        at /checkout/src/libstd/panicking.rs:537
   5:     0x5603bb539e2b - <webrender::gpu_cache::GpuDataRequest<'a> as core::ops::drop::Drop>::drop::h35ecc49d83fecd30
[...]

PR #2652, opened 2018-04-12
Pass dot / dash locations to clip shader in the VAO

mrobinson:
Instead of loading dot and dash positions into the GPU cache, send them
as VAO arguments.  This avoids exceeding the maximum GPU cache block
size with very large border corners with hundreds or thousands of dot /
dashes.

This fixes the first panic from #2469, though performance is still
abysmal. Later changes can address performance issues.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2652)

### DIFF
--- a/webrender/res/cs_clip_border.glsl
+++ b/webrender/res/cs_clip_border.glsl
@@ -4,6 +4,9 @@
 
 #include shared,clip_shared
 
+in vec4 aDashOrDot0;
+in vec4 aDashOrDot1;
+
 varying vec3 vPos;
 
 flat varying vec2 vClipCenter;
@@ -46,9 +49,8 @@ struct BorderClipDash {
     vec4 point_tangent_1;
 };
 
-BorderClipDash fetch_border_clip_dash(ivec2 address, int segment) {
-    vec4 data[2] = fetch_from_resource_cache_2_direct(address + ivec2(2 + 2 * (segment - 1), 0));
-    return BorderClipDash(data[0], data[1]);
+BorderClipDash fetch_border_clip_dash(ivec2 address) {
+    return BorderClipDash(aDashOrDot0, aDashOrDot1);
 }
 
 // Per-dot clip information.
@@ -56,9 +58,8 @@ struct BorderClipDot {
     vec3 center_radius;
 };
 
-BorderClipDot fetch_border_clip_dot(ivec2 address, int segment) {
-    vec4 data = fetch_from_resource_cache_1_direct(address + ivec2(2 + (segment - 1), 0));
-    return BorderClipDot(data.xyz);
+BorderClipDot fetch_border_clip_dot(ivec2 address) {
+    return BorderClipDot(aDashOrDot0.xyz);
 }
 
 void main(void) {
@@ -98,7 +99,7 @@ void main(void) {
         switch (corner.clip_mode) {
             case CLIP_MODE_DASH: {
                 // Fetch the information about this particular dash.
-                BorderClipDash dash = fetch_border_clip_dash(cmi.clip_data_address, cmi.segment);
+                BorderClipDash dash = fetch_border_clip_dash(cmi.clip_data_address);
                 vPoint_Tangent0 = dash.point_tangent_0 * sign_modifier.xyxy;
                 vPoint_Tangent1 = dash.point_tangent_1 * sign_modifier.xyxy;
                 vDotParams = vec3(0.0);
@@ -106,7 +107,7 @@ void main(void) {
                 break;
             }
             case CLIP_MODE_DOT: {
-                BorderClipDot cdot = fetch_border_clip_dot(cmi.clip_data_address, cmi.segment);
+                BorderClipDot cdot = fetch_border_clip_dot(cmi.clip_data_address);
                 vPoint_Tangent0 = vec4(1.0);
                 vPoint_Tangent1 = vec4(1.0);
                 vDotParams = vec3(cdot.center_radius.xy * sign_modifier, cdot.center_radius.z);
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -12,9 +12,10 @@ use clip_scroll_tree::{CoordinateSystemId};
 use euclid::{TypedTransform3D, vec3};
 use glyph_rasterizer::GlyphFormat;
 use gpu_cache::{GpuCache, GpuCacheAddress};
-use gpu_types::{BrushFlags, BrushInstance, ClipChainRectIndex, ZBufferId, ZBufferIdGenerator};
-use gpu_types::{ClipMaskInstance, ClipScrollNodeIndex, RasterizationSpace};
-use gpu_types::{CompositePrimitiveInstance, PrimitiveInstance, SimplePrimitiveInstance};
+use gpu_types::{BrushFlags, BrushInstance, ClipChainRectIndex, ClipMaskBorderCornerDotDash};
+use gpu_types::{ClipMaskInstance, ClipScrollNodeIndex, CompositePrimitiveInstance};
+use gpu_types::{PrimitiveInstance, RasterizationSpace, SimplePrimitiveInstance, ZBufferId};
+use gpu_types::ZBufferIdGenerator;
 use internal_types::{FastHashMap, SavedTargetIndex, SourceTexture};
 use picture::{PictureCompositeMode, PicturePrimitive, PictureSurface};
 use picture::{IMAGE_BRUSH_BLOCKS, IMAGE_BRUSH_EXTRA_BLOCKS};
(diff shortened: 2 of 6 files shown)


mrobinson:
@glennw I'd love to get your feedback on my approach here. In particular I'm curious if I should be doing more to discards dots and dashes in the vertex shader. I'm also curious if my method of modifying the ClipMaskInstance data structure using GLSL preprocessor directives makes sense or if there is a more elegant method. I've decided to use a single data structure (a data blob like primitives use) for both dots and dashes. This seems like it would give a minor improvement to batching and prevent having to compile separate shaders for dots and dashes.

glennw:
Yup, this seems like a good general approach to me (I haven't looked in super detail).

I'm not sure that we need any of the pre-processor changes you mentioned - each clip shader will know the exact vertex format / kind it's reading (e.g. clip_dot vs clip_rect shader), so it's probably fine to define a GLSL type for each of those?
 
I think we do need to do some clipping work for dots / dashes. Since dots and dashes are drawn as geometry, it's possible they may extend outside the task rect of the clip mask, I *think*.

Ideally we can do this in the VS (cull the entire dot if it's outside the task rect), but I think we'll probably need a FS clip as well, to ensure that partial dots don't write outside the task rect. Another option might be to just use a scissor rect (and disable batching) on clip masks that have dots / dashes, since they are quite rare. This wouldn't be ideal, but might be fine for a first pass. The case where we will actually need clipping for this is quite rare, but I'm fairly sure it can occur - e.g. if a border with dots has a clip chain that results in part of the border being removed?

I like the idea of @kvark 's suggestion - but I'm a little wary of it. As we discussed previously, the existing Gecko code for dot / dash placement is absolutely massive. For now, we're assuming we can get away with a fairly simple placement algorithm, which would be fine to evaluate in a VS / texture, but if we end up needing to do something more complex, that won't be feasible.

mrobinson:
@glennw @kvark Thanks for looking at the PR. I've updated it to improve clipping in the fragment shader. Now the mask rect is clamped to the task rect, which I think properly handles partial dots. A good follow-up for this version would be to produce vertices representing a tight bounding rectangle of the dot or dash. I'm working on that now, but this version seems to fix the issues raised in the bug while maintaining the performance of the old approach.

As far as the shader goes, the dot/dash shader shares almost every input variable with the normal clip mask shader, so I used the GLSL preprocessor to reuse the `ClipMaskInstance` data structure from `clip_shared.gllsl`. I'm happy to remove this and just duplicate the data structure, but I think we'll still need to use the preprocessor (in order to adjust the list of per-vertex input arguments to the shader) or duplicate some code (`distance_to_line` if we don't include `clip_shared.glsl` in the dot/dash shader).

mrobinson:
This patch is read to go, except I'm seeing a timeout during the crashtests: https://treeherder.mozilla.org/#/jobs?repo=try&revision=1f61787a13f729be243398b316548031ad5cbdb3&selectedJob=174110305

When I run these tests locally, another crashtest times out, both with and without my changes. I'm a bit stumped what is causing this behavior.

mrobinson:
I'm still having trouble debugging the crashtest timeout (though the Gecko tree is closed). I'm hoping to make some progress once it reopens.

mrobinson:
@glennw I was finally able to get a passing try run for Gecko (modulo a few intermittent failures):

https://treeherder.mozilla.org/#/jobs?repo=try&revision=38bf3c0a42ab4ffb0b520f1325076668be37bb7d&selectedJob=174545800

I had to make some changes in order to make this happen:
 * I reverted some of the cleaning up I did around `BorderCornerClipSource` initialization. This wasn't strictly necessary, though it made the code more efficient. This can be introduced gradually in order to narrow down what was causing the timeouts during the crash tests.
 * I had to remove the shader clipping. This is important for performance, but I verified that the old approach was just as bad. Would it be okay to get this landed and then tackle clipping. I'm thinking it might take a couple days to figure out what is causing the fuzziness and I think it would help to get this cleared away first.

Thanks!

mrobinson:
Looks like I spoke too soon. This does seem to change the details of the fuzziness (an already very fuzzy test) for layout/reftests/css-break/box-decoration-break-with-inset-box-shadow-1.html. It seems reasonable that we can just update the fuzz for this one.

glennw:
Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2652)*

staktrace:
I'm ok with the box-decoration-break-with-inset-box-shadow-1.html fuzz since that's just increasing pre-existing fuzziness by a little bit. But the position-sticky/inline-4.html fuzz is new, and has a user-visible difference (one of the dots is off). Is this expected, or is there something wrong with the test, or something else?

mrobinson:
@staktrace Sorry, it seems the `position-sticky/inline-4.html` failure came in later and I thought I had fixed it. It's really mysterious, because it's a Windows only failure. These changes should not change at all the behavior of the shaders. I'll continue investigating, but this is incredibly weird.

mrobinson:
Okay. After a lot of banging my head against this issue, I think I have figured out what was going on! The first versions of this PR changed the alignment of the VAO for dots and dashes. When I restored the unused element of the ClipMaskInstance the reference test failures disappeared. This also makes the diff much smaller. Here is the latest Gecko try job:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=7d8ae92639770c3bd37a8f776742a8ad44b3c862&selectedJob=175108604

I only see two seemingly unrelated intermittent failures.

glennw:
Reviewed 6 of 7 files at r2.

---

*[webrender/res/cs_clip_border.glsl, line 52 at r2](https://beta.reviewable.io/reviews/servo/webrender/2652#-LAoQFx5Buc8TAoiyqRt:-LAoQFx5Buc8TAoiyqRu:b-7b3ko4) ([raw file](https://github.com/servo/webrender/blob/8c5a85b2b57d039e35bb07b232d041e8fc2dd73e/webrender/res/cs_clip_border.glsl#L52)):*
> ```GLSL
> };
> 
> BorderClipDash fetch_border_clip_dash(ivec2 address, int segment) {
> ```

We can remove segment as a param from here.

---

*[webrender/res/cs_clip_border.glsl, line 61 at r2](https://beta.reviewable.io/reviews/servo/webrender/2652#-LAoQJ45DPUG3TNaIzos:-LAoQJ45DPUG3TNaIzot:bb73xfv) ([raw file](https://github.com/servo/webrender/blob/8c5a85b2b57d039e35bb07b232d041e8fc2dd73e/webrender/res/cs_clip_border.glsl#L61)):*
> ```GLSL
> };
> 
> BorderClipDot fetch_border_clip_dot(ivec2 address, int segment) {
> ```

Same

---

*[webrender/src/batch.rs, line 1788 at r2](https://beta.reviewable.io/reviews/servo/webrender/2652#-LAoQMTHAF5WhkItecGG:-LAoQMTHAF5WhkItecGH:baykbjw) ([raw file](https://github.com/servo/webrender/blob/8c5a85b2b57d039e35bb07b232d041e8fc2dd73e/webrender/src/batch.rs#L1788)):*
> ```Rust
>                             self.borders.push(ClipMaskBorderCornerDotDash {
>                                 clip_mask_instance: ClipMaskInstance {
>                                     segment: 1,
> ```

Does segment get used here at all, or could we just leave it as the base instance?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2652)*

glennw:
A couple of minor nits, r=me after those are addressed.

mrobinson:
Thanks for the review!

---

---

*[webrender/res/cs_clip_border.glsl, line 52 at r2](https://beta.reviewable.io/reviews/servo/webrender/2652#-LAoQFx5Buc8TAoiyqRt:-LAoSVfy8Eui7rrHOvtF:b-nusppu) ([raw file](https://github.com/servo/webrender/blob/8c5a85b2b57d039e35bb07b232d041e8fc2dd73e/webrender/res/cs_clip_border.glsl#L52)):*

We can remove segment as a param from here.
</blockquote></details>

Okay. I'll fix this.

---

*[webrender/src/batch.rs, line 1788 at r2](https://beta.reviewable.io/reviews/servo/webrender/2652#-LAoQMTHAF5WhkItecGG:-LAoS_7i5T9DK0bH_2YG:bmg81yj) ([raw file](https://github.com/servo/webrender/blob/8c5a85b2b57d039e35bb07b232d041e8fc2dd73e/webrender/src/batch.rs#L1788)):*

Does segment get used here at all, or could we just leave it as the base instance?
</blockquote></details>

The shader understands the 0 segment as the clear, so this just needs to be != 0. I can add a comment here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2652#-:-LAoSrGz9_s2giLa_4WZ:b-9qdoz3)*

PR #2653, opened 2018-04-12
Intersect the dirty rect with the tile

jrmuizel:
This prevents the dirty rect from extending beyond the size of tile
and causing explosions later on.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1453718

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2653)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -147,6 +147,29 @@ pub struct ResourceClassCache<K: Hash + Eq, V, U: Default> {
     pub user_data: U,
 }
 
+fn intersect_for_tile(
+    dirty: DeviceUintRect,
+    width: u32,
+    height: u32,
+    tile_size: TileSize,
+    tile_offset: TileOffset,
+
+) -> Option<DeviceUintRect> {
+        dirty.intersection(&DeviceUintRect::new(
+            DeviceUintPoint::new(
+                tile_offset.x as u32 * tile_size as u32,
+                tile_offset.y as u32 * tile_size as u32
+            ),
+            DeviceUintSize::new(width, height),
+        )).map(|mut r| {
+                // we can't translate by a negative size so do it manually
+                r.origin.x -= tile_offset.x as u32 * tile_size as u32;
+                r.origin.y -= tile_offset.y as u32 * tile_size as u32;
+                r
+            })
+}
+
+
 impl<K, V, U> ResourceClassCache<K, V, U>
 where
     K: Clone + Hash + Eq + Debug,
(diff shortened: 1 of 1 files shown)


webrender/src/resource_cache.rs
```diff
@@ -935,6 +936,20 @@ impl ResourceCache {
                 let (actual_width, actual_height) =
                     compute_tile_size(image_descriptor, tile_size, tile);
 
+                if let Some(dirty) = dirty_rect {
+                    dirty_rect = dirty
```
kvark: `dirty_rect` being `None` means that the whole image needs to be updated. In this case though, if there is zero intersection with the tile, we don't want any updates at all. I.e.:
```rust
if let Some(ref mut dirty) = dirty_rect {
  *dirty = match dirty.intersection(...) {
    Some(r) => {...; r}
    None => continue,
  };
}
```

webrender/src/resource_cache.rs
```diff
@@ -935,6 +936,20 @@ impl ResourceCache {
...
+                        .intersection(&DeviceUintRect::new(
+                            DeviceUintPoint::new(tile.x as u32 * tile_size as u32, tile.y as u32 * tile_size as u32),
+                            DeviceUintSize::new(actual_width, actual_height)))
+                        .map(|mut x| {
+                            // we can't translate by a negate size so do it manually
+                            x.origin.x -= tile.x as u32 * tile_size as u32;
+                            x.origin.y -= tile.y as u32 * tile_size as u32;
+                            x
+                        }
+                        )
```
kvark: formatting

webrender/src/resource_cache.rs
```diff
@@ -935,6 +936,24 @@ impl ResourceCache {
                 let (actual_width, actual_height) =
                     compute_tile_size(image_descriptor, tile_size, tile);
 
+                if let Some(ref mut dirty) = dirty_rect {
+                    *dirty = match dirty.intersection(&DeviceUintRect::new(
+                            DeviceUintPoint::new(
+                                tile.x as u32 * tile_size as u32,
```
kvark: I suppose this should be `(tile.x * tile_size) as u32`?

webrender/src/resource_cache.rs
```diff
@@ -935,6 +936,24 @@ impl ResourceCache {
...
+                                tile.y as u32 * tile_size as u32
+                            ),
+                            DeviceUintSize::new(actual_width, actual_height),
+                    )) {
+                        Some(mut r) => {
+                            // we can't translate by a negative size so do it manually
```
kvark: we could probably have a `Vector2D` for that offset?

jrmuizel: DeviceUintVector2D would still need positive values so I don't think that would work.

webrender/src/resource_cache.rs
jrmuizel: that would be u16 * u16 which we don't want.

kvark: ah, I assumed those were floats.

kvark:
@jrmuizel ok, I think this is fine to land now (r=me), unless you want to ship the raw test as well here.

jrmuizel:
It looks like we might need to clip in two places. This patch does that, but I'll try to separate it out into a function.

jrmuizel:
This should be ready now. I'll submit the test changes in follow ups.

PR #2655, opened 2018-04-12
Invert Y coordinates of DOM to Texture images

mbrubeck:
Fixes servo/servo#20540.  r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2655)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -2920,7 +2920,12 @@ impl Renderer {
                     }
                 };
                 let (src_rect, _) = render_tasks[output.task_id].get_target_rect();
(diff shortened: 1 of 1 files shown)


webrender/src/renderer.rs
```diff
@@ -2920,7 +2920,12 @@ impl Renderer {
                     }
                 };
                 let (src_rect, _) = render_tasks[output.task_id].get_target_rect();
-                let dest_rect = DeviceIntRect::new(DeviceIntPoint::zero(), output_size);
+                let mut dest_rect = DeviceIntRect::new(DeviceIntPoint::zero(), output_size);
```
kvark: @glennw we need to adjust the corresponding example code

PR #2656, opened 2018-04-13
Change drop-shadows to be drawn via a single primitive.

glennw:
Previously, drop-shadows were a Picture, with two brush image
primitives, each referencing the picture (one for the content,
one for the shadow with an offset).

Although conceptually reasonable, this exposes some problems
with the way the current prim_store visibility pass works.
Specifically, we can end up processing the picture more than
once. In general this is an inefficiency but doesn't affect
correctness, but it breaks some assumptions once the content
of the drop-shadow element is an item in the render task cache.

In the future, this problem should disappear as we refactor
how the visibility pass operates, but for now the simplest solution
is to draw the drop-shadow with a single primitive. To make this
work, there are a couple of hacks in this patch, which
push an extra brush image primitive into the GPU cache data.

The drop shadows still have an issue with the shadow disappearing
if the content of the drop shadow is off-screen, however this is
no worse than the existing code. Doing it this way actually makes
it easier to fix this is the future too.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2656)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -638,7 +638,7 @@ impl AlphaBatchBuilder {
                 let brush = &ctx.prim_store.cpu_brushes[prim_metadata.cpu_prim_index.0];
 
                 match brush.kind {
-                    BrushKind::Picture { pic_index, source_kind, .. } => {
+                    BrushKind::Picture { pic_index, .. } => {
                         let picture =
                             &ctx.prim_store.pictures[pic_index.0];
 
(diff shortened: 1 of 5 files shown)


glennw:
r? @kvark 

Pending try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=27f305d700495ca20b6f4525b3a00e93ae15c498

glennw:
Try run looks good to me.

kvark:
:lgtm:

>  In general this is an inefficiency but doesn't affect
correctness, but it breaks some assumptions once the content
of the drop-shadow element is an item in the render task cache.

I'd like to know more about this.

---

Reviewed 5 of 5 files at r1.

---

*[webrender/src/picture.rs, line 171 at r1](https://beta.reviewable.io/reviews/servo/webrender/2656#-L9z819V0W67xDLNEn9C:-L9z819V0W67xDLNEn9D:b-8fx2u7) ([raw file](https://github.com/servo/webrender/blob/cf9d1473410901ef689b67085876c7a40f1e005c/webrender/src/picture.rs#L171)):*
> ```Rust
>                 //           drop-shadows from disappearing if the main content
>                 //           rect is not visible. Something like:
>                 // let shadow_rect = local_content_rect
> ```

:+1: 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2656#-:-L9z8e050D1UiUlHDP5f:b-r1u5rx)*

PR #2661, opened 2018-04-13
wrench/blob: Only paint the dirty rect

jrmuizel:
Previously we were painting the entire blob and ignoring the dirty rect.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2661)


jrmuizel:
This needs #2653 to work properly

glennw:
#2653 has landed, so it looks like this is ready to go.

@bors-servo r=kvark

Issue #2657, opened 2018-04-13
Bad subpixel AA inside stacking context for mask


mstange:
https://bugzilla.mozilla.org/show_bug.cgi?id=1454097

Now that we're using stacking contexts for masks ([bug 1412375](https://bugzilla.mozilla.org/show_bug.cgi?id=1412375)), if you go to a site with a long URL, like [this one](https://treeherder.mozilla.org/#/jobs?repo=try&revision=75fd86d6099ab3ce8ed67ab6cec8cbda7759f319&selectedJob=173318894), the text in the URL bar will look funny. Firefox applies a fade-out mask effect to long URLs.

I think this is happening because we're incorrectly allowing the use of subpixel AA inside the intermediate surface for the stacking context, even though we're drawing onto transparency.

We should do the following:

 1. Force grayscale AA inside stacking contexts unless we know that the text is being drawn onto opaque pixels within that stacking context.
 2. As a future optimization and quality improvement, detect the case where the grouping provided by the stacking context is unnecessary, and discard the intermediate surface and apply the mask to the contents directly. This would give us the previous appearance for the URL bar.

mrobinson:
@mstange Do you mind expanding on why the grouping is unnecessary here?

glennw:
I *think* the problem is:

https://github.com/servo/webrender/blob/1875203eb3f0afade5677a4a420fc70aef8ded8f/webrender/src/display_list_flattener.rs#L1274

If this becomes:

```
let allow_subpixel_aa = composite_mode.is_none();
```

Then I think this will fix the problem.

The test itself is still super-conservative - in addition to the optimization mentioned in (2) above, we'll also want to detect cases where we have an intermediate surface but we know the background is opaque.

We don't need to do this now, but we should do it in the future to allow subpixel AA in more cases than we currently do. Detecting (what I expect to be) the common case for this should be really simple - we just determine when checking the visibility of a picture if the local rect of any solid rects in this picture cover the entire local rect of the picture. This would mean that it's safe to use subpixel AA (and could also be used as a cheap occlusion cull for any primitives that come before such a rect in this picture).

Anyway, hopefully the one line fix mentioned above should be enough to fix the reported bug here. I can try it next week if no-one else gets to it first.

Gankro:
@mrobinson I believe grouping is unnecessary in this case because there's only one primitive that's getting masked (a textrun). Potentially an easy case to notice in gecko.

mstange:
[...]

PR #2662, opened 2018-04-15
hotfix for #2657, prevent subpixel in masks

Gankro:
people are noticing this a lot, so let's just do this now and figure out the more robust solutions later.

#2657 still needs a proper regression test, which I'll try to knock out tomorrow

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2662)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -1225,7 +1225,8 @@ impl<'a> DisplayListFlattener<'a> {
         // TODO(gw): This is super conservative. We can expand on this a lot
         //           once all the picture code is in place and landed.
         let allow_subpixel_aa = composite_ops.count() == 0 &&
-                                transform_style == TransformStyle::Flat;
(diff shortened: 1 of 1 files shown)


glennw:
One CI failure:

```
REFTEST reftests/text/white-opacity.yaml == reftests/text/white-opacity.png
REFTEST TEST-UNEXPECTED-FAIL | reftests/text/white-opacity.yaml == reftests/text/white-opacity.png | image comparison, max difference: 72, number of differing pixels: 3247
```

I think it's probably just a reference image update, but I haven't looked in detail.

Gankro:
It's weird that this is only affecting macos release..?

jrmuizel:
We don't run reftests on macOS debug.

glennw:
Ah, yes - I was thinking the composite mode would account for it, but I guess the parent of the 3d content will still have flat transform style.

@bors-servo r+

Issue #2659, opened 2018-04-13
Repeated primitives with sub-px tile spacing.


nical:
Tile spacing exists so that we can implement [bacground-repeat: space](https://developer.mozilla.org/en-US/docs/Web/CSS/background-repeat) which more or less means add some space between repetitions such that the first repetition starts exactly at the top left of the element and the last repetition ends exactly at the lower right of the element.

Example: 

```
data:text/html,<div style="width:500px;height:200px;border:1px solid lightgrey; background:linear-gradient(to right, black 1px, lime 0, lime 199px, black 0); background-size: 200px 200px; background-repeat: space"> 
```

The current plan to support tile spacing is to bake the spacing in a render target by blitting the image with some transparent pixels around it, and then use the regular image shader to render it.

Repeating in the fragment shader makes it hard to support variable spacing to adjust repetitions so that they line up with the element's edge. We could support this by allowing non-integer spacing, but this means that a repeated image that would have been crisp in gecko today could be a bit blurry with webrender.

I'm tempted to say "meh, a bit of blurriness in with background-repeat: space is fine.", but others will (and already do) think otherwise.

If not having blurry repeated background-repeat: space images is important we could:

- Keep decomposing repeated images/gradients with spacing on the CPU. This means we potentially will generate a lot of primitives each frame and can't use a brush segments per repetition because of the limit of (MAX_VERTEX_TEXTURE_WIDTH - VECS_PER_PRIM) / VECS_PER_SEGMENT which means a bit less than 512 segments per primitive.
- come up with a way to avoid the blurriness in the shader
  - maybe having a dedicated repeated_image_with_annoying_parameters.glsl that compensates the factional offset to avoid the extra blurriness somehow.

cc @glennw @mstange @jrmuizel

nical:
To be more specific, the blurriness issue is when `layout_tile_spacing / layout_stretch_size * device_img_size` is not an integer quantity (although I suppose the common case is that layout_stretch_size is a multiple of device_img_size).

nical:
I just came up with another awful workaround:

manually split repeated backgrounds into 9 primitives so that we can manually position the first and last rows/column.

```
+--+------------------+--+
|x |x x x x x x x x x | x|
+--+------------------+--+
|x |x x x x x x x x x | x|
|  |                  |  |
|x |x x x x x x x x x | x|
|  |                  |  |
|x |x x x x x x x x x | x|
+--+------------------+--+
|x |x x x x x x x x x | x|
[...]

PR #2664, opened 2018-04-16
Prerender tile spacing and use the brush shader.

nical:
On top of #2621 and #2644, bake the tile spacing into an image using a blit render task.
This lets us use the brush shader with tile spacing. In order for spacing to fully work the way we intend it to we need a bit more work (see #2659), but since the image shader we use for tile spacing doesn't do the right thing either, I don't think it would be a problem to move forward with this independently of #2659's resolution.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2664)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -1250,8 +1250,8 @@ impl Device {
             src_rect.origin.y + src_rect.size.height,
             dest_rect.origin.x,
             dest_rect.origin.y,
-            dest_rect.origin.x + dest_rect.size.width,
-            dest_rect.origin.y + dest_rect.size.height,
+            dest_rect.origin.x + src_rect.size.width,
+            dest_rect.origin.y + src_rect.size.height,
             gl::COLOR_BUFFER_BIT,
             gl::LINEAR,
         );
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -1250,8 +1250,8 @@ impl Device {
             src_rect.origin.y + src_rect.size.height,
             dest_rect.origin.x,
             dest_rect.origin.y,
-            dest_rect.origin.x + src_rect.size.width,
-            dest_rect.origin.y + src_rect.size.height,
+            dest_rect.origin.x + dest_rect.size.width,
+            dest_rect.origin.y + dest_rect.size.height,
             gl::COLOR_BUFFER_BIT,
             gl::LINEAR,
         );
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -22,7 +22,7 @@ use euclid::{SideOffsets2D, vec2};
 use frame_builder::{FrameBuilder, FrameBuilderConfig};
 use glyph_rasterizer::FontInstance;
 use hit_test::{HitTestingItem, HitTestingRun};
-use image::{decompose_image, TiledImageInfo};
+use image::{decompose_image, TiledImageInfo, simplify_repeated_primitive};
 use internal_types::{FastHashMap, FastHashSet};
 use picture::PictureCompositeMode;
 use prim_store::{BrushKind, BrushPrimitive, BrushSegmentDescriptor, CachedGradient};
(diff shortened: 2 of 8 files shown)


nical:
This might only accidentally work, I didn't write any code to clear the target where the transparent pixels are (I'm guessing we don't automatically clear before blits), and we might want to implement this some other way. But it seems to work (at least in simple cases) with a relatively simple change which is satisfying.

nical:
I am still a bit unsure about where the line is between the two but we might want to use a picture instead of a render task for this in order to support baking gradients.

glennw:
Color render targets are always cleared to transparent before any render tasks are drawn into them (there is a possible optimization opportunity to skip this in cases where it doesn't matter).

The gradient + spacing question is an interesting one - baking them might be a bit tricky in cases where there is a transform active, I think. We can ponder that once we get the image + spacing baking working though :)

nical:
Relevant review comment from #2695 

> webrender/src/device.rs, line 1253 at r1 (raw file):
> 
> ```
>                dest_rect.origin.x,
>                dest_rect.origin.y,
>                dest_rect.origin.x + src_rect.size.width,
> ```
> I don't understand this change. This would mean that we can never do a blit which is a stretch?

Oops, I didn't know we support/need to do blits that stretch. I'll update the PR to support that properly.

nical:
@glennw I added a commit that adds explicit support for padding in blit render tasks instead of using the target rect size and disallowing stretching blits (although we currently have an assertion checking that the source and destination sizes are equal).

nical:
`images/tile-with-spacing-ref.yaml` renders with some garbage in the padded region. I added a bit logging and I am not seeing clears associated to the second blit, only the first one. Putting the padding in the first blit seems to fix the issue (not 100% sure though since it depends on what's in the texture). It's odd though, I'd like to know why I don't seem to be getting all of the clears I expect.

Actually, now that I am poking at this I wonder why we even do two blits and not just one when isolating sub-rects (and prerendering spacing since it uses the same code). I can see things don't show up when I do a single blit but it's not intuitive to me why we need it.

glennw:
When we pre-render spacing, we're doing the following:

```
Texture cache -> render task surface -> Texture cache. 
```

If we don't do the intermediate blit, the source texture is the same as the destination FBO texture, which is undefined behaviour in GL.

glennw:
(I haven't looked at this in detail, below is just a guess).

I suspect that when drawing items to the texture cache as a final destination, we don't issue any clears on that target type?

So the implicit assumption is that a task that targets the texture cache ends up writing every pixel. If that's the case - we should perhaps add support for clear mode (per render task) to texture cache targets, and a clear mode of "DontCare" if the task knows it will write to every pixel, to avoid clearing the task area in that case.

nical:
I punted on clearing the padding in the second blit and just used the first blit to insert the spacing. We can look into having more involved mechanisms later, but I suspect in the more general case we'll want to move to actually use a picture with more commands instead of just a couple of blit render tasks to better support some of the cases where we currently get sampling artifacts.

@glennw This PR should be ready to land if you are OK with that and the resolution in my previous comment.

glennw:
The reftest change seems fine. I don't quite follow how the padding / clearing works though - if we don't clear, won't the previous contents of that texture cache allocation be left in the empty space?

nical:
The first blit properly clears beforehand and we add the padding there. The second one blits the image including the pixels of the padding we prerendered so it shouldn't need a clear (unless blitting does some kind of blending).

glennw:
Oh, of course. That makes sense :) Is this ready to go then, or do we want to do another try push as you mentioned above?

nical:
Yeah, waiting for https://treeherder.mozilla.org/#/jobs?repo=try&revision=d20a0909031b4cd589f4b064e59afe4c81796af9 On [the other PR's try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=c7afdfa3b826df94ff523bbbf9c2850fb8590b22) there is a number of tests with showing shape with a bit of blurriness at the bottom that I suspect come from here, unfortunately.

nical:
I think I figured the main problem out, hopefully this try push will coe back green: https://treeherder.mozilla.org/#/jobs?repo=try&revision=95142a60f5aa1e2b32b383b578aaea41eaeee372

glennw:
Reviewed 2 of 5 files at r1, 3 of 6 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2664)*

PR #2669, opened 2018-04-18
Fix filtered elements being clipped during property animation.

### LINKED ISSUE #2668: Filtered elements are clipped if their bounds change via animated properties
https://bugzilla.mozilla.org/show_bug.cgi?id=1454763

\>> [Testcase](https://bug1454763.bmoattachments.org/attachment.cgi?id=8968651) <<

In this testcase, the rounded border is sometimes not rendered completely. The testcase applies a filter to a stationary element which wraps an element whose transform is animated. The bounds of the filter surface are dependent on the animated property.

glennw:
When the local rect of a picture primitive changes (due to a
child primitive or picture being attached to an animating
scroll node), we need to invalidate the GPU cache entry for
this picture. This ensures that the correct local rect is
uploaded to the GPU shaders.

Fixes #2668.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2669)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -1958,7 +1958,13 @@ impl PrimitiveStore {
                 pic.runs = pic_context_for_children.prim_runs;
 
                 let metadata = &mut self.cpu_metadata[prim_index.0];
-                metadata.local_rect = pic.update_local_rect(result);
+
+                let new_local_rect = pic.update_local_rect(result);
+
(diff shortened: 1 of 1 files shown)


glennw:
r? @kvark or anyone

cc @mstange - this is the fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1454763

I don't think we have any way to add a regression test for this right now, unfortunately.

PR #2674, opened 2018-04-19
Various small changes related to caching text runs with brushes.

glennw:
* Remove unused subpixel blend mode (can retrieve via git if we need later).
* Switch brush_image shader to do mask swizzling the same way as text shader.
  * This allows handling the various subpx modes in brush_image for later.
  * It's more flexible, a superset of existing brush_image swizzle modes.
* Remove BrushImageSourceKind.
* Store two colors inline with the brush primitive data for images / pictures.
  * These are always fetched in the brush_image VS.
  * Extra data is only fetched when drawing drop-shadows / screen-space images.
* Rename TextShaderMode -> ShaderColorMode, move to prim_shared.
* Allow batching code to correctly handle multi-pass mode for any batch kind.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2674)

### DIFF
--- a/webrender/res/brush_blend.glsl
+++ b/webrender/res/brush_blend.glsl
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#define VECS_PER_SPECIFIC_BRUSH 0
+#define VECS_PER_SPECIFIC_BRUSH 2
 #define FORCE_NO_PERSPECTIVE
 
 #include shared,prim_shared,brush
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -19,18 +19,12 @@ flat varying vec4 vUvBounds;
 flat varying vec4 vUvSampleBounds;
 
 #ifdef WR_FEATURE_ALPHA_PASS
-flat varying vec2 vSelect;
 flat varying vec4 vColor;
+flat varying vec2 vMaskSwizzle;
 flat varying vec2 vTileRepeat;
 #endif
 
 #ifdef WR_VERTEX_SHADER
-
-#ifdef WR_FEATURE_ALPHA_PASS
-    #define IMAGE_SOURCE_COLOR              0
-    #define IMAGE_SOURCE_ALPHA              1
-    #define IMAGE_SOURCE_MASK_FROM_COLOR    2
-#endif
 
 struct ImageBrushData {
     vec4 color;
@@ -114,11 +108,9 @@ void brush_vs(
     vec2 f;
 
 #ifdef WR_FEATURE_ALPHA_PASS
-    int image_source = user_data.y >> 16;
+    int color_mode = user_data.y >> 16;
     int raster_space = user_data.y & 0xffff;
-
     ImageBrushData image_data = fetch_image_data(prim_address);
-    vColor = image_data.color;
 
     // Derive the texture coordinates for this image, based on
     // whether the source image is a local-space or screen-space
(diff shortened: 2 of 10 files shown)


glennw:
These are small changes that I can land incrementally, pulled from my local
branch where I am working on drawing (and caching) text runs as brushes. Some
of them probably don't make a heap of sense individually, but are hopefully
easier to review if we can land bits incrementally.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=99dfc3dde70c4a8b8a6a52cebe645c92854a51b7

r? @kvark

staktrace:
Yeah the oranges look unrelated, at least two of them I've seen before and the Wr6 orange looks like a random intermittent.

kvark:
:lgtm: left a few notes

---

---

*[webrender/res/ps_text_run.glsl, line 143 at r1](https://beta.reviewable.io/reviews/servo/webrender/2674#-LAU-BJ85vuw1jQci_Cm:-LAU-BJ85vuw1jQci_Cn:b62njme) ([raw file](https://github.com/servo/webrender/blob/6394ce5e7ea8230e2084010d896e98c7a2be1878/webrender/res/ps_text_run.glsl#L143)):*
> ```GLSL
>             vColor = vec4(text.color.a);
>             break;
>         case COLOR_MODE_SUBPX_BG_PASS1:
> ```

looks like an opportunity to share this logic with `brush_image`

---

*[webrender/src/batch.rs, line 753 at r1](https://beta.reviewable.io/reviews/servo/webrender/2674#-LAU-OCH1sNymYhY1fMG:-LAU-OCH1sNymYhY1fMH:bhxmbux) ([raw file](https://github.com/servo/webrender/blob/6394ce5e7ea8230e2084010d896e98c7a2be1878/webrender/src/batch.rs#L753)):*
> ```Rust
>                                             let extra_data_address = gpu_cache.get_address(&picture.extra_gpu_data_handle);
>                                             let shadow_prim_address = extra_data_address.offset(2);
>                                             let shadow_data_address = extra_data_address.offset(8);
> ```

can we have those defined somewhere? (bonus points for use elsewhere and/or assertions to match the size during the GPU data population)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2674#-:-LAU-wxu07FX-BoCoNis:blwyo0g)*

glennw:
Review status: 8 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[webrender/res/ps_text_run.glsl, line 143 at r1](https://beta.reviewable.io/reviews/servo/webrender/2674#-LAU-BJ85vuw1jQci_Cm:-LAUedLe7WyUNlMLFVeO:bgtzyn5) ([raw file](https://github.com/servo/webrender/blob/6394ce5e7ea8230e2084010d896e98c7a2be1878/webrender/res/ps_text_run.glsl#L143)):*

looks like an opportunity to share this logic with `brush_image`
</blockquote></details>

I haven't done that yet, since if things work out here, ps_text_run will be disappearing soon :) If that doesn't happen, then we should definitely share this code in a function.

---

*[webrender/src/batch.rs, line 753 at r1](https://beta.reviewable.io/reviews/servo/webrender/2674#-LAU-OCH1sNymYhY1fMG:-LAUeoR19PbVb9cvzOgu:b-wmd0jo) ([raw file](https://github.com/servo/webrender/blob/6394ce5e7ea8230e2084010d896e98c7a2be1878/webrender/src/batch.rs#L753)):*

can we have those defined somewhere? (bonus points for use elsewhere and/or assertions to match the size during the GPU data population)
</blockquote></details>

For now, I've made them constants and added a comment near the code in picture.rs that they need to be updated. We should aim to make this type-safe and/or assert in the future though, for sure.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2674)*

kvark:
thanks!
@bors-servo r+

PR #2677, opened 2018-04-20
Some more preparatory work for text run caching.

glennw:
* Support dual-source blending in brush shaders (optionally).
* Optimize how blend enable state is set in alpha batches. 
* Refactor and simplify how submit_batch works for subpixel modes.
  * In most cases, pass the shader color mode via instances.
  * The instance can have a value that says to take the color mode from uMode instead. This is used for multi-pass renderings.
  * Simplify the submit_batch code in a way that text runs are not special cased. This allows any primitive kind to work with any of the subpixel blend modes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2677)

### DIFF
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -112,6 +112,10 @@ void brush_vs(
     int raster_space = user_data.y & 0xffff;
     ImageBrushData image_data = fetch_image_data(prim_address);
 
+    if (color_mode == COLOR_MODE_FROM_PASS) {
+        color_mode = uMode;
+    }
+
     // Derive the texture coordinates for this image, based on
     // whether the source image is a local-space or screen-space
     // image.
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -205,12 +205,7 @@ void brush_vs(
 
 #ifdef WR_FRAGMENT_SHADER
 
-vec4 brush_fs(
-#ifdef WR_FEATURE_DUAL_SOURCE_BLENDING
-    out vec4 blend
-#endif
-) {
-
+Fragment brush_fs() {
     vec2 uv_size = vUvBounds.zw - vUvBounds.xy;
 
 #ifdef WR_FEATURE_ALPHA_PASS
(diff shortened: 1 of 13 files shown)


glennw:
r? @kvark

There's a few subtle bits to this code, and some of the text permutations don't have great test coverage, so please look over carefully :)

Having said that, I did some manual testing in Gecko and didn't notice anything obviously wrong.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=1619606ce2e4714932ca500fa6fb9fba69cafe2f

glennw:
Review status: 5 of 13 files reviewed at latest revision, 3 unresolved discussions.

---

*[webrender/res/brush.glsl, line 161 at r1](https://beta.reviewable.io/reviews/servo/webrender/2677#-LAYomJ02slZL4l3IqAd:-LAZpvZ_6wZn_nu3YcP6:byglz19) ([raw file](https://github.com/servo/webrender/blob/8097ac058090d16e062361b1a4a290202b5a6d67/webrender/res/brush.glsl#L161)):*

can we please avoid `out` parameters in functions?
instead - have a struct returned, and the define would only wrap around one of the struct members
</blockquote></details>

I avoided it thinking we had shader compiler issues in the past, but that was with structures for interpolants. I've updated to use a Fragment structure.

---

*[webrender/res/ps_text_run.glsl, line 71 at r1](https://beta.reviewable.io/reviews/servo/webrender/2677#-LAYpHVnBVb5lMtTLd0q:-LAZq3Y_2Z2O57BnVJGP:batsk0d) ([raw file](https://github.com/servo/webrender/blob/8097ac058090d16e062361b1a4a290202b5a6d67/webrender/res/ps_text_run.glsl#L71)):*

so, correct me if I'm wrong, but having the color mode to come from the uniform allows us to re-use the instance list for a number of different calls,  so `uMode` is still needed?
</blockquote></details>

Yup, that's right - so we place a color mode in the instance for every case except the SubpixelBg 3-pass technique. So we get better batching for most cases, but retain the ability to override with a uniform value when we are drawing a multi-pass technique by submitting the same geometry multiple times.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/webrender/2677)*

glennw:
@kvark Thanks for the quick review! Comments addressed.

webrender/res/brush_image.glsl
```diff
@@ -232,14 +236,23 @@ vec4 brush_fs() {
...
+    Fragment frag;
+
 #ifdef WR_FEATURE_ALPHA_PASS
     float alpha = init_transform_fs(vLocalPos);
     texel.rgb = texel.rgb * vMaskSwizzle.x + texel.aaa * vMaskSwizzle.y;
-    vec4 color = vColor * texel * alpha;
+
+    #ifdef WR_FEATURE_DUAL_SOURCE_BLENDING
+        vec4 alpha_mask = texel * alpha;
+        frag.color = vColor * alpha_mask;
```
kvark: let's unify these paths a little:
```
vec4 alpha_mask = texel * alpha;
frag.color = vColor * alpha_mask;
#ifdef WR_FEATURE_DUAL_SOURCE_BLENDING
frag.blend = alpha_mask * vColor.a;
#endif
```

glennw: Well spotted, fixed!

PR #2679, opened 2018-04-21
Remove "Layer" space

### LINKED ISSUE #2672: Delete "Layer" space from the codebase
> \<nical> Gankro: (re: Layer/Layout) back then the terminology in place was "layer" in webrender, but it didn't make much sense for the API where turned out to be layout coordinates. We weren't sure whether the layout coordinates at the API boundary were going to remain the same as the "layer" coordinates used internally so we just kept both names with a typedef.
>
> \<Gankro> nical: can we kill it with fire now?
>
> \<nical> yeah
> let's just call it all layout

nc4rrillo:
Fixes #2672

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2679)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -3,9 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use api::{AlphaType, ClipMode, DeviceIntRect, DeviceIntSize};
-use api::{DeviceUintRect, DeviceUintPoint, DeviceUintSize, ExternalImageType, FilterOp, ImageRendering, LayerRect};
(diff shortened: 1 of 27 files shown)


glennw:
Thanks, looks good to me. I built Gecko and it compiles and runs with this PR, so :rocket: 

@bors-servo r+

staktrace:
For the record, this did break the Gecko build.

glennw:
(from discussion with kats on IRC)

The reason I was able to build locally in this case is that I didn't do a FFI bindings regeneration step, so it just happened to work since the memory layout between the two types is exactly the same. Sorry!

PR #2681, opened 2018-04-23
Robustify shutdown for the scene builder thread

staktrace:
Currently the scene builder thread can be processing a scene when the RB
thread shuts down. In this case, the scene builder thread may end up
trying to unwrap() the send result from the
SceneBuilderResult::Transaction, but that would fail and panic. Although one
alternative is to simply remove that unwrap() call, it seems better to
have a proper shutdown sequence and assert that things are happening as
we expect them to.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2681)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -731,6 +731,9 @@ impl RenderBackend {
                                 &mut profile_counters
                             );
                         }
+                    },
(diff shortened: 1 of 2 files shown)


glennw:
I'm a bit out of the loop with the scene building code, but this seems reasonable.

@bors-servo r+

PR #2682, opened 2018-04-24
Run reftests on Windows CI.

glennw:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2682)


glennw:
This is just a testing PR for experimenting with running the reftests on Windows CI + ANGLE.

Once we get the basics running in the CI environment, we'll work out which tests need to be fuzzed / disabled on Windows for now.

glennw:
I think this PR is ready for review now.

It disables ~20% of the reftests that use PNG images as a reference. I verified manually in the reftest analyzer that each of these differences are very small pixel differences related to rasterizer differences.

In the future, we'll need to either work out a way to drastically reduce the number of PNG references we use, or more likely, support reference image generation per platform (possibly with an automated system to re-generate reference images).

In the mean time, this gives us a lot more Windows coverage than we had previously, so I think it's worth enabling now.

r? @jrmuizel

PR #2683, opened 2018-04-24
Clip dots/dashes to task rect

mrobinson:
This ensures that when rendering dots and dashes that are clipped, they
don't render outside of the task rectangle. This should increase
performance (since dots and dashes outside the task rectangle produce
degenerate triangles) and avoid corrupting nearby areas of the
RenderTarget texture.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2683)

### DIFF
--- a/webrender/res/cs_clip_border.glsl
+++ b/webrender/res/cs_clip_border.glsl
@@ -134,9 +134,16 @@ void main(void) {
     vec2 device_pos = world_pos.xy * uDevicePixelRatio;
 
     // Position vertex within the render task area.
-    vec2 final_pos = device_pos -
-                     area.screen_origin +
-                     area.common_data.task_rect.p0;
+    vec2 task_rect_origin = area.common_data.task_rect.p0;
+    vec2 final_pos = device_pos - area.screen_origin + task_rect_origin;
+
+    // Clamp the mask rectangle to the render task area.
+    final_pos = clamp(final_pos,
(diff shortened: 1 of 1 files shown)


mrobinson:
There is a pending Gecko try job here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=692d39841ca1469d3b3e915f0b4cef356468af8a

I've been having a little trouble with the infrastructure (builds timing out), but hopefully this one is successful.

mrobinson:
The try run is looking good, apart from one intermittent which I believe is unrelated.

glennw:
Does this work in the case of a transform?

I think that the world rect of the corner could in theory end up crossing two edges of the task rect, and now be possible to clip correctly via a VS clamp because of that?

I could easily be missing something though, haven't looked in depth.

glennw:
Hmm, maybe that doesn't matter as long as we take a bounding rect that will include enough pixels, which this PR might handle? Hmm, I might be overthinking this :grimacing:

mrobinson:
That's an interesting question! Let me do some tests and come back with a solid answer.

mrobinson:
This does seem to have an issue with transformations. I'm working through those issues now.

glennw:
@mrobinson Do we still want this one open?

mrobinson:
Nope! This should be closed now.

PR #2685, opened 2018-04-24
Add a mechanism to flush the scene builder thread

staktrace:
We need this for things like reftests in Gecko where we need to ensure
any async scene builds are done before we take a snapshot.

r? @nical or @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2685)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -732,6 +732,9 @@ impl RenderBackend {
                             );
                         }
                     },
+                    SceneBuilderResult::FlushComplete(tx) => {
+                        tx.send(()).ok();
+                    }
                     SceneBuilderResult::Stopped => {
                         panic!("We haven't sent a Stop yet, how did we get a Stopped back?");
                     }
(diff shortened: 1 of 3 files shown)


webrender/src/render_backend.rs
```diff
@@ -754,6 +757,13 @@ impl RenderBackend {
         // inflight messages, otherwise the scene builder might panic.
         while let Ok(msg) = self.scene_rx.recv() {
             match msg {
+                SceneBuilderResult::FlushComplete(tx) => {
+                    // If somebody's blocked waiting for a flush, how did they
+                    // trigger the RB thread to shut down? This shouldn't happen
+                    // but handle it gracefully anyway.
+                    debug_assert!(false);
```
glennw: This would be better as ```unreachable!("bug: [reasons]");```

webrender/src/render_backend.rs
```diff
@@ -754,6 +757,13 @@ impl RenderBackend {
...
+                    let _ = tx.send(());
```
glennw: Another way to write that you don't care about the result here is ```tx.send(()).ok();```. I'm not sure which is actually more preferred, so feel free to leave as is if you like.

webrender/src/render_backend.rs
staktrace: unreachable! will panic in release builds, right? I'd rather avoid that in this case because there is a reasonable way to handle it gracefully. That's why I went with the debug_assert instead. This scenario could occur if somebody uses WR with multiple API instances one of which is used to shut down the RB thread while the other is blocked on a flush. I don't believe it's possible with either gecko or servo at the moment, but other WR consumers might want to do that. At the same time, if that scenario does occur we might want to revisit this.

webrender/src/render_backend.rs
staktrace: Ah, this is cleaner. Will do this, thanks!

staktrace:
Mac CI says: `fatal: reference is not a tree: 367eafdaad18b0aa7581dfb904d041d6e576565c` which is weird since the Linux CI worked fine on the same revision.

@bors-servo retry

PR #2687, opened 2018-04-24
Generate tight bounding rects for dotted corners

mrobinson:
Instead of producing a quad the size of the entire mask for every dot,
generate a tight bounding rect. This seems to increase performance a
great deal on my machine.

Given the test case
```yaml
---
root:
  items:
    -
      bounds: [16, 16, 36720, 50]
      "clip-rect": [16, 16, 36720, 50]
      type: border
      width: 6
      "border-type": normal
      color: black
      style: dotted
      radius: [18360, 25]
```
...this change reduces GPU time from 24ms to 1.7ms and brings framerate up from 37fps to over 60fps.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2687)

### DIFF
--- a/webrender/res/cs_clip_border.glsl
+++ b/webrender/res/cs_clip_border.glsl
@@ -71,6 +71,11 @@ void main(void) {
     BorderCorner corner = fetch_border_corner(cmi.clip_data_address);
     vClipCenter = corner.clip_center;
 
+    // Get local vertex position for the corner rect.
+    // TODO(gw): We could reduce the number of pixels written here by calculating a tight
+    // fitting bounding box of the dash itself like we do for dots below.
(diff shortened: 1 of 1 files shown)


mrobinson:
Okay. Looks like most of the redness was due to a shader validation issue. Now there is seemingly only one test with a small difference (not visible to the eye, as far as I can tell). I'll see if I can eliminate it tomorrow though.

mrobinson:
The latest version of this patch should fix these issues. The changes I made were to:
* Fix the shader validation issue.
* Ensure that the dot rectangle is clamped to the border corner rectangle. It looks like dots naturally escape the border rectangle, perhaps pointing to a small issue with dot positioning.

mrobinson:
(this should be ready to review)

PR #2690, opened 2018-04-25
Cleanups

waywardmonkeys:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2690)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -145,7 +145,7 @@ impl ClipScrollNode {
             parent: parent_index,
             children: Vec::new(),
             pipeline_id,
-            node_type: node_type,
+            node_type,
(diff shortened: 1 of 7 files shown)


webrender_api/src/api.rs
```diff
@@ -40,9 +40,7 @@ pub enum ResourceUpdate {
 
 impl ResourceUpdates {
     pub fn new() -> Self {
-        ResourceUpdates {
-            updates: Vec::new(),
-        }
+        Self::default()
```
Gankro: While technically more compact, this really just seems like it's making the code more opaque and harder to change later, especially since we have no use for the actual Default impl otherwise. I'd rather not land this particular change.

waywardmonkeys:
I've removed that commit from this patch series, which should address the comment.

PR #2691, opened 2018-04-25
Make the ready callback more consistent

staktrace:
For gecko we need to ensure that we get a callback for every single
transaction that has a generate_frame flag set, because Gecko uses this
callback as part of a mechanism to throttle frame requests. If the
callback never arrives, it gets stuck thinking there are still inflight
frame requests, and doesn't request more.

This patch slightly modifies the semantics of the callback to ensure it
meets Gecko's needs. The main semantic change is that it may now get
called with both the `scrolled` and `composite` arguments as false,
whereas before at least one of these was always true.

This patch also renames the callback to better match the new semantics.
The new name is actually the original name of the callback, as can be
seen from the out-of-date documentation, which I'm also fixing in this
patch.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2691)

### DIFF
--- a/direct-composition/src/main_windows.rs
+++ b/direct-composition/src/main_windows.rs
@@ -186,7 +186,7 @@ impl api::RenderNotifier for Notifier {
         let _ = self.events_proxy.wakeup();
     }
 
-    fn new_document_ready(&self, _: api::DocumentId, _: bool, _: bool) {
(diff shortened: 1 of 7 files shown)


staktrace:
r? anybody who feels comfortable with this

This is needed for https://bugzilla.mozilla.org/show_bug.cgi?id=1456978

glennw:
Seems reasonable to me.

@bors-servo r+

PR #2692, opened 2018-04-25
Update Pathfinder integration for the texture cache changes.

pcwalton:
This also contains a small fix that I needed to compile on the current Rust nightly compiler. It should not impact compilation on stable unless the `serde` Cargo feature is used.

r? @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2692)

### DIFF
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -37,9 +37,9 @@ use rayon::prelude::*;
 use render_backend::FrameId;
 use render_task::{RenderTaskCache, RenderTaskTree};
 #[cfg(feature = "pathfinder")]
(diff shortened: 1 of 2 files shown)


glennw:
The texture cache change looks good to me. I'll get @Gankro to double check the custom attribute change - does that cause any issues with Gecko?

glennw:
CI error:

```
error[E0554](B: #![feature] may not be used on the stable release channel(B
  (B--> (Bwebrender/src/lib.rs:43:32(B
   (B|(B
43(B (B| (B#![cfg_attr(feature = "serde", feature(custom_attribute))](B
   (B| (B                               (B^^^^^^^^^^^^^^^^^^^^^^^^^^(B
error(B: aborting due to previous error(B

```

PR #2694, opened 2018-04-26
Allow texture cache entries to store non-rectangular UVs.

glennw:
Most of the time, UVs are stored as a (top left, bottom right) pair
of coordinates in the texture cache. The exception to this is when
we have an off-screen render target that was transformed.

In this case, we need access to the four corners within the render
surface in order to generate correct UVs when we composite that
surface into the framebuffer.

Previously, we generated these in the brush shader by transforming
the vertex position, and using that to generate a screen space
texture coordinate.

This patch removes that code, and instead does the calculation of
the four corner UVs during render task generation on the CPU. The
corner coordinates are then stored in the texture cache entry, and
the shaders do a simple bilerp to find the correct UV at each vertex
when drawing these images.

This has a number of benefits:
 * We can draw a screen-space rendered surface at any location now
   with the same logic (we don't need hacks for applying local offsets
   etc when drawing drop-shadows).
 * Most importantly, when we store an off-screen surface in the texture
   cache for use on subsequent frames, the shader that draws that doesn't
   need to know anything about the transform of the surface when it was
   rendered. This means the screen-space UV generation doesn't rely
   on access to the clip-scroll node transform of the source surface.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2694)


glennw:
Although this has general benefits for drop-shadows etc, the primary motivation is making it simpler to cache screen-space rendered surfaces between frames and display lists.

Pending try here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=dd8e97d7887ad6afea4bd8209face2bf1636c39b

r? @mrobinson or @kvark

glennw:
Hmmm, quite a lot of orange on the try run.

I'm not sure if they are related to this PR or something else that got merged today.

I'll investigate further in the morning.

Darkspirit:
mozregression --repo try --launch dd8e97d7887ad6afea4bd8209face2bf1636c39b --pref gfx.webrender.all:true startup.homepage_welcome_url:'https://www.mozilla.org/de/'



screenshot_20180426_154341

glennw:
@Darkspirit Thank you! I'll investigate this regression and the try issue this morning.

glennw:
Try run looks good with the fix above:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=449c4c3f81c9120c5c39548a24e9acb39d68a31a

r? @mrobinson

Issue #2689, opened 2018-04-25
async scene building breaks tile decomposition rawtests


jrmuizel:
Running the rawtests with a patch like this:

```
diff --git a/webrender_api/src/api.rs b/webrender_api/src/api.rs
index 35a123bc..115534ba 100644
--- a/webrender_api/src/api.rs
+++ b/webrender_api/src/api.rs
@@ -151,21 +151,21 @@ pub struct Transaction {
     generate_frame: bool,
 }

 impl Transaction {
     pub fn new() -> Self {
         Transaction {
             scene_ops: Vec::new(),
             frame_ops: Vec::new(),
             resource_updates: ResourceUpdates::new(),
             payloads: Vec::new(),
-            use_scene_builder_thread: false, // TODO: make this true by default.
+            use_scene_builder_thread: true, // TODO: make this true by default.
             generate_frame: false,
         }
     }
```

gives
```
Shader override path: None
	hit testing test...
	retained blob images test...
	blob update test...
	blob update epoch test...
	tile decomposition...
thread 'WRRenderBackend#0' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
@nical @staktrace

staktrace:
It might be the same problem as #2686, although I wouldn't expect tile decomposition to switch an image between tiled and non-tiled? I don't really know this code, but nical said he'd look into this problem when he's back from training (so presumably tomorrow).

nical:
I had a look and what's actually happening here is that with async transactions we apply resource updates *after* scene building which makes sense, but we also rely on knowing which images are tiled during scene building to decompose them into several primitives if need be.

So this isn't #2686 although i suspect it is the real issue that made us realize we can have issues with images that change tiling when they are resized (without being recreated).

I am working on a patch to move decomposition out of scene building which should fix this.

nical:
I checked and this is indeed fixed.

PR #2695, opened 2018-04-26
Decompose tiled images during frame building (v3).

nical:
This PR implements decomposing tiled images during frame building using visibility information to avoid issues with very large blob images.
This version does not use segments, avoiding issues with the maximum contiguous gpu cache alloc size as well as a few hacks that I had to write to get repetitions to work with segments.

As a bonus the non-brush image primitive and ps_image shader are now obsolete so I removed them.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2695)


nical:
Fixes #2689.

glennw:

---

*[webrender/src/batch.rs, line 617 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2vCLjEoKTa7Lwy_ln:-LB2vCLjEoKTa7Lwy_lo:b-y96u58) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/batch.rs#L617)):*
> ```Rust
>         // Tiled images have a cache handle per visible tile and an invalid handle in the
>         // primitive metadata.
>         let prim_cache_address = gpu_cache.try_get_address(&prim_metadata.gpu_location);
> ```

I think we'd be better to match on the prim kind here. Having a fallible call here may mask bugs for other primitive types, if the cache address is not valid when it's expected to be.

---

*[webrender/src/device.rs, line 1253 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2txbb0PxJQjKuWXO3:-LB2txbb0PxJQjKuWXO4:b-bl3flw) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/device.rs#L1253)):*
> ```Rust
>             dest_rect.origin.x,
>             dest_rect.origin.y,
>             dest_rect.origin.x + src_rect.size.width,
> ```

I don't understand this change. This would mean that we can never do a blit which is a stretch?

---

*[webrender/src/gpu_cache.rs, line 660 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2vVDY5alEivtKhhHI:-LB2vVDY5alEivtKhhHJ:bv4xo0p) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/gpu_cache.rs#L660)):*
> ```Rust
>     /// Similar to get_address except that it accepts invalid cache handles in which case it
>     /// returns an invalid cache address.
>     pub fn try_get_address(&self, id: &GpuCacheHandle) -> GpuCacheAddress {
> ```

We can remove this, given the comment above about this method call.

---

*[webrender/src/prim_store.rs, line 1360 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2x8Rv9dTDZ0wT0dLs:-LB2x8Rv9dTDZ0wT0dLt:bdar4v3) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/prim_store.rs#L1360)):*
> ```Rust
>                                                     request.push(*tile_rect);
>                                                     request.push(tight_clip_rect);
>                                                     request.write_segment(*tile_rect, [1.0, 1.0, 0.0, 0.0]);
> ```

The GPU cache format for an image brush has changed recently - in between the prim rect and the segment info are two colors - they can just be set to PremultipliedColorF::WHITE in this case. We should abstract the code to write a brush image GPU request into a function (doesn't need to be done this PR though).

---

*[webrender/src/prim_store.rs, line 2162 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2ww_UAla77z29D2EF:-LB2ww_UAla77z29D2EG:b9bdier) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/prim_store.rs#L2162)):*
> ```Rust
> }
> 
> fn compute_conservatrive_visible_rect(
> ```

nit: conservative spelling

---

*[webrender/src/prim_store.rs, line 2183 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2x1jE3eIytMbZ966i:-LB2x1jE3eIytMbZ966j:b-xb916y) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/prim_store.rs#L2183)):*
> ```Rust
> 
> fn edge_flags_for_tile_spacing(tile_spacing: &LayerSize) -> EdgeAaSegmentMask {
>     // If the number of gpu blocks for the request changes we can't reuse
> ```

This comment doesn't seem relevant here?

---

*[wrench/reftests/transforms/local-clip.png, line 0 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2v1pI1mplsgJ9coCZ:-LB2v1pI1mplsgJ9coC_:b-a8asis) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/wrench/reftests/transforms/local-clip.png#L0)):*
Do we know why this reference image changed?

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2695)*

glennw:
@nical Amazing PR, thanks!

This looks great to me - it will be fantastic to finally remove the legacy image primitive.

There are a few comments from reviewable above, but they are all minor details. The sooner we can get this merged the better, since it will bitrot easily. Things left to complete below:

- [ ] Address the reviewable comments above.
- [ ] Rebase.
- [ ] Gecko try run.
- [ ] Merge!

nical:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[webrender/src/prim_store.rs, line 2183 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2x1jE3eIytMbZ966i:-LB5WRqOAGf5dVMkXYxq:bxefj9u) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/prim_store.rs#L2183)):*

This comment doesn't seem relevant here?
</blockquote></details>

Oops, copy-pasta accident.

---

*[wrench/reftests/transforms/local-clip.png, line  at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2v1pI1mplsgJ9coCZ:-LB5WaoQ94vUwq7L6Rrb:bux1hgh) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/wrench/reftests/transforms/local-clip.png#L)):*

Do we know why this reference image changed?
</blockquote></details>

This shouldn't be part of the final PR. I updated this reference image after changing the way the antialiasing is clipped in the image brush shader but this was in the PR about support repetitions in brush shaders and ended up in this commit queue by accident (this branch has been rebased many times). This PR will be quite a bit smaller once I have merged the two PRs that it sits on top of (#2644 and #2664).

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2695)*

nical:
*[webrender/src/prim_store.rs, line 1360 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2x8Rv9dTDZ0wT0dLs:-LB5sMCx5Rdf2eNQbq12:b-t792h) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/prim_store.rs#L1360)):*

The GPU cache format for an image brush has changed recently - in between the prim rect and the segment info are two colors - they can just be set to PremultipliedColorF::WHITE in this case. We should abstract the code to write a brush image GPU request into a function (doesn't need to be done this PR though).
</blockquote></details>

I don't see where these colors are being added currently. Has this changed landed?

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2695)*

nical:
*[webrender/src/prim_store.rs, line 1360 at r1](https://reviewable.io/reviews/servo/webrender/2695#-LB2x8Rv9dTDZ0wT0dLs:-LB5v7dCE5VN2pUZUakd:bvmut1m) ([raw file](https://github.com/servo/webrender/blob/534631233626f5ad628b0d2b197381b99db7e19b/webrender/src/prim_store.rs#L1360)):*

I don't see where these colors are being added currently. Has this changed landed?
</blockquote></details>

Err nevermind, I hadn't properly rebased.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2695)*

PR #2696, opened 2018-04-26
Clip to task rect while rendering dot/dash corner masks

mrobinson:
We can use the task rectangle to do fragment shader clipping while
rendering dots and dashes for border corners. This prevents dot and
dash masks from rendering outside of the task rectangle.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2696)


mrobinson:
Gecko try job is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=0cb612b3b9489b6e39115813a5d445d631a881bd

PR #2699, opened 2018-04-26
Panic more verbosely if we don't find the descriptor

jrmuizel:
We're running into a panic on unwrap() in
https://bugzilla.mozilla.org/show_bug.cgi?id=1455848
This should help us find out more information.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2699)

### DIFF
--- a/webrender/src/platform/windows/font.rs
+++ b/webrender/src/platform/windows/font.rs
@@ -131,7 +131,10 @@ impl FontContext {
         }
 
         let system_fc = dwrote::FontCollection::system();
-        let font = system_fc.get_font_from_descriptor(&font_handle).unwrap();
+        let font = match system_fc.get_font_from_descriptor(&font_handle) {
+            Some(font) => font,
+            None => { panic!("missing descriptor {:?}", font_handle) }
+        };
         let face = font.create_font_face();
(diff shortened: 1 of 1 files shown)


webrender/src/platform/windows/font.rs
```diff
@@ -131,7 +131,10 @@ impl FontContext {
         }
 
         let system_fc = dwrote::FontCollection::system();
-        let font = system_fc.get_font_from_descriptor(&font_handle).unwrap();
+        let font = match system_fc.get_font_from_descriptor(&font_handle) {
```
glennw: This would be better written as:

```
let font = system_fc
  .get_font_from_descriptor(&font_handle)
  .expect(&format!("missing descriptor {:?}", font_handle));
```

jrmuizel: Yeah it would

glennw: Fair point :)

lqd: Maybe then this ?
```
let font = system_fc
  .get_font_from_descriptor(&font_handle)
  .unwrap_or_else(|| panic!("missing descriptor {:?}", font_handle));
```

PR #2701, opened 2018-04-27
Allow subpixel with specified background color in offscreen targets.

### LINKED ISSUE #2670: BlendMode::SubpixelWithBgColor text rendering should still be allowed inside transparent groups
With the default theme on Mac, overflowing tab titles no longer use subpixel AA. However, the text in the tab titles uses a font smoothing background color, and `BlendMode::SubpixelWithBgColor` is able to render onto transparency without any problems. So I think we need to adjust the logic that disables subpixel AA in this case, so that SubpixelWithBgColor text blending is still allowed.

glennw:
When a subpixel text run has a specified background color, we can
safely draw it into an off-screen surface.

Fix a bug in the renderer where the shader color mode was being
set before the shader was bound.

Add support in wrench for specifying the background color of
a text run.

Add a reftest that checks offscreen surface + bg color subpixel
text is the same as subpixel text on a normal background.

Fixes #2670.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2701)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -728,7 +728,7 @@ impl TextRunPrimitiveCpu {
         display_list: &BuiltDisplayList,
         frame_building_state: &mut FrameBuildingState,
     ) {
-        if !allow_subpixel_aa {
+        if !allow_subpixel_aa && self.font.bg_color.a == 0 {
             self.font.render_mode = self.font.render_mode.limit_by(FontRenderMode::Alpha);
         }
 
--- a/webrender/src/glyph_rasterizer.rs
+++ b/webrender/src/glyph_rasterizer.rs
@@ -224,6 +224,10 @@ impl FontInstance {
         platform_options: Option<FontInstancePlatformOptions>,
         variations: Vec<FontVariation>,
     ) -> Self {
+        // If a background color is enabled, it only makes sense
+        // for it to be completely opaque.
+        debug_assert!(bg_color.a == 0 || bg_color.a == 255);
+
         FontInstance {
             font_key,
             size,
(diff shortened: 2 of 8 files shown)


glennw:
r? anyone

(will kick off a try run shortly)

glennw:
Pending try:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=491dbd606f5d76858e5e865a2c6449c74ca612c0

glennw:
Try run looks :green_apple:

webrender/src/prim_store.rs
```diff
@@ -726,7 +726,7 @@ impl TextRunPrimitiveCpu {
         display_list: &BuiltDisplayList,
         frame_building_state: &mut FrameBuildingState,
     ) {
-        if !allow_subpixel_aa {
+        if !allow_subpixel_aa && self.font.bg_color.a == 0 {
```
Gankro: shouldn't this be `!= 1.0` ?

Gankro:
r=me with comment addressed.

webrender/src/prim_store.rs
glennw: We use the ```!= 0``` check in a few places, and the comment where the field is declared says:

```
    /// When bg_color.a is != 0 and render_mode is FontRenderMode::Subpixel,
    /// the text will be rendered with bg_color.r/g/b as an opaque estimated
    /// background color.
```

So I think it's fine to leave it as is? Perhaps I should add an assert that the alpha in this case is either 0 or 255 though, since any other value doesn't really make sense here?

Gankro: yeah, the assert sounds like a good compromise!

glennw: OK, rebased + added an assert in ```FontInstance::new```.

PR #2702, opened 2018-04-27
Support non-square slab allocator pages in texture cache.

glennw:
Text runs are typically rectangles that are much longer in one
dimension than the other dimension (depending on horizontal or
vertical layout direction).

Storing rectangles of this shape in the texture cache was very
inefficient, since only square allocations are supported.

This patch adds support for texture cache regions to support
non-square slab sizes.

This is an efficiency win for any image that is rectangular,
and will become particularly important once we start caching
text runs in the texture cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2702)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -639,7 +639,7 @@ impl TextureCache {
         // - We have freed an item that will definitely allow us to
         //   fit the currently requested allocation.
         let needed_slab_size =
-            SlabSize::new(required_alloc.width, required_alloc.height).get_size();
+            SlabSize::new(required_alloc.width, required_alloc.height);
         let mut found_matching_slab = false;
         let mut freed_complete_page = false;
         let mut evicted_items = 0;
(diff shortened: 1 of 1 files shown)


glennw:
r? @mrobinson or anyone

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=acae109e956c21ccc502c372970a72d8d3b66439

webrender/src/texture_cache.rs
```diff
@@ -900,43 +901,49 @@ impl TextureCache {
...
-            _ => panic!("Invalid dimensions for cache!"),
+        let x_size = quantize_dimension(width);
+        let y_size = quantize_dimension(height);
+
+        assert!(x_size > 0 && x_size <= TEXTURE_REGION_DIMENSIONS);
+        assert!(y_size > 0 && y_size <= TEXTURE_REGION_DIMENSIONS);
+
+        let (width, height) = match (x_size, y_size) {
+            // Special cased rectangular slab pages.
+            (512, 512) => (512, 512),
```
mrobinson: It seems like this line can be removed.

mrobinson: It seems like this line can be removed, since it is just the square case?

glennw: Good point! Fixed.

PR #2703, opened 2018-04-27
Improve the scene swap for gecko

staktrace:
This is a better version of #2688 and fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1457470 as well.

r? @nical 

I'm not super happy with having to pass the extra channel back in the second patch but I couldn't come up with a better way to do this. Suggestions welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2703)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -264,7 +264,6 @@ impl Document {
             frame_ops: transaction_msg.frame_ops,
             render: transaction_msg.generate_frame,
             document_id,
-            current_epochs: self.current.scene.pipeline_epochs.clone(),
         }).unwrap();
     }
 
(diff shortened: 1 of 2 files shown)


webrender/src/render_backend.rs
```diff
@@ -706,12 +705,22 @@ impl RenderBackend {
                                 doc.new_async_scene_ready(built_scene);
                                 doc.render_on_hittest = true;
                             }
-                            result_tx.send(SceneSwapResult::Complete).unwrap();
+                            if let Some(tx) = result_tx {
+                                let (resume_tx, resume_rx) = channel();
+                                tx.send(SceneSwapResult::Complete(resume_tx)).unwrap();
+                                // Block until the post-swap hook has completed on
```
nical: I suppose we'll have to do with this for now but I suspect that it'll introduce long latencies on busy CPUs. We're seeing similar issues in WebRTC code right now where an audio callback sits idle for 10ms because it tried to take a mutex that was held for 0.1ms by another thread and isn't scheduled for the next 9.9ms because the scheduler chose to run other things (which is terrible in an audio callback, and not great in rendering).

staktrace: Hm, that seems pretty bad. We can certainly do a yield from the scene builder thread after sending the resume message that wakes up the RB thread, but whether or not it helps will depend on the CPU load and scheduler implementation details.

Issue #2645, opened 2018-04-11
webrender does not compile on Windows with feature=pathfinder


fschutt:
After the latest pathfinder (currently https://github.com/pcwalton/pathfinder/commit/235c51e160ed651b3e8e273e14d9a66b9c9d114b) changes have been merged, webrender does not compile on Windows anymore:

```rust
   Compiling pathfinder_gfx_utils v0.1.0 (https://github.com/pcwalton/pathfinder#235c51e1)
   Compiling plane-split v0.8.0
   Compiling lyon_geom v0.10.0
   Compiling webrender_api v0.57.2 (https://github.com/servo/webrender?rev=1ff66239b728083a37c53c98d51e070d4aae8ca0#1ff66239)
   Compiling lyon_path v0.10.0
   Compiling pathfinder_font_renderer v0.3.0 (https://github.com/pcwalton/pathfinder#235c51e1)
   Compiling pathfinder_path_utils v0.1.0 (https://github.com/pcwalton/pathfinder#235c51e1)
   Compiling pathfinder_partitioner v0.1.0 (https://github.com/pcwalton/pathfinder#235c51e1)

error[E0061]: this function takes 2 parameters but 3 parameters were supplied
   --> C:\Users\Felix Schütt\.cargo\git\checkouts\webrender-c3596abe1cf4f320\1ff6623\webrender\src\glyph_rasterizer.rs:579:55
    |
579 |                         match pathfinder_font_context.glyph_dimensions(&pathfinder_font_instance,
    |                                                       ^^^^^^^^^^^^^^^^ expected 2 parameters

error[E0308]: mismatched types
   --> C:\Users\Felix Schütt\.cargo\git\checkouts\webrender-c3596abe1cf4f320\1ff6623\webrender\src\glyph_rasterizer.rs:582:29
    |
582 |                             Ok(glyph_dimensions) => glyph_dimensions,
    |                             ^^^^^^^^^^^^^^^^^^^^ expected enum `std::option::Option`, found enum `std::result::Result`
    |
    = note: expected type `std::option::Option<pathfinder_font_renderer::GlyphDimensions>`
               found type `std::result::Result<_, _>`

error[E0308]: mismatched types
   --> C:\Users\Felix Schütt\.cargo\git\checkouts\webrender-c3596abe1cf4f320\1ff6623\webrender\src\glyph_rasterizer.rs:583:29
    |
583 |                             Err(_) => continue,
[...]

PR #2705, opened 2018-04-28
Fix From conversion from NativeFontHandleWrapper to PathfinderComPtr

fschutt:
See https://github.com/pcwalton/pathfinder/pull/81 and https://github.com/servo/webrender/issues/2645 - this fixes 1 compile error, where we can't convert from a NativeFontHandleWrapper to a ComPtr. We need this, so that PathfinderFontContext::add_native_font() works correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2705)


webrender/src/platform/windows/font.rs
```diff
@@ -439,3 +443,16 @@ impl FontContext {
...
     }
 }
+
+#[cfg(feature = "pathfinder")]
+impl<'a> From<NativeFontHandleWrapper<'a>> for PathfinderComPtr<IDWriteFontFace> {
+    fn from(font_handle: NativeFontHandleWrapper<'a>) -> Self {
+        let system_fc = ::dwrote::FontCollection::system();
+        let font = match system_fc.get_font_from_descriptor(&font_handle.0) {
+            Some(font) => font,
+            None => { panic!("missing descriptor {:?}", font_handle.0) }
```
pcwalton: nit: no need for {} here

fschutt:
@pcwalton alright, fixed it

PR #2709, opened 2018-04-30
Output gradientstops in serialization and debug output

### LINKED ISSUE #2328: SetGradientStops missing with debug-serialization

Gankro:
Fixes #2328 (I think?)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2709)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -236,35 +236,46 @@ impl<'a> BuiltDisplayListIter<'a> {
         self.cur_clip_chain_items = ItemRange::default();
 
         loop {
-            if self.data.is_empty() {
-                return None;
+            self.next_raw();
+            if let SetGradientStops = self.cur_item.item {
+                // SetGradientStops is a dummy item that most consumers should ignore
+                continue;
             }
-
-            {
-                let reader = bincode::IoReader::new(UnsafeReader::new(&mut self.data));
-                bincode::deserialize_in_place(reader, &mut self.cur_item)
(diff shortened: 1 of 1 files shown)


webrender_api/src/display_list.rs
```diff
@@ -236,35 +236,46 @@ impl<'a> BuiltDisplayListIter<'a> {
         self.cur_clip_chain_items = ItemRange::default();
 
         loop {
-            if self.data.is_empty() {
-                return None;
+            self.next_raw();
+            if let SetGradientStops = self.cur_item.item {
+                // SetGradientStops is a dummy item that most consumers should ignore
```
kvark: can we leave this decision to the clients, given that they do know which items they need and which they don't?

kvark:
I do confirm the PR fixes the issue with capturing gradients :+1: :tada:

webrender_api/src/display_list.rs
Gankro: They can just use `_raw` if they want to make that decision, right? This method is for the "intended" case, where SetGradientStops is really just a quirk of our encoding.

PR #2710, opened 2018-05-01
Basic chasing infrastructure

kvark:
While the capturing infrastructure allows users to inspect the scene and frame level state of the render pipeline, it's not able to answer the question of how does an item make it's way from one way to another. Or, often, where and why does it get culled. This is what the chasing is for: given a property of an item (currently, the local rectangle is supported), it tracks it through the pipeline and prints out all the paths taken, e.g.;
```
Chasing PrimitiveIndex(298)
	preparing a run of length 1 in pipeline PipelineId(1, 1)
	updating clip task with screen rect TypedRect(1712×32 at (0,29))
	base combined outer rect Some(TypedRect(1712×32 at (0,29)))
	need no task: no clips are left
	considered visible and ready with local rect TypedRect(1712.0×32.0 at (0.0,29.0))
```

This is something I've been needing for a few cases already, and I suspect will need more in the future.
Let me know if you have any ideas on how to make it better, or how to accomplish the task without this code at all ;)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2710)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -17,7 +17,7 @@ use clip::{ClipRegion, ClipSource, ClipSources, ClipStore};
 use clip_scroll_node::{ClipScrollNode, NodeType, StickyFrameInfo};
 use clip_scroll_tree::{ClipChainIndex, ClipScrollNodeIndex, ClipScrollTree};
 use euclid::{SideOffsets2D, vec2};
-use frame_builder::{FrameBuilder, FrameBuilderConfig};
+use frame_builder::{ChasePrimitive, FrameBuilder, FrameBuilderConfig};
 use glyph_rasterizer::FontInstance;
 use gpu_cache::GpuCacheHandle;
 use gpu_types::BrushFlags;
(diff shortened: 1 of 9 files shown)


glennw:
I *do* like the idea of this, I can see it being useful.

It feels like it might be fragile to keep in sync with WR changes though, not that I have any better ideas on how to implement such a feature.

glennw:
Just a thought - would an implementation be any simpler if we stored the ```chase``` flag internally with the data in question? For example, have a ```chase``` flag on the base display item, that gets propagated to prim store items, that gets propagated to batch structures etc?

glennw:
It feels like that *might* make it a bit easier to keep in sync, but that's just a guess.

nical:
I can think of a lot of times where something like this would have been very useful to me. Looks like a few simple macros could reduce the visual noise of adding branches everywhere, fro example turning

```rust
if cfg!(debug_assertions) && Some(prim_index) == self.chase_id {
    println!("\tupdating clip task with screen rect {:?}", prim_screen_rect);
}
```

into

```rust
// or track_prim_index?
chase_prim_index!(self, prim_index, "\tupdating clip task with screen rect {:?}", prim_screen_rect);
```

nical:
I didn't mean this in a meta-programming sense as much as just factoring out the branch and a bit of boilerplate ('cause the less control flow the easier things are to read in general). A plain function would do just as well if it was possible to write `"rect: {:?}", rect` in the argument list.

kvark:
@glennw I implemented the approach you proposed in https://github.com/kvark/webrender/commit/dbdb9607bbe58f177ad20b79fd6bb6a5ba7275f8
It appears to be slightly more verbose, but I believe it's better for the long run, since it has support for tracking multiple items at once built-in.

glennw:
@kvark Should we close this and open a PR for https://github.com/kvark/webrender/commit/dbdb9607bbe58f177ad20b79fd6bb6a5ba7275f8 ? Or should we merge this implementation?

kvark:
@glennw I wanted to get feedback from you first. Anyhow, the alternative code is now #2725 for review

kvark:
Given the issue with `PrimitiveInfo` serialization that got in the way of #2725, I suggest us to proceed with this variant, which is now rebased.

kvark:
@gw3583 please take a look

PR #2711, opened 2018-05-01
Store stretch size in primitives, free up extra segment block.

glennw:
The data that was being stored in the segment extra data field is
constant, per segment. Instead, store the stretch size in each
primitive as appropriate, and derive the repeat count in the vertex
shader from the local rect size and stretch size.

This makes a complete 4 floats available in the per-segment data,
which I intend to use as part of porting borders to be brushes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2711)


glennw:
r? @nical

I don't want to cause you rebase issues - if you want to hold off on merging this until your related PRs are merged, that's fine.

It would be good if you could confirm that this change seems OK, and isn't going to interfere with anything you have planned in your pending patches though.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=b989cb4e57f4636869f30c8f80f604af3dc20317

glennw:
It looks like the orange items are existing intermittents.

PR #2716, opened 2018-05-02
Add initial brush border primitive.

glennw:
This is currently only used for image borders. We can incrementally
port more border types to use this brush primitive, and then we
can eventually remove the old border shader and primitive type.

This is also a small optimization for image borders. Previously
image borders created 9 top level primitives via add_image, which
involves extra culling, clip-chain tests etc.

This also removes the only use of sub_rect in the add_image API.
We can probably remove support for that as a follow up. I have
left that in place for now due to other in-progress work that
is modifiying the image primitive APIs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2716)

### DIFF
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -59,11 +59,13 @@ vec2 transform_point_snapped(
 void brush_vs(
     VertexInfo vi,
     int prim_address,
-    RectWithSize local_rect,
+    RectWithSize prim_rect,
+    RectWithSize segment_rect,
     ivec3 user_data,
     mat4 transform,
     PictureTask pic_task,
-    vec4 unused
+    int brush_flags,
+    vec4 texel_rect
 ) {
     ImageBrushData image_data = fetch_image_data(prim_address);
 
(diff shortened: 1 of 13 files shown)


glennw:
r? @nical

Again, I'm happy to wait until your WIP image PRs land before landing this, if you think it will conflict badly.

Note this part from the commit message in particular:

```
This also removes the only use of sub_rect in the add_image API.
We can probably remove support for that as a follow up. I have
left that in place for now due to other in-progress work that
is modifiying the image primitive APIs.
```

I'm not sure if removing that now would simplify or complicate your in-progress PRs?

glennw:
Pending try:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=f60c0beac8a38d77676f9a0fd8b414e81a971af7

webrender/res/brush_image.glsl
```diff
@@ -79,6 +81,26 @@ void brush_vs(
...
+    vec2 stretch_size = image_data.stretch_size;
+
+    // If this segment should interpolate relative to the
+    // segment, modify the parameters for that.
+    if ((brush_flags & BRUSH_FLAG_SEGMENT_RELATIVE) != 0) {
+        local_rect = segment_rect;
+        stretch_size = local_rect.size;
+
+        if ((brush_flags & BRUSH_FLAG_SEGMENT_REPEAT_X) != 0) {
+            stretch_size.x = texel_rect.z - texel_rect.x;
```
nical: I thought that `stretch_size` was in layout space while `texel_rect` in device space, how does this work?

glennw: Since the source image is provided by the user, it doesn't depend on device pixel ratio - it effectively maps one texel to one logical pixel, which (I think) is what we want.

glennw:
Yep, we might be able to tidy this up a bit in the future (to not select these fields in the shader). Thanks for the quick review!

glennw:
Added a comment describing the device pixels vs. local space.

@bors-servo r=nical (via IRC).

PR #2717, opened 2018-05-02
Update to gleam 0.5

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2717)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -701,11 +701,19 @@ impl Device {
         _file_changed_handler: Box<FileWatcherHandler>,
         cached_programs: Option<Rc<ProgramCache>>,
     ) -> Device {
-        let max_texture_size = gl.get_integer_v(gl::MAX_TEXTURE_SIZE) as u32;
+        let mut max_texture_size = [0];
+        unsafe {
+            gl.get_integer_v(gl::MAX_TEXTURE_SIZE, &mut max_texture_size);
+        }
+        let max_texture_size = max_texture_size[0] as u32;
         let renderer_name = gl.get_string(gl::RENDERER);
 
+        let mut extension_count = [0];
+        unsafe {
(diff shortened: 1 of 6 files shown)


webrender/src/device.rs
```diff
@@ -701,11 +702,19 @@ impl Device {
         _file_changed_handler: Box<FileWatcherHandler>,
         cached_programs: Option<Rc<ProgramCache>>,
     ) -> Device {
-        let max_texture_size = gl.get_integer_v(gl::MAX_TEXTURE_SIZE) as u32;
+        let mut max_texture_size = 0;
```
kvark: wouldn't it be easier to just go with `let mut max_texture_size = [0]`, followed by `&mut max_texture_size` and then taking `[0]` from it as opposed to relying on external `ref_slice` crate?

nox: Maybe, which way do you prefer? I don't particularly care either way, I went for the simpler code that made the less invasive changes, not having to do `[0]` everywhere.

nox:
I don't understand what failed, and I don't understand Taskcluster's UI.

staktrace:
Clicking on the link takes you to the task group view, which shows you the different tasks. The one that is marked failed as of right now is "Linux release tests", so you can click on that to view the task detail. Then you need to click on the "Run Logs" tag and select the log (either live.log or live_backing.log will do, they should be the same). The error is:

```
[taskcluster:error] Task was aborted because states could not be created successfully. Error: Error calling 'link' for localLiveLog : Error: (HTTP code 500) server error - connection error: desc = "transport: dial unix:///var/run/docker/containerd/docker-containerd.sock: timeout" 
```

which is some sort of intermittent network failure. We ran into this yesterday as well. Waiting a bit and retrying will probably resolve it

nox:
@staktrace Oh thanks! Will retry in a bit then.

PR #2720, opened 2018-05-02
Return the generated clip id for reference frames.

staktrace:
In gecko we need to know this clip id in order to ensure that items
inside the reference frame are associated with the right scrollframe.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2720)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -1288,7 +1288,7 @@ impl DisplayListBuilder {
         mix_blend_mode: MixBlendMode,
         filters: Vec<FilterOp>,
         glyph_raster_space: GlyphRasterSpace,
-    ) {
+    ) -> Option<ClipId> {
         let reference_frame_id = if transform.is_some() || perspective.is_some() {
             Some(self.generate_clip_id())
         } else {
@@ -1310,6 +1310,8 @@ impl DisplayListBuilder {
 
         self.push_item(item, info);
         self.push_iter(&filters);
+
+        reference_frame_id
     }
 
     pub fn pop_stacking_context(&mut self) {


staktrace:
r? anyone, it's a pretty trivial change

mrobinson:
@staktrace Will you be using this id to assign scroll ids to elements?

staktrace:
@mrobinson Yes. https://bugzilla.mozilla.org/show_bug.cgi?id=1458598#c1 has a bit of a description of the scenario this is needed for.

mrobinson:
@staktrace If you are needing this in Gecko, perhaps it is time to fully separate the ClipScrollTree from the display list. This will make a lot of the internals of WebRender much simpler.

staktrace:
@mrobinson Would that change the exposed API to gecko? I have no objections to the change if it's internal to WebRender. If it's going to shift complexity from WebRender to gecko then I'd like to get a better idea of what would be involved first.

mrobinson:
@staktrace There would be two main pieces of new complexity for clients. The first is dealing with ClipIds created by reference frames. Am I correct in understanding that Gecko will now track all newly created reference frames and use their ClipIds? If so, I think this won't change anything.

The second would be handling coordinate systems. After this change, it might make sense for all coordinates to be given as reference frame relative to the WebRender API. I can try to figure out a way to avoid that change though.

mrobinson:
Correction: Only stacking context coordinates would need to be given relative to reference frames.. WebRender could easily continue to do the per-stacking-context offset.

staktrace:
That sounds reasonable enough. It would be good to wait for bug 1377187 and bug 1458598 to land before working on that though, so that we can test the changes with the new gecko-side clipping glue.

mrobinson:
@staktrace No problem. There is a bit of preparation that is also along the lines of cleanup that I can focus on until it is a good moment to do more experimentation in Gecko.

PR #2728, opened 2018-05-03
Update core-foundation and core-graphics

jrmuizel:
I don't think we want to land this yet, but I'm posting it to avoid anybody duplicating the work.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2728)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -181,12 +181,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


kvark:
Looks fine by me.
> I don't think we want to land this yet

what are the concerns?

jrmuizel:
I want to let some other dependencies update first. That way we don't end up with lots of versions of core-graphics.

jrmuizel:
Also, we could wait for https://github.com/servo/core-foundation-rs/pull/213 which will make this patch use even less unsafe.

gw3583:
@jrmuizel Should we keep this open?

PR #2735, opened 2018-05-06
Don't intersect the dirty rect with the tile rect.

jrmuizel:
Gecko's blob renderer assumes that the dirty rect is in image
space and not in tile space. To avoid having to translate back
to image space in Gecko we can just not clip to the tile in WebRender.

This also fixes the wrench's blob renderer to store the tile size
and clip to the tile rect before rendering.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2735)

### DIFF
--- a/wrench/src/blob.rs
+++ b/wrench/src/blob.rs
@@ -8,6 +8,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use std::sync::Mutex;
 use webrender::api::*;
+use webrender::intersect_for_tile;
 
 // Serialize/deserialize the blob.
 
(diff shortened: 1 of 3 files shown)


wrench/src/blob.rs
```diff
@@ -24,11 +24,31 @@ fn deserialize_blob(blob: &[u8]) -> Result<ColorU, ()> {
     };
 }
 
+fn intersect_for_tile(
```
kvark: looks like a duplicate of the function from `resource_cache`?

wrench/src/blob.rs
```diff
@@ -24,11 +24,31 @@ fn deserialize_blob(blob: &[u8]) -> Result<ColorU, ()> {
...
+            })
+}
+
 // This is the function that applies the deserialized drawing commands and generates
 // actual image data.
 fn render_blob(
     color: ColorU,
     descriptor: &BlobImageDescriptor,
+    tile_size: Option<TileSize>,
     tile: Option<TileOffset>,
```
kvark: let's combine the two as `Option<(TileSize, TileOffset)>`

kvark:
@jrmuizel nit: r=me

PR #2736, opened 2018-05-07
Scroll via HitTester

mrobinson:
This change uses the HitTester to do scrolling, which allows us to
remove some redundant code. It also will help us to remove bounds for
reference frames, which will be useful when we separate the concept of
reference frames and stacking contexts.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2736)


mrobinson:
r? anyone

webrender/src/render_backend.rs
```diff
@@ -620,9 +620,22 @@ impl RenderBackend {
             FrameMsg::Scroll(delta, cursor) => {
                 profile_scope!("Scroll");
 
-                let should_render = doc.scroll(delta, cursor)
-                    && doc.render_on_scroll == Some(true);
+                let mut should_render = true;
+                let node_index = match doc.hit_tester {
+                    Some(ref hit_tester) => {
+                        let test = HitTest::new(None, cursor, HitTestFlags::empty());
```
kvark: since this is where we have the node index and this is where checking for `scroll_nearest_scrolling_ancestor` is really needed, can we rewrite this part as:
```rust
let should_render = match doc.hit_tester {
...
};
```

webrender/src/clip_scroll_tree.rs
```diff
@@ -214,16 +184,33 @@ impl ClipScrollTree {
...
+            None => return self.topmost_scroll_node_index(),
+        };
+
+        let node = &self.nodes[index.0];
+        match node.node_type {
+            NodeType::ScrollFrame(state) if state.sensitive_to_input_events() => return index,
+            _ => {}
+        }
+
+        return self.find_nearest_scrolling_ancestor(node.parent)
```
kvark: nit: let's move this into `_ => ` arm and remove `return` calls

webrender/src/geometry.rs
```diff
@@ -1,113 +0,0 @@
...
-
-/*
- A naive port of "An Efficient and Robust Ray–Box Intersection Algorithm"
- from https://www.cs.utah.edu/~awilliam/box/box.pdf
-
- This should be cleaned up and extracted into more useful types!
- */
-
-// Assumes rect is in the z=0 plane!
-pub fn ray_intersects_rect(
```
kvark: oh god, so nice to see the cleanup here!

webrender/src/hit_test.rs
```diff
@@ -223,6 +223,39 @@ impl HitTester {
...
+
+            let mut clipped_in = false;
+            for item in items.iter().rev() {
+                if !item.rect.contains(&point_in_layer) ||
+                    !item.clip_rect.contains(&point_in_layer) {
+                    continue;
+                }
+
+                let clip_chain_index = clip_and_scroll.clip_chain_index;
+                clipped_in |=
```
kvark: should probably avoid calling this method if `clipped_in` is already true

webrender/src/render_backend.rs
mrobinson: This bit of ugliness is necessary to avoid a double-borrow of doc, unfortunately. I can leave a comment about that.

webrender/src/clip_scroll_tree.rs
mrobinson: Sure thing.

webrender/src/hit_test.rs
mrobinson: Yikes! Thanks for pointing out this bug. This should really be 

```rust
clipped_in = clipped_in || 
    self.is_point_clipped_in_for_clip_chain(point, clip_chain_index, &mut test);
```

so that we can rely on short-circuit logic. If it's okay with you, I'll do this in another PR since the change is unrelated to this one.

mrobinson: A fix for this has landed here: https://github.com/servo/webrender/pull/2739

mrobinson:
Interesting. Looks like a rawtest failure, but I'm not sure how this change could affect it:

```
thread 'main' panicked at 'assertion failed: called.load(Ordering::SeqCst) == 1', wrench/src/rawtest.rs:175:9
```

Gankro:
Could we just disable that particular rawtest instead of all of them?

PR #2739, opened 2018-05-08
Fix short-circuit logic in HitTester

mrobinson:
|= (the bitwise or-assignement operator) does not do short-circuiting.
It isn't an error to call is_point_clipped_in_for_clip_chain again in
this case, but it has the potential to do a lot of extra work.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2739)


mrobinson:
cc @kvark 

r? anyone

Issue #2429, opened 2018-02-15
http://keithclark.co.uk/labs/css-fps/nojs/ runs really badly


jrmuizel:
Rendering is broken and performance is terrible
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1422051

glennw:
There's (at least) two known issues here (that we have existing bugs for).

One is that we don't correctly handle bounding rect and plane-splitting when a plane goes behind the near plane (@kvark is planning  to take a look at fixing this).

The other one is an optimization case - at the moment as soon as we enter a 3d rendering context, we apply plane-splitting and off-screen rendering to every plane encountered. In the vast majority of real cases, the plane-splitting is actually redundant (e.g. if there's no intersections) and we can detect this and skip the off-screen rendering for this context.

PR #2741, opened 2018-05-08
Clip against near plane for bounds calculus

### LINKED ISSUE #2272: regression: Animation rendered all over the place with webrender
Filing for https://bugzilla.mozilla.org/show_bug.cgi?id=1420754

This was determined to be a regression from #2069 but was not fixed by the fixes for #2133.

### LINKED ISSUE #2389: 3d transforms that extend behind the near plane are broken.
For an easy repro case, take the ```perspective-origin.yaml``` reftest, and extend the height of the image such that it extends beyond the near plane.

There are (at least) a couple of issues here.

* The ```calculate_screen_bounding_rect``` function returns incorrect results when some vertices are behind the near plane. We could detect this and clip the polygon to the near plane (this is probably the proper solution). Alternatively, we may be able to take advantage of https://github.com/servo/webrender/pull/2382 - detect this case, set the rect to a maximum value and rely on it being clipped by the outer rect of any parent clips.

* The plane-splitting crate (from a quick look only) seems to assume that the result of any plane split will be a simple polygon with only four vertices. We probably need to clip against the near plane, and produce an arbitrary number of triangles in this case.

There might be other issues, and there might be better solutions than what I've mentioned above. But solving this would fix a number of existing bugzilla issues and test cases.

kvark:
Fixes #2389
Fixes #2272
Improves #2429 a great deal (there are general plane splitting issues in there still)

TODO:
- [x] reviews
- [x] green CI
- [x] try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=ddccbd9f16aea173fe950293f1a4f429650dea6a

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2741)


kvark:
Inviting reviewers! cc @glennw

webrender/src/clip_scroll_node.rs
```diff
@@ -369,13 +369,20 @@ impl ClipScrollNode {
...
+            &self.world_viewport_transform,
+            device_pixel_scale,
+            None,
+        );
 
         // All clipping ClipScrollNodes should have outer rectangles, because they never
         // use the BorderCorner clip type and they always have at last one non-ClipOut
         // Rectangle ClipSource.
-        let screen_outer_rect = screen_outer_rect.expect("Clipping node didn't have outer rect.");
+        let screen_outer_rect = match screen_outer_rect {
```
mrobinson: Do ClipScrollNodes not always have an outer rectangle now?

webrender/src/util.rs
```diff
@@ -160,21 +162,114 @@ pub fn calculate_screen_bounding_rect(
...
     rect: &LayoutRect,
     device_pixel_scale: DevicePixelScale,
-) -> DeviceIntRect {
-    let points = [
-        transform.transform_point2d(&rect.origin),
-        transform.transform_point2d(&rect.top_right()),
-        transform.transform_point2d(&rect.bottom_left()),
-        transform.transform_point2d(&rect.bottom_right()),
+    screen_bounds: Option<&DeviceIntRect>,
+) -> Option<DeviceIntRect> {
```
mrobinson: I recall this function as being extremely performance sensitive, so is there any way to have a fast path when it isn't necessary to do all of these calculations?

kvark:
I'm a bit puzzled over AppVeyor reftest failure...

webrender/src/clip_scroll_node.rs
kvark: In this code, it may be `None` if either `local_outer_rect` of the clip source is `None`, or the computed bounding rect is outside the screen.

webrender/src/util.rs
kvark: The way it's build is to avoid extra steps unless necessary. So the old `transform_point2d` for each corner is equivalent to `transform_point_2d_homogeneous` followed by `to_point2d` at the end.
Extra overhead may come from the following sources:
  1. more stuff being moved around (copies are the doom of Rust code)
  2. comparing 4 f32 values to 0 in `homogens.iter().any(|h| h.w <= 0.0)`
  3. worse instruction cache utilization for the function body being grown

Overall, I think it's a reasonable compromise, and it's not clear to me if we can do better here.

kvark:
@mrobinson thanks for taking a look! I think we are all set, unless there are more concerns.

wrench/src/main.rs
```diff
@@ -15,7 +15,7 @@ extern crate core_graphics;
 extern crate crossbeam;
 #[cfg(target_os = "windows")]
 extern crate dwrote;
-#[cfg(feature = "logging")]
+#[cfg(feature = "env_logger")]
```
mrobinson: I think this was landed in another patch?

kvark:
@mrobinson 
Rebased and fixed the local outer rect now (hopefully).

> One thing we could do here is to detect when transformation doesn't project into three dimensions and to use the old code for this situation. You could have this path be an early return and avoid the long if statement in the body.

It would involve more logic duplication and control flow tweaks, so I'm not sure it's worth it. Maybe we can leave a TODO comment in the code saying to keep an eye for this branch when profiling?

kvark:
Updated try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=7dd1dfe5c8453fe34e37bcfbfd23f54e4bcc9a2e

kvark:
I don't want this to stall any longer.
@bors-servo r=mrobinson

PR #2704, opened 2018-04-27
Decompose tiled images during frame building (v3)

nical:
This PR implements decomposing tiled images during frame building using visibility information to avoid issues with very large blob images.
This version does not use segments, avoiding issues with the maximum contiguous gpu cache alloc size as well as a few hacks that I had to write to get repetitions to work with segments.

As a bonus the non-brush image primitive and ps_image shader are now obsolete so I removed them.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2704)

PR #2742, opened 2018-05-08
Decompose tiled images during frame building (v4)

jrmuizel:
This is a continuation of #2704

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2742)


kvark:
Reviewed 11 of 11 files at r3.

---

*[webrender/src/batch.rs, line 613 at r3](https://reviewable.io/reviews/servo/webrender/2742#-LC4H5r96KZrV-J5BNOk:-LC4H5r96KZrV-J5BNOl:b-93os0) ([raw file](https://github.com/servo/webrender/blob/c2ec519708c63448a77b6a0ab98426aa89af31c9/webrender/src/batch.rs#L613)):*
> ```Rust
>         // use some of the per-primitive data typically stored in PrimitiveMetadata and get
>         // it from each sub-primitive instead.
>         let is_multiple_primitives = match prim_metadata.prim_kind {
> ```

is this a new concept introduced by the PR? We already have segments, so this is a bit confusing

---

*[webrender/src/batch.rs, line 995 at r3](https://reviewable.io/reviews/servo/webrender/2742#-LC4IhxI0xGH7YP1aC1h:-LC4IhxJ-c208qUx3v1Y:b-jmlies) ([raw file](https://github.com/servo/webrender/blob/c2ec519708c63448a77b6a0ab98426aa89af31c9/webrender/src/batch.rs#L995)):*
> ```Rust
>                         }
>                     }
>                     BrushKind::Image { request, ref visible_tiles, .. } if !visible_tiles.is_empty() => {
> ```

does it make sense to take the `_` arm if no visible tiles are there? It seems that we are better just going the `for` loop unconditionally here

---

*[webrender/src/batch.rs, line 1365 at r3](https://reviewable.io/reviews/servo/webrender/2742#-LC4KjSk3EXkKTXQ3J1o:-LC4KjSk3EXkKTXQ3J1p:b8ejclq) ([raw file](https://github.com/servo/webrender/blob/c2ec519708c63448a77b6a0ab98426aa89af31c9/webrender/src/batch.rs#L1365)):*
> ```Rust
>     );
> 
>     if cache_item.texture_id == SourceTexture::Invalid {
> ```

I'd really like to nuke that `Invalid` variant, perhaps later... but it's unfortunate to see more use of it

---

*[webrender/src/image.rs, line 183 at r3](https://reviewable.io/reviews/servo/webrender/2742#-LC4N_0KDcUzUZBR5Lao:-LC4N_0KDcUzUZBR5Lap:bctulwn) ([raw file](https://github.com/servo/webrender/blob/c2ec519708c63448a77b6a0ab98426aa89af31c9/webrender/src/image.rs#L183)):*
> ```Rust
> 
>     // Position of the first visible tile (top-left) in layer space.
>     let x0 = prim_rect.origin.x + t0.x as f32 * layer_tile_size.width;
> ```

nit: I don't think we need those, the `x_count` computation is not getting more complex if we just do it all in there

---

*[webrender/src/prim_store.rs, line 2480 at r3](https://reviewable.io/reviews/servo/webrender/2742#-LC4OPlrEE8xZyr-RMnW:-LC4OPlrEE8xZyr-RMnX:b604ty1) ([raw file](https://github.com/servo/webrender/blob/c2ec519708c63448a77b6a0ab98426aa89af31c9/webrender/src/prim_store.rs#L2480)):*
> ```Rust
> }
> 
> fn compute_conservative_visible_rect(
> ```

as mentioned in https://github.com/servo/webrender/pull/2704#issuecomment-387506329 we shouldn't have this method

---

*[webrender/src/renderer.rs, line 139 at r3](https://reviewable.io/reviews/servo/webrender/2742#-LC4O_nbBUjgX55PwHtO:-LC4O_nbBUjgX55PwHtP:bjwa7hh) ([raw file](https://github.com/servo/webrender/blob/1eb46da9a8ff5fc092834b08a1578cb636cac560/webrender/src/renderer.rs#L139)):*
> ```Rust
>     color: debug_colors::LIGHTGREY,
> };
> const GPU_TAG_PRIM_IMAGE: GpuProfileTag = GpuProfileTag {
> ```

:tada: 

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2742)*

jrmuizel:
This has one of the review comments applied. The others seemed less actionable, or things we can do in a follow up once nical is back.

nical:
> is this a new concept introduced by the PR? We already have segments, so this is a bit confusing

Yep, version 2 of this tile decomposition saga was an attempt to use segments and it went wrong in various ways:
 - We ran into situations where we need more than a couple hundred tiles which blows up the max amount of segments (which are always contiguously allocated in the gpu cache).
 - I had to add a new way way to select the source rect per segment which added a fair bit of code to the shaders.
 - I had to change the way segments selected their batches in that particular case (because tiles of a given primitive wouldn't necessarily end up sampling the same texture).
 - At the end none of the segment infrastructure ended up being useful to the tiling case and the implementation ended up twice as complicated as this one even though they pretty much do the same thing.

That said, with some extra work we could actually segment individual tiles just like we do for regular primitives to move more pixels to the opaque pass when there is aa or clipping involved.

> does it make sense to take the _ arm if no visible tiles are there? It seems that we are better just going the for loop unconditionally here

This arm is specific to the tiled case. we could rewrite it to handle both tiled and non-tiled images but the non-tiled image path is shares more with the rest of the primitives than it shares with the tiled image path so it wouldn't buy us much.

webrender/src/image.rs
```diff
@@ -224,3 +224,77 @@ pub fn for_each_tile(
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::collections::HashSet;
+    use api::{LayoutRect, DeviceUintSize};
+    use euclid::{TypedRect, TypedSize2D, TypedPoint2D};
+
+    fn point<T: Copy, U>(x: T, y: T) -> TypedPoint2D<T, U> {
+        TypedPoint2D::new(x, y)
```
kvark: let's just use existing euclid's `point2`, `size2`, and possibly `rect` helpers

webrender/src/image.rs
```diff
@@ -224,3 +224,77 @@ pub fn for_each_tile(
...
+    ) {
+        let mut coverage = LayoutRect::zero();
+        let mut tiles = HashSet::new();
+        for_each_tile(prim_rect,
+                      visible_rect,
+                      device_image_size,
+                      device_tile_size,
+                      &mut |tile_rect, tile_offset, tile_flags| {
+                          // make sure we don't get sent duplicate tiles
+                          assert!(!tiles.contains(&tile_offset));
```
kvark: can assert that `prim_rect.contains(&tile_rect)` as well as
`!tiles.any(|t| t.intersects(&tile_rect))` which is superior to `!tiles.contains(&tile_offset)`

webrender/src/image.rs
```diff
@@ -224,3 +224,77 @@ pub fn for_each_tile(
...
+                      },
+        );
+        assert!(prim_rect.contains_rect(&coverage));
+        assert!(coverage.contains_rect(&visible_rect.intersection(&prim_rect).unwrap_or(LayoutRect::zero())));
+    }
+
+    #[test]
+    fn basic() {
+        let mut count = 0;
+        checked_for_each_tile(&rect(0., 0., 1000., 1000.),
```
kvark: we can just make it to count things right away instead of accepting a generic closure

webrender/src/image.rs
```diff
@@ -224,3 +224,77 @@ pub fn for_each_tile(
...
+                              &rect(75., 75., 400., 400.),
+                              &size(400, 400),
+                              36,
+                              &mut |_tile_rect, _tile_offset, _tile_flags| {
+                                  count += 1;
+                              },
+        );
+        assert_eq!(count, 0);
+    }
+}
```
kvark: newline!

kvark:
:ship: it
@bors-servo r+

PR #2743, opened 2018-05-09
Eliminate ClipScrollNode::local_viewport_rect

mrobinson:
It never really made sense for all types of ClipScrollNodes to have
bounding rectangles. For instance, for quite some time ReferenceFrames
always had bounding rectangles that started at zero and with unused
dimensions. With some recent refactoring we can eliminate this property
and have only those node types that need a bounding rectangle keep it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2743)


mrobinson:
r? anyone

PR #2745, opened 2018-05-10
Don't select already used texture

dati91:
The same restriction is present in the [next check](https://github.com/servo/webrender/blob/master/webrender/src/renderer.rs#L3425) when we did't found the index.

The incorrect reuse happens in e.g. `reftests\filters\filter-large-blur-radius.yaml` test.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2745)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -3414,7 +3414,7 @@ impl Renderer {
                     size: texture.get_dimensions(),
                     num_layers: texture.get_render_target_layer_count(),
                     format: texture.get_format(),
-                }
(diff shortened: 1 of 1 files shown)


gw3583:
I thought it might be something like that. @dati91 Could you expand on what problem this fixes and what the issue is with the reftest you mentioned above?

dati91:
@kvark thanks for clarifying
@gw3583 then sorry for the noise, because it only affects our vulkan implementation

kvark:
Alright, we'll figure it out downstream then ;)

PR #2747, opened 2018-05-10
Remove StackingContext::scroll_policy

mrobinson:
This can be handled in clients now that they have access to reference
frame ciip ids. Removing fixed positioning support from WebRender
simplifies the code greatly and also opens the way to explicit creation
of reference frames in the API. It will also allow us to get rid of the
hacky replacements code during display list flattening.

This also requires exposing the root reference frame to Wrench tests in
order to enable them retain the fixed positioning feature. This change
exposes new aliases to wrench "root-scroll-node" and
"root-reference-frame" in order to make this clearer. These nodes
correspond to the ids 1 and 0 respectively and Wrench will halt if it
encounters these used for other nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2747)

### DIFF
--- a/webrender/examples/alpha_perf.rs
+++ b/webrender/examples/alpha_perf.rs
@@ -34,7 +34,6 @@ impl Example for App {
         builder.push_stacking_context(
             &info,
             None,
-            ScrollPolicy::Scrollable,
             None,
             TransformStyle::Flat,
             None,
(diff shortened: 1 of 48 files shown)


mrobinson:
Since Gecko only uses ScrollPolicy:Scrollable, the only change necessary there should be to remove this argument from `DisplayListBuilder::push_stacking_context`. There should be no behavior change.

gw3583:
Reviewed 48 of 48 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2747)*

PR #2750, opened 2018-05-11
Use non-animated (static) value if the animation is in delay phase

hiikezoe:
When animation is in-delay phase, there is no corresponding value in SceneProperties, so we have to store the static value and use it for the delay phase.
   
This patch lets PropertyBinding::Binding having the static value along with the key.

This is the change for https://bugzilla.mozilla.org/show_bug.cgi?id=1460389 .

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2750)

### DIFF
--- a/webrender/examples/animation.rs
+++ b/webrender/examples/animation.rs
@@ -43,14 +43,14 @@ impl Example for App {
         let bounds = (0, 0).to(200, 200);
 
         let filters = vec![
-            FilterOp::Opacity(PropertyBinding::Binding(self.opacity_key), self.opacity),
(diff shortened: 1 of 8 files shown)


hiikezoe:
Can I use bors-servo here?  I will try.
@bors-servo try

@staktrace r?

staktrace:
The changes looks reasonable enough to me, so r+ from my point of view once the CI status is green. One of the recognized owners of the repo will have to tell bors though.

hiikezoe:
Gah, one more place. 

Thanks kats for the info!

PR #2752, opened 2018-05-11
Decompose gradients during scene building.

nical:
This completes the tiling/decomposition saga by decomposing gradients with spacing during frame building. I hope that gradients with spacing are rare enough that we don't need an optimized path that bakes the spacing and repeats with a single primitive. We'll see.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2752)


gw3583:
Do we need a gecko try for this?

nical:
This is pretty straightforward since it directly uses the infrastructure from the previous episodes of the tiling saga, so I wouldn't worry about it more than our usual patch (as opposed to the rest of the image tiling work).
We might get unexpected passes now that we can render more than a couple hundred gradient repetitions at once.

gw3583:
Reviewed 3 of 3 files at r1.

---

*[webrender/src/batch.rs, line 1414 at r1](https://reviewable.io/reviews/servo/webrender/2752#-LCFOvCN0ledwC31B76i:-LCFOvCN0ledwC31B76j:bbkbh4c) ([raw file](https://github.com/servo/webrender/blob/13ee36a8fe50be0d8ed0dab54924d416e2c5c3e6/webrender/src/batch.rs#L1414)):*
> ```Rust
> 
>     for tile in visible_tiles {
>         let stops_handle = &ctx.cached_gradients[gradient_index.0].handle;
> ```

We could probably take this and a base_instance outside the loop, not a big deal though.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2752)*

gw3583:
Looks good to me. I'll leave it to you whether to fix the nit above and/or do a gecko try, then this should be good to go.

@bors-servo delegate+

nical:
Nit fixed. @bors-servo r=glennw

PR #2757, opened 2018-05-14
Properly update the viewport rect on scroll nodes

mrobinson:
I introduced a regression in PR #2743 with the handling of updated
scroll frames. Instead of the new viewport size, the old one is used.
This lead to misbehavior of sticky elements in Gecko. Unfortunately, I
don't know a good way to test this with an automated test, but I've
confirmed that this fixes the issue manually.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2757)


Darkspirit:
Mentioned Nightly regression: https://bugzilla.mozilla.org/show_bug.cgi?id=1461154

mrobinson:
@kvark Yeah. I totally agree. I'll circle back on this to try to come up with a solution that is less fragile.

mrobinson:
@kvark Thanks for the review!

PR #2761, opened 2018-05-16
Allocate more bits for the clip chain rectangle and node IDs

kvark:
We easily hit the max number of local rectangles, and the batch code happily converts those indices to negative... The PR is not a general solution, but gets us closer to a situation where we might as well crash out of memory than exceed our internal indices.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1439386 ~~(TODO: confirm)~~
Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=74e50c08e063f049d1d11000278dd1719224f6c0

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2761)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -24,6 +24,14 @@ void brush_vs(
 #define BRUSH_FLAG_SEGMENT_REPEAT_X             4
 #define BRUSH_FLAG_SEGMENT_REPEAT_Y             8
 
+//Note: these have to match `gpu_types` constants
+#define INT_BITS    (31)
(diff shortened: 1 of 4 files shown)


kvark:
The fix is confirmed. Try looks green(-ish). Ready to fly now!
FYI, the max number of local rectangles has now 7 more bits.

gw3583:
I wonder if we'd be better to just use an ```i32``` for these fields and avoid all the shifts. But we can always consider that as something for later on, so r=me once you're ready to merge this.

PR #2762, opened 2018-05-16
Allow explicit indication of which missing pipelines are acceptable

staktrace:
In some cases Gecko includes iframes for pipelines that are in other
content processes, and the order in which the display lists reach the
compositor is not deterministic. So it would be better if we could just
have WebRender ignore those pipeline references. In all other cases we
can assert.

This is desirable for https://bugzilla.mozilla.org/show_bug.cgi?id=1454042

r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2762)

### DIFF
--- a/webrender/examples/iframe.rs
+++ b/webrender/examples/iframe.rs
@@ -74,7 +74,7 @@ impl Example for App {
         );
         // red rect under the iframe: if this is visible, things have gone wrong
         builder.push_rect(&info, ColorF::new(1.0, 0.0, 0.0, 1.0));
-        builder.push_iframe(&info, sub_pipeline_id);
+        builder.push_iframe(&info, sub_pipeline_id, false);
         builder.pop_stacking_context();
     }
 }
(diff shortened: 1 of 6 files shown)


kvark:
Did you consider instead to have Gecko sending a provisional pipeline in case it needs to update it later? I don't quite like the fact that WR will just skip some elements, whatever they are, since it would make visual surprises for us, non-deterministically. If Gecko doesn't know when it will get the contents, it can at least figure out what the temporary contents should be for the mean time.

staktrace:
I'm not sure how Gecko could send a provisional pipeline, it would require the UI process to block on the content process in order to know whether it needs to send one, and that is not allowed because it can lead to chrome hangs which are bad. Also the temporary contents would just be an empty display list so in practice the results should be the same.

staktrace:
Patch updated to fix style nit

kvark:
> Also the temporary contents would just be an empty display list so in practice the results should be the same.

My point here is that it's Gecko who wants to mess things up by not providing the pipeline upfront for scene building, so it's the entity that should take responsibility for missing content, not the special handling in WR.

kvark:
Let's get this in, and continue discussing later.
@bors-servo r+

Darkspirit:
#TechnicalDebt

PR #2763, opened 2018-05-16
Preliminary cleanup for the blob image stuff

nical:
This PR stores the blob image data as an `Arc<Vec<u8>>` internally instead of a `Vec<u8>`. This is needed to do the pre-post scene building tracking of the blob data without needlessly cloning everything all the time.
Since I was going to do a slight breaking change I took the occasion to fix something that's been bothering me forever: storing image sizes as `DeviceUintSize` instead of pairs of `u32`s.

@staktrace heads up it's a slight API change, I'll put up a patch to fixup the bindings tomorrow.

r? anyone, all of the changes are mechanical.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2763)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{AddFont, BlobImageData, BlobImageResources, ResourceUpdate, ResourceUpdates};
+use api::{AddFont, BlobImageResources, ResourceUpdate, ResourceUpdates};
 use api::{BlobImageDescriptor, BlobImageError, BlobImageRenderer, BlobImageRequest};
 use api::{ClearCache, ColorF, DevicePoint, DeviceUintPoint, DeviceUintRect, DeviceUintSize};
 use api::{Epoch, FontInstanceKey, FontKey, FontTemplate};
@@ -34,7 +34,6 @@ use std::collections::hash_map::Entry::{self, Occupied, Vacant};
 use std::cmp;
 use std::fmt::Debug;
 use std::hash::Hash;
-use std::mem;
 #[cfg(any(feature = "capture", feature = "replay"))]
 use std::path::PathBuf;
 use std::sync::{Arc, RwLock};
@@ -495,7 +494,7 @@ impl ResourceCache {
         &mut self,
         image_key: ImageKey,
         descriptor: ImageDescriptor,
-        mut data: ImageData,
+        data: ImageData,
         mut tiling: Option<TileSize>,
     ) {
         if tiling.is_none() && Self::should_tile(self.max_texture_size(), &descriptor, &data) {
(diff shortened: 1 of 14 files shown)


webrender/src/resource_cache.rs
```diff
@@ -149,24 +149,23 @@ pub struct ResourceClassCache<K: Hash + Eq, V, U: Default> {
 
 pub fn intersect_for_tile(
     dirty: DeviceUintRect,
-    width: u32,
-    height: u32,
+    actual_tile_size: DeviceUintSize,
```
kvark: perhaps, `clipped_tile_size`? We can even re-use the `ScreenRect` struct here

webrender/src/resource_cache.rs
```diff
@@ -149,24 +149,23 @@ pub struct ResourceClassCache<K: Hash + Eq, V, U: Default> {
...
-            DeviceUintSize::new(width, height),
-        )).map(|mut r| {
-                // we can't translate by a negative size so do it manually
-                r.origin.x -= tile_offset.x as u32 * tile_size as u32;
-                r.origin.y -= tile_offset.y as u32 * tile_size as u32;
-                r
-            })
+    dirty.intersection(&DeviceUintRect::new(
+        DeviceUintPoint::new(
+            tile_offset.x as u32 * tile_size as u32,
```
kvark: any plans to vectorize this? :)

nical: I am not not sure what you mean here. `tile_offset` is a u16 "in number of tiles" vector, which we multiply by a scalar. Are your referring to something like creating a `ScaleFactor<u32, Tiles, DeviceSpace>` from `tile_size as u32` and multiply `tile_offset`?

webrender/src/resource_cache.rs
nical: `clipped_tile_size` sounds good to me. Which ScreenRect are you referring to?

kvark: The only one we have - in `prim_store`:
```rust
#[derive(Debug, Copy, Clone)]
pub struct ScreenRect {
    pub clipped: DeviceIntRect,
    pub unclipped: DeviceIntRect,
}
```

nical: Oh, I doubt it'd make things any simpler in this case.

webrender/src/resource_cache.rs
kvark: Don't worry about this, but I was thinking about something like:
```rust
dirty.intersection(&DeviceUintRect::new(tile_offset.to_u32() * tile_size as u32, actual_tile_size)
    .map(|r| r.translate_neg(tile_offset.to_u32() * tile_size as u32))
            
```

webrender/src/resource_cache.rs
kvark: yeah, I'm fine with just renaming the thing to `clipped_tile_size`

webrender/src/resource_cache.rs
nical: It'd also need tile_offset to be turned into a vector in device space instead of a point in tiles for the thing to chain properly.

webrender/src/resource_cache.rs
nical: I'll do it tomorrow, the build is also whining about a long line.

nical:
The gecko side of the API changes are in the patch in [bug 1461342](https://bugzilla.mozilla.org/show_bug.cgi?id=1461342).

PR #2764, opened 2018-05-17
Add ProgramCacheObserver for an application can handle ProgramCache update

sotaroikeda:
The pull request has some changes that  application can pre-load program binary to ProgramCache and can handle ProgramCache update for saving program binary.

Related info is in [Bug 1418202](https://bugzilla.mozilla.org/show_bug.cgi?id=1418202)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2764)

### DIFF
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -15,6 +15,7 @@ capture = ["webrender_api/serialize", "ron", "serde", "debug_renderer"]
 replay = ["webrender_api/deserialize", "ron", "serde"]
 debug_renderer = []
 pathfinder = ["pathfinder_font_renderer", "pathfinder_gfx_utils", "pathfinder_partitioner", "pathfinder_path_utils"]
+serialize_program = ["serde"]
 
 [dependencies]
 app_units = "0.6"
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -23,6 +23,7 @@ use std::path::PathBuf;
 use std::ptr;
 use std::rc::Rc;
 use std::slice;
+use std::sync::Arc;
 use std::thread;
 
 #[derive(Debug, Copy, Clone, PartialEq, Ord, Eq, PartialOrd)]
@@ -575,7 +576,8 @@ pub struct VBOId(gl::GLuint);
 #[derive(PartialEq, Eq, Hash, Debug, Copy, Clone)]
 struct IBOId(gl::GLuint);
 
-#[derive(PartialEq, Eq, Hash, Debug)]
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+#[cfg_attr(feature = "serialize_program", derive(Deserialize, Serialize))]
 pub struct ProgramSources {
     renderer_name: String,
     vs_source: String,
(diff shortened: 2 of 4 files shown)


sotaroikeda:
r? anyone.

webrender/src/device.rs
```diff
@@ -592,32 +594,50 @@ impl ProgramSources {
     }
 }
 
+#[cfg_attr(feature = "serialize_program", derive(Deserialize, Serialize))]
 pub struct ProgramBinary {
+    sources: ProgramSources,
```
kvark: can we only have this field for when `serialize_program` is featured?

webrender/src/device.rs
```diff
@@ -592,32 +594,50 @@ impl ProgramSources {
...
+            sources: sources.clone(),
             binary,
             format
         }
     }
 }
 
+/// The interfaces that an application can implement to handle ProgramCache update
+pub trait ProgramCacheObserver {
+    fn notify_binary_added(&self, program_binary: &Arc<ProgramBinary>);
```
kvark: does the application need to have `Arc<>` here? would it make sense to provide `&ProgramBinary` instead?

webrender/src/lib.rs
```diff
@@ -48,7 +48,7 @@ extern crate cfg_if;
 extern crate lazy_static;
 #[macro_use]
 extern crate log;
-#[cfg(any(feature = "debugger", feature = "capture", feature = "replay"))]
+#[cfg(any(feature = "debugger", feature = "capture", feature = "replay", feature = "serialize_program"))]
```
kvark: I wonder if we can just do `feature = "serde"` here

webrender/src/device.rs
```diff
@@ -592,32 +594,50 @@ impl ProgramSources {
...
-    pub fn new() -> Rc<Self> {
+    pub fn new(program_cache_observer: Option<Box<ProgramCacheObserver>>) -> Rc<Self> {
         Rc::new(
             ProgramCache {
                 binaries: RefCell::new(FastHashMap::default()),
+                program_cache_handler: program_cache_observer,
             }
         )
     }
+    pub fn load_program_binary(&self, program_binary: Arc<ProgramBinary>) {
```
kvark: where is this used or meant to be used?

webrender/Cargo.toml
```diff
@@ -15,6 +15,7 @@ capture = ["webrender_api/serialize", "ron", "serde", "debug_renderer"]
 replay = ["webrender_api/deserialize", "ron", "serde"]
 debug_renderer = []
 pathfinder = ["pathfinder_font_renderer", "pathfinder_gfx_utils", "pathfinder_partitioner", "pathfinder_path_utils"]
+serialize_program = ["serde"]
```
kvark: need to add testing for it to CI

webrender/src/device.rs
sotaroikeda: Yes, I am going to address it.

webrender/src/device.rs
sotaroikeda: Arc<> is necessary here, since ProgramBinary is serialized and is stored to disk on worker thread.

webrender/src/lib.rs
sotaroikeda: I will update it in a next pull request.

webrender/Cargo.toml
sotaroikeda: I will add the testing.

sotaroikeda:
I applied the comment, but Arc<ProgramBinary> was kept since since ProgramBinary is serialized and is stored to disk on worker thread. [diagram](https://bugzilla.mozilla.org/attachment.cgi?id=8976401) shows gecko side implementation.

@kvark  can you review it again?

kvark:
thank you!
@bors-servo r+

PR #2766, opened 2018-05-18
Ensure epoch updates and callbacks fire for zero-pixel frames

staktrace:
Even if the window size is zero in some dimensions (which happens
legitimately in Gecko), we don't want to bail out completely at the
scene building step. We need to pretend to do the work, and send the
necessary epoch updates and callbacks back to Gecko. If we fail to do
this Gecko can end up in a stuck state where it is waiting for a render
that it thinks is going to happen but never does. This patch optimizes
out the rendering step in these cases, rather than doing an early-exit
on the entire codepath.

This is needed for https://bugzilla.mozilla.org/show_bug.cgi?id=1451305

r? @glennw - let me know if there's a better way to do this, or more things we can optimize away. I didn't spend too much time on it, there might be further improvements possible.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2766)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -42,6 +42,7 @@ use std::sync::atomic::{ATOMIC_USIZE_INIT, AtomicUsize, Ordering};
 use std::mem::replace;
 use std::sync::mpsc::{channel, Sender, Receiver};
 use std::u32;
+#[cfg(feature = "replay")]
 use tiling::Frame;
 use time::precise_time_ns;
 
@@ -170,13 +171,15 @@ impl Document {
 
     fn can_render(&self) -> bool { self.frame_builder.is_some() }
 
+    fn has_pixels(&self) -> bool {
+        !self.view.window_size.is_empty_or_negative()
+    }
+
     // TODO: We will probably get rid of this soon and always forward to the scene building thread.
     fn build_scene(&mut self, resource_cache: &mut ResourceCache) {
         let max_texture_size = resource_cache.max_texture_size();
 
-        if self.view.window_size.width == 0 ||
-           self.view.window_size.height == 0 ||
-           self.view.window_size.width > max_texture_size ||
+        if self.view.window_size.width > max_texture_size ||
            self.view.window_size.height > max_texture_size {
             error!("ERROR: Invalid window dimensions {}x{}. Please call api.set_window_size()",
                 self.view.window_size.width,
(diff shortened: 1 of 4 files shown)


webrender/src/render_backend.rs
```diff
@@ -170,13 +170,15 @@ impl Document {
 
     fn can_render(&self) -> bool { self.frame_builder.is_some() }
 
+    fn has_pixels(&self) -> bool {
+        self.view.window_size.width > 0 || self.view.window_size.height > 0
```
kvark: nit: could use `!self.view.window_size.is_empty_or_negative()` instead

webrender/src/render_backend.rs
```diff
@@ -307,6 +305,14 @@ impl Document {
         )
     }
 
+    pub fn get_updated_pipeline_info(&mut self) -> PipelineInfo {
```
kvark: let's not have the name starting with `get_` since it's mutating `self`.

webrender/src/render_backend.rs
```diff
@@ -1079,6 +1085,13 @@ impl RenderBackend {
             self.result_tx.send(msg).unwrap();
             profile_counters.reset();
             doc.render_on_hittest = false;
+        } else if op.render {
+            // WR-internal optimization to avoid doing a bunch of render work if
+            // there's no pixels. We still want to pretend to render and request
+            // a composite to make sure that the callbacks (particularly the
+            // new_frame_ready callback below) has the right flags.
+            let msg = ResultMsg::PublishPipelineInfo(doc.get_updated_pipeline_info());
```
kvark: It's duplicating the way Renderer gets the pipeline updates. Perhaps, we can remove this from `PublishDocument` one and always make it so the `PublishPipelineInfo` is sent beforehand?

kvark:
Oh, and also - would be really nice to have a reftest checking those code paths.

wrench/src/rawtest.rs
```diff
@@ -47,6 +48,8 @@ impl<'a> RawtestHarness<'a> {
         self.test_very_large_blob();
         self.test_save_restore();
         self.test_capture();
+        self.test_zero_height_window();
+        self.test_zero_height_window();
```
kvark: why is it repeated many times?

staktrace: Whoops, sorry, that was from local testing to make sure it wasn't leaving stuff in a bad state. Updated patch so that it only runs once.

PR #2767, opened 2018-05-23
Add initial port of borders to brush primitives, with task caching.

gw3583:
This adds the infrastructure for drawing borders as brushes. The
border primitive is drawn via a render task, and cached between
both scroll frames and also new display lists, since the cache key
is based on the content of the border.

The border is drawn as a nine-patch, in order to minimize the
amount of space that is allocated in the texture cache for the
primitive. The edge segments use the repetition support in the
brush shader to draw the complete edge. This makes normal CSS
borders behave almost exactly the same in the rendering phase
as image-borders, which improves batching (since they are drawn
with a simple image shader).

This initial patch supports only a small subset of the possible
border primitives - once this lands I'll be incrementally adding
the other border styles to this system, and eventually removing
the legacy border primitive.

I had to add some fuzziness to a couple of the reftests, due to
the test running through the new path, vs. the reference running
through the old border path, which has slight differences in the
AA between edges. Once all the border types are ported to the new
path, we can remove this fuzziness.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2767)


gw3583:
This still needs a gecko try, and probably a bit of code cleanup. But it should mostly be ready for review.

r? @kvark

kvark:
I assume this change can also show up as a performance regression, for the fact those simple borders now go through the cached targets mechanism before being shown to screen. E.g. if all the borders are animated, and the cache gets invalidated all the time, we'll get slower rendering.

---

---

*[webrender/res/cs_border_segment.glsl, line 66 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDY-QbBxP2PuhDQdif:-LDDY-Qc9C-LgXgKVpv5:b-lu34ba) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L66)):*
> ```GLSL
> }
> 
> vec2 get_clip_sign(int segment) {
> ```

It does appear to me that this method and get_outer_corner are redundant. Would probably be easier to have one method returning just a vec2 of (0,0), (0,1), (1,0), (1,1) depending on the corner, and then easily derive both the outer corner and the clip sing from it

---

*[webrender/res/cs_border_segment.glsl, line 126 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDYzAAEat8oEg7QXSe:-LDDYzAAEat8oEg7QXSf:b-dvg4pm) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L126)):*
> ```GLSL
>     }
> 
>     gl_Position = uTransform * vec4(aTaskOrigin + pos, 0.0, 1.0);
> ```

can we just include the `aTaskOrigin` value in `aRect`?

---

*[webrender/res/cs_border_segment.glsl, line 136 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDZHYR7Eu1z6kRXm9e:-LDDZHYS-aQnJjfXZyhB:b-qdkmd6) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L136)):*
> ```GLSL
> 
>     // Apply clip-lines
>     float d_line = distance_to_line(vClipLine.xy, vClipLine.zw, vPos);
> ```

can we move this under the following `if`?

---

*[webrender/res/cs_border_segment.glsl, line 149 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDZSTg7hIfKSHkC5Pi:-LDDZSTg7hIfKSHkC5Pj:b1s8ek7) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L149)):*
> ```GLSL
>     if ((vClipFlags & CLIP_RADII) != 0) {
>         if (vClipSign.x * p.x < 0.0 && vClipSign.y * p.y < 0.0) {
>             d = max(d, d_radii);
> ```

similarly, can we move all the `d_radii` computation under this `if`?

---

*[webrender/src/border.rs, line 59 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDD_LOw9jMCONcPLQhb:-LDD_LOw9jMCONcPLQhc:b-y9mpe6) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L59)):*
> ```Rust
>     fn from(radius: BorderRadius) -> BorderRadiusAu {
>         BorderRadiusAu {
>             top_left: LayoutSizeAu::new(
> ```

this is begging  the `LayoutSize -> LayoutSizeAu` helper

---

*[webrender/src/border.rs, line 89 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDD_b33A9WNAq6rYdnG:-LDD_b377qlEXRUs2v5u:b-do20hn) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L89)):*
> ```Rust
> }
> 
> impl From<BorderWidths> for BorderWidthsAu {
> ```

one way to reduce the code duplication would be to have those structs like `BorderWidths` to be generic over T, which we'll use with `f32` and `Au`. We'd still need to write the conversions though.

---

*[webrender/src/border.rs, line 103 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDa7l9D0ybzVs0ZH5E:-LDDa7l9D0ybzVs0ZH5F:bkxp74j) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L103)):*
> ```Rust
> #[cfg_attr(feature = "capture", derive(Serialize))]
> #[cfg_attr(feature = "replay", derive(Deserialize))]
> pub struct BorderSideAu {
> ```

we could move the width in here, so that `BorderWidthsAu` is not needed

---

*[webrender/src/border.rs, line 473 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDaJdl6j5neKB5E76o:-LDDaJdl6j5neKB5E76p:b4p69kg) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L473)):*
> ```Rust
>                 BorderStyle::Hidden |
>                 BorderStyle::None => {
>                     true
> ```

is this the case even for `edge.color != constant_color`?

---

*[webrender/src/border.rs, line 1045 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDaxUEDOyos28jVrOw:-LDDaxUF9RDqb3pwhlyM:b-8wvpm7) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L1045)):*
> ```Rust
>         let dp_height_br = (border.radius.bottom_right.height * scale).ceil();
> 
>         let dp_width_tl = (border.radius.top_left.width * scale).ceil();
> ```

can we vectorize this a little by doing `let dp_tl = (border.radius.top_left * scale).ceil()`?

---

*[webrender/src/border.rs, line 1084 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDc0v51CMEAnnZ2nFQ:-LDDc0v6-3YxLj9yUIt4:b-3km79j) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L1084)):*
> ```Rust
>         );
> 
>         let left_color = border.left.border_color(1.0, 2.0 / 3.0, 0.3, 0.7);
> ```

should we define those 0.3 and 0.7 magic constants somewhere?

---

*[webrender/src/border.rs, line 1252 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDdKTQAW6KeLY8RZn8:-LDDdKTQAW6KeLY8RZn9:b-rxeqww) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L1252)):*
> ```Rust
>         );
> 
>         add_corner_segment(
> ```

This is somewhat difficult to validate with eyes. Could we put some unit tests in place? At least, we can verify that the produced segments/instances don't overlap and cover the whole border area.

---

*[webrender/src/prim_store.rs, line 1432 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDeLoV23ihPhr5aB6K:-LDDeLoW7ebKWBHYxrXE:bw3hr9s) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/prim_store.rs#L1432)):*
> ```Rust
>         let brush = &mut self.cpu_brushes[metadata.cpu_prim_index.0];
> 
>         if let BrushKind::Border { ref mut source, .. } = brush.kind {
> ```

nit: we could avoid indentation of the whole method (by returning out of a match arm) and/or combine two if-lets into one

---

*[webrender/src/prim_store.rs, line 1775 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDfB5-9le0vrCF7JfS:-LDDfB50CJBAUHGrp5Ge:b-c5iog8) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/prim_store.rs#L1775)):*
> ```Rust
>                             BorderSource::Border { .. } => {
>                                 // Handled earlier
> ```

any reason we can't move that "earlier" here? :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2767#-:-LDDfep2Eg3A4TDtavxx:byuge9p)*

gw3583:
Review status: 16 of 22 files reviewed at latest revision, 13 unresolved discussions, some commit checks pending.

---

*[webrender/res/cs_border_segment.glsl, line 66 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDY-QbBxP2PuhDQdif:-LDEHpQs9yq6sE1DP9Oi:b-896fix) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L66)):*

It does appear to me that this method and get_outer_corner are redundant. Would probably be easier to have one method returning just a vec2 of (0,0), (0,1), (1,0), (1,1) depending on the corner, and then easily derive both the outer corner and the clip sing from it
</blockquote></details>

Done.

---

*[webrender/res/cs_border_segment.glsl, line 126 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDYzAAEat8oEg7QXSe:-LDEHqnNB6Xn-ccQhQjy:bybzvx3) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L126)):*

can we just include the `aTaskOrigin` value in `aRect`?
</blockquote></details>

Not in this case, since we need to be able to operate on the local-space position for some of the other calculations (such as clip distance).

---

*[webrender/res/cs_border_segment.glsl, line 136 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDZHYR7Eu1z6kRXm9e:-LDEIa5P4650heJ9bXuw:b-896fix) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L136)):*

can we move this under the following `if`?
</blockquote></details>

Done.

---

*[webrender/res/cs_border_segment.glsl, line 149 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDZSTg7hIfKSHkC5Pi:-LDEIaZ56tNWnbC5v3wO:b-896fix) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/res/cs_border_segment.glsl#L149)):*

similarly, can we move all the `d_radii` computation under this `if`?
</blockquote></details>

Done.

---

*[webrender/src/border.rs, line 59 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDD_LOw9jMCONcPLQhb:-LDEKjFX3VbkiLYlhhby:b-896fix) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L59)):*

this is begging  the `LayoutSize -> LayoutSizeAu` helper
</blockquote></details>

Done.

---

*[webrender/src/border.rs, line 89 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDD_b33A9WNAq6rYdnG:-LDEIfF518xF25g1KmeL:bdrqntj) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L89)):*

one way to reduce the code duplication would be to have those structs like `BorderWidths` to be generic over T, which we'll use with `f32` and `Au`. We'd still need to write the conversions though.
</blockquote></details>

I did consider this, but avoided it for now since I'm not sure what effect (if any) it might have on the codegen that serde produces for (de)serializing these types.

---

*[webrender/src/border.rs, line 103 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDa7l9D0ybzVs0ZH5E:-LDEInre7mVqX8w88GUm:bjuidof) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L103)):*

we could move the width in here, so that `BorderWidthsAu` is not needed
</blockquote></details>

I tried to mirror the public struct here, which seems a bit clearer. But I can combined these if you'd prefer?

---

*[webrender/src/border.rs, line 473 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDaJdl6j5neKB5E76o:-LDELKHuCl_m5eDVEXer:b5bj2d3) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L473)):*

is this the case even for `edge.color != constant_color`?
</blockquote></details>

Yep, the initial border brush implement does now handle the edge colors not being equal. I updated the variable name to be clearer what this is referring to. 

---

*[webrender/src/border.rs, line 1045 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDaxUEDOyos28jVrOw:-LDEIufa-jICaL69tJGR:b-896fix) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L1045)):*

can we vectorize this a little by doing `let dp_tl = (border.radius.top_left * scale).ceil()`?
</blockquote></details>

Done.

---

*[webrender/src/border.rs, line 1084 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDc0v51CMEAnnZ2nFQ:-LDEIvcXChZ7_KZ6V1Fb:b49y6i3) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L1084)):*

should we define those 0.3 and 0.7 magic constants somewhere?
</blockquote></details>

I added a comment describing where they come from.

---

*[webrender/src/border.rs, line 1252 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDdKTQAW6KeLY8RZn8:-LDEJ8rT5Lk77Qb81KVA:b-8ku93p) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/border.rs#L1252)):*

This is somewhat difficult to validate with eyes. Could we put some unit tests in place? At least, we can verify that the produced segments/instances don't overlap and cover the whole border area.
</blockquote></details>

We have quite good coverage of this in the border reftest suite, which contains a lot of different variations of widths / radii. But I can add some unit tests in addition to these if you prefer.

---

*[webrender/src/prim_store.rs, line 1432 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDeLoV23ihPhr5aB6K:-LDEL5aH5ocI3FIbEa_J:ba10rg) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/prim_store.rs#L1432)):*

nit: we could avoid indentation of the whole method (by returning out of a match arm) and/or combine two if-lets into one
</blockquote></details>

This seems like it would be less readable to me - could you show me what you would expect it to look like?

---

*[webrender/src/prim_store.rs, line 1775 at r1](https://reviewable.io/reviews/servo/webrender/2767#-LDDfB5-9le0vrCF7JfS:-LDEJLkT-3HAhFerHtdt:b-q78xdk) ([raw file](https://github.com/servo/webrender/blob/80ad66afe57bb342165a79da45e518467ebf0ff4/webrender/src/prim_store.rs#L1775)):*

any reason we can't move that "earlier" here? :)
</blockquote></details>

We need to determine the segment descriptor before update_clip_task is called. I added a comment to clarify this.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2767)*

gw3583:
It can show up as a performance regression, you're right. I think in practice this would only occur in contrived test cases, for the following reasons:

* If a single frame contains multiple animated borders with the same cache key, that is a win since we only draw that border once and then use the simple image shader to composite all of them.
* We draw the borders as nine-patches, so the cache key will often match so long as the radii match, even if the overall size of the primitive is different, which improves cache hits (this also means the number of pixels spent in the expensive border shader is much smaller than previously).
* We typically get better batching during scene composite, since there are often other primitives interleaved with borders.

You could definitely come up with a test case where this is a performance regression, but I expect it to be a significant performance win in all real world cases :)

I *think* all the review comments are addressed. However, it looks like there are a couple of failures in the gecko try - I still need to address those.

gw3583:
OK, added a new commit to fix the gecko try failures. Current try is:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=a024c4057a7a2f9e85060667b8114c7f9dd16455&selectedJob=179941115

Despite the large amount of orange, I *think* this is ready for merge once @kvark is happy with the review comments, but @staktrace should sign off on the orange try results:

```1, tp6, Wd, wpt2, wpt5, 15, gl1, gpu, Wd, wpt11, x, R4``` - all appear to be unrelated intermittents. Do we know what is causing these?

```R8, R8, R1, R1, R1, R1``` - all of these are the same 4 failures, but these are fuzzed on all other rendering backends. Now WR matches them in fuzziness :)

staktrace:
@glennw Try push looks good. Random orange factor is pretty bad right now even on m-c. It's not specific to WebRender.

kvark:
Alright, I think all my bullets have been deflected, and it's time to ship this :)
@bors-servo r+

---

Reviewed 2 of 2 files at r2, 2 of 4 files at r3, 8 of 8 files at r4.

---

*[webrender/src/border.rs, line 42 at r4](https://reviewable.io/reviews/servo/webrender/2767#-LDHdNvg46NLdedWdyPQ:-LDHdNvg46NLdedWdyPR:bgfv7m7) ([raw file](https://github.com/servo/webrender/blob/9e1d61d22de1263214fc7968c3b5380db9aef588/webrender/src/border.rs#L42)):*
> ```Rust
> 
> trait AuSizeConverter {
>     fn to_au(&self) -> LayoutSizeAu;
> ```

oh, what I had in mind is a simple `impl From<LayoutSize> for LayoutSizeAu {..}`, but that trait works too

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2767#-:-LDHe65m3qU0kRz9Kkzc:bzavxdd)*

PR #2718, opened 2018-05-02
Merge ResourceUpdate and Transaction

hgallagher1993:
Fixes #2322

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2718)

PR #2768, opened 2018-05-23
Add a hook for resource-update-only SB transaction completion.

staktrace:
The scene builder might process a transaction which has only resource
updates and no actual display list. In this case we still need to
trigger a composite after the transaction is processed, but we don't
need to the more heavyweight stuff that goes with a scene swap. This
patch adds a new hook that enables the caller to trigger the composite.

This is needed to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1463313

r? @nical or anybody else

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2768)


staktrace:
There was a merge error with #2718 - I've rebased the patch now.

PR #2770, opened 2018-05-23
Fix document placement and scissoring

kvark:
This PR fixes our document example.
r? @gw3583 

Edit: actually, still working on getting https://github.com/servo/servo/issues/20847 fixed by this

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2770)

### DIFF
--- a/webrender/src/render_task.rs
+++ b/webrender/src/render_task.rs
@@ -701,7 +701,13 @@ impl RenderTask {
             }
         };
 
-        let (target_rect, target_index) = self.get_target_rect();
+        let (mut target_rect, target_index) = self.get_target_rect();
+        // The primitives inside a fixed-location render task
+        // are already placed to their corresponding positions,
+        // so the shader doesn't need to shift by the origin.
+        if let RenderTaskLocation::Fixed(_) = self.location {
+            target_rect.origin = DeviceIntPoint::origin();
+        };
 
         RenderTaskData {
             data: [
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -2887,8 +2887,18 @@ impl Renderer {
 
             for alpha_batch_container in &target.alpha_batch_containers {
                 if let Some(target_rect) = alpha_batch_container.target_rect {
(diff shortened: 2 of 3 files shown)


kvark:
@gw3583 can be reviewed now

webrender/src/renderer.rs
```diff
@@ -2843,8 +2843,18 @@ impl Renderer {
 
             for alpha_batch_container in &target.alpha_batch_containers {
                 if let Some(target_rect) = alpha_batch_container.target_rect {
+                    // Note: `framebuffer_target_rect` needs a Y-flip before going to GL
+                    let rect = if render_target.is_none() {
```
gw3583: Probably worth putting this code in a function instead of duplicating below?

kvark:
Apparently, I still need to work out the proper device/pixel ration support here.

webrender/src/renderer.rs
kvark: I think this should be done to serve both these scissors and ones for clearing. It's not even going to save LOC, just give that piece of logic a name.

webrender/src/render_task.rs
```diff
@@ -701,7 +701,13 @@ impl RenderTask {
             }
         };
 
-        let (target_rect, target_index) = self.get_target_rect();
+        let (mut target_rect, target_index) = self.get_target_rect();
+        // The primitives inside a fixed-location render task
+        // are already placed to their corresponding positions,
+        // so the shader doesn't need to shift by the origin.
+        if let RenderTaskLocation::Fixed(_) = self.location {
+            target_rect.origin = DeviceIntPoint::origin();
```
gw3583: It seems like it might be tidier to ensure the target rect origin is initialized to zero for fixed tasks, instead of modified here? Not a bit deal either way.

kvark: Unfortunately, we can't do this. There are other places where `Fixed(rect)` is expected to have the actual origin, e.g. `get_target_rect`.

kvark:
@gw3583 thanks!
@bors-servo r=gw3583

PR #2775, opened 2018-05-24
This adds an assertion to a rawtest to catch an intermittent

jrmuizel:
It seems like this test sometimes fails. This should move the crash
earlier.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2775)


mrobinson:
If I'm not mistaken the bot failure is the assertion that you added doing its job.

PR #2776, opened 2018-05-24
pipe through a scene_id to distinguish cached pictures

### LINKED ISSUE #2769: async scene building: cache collisions from reused picture ids
On my linux machine with async-scene-building, if I open this in one tab:

```
data:text/html,<div style="width: 400px;text-shadow: 1px 1px 2px green;">Hello there my name is alexis and this is my amazing test-case for shadows being broken under async scene building
```

and then open this in a second tab (identical except the shadow is red):

```
data:text/html,<div style="width: 400px;text-shadow: 1px 1px 2px red;">Hello there my name is alexis and this is my amazing test-case for shadows being broken under async scene building
```

the second tab has a green shadow, as it is using the cached shadow picture (really the final blur picture) from the other tab.

The issue is that async scene building repeatedly creates new `FrameBuilder::empty()`'s, resetting the picture id to 0 *and* that the picture cache doesn't factor in scene/document id's.

Gankro:
fixes #2769

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2776)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -176,7 +176,7 @@ impl Document {
     }
 
     // TODO: We will probably get rid of this soon and always forward to the scene building thread.
-    fn build_scene(&mut self, resource_cache: &mut ResourceCache) {
+    fn build_scene(&mut self, resource_cache: &mut ResourceCache, scene_id: u64) {
         let max_texture_size = resource_cache.max_texture_size();
 
         if self.view.window_size.width > max_texture_size ||
@@ -199,7 +199,7 @@ impl Document {
             return;
         }
 
-        // The DisplayListFlattener will re-create the up-to-date current scene's pipeline epoch
+        // The DisplayListFlattener  re-create the up-to-date current scene's pipeline epoch
         // map and clip scroll tree from the information in the pending scene.
         self.current.scene.pipeline_epochs.clear();
         let old_scrolling_states = self.clip_scroll_tree.drain();
@@ -213,6 +213,7 @@ impl Document {
             &self.output_pipelines,
             &self.frame_builder_config,
             &mut self.current.scene,
+            scene_id,
         );
 
         self.clip_scroll_tree.finalize_and_apply_pending_scroll_offsets(old_scrolling_states);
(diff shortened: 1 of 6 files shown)


Gankro:
CI failure looks like an intermittent?

gw3583:
Yep, probably related to #2775.

webrender/src/render_backend.rs
```diff
@@ -213,6 +213,7 @@ impl Document {
             &self.output_pipelines,
             &self.frame_builder_config,
             &mut self.current.scene,
+            0,
```
kvark: why are we passing 0 here?

webrender/src/render_backend.rs
```diff
@@ -424,6 +427,7 @@ pub struct RenderBackend {
     recorder: Option<Box<ApiRecordingReceiver>>,
     sampler: Option<Box<AsyncPropertySampler + Send>>,
 
+    last_scene_id: u64,
```
kvark: nit: any specific reason for not using `usize`?

Gankro: because all our other ids are 64-bit unconditionally (less platform-specificness)

webrender/src/render_backend.rs
Gankro: because the value isn't necessary on the sync path, and also we plan to remove this path altogether soon, right?

kvark: please put a comment then saying that this is not used because the picture IDs get reset. Frankly, it seems easier to just put the correct value in here in the first place

Gankro: there is a comment elsewhere, where the purpose of this value is explained, but you're right, i'll look at piping it through properly. i did this when i was just figuring out how everything piped together

kvark:
Thanks, all good from my side to ship.

kvark:
@Gankro please rebase and r=me

gw3583:
@Gankro I think this needs a rebase in order to pick up some unrelated rawtest fixes, then r=kvark.

Gankro:
(sorry been sick this week, will get to this when i'm feeling better)

PR #2777, opened 2018-05-24
Update serde to 1.0.58 + serde-rs/serde#1278

SimonSapin:


Gankro:
I thought we weren't supposed to update serde due to some knock on dependency effects? Was that resolved?

Eijebong:
@Gankro The problem was depending to the latest serde_derive which depends on syn 0.14. (Not that it really matters now that we finally removed syn 0.11 (:tada:) from the dependency tree, but having three versions was a bit much)

SimonSapin:
https://github.com/serde-rs/serde/pull/1278 and servo/deserialize_from_enums7 are based on 1.0.58 rather than .59 or upstream/master to avoid picking up syn 0.14 yet.

PR #2781, opened 2018-05-26
Use a proper allow attribute.

emilio:
`#![allow(..)]` is supposed to be global, I wouldn't have expected the code to compile, but in any case I suspect it may be allowing unused variables in the whole module.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2781)

### DIFF
--- a/webrender/src/device.rs
+++ b/webrender/src/device.rs
@@ -603,7 +603,7 @@ pub struct ProgramBinary {
 }
 
 impl ProgramBinary {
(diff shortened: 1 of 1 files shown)


staktrace:
According to https://doc.rust-lang.org/reference/attributes.html it will allow unused variables in the enclosing scope, so presumably in any function in the `impl`. But yeah the change looks good. r=me (I don't have bors powers so you'll have to tell it)

PR #2785, opened 2018-05-28
Allow rasterizing blobs on the scene builder thread

nical:
The general idea is to move the rasterization of blob images that we think will be rendered soon to the scene builder thread so that it can happen asynchronously and not block scrolling.
The current heuristic for what "we think will be rendered soon" is to rasterize anything that is in the bounds of the display list. This is simple but will typically rasterize more than necessary when off-screen blobs are animated. We can look into tweaking this as a followup.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2785)


kvark:
I looked at it briefly, comments below

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[webrender/examples/blob.rs, line 224 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEForCt3cORBPnN01_t:-LEForCt3cORBPnN01_u:brwndvs) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/examples/blob.rs#L224)):*
> ```Rust
>         &mut self,
>         _services: &api::BlobImageResources,
>         requests: Vec<(api::BlobImageRequest, api::BlobImageDescriptor, Option<api::DeviceUintRect>)>,
> ```

given that the same trio appears in `BlobImageRenderer::request` signature , they deserve a struct home (unless you want to remove that other use eventually)

---

*[webrender/examples/blob.rs, line 226 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEFog5_0Knl8sJ24ie8:-LEFog5_0Knl8sJ24ie9:b-qsrzok) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/examples/blob.rs#L226)):*
> ```Rust
>         requests: Vec<(api::BlobImageRequest, api::BlobImageDescriptor, Option<api::DeviceUintRect>)>,
>     ) -> Option<Box<api::BlobSceneBuilderRequest>> {
>         Some(Box::new(SceneBuilderRequest {
> ```

would using an associated type be more appropriate here instead of a trait object?

---

*[webrender/examples/blob.rs, line 246 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEFpnyrDeM3we_cQsC0:-LEFpnyt8eE7f49cPYvC:b-hnudl3) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/examples/blob.rs#L246)):*
> ```Rust
>         let requests = mem::replace(&mut self.requests, Vec::new());
>         let workers = Arc::clone(&self.workers);
>         workers.install(||{
> ```

can't we do `self.workers.install()` here?

---

*[webrender/src/render_backend.rs, line 236 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEFq0FE-vIKhHJ2dGts:-LEFq0FFAID89c7jK30h:b-if2z6v) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/src/render_backend.rs#L236)):*
> ```Rust
>         document_ops: &DocumentOps,
>         document_id: DocumentId,
>         resource_cache: &mut ResourceCache,
> ```

this is a little concerning to me. Semantically, this is no longer just forwarding the transaction but actually starting the work on it (by rasterizing blobs)

---

*[webrender/src/resource_cache.rs, line 429 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEG6Wp42HZITj8C1pj6:-LEG6Wp5BO9rjjI_DSOv:b-uw19wv) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/src/resource_cache.rs#L429)):*
> ```Rust
>     ) {
>         let mut new_updates = Vec::with_capacity(updates.len());
>         for update in mem::replace(updates, Vec::new()) {
> ```

did you try going with `retain()` instead of doing the replace dance here?

---

*[webrender/src/resource_cache.rs, line 433 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEG6hlI6JWX7d_rg6AZ:-LEG6hlI6JWX7d_rg6A_:b-vbhhyj) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/src/resource_cache.rs#L433)):*
> ```Rust
>                 ResourceUpdate::AddImage(ref img) => {
>                     if let ImageData::Blob(ref blob_data) = img.data {
>                         let tiling = None; // TODO: tiled blob images.
> ```

`assert_eq!(img.tiling, None)`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2785#-:-LEG89p37hDsm7mnqfz0:by7fvle)*

nical:
Review status: 0 of 7 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[webrender/examples/blob.rs, line 224 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEForCt3cORBPnN01_t:-LEOxXHY3Ae5hgrI_pCf:b-yc5bab) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/examples/blob.rs#L224)):*

given that the same trio appears in `BlobImageRenderer::request` signature , they deserve a struct home (unless you want to remove that other use eventually)
</blockquote></details>

I am not sure what the actual interface will be by the time this merges (do we pass a rect decide which tiles to render or do we pass tiles explicitly, etc), so this is the simple version, but It'll be a struct by the time this PR is finished.

---

*[webrender/examples/blob.rs, line 226 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEFog5_0Knl8sJ24ie8:-LEOxxe34Nxgl13TFpt2:bsgtrh1) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/examples/blob.rs#L226)):*

would using an associated type be more appropriate here instead of a trait object?
</blockquote></details>

I don't use associated types very often but my understanding is that the actual type needs to be known at compile time which wouldn't work with BlobImageRenderer and its request object that are selected at runtime.

---

*[webrender/examples/blob.rs, line 246 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEFpnyrDeM3we_cQsC0:-LEP-3NG5djyUsR_yLkR:bs5qe5x) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/examples/blob.rs#L246)):*

can't we do `self.workers.install()` here?
</blockquote></details>

Yeah that looks like a remnant of a previous version of the code where this was needed.

---

*[webrender/src/render_backend.rs, line 236 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEFq0FE-vIKhHJ2dGts:-LEOtOfyCCzmiDq5D8L-:b3cdm0g) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/src/render_backend.rs#L236)):*

this is a little concerning to me. Semantically, this is no longer just forwarding the transaction but actually starting the work on it (by rasterizing blobs)
</blockquote></details>

We don't necessarily start any work here. We just take the blob image renderer as &mut in order to let it do a bit of internal bookkeeping when creating the request object. We can hide that behind a `Cell` if you feel strongly about keeping the interface non-&mut.

---

*[webrender/src/resource_cache.rs, line 429 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEG6Wp42HZITj8C1pj6:-LEOxGOn3yXpuTiSYQsw:bpzxshn) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/src/resource_cache.rs#L429)):*

did you try going with `retain()` instead of doing the replace dance here?
</blockquote></details>

I'll  give it a try, there's also the option of maintaining a separate array.

---

*[webrender/src/resource_cache.rs, line 433 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEG6hlI6JWX7d_rg6AZ:-LEOwbSF6Czb1YVKvrf3:bxazmz4) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/src/resource_cache.rs#L433)):*

`assert_eq!(img.tiling, None)`?
</blockquote></details>

There's two kinds of tiling: whether the blob image is rendered into separate tiles and whether we upload it into separate tiles in the texture cache. When we support rendering blobs in tiles I don't see a reason to keep the two separate but in the mean time we need to support splitting a big non-tiled blob image into texture cache tiles.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2785)*

nical:
Review status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @kvark and @nical)

---

*[webrender/src/resource_cache.rs, line 429 at r1](https://reviewable.io/reviews/servo/webrender/2785#-LEG6Wp42HZITj8C1pj6:-LFSI78pByJd43Rv1WYi:bio7l0l) ([raw file](https://github.com/servo/webrender/blob/122d73920175912d86864b7344f56e8109ee48f9/webrender/src/resource_cache.rs#L429)):*

So I looked into this again and the reason I did the replace dance was the classic "gotta call a &mut method while iterating on member array" story.
</blockquote></details>

Sorry, actually it was because the font variants need to move non-copy stuff out of the vector.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2785)*

nical:
I implemented rasterizing blob tiles and the API to specify the visible rect of an image, which restricts blob rasterization to the tiles that intersect it.
That's enough to pass the `very_large_blob` rawtest, but there are still some missing pieces, the main one being that the rasterized tiles are currently never discarded and getting that right requires a bit of thinking.

On the gecko side I think I figured out how to compute the intersection of the blob image and the viewport, but I haven't integrated that in a way that I can test with the webrender parts yet.

nical:
The PR is not ready to land because some stuff on the gecko side still block testing it thoroughly, but it's ready for reviews.

## How this works

The blob image mechanism now has two traits:
- [`BlobImageHandler`](https://github.com/servo/webrender/pull/2785/files#diff-2b72a28a40b83edf41a59adfd46b1a11R188) is roughly the equivalent of the previous `BlobImageRenderer` except that it doesn't do any rendering (it manages the state of the blob commands, and resources like fonts).
- [`AsyncBlobImageRasterizer`](https://github.com/servo/webrender/pull/2785/files#diff-2b72a28a40b83edf41a59adfd46b1a11R211) is created by the handler and sent over to the scene builder thread. the async rasterizer is meant to be a snapshot of the state of blob image commands that can execute the commands if provided some requests.

When receiving a transaction, the render backend / resource cache look at the list of added and updated blob images in that transaction, [collect the list of blob images and tiles that need to be rendered](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R848), create a rasterizer, and ship the two to the scene builder.
After building the scene the rasterizer gets handed the list of blob requests and [does all of the rasterization](https://github.com/servo/webrender/pull/2785/files#diff-856af4d4ff2333d4204e7e5a87a93c58R153), blocking the scene builder thread until the work is done.

When the scene building and rasterization is done, the render backend receives the rasterized blobs and [stores them](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R520) so that they are available when frame building needs them.

Because blob images can be huge, we don't always want to rasterize them entirely during scene building. To decide what should be rasterized, we rely on gecko giving us a hint through the added `set_image_visible_area` API. When the render backend receives that message [it decides which tiles are going to be rasterized](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R469). This information is also used to [decide which tiles to evict](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R430), so that we don't keep thousands of tiles if we scroll through a massive blob image. The idea is for the visible area to correspond to the size of the display list.

Sometimes, however, Gecko gets this visible area "wrong", or at least gives webrender a certain visible area but eventually webrender requests tiles during frame building that weren't in that area. I think that this is inevitable because the culling logic in gecko and webrender works very differently, so relying on them to match exactly is fragile at best.
So to work around this type of situation, [keep around the async blob rasterizer](https://github.com/servo/webrender/pull/2785/files#diff-3722af8f0bcba9c3ce197a9aa3052014R769) that we sent to the scene builder, and store it in the resource cache when we swap the scene. This blob rasterizer represents the state of the blob commands at the time the transaction was built (and is potentially different from the state of the blob image handler). Frame building [collects a list of blob images](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R811) (or blob tiles) that are not already rasterized, and asks the current async blob rasterizer to rasterize them synchronously on the render backend. The hope is that this would happen rarely.

Another important detail is that for this to work, resources that are used by blob images (so currently only fonts), need to be in sync with the blobs. Fortunately, fonts are currently immutable so we mostly need to make sure they are added {before the transaction](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R440) is built and [removed after](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R400) the transaction is swapped. If blob images were to use images, then we'd have to either do the same for these images (and disallow updating them), or maintain the state of images before and after scene building like we effectively do for blob.

## Shortcomings

I'll  admit this whole mechanism isn't particularly exciting or beautiful. It's the simplest I could come up with that satisfies the transaction rules and need to b, etc. In its current state it has a few ugly corners:

- Because we eagerly rasterize blobs for the entire scene during scene building, we rasterize a lot more than we do before with lazy rasterization during frame building.
- Currently restricting blob rasterization to the dirty rect relies on guarantees that we get from rasterizing synchronously on the scene builder, namely: we rasterize only if we need to upload so we know it will be uploaded, and we know at the time of rasterization whether the thing is still on the GPU. Moving this to the scene builder we don't know whether the blob will be in the texture cache by the time we finish rasterizing and therefore we have to produce the whole image or whole tile.

Both of these things can be mitigated by incorporating the same kind of optimizations we do in gecko's layers tiling logic (copy-on-write tiles, checkerboarding, clever heuristics to decide which tiles are really important to rasterize depending on scroll speed, etc). But it's not clear to me how much effort should go into these optimizations, versus into using blob images less (certainly a mix of both). More importantly it'll be miraculous if this lands as is without making waves, so let's not make it even more complicated, prepare for some regressions and land this anyway, since this makes APZ a ton smoother even if at the expense extra CPU time spent during rasterization.

Gankro:
> Because we eagerly rasterize blobs for the entire scene during scene building, we rasterize a lot more than we do before with lazy rasterization during frame building.

Checking i understand what this means: ideally we would only rasterize blobs that are actually on screen, but with this design we need to rasterize all the tiles which *might* be APZ scrolled onto the screen with the current scene?

nical:
> Checking i understand what this means: ideally we would only rasterize blobs that are actually on screen, but with this design we need to rasterize all the tiles which might be APZ scrolled onto the screen with the current scene?

Yes. We can make extensions to this implementation where the render backend decides what critical area needs to be rasterized eagerly and asynchronously schedule rasterization of blobs we think we will see soon because of scroll speed, etc. That would be similar to gecko's layer tiling logic.
That's possible but I hope we won't need to do it.

nical:
Review ping @kvark @Gankro @gw3583 @staktrace (and anyone else interested).

gw3583:
Added some minor comments in the review above, nothing major.

The overview description above sounds reasonable, although I haven't thought it through completely. It's perhaps worth including that description somewhere in the ```doc/``` directory in tree?

Overall, I'm happy to get this merged once others are - like you said, it's probably best to get it in and working, and deal with any fallout from it.

There is a CI failure - I haven't investigated that, and I expect we'll also want a try run, plus signoff from @kvark before merging.

Nice work!

Darkspirit:
webrender/doc/blob.md:
`added {before the transaction]` -> `added [before the transaction]`

webrender/doc/blob.md
```diff
@@ -0,0 +1,17 @@
...
+
+Another important detail is that for this to work, resources that are used by blob images (so currently only fonts), need to be in sync with the blobs. Fortunately, fonts are currently immutable so we mostly need to make sure they are added {before the transaction](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R440) is built and [removed after](https://github.com/servo/webrender/pull/2785/files#diff-77cbdf7ba9ebae81feb38a64c21b8454R400) the transaction is swapped. If blob images were to use images, then we'd have to either do the same for these images (and disallow updating them), or maintain the state of images before and after scene building like we effectively do for blob.
```
Darkspirit: `added {before the transaction]` -> `added [before the transaction]`

nical:
> So this appears to be related to display port of Gecko, in which case - why don't we communicate it as such (i.e. a general set_display_port or something) instead of having a state per image?

To do this we'd need to resolve the scroll tree during frame building to figure out where each image item is with respect to the display port (which long term would be my favorite solution, but requires quite a bit more work).

> Why would we want to rely on it? In my understanding, Gecko's viewport should be considered an optimization hint by WR, (e.g. this is how much stuff you should expect to be shown) not much more.

The way I read your question, I think that we are agreeing: relying on gecko's culling and webrender's to match perfectly would be hard/fragile, hence the current approach being to treat the visible area information provided by gecko more as an optimization hint than a drawing parameter like a clip.

> I don't understand this part.

currently (non-blob) image updates are applied to the resource cache after we swap the scene. This is because it is information relevant to frame building, so if we were to add image updates to the resource cache when the transaction arrives before scene building, we would expose APZ frames which don't have the new scene built to the updated versions of the images that should arrive later with the scene.
As a result, if blob images were to read images from the resource cache they would not see any potential image update from the same transaction. This isn't a problem with fonts because we don't have a way to update them. It's also not in practice a problem with images because blob images don't actually read from them right now but it's something to know about if we want to do that in the future.

> Aren't we only rasterizing the current viewport blobs (as opposed to "entire scene")?

So there is a bit of confusion around viewport/displayport and I totally get it wrong half of the time. In this PR the rectangle that we use to decide what we rasterize is one that corresponds to the size of the entire scene, which in practice represents a few screenfuls of content (it's not the entire page).

> Also, it would be nice if the Renderer didn't wait for rasterization of tiles that it knows are outside of the current frame. Do you see a way for us to make it so?

This comes back to the problem  of being able to know where the image items are in screen space and requires resolving the scroll tree during scene building and doing a subset of the culling phase there. It'd be really cool because we could be much smarter about what to rasterize ahead of time (could take into account scroll direction and velocity, etc), but that requires some work.

> Didn't we previously decide to force those images to be in the texture cache for the duration of async blob rasterization?

It was one of the possibilities we talked about but I believe we ended up leaning towards rasterizing more instead of uploading more. The reason is that doing extra rasterization on the scene builder is costly for the CPU but won't cause janky scrolling, while uploading too much puts us at risk of missing the frame budget in the critical path.
So uploading textures is still lazy, even if rasterization is eager.

kvark:
Thanks for addressing my concerns @nical ! I was just about to let bors do it's thing but noticed the lack of a try push. Guess this is blocked now until we see it passing.

nical:
> interesting. Could you explain why it has to go into euclid?

TileRange is a TypedRect which is in euclid even if one of type parameter is defined in webrender (I tried to be sure).

There is a green try push from a week ago but there was a few rather bumpy rebases in between so I do want do another one before landing (I tried on Friday but got stuck with non-trivial wr update)

nical:
Latest try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=a167b8873fba6c7b9a6802b88b6046d6580686c3 (so far the failures are unexpected passes)

kvark:
Looks good, thank you!
@bors-servo r+

staktrace:
Is bors stuck? The taskcluster jobs on the merge commit passed but it didn't finish the merge...

gw3583:
@staktrace it looks like bors was stuck. I issued a retry - if that doesn't work, we can manually merge, or if @nical pushes an amended commit hash, that is generally enough to fix up the state that bors is in.

kvark:
>spurious network error (2 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)

@bors-servo retry

nical:
Rebased and squashed. @bors-servo r=kvark

PR #2725, opened 2018-05-03
Debug tracking of display items

kvark:
This is a different take on #2710

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2725)

PR #2786, opened 2018-05-28
Revert "Debug tracking of display items"

kvark:
This reverts #2725, which causes chaos and destruction in Gecko tests: https://treeherder.mozilla.org/#/jobs?repo=try&revision=12a0fba7254666c7dfba9f4b738093e076de2381&selectedJob=180600645

I'll make sure to test the Gecko binding side before bringing this back.

r? @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2786)

### DIFF
--- a/webrender_api/src/display_item.rs
+++ b/webrender_api/src/display_item.rs
@@ -67,9 +67,6 @@ pub struct PrimitiveInfo<T> {
     pub clip_rect: TypedRect<f32, T>,
     pub is_backface_visible: bool,
     pub tag: Option<ItemTag>,
-    #[cfg(debug_assertions)]
-    #[serde(skip_serializing, default)]
-    pub is_tracked_for_debug: bool,
 }
 
 impl LayoutPrimitiveInfo {
@@ -86,20 +83,7 @@ impl LayoutPrimitiveInfo {
             clip_rect,
             is_backface_visible: true,
             tag: None,
-            #[cfg(debug_assertions)]
-            is_tracked_for_debug: false,
         }
-    }
-
-    #[cfg(debug_assertions)]
-    #[inline]
-    pub fn is_tracked(&self) -> bool {
-        self.is_tracked_for_debug
-    }
-    #[cfg(not(debug_assertions))]
-    #[inline]
-    pub fn is_tracked(&self) -> bool {
-        false
     }
 }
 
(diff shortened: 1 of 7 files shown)


staktrace:
r=me

webrender_api/src/display_item.rs
```diff
@@ -67,9 +67,6 @@ pub struct PrimitiveInfo<T> {
     pub clip_rect: TypedRect<f32, T>,
     pub is_backface_visible: bool,
     pub tag: Option<ItemTag>,
-    #[cfg(debug_assertions)]
-    #[serde(skip_serializing, default)]
```
Gankro: just for the record, the issue is this line, which shouldn't be included (or should be `(skip_serializing, skip_deserializing)`).

kvark:
@Gankro for the record, I believe this is an issue between Bincode and Serde, not our/user code. They need to agree on how optional fields are handled and implement this *consistently*: either bail out on the encoder creation because bincode doesn't like optional fields, or ignore the optionality in both encoder and decoder. There is no value in handling this separately, it's more of a trap.

kvark:
uh:
> [taskcluster:error] Task timeout after 3600 seconds. Force killing container.

@bors-servo retry

Gankro:
@kvark i agree that this is really bad UX, but it's just how serde works. It's a bit messy because it's trying to support viewing structs as sequences and maps, and they want it to be possible for an application to do things that only really make sense for one or the other without introducing a messy type-level distinction (if it's even possible).

staktrace:
> [taskcluster:error] Task timeout after 3600 seconds. Force killing container.

Not sure where you're seeing this. All the three failures have this in the log:

```
	hit testing test...
	retained blob images test...
thread 'main' panicked at 'assertion failed: called.load(Ordering::SeqCst) == 1', wrench/src/rawtest.rs:391:9
stack backtrace:
```

i.e. the retained blob images rawtest is failing consistently.

kvark:
@staktrace interesting. This appears unrelated to me.

gw3583:
There is some kind of intermittent failure in the raw tests, that I haven't looked into. If it's a perma-fail on this revert, we should just disable those raw tests until someone can look into the root cause.

kvark:
unused happens...
@bors-servo r=staktrace,gw3583

kvark:
what's going on, our rawtests are going nuts :fire:

kvark:
Looks like we'll have to block this until #2788 comes through

PR #2789, opened 2018-05-29
Give Linux CI more time to run

staktrace:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2789)


staktrace:
Note that Linux CI reinstalls rust every time, so that does make it slower than Mac CI, which reuses the rust install from one run to the next.

PR #2796, opened 2018-06-01
Don't try to resolve blobs that weren't requested

staktrace:
r? @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2796)


staktrace:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1455490 which is a WR topcrasher in firefox

PR #2797, opened 2018-06-01
Rebuild the hit-tester as part of the scene build.

staktrace:
In order for Gecko/APZ to get useful results from hit-testing, the
hit-test tree needs to stay in sync with the scene and clip-scroll tree.
This patch accomplishes this by rebuilding the hit-tester at the
same time we rebuild or swap in these other structures, rather than at
render time.

r? @mrobinson

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2797)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -111,11 +111,6 @@ struct Document {
     // the first frame would produce inconsistent rendering results, because
     // scroll events are not necessarily received in deterministic order.
     render_on_scroll: Option<bool>,
-    // A helper flag to prevent any hit-tests from happening between calls
-    // to build_scene and rendering the document. In between these two calls,
-    // hit-tests produce inconsistent results because the clip_scroll_tree
-    // is out of sync with the display list.
-    render_on_hittest: bool,
 
     /// A data structure to allow hit testing against rendered frames. This is updated
     /// every time we produce a fully rendered frame.
@@ -163,7 +158,6 @@ impl Document {
             frame_builder: None,
             output_pipelines: FastHashSet::default(),
             render_on_scroll,
-            render_on_hittest: false,
             hit_tester: None,
             dynamic_properties: SceneProperties::new(),
         }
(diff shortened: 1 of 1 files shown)


staktrace:
(This is needed for bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1465892)

mrobinson:
@bors-servo r-

@kats raised a good point, which is that this change makes it so that the hit tester is not updated during scrolling.

staktrace:
So I spent more time thinking about this and realized that in addition to the previous problem (hit tester not getting updated during scrolling), there are other problems with this patch.

Fundamentally, the problem I'm trying to solve here is that there is a period of time after a scene swap but before the next render where the hit_tester in WR has internal state that does not correspond to the APZ state. The original patch I had for this PR attempted to fix that by updating the hit-tester right after the scene swap. However, the "APZ state" incorporates not just the display list, but async scrolling offsets and scrollbar transforms. These are applied at [here](https://github.com/servo/webrender/blob/8e697f8cb1f1aab2e5f6b9b903eb7191340b10c5/webrender/src/render_backend.rs#L1011) and [here](https://github.com/servo/webrender/blob/8e697f8cb1f1aab2e5f6b9b903eb7191340b10c5/webrender/src/clip_scroll_node.rs#L465) respectively, which are during the render phase. So the hit-tester really only have a correct state if we update it after a render (which is what we're doing now), or if we extract those bits from the rendering step and have the ability to do them independently of a full render (which seems really convoluted).

That leaves us with one of two possible solutions: (1) Start respecting the `render_on_hittest` flag again (which is currently set in the right places but never respected) and do a render if we get a hit-test request while this flag is set. This will effectively resurrect the code I added in #2343. (2) Always do a render after a scene swap. We generally need to do a render after a scene swap anyway; currently we wait until the following vsync to do it but if we are caching enough stuff across renders then we might as well do it right after the scene swap and then at the vsync the render will be mostly a no-op.

I'm leaning heavily towards approach (2). I had actually wanted to do this earlier but I ran into issues where Gecko would get confused as to how many frames were inflight and would get stuck always waiting for a non-existent frame. However I have a better solution now that seems to work. I'll put up the PR for that but if anybody prefers approach (1) I'm happy to discuss it further.

staktrace:
Patch updated. @kvark and @nical you might want to take a look as well in case you have any thoughts on this.

nical:
I also prefer (2). Looks good to me!

staktrace:
rebased ^

This is ready for proper review (in case it wasn't clear)

PR #2798, opened 2018-06-01
Add a debug indicator showing when frames and scene are built.

nical:
This adds a simple way to see when we are rendering frames and building scenes by having a little bar at the top left of the screen that moves if anything changes (the blue one is for frames and the red one for scenes).

screenshot from 2018-06-01 17-43-42

We seem to run into situations in gecko sometimes where we continuously render at 60fps even though nothing seems to be changing on screen and the motivation is to have a non-intrusive way to spot when that happens (right now the only way is to is to have the profiler HUD on at all times).

This is a simple system and could be used for more things, for example have a pref that sets a time budget per frame beyond which we move an indicator, or to know when lots of textures get uploaded or when the number of batches is higher than what we'd like, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2798)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -266,6 +266,7 @@ impl Document {
         resource_cache: &mut ResourceCache,
         gpu_cache: &mut GpuCache,
         resource_profile: &mut ResourceProfileCounters,
+        is_new_scene: bool,
     ) -> RenderedDocument {
         let accumulated_scale_factor = self.view.accumulated_scale_factor();
         let pan = self.view.pan.to_f32() / accumulated_scale_factor;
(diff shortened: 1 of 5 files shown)


nical:
r? anyone

webrender/src/render_backend.rs
```diff
@@ -1040,6 +1040,10 @@ impl RenderBackend {
 
         debug_assert!(op.render || !op.composite);
 
+        if op.build {
```
gw3583: If I'm understanding the code correctly, this won't actually be seen until the next render anyway.

Perhaps it would be better to store if the scene was built in the render backend, and pass this as a flag during the next render call?

gw3583:
I like the idea of this, nice! One comment above and a compile error below:

```
error[E0432]: unresolved import `debug_render`
 --> webrender/src/profiler.rs:6:5
  |
6 | use debug_render::DebugRenderer;
  |     ^^^^^^^^^^^^ Maybe a missing `extern crate debug_render;`?
```

nical:
@glennw I updated the PR, it now stores a field in the rendered document instead of adding a message.

PR #2799, opened 2018-06-04
Invalidate drop-shadow data when local rect is animating.

### LINKED ISSUE #2721: drop-shadow() filter is not updated in response to off-main-thread animations
https://bugzilla.mozilla.org/show_bug.cgi?id=1458661

[Testcase 1](https://bug1458661.bmoattachments.org/attachment.cgi?id=8972679)
[Testcase 2](https://bug1458661.bmoattachments.org/attachment.cgi?id=8972680)
[Testcase 3](https://bug1458661.bmoattachments.org/attachment.cgi?id=8972944) (works correctly but might still be worth confirming with once the bug is fixed)

In these testcases the black shadow lags behind. In the second testcase, the shadow gets squeezed and stretched in funny ways as the bounds of the filter input change.

gw3583:
If the local rect of the content of a drop-shadow is changing
each frame, we also need to invalidate the GPU cache data for
the drop-shadow primitive.

Unfortunately we don't have a way in wrench right now to create
a regression test for this.

Fixes #2721.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2799)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -467,6 +467,14 @@ impl PicturePrimitive {
                 pic_state.tasks.push(render_task_id);
                 self.surface = Some(PictureSurface::RenderTask(render_task_id));
 
+                // If the local rect of the contents changed, force the cache handle
+                // to be invalidated so that the primitive data below will get
+                // uploaded to the GPU this frame. This can occur during property
+                // animation.
+                if pic_state.local_rect_changed {
+                    frame_state.gpu_cache.invalidate(&mut self.extra_gpu_data_handle);
+                }
+
                 if let Some(mut request) = frame_state.gpu_cache.request(&mut self.extra_gpu_data_handle) {
                     // TODO(gw): This is very hacky code below! It stores an extra
                     //           brush primitive below for the special case of a
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -2473,6 +2473,7 @@ impl PrimitiveStore {
                 if new_local_rect != metadata.local_rect {
                     metadata.local_rect = new_local_rect;
                     frame_state.gpu_cache.invalidate(&mut metadata.gpu_location);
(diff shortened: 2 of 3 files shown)


webrender/src/picture.rs
```diff
@@ -467,6 +467,14 @@ impl PicturePrimitive {
                 pic_state.tasks.push(render_task_id);
                 self.surface = Some(PictureSurface::RenderTask(render_task_id));
 
+                // If the local rect of the contents changed, force the cache handle
+                // to be invalidated so that the primitive data below will get
+                // uploaded to the GPU this frame. This can occur during property
+                // animation.
+                if pic_state.local_rect_changed {
```
kvark: when is it ever reset to `false`?

gw3583: The ```PictureState``` struct is constructed every frame, during the recursive traversal.

gw3583: (It's just a stack allocated helper struct passed down to child pictures).

webrender/src/prim_store.rs
```diff
@@ -2473,6 +2473,7 @@ impl PrimitiveStore {
                 if new_local_rect != metadata.local_rect {
                     metadata.local_rect = new_local_rect;
                     frame_state.gpu_cache.invalidate(&mut metadata.gpu_location);
+                    pic_state.local_rect_changed = true;
```
kvark: any reason we can't just invalidate `pic_state.extra_gpu_data_handle` right here?

gw3583: ```extra_gpu_data_handle``` exists on the ```Picture``` struct, not in ```PictureState```. We *do* have access to the picture here, and *could* invalidate it here - however, the way that ```extra_gpu_data_handle``` is used depends on the type of ```Picture```, and so it seems tidier to have that done in the code that deals with specific picture types, rather than adding a new ```match``` here on the picture type.

PR #2802, opened 2018-06-05
Remove TRANSFORM variant of text run shader.

gw3583:
This is a step towards converting the text run shader to be
a brush shader. Like brush shaders, we now determine whether to
apply pixel snapping based on a conditional, rather than a new
shader and batch break.

Remove TransformBatchKind enum, since it's no longer used.

Change the color of B_Blend shader in the profiler, since the
existing color was very close to the color used by the
B_LinearGradient shader, and hard to distinguish between them.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2802)

### DIFF
--- a/webrender/res/ps_text_run.glsl
+++ b/webrender/res/ps_text_run.glsl
@@ -22,11 +22,17 @@ VertexInfo write_text_vertex(vec2 clamped_local_pos,
                              PictureTask task,
                              RectWithSize snap_rect,
                              vec2 snap_bias) {
-#if defined(WR_FEATURE_GLYPH_TRANSFORM) || !defined(WR_FEATURE_TRANSFORM)
     // Ensure the transform does not contain a subpixel translation to ensure
     // that glyph snapping is stable for equivalent glyph subpixel positions.
-    scroll_node.transform[3].xy = floor(scroll_node.transform[3].xy + 0.5);
+#if defined(WR_FEATURE_GLYPH_TRANSFORM)
+    bool remove_subpx_offset = true;
+#else
+    bool remove_subpx_offset = scroll_node.is_axis_aligned;
 #endif
+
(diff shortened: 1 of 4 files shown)


gw3583:
Pending try run for completeness, although this change should not cause any change in behavior:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=62f61d9b23308388f0418f820a673880b51541ef

nical:
R2 failures look unrelated. Coming from the border brush changes?

webrender/res/ps_text_run.glsl
```diff
@@ -22,12 +22,18 @@ VertexInfo write_text_vertex(vec2 clamped_local_pos,
...
                              RectWithSize snap_rect,
                              vec2 snap_bias) {
-#if defined(WR_FEATURE_GLYPH_TRANSFORM) || !defined(WR_FEATURE_TRANSFORM)
     // Ensure the transform does not contain a subpixel translation to ensure
     // that glyph snapping is stable for equivalent glyph subpixel positions.
-    scroll_node.transform[3].xy = floor(scroll_node.transform[3].xy + 0.5);
+#if defined(WR_FEATURE_GLYPH_TRANSFORM)
+    bool remove_subpx_offset = true;
+#else
+    bool remove_subpx_offset = scroll_node.is_axis_aligned;
```
kvark: shouldn't this be the opposite of `is_axis_aligned`?

gw3583: The previous logic would include this conditional compilation on ```!defined(WR_FEATURE_TRANSFORM)``` which is the same as ```is_axis_aligned```, I think.

gw3583:
Yup, the R2 failures are unrelated failures from the border brush PR.

PR #2804, opened 2018-06-05
Remove a couple of unused fields.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2804)


gw3583:
Noticed these are unused while profiling some gradient code.

r? anyone

Issue #2666, opened 2018-04-17
Document things.


nical:
There is a lot of concepts in WebRender, most of which aren't explained anywhere and the knowledge is spread over many timezones (well, mostly concentrated on a timezone far away from Europe). We should at least have a header comment on the top of the files that contain important things and that comment should explain the general idea and how it interacts with other important things.

- [ ] render task
- [ ] picture (and what's the difference with a render task)
- [ ] render pass
- [ ] brush and segments

mrobinson:
We should also update the clipping documentation which is wildly out of date.

PR #2806, opened 2018-06-05
Improve clipping documentation

mrobinson:
The old documentation was very out of date. The new documentation
is slightly less complete, but should be much more accurate.

This is part of #2666.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2806)

### DIFF
--- a/webrender/doc/CLIPPING.md
+++ b/webrender/doc/CLIPPING.md
@@ -1,97 +1,53 @@
-# Clipping in WebRender
+# Clipping and Positioning in WebRender
 
-The WebRender display list allows defining clips in two different ways. The
-first is specified directly on each display item and cannot be reused between
-items. The second is specified using the `SpecificDisplayItem::Clip` display item
-and can be reused between items as well as used to define scrollable regions.
+Each non-structural WebRender display list item has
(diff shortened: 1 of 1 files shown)


webrender/doc/CLIPPING.md
```diff
@@ -1,97 +1,53 @@
...
-   fixed.  ([github issue](https://github.com/servo/webrender/issues/1090))
-
-1. It should be possible to specify more than one predefined clip for an item.
-   This is necessary for items that live in a scrolling frame, but are also
-   clipped by a clip that lives outside that frame.
-   ([github issue](https://github.com/servo/webrender/issues/840))
+# Clipping and Positioning in WebRender
+
+Each non-structural WebRender display list item has
+ * A `ClipId` of a positioning node
```
kvark: isn't it a little weird that the positioning node is defined by `ClipId` which has nothing to do with clipping?

mrobinson: It certainly is. I think if we do switch to using two different ids, it makes sense to call one `ClipScrollNodeId` and `ClipChainId` or something like that.

PR #2809, opened 2018-06-06
support arbitrarily large font sizes

lsalzman:
This fixes Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1432241

At present, we just have a rudimentary check that ignores all fonts whose size is exceeds 4096, with no regard for device pixel scale or even transforms. This thus leads to probably large text will simply disappear from web-pages.

Gecko's Skia backend would normally, after the actual device size hits 1024, would render a glyph as a path instead of using glyph rasterization, which limits the growth of the glyph cache to be within reason.

Since we don't currently have an easy way of using paths right now in WR, as discussed with Glenn, this instead limits the maximum device size of the font to 1024 (like Skia), but from there just scales the glyph in the shader using the same mechanism we use for bitmap glyphs to support arbitrarily large scales. 

While not perfect, this at least allows large text to show up, rather than completely disappear and has similar benefits with respect to limiting the growth of the glyph cache without yet requiring us to conquer native path rendering.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2809)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1094,7 +1094,7 @@ impl AlphaBatchBuilder {
 
                 let font = text_cpu.get_font(
                     ctx.device_pixel_scale,
-                    Some(scroll_node.transform),
+                    scroll_node.transform,
                 );
                 let subpx_dir = font.get_subpx_dir();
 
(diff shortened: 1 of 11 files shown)


lsalzman:
Fixed all review comments. Reftest still in progress.

PR #2811, opened 2018-06-07
Move examples in a separate crate

zakorgy:
This change separates the examples from the WR crate, and removes unrelated dependencies from it.

In the future this can help the transition to new backends, like adding new features which should not exist in the WR crate.

Any thoughts? @kvark @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2811)


mrobinson:
@zakorgy Do you explaining a bit more how this helps in the development of new backends?

webrender/Cargo.toml
```diff
@@ -63,10 +63,6 @@ optional = true
 
 [dev-dependencies]
 mozangle = "0.1"
```
kvark: why is `mozangle` left?

zakorgy: Because it is used by the `tests` folder not by the `examples`.

kvark:
@mrobinson the motivation is to have "webrender" crate independent of the backend at the type level. I.e. wrench, examples, servo,etc need to choose the Metal/D3D12/Vulkan graphics backend with accompanying window integration crates (like `winit`), so it makes sense to have those dependencies moved out.

Examples is the only thing that's not a separate crate among the ones that need to actually create a window and initialize WR. Other projects, like gfx-rs or Amethyst, also move their examples out for this reason.

mrobinson:
@kvark Thanks for the explanation.

kvark:
There doesn't appear to be strong opposition to doing this, even though there isn't much support either. Let's get it moving.
@bors-servo r+

PR #2813, opened 2018-06-08
Unify show, load, and replay Wrench commands

mrobinson:
The three commands all load different kinds of traces and recordings.
Since we know what behavior we want based on the argument given, we can
unify these three commands.

Currently, the arguments are still specific to each particular recording
type, but eventually they can be made functional for all of them or
removed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2813)

### DIFF
--- a/wrench/src/main.rs
+++ b/wrench/src/main.rs
@@ -363,6 +363,31 @@ fn create_notifier() -> (Box<RenderNotifier>, Receiver<NotifierEvent>) {
     (Box::new(Notifier { tx: tx }), rx)
 }
 
+fn rawtest(mut wrench: Wrench, window: &mut WindowWrapper, rx: Receiver<NotifierEvent>) {
+    RawtestHarness::new(&mut wrench, window, &rx).run();
+    wrench.shut_down(rx);
+}
+
+fn reftest<'a>(
+    mut wrench: Wrench,
+    window: &mut WindowWrapper,
+    subargs: &clap::ArgMatches<'a>,
+    rx: Receiver<NotifierEvent>
+) -> usize {
+    let dim = window.get_inner_size();
+    let base_manifest = Path::new("reftests/reftest.list");
+    let specific_reftest = subargs.value_of("REFTEST").map(|x| Path::new(x));
+    let mut reftest_options = ReftestOptions::default();
+    if let Some(allow_max_diff) = subargs.value_of("fuzz_tolerance") {
+        reftest_options.allow_max_difference = allow_max_diff.parse().unwrap_or(1);
+        reftest_options.allow_num_differences = dim.width as usize * dim.height as usize;
+    }
+    let num_failures = ReftestHarness::new(&mut wrench, window, &rx)
+        .run(base_manifest, specific_reftest, &reftest_options);
+    wrench.shut_down(rx);
+    num_failures
+}
+
 fn main() {
     #[cfg(feature = "env_logger")]
     env_logger::init();
@@ -440,10 +465,8 @@ fn main() {
         notifier,
     );
 
-    let mut thing = if let Some(subargs) = args.subcommand_matches("show") {
-        Box::new(YamlFrameReader::new_from_args(subargs)) as Box<WrenchThing>
-    } else if let Some(subargs) = args.subcommand_matches("replay") {
-        Box::new(BinaryFrameReader::new_from_args(subargs)) as Box<WrenchThing>
+    if let Some(subargs) = args.subcommand_matches("show") {
+        render(&mut wrench, &mut window, size, &mut events_loop, subargs);
     } else if let Some(subargs) = args.subcommand_matches("png") {
         let surface = match subargs.value_of("surface") {
             Some("screen") | None => png::ReadSurface::Screen,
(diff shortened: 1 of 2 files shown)


wrench/src/main.rs
```diff
@@ -440,10 +468,8 @@ fn main() {
         notifier,
     );
 
-    let mut thing = if let Some(subargs) = args.subcommand_matches("show") {
-        Box::new(YamlFrameReader::new_from_args(subargs)) as Box<WrenchThing>
-    } else if let Some(subargs) = args.subcommand_matches("replay") {
-        Box::new(BinaryFrameReader::new_from_args(subargs)) as Box<WrenchThing>
+    if let Some(subargs) = args.subcommand_matches("show") {
+        render(&mut wrench, &mut window, size, &mut events_loop, subargs);
```
Gankro: does this need to check for the aliases or are they all mapped to "show" by the argument parser?

wrench/src/main.rs
```diff
@@ -665,17 +700,13 @@ fn main() {
...
             let pixels = wrench.renderer.read_pixels_rgba8(rect);
             save_flipped("screenshot.png", pixels, size);
         }
-        Some(ref mut events_loop) => {
-            events_loop.run_forever(|event| body(&mut wrench, event));
-        }
+        Some(ref mut events_loop) => events_loop.run_forever(|event| body(wrench, event)),
     }
-
-    wrench.renderer.deinit();
```
Gankro: what's this part of the change about?

Gankro:
r=me with comments addressed, and assuming kvark is ok with doing this (i am very happy that we are).

wrench/src/main.rs
mrobinson: I was a little unsure about this as well, but I confirmed that 'show', 'load', and 'replay' all work.

wrench/src/main.rs
mrobinson: The call do `deinit()` moved to the caller in order to keep it in the same function as the call to the constructor.

mrobinson:
@Gankro Thanks for the review! I'll wait for feedback from @kvark.

wrench/src/main.rs
```diff
@@ -363,6 +363,34 @@ fn create_notifier() -> (Box<RenderNotifier>, Receiver<NotifierEvent>) {
     (Box::new(Notifier { tx: tx }), rx)
 }
 
+fn rawtest(mut wrench: Wrench, window: &mut WindowWrapper, rx: Receiver<NotifierEvent>) {
+    {
+        let harness = RawtestHarness::new(&mut wrench, window, &rx);
```
kvark: nit: any reason to not just do `RawtestHarness::new(...).run()` without an explicit scope?

wrench/src/main.rs
```diff
@@ -491,16 +498,44 @@ fn main() {
...
-        let path = PathBuf::from(subargs.value_of("path").unwrap());
-        let mut documents = wrench.api.load_capture(path);
+    } else {
+        panic!("Should never have gotten here! {:?}", args);
+    };
+
+    wrench.renderer.deinit();
+}
+
+fn render<'a>(
```
kvark: nice moving logic out into standalone functions!
I wonder if we should make those to be methods of `Wrench`?

wrench/src/main.rs
```diff
@@ -491,16 +498,44 @@ fn main() {
...
+    wrench: &mut Wrench,
+    window: &mut WindowWrapper,
+    size: DeviceUintSize,
+    events_loop: &mut Option<winit::EventsLoop>,
+    subargs: &clap::ArgMatches<'a>,
+) {
+    let input_path = subargs.value_of("INPUT").map(|s| PathBuf::from(s)).unwrap();
```
kvark: nit: `map(PathBuf::from)`

wrench/src/main.rs
```diff
@@ -491,16 +498,44 @@ fn main() {
...
         println!("loaded {:?}", documents.iter().map(|cd| cd.document_id).collect::<Vec<_>>());
         let captured = documents.swap_remove(0);
         window.resize(captured.window_size);
         wrench.document_id = captured.document_id;
         Box::new(captured) as Box<WrenchThing>
     } else {
-        panic!("Should never have gotten here! {:?}", args);
+        let extension = input_path
+            .extension()
+            .unwrap_or_else(|| panic!("Tried to render with an unknown file type."))
```
kvark: le'ts just `expect()`

wrench/src/main.rs
```diff
@@ -491,16 +498,44 @@ fn main() {
...
+            .to_str()
+            .unwrap_or_else(|| panic!("Tried to render with an unknown file type."));
+
+        if "yaml" == extension {
```
kvark: let's just `match extension`

wrench/src/main.rs
mrobinson: No reason. I just moved this code without modifying it. This should be fixed now though.

wrench/src/main.rs
mrobinson: I think that'd be a great idea. Perhaps a followup patch to keep the change as small as possible?

wrench/src/main.rs
mrobinson: Fixed.

wrench/src/main.rs
mrobinson: Much better. Fixed.

wrench/src/main.rs
mrobinson: Okay. This should be fixed.

mrobinson:
Thanks for the review! The new version of the branch should have all of these issues fixed.

kvark:
Thank you! We can certainly leave the wrench methodification for follow-ups ;)
@bors-servo r+

PR #2822, opened 2018-06-18
Handle zero-width sized borders without crashing.

### LINKED ISSUE #2812: Zero width borders break WR
While updating to the latest WR I discovered the following YAML:

<pre>
root:
  items:
    - type: stacking-context
      bounds: [0, 0, 500, 500]
      items:
        - type: border
          bounds: [ 0, 0, 50, 50 ]
          width: [ 0, 0, 0, 0 ]
          border-type: normal
          style: [ solid, solid, solid, solid ]
          color: [ blue, blue, blue, blue ]
</pre>

breaks WR with the following assertion:

> thread 'WRRenderBackend#0' panicked at 'assertion failed: descriptor.size.width > 0 && descriptor.size.height > 0', webrender\src\texture_cache.rs:801:9

I think this has to do with moving borders to the brush code path and caching them in the texture cache. 

Probably Gecko/Servo stop completely zero width borders from hitting WR at all. Should WR treat this as a no op rather than hitting an assertion?

gw3583:
Fixes #2812.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2822)


gw3583:
Pending try here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=30a097c1045a0c846a6aedd395a707859fa68640

r? @anyone

gw3583:
Try run looks good, and the appveyor failure appears to be a network intermittent failure.

Issue #2583, opened 2018-03-28
`wrench --save yaml load capture` doesn't work


Gankro:
The example renders fine, and the yaml_frames dir is created, but it's empty.

Gankro:
note: this appears to not work because `load` injects content at a level below `--save yaml` (yaml_frame_writer) reads. A potential solution is to just push yaml_frame_writer lower down the stack.

Issue #2411, opened 2018-02-13
API recorder doesn't support external images


staktrace:
Right now running Firefox with the ENABLE_WR_RECORDING=1 changes behaviour in that Firefox stops sending WR external images. This can make bugs disappear under recording which makes debugging more painful.

We should have WR record external images to the binary recorder and make wrench provide an external image handler that returns the data from the recorded file.

This goes with https://bugzilla.mozilla.org/show_bug.cgi?id=1437925

kvark:
I'd rather focused our efforts on polishing the external image support on capturing infrastructure (e.g. making it possible to replay an scene with `ImageBufferKind::TextureExternal` on non-embedded GL) than trying to upgrade the binary recordings.

glennw:
Should this be closed then?

Issue #2239, opened 2017-12-18
[RFC] Capture: stable hierarchical textual representation


kvark:
There is one case where YAML is most relevant: reftests. We hand-rolled our text reftest format with the detached read/write implementations that we've been using for a year now:
  - *good*: syntax is way more stable than the actual `DisplayItem` representation (which is currently directly serialized by capturing as well as saving as RON/JSON).
  - *bad*: read and write fall out of sync, there is a lot of hand-written code in there.

---

There is something that we could build though, that would be as easy to read/write as the current reftest format and yet much easier to support. Basic idea is to define a number of types for the pure purpose of stable and pretty serialization (so the types are going to derive the serde traits). These types are simplified version of our display item types.

For example, a transform type could be:
```rust
enum TransformOp {
  Translate(Vector),
  Rotate(Axis, Angle),
  Scale(Scalar),
}
enum Transform {
  Composite(Vec<TransformOp>),
  Raw(Matrix4),
}
```

When writing the serialized syntax for these, ability to skip type names is much welcome:
```ron
transform: Composite([Translate([1,2,3]), Rotate(X, 25)]),
```
Arguably, it's not exactly as simple as it is now, so that's a possible downside.

---
The idea behind this prettified serialization layer is that:
  1. It stays stable! When it changes - we should be able to detect it at compile time and adjust all the reftests (that would otherwise straight fail to load).
  2. The serialization code is derived automatically, so it's one less thing to worry about.
  3. We only need to write conversion to/from this stable formats from our internal `DisplayItem` things.
  4. It's readable/writable and hierarchical(!), unlike the current JSON/RON dumps that are linear.

jrmuizel:
I don't feel that strongly about this. It would be interesting to see what test cases look like in this format. As an intermediate step we could keep around both forms to get a feel for the advantages/disadvantages.

PR #2825, opened 2018-06-18
Remove old recording APIs.

gw3583:
These are not maintained or used now that we have scene capture
functionality working.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2825)


gw3583:
In theory, there is still a use case for the old binary recording system - it supports recording a sequence of frames.

In practice, this seems un-maintained (it didn't work when I tested it today), and I don't intend to use it again now that we have scene capture functionality working reliably.

What do you think @kvark @Gankro @mrobinson @jrmuizel ?

gw3583:
If / when this lands, we can close the following:

https://github.com/servo/webrender/issues/2583
https://github.com/servo/webrender/issues/2411
https://github.com/servo/webrender/issues/2231
https://github.com/servo/webrender/issues/2131
https://github.com/servo/webrender/issues/2201
https://github.com/servo/webrender/issues/1253
https://github.com/servo/webrender/issues/1402
https://github.com/servo/webrender/issues/1826

mrobinson:
@gw3583 I'm okay with this. We briefly discussed a better high level API (like the current YAML one) last week, but I think the consensus was that we should be using serde serialization instead of a custom deserializer.

kvark:
@mrobinson 
> We briefly discussed a better high level API (like the current YAML one) last week, but I think the consensus was that we should be using serde serialization instead of a custom serializer.

To be fair, the high-level API we discussed is mostly unrelated to the binary recordings. And the binary stuff uses Serde for encoding/decoding (unlike YAML).

@gw3583 
> In practice, this seems un-maintained (it didn't work when I tested it today), and I don't intend to use it again now that we have scene capture functionality working reliably.

I think a better solution to this problem in the long term would be sequential capturing. Most of the capture time today comes from reading back texture data, and this would be minimal as long as we keep track of what's already on disk and just re-use it. So at the end of the day we'll end up with a sequence of captured frames sharing the resources, and we'll add some controls to wrench to go back and forth between the frames.

So yeah, it does seem like the binary recording can be dropped, even though it's a fairly small chunk of code.

mrobinson:
> To be fair, the high-level API we discussed is mostly unrelated to the binary recordings. And the binary stuff uses Serde for encoding/decoding (unlike YAML).

My comment was just in relation to the bits of this PR that remove the YAML frame writer. I have no reservations at all about removing the binary bits.

kvark:
> ... even though it's a fairly small chunk of code.

...
> +1 −2,470

:fearful:

jrmuizel:
The big advantage of being able to record to yaml is that the output is stable across revisions of WebRender. I would be sad to lose this.

gw3583:
@kvark How feasible do you think it would be to make the scene capture functionality also write out a stable representation of the scene-1-0.ron file? I guess this would semantically be writing a YAML file from the display list in the scene.ron file? That would probably give us a stable representation across WR versions. Or do you have other ideas that might provide that functionality?

kvark:
@gw3583 the idea we've been discussing at All Hands is making a semi-stable structures defined that would be serialized to/from *instead* of RON. That would make the current YAML parser irrelevant, and all the conversion logic will by type-checked properly (but we still need to write/maintain the conversion code from the current internal representation of scenes into that stable definition). This is exactly what #2239 is proposing.

Another (probably simpler, but not as beneficial in the long term) solution would be to attempt to address #2583 (get internal RON -> YAML conversion with Wrench).

gw3583:
Let's close this for now - we probably don't want to remove this completely until we have in place a solution for multi-frame recordings.

PR #2826, opened 2018-06-18
Remove arrays from vertex output

dati91:
We are trying to generate metal shader with `spirv_cross` and hitting and error:
```
"The vertex function output structure may not include a matrix or array."
```

cc @kvark @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2826)

### DIFF
--- a/webrender/res/cs_border_segment.glsl
+++ b/webrender/res/cs_border_segment.glsl
@@ -6,8 +6,10 @@
 
 // For edges, the colors are the same. For corners, these
 // are the colors of each edge making up the corner.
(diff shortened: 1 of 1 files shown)


gw3583:
@dati91 Looks good, thanks. Could you add an entry to https://github.com/servo/webrender/wiki/Driver-issues describing this issue?

@bors-servo r+

dati91:
@gw3583 done: https://github.com/servo/webrender/wiki/Driver-issues#2826-varying-arrays

PR #2829, opened 2018-06-19
Trust image requests in their dirtyness and verify with an extra assertion

kvark:
This is a follow-up to #2796 and #2774 that should fix https://bugzilla.mozilla.org/show_bug.cgi?id=1468713
The use case (supposedly) happening here is:
  - during an image request, we figured that the texture has been evicted, so we ignore the dirty rectangle and don't use any intersection result with the viewport
  - during the uploading of textures, we try to intersect the original dirty rectangle with the viewport and realize that it's not actually visible. Problem here is not taking into account the forced upload switch, which is now fixed.

Instead of ignoring a request, we are now asserting for the dirty rect visibility result to match our expectation made during `request_image`.

r? @staktrace 
Note: the second commit is cosmetic

Edit: the third commit fixes the coordinate space of the dirty rectangle. When it reaches our texture cache, there is no longer any knowledge of the full image size if it's tiled, so we correctly transform it into the local space. This fixes the slice addressing later on when the image needs to be uploaded.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2829)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -525,7 +525,7 @@ impl ResourceCache {
         &mut self,
         image_key: ImageKey,
         descriptor: ImageDescriptor,
-        mut data: ImageData,
+        data: ImageData,
         dirty_rect: Option<DeviceUintRect>,
     ) {
         let max_texture_size = self.max_texture_size();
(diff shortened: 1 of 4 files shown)


kvark:
@staktrace this should be ready now
try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=7b480ad64ed6306a9987fb8be3883f5f6e5c772e

staktrace:
@kvark this looks good and fixes the crash. There is another crash on the same page but I'll chase that down separately. r+

kvark:
@staktrace thanks for verifying! I do wonder about that other crash though, if it's related or not...
@bors-servo r=staktrace

kvark:
(as per IRC, last commit is reviewed by Glenn)
@bors-servo r=staktrace,gw3583

PR #2830, opened 2018-06-19
snap the text run offset and glyph offsets separately

lsalzman:
This fixes issues introduced by #2738. Since the transform offset was snapped separately from the text/glyph offsets, this led to problems where certain Servo reftests might migrate a subpixel offset between the transform and the text offset.

As a compromise, Glenn and I agreed it would be best to combine the transform and text offset for snapping, then separately base the subpixel positioning off of only the glyph offset. This both keeps the resolution of Gecko's scrolling issue while addressing the Servo issue.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2830)


lsalzman:
Try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e494ef861d68050b897139e23abf43a248faf1e1

gw3583:
This looks good to me! Could we add a reftest for it?

lsalzman:
Fixes shadow-transforms and add a new reftest

PR #2832, opened 2018-06-20
Get the new scene indicator to work with async scene building.

nical:
The recently added new scene indicator only works without async scene building, we just need to patch the information that we just built the scene through `update_document`'s arguments.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2832)


nical:
r? anyone

staktrace:
LGTM

PR #2837, opened 2018-06-21
Update image and png

Eijebong:
This also removes a dependency on num-traits 0.1

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2837)


kvark:
nice!
@bors-servo r+

PR #2838, opened 2018-06-22
use WS_EX_NOREDIRECTIONBITMAP direct-composition app

ekicyou:
[[with_no_redirection_bitmap](https://github.com/tomaka/winit/commit/047c67baf31cf9bcdc6f0275037bb3b56f2bd268)] API added to master branch of [winit].

It is useful for direct-composition samples.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2838)


ekicyou:
There is a version mismatch of winapi, and appveyor does not pass.
Because it is difficult to solve, PR will be done as it is.

atouchet:
For what it's worth winit 0.16 is now out which includes WS_EX_NOREDIRECTIONBITMAP .

ekicyou:
Thanks. Changed winit.

kvark:
There are legitimate CI errors that need to be addressed before we move this forward.

ekicyou:
In order to use winit = "0.16.0" you need rust = "1.24.0",
but in AppVeyor environment rust = "1.23.0" is used.

```rust
#[inline]
#[stable(feature = "refcell_replace", since="1.24.0")]
pub fn replace(&self, t: T) -> T {
    mem::replace(&mut *self.borrow_mut(), t)
}
```

```
Start-FileDownload "https://static.rust-lang.org/dist/rust-1.23.0-${env:TARGET}.msi"
Downloading rust-1.23.0-x86_64-pc-windows-msvc.msi (111,887,897 bytes)...100%
msiexec /passive /i "rust-1.23.0-%TARGET%.msi" ADDLOCAL=Rustc,Cargo,Std INSTALLDIR=C:\Rust
rustc -V
rustc 1.23.0 (766bd11c8 2018-01-01)
```


Since we can not judge the build environment update, we suspend it as it is.

kvark:
According to https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox, we should be safe bumping all CIs to use Rust 1.26. Let's just do this in a separate PR.

ekicyou:
Thanks for the suggestion.
Since July 5 it seems that Rust 1.27.0 can be used, so please wait a while.

However, the task of unifying "winapi = 0.3.5" appears complicated.
I am sorry but I do not have the motivation to grasp the "webrender" body, which is difficult.

I think that it may be necessary to separate PR.

* rust 1.27.0 update
* winit = "0.16.0" update
  * winapi = "0.3.5" update
* use WS_EX_NOREDIRECTIONBITMAP

ekicyou:
fnn... Update https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox.
Check with PR # 2870.

ekicyou:
This PR has more steps than expected, and an unreleased rustc1.28.0-stable is required.

Judge it as premature, close PR once.
Thank you for your advice.

PR #2839, opened 2018-06-22
Reduce per-instance size from 32 -> 16 bytes.

gw3583:
Add a new PrimitiveHeader concept, which contains all the common
information for a given primitive (such as scroll node, render
task, z value etc).

Instances contain an index to the primitive header, and read the
information from that. This allows the instance size to be
halved. There is a slight extra vertex shader cost, due to
an extra indirection. However, this was not apparent in any
profiling I did, since the vertex shader time typically makes
up a very small amount of overall time. Additionally, the
primitive header texels are likely to be in the texture cache
for each glyph that is fetched. A possible future improvement
is to remove that extra indirection by writing some of that
data, such as the render task address, directly into the primitive
header.

On sites such as nytimes.com or wikipedia.org, where there are
typically a large number of glyphs, this saves several hundred
kB per frame of data being sent to the GPU, which improves the
compositor / driver CPU time.

Also remove the local_clip_chains vector and texture. Now, we just
calculate this as part of the primitive header and include it
there. This can save significant amounts of texture upload on
pages that have a lot of clip chains or clip nodes, relative to
the visible primitive count.

In future, both the PrimitiveHeader and instance structures can
easily be further compressed again, but this can be done as a
follow up.

Finally, although this is a useful optimization by itself, it's
mostly prep work for some changes I have planned related to
how we store and upload scroll nodes. Hopefully these changes will
be both an optimization and make it easier to fix some of the
correctness issues we have with nested 3d transforms.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2839)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -16,7 +16,6 @@ void brush_vs(
     vec4 segment_data
 );
 
-#define VECS_PER_BRUSH_PRIM                 2
 #define VECS_PER_SEGMENT                    2
 
 #define BRUSH_FLAG_PERSPECTIVE_INTERPOLATION    1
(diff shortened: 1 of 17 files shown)


staktrace:
I haven't seen that R8 failure anywhere else so I would lean towards it being related to this PR.

gw3583:
Review status: 16 of 17 files reviewed, 3 unresolved discussions (waiting on @kvark and @gw3583)

---

*[webrender/res/ps_text_run.glsl, line 129 at r1](https://reviewable.io/reviews/servo/webrender/2839#-LFebP7a0J4MPnDpvJdI:-LFnumZD-ehyVHkShoQw:b-nxdyqe) ([raw file](https://github.com/servo/webrender/blob/2a8355f4d3093f9a60a3c97736e12e137bd9dd71/webrender/res/ps_text_run.glsl#L129)):*

definitely feels like we can shave a few more bytes off an instance in the future
</blockquote></details>

Yup, plenty of room for further compression.

---

*[webrender/src/gpu_types.rs, line 211 at r1](https://reviewable.io/reviews/servo/webrender/2839#-LFed2-o8bN4AG6Vj6wr:-LFnuqkwAy7WvNCg6h5v:bhylzq2) ([raw file](https://github.com/servo/webrender/blob/2a8355f4d3093f9a60a3c97736e12e137bd9dd71/webrender/src/gpu_types.rs#L211)):*

I wonder if we could just quantize those, compressing and letting us have all the header info in the same texture
</blockquote></details>

I think that's quite possible as a follow up, yes.

---

*[webrender/src/renderer.rs, line 310 at r1](https://reviewable.io/reviews/servo/webrender/2839#-LFedvst8V_zVPnDMsUr:-LFnuvil67V91-hgkUpm:b-dp85x3) ([raw file](https://github.com/servo/webrender/blob/2a8355f4d3093f9a60a3c97736e12e137bd9dd71/webrender/src/renderer.rs#L310)):*

where did 10 go?
</blockquote></details>

Oops, forgot to reorder when I removed the clip rects texture. Fixed.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2839)*

gw3583:
@kvark Addressed the review comments. In terms of profiling, I typically see a very small cost in VS time, and a significant saving in render backend / compositor time (for example, on nytimes.com - VS time is now 5.4% of the scene, previously 5.2%, but compositor and backend time both drop by ~0.3ms, which is a significant portion of the total CPU time).

@staktrace I verified that the failure in R8 is occurring in a nightly build when WR is enabled - I'm not sure how that occurred, but it seems unrelated to this PR.

gw3583:
TC failure appears to be a network issue:

```
[taskcluster:error] Error: Error: (HTTP code 500) server error - {"message":"Get https://registry-1.docker.io/v2/staktrace/webrender-test/manifests/freetype28: received unexpected HTTP status: 503 Service Unavailable"}
 ```

kvark:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed, 2 unresolved discussions (waiting on @kvark)

---



*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2839#-:-LFoLf70EtFPhUcxmCsn:bnfp4nl)*

staktrace:
> @staktrace I verified that the failure in R8 is occurring in a nightly build when WR is enabled - I'm not sure how that occurred, but it seems unrelated to this PR.

@glennw did you verify this on try or on a local build? On try the R8 is 100% green without your patch and 100% failing with your patch. https://bugzilla.mozilla.org/show_bug.cgi?id=1470125#c4

gw3583:
@staktrace I did only verify with a local reftest run. So, (unless I messed up my testing somewhere) this means:

* That test fails on a nightly build running reftests locally, but not on CI.
* This patch makes that test failure also occur on CI.

Taking a closer look at the test, it uses ```scrollLeft``` - could it be that the screenshot is being taken in Gecko + WR for the reftest before that scroll applies? It seems unlikely, but that may explain why CI reports it as succeeding?

I'll investigate further today and see if I can work out what is happening.

gw3583:
@staktrace OK, so there are two bugs in play here, which was confusing me. There's a genuine timing bug where the DLs sent to WR aren't always the same for this test case. That was masking that there was also a genuine bug in this patch, related to local clip rects on images that are large enough to require tiling. The fix is https://github.com/servo/webrender/pull/2843.

staktrace:
Thanks for chasing this down! I've confirmed that #2843 fixes it on my try pushes. It's odd that the test is sending different DLs to WR though; I guess it might decide to flush a paint before the scroll but the reftest snapshot shouldn't happen until after the scroll.

PR #2840, opened 2018-06-22
Move device.rs in a folder

zakorgy:
Created a `device` folder where we can place new backend implementations in the future, and moved the existing `device.rs` implementation into it with a new name `(gl.rs) `.

cc @kvark @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2840)


gw3583:
Looks good to me - although probably worth waiting for #2839 to land before we merge this.

PR #2841, opened 2018-06-25
Cache the used font instance during prepare, for batching.

gw3583:
Instead of calling get_font() during batching, this is now
stored as a used_font field in the text run primitive.

This is a small optimization, but is mostly tidy up work in
preparation for some upcoming picture caching patches. The
reason for this change is to ensure that the batching code
doesn't need to access the *value* of the scroll node
matrices, which will be important once pictures can be
rasterized in different spaces.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2841)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1077,17 +1077,13 @@ impl AlphaBatchBuilder {
                 let text_cpu =
                     &ctx.prim_store.cpu_text_runs[prim_metadata.cpu_prim_index.0];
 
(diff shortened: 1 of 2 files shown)


gw3583:
r? @lsalzman 

Pending try here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=6974fd164f65dcaf26e0158f4f7dd0f6a01b2d70

PR #2846, opened 2018-06-26
Group scroll, reference and sticky node types into positioning nodes.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2846)

### DIFF
--- a/webrender/src/clip_scroll_node.rs
+++ b/webrender/src/clip_scroll_node.rs
@@ -46,9 +46,26 @@ impl StickyFrameInfo {
 }
 
 #[derive(Debug)]
-pub enum NodeType {
+pub enum SpatialNodeKind {
+    /// A special kind of node that adjusts its position based on the position
+    /// of its parent node and a given set of sticky positioning offset bounds.
+    /// Sticky positioned is described in the CSS Positioned Layout Module Level 3 here:
+    /// https://www.w3.org/TR/css-position-3/#sticky-pos
+    StickyFrame(StickyFrameInfo),
+
+    /// Transforms it's content, but doesn't clip it. Can also be adjusted
+    /// by scroll events or setting scroll offsets.
+    ScrollFrame(ScrollFrameInfo),
+
     /// A reference frame establishes a new coordinate space in the tree.
     ReferenceFrame(ReferenceFrameInfo),
+}
+
+#[derive(Debug)]
+pub enum NodeType {
+    Spatial {
+        kind: SpatialNodeKind,
+    },
 
     /// Other nodes just do clipping, but no transformation.
     Clip {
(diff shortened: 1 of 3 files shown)


gw3583:
I don't know if this makes sense - but I wanted to get some feedback on it.

First, some background. I'm prototyping some code to be able to rasterize a Picture in a different coordinate space. There's some correctness reasons for this (e.g. rasterizing nested 3d rendering contexts in local space) and some performance reasons (caching local space pictures).

To do this, I would like to minimize the size of the transform matrices we need to convert between spaces. Currently, we store and upload the clip-scroll nodes texture for every node in the CST, even clip nodes which don't actually provide different positioning information (although primitives do reference clip nodes, more on that later).

So the basic idea here is to group all positioning node types. From that, I'll be able to create a much smaller set of transform matrices, and store some common information that is relevant to all positioning nodes. The ```positioning_node_index``` field in the clip node provides a direct reference from a given clip node to the CST node which contains the actual positioning transform for this node.

I'm not totally convinced this is the right way to go - it's hard to know without a full implementation. But I thought I'd put this up now for feedback to see if it seems like a reasonable idea.

Thoughts @mrobinson @kvark ?

webrender/src/clip_scroll_node.rs
```diff
@@ -46,9 +46,26 @@ impl StickyFrameInfo {
 }
 
 #[derive(Debug)]
-pub enum NodeType {
+pub enum PositioningNodeType {
```
kvark: possibly alternative names: "SpatialNode", "PositionFrame"

webrender/src/clip_scroll_node.rs
```diff
@@ -46,9 +46,26 @@ impl StickyFrameInfo {
...
+    ScrollFrame(ScrollFrameInfo),
+
     /// A reference frame establishes a new coordinate space in the tree.
     ReferenceFrame(ReferenceFrameInfo),
+}
+
+#[derive(Debug)]
+pub enum NodeType {
+    Positioning {
+        kind: PositioningNodeType,
```
kvark: nit: minor semantics inconsistency - "kind" vs "XXType"

webrender/src/clip_scroll_node.rs
```diff
@@ -80,7 +90,7 @@ pub enum NodeType {
 impl NodeType {
     fn is_reference_frame(&self) -> bool {
         match *self {
-            NodeType::ReferenceFrame(_) => true,
+            NodeType::Positioning { kind: PositioningNodeType::ReferenceFrame(_), .. } => true,
```
kvark: do we need all those `..` sprinkled in `Positioning` matches?

webrender/src/clip_scroll_node.rs
gw3583: Renamed to Spatial :+1:

webrender/src/clip_scroll_node.rs
gw3583: Fixed

webrender/src/clip_scroll_node.rs
gw3583: Yep - if all goes according to plan, I'll need to add one common field to the Spatial node type, which will be used as an index / identifier into a (typically much smaller) array of transforms that we now need to send to the GPU.

gw3583:
OK, the local patches I'm working on to optimize the number of transforms we upload to the GPU (that are built on this) seem to be going well, so I think this is ready for review now.

kvark:
Thank you!
@bors-servo r+

PR #2847, opened 2018-06-26
Fix tiling and dirty rect calculations for external images

aosmond:
- When we have a dirty rect, we should cull all tiled requests which do
not overlap with the dirty rect, not just blob image requests.

- We should not take the dirty rect from the image template until we
have processed all of the requests. This will allow each tile to
re-upload only the modified segment.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2847)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -641,10 +641,17 @@ impl ResourceCache {
             return
         };
 
+        if !self.pending_image_requests.insert(request) {
+            return
+        }
+
+        // If we are tiling, then we need to confirm the dirty rect intersects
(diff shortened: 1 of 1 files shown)


staktrace:
@aosmond This PR causes crashtest failures in gecko, see try push at https://treeherder.mozilla.org/#/jobs?repo=try&revision=e989ff39da6b5c43911c03c9c1efa133a7e15adf

aosmond:
The backtraces don't spell out the offending line number, but I suspect the logic for unwrapping request.tile vs template.tiling was incorrectly inverted, but was fine before because it only applied to blob images. update_texture_cache does it the other way.

PR #2848, opened 2018-06-26
Implement nine patch support for border gradients

### LINKED ISSUE #1875: Are Border Gradients Sufficiently Expressive?
I'm still taking this all in, but here's a weird case that @mstange believes can't be captured with the arguments we pass to wr:

screen shot 2017-10-13 at 4 03 59 pm

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title>border-image with linear-gradient</title>
<style>

div {
  border: 30px solid;
  box-sizing: border-box;
  width: 200px;
  height: 200px;
  border-image: linear-gradient(45deg, red 50px, yellow 50px, yellow 150px, red 150px) 50;
}

</style>

<div></div>
```

mrobinson:
This will allow border-gradients to support the full richness of the CSS
API. There are some areas where the implementation doesn't match what
CSS expects, but this should move us a good deal closer. We can
gradually fix those issues and turn this on in Gecko.

Fixes #1875.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2848)


mrobinson:
This doesn't allow us to turn on WebRender border-image in Gecko yet, but gets us a little closer. We would need to investigate a bit more into why so many reference tests still fail.

gw3583:
Review status: all files reviewed, 1 unresolved discussion (waiting on @mrobinson)

---

*[webrender_api/src/display_item.rs, line 276 at r1](https://reviewable.io/reviews/servo/webrender/2848#-LFyHoET3kthta2ywYFe:-LFyHoET3kthta2ywYFf:b-xtcpdw) ([raw file](https://github.com/servo/webrender/blob/0e73f3fbd89b14b925e427970036cd2a327894f2/webrender_api/src/display_item.rs#L276)):*
> ```Rust
> 
> #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize)]
> /// TODO(mrobinson): Currently only images are supported, but we will
> ```

We can remove this comment now.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/webrender/2848)*

gw3583:
One minor nit above, r=me otherwise. It also looks like there is no edge AA between gradient stops in the radial gradient image, but I think that's just a pre-existing issue with the radial gradient shader.

mrobinson:
@bors-servo r=gw3583

@gw3583 Thanks for the review!

mrobinson:
For the purposes of not breaking the build, here's a Gecko try job (should not change behavior at all) for this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=76b9810cacc13df3dc227b73026f49775815317e&selectedJob=185117848

PR #2849, opened 2018-06-26
Report time spent for telemetry recording.

staktrace:
This records the amount of time spent per scene build, scene swap, and
render, and reports it via the gecko callback hooks so that it can be
recorded in telemetry.

r? anybody

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2849)

### DIFF
--- a/direct-composition/src/main_windows.rs
+++ b/direct-composition/src/main_windows.rs
@@ -186,7 +186,11 @@ impl api::RenderNotifier for Notifier {
         let _ = self.events_proxy.wakeup();
     }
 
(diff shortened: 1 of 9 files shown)


nical:
Looks good, I think we should add blob rasterization time. It would be collected in the same place as the scene building time (unless we want to group blob rasterization and scene building under the same measurement).
I can add it in the async blob PR but since it will require the same mechanical changes to the public API, we might as well do it here.

staktrace:
I forgot to mention: the gecko-side changes that go with this are in https://bugzilla.mozilla.org/show_bug.cgi?id=1470901

PR #2850, opened 2018-06-27
Forbid segmentation of tiled images

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1470855
r? @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2850)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -28,7 +28,7 @@ use render_backend::FrameId;
 use render_task::{BlitSource, RenderTask, RenderTaskCacheKey};
 use render_task::{RenderTaskCacheKeyKind, RenderTaskId, RenderTaskCacheEntryHandle};
 use renderer::{MAX_VERTEX_TEXTURE_WIDTH};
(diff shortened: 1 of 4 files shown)


gw3583:
Could we store (if we don't already) whether tiling is enabled inside ```BrushKind::Image``` I wonder? That would mean we don't need to pass resource cache, and avoid a resource cache hash lookup each time we call supports_segments? In general, looks good though!

kvark:
I think it would be semantically nicer to have `visible_tiles: Option<Vec<...>>` in there, as you suggest. I'll give it a try tomorrow.

kvark:
@gw3583 actually, we don't have easy access to the tiling information during flattening, where `visible_tiles` is first initialized, so I don't think this is a good path. I think we should just merge the PR as is. cc @nical

gw3583:
@kvark OK, sounds good. Is it possible to add a reftest for this?

kvark:
Added now. I also confirmed that it fails with the previous revision.

PR #2851, opened 2018-06-27
Fix a small typo in the wrench window title

mrobinson:
This is almost inconsequential, but always makes me do a double-take.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2851)


staktrace:
r+

PR #2853, opened 2018-06-27
Fix crashtest panic in ResourceCache::request_image.

aosmond:
ResourceCache::update_texture_cache checks if the request has a tile
associated with it, and if so, adjusts the descriptor and dirty rect for
the individual texture updates. If the image should have been tiled, but
the request did not provide such information, it still attempts to
proceed without tiling.

ResourceCache::request_image was changed to verify that the dirty rect
for the associated image intersects with the request, to avoid
requesting updates for tiled requests which are unchanged. Before it
only did this check for blob images, but it is now necessary for
external images as well which can provide a dirty rect.

We assume here that all requests for an image that must be tiled will have
the necessary tiling information provided, but that does not appear to be
the case. Clipped and YUV images do not provide tiling information as an
example. This patch reverses the check ordering to be consistent with
update_texture_cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2853)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -652,9 +652,9 @@ impl ResourceCache {
         //  - The image is a blob.
         //  - The blob hasn't already been requested this frame.
         if template.data.is_blob() || dirty_rect.is_some() {
-            let (offset, size) = match template.tiling {
-                Some(tile_size) => {
-                    let tile_offset = request.tile.unwrap();
+            let (offset, size) = match request.tile {
+                Some(tile_offset) => {
+                    let tile_size = template.tiling.unwrap();
                     let actual_size = compute_tile_size(
                         &template.descriptor,
                         tile_size,


aosmond:
Try with my original PR, without this fix: https://treeherder.mozilla.org/#/jobs?repo=try&revision=62bb98e3f84f903cfc5e743690f9cac858a5eb5e&selectedJob=185141784

Try with this fix: https://treeherder.mozilla.org/#/jobs?repo=try&revision=ad952a82cf8fa5d8ba8f8a865c6d745ca759d653
https://treeherder.mozilla.org/#/jobs?repo=try&revision=db09a75fb3917f5d051589c4c1e7db4e12a988b1

kvark:
I believe this is wrong, but given that it's patching a PR that was also not correct in the first place (and which we don't want to back out), we should go ahead to fix the crash so that we can clean up the mess afterwards (discussed the proper solution on a call today with @aosmond, @nical and @staktrace ).
@bors-servo r+

PR #2854, opened 2018-06-27
Allow scrolling up and down in Wrench

mrobinson:
This makes it easier to debug scrolling issues by allowing interactive
scrolling when looking at Wrench renders.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2854)


mrobinson:
I have found this to be essential in debugging https://bugzilla.mozilla.org/show_bug.cgi?id=1446626.

PR #2855, opened 2018-06-27
configurable angle for synthetic italics

lsalzman:
This fixes Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1460259

Gecko no longer uses a fixed angle for synthetic italics, so the old strategy of just trying to use fixed per-platform values that match Gecko also no longer works.

We thus need to be able to pass in a per-font angle.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2855)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -1832,6 +1832,7 @@ impl<'a> DisplayListFlattener<'a> {
                 font_instance.bg_color,
                 render_mode,
                 flags,
+                font_instance.synthetic_italics,
                 font_instance.platform_options,
                 font_instance.variations.clone(),
             );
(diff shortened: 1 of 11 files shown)


gw3583:
Do we need a try run for this?

gw3583:
Looks good to me. Is it possible to add a reftest for this?

lsalzman:
Already did a try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=587c4a551808f98a18599dd893eea415aaa0c1d2

The only thing that went wrong there is the R2 which was marked as a perma-fail, but now passes because of this change. Just needs to be fixed in the Gecko reftest.list.

lsalzman:
Added reftests

gw3583:
Looks like one of the ref yaml files is missing? I'll delegate anyway so feel free to r+ when that file is added, or now if I just missed it.

@bors-servo delegate+

PR #2856, opened 2018-06-27
Apply dirty rect updates consistently.

aosmond:
This patch moves the dirty rect storage from ImageResource to
CachedImageInfo. This is because cache entries may be updated
independently across frame updates, and thus cannot share the dirty rect
state.

We no longer clear the dirty rect when we get an update without a dirty
rect. This is because we can get a subsequent update (before applying
the previous) which provides a new dirty rect, and we still want to
reupload the whole image. Instead we now store a dirty rect for the
whole descriptor.

In addition to special handling for tiled requests, we also properly
handle requests which change the ImageRendering type. Before we would
take the dirty rect for the first request and ignore it for the rest.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2856)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -32,8 +32,8 @@ use render_backend::FrameId;
 use render_task::{RenderTaskCache, RenderTaskCacheKey, RenderTaskId};
 use render_task::{RenderTaskCacheEntry, RenderTaskCacheEntryHandle, RenderTaskTree};
 use std::collections::hash_map::Entry::{self, Occupied, Vacant};
-use std::collections::hash_map::IterMut;
-use std::cmp;
+use std::collections::hash_map::ValuesMut;
+use std::{cmp, mem};
 use std::fmt::Debug;
 use std::hash::Hash;
 #[cfg(any(feature = "capture", feature = "replay"))]
@@ -169,21 +169,21 @@ pub fn intersect_for_tile(
 }
 
 fn merge_dirty_rect(
-    prev_dirty_rect: Option<DeviceUintRect>,
-    dirty_rect: Option<DeviceUintRect>,
-    descriptor: ImageDescriptor,
+    prev_dirty_rect: &Option<DeviceUintRect>,
+    dirty_rect: &Option<DeviceUintRect>,
+    descriptor: &ImageDescriptor,
 ) -> Option<DeviceUintRect> {
     // It is important to never assume an empty dirty rect implies a full reupload here,
-    // although we are able to do so elsewhere. We use store the descriptor's full rect
-    // instead. There are update sequences which could cause us to forget the correct
-    // dirty regions if we cleared the dirty rect when we received None, e.g.:
+    // although we are able to do so elsewhere. We store the descriptor's full rect instead
+    // There are update sequences which could cause us to forget the correct dirty regions
+    // regions if we cleared the dirty rect when we received None, e.g.:
     //      1) Update with no dirty rect. We want to reupload everything.
     //      2) Update with dirty rect B. We still want to reupload everything, not just B.
     //      3) Perform the upload some time later.
     match (dirty_rect, prev_dirty_rect) {
-        (Some(rect), Some(prev_rect)) => Some(rect.union(&prev_rect)),
-        (Some(rect), None) => Some(rect),
-        (None, _) => Some(descriptor.full_rect()),
+        (&Some(ref rect), &Some(ref prev_rect)) => Some(rect.union(&prev_rect)),
+        (&Some(ref rect), &None) => Some(*rect),
+        (&None, _) => Some(descriptor.full_rect()),
     }
 }
 
@@ -208,6 +208,10 @@ where
         self.resources.insert(key, value);
     }
 
+    pub fn remove(&mut self, key: &K) {
+        self.resources.remove(key);
+    }
+
     pub fn get_mut(&mut self, key: &K) -> &mut V {
         self.resources.get_mut(key)
             .expect("Didn't find a cached resource with that ID!")
@@ -221,8 +225,8 @@ where
         self.resources.entry(key)
     }
 
-    pub fn iter_mut(&mut self) -> IterMut<K, V> {
-        self.resources.iter_mut()
+    pub fn values_mut(&mut self) -> ValuesMut<K, V> {
+        self.resources.values_mut()
     }
 
     pub fn clear(&mut self) {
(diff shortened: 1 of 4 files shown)


aosmond:
Works well from my local testing with the external image dirty rect changes. Let's make sure it passes the tests standalone too:

try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2144458335b7e456e59609e1206e95e9284524bc

aosmond:
Let's try this again: https://treeherder.mozilla.org/#/jobs?repo=try&revision=175ab1e476a37b5e1524f854bc9335ec9d9a9004

aosmond:
@kvark @nical

webrender/src/resource_cache.rs
```diff
@@ -546,16 +553,53 @@ impl ResourceCache {
...
+        } else {
+            for &render in &[ImageRendering::Auto, ImageRendering::CrispEdges, ImageRendering::Pixelated] {
+                let request = ImageRequest {
+                    key: image_key,
+                    rendering: render,
+                    tile: None
+                };
+
+                if let Some(&mut Ok(ref mut entry)) = self.cached_images.try_get_mut(&request) {
+                    let merged_dirty_rect = match (dirty_rect, entry.dirty_rect) {
```
kvark: we should not duplicate this `match` logic between the condition arms

webrender/src/resource_cache.rs
```diff
@@ -546,16 +553,53 @@ impl ResourceCache {
...
+        // although we are able to do so elsewhere. We must store the descriptor's full rect
+        // instead. This is because there are update sequences which could cause us to forget the
+        // correct dirty regions if we cleared the dirty rect when we received None, e.g.:
+        //
+        //      1) Update with no dirty rect. We want to reupload everything.
+        //      2) Update with dirty rect B. We still want to reupload everything, not just B.
+        //      3) Perform the upload some time later.
+        if tiling.is_some() {
+            for (request, entry) in self.cached_images.iter_mut() {
+                if request.tile.is_some() && request.key == image_key && entry.is_ok() {
```
kvark: let's do `if let Ok(ref mut entry) = entry { .. }` here to avoid the following `as_ref().unwrap()` and `as_mut().unwrap()`

webrender/src/resource_cache.rs
```diff
@@ -546,16 +553,53 @@ impl ResourceCache {
...
+        // It is important to never assume an empty dirty rect implies a full reupload here,
+        // although we are able to do so elsewhere. We must store the descriptor's full rect
+        // instead. This is because there are update sequences which could cause us to forget the
+        // correct dirty regions if we cleared the dirty rect when we received None, e.g.:
+        //
+        //      1) Update with no dirty rect. We want to reupload everything.
+        //      2) Update with dirty rect B. We still want to reupload everything, not just B.
+        //      3) Perform the upload some time later.
+        if tiling.is_some() {
+            for (request, entry) in self.cached_images.iter_mut() {
```
kvark: Let's leave a warning here that this might need optimization to avoid the O(N * M) complexity (where N is the number of updates, and M is the number of cache entries).

aosmond:
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=789dee0057e595c18d40c41e4861042eed87b10e&selectedJob=186456340

aosmond:
@kvark r?

webrender/src/resource_cache.rs
```diff
@@ -168,6 +166,24 @@ pub fn intersect_for_tile(
     })
 }
 
+fn merge_dirty_rect(
+    prev_dirty_rect: Option<DeviceUintRect>,
+    dirty_rect: Option<DeviceUintRect>,
+    descriptor: ImageDescriptor,
+) -> Option<DeviceUintRect> {
+    // It is important to never assume an empty dirty rect implies a full reupload here,
+    // although we are able to do so elsewhere. We use store the descriptor's full rect
```
kvark: nit: "We use store"

webrender/src/resource_cache.rs
```diff
@@ -168,6 +166,24 @@ pub fn intersect_for_tile(
     })
 }
 
+fn merge_dirty_rect(
+    prev_dirty_rect: Option<DeviceUintRect>,
+    dirty_rect: Option<DeviceUintRect>,
+    descriptor: ImageDescriptor,
```
kvark: let's pass it by reference here

webrender/src/resource_cache.rs
```diff
@@ -548,24 +601,33 @@ impl ResourceCache {
                 .update(image_key, Arc::clone(blob), dirty_rect);
         }
 
+        // Each cache entry stores its own copy of the image's dirty rect. This allows them to be
+        // updated independently.
+        if let Some(entries) = self.cached_images.try_get_mut(&image_key) {
```
kvark: nit: can combine into a single
```rust
match self.cached_images.try_get_mut(..) {
  Some(ImageResult::...) => {..}
  _ => {..}
}
```

webrender/src/resource_cache.rs
```diff
@@ -606,41 +668,75 @@ impl ResourceCache {
...
-                (info, needs_update)
+        let storage = match self.cached_images.entry(request.key) {
+            Occupied(e) => {
+                // We might have an existing untiled entry, and need to insert
+                // a second entry. In such cases we need to move the old entry
+                // out first, replacing it with a dummy entry, and then creating
+                // the tiled/multi-entry variant.
+                let entry = e.into_mut();
+                if !request.is_untiled_auto() {
+                    let untiled_entry = if let &mut ImageResult::UntiledAuto(ref mut entry) = entry {
```
kvark: nit: let's just `match`

webrender/src/resource_cache.rs
```diff
@@ -606,41 +668,75 @@ impl ResourceCache {
...
-        // in the cache matches that of the template, then it is
-        // valid to use as-is.
-        let (entry, needs_update) = match self.cached_images.entry(request) {
-            Occupied(entry) => {
-                let info = entry.into_mut();
-                let needs_update = info.as_mut().unwrap().epoch != template.epoch;
-                info.as_mut().unwrap().epoch = template.epoch;
-                (info, needs_update)
+        let storage = match self.cached_images.entry(request.key) {
+            Occupied(e) => {
```
kvark: perhaps, it would be cleaner to lift the condition up here:
```rust
Occupied(e) if e.is_untiled_auto() => {
...
}
Occupied(e) => {
...
}
Vacant(e) => {..}
```

webrender/src/resource_cache.rs
```diff
@@ -666,6 +762,7 @@ impl ResourceCache {
                     if let Some(dirty) = dirty_rect {
                         if intersect_for_tile(dirty, actual_size, tile_size, tile_offset).is_none() {
                             // don't bother requesting unchanged tiles
+                            entry.dirty_rect.take();
```
kvark: would be cleaner to just do `= None` here

wrench/src/blob.rs
```diff
@@ -55,7 +55,7 @@ fn render_blob(
     };
 
     let mut dirty_rect = dirty_rect.unwrap_or(DeviceUintRect::new(
-        DeviceUintPoint::origin(),
+        DeviceUintPoint::new(descriptor.offset.x as u32, descriptor.offset.y as u32),
```
kvark: let's just do `descriptor.offset.to_u32()` here

aosmond:
@kvark incorporated your feedback

kvark:
Sweet, :shipit: !
@bors-servo r+

PR #2857, opened 2018-06-28
Reduce size of scroll nodes texture, introduce transform palette.

gw3583:
Previously, we created a transform structure for every clip-scroll
node in the tree, and uploaded these to the GPU each frame. This
is a bit wasteful, because in a typical scene most CST nodes are
actually clip nodes, which don't affect position transforms.

We now maintain a TransformIndex in the CST. This is a unique
identifier for spatial nodes. For clip nodes, it is a reference
to the spatial node for this clip node.

A new type, the TransformPalette is introduced. This creates a
transform entry only for spatial nodes, which reduces the size
of the transform texture upload significantly.

Additionaly, we store the is_axis_aligned flag in the transform
node ID itself, to reduce the number of texel fetches that the
vertex shader does to fetch a transform.

The change described above is a decent optimization, but that is
mostly a side effect. The actual goal of this patch is to
introduce the functionality we need to be able to rasterize
a Picture in a coordinate space relative to a given parent. That
functionality isn't part of this patch, but the TransformPalette
is the basis of how this will work, allowing us to retrive a
transform for a given TransformIndex, but relative to a given
CoordinateSystemId.

The benchmarks below show the before / after size of the transform
texture that is being uploaded per frame:

nytimes.com | 19.9 kB => 0.9 kB
en.wikipedia.org | 14.6 kB => 0.9 kB
reddit.com | 30.7 kB => 4 kB
cnn.com | 11.7 kB => 1 kB
twitter.com/servodev | 26.2 kB => 3.3 kB

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2857)


gw3583:
r? @mrobinson and / or @kvark 

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9b5af357909fa58aa302a11991b3865bbd00b53c

PR #2859, opened 2018-06-28
Don't segment using clips with different positioning nodes

mrobinson:
When segmenting brushes, don't consider clips if they have a different
positioning node. The issue here is that since they have a different
positioning node, the segmentation results may change during scrolling.
This means that we would need to invalidate the cache result and
re-segment. We can do that in the future, but in the meantime this fixes
the rendering issues the bug causes.

Unfortunately, this bug is triggered by rendering, scrolling, and then
re-rendering, so it is tricky to make a test for it.

Original bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1446626

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2859)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -257,19 +257,39 @@ impl ClipSource {
             }
         }
     }
+
+    pub fn is_rect(&self) -> bool {
+        match *self {
+            ClipSource::Rectangle(..) => true,
+            _ => false,
+        }
+    }
+
+    pub fn is_image_or_line_decoration_clip(&self) -> bool {
+        match *self {
+            ClipSource::Image(..) | ClipSource::LineDecoration(..) => true,
+            _ => false,
+        }
+    }
 }
 
 #[derive(Debug)]
 pub struct ClipSources {
     pub clips: Vec<(ClipSource, GpuCacheHandle)>,
     pub local_inner_rect: LayoutRect,
-    pub local_outer_rect: Option<LayoutRect>
+    pub local_outer_rect: Option<LayoutRect>,
+    pub only_rectangular_clips: bool,
+    pub has_image_or_line_decoration_clip: bool,
 }
 
 impl ClipSources {
     pub fn new(clips: Vec<ClipSource>) -> Self {
         let (local_inner_rect, local_outer_rect) = Self::calculate_inner_and_outer_rects(&clips);
 
+        let has_image_or_line_decoration_clip =
+            clips.iter().any(|clip| clip.is_image_or_line_decoration_clip());
+        let only_rectangular_clips =
+            !has_image_or_line_decoration_clip && clips.iter().all(|clip| clip.is_rect());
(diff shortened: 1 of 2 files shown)


gw3583:
Ah, this will need to be rebased after the transform node PR landed.

```
    --> src\prim_store.rs:2046:26
     |
2046 |             if clip_item.scroll_node_data_index != prim_run_context.scroll_node.node_data_index {
     |                          ^^^^^^^^^^^^^^^^^^^^^^
error[E0609]: no field `node_data_index` on type `&clip_scroll_node::ClipScrollNode`
    --> src\prim_store.rs:2046:81
     |
2046 |             if clip_item.scroll_node_data_index != prim_run_context.scroll_node.node_data_index {
     |                                                                                 ^^^^^^^^^^^^^^^
```

mrobinson:
@gw3583 This should be rebased and ready to go now.

staktrace:
Looks like there's a couple of failing reftests?

mrobinson:
@staktrace Yeah, unfortunately this causes a few regressions. I'm looking at those now and hoping to have a new branch up soon.

kvark:
@mrobinson could you make a try push?

mrobinson:
Gecko try push pending here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=45c352f270594c5b706de7916af8df1707455c74

mrobinson:
Looks like there is one reftest failure. I'll try to fix that tomorrow and post a new version of this PR.

mrobinson:
Update: This one is a bit of a mystery. I have tried everything I can think of to reproduce the failure locally, but I haven't figured out to yet. I've confirmed that the part of my change that is causing the failure is the early continue and not the new calculation of `local_rects_only`. I just need to figure out why now.

mrobinson:
Sorry about the lack of movement on this. I have some cycles now to look at the mysterious test failure.

gw3583:
@mrobinson Is this ready for review now? Do we need another gecko try push or is the one above sufficient?

mrobinson:
@gw3583 This works, but I wanted to try to write a version that doesn't always use a Global clip mask today.

mrobinson:
@gw3583 Okay. This should be ready to review now. I'm particularly worried about the situations where we now need to fall back to a global mask. This happens for items which are clipped by non-rectangular clips from other position nodes. For instance, if a scroll area is surrounded by a rectangular clip, we will need to generate a rectangular mask for items that intersect that clip.

kvark:
What's up with all the failures in "Windows 10 x64 QuantumRender debug"?

staktrace:
You can ignore those failures. The mochitests on windows debug only run on try pushes but not on m-c or other branches. Annoyingly there's no easy way to turn it off in the try pushes.

kvark:
Thanks @staktrace 
@bors-servo r+

PR #2860, opened 2018-06-29
Add a simple benchmark for overdraw due to many text shadows.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2860)

### DIFF
--- a/wrench/benchmarks/overlapping-text-shadows.yaml
+++ b/wrench/benchmarks/overlapping-text-shadows.yaml
@@ -1,0 +1,807 @@
+root:
+  items:
+    -
+      type: "shadow"
+      offset: [0, 0]
+      color: red
+    -
+      type: "shadow"
+      offset: [1, 1]
+      color: red
+    -
+      type: "shadow"
+      offset: [2, 2]
+      color: red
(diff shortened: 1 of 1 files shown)


gw3583:
FWIW, I did a quick hack / prototype of using a z-pass on text runs on this benchmark. On my HD4600, the results are:

Normal:
 GPU time: 7.4ms
 Alpha overdraw: 3.81x

With z-pass:
 GPU time: 3.2ms
 Alpha overdraw: 0.78x

So this seems like quite a promising technique :tada:

gw3583:
Related to bug https://bugzilla.mozilla.org/show_bug.cgi?id=1463318, it brings the GPU time on a 4k screen + HD4600 down from ~50ms to ~20ms. Which is a huge win, but we will probably still want to cache that in a picture to maintain 60fps on low end hardware.

kvark:
> So this seems like quite a promising technique

Indeed :) I'm more concerned about the cost in the bad/normal case, and our ability to come up with a good heuristic on when to enable it.

Anyhow,
@bors-servo r+

PR #2865, opened 2018-07-04
Store the built state in the PipelineInfo

staktrace:
This adds a flag to the entries in the PipelineInfo that records whether
or not a pipeline was actually built at a given epoch. It is possible
for pipelines to have display lists that are not built if they are not
connected to the root pipeline via a chain of iframe items, and in some
cases Gecko needs to be able to distinguish between whether a pipeline
was built or not during a particular render.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2865)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -5,7 +5,7 @@
 
 use api::{AlphaType, BorderDetails, BorderDisplayItem, BuiltDisplayListIter, ClipAndScrollInfo};
 use api::{ClipId, ColorF, ComplexClipRegion, DeviceIntPoint, DeviceIntRect, DeviceIntSize};
-use api::{DevicePixelScale, DeviceUintRect, DisplayItemRef, Epoch, ExtendMode, ExternalScrollId};
+use api::{DevicePixelScale, DeviceUintRect, DisplayItemRef, ExtendMode, ExternalScrollId};
 use api::{FilterOp, FontInstanceKey, GlyphInstance, GlyphOptions, GlyphRasterSpace, GradientStop};
 use api::{IframeDisplayItem, ImageKey, ImageRendering, ItemRange, LayoutPoint};
 use api::{LayoutPrimitiveInfo, LayoutRect, LayoutSize, LayoutTransform, LayoutVector2D};
@@ -150,9 +150,6 @@ pub struct DisplayListFlattener<'a> {
     /// The map of all font instances.
     font_instances: FontInstanceMap,
 
-    /// Used to track the latest flattened epoch for each pipeline.
-    pipeline_epochs: Vec<(PipelineId, Epoch)>,
-
     /// A set of pipelines that the caller has requested be made available as
     /// output textures.
     output_pipelines: &'a FastHashSet<PipelineId>,
@@ -207,8 +204,6 @@ impl<'a> DisplayListFlattener<'a> {
         let root_pipeline_id = scene.root_pipeline_id.unwrap();
         let root_pipeline = scene.pipelines.get(&root_pipeline_id).unwrap();
 
-        let root_epoch = scene.pipeline_epochs[&root_pipeline_id];
-
         let background_color = root_pipeline
             .background_color
             .and_then(|color| if color.a > 0.0 { Some(color) } else { None });
@@ -218,7 +213,6 @@ impl<'a> DisplayListFlattener<'a> {
             clip_scroll_tree,
             font_instances,
             config: *frame_builder_config,
-            pipeline_epochs: Vec::new(),
             output_pipelines,
             id_to_index_mapper: ClipIdToIndexMapper::default(),
             hit_testing_runs: recycle_vec(old_builder.hit_testing_runs),
@@ -243,8 +237,7 @@ impl<'a> DisplayListFlattener<'a> {
         debug_assert!(flattener.picture_stack.is_empty());
 
         new_scene.root_pipeline_id = Some(root_pipeline_id);
-        new_scene.pipeline_epochs.insert(root_pipeline_id, root_epoch);
-        new_scene.pipeline_epochs.extend(flattener.pipeline_epochs.drain(..));
+        new_scene.pipeline_epochs = scene.pipeline_epochs.clone();
         new_scene.pipelines = scene.pipelines.clone();
 
         FrameBuilder::with_display_list_flattener(
@@ -528,9 +521,6 @@ impl<'a> DisplayListFlattener<'a> {
                 reference_frame_relative_offset
             ),
         );
-
-        let epoch = self.scene.pipeline_epochs[&iframe_pipeline_id];
-        self.pipeline_epochs.push((iframe_pipeline_id, epoch));
 
         let bounds = item.rect();
         let origin = *reference_frame_relative_offset + bounds.origin.to_vector();


staktrace:
This is desired for https://bugzilla.mozilla.org/show_bug.cgi?id=1473352

kvark:
@stacktrace The building is triggered, as you say, by having the pipeline connected to the root pipeline. This connection is defined by Gecko essentially, so it has all the info it needs to answer the question if a pipeline was actually built.

Can you confirm that this is just needed because it's more convenient to track this on our side?

staktrace:
Agreed - this is mostly for convenience because we can have multiple scene build requests in flight at the same time and it to track it on the gecko side means we'd have to duplicate a bunch of stuff there. It seemed easier/more reliable to do it this way.

staktrace:
(Also it's a little complicated on the Gecko side because the knowledge of whether the content display list was attached is buried deep in the display list creation, and extracting that and associating it with data from the content process is messy)

webrender/src/renderer.rs
```diff
@@ -1833,7 +1833,10 @@ impl Renderer {
 
     /// Returns the Epoch of the current frame in a pipeline.
     pub fn current_epoch(&self, pipeline_id: PipelineId) -> Option<Epoch> {
-        self.pipeline_info.epochs.get(&pipeline_id).cloned()
+        self.pipeline_info.epochs
```
kvark: is the current epoch useful for non-built pipelines?

webrender/src/scene.rs
```diff
@@ -99,7 +99,7 @@ pub struct ScenePipeline {
 pub struct Scene {
     pub root_pipeline_id: Option<PipelineId>,
     pub pipelines: FastHashMap<PipelineId, Arc<ScenePipeline>>,
-    pub pipeline_epochs: FastHashMap<PipelineId, Epoch>,
+    pub pipeline_epochs: FastHashMap<PipelineId, (Epoch, BuildState)>,
```
kvark: hmm, this seems misplaced. Scene is the source for building, it's not supposed to know about build artifacts (like this flag)

webrender_api/src/api.rs
```diff
@@ -645,6 +645,23 @@ impl fmt::Debug for ApiMsg {
     }
 }
 
+// Used in a PipelineInfo to indicate whether or not a pipeline at a particular
+// epoch was built or not. A pipeline may have not been built if it it was not
+// attached to the root pipeline via iframe items.
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+pub struct BuildState(pub bool);
```
kvark: Any reason this is not just an `enum`?

webrender/src/renderer.rs
staktrace: Yes, knowing the epoch lets us know which gecko transaction has been completed, and lets us fire the MozAfterPaint event on the gecko side, which is the desired goal of this change.

kvark: but if the pipeline isn't built, it's not painted either, so how does `MozAfterPaint` make sense?

webrender/src/scene.rs
staktrace: Ok. I should be able to take this out, will update the patch.

webrender_api/src/api.rs
staktrace: Not really. I can change it to an enum.

webrender/src/renderer.rs
staktrace: Because Gecko. You could consider it a bug, but in this case it's easier to bug-compatible with gecko than rewrite large chunks of the gecko frontend :(

mstange: Yeah it should really be called `MozAfterPrepareToComposite` or `MozReadyToCompositeInstantly`.

kvark: ok, @mstange , but it's not even prepared/ready if it's not built!

staktrace: Maybe MozCompositorHasSomeIdeaThatThisExists :)

webrender/src/scene.rs
staktrace: @kvark Would the frame builder be a better place for this? The flattener can construct the map with all the built and unbuilt entries, pass it to the frame builder, and the render_backend can read it from the frame builder in `updated_pipeline_info` instead of the from the scene.

kvark: sounds reasonable

webrender/src/renderer.rs
mstange: Actually, sorry, I was mixing up `MozAfterPaint` with `MozLayerTreeReady`. I concur with kats now; Gecko is being stupid and using `MozAfterPaint` in a confused way, but matching the bug is the easy way out.

Darkspirit:
[#TechnicalDebt](https://github.com/servo/webrender/search?q=TechnicalDebt&type=Issues)

staktrace:
This is turning out more complicated than I thought. Right now there is code in [update_epoch](https://github.com/servo/webrender/blob/311c20eaa156a4e0816e2d9db8469d3e30b67c25/webrender_api/src/api.rs#L96) that sends a `FrameMsg::UpdateEpoch`. Ordinarily, after scene building, the `new_scene` only has the built pipelines in its `pipeline_epochs` map, but this `FrameMsg::UpdateEpoch` means that pipelines that were *not* built but *were* updated also end up in the new scene's `pipeline_epochs` map. And so right now when we send the `PipelineInfo` to gecko, it already contains a mix of built and not-built pipelines, and apparently there are tests that rely on this behaviour.

This patch as-is doesn't preserve this behaviour correctly, although it's simple enough to modify the patch to do so. The problem is preserving this behaviour *without* putting the BuildState on the scene's copy of the `pipeline_epoch` map (i.e. the change requested in the review comment above). We need a way to distinguish between pipelines added via `scene::set_display_list` and `scene::update_epoch` because the former should be marked as `NotBuilt` and the latter should be marked as `Built`, but if we don't store the BuildState on the scene then we lose this information.

staktrace:
> We need a way to distinguish between pipelines added via scene::set_display_list and scene::update_epoch because the former should be marked as NotBuilt and the latter should be marked as Built, but if we don't store the BuildState on the scene then we lose this information.

I realized we can do this by checking if `scene.pipelines` has an entry for the pipeline id. Doing try pushes and stuff to verify. Although honestly it doesn't feel like the best solution so I'll let it simmer a bit more. It might be possible to remove the Built vs NotBuilt distinction entirely.

kvark:
Interesting! What if we store the pipeline epoch in the pipeline itself? This way, we can trivially find out if the pipeline was added with a display list by comparing it's epoch with the map key.

staktrace:
So yeah, I don't think we need the distinction between "built" and "not built" pipelines on the gecko side. As long as all the pipeline epochs get propagated from the pending scene to the new scene things are good. Here's a try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=31c488c1216646ac325e71f2cd527ea8debd79c6

The WR patch is much simpler, as it is mostly just code deletion now. I don't know if servo relies on the pipeline info only containing epochs for built scenes or not, that's the only potential concern. But really the code as it is now is not really providing a easily-describable set of epochs in the pipeline info because of this `update_epoch` business so if servo is relying on it they're probably going to have the same issues gecko does, and this patch should help.

kvark:
Haven't seen such a clean try push in a while :)
I guess the intermittents are getting nailed (perhaps, even thanks to Pernosco?) faster than we introduce them.

Looking at the Servo code briefly, I'm not seeing them using the epochs.
@bors-servo r+

staktrace:
Trivial merge error; I rebased the patch and took out the now-unneeded "use Epoch". ^

PR #2866, opened 2018-07-04
Move processing of frame messages to before call to build_scene

### LINKED ISSUE #2820: First call to set_output_document has no effect
The frame output example stopped working some time ago. 

The issue seems to be that frame messages (which contain `EnableFrameOutput` are processed after the call to `build_scene` resulting in an empty list of `output_pipelines` getting passed to the `FrameBuilder`.

Moving the processing of frame messages to before the call to `build_scene` fixes the issue. There is a separate issue where nothing is rendered in the example.

hgallagher1993:
Fixes #2820

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2866)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -1015,15 +1015,6 @@ impl RenderBackend {
             &mut profile_counters.resources,
         );
 
-        if op.build {
-            let scene_id = self.make_unique_scene_id();
-            let doc = self.documents.get_mut(&document_id).unwrap();
-            let _timer = profile_counters.total_time.timer();
-            profile_scope!("build scene");
-
-            doc.build_scene(&mut self.resource_cache, scene_id);
-        }
-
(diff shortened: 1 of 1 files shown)


staktrace:
This is going to break APZ in gecko, because the sampler callback will produce incorrect values

staktrace:
#2820 could also be fixed by moving the enable_frame_output into a separate transaction that is submitted before the transaction that has the display list.

hgallagher1993:
:+1: Just an FYI as well I'll be gone from this evening until probably Sunday so it'll be a good few days before I get back to this

nc4rrillo:
So if we can just move `enable_frame_output` to a separate transaction it may be enough to just update the frame output sample.

hgallagher1993:
Sorry about the delay with this, my home directory got wiped last week so I've just been busy trying to restore everything (like all the opensource projects I had downloaded) so I'll try and get this finished in the next day or 2

PR #2870, opened 2018-07-06
update rust 1.27.0

ekicyou:
separate from #2838
Some modules do not pass builds with rust on ci, such as winit = "0.16".

**Update https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox,**
Are you being told to skip the old version and use Rust 1.28.0?
If there is a problem cancel PR.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2870)


kvark:
Thank you for the PR!
It appears to be blocked currently on the bugzilla bug posted by @SimonSapin .
Please make sure to rebase the PR properly by the time it's ready for merging (having 3 commits, 2 of which are merge commits, is not desired).

ekicyou:
Thanks!

I also examined how to switch to 1.28-beta, but I thought it better to wait for the 1.28 stable version rather than being simple.

There is no reason to hurry, so there are problems with rebasing, so I'd like to close the PR and see the situation.

atouchet:
The Bugzilla bug is now closed.

PR #2871, opened 2018-07-06
Split the ClipScrollTree

mrobinson:
Split the ClipScrollTree into a collection of spatial nodes and a
collection of clipping nodes. This makes it so that no unnecessary work
is done or storage is used for clipping nodes. It also allows us to
simplify certain bits of code that had to deal with both types of nodes.

During display list "flattening," positioning ClipIds are converted to
SpatialNodeIndices by consulting the ClipNode for their positioning
node. ClipIds were already converted directly into ClipChainIndices.

Eventually this change should make it to the API, meaning that we expose
the difference between the different node types. This will make it
harder to misuse the API and allow us to do away with looking up the
positioning node of clips during flattening.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2871)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -475,8 +475,8 @@ impl AlphaBatchBuilder {
 
         // Add each run in this picture to the batch.
         for run in &pic.runs {
-            let scroll_node = &ctx.clip_scroll_tree.nodes[run.clip_and_scroll.scroll_node_id.0];
-            let transform_id = ctx.transforms.get_id(scroll_node.transform_index);
+            let transform_id =
+                ctx.transforms.get_id(run.clip_and_scroll.scroll_node_id.transform_index());
             self.add_run_to_batch(
                 run,
                 transform_id,
@@ -670,7 +670,7 @@ impl AlphaBatchBuilder {
                             debug_assert!(picture.surface.is_some());
 
                             let real_xf = &ctx.clip_scroll_tree
-                                .nodes[picture.reference_frame_index.0]
+                                .spatial_nodes[picture.reference_frame_index.0]
                                 .world_content_transform
                                 .into();
                             let polygon = make_polygon(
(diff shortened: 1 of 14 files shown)


mrobinson:
A few notes on this change:
 1. I have used the "spatial" terminology that @gw3583 used previously. I'm happy to use something else if it's more appropriate.
 2. This change seems to have exposed a pre-existing quirk where some clip indices don't make it into the display list in Gecko. This is a bit of a mystery to me and I'm still investigating. For now, the code just skips over empty ClipNodes.
 3. Gecko try job is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=33639cd81873baf1c23e6e061900ea3433e4fec1

gw3583:
Looks good to me from a quick look!

@bors-servo r=kvark,gw3583

johnp:
Given Bugzilla is down right now: There's a new webrender [crash](https://crash-stats.mozilla.com/report/index/e9ad7ae5-ef7e-4660-bb3a-c069b0180714) on Nightly (100% reproducible on my profile) and looking at the stuff that landed with the latest webrender update on Nightly this PR might have caused it. I'll file a proper bug tomorrow (europe) if no one beats me to it.

staktrace:
@johnp: Could you check the try build from https://treeherder.mozilla.org/#/jobs?repo=try&revision=721ea8ff7f60f1eae6aee01a5571b40777b6b2f5 on your profile/session to see if that also crashes? The try build has the fix from #2899 . Thanks!

Darkspirit:
([Linux x64 opt > B > Job Details > target.tar.bz2](https://queue.taskcluster.net/v1/task/caEa_JcJRwe5Gx76IOf02g/runs/0/artifacts/public/build/target.tar.bz2))

staktrace:
@johnp Do you think you'll have a chance to verify the fix soon? We don't have any STR for the crash and it was fairly high volume so it would be great to get confirmation that it was fixed before we push the next WR update into firefox.

PR #2873, opened 2018-07-06
Update angle shader validation features and add a missing newline

### LINKED ISSUE #2872: Missing new line from feature strings in angle_shader_validation
There is a missing new line when creating feature definitions for shaders in the angle shader validation: 
https://github.com/servo/webrender/blob/311c20eaa156a4e0816e2d9db8469d3e30b67c25/webrender/tests/angle_shader_validation.rs#L111

This should be 
```rust
features.push_str(&format!("#define WR_FEATURE_{}\n", feature));
```
However adding the missing new line here brings out an error, because the validation fails at the `brush_yuv_image` shader with the `"YUV_NV12,TEXTURE_RECT"` configuration:
```
thread 'validate_shaders' panicked at 'Shader compilation failed: brush_yuv_image
ERROR: 0:172: 'GL_ARB_texture_rectangle' : extension is not supported
ERROR: 0:172: 'sampler2DRect' : unsupported type
ERROR: 0:173: 'GL_ARB_texture_rectangle' : extension is not supported
ERROR: 0:173: 'sampler2DRect' : unsupported type
ERROR: 0:174: 'GL_ARB_texture_rectangle' : extension is not supported
ERROR: 0:174: 'sampler2DRect' : unsupported type
```
Maybe the reason this error didn't show before, that the feature definitions were in line, and just the first one get processed? Because if I change the order of those features ("TEXTURE_RECT,YUV_NV12") this error also occurs (without the new line).

I could send a PR to fix this with some other small things in the `angle_shader_validtaiom.rs` (e.g. the `TRANSFORM` [variant](https://github.com/servo/webrender/blob/master/webrender/tests/angle_shader_validation.rs#L23) of primitive shaders doesn't exist anmyore).

zakorgy:
This adds the missing new line to the end of the the feature strings, which I described in https://github.com/servo/webrender/issues/2872.
Also updated the features which the angle validator uses for testing, since `TRANSFORM` is no longer a Primitive shader feature and Brush shaders have their own feature set too.

Fixes #2872

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2873)

### DIFF
--- a/webrender/tests/angle_shader_validation.rs
+++ b/webrender/tests/angle_shader_validation.rs
@@ -18,9 +18,11 @@ struct Shader {
 
 const SHADER_PREFIX: &str = "#define WR_MAX_VERTEX_TEXTURE_WIDTH 1024\n";
 
+const BRUSH_FEATURES: &[&str] = &["", "ALPHA_PASS"];
 const CLIP_FEATURES: &[&str] = &["TRANSFORM"];
 const CACHE_FEATURES: &[&str] = &[""];
-const PRIM_FEATURES: &[&str] = &["", "TRANSFORM"];
+const GRADIENT_FEATURES: &[&str] = &[ "", "DITHERING", "ALPHA_PASS", "DITHERING,ALPHA_PASS" ];
+const PRIM_FEATURES: &[&str] = &[""];
 
 const SHADERS: &[Shader] = &[
     // Clip mask shaders
@@ -43,7 +45,7 @@ const SHADERS: &[Shader] = &[
     // Cache shaders
     Shader {
         name: "cs_blur",
-        features: CACHE_FEATURES,
+        features: &[ "ALPHA_TARGET", "COLOR_TARGET" ],
     },
     Shader {
         name: "cs_border_segment",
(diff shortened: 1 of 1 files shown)


kvark:
This is blocked until we figure out the exact situation happening in #2872

webrender/tests/angle_shader_validation.rs
```diff
@@ -43,7 +44,7 @@ const SHADERS: &[Shader] = &[
     // Cache shaders
     Shader {
         name: "cs_blur",
-        features: CACHE_FEATURES,
+        features: &[ "ALPHA_TARGET", "COLOR_TARGET" ],
```
kvark: don't we need the same for `cs_border_segment`?

webrender/tests/angle_shader_validation.rs
```diff
@@ -56,36 +57,43 @@ const SHADERS: &[Shader] = &[
...
     },
     Shader {
         name: "brush_mix_blend",
-        features: &[],
+        features: BRUSH_FEATURES,
     },
     Shader {
         name: "brush_radial_gradient",
-        features: &[ "DITHERING" ],
+        features: &[ "", "DITHERING", "ALPHA_PASS", "DITHERING,ALPHA_PASS" ],
```
kvark: let's have a `GRADIENT_FEATURES` define

webrender/tests/angle_shader_validation.rs
```diff
@@ -56,36 +57,43 @@ const SHADERS: &[Shader] = &[
...
     },
     Shader {
         name: "brush_linear_gradient",
-        features: &[],
+        features: &[ "", "DITHERING", "ALPHA_PASS", "DITHERING,ALPHA_PASS" ],
```
kvark: dithering wasn't used here before, why is this needed now?

kvark:
Strongly related to #2436

webrender/tests/angle_shader_validation.rs
zakorgy: No, because it is a LazilyCompiledShader and has no features at all: https://github.com/servo/webrender/blob/master/webrender/src/shade.rs#L629
These two are specified explicitly: 
https://github.com/servo/webrender/blob/master/webrender/src/shade.rs#L501
https://github.com/servo/webrender/blob/master/webrender/src/shade.rs#L509

webrender/tests/angle_shader_validation.rs
zakorgy: I just tought it is a good idea to test both gradient tests with all the possible features.

kvark: yeah, if the dithering is indeed used, we should test for it :) let's just have a common define for those feature sets

zakorgy:
@kvark Could you please take a look at this PR? I think, I have addressed all of your comments.

PR #2875, opened 2018-07-06
Inherit the clip chain of IFrame

### LINKED ISSUE #2834: Content area stuff is drawn over the chrome in Firefox
https://bugzilla.mozilla.org/show_bug.cgi?id=1458373

STR:
- Start firefox nightly with webrender
- Go to https://addons.mozilla.org/en-US/firefox/addon/steven-universe-sky/ and install (or just "tap to preview")
- Go to https://bugzilla.mozilla.org/attachment.cgi?id=8980322

kvark:
~~This is an experiment to fix #2834 on our side by inheriting the clip chains of the iFrames.~~
Fixes #2834
~~I really don't like the way it turns out... and the old code got me confused because the `info.clip_node_id` was completely ignored in `flatten_iframe`.~~
@mrobinson please take a look, cc @staktrace

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2875)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -190,6 +190,9 @@ pub struct DisplayListFlattener<'a> {
     /// A stack of the currently active shadows
     shadow_stack: Vec<(Shadow, PictureIndex)>,
 
+    /// The stack keeping track of the root clip chains associated with pipelines.
+    pipeline_clip_chain_stack: Vec<ClipChainIndex>,
+
     /// A list of scrollbar primitives.
     pub scrollbar_prims: Vec<ScrollbarPrimitive>,
 
@@ -240,6 +243,7 @@ impl<'a> DisplayListFlattener<'a> {
             picture_stack: Vec::new(),
             shadow_stack: Vec::new(),
             sc_stack: Vec::new(),
+            pipeline_clip_chain_stack: Vec::new(),
             prim_store: old_builder.prim_store.recycle(),
             clip_store: old_builder.clip_store.recycle(),
         };
@@ -530,7 +534,8 @@ impl<'a> DisplayListFlattener<'a> {
 
         self.id_to_index_mapper.initialize_for_pipeline(pipeline);
 
-        self.add_clip_node(
+        //TODO: use or assert on `clip_and_scroll_ids.clip_node_id` ?
+        let clip_chain_index = self.add_clip_node(
             info.clip_id,
             clip_and_scroll_ids.scroll_node_id,
             ClipRegion::create_for_clip_node_with_local_clip(
@@ -538,6 +543,7 @@ impl<'a> DisplayListFlattener<'a> {
                 reference_frame_relative_offset
             ),
         );
+        self.pipeline_clip_chain_stack.push(clip_chain_index);
 
         let bounds = item.rect();
         let origin = *reference_frame_relative_offset + bounds.origin.to_vector();
@@ -564,6 +570,7 @@ impl<'a> DisplayListFlattener<'a> {
         self.flatten_root(pipeline, &iframe_rect.size);
 
         self.pop_reference_frame();
+        self.pipeline_clip_chain_stack.pop();
     }
 
     fn flatten_item<'b>(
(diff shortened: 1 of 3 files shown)


mrobinson:
Hrm. Weird. It doesn't seem like this should be necessary. I left an explanation here: https://github.com/servo/webrender/issues/2834#issuecomment-403394224

mrobinson:
Apologies! I didn't understand what this PR did before, but now I do. I think this is one way that we can fix the issue. Tomorrow I'll take another look at this.

kvark:
I updated (or, rather reimplemented) the code on latest master.

webrender/src/display_list_flattener.rs
```diff
@@ -169,6 +169,9 @@ pub struct DisplayListFlattener<'a> {
     /// The map of all font instances.
     font_instances: FontInstanceMap,
 
+    /// The map keeping track of the root clip chains associated with pipelines.
+    pipeline_clip_chains: FastHashMap<PipelineId, ClipChainIndex>,
```
mrobinson: If you want to avoid the cost of hashing here, this can just be a stack. You would push ClipChainIndex(0) for the root pipeline and then push again when processing an iframe element.

webrender/src/display_list_flattener.rs
```diff
@@ -539,6 +543,17 @@ impl<'a> DisplayListFlattener<'a> {
             ),
         );
 
+        match clip_and_scroll_ids.clip_node_id {
```
mrobinson: I think what you want to do here is to simply use the newly added clip node. So

```rust
self.pipeline_clip_chains.insert(
    iframe_pipeline_id, 
    self.id_to_index_mapper.get_node_index(info.clip_id)
);
```

webrender/src/display_list_flattener.rs
```diff
@@ -733,12 +748,19 @@ impl<'a> DisplayListFlattener<'a> {
...
-                                .iter()
-                                .map(|id| self.id_to_index_mapper.get_clip_node_index(*id))
-                                .collect();
-                let parent = info.parent.map(|id|
-                     self.id_to_index_mapper.get_clip_chain_index(&ClipId::ClipChain(id))
-                );
+                    .iter()
+                    .map(|id| self.id_to_index_mapper.get_clip_node_index(*id))
+                    .collect();
+                let parent = match info.parent {
```
mrobinson: I think you can `map_or_else` here:

```rust
let parent = info.parent.map_or_else(
    || self.pipeline_clip_chains.get(&pipeline_id).cloned(),
    |id| self.id_to_index_mapper.get_clip_node_index(*id))
);
```

mrobinson: Oh, I see that map_or_else won't wok here with parent accepting an `Option`. Perhaps it makes sense to always fall back to `ClipChainIndex(0)`.

webrender/src/display_list_flattener.rs
kvark: good point!

webrender/src/display_list_flattener.rs
kvark: but that's the whole point of the PR - to take `clip_and_scroll_ids` into account

kvark:
Thanks for the review @mrobinson ! Please take another look.

webrender/src/display_list_flattener.rs
```diff
@@ -539,6 +543,16 @@ impl<'a> DisplayListFlattener<'a> {
             ),
         );
 
+        let clip_chain_index = match clip_and_scroll_ids.clip_node_id {
```
mrobinson: Here I think you always want to use the newly added clip node which is represented by `info.clip_id`. This ensures that contents don't escape the iframe at all. Trying to reach for the parent means that the content might escape the iframe and only be contained by whatever clip happens to be the parent here. If Gecko and Servo want to contain the content to the iframe (which I think they always do), they would end up needing to push an extra clip node. I think that `info.clip_id` is the correct choice here to ensure consistent behavior.

In the future maybe it makes sense that iframes don't automatically clip, in which case we wouldn't push this clip node and we would in that case look at the clipping node of `clip_and_scroll_ids` in this spot. That would allow clients to allow iframe contents to escape their rectangles, while still avoiding having to push an extra clip node to prevent it.

staktrace: I agree that Gecko at least always wants to contain stuff inside the iframe to the iframe clip. If I'm understanding this right, what @mrobinson is suggesting would fix https://bugzilla.mozilla.org/show_bug.cgi?id=1473940

kvark:
Here is a thing. I tested the following variants on @staktrace reduced test case:
  1. original FF from m-c
  2. my PR
  3. @mrobinson 's proposed change

And I'm seeing all these producing the same faulty behavior: content is *mostly* clipped, except when scrolling. I was able to (wasn't easy!) catch a screenshot of it:
wr-content-override

So, looks like we are back to square one? The behavior was certainly fixed when I first filed the PR. Something has changed now.

staktrace:
It's possible that there are two different bugs here. I'll test the other STR (from bug 1462955) and see if this patch fixes that.

staktrace:
Confirmed that the build from your [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=2af2bc05f7ca7043c234805de9f2eeaaca7bcf92) fixes bug 1462955. So I think this is still worth landing. Bug 1458373 is likely a different root cause.

kvark:
@staktrace thanks for verifying!
I pushed another try with the change proposed by @mrobinson , gotta see if it works as well as my original change: https://hg.mozilla.org/try/rev/0c5e83647b894b6585ad0b249801ffab82df9f67

staktrace:
@kvark The try push in the previous comment also fixes the bug

kvark:
Alright, I changed the code according to @mrobinson suggestion, we had a good try push, and I believe all the review concerns are addressed. Let's not hold this patch for too much longer then.
@bors-servo r=mrobinson

PR #2878, opened 2018-07-10
Update dependencies

Eijebong:
Including glutin, winit, core-* stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2878)


wrench/src/main.rs
```diff
@@ -259,7 +260,7 @@ fn make_window(
             let window_builder = winit::WindowBuilder::new()
                 .with_title("WRench")
                 .with_multitouch()
-                .with_dimensions(size.width, size.height);
+                .with_dimensions(glutin::dpi::LogicalSize::new(size.width as f64, size.height as f64));
```
Eijebong: Woups

examples/multiwindow.rs
```diff
@@ -171,10 +172,10 @@ impl Window {
         }
 
         let framebuffer_size = {
-            let (width, height) = self.window.get_inner_size().unwrap();
-            DeviceUintSize::new(width, height)
+            let glutin::dpi::LogicalSize {width, height} = self.window.get_inner_size().unwrap();
```
Eijebong: woups²

staktrace:
Right now all the taskcluster failures are just linting problems.

But I can help with updating rust versions - what version of rust do you want to be running on?

Eijebong:
Mhh maybe it's not necessary then, 1.25 seems to be working but if we can get to the same version as firefox it'd be nice :)

PR #2880, opened 2018-07-10
Reset the active shader to whatever it was before after rendering

fschutt:
Some external toolkits (such as glium) can't see that webrender
has changed the shader under their feet. glium, for example,
assumes that the active shader program hasn't changed, but in
reality, webrender sets the active shader to 0 after a draw call,
so glium::use_program() doesn't do anything because glium thinks
that the original shader is still active.

Generally, to be a good citizen in OpenGL-land, it's not a good
idea to leave and state changes behind after one frame - the
webrenders rendering should be invisible to any other toolkits,
otherwise workarounds around this no-cleanup "feature" would be
required.

Also see: https://github.com/glium/glium/issues/1696

cc @glennw

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2880)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -2241,7 +2241,11 @@ impl Renderer {
         &mut self,
         framebuffer_size: DeviceUintSize,
     ) -> Result<RendererStats, Vec<RendererError>> {
-        self.render_impl(Some(framebuffer_size))
+        let mut current_program = [0_i32];
+        unsafe { self.device.gl().get_integer_v(gl::CURRENT_PROGRAM, &mut current_program) };
+        let result = self.render_impl(Some(framebuffer_size));
+        self.device.gl().use_program(current_program[0] as u32);
+        result
     }
 
     // If framebuffer_size is None, don't render


fschutt:
Note: I'm not happy about putting it all into the `render` function instead of the `render_impl` function, but on the other hand, I don't know why `render` and `render_impl` are two functions in the first place and where else I should put this code. If anyone has a better idea where to put this, I'm welcome for suggestions.

webrender/src/renderer.rs
```diff
@@ -2241,7 +2241,11 @@ impl Renderer {
         &mut self,
         framebuffer_size: DeviceUintSize,
     ) -> Result<RendererStats, Vec<RendererError>> {
-        self.render_impl(Some(framebuffer_size))
+        let mut current_program = [0_i32];
+        unsafe { self.device.gl().get_integer_v(gl::CURRENT_PROGRAM, &mut current_program) };
```
kvark: I don't really like the idea of querying the current state in WR and trying to preserve it on exiting the frame. If you want to go there, you'd have to take care of the texture bindings, buffer bindings, framebuffers, and possibly more.

On the other hand, leaving a complete mess after ourselves is not a polite way to end a frame either. Hence, we have the `end_frame` function that resets the state to defaults, including the current program.

If you want to embed WebRender in glium, I'd prefer the gluing code taking care of restoring the glium state, not WR.

fschutt: Well okay, I get that - but then there should be more documentation on what GL state webrender actually modifies. I only noticed that the framebuffer doesn't get unbound because of [this comment in limn](https://github.com/christolliday/limn/blob/0862eb02a46bc6a6610bc6911c3420b5f1a32665/examples/opengl_boxes.rs#L36-L37). Otherwise, I'd have stepped around in the dark wondering why it wouldn't render to the screen.

Or we could modify the `end_frame` function and do as you said - reset the texture, framebuffer, etc. bindings to what it was before. I'd actually prefer that - then we have a `start_frame`, which saves the current state and the `end_frame` doesn't resets everything to the default values, but rather the saved values. So that webrenders state manipulation doesn't affect anything outside of it.

Mainly this is because of ´ExternalImage` - if you want to draw an OpenGL texture, you have to do it somewhere in between frames (i.e. after webrender has rendered something and before the next display list gets built). So if webrender doesn't clean up after the frame, you have to know about these issues (right now). 

I'm not sure about this decision either. I'll look into it tomorrow how much work it would actually be to do a clean reset (cleaning up all GL state in `end_frame` correctly).

kvark: To clarify, I'm not proposing `end_frame` to restore the state of `start_frame`. I think what it currently does is reasonable, and all we need is documenting the set of states, so that any embedding code knows what to stash.

fschutt:
Alright, then I'll close this and do it in the wrapper instead.

PR #2883, opened 2018-07-11
Move SpatialNode and ClipNode to their own files

mrobinson:
These nodes types are not really related so it makes sense to give them
their own source files. 'clip_scroll_node' as a concept doesn't exist
any longer.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2883)


kvark:
Looks like it's just moving things around, and making them nicer. Thank you!
@bors-servo r+

Issue #2877, opened 2018-07-09
Switch to winit = "0.16"


ekicyou:
In webrender, winit = "0.13" is used, but "0.16" is already used in upstream servo.

As I noticed at #2838, in "0.16" there are many destructive changes to the API, and the organization of functions has advanced. Since PR currently related to winit is difficult, I would like to switch.

kvark:
Note that WebRender itself doesn't care about `winit`. It's the `Wrench` and examples that use it, and they are perfectly safe to update without breaking anything else.

atouchet:
This was done in #2878.

PR #2886, opened 2018-07-12
use WS_EX_NOREDIRECTIONBITMAP direct-composition app

ekicyou:
`with_no_redirection_bitmap` API added to master branch of `winit=0.16.0`.
It is useful for direct-composition samples.

resubmit #2838 PR,
Thanks! #2878, #2877.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2886)

### DIFF
--- a/direct-composition/src/main_windows.rs
+++ b/direct-composition/src/main_windows.rs
@@ -12,7 +12,7 @@ extern crate winit;
 use direct_composition::DirectComposition;
 use std::sync::mpsc;
 use webrender::api;
-use winit::os::windows::WindowExt;
+use winit::os::windows::{WindowExt, WindowBuilderExt};
 use glutin::dpi::LogicalSize;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


SimonSapin:
Looks good, thanks!

@bors-servo r+

I didn’t know what `WS_EX_NOREDIRECTIONBITMAP` is, and https://docs.microsoft.com/en-us/windows/desktop/winmsg/extended-window-styles was not super informative. https://stackoverflow.com/questions/38179033/when-should-ws-ex-noredirectionbitmap-be-used and https://msdn.microsoft.com/magazine/dn745861.aspx have a better explanation. (That last article would have been useful to find while figuring out how to use DirectComposition to make this demo!)

For posterity, this is what the demo app looks like now (with in the background the terminal where I compiled it).
capture3

ekicyou:
Thank you for adding explanation.
Although commit is dare to mention `.with_decorations(true)`,When false, the world without windows spreads.

Even though it is Windows :-)

By using WS_EX_NOREDIRECTIONBITMAP and DirectComposition, you can develop applications without windows by GPU rendering (2D / 3D) (such as desktop mascot).

PR #2888, opened 2018-07-12
Remove redundant code, now that only local clips are segmented.

gw3583:
As of #2859, this code path can never be executed. Removing it
also simplifies some of the code I'm working on for local space
picture rasterization.

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -33,7 +33,7 @@ use segment::SegmentBuilder;
 use spatial_node::SpatialNode;
 use std::{mem, usize};
 use std::sync::Arc;
-use util::{MatrixHelpers, WorldToLayoutFastTransform, calculate_screen_bounding_rect};
+use util::{MatrixHelpers, calculate_screen_bounding_rect};
 use util::{pack_as_float, recycle_vec};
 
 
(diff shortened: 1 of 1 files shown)


gw3583:
r? @mrobinson or @kvark 

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=8bf7fcf0c5a0ddcdbf8e59be6396355782a1168f

gw3583:
Holy cow, that's a lot of orange on the try run! @staktrace Is this a known issue on try runs right now, or did I somehow manage to break things?

gw3583:
A lot of the errors are:

```
'internal error: entered unreachable code: Got a non Clip ClipId for root reference frame.', gfx/webrender_bindings/src/bindings.rs:2444:14
```

The code I've changed here is clip related, but I can't see how it would cause that. Maybe the bindings are out of sync or something like that?

gw3583:
Ah, I think the following code in ```bindings.rs``` probably needs to be updated after #2871 and #2883.

```
#[no_mangle]
pub extern "C" fn wr_root_scroll_node_id() -> usize {
    // The PipelineId doesn't matter here, since we just want the numeric part of the id
    // produced for any given root reference frame.
    match ClipId::root_scroll_node(PipelineId(0, 0)) {
        ClipId::Clip(id, _) => id,
        _ => unreachable!("Got a non Clip ClipId for root reference frame."),
    }
}
```

Is that right @mrobinson ?

gw3583:
@mrobinson Ah, sure - I'll apply that patch and do a new try run. Thanks!

gw3583:
Pending try with that patch:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=f77c2ce082b48239229d7d917f3af7ad2bcaafd5

mrobinson:
The try job look pretty good to me. There are a couple crashes, but I'm not sure that they are related.

Issue #2887, opened 2018-07-12
FPS counter of profiler does not count number of frame generation


sotaroikeda:
With gecko, there is a case that the FPS counter become 120 fps like [Bug 1474595](https://bugzilla.mozilla.org/show_bug.cgi?id=1474595) and [Bug 1470327](https://bugzilla.mozilla.org/show_bug.cgi?id=1470327). In these cases, gecko requests to generate frame by 60fps. But FPS counter shows 120fps.

sotaroikeda:
It seems to related to number of  Renderer::render_impl(). The function was called from Renderer::update() and from Renderer::render().

kvark:
Fixed by  #2889

PR #2889, opened 2018-07-12
Fix fps counter

sotaroikeda:
Address #2887.

Current FPS counter count calls of render_impl()s. But we normally expects frame generation counts.

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -2361,8 +2361,10 @@ impl Renderer {
         });
 
         let current_time = precise_time_ns();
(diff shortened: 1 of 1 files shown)


nical:
Oh boy.
@bors-servo r+

kvark:
@sotaroikeda FYI, github recognizes some of the words like "Fixes xxx" or "Closes xxx", so if you used those instead of "Address xxx", it would automatically close the issue upon merging a PR, which is considered a good practice.

sotaroikeda:
Thanks!

PR #2891, opened 2018-07-12
Generate shaders in build time instead of run time

zakorgy:
The discussion about this stared here: https://github.com/szeged/webrender/issues/197#issue-337915626

1. Added a helper `shaders.ron` file which contains the possible feature sets for each shader.
2. In the build script we generate shaders using the information from this `ron` file, for each possible feature configuration. The generated shader file names contain the feature names in lower case separated by a `_`.
The shader processing code parts (get_shader_source, parse_shader_source, build_shader_strings) are moved in the build script from `device/gl.rs`.
3. In run-time we pass the base name of the shader and the used features to load the proper shader file.

There is a hacky [part](https://github.com/servo/webrender/compare/master...szeged:build-shaders#diff-0ad2b4b9ac27cf5624c19208f294d671R186), because we don't always know which shader version (GL or GLES) we intend to use in build time. Replacing it in run-time, if we differ from the default `#version 150` (GL), seemed the best solution.

@kvark, @gw3583 could you please take a look?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2891)


webrender/build.rs
```diff
@@ -27,22 +38,143 @@ fn write_shaders(glsl_files: Vec<PathBuf>, shader_file_path: &Path) {
...
+        shader_map.insert(shader_name.clone(), full_name.clone());
+    }
+    shader_map
+}
+
+// Get a shader string by name
+fn get_shader_source(shader_name: &str, shaders: &HashMap<String, String>) -> Option<String> {
+    if let Some(shader_file) = shaders.get(shader_name) {
+        let shader_file_path = Path::new(shader_file);
+        if let Ok(mut shader_source_file) = File::open(shader_file_path) {
```
kvark: shouldn't we panic if the shader can't be opened?

webrender/build.rs
```diff
@@ -27,22 +38,143 @@ fn write_shaders(glsl_files: Vec<PathBuf>, shader_file_path: &Path) {
...
+    fs_source.push_str(&shared_result);
+
+    (vs_source, fs_source)
+}
+
+fn generate_shaders(out_dir: &str, shaders: HashMap<String, String>) {
+    let mut file = File::open("res/shaders.ron").expect("Unable to open shaders.ron");
+    let mut source = String::new();
+    file.read_to_string(&mut source).expect("Unable to read shaders.ron");
+    let shader_configs: Vec<Shader> = de::from_str(&source).expect("Unable to deserialize shaders.ron");
```
kvark: would probably be easier to do [from_reader](https://docs.rs/ron/0.3.0/ron/de/fn.from_reader.html)

webrender/res/shaders.ron
```diff
@@ -0,0 +1,130 @@
...
+        ],
+    ),// [17]
+    (
+    name: "pf_vector_cover",
+    feature_sets: [
+            "TEXTURE_2D",
+        ],
+    ),// [18]
+    (
+    name: "pf_vector_stencil",
```
kvark: something weird with the indentation here

webrender/tests/angle_shader_validation.rs
```diff
@@ -16,8 +16,6 @@ struct Shader {
     features: &'static [&'static str],
 }
 
-const SHADER_PREFIX: &str = "#define WR_MAX_VERTEX_TEXTURE_WIDTH 1024\n";
```
kvark: why does the shader validator need to have a separate list of shaders and features, given that we already have them described in RON?

zakorgy:
@kvark I have addressed your comments. The validator now loads the shader configs from the RON file, however I had to add a list of excluded features, because the extensions used by these features are not supported by the angle validator.

kvark:
@zakorgy thank you! Please also not leave the "fixup! ..." commits and instead do a real git fixup for those, making sure that each is buildable and has a concrete purpose.
For merging, we'll need to have @gw3583 to review it.

gw3583:
@kvark Looks like this is ready to go?

kvark:
@gw3583 Thanks for looking (I assume you did, right?)
@bors-servo r+

staktrace:
I'm guessing this PR is the one that broke firefox CI with shader errors: https://treeherder.mozilla.org/#/jobs?repo=try&revision=059d4f6b01e13547633496cc649c76b0f8c6e282

kvark:
@staktrace that is unfortunate. Reverting in #2912 until the Gecko side is clear.

zakorgy:
@staktrace could you please point me how can I reproduce the error from the CI. I'm not familiar with gecko tests, and it would be nice to know which command/script should I use.

staktrace:
@zakorgy The simplest way is to build Firefox using the [instructions] (https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions) and then apply your changes to the `gfx/webrender{,api}` folders. Then rebuild and run to make sure it doesn't blow up. You can also run reftests using `./mach reftest` . If you get the reftests in `layout/reftests/reftest-sanity` working that's probably a good sign and we can do a try push to see if anything else fails.

PR #2756, opened 2018-05-14
Add an explicit API to add reference frames

mrobinson:
Previously reference frames were inserted into the ClipScrollTree when a
stacking context had a transformation or a perspective. This changes
makes the API for adding reference frames explicit in order to give
clients better control over when and how they are created. This change
also opens up future opportunities for shrinking the size of the display
list (by separating out the ClipScrollTree from the list of display
items) and for concurrency during the flattening process.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2756)

Issue #2551, opened 2018-03-20
[RFC] Simplifying the clipping API.


glennw:
This is a writeup of a conversation I had with @mstange. I'll try to summarize what we discussed, and list what I think would change in WR code.

One of the things that complicates drawing a Picture in local space is that clips that affect the Picture can be defined in other coordinate spaces, due to the way ancestor clip-scroll nodes can be referenced. See https://github.com/servo/webrender/issues/2533 for details.

Gecko doesn't actually require this. The Gecko requirements for clipping are that:
 * A primitive can have clips defined that are in the same reference frame (stacking context from public API point of view).
 * Any clip from outside that reference frame is applied to the *stacking context*. This is done by drawing the stacking context to an intermediate surface, and then applying the ancestor clips to the stacking context itself, rather than the primitives *within* the stacking context.

Previously, this was not really feasible in WR, because we didn't support clip masks on stacking contexts. Now that stacking context blits are drawn as brush primitives, we do (mostly) support this, and expanding it to support in all cases should be trivial (more details below).

If we were to change WR such that it supports clipping the way Gecko uses (that is, local clips on primitives, and ancestor clips on stacking contexts), that has the potential to:
 * Drastically simplify implementing local-space rendering (#2533) should we decide to.
 * Make the existing clipping code simpler (I think) and significantly faster. The clip mask for an ancestor clip-chain would only need to be drawn once, and since Picture drawing supports segments just like normal primitives, we would render only minimal clip masks in this case.

These are the steps that I envisage we'd take:
 * Complete the WR support for clipping stacking contexts. Effectively, this should work now for stacking contexts that already draw to an intermediate surface. All we should need to do here is add a piece of code that marks a Picture as needing an intermediate surface if it contains a complex clip chain that affects the stacking context rect (the code to do the intermediate rendering / clipping is already in place and works).
 * Update Gecko to make use of that API (not sure how complex this is?).
 * Deprecate and remove support in the WR display list builder for adding clips to primitives that are outside its reference frame.

@mstange @mrobinson @staktrace @kvark Does that sound like a correct summary? Any corrections / thoughts / questions?

glennw:
cc @jrmuizel too

staktrace:
[...]

PR #2892, opened 2018-07-13
Improve clipping and positioning documentation

mrobinson:
The new documentation gives a more complete and up-to-date picture of
WebRender's clipping and positioning architecture.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2892)


leeoniya:
is anything mentioned these issues still relevant for Ideas for the Future?

https://github.com/servo/webrender/issues/2551

> ...when we separate the concept of
reference frames and stacking contexts.

https://github.com/servo/webrender/pull/2736

mrobinson:
@leeoniya Stacking contexts and reference frames are now two separate concepts after #2756. The idea of applying clips to stacking contexts was implemented in #2600. We are still working out exactly how flexible Gecko requires the display list to be.

staktrace:
Thanks for writing this documentation! I read through it and it clarified a few things for me. I agree that differentiating between the three kinds of things in the API would be good and make things more robust. I was looking at the gecko side code and it seems like every clip node that we create with a parent has what is effectively a spatial node as the parent (i.e. it's the WrClipId corresponding to a scroll frame, a reference frame, or a sticky node) with one exception: [here](https://searchfox.org/mozilla-central/rev/46292b1212d2d61d7b5a7df184406774727085b8/layout/painting/nsDisplayList.cpp#9366) we create a clip node with a mask, and then use that clip code as the parent for other clip nodes that might get defined on items nested inside the mask.

If we can do this in a different way (e.g. including the mask in the clip chain of the nested items) then that would clean this up and we should be able to separate WrClipId from a WrSpatialNodeId on the gecko side. That would be a good first step towards a better API, I think.

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
...
+    transformation of each node depending on the relative transformation of the
+    node combined with the transformations of all of its ancestors. These nodes
+    are responsible for positioning display list items and clips.
+ 2. A collection of `ClipNodes` which specify a rectangular clip and, optionally,
+    a set of rounded rectangle clips and a masking image.
+ 3. A collection of `ClipChains`. Each `ClipChain` is a list of `ClipNode`
+    elements. Every display list item has an assigned `ClipChain` which
+    specifies what `ClipNodes` are applied to that item.
+
+The `SpatialNode` of the clips applied to each item are completely independent of
```
kvark: `SpatialNodes`

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
...
+the `SpatialNode` applied to the item itself.
+
+One holdover from the previous design is that both `ClipNode` and `SpatialNodes`
+have a parent node, which is either a `SpatialNode` or a `ClipNode`.  From this
+node WebRender can determine both a parent `ClipNode` and a parent `SpatialNode`
```
kvark: This is confusing me quite a bit. Why do we want to support the `ClipNode` being a parent, ever?

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
...
+any `ClipNode` appended to the chain. An example of the goal of this process is
+to avoid having to render a mask for a large rounded rectangle when the rest of
+the clip chain constrains the content to an area completely inside that
+rectangle. Avoiding mask rasterization in this case and others has large
+performance impacts on WebRender.
+
+# Clipping and Positioning in the Display List
+
+Each non-structural WebRender display list item has
+ * A `ClipId` of a `SpatialNode` or `ClipNode` for positioning
```
kvark: a clip id of a clip node for positioning - this rings a bell for me

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
...
+careful when using the API. For instance, when specifying the parent of
+`ClipNode` one can use the `ClipId` or another `ClipNode` or a `SpatialNode`,
+but not one for a `ClipChain`.
+
+WebRender's internal representation of clipping and positioning is not a perfect
+match to the display list representation of these concepts. This is due, again,
+to the evolutionary nature of the design. The general trend is that the display
+list gradually moves toward the internal representation. The most important of
+these incongruities is that while `ClipNodes`, sticky frames, and scroll frames
+are defined and simply return a ClipId, reference frames return a ClipId, but
```
kvark: nit: codify `ClipId`

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
...
+also are pushed and popped like stacking contexts.
+
+## Converting `ClipId` to global `ClipScrollTree` indices
+
+WebRender must access `ClipNodes` and `SpatialNodes` quite a bit when building
+scenes and frames, so it tries to convert `CiipIds`, which are already
```
kvark: "CiipIds"

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
...
+per-pipeline indices, to global scene-wide indices.  Internally this a
```
kvark: a verb is missing? ("this a")

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
...
+conversion from `ClipId` into `SpatialNodeIndex`, `ClipNodeIndex` or
+`ClipChainIndex`. In order to make this conversion cheaper, the
+`DisplayListFlattner` assigns offsets for each pipeline and node type in the
+scene-wide `ClipScrollTree`.
+
+Nodes are added to their respective arrays sequentially as the display list is
+processed during "flattening." When encountering an iframe, the
```
kvark: let's not include the "." in quote marks

webrender/doc/CLIPPING_AND_POSITIONING.md
```diff
@@ -0,0 +1,167 @@
+# Original Design
+
+To understand the current design for clipping and positioning (transformations
+and scrolling) in WebRender if can be useful to have a little background about
```
mrobinson: Fixed.

webrender/doc/CLIPPING_AND_POSITIONING.md
mrobinson: I changed this to make the inflection of the noun match the subject.

webrender/doc/CLIPPING_AND_POSITIONING.md
mrobinson: Nowadays there is no reason this has to be the case, but consider that in the past the only way to combine clip nodes was to have them be part of the ancestry in the `ClipScrollTree`. With the ClipChain API and the fact that an item can have both a positioning node and clip node assigned, it isn't strictly necessary. Once both Gecko and Servo are using the ClipChain API for all clips and never using ClipNodes as positioning nodes, we can fix this for good.

webrender/doc/CLIPPING_AND_POSITIONING.md
mrobinson: Fixed.

webrender/doc/CLIPPING_AND_POSITIONING.md
mrobinson: Fixed.

webrender/doc/CLIPPING_AND_POSITIONING.md
mrobinson: Fixed.

webrender/doc/CLIPPING_AND_POSITIONING.md
mrobinson: So weirdly this is the style in American English, which I admit is bizarre. Since the rest of the document is written in American English though, maybe it makes sense to be consistent here.

webrender/doc/CLIPPING_AND_POSITIONING.md
kvark: Thank you for explanation! @staktrace do you know if there is an issue to file/track for this on Gecko side?

staktrace: @kvark I don't believe there is; please feel free to file one

kvark: Thanks! Filed as https://bugzilla.mozilla.org/show_bug.cgi?id=1476044

Issue #2890, opened 2018-07-12
Very high CPU usage with a plain text file


automatedbugreportingfacility:
This source files causes CPU usage spikes when scrolling horizontally, trying to highlight any text, or trying to move the tab around on the tab bar (supposedly, due to the tab thumbnail overlapping with the content area, so curiously, even drawing over static content causes issues here). Also, the text may disappear while scrolling.

URL (view-source: URLs are not web accessible, please copy and paste into the URL bar):
view-source:https[]()://cdn.rawgit.com/WebAssembly/wabt/aae5a4b7/demo/libwabt.js

kvark:
I'm seeing 70ms - 80ms render backend CPU time, and about 7 million vertices on the GPU side.

kvark:
There is less then 800 primitives on the page, but they appear to contain very long text runs. Our logic currently has to go through each glyph in a run and request it, which involves a hashmap access among other things. Perhaps, we could find out only a subset of glyphs to be requested based on the screen rectangle?..
Also, those appear related:
```
        // TODO(gw): In the future, remove `glyph_instances`
        //           completely, and just reference the glyphs
        //           directly from the display list.
```
```
            // TODO(gw): If we support chunks() on AuxIter
            //           in the future, this code below could
            //           be much simpler...
```

nical:
See also #1502.

PR #2894, opened 2018-07-13
Reduce glyph key to 32 bits

kvark:
Related to #2890.

As we do quite a few hashmap accesses (at least one per frame per glyph) by this key, reducing its size should help.

r? anyone

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2894)

### DIFF
--- a/webrender/src/glyph_rasterizer/mod.rs
+++ b/webrender/src/glyph_rasterizer/mod.rs
@@ -249,7 +249,7 @@ impl FontInstance {
     #[allow(dead_code)]
     pub fn get_subpx_offset(&self, glyph: &GlyphKey) -> (f64, f64) {
         if self.use_subpixel_position() {
-            let (dx, dy) = glyph.subpixel_offset;
+            let (dx, dy) = glyph.subpixel_offset();
             (dx.into(), dy.into())
         } else {
             (0.0, 0.0)
@@ -287,7 +287,8 @@ impl FontInstance {
         if max_size > FONT_SIZE_LIMIT &&
            self.transform.is_identity() &&
            self.render_mode != FontRenderMode::Subpixel &&
-           !self.use_subpixel_position() {
+           !self.use_subpixel_position()
+        {
             max_size / FONT_SIZE_LIMIT
         } else {
             1.0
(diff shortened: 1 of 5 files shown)


gw3583:
```
	reftests/text/shadow-red.yaml == reftests/text/shadow-red-ref.yaml
	reftests/text/shadow-grey.yaml == reftests/text/shadow-grey-ref.yaml
	reftests/text/shadow-grey-transparent.yaml == reftests/text/shadow-grey-ref.yaml
	reftests/text/synthetic-bold-transparent.yaml == reftests/text/synthetic-bold-transparent-ref.yaml
	reftests/text/ahem.yaml == reftests/text/ahem-ref.yaml
	reftests/text/isolated-text.yaml == reftests/text/isolated-text.png
	reftests/text/colors.yaml == reftests/text/colors-alpha.png
	reftests/text/colors.yaml == reftests/text/colors-subpx.png
	reftests/text/colors.yaml == reftests/text/colors-subpx.png
	reftests/text/border-radius.yaml == reftests/text/border-radius-alpha.png
	reftests/text/border-radius.yaml == reftests/text/border-radius-subpx.png
	reftests/text/text-masking.yaml == reftests/text/text-masking-alpha.png
	reftests/text/text-masking.yaml == reftests/text/text-masking-subpx.png
	reftests/text/alpha-transform.yaml == reftests/text/alpha-transform.png
	reftests/text/subpixel-rotate.yaml == reftests/text/subpixel-rotate.png
	reftests/text/subpixel-scale.yaml == reftests/text/subpixel-scale.png
	reftests/text/subpixel-skew.yaml == reftests/text/subpixel-skew.png
	reftests/text/embedded-bitmaps.yaml == reftests/text/embedded-bitmaps.png
	reftests/text/clipped-transform.yaml == reftests/text/clipped-transform.png
	reftests/text/blurred-shadow-local-clip-rect.yaml == reftests/text/blurred-shadow-local-clip-rect-ref.png
	reftests/text/two-shadows.yaml == reftests/text/two-shadows.png
	reftests/text/shadow-clip.yaml == reftests/text/shadow-clip-ref.yaml
	reftests/text/shadow-partial-glyph.yaml == reftests/text/shadow-partial-glyph-ref.yaml
	reftests/text/shadow-transforms.yaml == reftests/text/shadow-transforms.png
	reftests/text/raster-space.yaml == reftests/text/raster-space.png
```

I'm not sure this is a valid change - I think the index value for a glyph can make use of the full 32 bit range on some fonts / platforms?

kvark:
I double checked our `macos` and `windows` modules and they both appear to manage glyph indices as `u16`. As for Unix's FreeType, I wasn't able to find any concrete description of the glyph index format. I suppose it's opaque and we can't actually cut away 4 bits from it. Could you confirm, @lsalzman ?
If that's the case, I'll just make `GlyphKey` be platform-defined, with Unix taking a slightly slower route.

lsalzman:
In practice, glyph ids in faces will probably never exceed a u16, though the types allow for it.

One idea in the worst case for GlyphKey would be to encode everything in a u32 bitfield. 28 bits for index, 2 bits for subpixel X, bits for subpixel Y, since the subpixel offset only requires 2 bits per field? Then you basically wouldn't have any issues fitting large fonts?

kvark:
@lsalzman ok, good, thanks for the confirmation!
> One idea in the worst case for GlyphKey would be to encode everything in a u32 bitfield. 28 bits for index, 2 bits for subpixel X, bits for subpixel Y, since the subpixel offset only requires 2 bits per field? Then you basically wouldn't have any issues fitting large fonts?

Funny thing is that this PR does exactly that. I just made a mistake somewhere, probably.

kvark:
Alright, this should be fixed now. Got bitten by the operator precedence, ouch.

gw3583:
@kvark If mac and windows platforms use a u16, then it's probably reasonable to assume the same on unix platforms. So r=me on this - my only question is whether we need a try run first?

kvark:
I'll make one tomorrow @gw3583 . Thanks for taking a look!

kvark:
Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=58912e858963725d9d762abe8ab1b99227fd8356&selectedJob=188596690

kvark:
There appear to be a bunch of intermittent timeouts, but I don't see anything related to the PR. Am I using the wrong try settings, or just an unlucky base revision?

kvark:
The new [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=dccb314da45eef002ef011d910fafe6c9f316f35) looks better.
@bors-servo r=gw3583

PR #2896, opened 2018-07-16
allow custom gamma and contrast for DWrite fonts

lsalzman:
This is to resolve Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1474968

This passes in gamma and contrast so we can parameterize preblending more closely to how Skia is doing so in Gecko.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2896)

### DIFF
--- a/webrender_api/src/font.rs
+++ b/webrender_api/src/font.rs
@@ -268,14 +268,16 @@ impl Default for FontInstanceOptions {
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Deserialize, Hash, Eq, PartialEq, PartialOrd, Ord, Serialize)]
 pub struct FontInstancePlatformOptions {
-    pub unused: u32,
+    pub gamma: u16, // percent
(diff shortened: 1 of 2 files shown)


webrender_api/src/font.rs
```diff
@@ -268,14 +268,16 @@ impl Default for FontInstanceOptions {
...
+    pub gamma: u16, // percent
+    pub contrast: u16, // percent
 }
 
 #[cfg(target_os = "windows")]
 impl Default for FontInstancePlatformOptions {
     fn default() -> FontInstancePlatformOptions {
         FontInstancePlatformOptions {
-            unused: 0,
+            gamma: 220,
```
gw3583: The previous default constant used was 230 - is 220 correct? Also, could we add a comment here describing why this is the default?

gw3583:
Just one minor comment - otherwise looks good!

PR #2902, opened 2018-07-17
Add Transform struct, decouple some prim code from SpatialNode.

gw3583:
The transform information in a SpatialNode contains a complete
local -> screen space transform. In the future, we want to be
able to selectively rasterize certain pictures in local space
(e.g. in a nested 3d transform with perspective).

To achieve this, TransformPalette will be extended to allow
construction of a Transform relative to a given coordinate
system / spatial node.

This patch starts decoupling the culling / prim CPU code from
directly referencing a SpatialNode, and instead retrieving the
relevant information from a Transform provided by the
TransformPalette.

Also rename some instances of scroll_node_id to spatial_node_index
as appropriate.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2902)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -477,7 +477,7 @@ impl AlphaBatchBuilder {
         for run in &pic.runs {
             let transform_id = ctx
                 .transforms
-                .get_id(run.clip_and_scroll.scroll_node_id);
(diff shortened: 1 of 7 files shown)


mrobinson:
@bors-servo r+

Thanks for fixing the remaining references to "scroll_node" as well!

PR #2903, opened 2018-07-17
Use dpi from winit instead of glutin

zakorgy:
For future backends it would be better to use `dpi` module from `winit`, since `glutin` is gl related. Also this PR reduces the dependencies of `direct-composition`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2903)


kvark:
Thank you! The change looks safe to merge.
@bors-servo r+

PR #2904, opened 2018-07-17
Update euclid to 0.18

Eijebong:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2904)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: trusty
 language: rust
 rust:
-  - 1.25.0
+  - 1.27.0
   - nightly
 matrix:
   fast_finish: true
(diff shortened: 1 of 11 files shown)


staktrace:
This PR seems to have caused some panics in Firefox CI, see https://bugzilla.mozilla.org/show_bug.cgi?id=1476636#c1

staktrace:
Here's a sample log: https://treeherder.mozilla.org/logviewer.html#?job_id=188772693&repo=try&lineNumber=1438

nical:
The panic is because this https://github.com/servo/webrender/pull/2904/files#diff-9c1d8b23da9e37ae3377086804c5aaf2L588 used to always have `Some(transformed rect)` and now can have `None` since transforming a rect is fallible in the latest euclid, since https://github.com/servo/euclid/pull/291.

cc @kvark 

I suppose that in this case we could cull out entirely the rects that we couldn't project.

PR #2905, opened 2018-07-18
Remove some redundant rectangle transform and union code.

gw3583:
This was probably a merge error - this is already accounted for
a few lines above when we transform the clipped_rect field.

Saves a rectangle transform + union per primitive per frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2905)


gw3583:
Something I noticed while working on related code.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=6bb18fd45124fe32ff94cdeadc7d7b8633a46640

r? @kvark

PR #2906, opened 2018-07-18
Decouple batching from clip-scroll tree.

gw3583:
A couple small tidy ups in the spatial node code, and also remove
the last dependency during batching on direct references to the
clip-scroll tree.

Once this is complete for culling, we'll be able to selectively
pick which coordinate system an offscreen surface (picture) is
rendered with, via the TransformPalette.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2906)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -670,13 +670,12 @@ impl AlphaBatchBuilder {
                             // Push into parent plane splitter.
                             debug_assert!(picture.surface.is_some());
 
-                            let real_xf = &ctx.clip_scroll_tree
-                                .spatial_nodes[picture.reference_frame_index.0]
-                                .world_content_transform
-                                .into();
+                            let real_xf = &ctx
+                                .transforms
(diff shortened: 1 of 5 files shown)


gw3583:
r? @mrobinson or @kvark 

Apologies for all these small seemingly inconsequential commits - just trying to land small parts of this work as I can to hopefully make reviews easier :)

gw3583:
Sure - it will probably need to wait until we resolve the panics due to the euclid update PR, I guess?

gw3583:
I cherry-picked this commit on top of the current SHA that Gecko is using. Pending try here, hopefully that should work:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=76e20abee844dca8ace7450eb3d0bb866e914abc

gw3583:
Try run looks good, I think - the orange looks like unrelated intermittents.

kvark:
Thank you!
@bors-servo r+

PR #2910, opened 2018-07-18
Framebuffer size wrong in examples

s-panferov:
I believe framebuffer size should be in physical pixels. Please correct me if I'm wrong.

This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2910)

### DIFF
--- a/examples/common/boilerplate.rs
+++ b/examples/common/boilerplate.rs
@@ -152,7 +152,10 @@ pub fn main_wrapper<E: Example>(
     };
 
     let framebuffer_size = {
-        let size = window.get_inner_size().unwrap();
+        let size = window
+            .get_inner_size()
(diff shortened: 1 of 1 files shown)


examples/common/boilerplate.rs
```diff
@@ -152,7 +152,10 @@ pub fn main_wrapper<E: Example>(
     };
 
     let framebuffer_size = {
-        let size = window.get_inner_size().unwrap();
+        let size = window
+            .get_inner_size()
+            .map(|s| s.to_physical(device_pixel_ratio as f64))
```
kvark: nit: would be simpler to first `unwrap` and then convert

s-panferov:
@kvark fixed

Issue #2884, opened 2018-07-11
WebRender crashes when asked to render a border on a giant rect


staktrace:
Simple test page: https://searchfox.org/mozilla-central/source/layout/generic/crashtests/large-border-radius-dashed.html

Causes WebRender to blow up. Specifically from this border display item in the display list:

BorderDisplayItem { widths: BorderWidths { left: 20.0, top: 10.0, right: 20.0, bottom: 10.0 }, details: Normal(NormalBorder { left: BorderSide { color: ColorF { r: 0.0, g: 0.0, b: 0.0, a: 1.0 }, style: Dashed }, right: BorderSide { color: ColorF { r: 0.0, g: 0.0, b: 0.0, a: 1.0 }, style: Dashed }, top: BorderSide { color: ColorF { r: 0.0, g: 0.0, b: 0.0, a: 1.0 }, style: Dashed }, bottom: BorderSide { color: ColorF { r: 0.0, g: 0.0, b: 0.0, a: 1.0 }, style: Dashed }, radius: BorderRadius { top_left: 5000000.0×5000000.0, top_right: 5000000.0×5000000.0, bottom_left: 5000000.0×5000000.0, bottom_right: 5000000.0×5000000.0 } }) }

The numbers in it are large, and it ends up trying to allocate a too-large texture.

bugzilla tracking bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1474885

nical:
The size of the render task items we generate for borders depends directly [on the size of the border radii](https://github.com/servo/webrender/blob/3a85116a17eee9fe203bcff10299de1f3efef2bf/webrender/src/border.rs#L682).
Looking at other render task types this is not the only place where we have this problem.

It would be good to be able to render the borders in screen space rather if they are bigger than a certain size.

gw3583:
Fixed by #2916.

Issue #2915, opened 2018-07-19
Rendering lots of dots/dashes


nical:
WebRender currently generates per-dot and per-dash data is sent to the GPU (as segment data I think). In some unfortunate cases we can end up allocating hundreds of thousands of them which isn't working out very well.

One way we could go about this is to have dashes and dots drawn on the GPU given some initial information computed on the CPU about how many there are and whatever is needed to place them properly (which according to [the amount of effort that goes into it in gecko](https://searchfox.org/mozilla-central/rev/88199de427d3c5762b7f3c2a4860c10734abd867/layout/painting/nsCSSRenderingBorders.cpp) doesn't appear to be that simple).

kvark:
There are 2 problems here:
  1. the amount of data we may want to send via GPU cache. It's unlikely going to change per frame, so from the performance standpoint it's only a single spike, but it may theoretically hit the limit of the GPU cache vertical size.
  2. the fact we expect all the segments from a primitive to fit inside a single GPU cache line, which is very limited (i.e. 512 segments). This could be alleviated by splitting a primitives into sub-primitives, so that each has segments fitting into their corresponding GPU cache lines.

gw3583:
With the way borders are drawn and cached now, we don't (as far as I know) store this information in the GPU cache. @mrobinson and I changed it to store the dot / dash data inside the instance, so that we don't hit GPU cache size limits.

nical:
Ah I see. We still run into cases where we have so many dots and dashes that the CPU side can freeze for a second pushing hundreds of thousands of dots into a vector, though.

PR #2916, opened 2018-07-19
Don't panic when rendering huge border corners

nical:
This is a simple workaround for huge border corners which currently try to allocate very large surfaces to render the mask.

The basic idea is to render them at a lower resolution (the maximum resolution being 2048x2048 device pixels per corner) and let the brush stretch them. As a result huge corners (with a radius bigger than 2048 device pixels) can render blurry, however we don't crash anymore and the code is very simple.
I toyed with rendering only the visible portion of the corner without down scaling, but it required adding a lot of code for a use case that I think isn't worth the maintenance cost in my opinion.

The patch also sets a limit on the number of dots/dashes so that we don't hit the maximum number of segments per primitive. This isn't great but fixing it properly isn't our highest priority. I filed #2915.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2916)

### DIFF
--- a/webrender/src/border.rs
+++ b/webrender/src/border.rs
@@ -9,9 +9,19 @@ use app_units::Au;
 use ellipse::Ellipse;
 use display_list_flattener::DisplayListFlattener;
 use gpu_types::{BorderInstance, BorderSegment, BrushFlags};
-use prim_store::{BrushKind, BrushPrimitive, BrushSegment};
+use prim_store::{BrushKind, BrushPrimitive, BrushSegment, VECS_PER_SEGMENT};
 use prim_store::{BorderSource, EdgeAaSegmentMask, PrimitiveContainer, ScrollNodeAndClipChain};
+use renderer::{MAX_VERTEX_TEXTURE_WIDTH};
 use util::{lerp, RectHelpers};
+
+// Using 2048 as the maximum radius in device space before which we
+// start stretching is up for debate.
+// the value must be chosen so that the corners will not use an
+// unreasonable amount of memory but should allow crisp corners in the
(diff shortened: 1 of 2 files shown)


nical:
With this, gecko can survive the crashtest from [bug 1474885](https://bugzilla.mozilla.org/show_bug.cgi?id=1474885). Fixes #2884.

webrender/src/border.rs
```diff
@@ -345,18 +345,23 @@ impl BorderCornerClipSource {
             1.0 - 2.0 * outer_scale.y,
         );
 
+        // No point in pushing more clips as it will blow up the maximum amount of
+        // segments per primitive later down the road.
+        // See #2915 for a better fix.
+        let max_clip_count = self.max_clip_count.min(512);
```
kvark: nit: should be computed from constants, e.g. `MAX_VERTEX_TEXTURE_WIDTH` divided by the blocks per segment

webrender/src/border.rs
```diff
@@ -927,6 +932,28 @@ impl BorderRenderTaskInfo {
...
+            .max(radii.bottom_left.height)
+            .max(radii.bottom_right.width)
+            .max(radii.bottom_right.height);
+
+        // Using 2048 as the maximum radius in device space before which we
+        // start strecthing is up for debate.
+        // the value must be chosen so that the corners will not use an
+        // unreasonable amount of memory but should allow crips corners in the
+        // common cases.
+        LayoutToDeviceScale::new(2048.0 / r)
```
kvark: nit: define it as a constant (with sound name)

PR #2918, opened 2018-07-20
Add native texture animation to yuv example.

gw3583:
This updates the yuv example to animate the contents of an external
texture handle each composite. It's useful to confirm that an
external native texture will be updated even if no frame is being
built, since the render() step will always request an updated
external native texture.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2918)

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -54,7 +54,7 @@ name = "yuv"
 path = "yuv.rs"
 
 [features]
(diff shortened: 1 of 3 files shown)


gw3583:
r? @kvark or anyone

I'm using this to aid in debugging part of https://bugzilla.mozilla.org/show_bug.cgi?id=1474583

PR #2919, opened 2018-07-20
Set wrench's window title before the event loop starts.

### LINKED ISSUE #2465: Setting Wrench window title inside the event loop
Calling `set_title` from inside the `run_forever` loop is illegal. We need to either change the way our event loop is organized, or the way we provide title updates.

nical:
Fixes #2465.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2919)

### DIFF
--- a/wrench/src/main.rs
+++ b/wrench/src/main.rs
@@ -489,6 +489,12 @@ fn main() {
         notifier,
     );
 
+    if let Some(window_title) = wrench.take_title() {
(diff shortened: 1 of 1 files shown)


kvark:
IIRC, the problem was that the code actually wants to change the title per frame.

gw3583:
Closing / re-opening to try and wake up bors

@bors-servo r=kvark

PR #2920, opened 2018-07-20
Don't panic when compiling the debug overlay shaders fail.

### LINKED ISSUE #2808: debug_font shader has compilation failures
See https://bugzilla.mozilla.org/show_bug.cgi?id=1460442

I think the debug_font shader is only compiled by the DebugRenderer, which only gets used in non-default scenarios (e.g. if a pref is flipped). So that probably accounts for the relatively low volume of crashes. But it happens on a variety of graphics cards so presumably it shouldn't be hard to reproduce and fix.

nical:
Fixes #2808.
There is always going to be a silly driver out there that will fail for whatever reason so let's not panic for something that isn't important to the user.
I did leave the unwrap in wrench sort of on purpose because in the unlikely event that someone using wrench hits the driver bug, they'll be in a much better place to fix or report the issue than our average user, and if we are lucky enough to hit this kind of bug a developer's machine we'll be able can try to fix it.

I'm tempted to remove the debug_render feature. It's not buying much since the shader is compiled lazily and it adds noise to the code (and feature flags are annoyingly easy to break).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2920)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1306,6 +1306,7 @@ struct TargetSelector {
 #[cfg(feature = "debug_renderer")]
 struct LazyInitializedDebugRenderer {
     debug_renderer: Option<DebugRenderer>,
+    failed: bool,
 }
 
 #[cfg(feature = "debug_renderer")]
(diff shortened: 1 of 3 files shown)


webrender/src/renderer.rs
```diff
@@ -1306,18 +1306,33 @@ struct TargetSelector {
 #[cfg(feature = "debug_renderer")]
 struct LazyInitializedDebugRenderer {
     debug_renderer: Option<DebugRenderer>,
```
kvark: cleaner to have `Option<Result<_>>` instead

nical: Why would that make things better?

kvark: because the state expressed with an extra `bool` is less safe and more redundant (e.g. it doesn't prohibit Some() + true combo)

PR #2923, opened 2018-07-21
euclid update

paulrouget:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2923)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,7 +269,7 @@ dependencies = [
 name = "direct-composition"
 version = "0.1.0"
 dependencies = [
- "euclid 0.18.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


Eijebong:
You also need to update the lockfile and this is blocked on https://github.com/nical/lyon/pull/394

`cargo update -p euclid`

paulrouget:
Fixed all the duplicated dependencies.

r?

kvark:
I wonder why/how the euclid update works given that `num-traits` is still 0.1 here. Perhaps, they used the [semver hack](https://github.com/dtolnay/semver-trick) ?..

Eijebong:
@kvark They did. That euclid update is hopefully the last one before getting rid of `num-traits` 0.1

PR #2925, opened 2018-07-23
Simplify the ClipId -> CST id mapping code.

gw3583:
There are some upcoming changes which will modify how and when
we store and represent clip nodes and chains. These changes are
necessary to allow WR to rasterize pictures in a different coordinate
space than just screen space.

This patch simplifies the mapping of ClipId to array indices in
the clip-scroll tree. This will make it simpler to make the
changes discussed above - specifically, we can unify how we handle
clip nodes with the per-primitive complex clips.

The code this removes was a performance optimization to avoid
hashmap lookups. I did some profiling on a number of real world
sites and also on some gecko layer stress tests, and didn't find
any noticeable regression from simplifying this code. However,
once the work above has been completed, we could re-investigate
how to optimize this mapping process, if it ever shows up in a
profile.

In one of the stress tests [1] for spatial and clip node creation,
the code is now slightly faster with this patch (28ms vs 29ms).

[1] https://mattwoodrow.github.io/dl-test/dl-test.html?count=50000&layer=inactive

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2925)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1772,16 +1772,16 @@ impl ClipBatcher {
     ) {
         let mut coordinate_system_id = coordinate_system_id;
         for work_item in clips.iter() {
+            let info = clip_store
+                .get_opt(&work_item.clip_sources)
+                .expect("bug: clip handle should be valid");
             let instance = ClipMaskInstance {
                 render_task_address: task_address,
-                transform_id: transforms.get_id(work_item.spatial_node_index),
+                transform_id: transforms.get_id(info.spatial_node_index),
                 segment: 0,
                 clip_data_address: GpuCacheAddress::invalid(),
                 resource_address: GpuCacheAddress::invalid(),
             };
-            let info = clip_store
-                .get_opt(&work_item.clip_sources)
-                .expect("bug: clip handle should be valid");
 
             for &(ref source, ref handle) in &info.clips {
                 let gpu_address = gpu_cache.get_address(handle);
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -91,28 +91,6 @@ pub enum ClipSource {
     LineDecoration(LineDecorationClipSource),
 }
 
-impl From<ClipRegion> for ClipSources {
-    fn from(region: ClipRegion) -> ClipSources {
-        let mut clips = Vec::new();
-
-        if let Some(info) = region.image_mask {
-            clips.push(ClipSource::Image(info));
-        }
-
-        clips.push(ClipSource::Rectangle(region.main, ClipMode::Clip));
-
-        for complex in region.complex_clips {
-            clips.push(ClipSource::new_rounded_rect(
-                complex.rect,
-                complex.radii,
-                complex.mode,
-            ));
-        }
-
-        ClipSources::new(clips)
-    }
-}
-
 impl ClipSource {
     pub fn new_rounded_rect(
         rect: LayoutRect,
@@ -280,10 +258,14 @@ pub struct ClipSources {
     pub local_outer_rect: Option<LayoutRect>,
     pub only_rectangular_clips: bool,
     pub has_image_or_line_decoration_clip: bool,
+    pub spatial_node_index: SpatialNodeIndex,
 }
 
 impl ClipSources {
-    pub fn new(clips: Vec<ClipSource>) -> Self {
+    pub fn new(
+        clips: Vec<ClipSource>,
+        spatial_node_index: SpatialNodeIndex,
+    ) -> Self {
         let (local_inner_rect, local_outer_rect) = Self::calculate_inner_and_outer_rects(&clips);
 
         let has_image_or_line_decoration_clip =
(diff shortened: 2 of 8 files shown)


gw3583:
Pending gecko try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=5c67919389ff58985dac7edc03227130126ca542

r? @mrobinson or @kvark

gw3583:
Gecko try looks good.

gw3583:
Pending try with the two follow up commits:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7c9d06e2bbe3539f630593cdd5f30ef7bd292993

gw3583:
Try still looks good :+1:

webrender/src/batch.rs
```diff
@@ -1772,16 +1772,16 @@ impl ClipBatcher {
     ) {
         let mut coordinate_system_id = coordinate_system_id;
         for work_item in clips.iter() {
+            let info = clip_store
+                .get_opt(&work_item.clip_sources)
```
kvark: nit: why `get_opt` if we follow with `expect` right away?

webrender/src/clip.rs
```diff
@@ -280,10 +258,14 @@ pub struct ClipSources {
     pub local_outer_rect: Option<LayoutRect>,
     pub only_rectangular_clips: bool,
     pub has_image_or_line_decoration_clip: bool,
+    pub spatial_node_index: SpatialNodeIndex,
 }
 
 impl ClipSources {
-    pub fn new(clips: Vec<ClipSource>) -> Self {
+    pub fn new(
+        clips: Vec<ClipSource>,
```
kvark: we could use `Iterator<Item = ClipSource>` to avoid heap allocation on the client

webrender/src/display_list_flattener.rs
```diff
@@ -97,53 +65,24 @@ impl ClipIdToIndexMapper {
...
-
-        self.clip_chain_map[id]
-    }
-
-    pub fn get_clip_chain_index_and_cache_result(&mut self, id: &ClipId) -> ClipChainIndex {
-        let index = self.get_clip_chain_index(id);
-        self.cached_clip_chain_index = Some((*id, index));
-        index
+    pub fn map_spatial_node(&mut self, id: ClipId, index: SpatialNodeIndex) {
+        debug_assert!(!self.spatial_node_map.contains_key(&id));
```
kvark: nit: could use the returned value from `insert`

webrender/src/hit_test.rs
```diff
@@ -34,6 +34,29 @@ pub struct HitTestClipNode {
     regions: Vec<HitTestRegion>,
 }
 
+impl HitTestClipNode {
+    fn new(node: &ClipNode, clip_store: &ClipStore) -> HitTestClipNode {
```
kvark: nit: `-> Self`

webrender/src/batch.rs
gw3583: Because (for now) we only have a weak handle here, so the freelist API doesn't guarantee it's present, but we know that it should be due to usage here. I intend to refactor how we store clip sources as part of this work.

webrender/src/clip.rs
gw3583: It's not clear to me how this avoids an allocation in this case - could you expand on that?

webrender/src/display_list_flattener.rs
gw3583: Done

webrender/src/hit_test.rs
gw3583: Done

webrender/src/display_list_flattener.rs
```diff
@@ -56,8 +56,8 @@ pub struct ClipIdToIndexMapper {
 
 impl ClipIdToIndexMapper {
     pub fn add_clip_chain(&mut self, id: ClipId, index: ClipChainIndex) {
-        debug_assert!(!self.clip_chain_map.contains_key(&id));
-        self.clip_chain_map.insert(id, index);
+        let _is_new_key = self.clip_chain_map.insert(id, index).is_none();
```
kvark: nit: let's not confuse the reader with `is_none()` here and instead move it inside the assert

webrender/src/clip.rs
kvark: This function doesn't need a vec, since it goes through the elements and constructs other stuff from them. But it's currently requiring the caller to have that vec. If this function gets `IntoIterator` instead, the caller may still provide a vec, or they can just construct a complex iterator (with `chain()`, `zip()`, etc) that is completely on the stack. IIRC, there is at least one place in code where this would be handy.

webrender/src/display_list_flattener.rs
gw3583: Fixed

webrender/src/clip.rs
gw3583: Ah, you're right - I misread that it took ownership of it, missing the ```collect```. Although, having tried to make that change it seems to flow on to changing several other functions, so perhaps better to leave for now (since this PR doesn't actually change that code), unless I'm missing an obvious way to make this change locally?

kvark: It would be correct to change the calling functions as well, where possible, but at the same time `Vec` implements `IntoIterator` so those changes are not required. I'll be happy to follow up with this refactor ;)

mrobinson:
If the LRU cache and non-hashing mapping aren't important to the code, then these changes simplify the code a lot. Looks good to me.

mrobinson:
@glennw It's an interesting question. It could be that this approach is still probably hashing quite a bit less than when we were not storing ClipScrollTree nodes in vectors and converting the ids to indexes.

Issue #2917, opened 2018-07-20
Investigate high CPU usage on youtube playback


gw3583:
Relevant bugzilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1474583

It looks likely that we can skip the frame build step, when we detect only an external native texture is changing. Needs some investigation to see how easy this is to detect.

PR #2927, opened 2018-07-24
Only trigger a frame build if scene properties have changed.

gw3583:
In many important use cases (e.g. video playback on youtube) the
display list doesn't change very often. Typically, the only
thing changing between frames are the natively decoded video
texture surfaces.

In these cases, we can save a significant amount of CPU usage
and battery power by not building a frame at all.

Previously, any call to set/update dynamic properties would
trigger a frame rebuild. However, it's often the case that
Gecko sends dynamic properties updates that are either empty
or exactly the same as the previous frame.

With this change, WR will compare the scene properties and
only trigger a frame rebuild if they have actually changed.

This is a partial step towards fixing #2917 - a couple of
other changes are also needed to completely avoid a frame
rebuiild in this use case.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2927)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -169,6 +169,97 @@ impl Document {
         !self.view.window_size.is_empty_or_negative()
     }
 
+    fn process_frame_msg(
+        &mut self,
+        message: FrameMsg,
+    ) -> DocumentOps {
+        match message {
+            FrameMsg::UpdateEpoch(pipeline_id, epoch) => {
+                self.current.scene.update_epoch(pipeline_id, epoch);
+
+                DocumentOps::nop()
+            }
+            FrameMsg::EnableFrameOutput(pipeline_id, enable) => {
+                if enable {
+                    self.output_pipelines.insert(pipeline_id);
+                } else {
+                    self.output_pipelines.remove(&pipeline_id);
+                }
+                DocumentOps::nop()
+            }
+            FrameMsg::Scroll(delta, cursor) => {
+                profile_scope!("Scroll");
+
+                let mut should_render = true;
+                let node_index = match self.hit_tester {
+                    Some(ref hit_tester) => {
+                        // Ideally we would call self.scroll_nearest_scrolling_ancestor here, but
+                        // we need have to avoid a double-borrow.
+                        let test = HitTest::new(None, cursor, HitTestFlags::empty());
+                        hit_tester.find_node_under_point(test)
+                    }
+                    None => {
+                        should_render = false;
+                        None
+                    }
+                };
+
+                let should_render =
+                    should_render &&
+                    self.scroll_nearest_scrolling_ancestor(delta, node_index) &&
+                    self.render_on_scroll == Some(true);
+                DocumentOps {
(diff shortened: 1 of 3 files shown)


gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=fcb64cd563bd463c4b86e0346b8d8c4fedbaa17f

webrender/src/render_backend.rs
```diff
@@ -602,12 +602,9 @@ impl RenderBackend {
     }
 
     fn process_frame_msg(
-        &mut self,
-        document_id: DocumentId,
+        doc: &mut Document,
```
kvark: nit: might as well just make it a method of `Document`

webrender/src/scene.rs
```diff
@@ -13,38 +13,72 @@ use std::sync::Arc;
...
 
     /// Add to the current property list for this display list.
     pub fn add_properties(&mut self, properties: DynamicProperties) {
-        for property in properties.transforms {
-            self.transform_properties
-                .insert(property.key.id, property.value);
-        }
+        let mut pending_properties = self.pending_properties
+            .take()
+            .unwrap_or(DynamicProperties::new());
```
kvark: nit: `unwrap_or_default` could be handy

webrender/src/scene.rs
```diff
@@ -13,38 +13,72 @@ use std::sync::Arc;
...
+    /// saving implications, allowing a frame build to be skipped
+    /// if the properties haven't changed in many cases.
+    pub fn flush_pending_updates(&mut self) -> bool {
+        let mut properties_changed = false;
 
-        for property in properties.floats {
-            self.float_properties
-                .insert(property.key.id, property.value);
+        if let Some(pending_properties) = self.pending_properties.take() {
+            if pending_properties != self.current_properties {
```
kvark: would it be worse to just keep the `properties_changed` flag and update it in `set_properties` and `update_properties`?

webrender_api/src/api.rs
```diff
@@ -1099,12 +1099,21 @@ pub struct PropertyValue<T> {
...
 /// can optionally be supplied to provide the current value of any
 /// animated properties.
-#[derive(Clone, Deserialize, Serialize, Debug)]
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
 pub struct DynamicProperties {
     pub transforms: Vec<PropertyValue<LayoutTransform>>,
     pub floats: Vec<PropertyValue<f32>>,
 }
 
+impl DynamicProperties {
```
kvark: nit: `[derive(Default)]` could replace this

webrender/src/render_backend.rs
gw3583: Done

webrender/src/scene.rs
gw3583: Done

webrender_api/src/api.rs
gw3583: Done

Issue #2885, opened 2018-07-12
In debug mode, webrender doesn't check for EXT_debug_marker before calling push_group_marker


fschutt:
The debug renderer seems to be active in debug mode by default, and it seems to call GpuMarker::new(), which in turn calls `pop_group_marker_ext`:

https://github.com/servo/webrender/blob/209c94babb87065ca339224133534b5f87f87a5a/webrender/src/device/query_gl.rs#L292

The problem is that it doesn't check if the extension necessary for GPU markers (I think it's `EXT_debug_marker`) is available first. Which leads to this:

```
Debug message with high or medium severity: `GL_INVALID_OPERATION in unsupported function called (unsupported extension or deprecated function?)`.
Please report this error: https://github.com/tomaka/glium/issues
Backtrace:
   1 - 0x5620d55fb7d5
         backtrace::backtrace::libunwind::trace at /home/felix/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.9/src/backtrace/libunwind.rs:53
         backtrace::backtrace::trace at /home/felix/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.9/src/backtrace/mod.rs:42
   2 - 0x5620d555f00b
         glium::context::default_debug_callback at /home/felix/.cargo/registry/src/github.com-1ecc6299db9ec823/glium-0.21.0/src/context/mod.rs:852
   3 - 0x5620d558f073
         core::ops::function::FnMut::call_mut at /checkout/src/libcore/ops/function.rs:146
   4 - 0x5620d5560296
         glium::context::init_debug_callback::callback_wrapper at /home/felix/.cargo/registry/src/github.com-1ecc6299db9ec823/glium-0.21.0/src/context/mod.rs:939
   5 - 0x7f638460047c
         <unknown> at <unknown>:??
   6 - 0x7f63845db981
         <unknown> at <unknown>:??
   7 - 0x5620d5504058
         gleam::ffi_gl::Gl::PopGroupMarkerEXT at /home/felix/Development/azul/target/debug/build/gleam-0a7e5e9d5b367460/out/gl_bindings.rs:2678
   8 - 0x5620d54e9624
         <gleam::gl::GlFns as gleam::gl::Gl>::pop_group_marker_ext at /home/felix/.cargo/registry/src/github.com-1ecc6299db9ec823/gleam-0.5.1/src/gl_fns.rs:1658
   9 - 0x5620d4c26552
[...]

PR #2935, opened 2018-07-28
Check for EXT_debug_marker

csherratt:
Fix for #2885 

Sorry novice at contributing here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2935)

### DIFF
--- a/webrender/src/device/query_gl.rs
+++ b/webrender/src/device/query_gl.rs
@@ -71,16 +71,18 @@ pub struct GpuFrameProfile<T> {
     samplers: QuerySet<GpuSampler<T>>,
     frame_id: FrameId,
     inside_frame: bool,
+    ext_debug_marker: bool
 }
 
 impl<T> GpuFrameProfile<T> {
-    fn new(gl: Rc<gl::Gl>) -> Self {
+    fn new(gl: Rc<gl::Gl>, ext_debug_marker: bool) -> Self {
         GpuFrameProfile {
             gl,
             timers: QuerySet::new(),
             samplers: QuerySet::new(),
             frame_id: FrameId::new(0),
             inside_frame: false,
+            ext_debug_marker
         }
     }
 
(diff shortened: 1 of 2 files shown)


webrender/src/device/query_gl.rs
```diff
@@ -71,16 +71,18 @@ pub struct GpuFrameProfile<T> {
     samplers: QuerySet<GpuSampler<T>>,
     frame_id: FrameId,
     inside_frame: bool,
+    ext_debug_marker: bool
```
kvark: github shows something off with indentation here

webrender/src/device/query_gl.rs
```diff
@@ -71,16 +71,18 @@ pub struct GpuFrameProfile<T> {
...
 impl<T> GpuFrameProfile<T> {
-    fn new(gl: Rc<gl::Gl>) -> Self {
+    fn new(gl: Rc<gl::Gl>, ext_debug_marker: bool) -> Self {
         GpuFrameProfile {
             gl,
             timers: QuerySet::new(),
             samplers: QuerySet::new(),
             frame_id: FrameId::new(0),
             inside_frame: false,
+            ext_debug_marker: ext_debug_marker
```
kvark: nit: just `ext_debug_marker,`

webrender/src/device/query_gl.rs
```diff
@@ -263,33 +267,40 @@ impl<T: NamedTag> GpuProfiler<T> {
...
     pub fn place_marker(&mut self, label: &str) {
-        GpuMarker::fire(&self.gl, label)
+        GpuMarker::fire(&self.gl, label, self.ext_debug_marker)
     }
 }
 
 #[must_use]
 pub struct GpuMarker {
     gl: Rc<gl::Gl>,
+    ext_debug_marker: bool
```
kvark: The whole struct is rather useless without this extension. If the users can't make a decision not to use it at all, we might just make it `gl: Option<Rc<gl::Gl>>` here instead of adding a flag

webrender/src/device/query_gl.rs
csherratt: I don't see anything wrong in my editor.

screenshot 2018-07-30_08-35-16

kvark:
Thank you!
@bors-servo r+

csherratt:
Hmm, http://feedback.wildfiregames.com/report/opengl/feature/GL_EXT_debug_marker

I don't know how accurate those stats are. But looks like this is only officially supported my mac os 10.

The other one you purpose seems to be supported pretty much everywhere, http://www.g-truc.net/doc/OpenGL%204%20Hardware%20Matrix.pdf so that makes sense to move to this.

PR #2936, opened 2018-07-28
Trim the render work done after a scene build

### LINKED ISSUE #2807: Reduce the amount of extra work needed for hit testing
#2797 enabled rendering on async scene rebuilds. AFAIK, hit-testing should only need the results of `build_layer_screen_rects_and_cull_layers` function, and don't actually care about the batches and render task graph. Thus, we can save a few cycles here.

cc @staktrace @nical @gw3583

staktrace:
Fixes #2807. I'm not totally sure that I'm including all the work that is necessary for hit-testing but it seems to pass my ad-hoc testing and CI. But still, please review with care to make sure I'm not violating assumptions or leaving behind inconsistent state by just doing the bits of the clip scroll tree update that don't touch the resource cache, gpu cache, and transform palette.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2936)

### DIFF
--- a/webrender/src/clip_node.rs
+++ b/webrender/src/clip_node.rs
@@ -30,14 +30,16 @@ impl ClipNode {
         &mut self,
         device_pixel_scale: DevicePixelScale,
         clip_store: &mut ClipStore,
-        render_context: &mut ClipRenderContext,
+        render_context: &mut Option<ClipRenderContext>,
         clip_chains: &mut [ClipChain],
         spatial_nodes: &[SpatialNode],
     ) {
         let clip_sources = clip_store.get_mut(self.clip_sources_index);
-        clip_sources.update(render_context.gpu_cache,
-                            render_context.resource_cache,
-                            device_pixel_scale);
+        if let &mut Some(ref mut context) = render_context {
+            clip_sources.update(context.gpu_cache,
+                                context.resource_cache,
+                                device_pixel_scale);
+        }
         let spatial_node = &spatial_nodes[clip_sources.spatial_node_index.0];
 
         let (screen_inner_rect, screen_outer_rect) = clip_sources.get_screen_bounds(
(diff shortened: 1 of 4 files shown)


gw3583:
Apart from the review comment above, I have a couple of questions:

* What are the situation(s) where we'd want to build the hit tester but not do a render anyway (for my own understanding, more than anything else). I vaguely recall complexities around needing to copy the hit-tester data at some point, but the details escape me.

* The bug being fixed refers to needing to call ```build_layer_screen_rects_and_cull_layers``` but not the batching and task generation of that code. However, it seems this patch only does the ```clip_scroll_tree.update```, and not the culling code. There's a couple subtleties here I'm unsure of:
  * During ```build_layer_screen_rects_and_cull_layers``` we can optionally update the local rect of a primitive. I *think* we only do this for ```Picture``` types, and picture primitives should never have an ```ItemTag``` associated with them - so this shouldn't affect hit testing.
  * We only update the visibility of a primitive during ```build_layer_screen_rects_and_cull_layers```. This includes checking whether a primitive is considered backface visible. Updating the clip-scroll tree may result in a change to whether a primitive is backface visible, but this patch would not update the hit-test code. I'm not actually sure what the specifications say regarding how hit testing interacts with backface visibility culling - is this relevant here?

If we *can* get away with just updating the clip scroll tree, that's great - it should be a decent win. However, if we need to also call ```build_layer_screen_rects_and_cull_layers``` but not do batching / task generation, this may be slightly more involved (and we might be better to consider refactoring how the visibility test code works, to make this simpler).

nical:
I rebased this PR in [this branch](https://github.com/nical/webrender/commit/c48215bf67ba22bfc077359a5786adc749b5900d). I'll do some testing before submitting the pull request.

nical:
Replaced by #3052

PR #2939, opened 2018-07-30
Do perspective division in hardware

### LINKED ISSUE #2932: Let the hardware do perspective division of transformed primitives
Related to #2272

Our current `write_transform_vertex` code does the perspective division manually, even if the attributes are supposed to be perspective interpolated:
```glsl
// Transform the current vertex to the world cpace.
vec4 world_pos = transform.m * vec4(local_pos, 0.0, 1.0);
// Convert the world positions to device pixel space.
vec2 device_pos = world_pos.xy / world_pos.w * uDevicePixelRatio;
```

The problem here is that the division result makes no sense for `world_pos.w <= 0.0` vertices, and we end up with broken rendering. The CPU side does near-plane splitting when calculating the bounding boxes, but we don't want to near-split all the transformed primitives before passing them to GPU, only the ones in `TransformStyle::Preserve3D` resulted plane splits.

The hardware already does proper frustum clipping after the vertex stage, so we should take advantage of it.

kvark:
Fixes #2932

TODO:
- [x] update reftests - commits/ccb1650afeceba1091df42f87c15babd48fb9158
- [x] run gecko try push - https://treeherder.mozilla.org/#/jobs?repo=try&revision=7c537505d671aa629232b5563d21d0c1b01ab2a6

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2939)


kvark:
@gw3583 it looks like some snapping logic is still not correct, which I'm going to fix. Otherwise, reftest failures are expected within a reasonable threshold, given that the shader transform logic is now different.

kvark:
Snapping is fixed now. Launched a new try push in https://treeherder.mozilla.org/#/jobs?repo=try&revision=63fdc4c26c5eb7674d32aaf5649ff23215bde56e

kvark:
@gw3583 reftests show a few intermittents but nothing wrong, as far as I can see. This should be ready to land.

PR #2941, opened 2018-07-31
Eagerly clean removed image out of the texture cache.

nical:
This is the first step towards improving the use of dirty rects with async blobs.

This PR adds the concept of manual and automatic eviction strategies. Automatically evicted texture cache entries correspond to what we currently do where the texture cache evicts items that haven't been request for a number of frames, while manually evicted items never get evicted until the resource cache explicitly marks them as unused with `TextureCache::mark_unused`.

The manual eviction policy will be needed for blob images so that we get the guarantees that what's in the cache doesn't get evicted while rasterization is happening asynchronously, which we need for dirty rects to stay valid.

Non-blob items will remain automatically evicted. This PR also makes us evict removed image keys faster for all types of texture cache entries which is good since an existing texture cache entry for a removed image key will never be used again (might as well free space up in the cache as soon we know we won't use it again).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2941)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -112,6 +112,8 @@ struct CacheEntry {
     eviction_notice: Option<EvictionNotice>,
     // The type of UV rect this entry specifies.
     uv_rect_kind: UvRectKind,
+    // If true the cache entry may be evicted if unused for a number of frames.
+    eviction: Eviction,
 }
 
 impl CacheEntry {
@@ -136,6 +138,7 @@ impl CacheEntry {
             uv_rect_handle: GpuCacheHandle::new(),
             eviction_notice: None,
             uv_rect_kind,
+            eviction: Eviction::Auto,
         }
     }
 
@@ -190,6 +193,14 @@ impl TextureCacheHandle {
     pub fn new() -> Self {
         TextureCacheHandle { entry: None }
     }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "capture", derive(Serialize))]
+#[cfg_attr(feature = "replay", derive(Deserialize))]
+pub enum Eviction {
+    Auto,
+    Manual,
 }
 
 // An eviction notice is a shared condition useful for detecting
@@ -410,6 +421,7 @@ impl TextureCache {
         gpu_cache: &mut GpuCache,
         eviction_notice: Option<&EvictionNotice>,
         uv_rect_kind: UvRectKind,
+        eviction: Eviction,
     ) {
         // Determine if we need to allocate texture cache memory
         // for this item. We need to reallocate if any of the following
@@ -462,6 +474,8 @@ impl TextureCache {
 
         // Upload the resource rect and texture array layer.
         entry.update_gpu_cache(gpu_cache);
+
+        entry.eviction = eviction;
 
         // Create an update command, which the render thread processes
         // to upload the new image data into the correct location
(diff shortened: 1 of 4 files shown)


webrender/src/texture_cache.rs
```diff
@@ -112,6 +112,8 @@ struct CacheEntry {
     eviction_notice: Option<EvictionNotice>,
     // The type of UV rect this entry specifies.
     uv_rect_kind: UvRectKind,
+    // If true the cache entry may be evicted if unused for a number of frames.
```
kvark: comments needs an update. Also, why is it not a doc comment?

webrender/src/texture_cache.rs
```diff
@@ -581,6 +595,17 @@ impl TextureCache {
          DeviceUintRect::new(origin, entry.size))
     }
 
+    pub fn mark_unused(&mut self, handle: &TextureCacheHandle) {
+        if let Some(ref handle) = handle.entry {
+            if let Some(entry) = self.entries.get_opt_mut(handle) {
+                // Set a very low last accessed frame to make it very likely that this entry
+                // will get cleaned up next time we try to expire entries.
+                entry.last_access = FrameId(0);
+                entry.eviction = Eviction::Auto;
```
kvark: given that `last_access` doesn't make sense for `Manual` anyway, can we make it a field of `Eviction::Auto`?

webrender/src/texture_cache.rs
nical: It's not a doc comment because the other comments were not doc comments either. Does it make a difference for non-public fields?

webrender/src/texture_cache.rs
nical: The `Eviction` enum was primarily meant to avoid adding a boolean parameter to the update function so I'd rather keep it this way.

webrender/src/texture_cache.rs
kvark: I secretly hope that we can (one day) generate a private documentation for a module/crate, where those will be useful. In any way, there isn't any downside of using doc comments here, but you aren't required to do the change since you were just following the old code.

webrender/src/texture_cache.rs
kvark: I don't understand. If we can get a better semantics, better use of types, then why neglect it?

nical: If you insist I can add another enum for the thing stored in here that has a last access time and keep this enum for the signature fo `TextureCache::update`. Where `Eviction` shows up most right now is as an argument in the various calls to `TextureCache::update` where the last access time member doesn't make sense (I think).

webrender/src/texture_cache.rs
nical: I don't mind doing the change either in light of your sercet hope. I'll update it.

nical:
> It sounds like these proposed changes may break that assumption? And if so, would that complicate things such as device reset etc?

Not necessarily: for memory pressures or device loss we can detect it and throw away the rendered blobs that only have a sub-rect available (that would fall back to synchronous rasterization on the next frame but look correct). That way we could still throw away manually evicted items if we want to.

If we don't have a way to maintain blobs in the cache, we can't rasterize sub-rects asynchronously, since doing so relies on the fact that the rest of the blob is in the cache and will still be there by the time we are done rasterizing.

webrender/src/texture_cache.rs
kvark: ok, fair enough. I can follow up with this myself if I see it feasible

kvark:
Given the concern raised, leaving for @gw3583 to merge.

gw3583:
OK, I think that sounds reasonable, as long as it's easy to detect those cases. I guess we'll just need to add a bit of code to the message that clears the caches to ensure that we hook up this logic to the blob code?

nical:
All we need is to throw away the rasterized blobs and the code that queries images during frame building will fallback to rasterizing the blobs it can't find synchronously on demand.

gw3583:
@nical Sounds good to me then!

PR #2942, opened 2018-07-31
Avoid transform copies where not necessary

kvark:
This should be harmless. Just using references and `Cow` where we did copies before. Those copies are sometimes difficult to track down in the profiler because they just make the functions heavier and not all the time manifest in separate system move calls.
cc @nical

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2942)


gw3583:
This looks good to me. It might be worth a try run, just in case, since some of those changes when detecting transform kind vs. preserves_2d etc can be subtle - but I'll leave that to your judgment :) r=me.

kvark:
A few intermittents, no blockers found.
@bors-servo r=gw3583

PR #2948, opened 2018-08-02
Only paint the dirty rect of non-tiled blobs.

nical:
Unless we settle with a small tile size we should follow up with the same optimization for tiled blobs.
This is built on top of #2941.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2948)

### DIFF
--- a/examples/blob.rs
+++ b/examples/blob.rs
@@ -17,6 +17,7 @@ use rayon::prelude::*;
 use std::collections::HashMap;
 use std::sync::Arc;
 use webrender::api::{self, DisplayListBuilder, DocumentId, PipelineId, RenderApi, Transaction};
+use webrender::api::{DeviceUintRect, DeviceUintPoint};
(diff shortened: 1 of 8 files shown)


nical:
try push with reftests green https://treeherder.mozilla.org/#/jobs?repo=try&revision=3ea9ba2961163832ab9b4279d0b97d0d46dd3c90

examples/blob.rs
```diff
@@ -97,7 +98,10 @@ fn render_blob(
 
     Ok(api::RasterizedBlobImage {
         data: Arc::new(texels),
-        size: descriptor.size,
+        rasterized_rect: DeviceUintRect {
```
kvark: nit: could use [from_size](https://docs.rs/euclid/0.19.0/euclid/struct.TypedRect.html#method.from_size)

PR #2950, opened 2018-08-03
Fix texture upload size and track it

### LINKED ISSUE #2930: Track number of bytes uploaded per frame
It would be nice to see this in the profiler hud. It can be an easy way to diagnose the source of performance problems that we run into.

kvark:
Fixes #2930
Also, apparently, we were uploading way more data then we needed. Say, if the client (Gecko) stores texture data in a 1GB memory buffer, and it wants to upload the left-top texel only, we'd fill up the PBO with all of that 1GB... But the good news is - this should be fixed with this PR.

~~TODO:~~ gecko try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=78f3185784677b34223ed2cce9cd2358f0c93209

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2950)

### DIFF
--- a/webrender/src/device/gl.rs
+++ b/webrender/src/device/gl.rs
@@ -2313,10 +2313,20 @@ impl<'a, T> TextureUploader<'a, T> {
         layer_index: i32,
         stride: Option<u32>,
         data: &[T],
-    ) {
+    ) -> usize {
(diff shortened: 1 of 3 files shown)


gw3583:
Oh, nice find! :+1: 

@bors-servo r+

PR #2947, opened 2018-08-01
Proper near plane splitting

kvark:
The PR is built on the shoulders of #2913, #2741, https://github.com/servo/euclid/pull/277, https://github.com/servo/euclid/pull/291, https://github.com/servo/plane-split/pull/12, and (last but not the least!) https://github.com/servo/plane-split/pull/15

It uses the new clipping semantics in `plane-split` crate uniformly for both bounding box computation and 3d plane splitting itself, ensuring that no incorrect perspective divisions are performed (and closing quite a few TODO entries). Also adds a small reftest for the latter.

Fixes #2272

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2947)

PR #2951, opened 2018-08-06
Clip chain refactor for local picture raster support.

gw3583:
There's a large number of related changes in this
patch. The overall goal is to make clip chain handling
not depend on screen rectangle generation. This
will make it much easier to implement local space
picture rasterization for perspective elements. In
addition to that, we unify how user / hierarchical and
primitive clip chain nodes are handled.

Clip chain instances are built on demand now, as
required by primitives. Right now, none of the
clip chain information is cached between primitives.
In general, now that the calculations are done in
local space, they are much faster - however we can
easily start to cache the common clip-chain state
(per raster coordinate system) if it ever shows up
in a profile. Additionally, since we only build
clip chains on demand, and clip chains for non-
visible primitives are not built, which is a
decent saving on many pages.

Remove ClipNode altogether. Instead, during flattening
we map directly from a ClipId to a ClipChainId
as appropriate. Clip chains are now created during
flattening and the links remain constant. It's only
during construction of a clip chain instance that we
include and exclude parts of the specified clip
chain. This unifies how we handle user defined clip
chains and legacy hierarchical clip chains. At the
same time, unify how the optional primitive specific
complex clip is handled with clip chains. Now, if
a primitive has a complex clip, it's added as a
normal clip chain node, and linked to the parent
clip chain that this primitive would otherwise
be attached to.

Primitive runs now only depend on the spatial node.
This results in far fewer primitive run breaks.

ClipWorkItems are now stored as a single contiguous
array in the ClipStore. To achieve this they are
stored as an index buffer, and a clip node range
is simply an (index, count). These are a lot
cheaper to pass to clip masks, and also require
fewer heap allocations.

Port hit testing to use the same clip chain data
structures that the main clip code uses, now that
the hierarchical and user defined clip chains are
handled in a uniform way.

Remove the need for Arc in clip chain code, since
clip chains are now stored in a contiguous index
buffer style structure.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2951)


gw3583:
This still needs some more work, specifically:

- [ ] Several test failures on a try run.
- [ ] More code comments / documentation.
- [ ] Code tidy up.

Anecdotally it seems to be quite a bit faster on the sites I've tested against than previously, although I haven't done detailed performance numbers yet.

It's probably not quite ready for review yet, but if anyone wants to start taking a look at it, go for it :+1:

gw3583:
If anyone has any general comments on the approach taken, and anything obvious missing, please let me know :)

gw3583:
I added some poorly drawn ASCII diagrams to try and describe how the data structures fit together (https://github.com/servo/webrender/pull/2951/commits/a65f56424d6d7108c6ae2db0a65afa170da8058f#diff-97cd013eed8f4591bd7659faf4352fd2R23).

I have fixed the test failues in wrench, and _most_ of the test failures in gecko try.

Tomorrow I will try to resolve the remaining failures in gecko try, which are mostly 3d transform related.

gw3583:
Try run is looking very green now - just a few orange items with unrelated intermittents.
https://treeherder.mozilla.org/#/jobs?repo=try&revision=e084e4749dbb158bc76efcb504db8e10173b2e1d

r? @kvark 
cc @mrobinson

gw3583:
@kvark Thanks for the review! Those naming suggestions are much better than I had :+1:

gw3583:
Once this lands we should be on the lookout for any regressions in number of clip masks we draw, or any hit-test issues. I wouldn't be surprised to have a few issues like that which need to be fixed up :fearful:

gw3583:
@bors-servo r=kvark

(is bors stuck?)

gw3583:
Hmmm, on subsequent tries I'm seeing failures, but when I revert those back to this patch, I'm still seeing the same failures - but those same failures were not occurring on the try run above. :confused: Investigating now...

gw3583:
Hmm, and it's actually a single test that is showing up in multiple places due to being run on each platform - ```transform-3d/1035611-1.html```.

gw3583:
Oh, I think I see what has happened. Ugh.

* #2947 fixed ```1035611-1.html``` and made it pass.
* This patch re-breaks that test.
* My try run above was on m-c before the WR update landed, and so didn't have updated reftest annotations.

Investigating now to see why this patch breaks that test.

sotaroikeda:
By [Bug 1483610 Comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1483610#c5), the change might cause huge gecko performance regression.

PR #2952, opened 2018-08-07
Submit transaction with frame output enabled before transaction with …

hgallagher1993:
…DL in frame_output.rs

Fix for #2820 and closed #2866 in favour of this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2952)


kvark:
Thank you!
@bors-servo r+

Issue #2955, opened 2018-08-08
WebRender some times tries to allocate huge render task surfaces


nical:
#2916 fixed one of the possible ways we can run into huge render tasks but there appear to be others.

I see [4 crashes in crash-stats](https://crash-stats.mozilla.com/signature/?signature=webrender%3A%3Arenderer%3A%3ARenderer%3A%3Adraw_tile_frame&date=%3E%3D2018-08-01T13%3A59%3A18.000Z&date=%3C2018-08-08T13%3A59%3A18.000Z&_columns=date&_columns=product&_columns=version&_columns=build_id&_columns=platform&_columns=reason&_columns=address&_columns=install_time&_sort=-date&page=1#reports) in the last seven days, they fail when initializing the texture in the gl context which logs

> [0][GFX1-]: Attempting to allocate a texture of size 20008x20008 above the limit, trimming (t=22.2719)

in the critical error section.

kvark:
I believe https://phabricator.services.mozilla.com/D16560 put the last nail into this coffin ;)

PR #2956, opened 2018-08-08
Add a sanity check for render task sizes

nical:
This PR moves the crash happening when failing to allocate large render task textures to the place where the size is assigned in the hope that the stack trace will be more actionable. This will make it easier to investigate #2955.

I'm not sure whether the explicit `panic!()` is needed or if the `error!()` macro already panics.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2956)

### DIFF
--- a/webrender/src/render_task.rs
+++ b/webrender/src/render_task.rs
@@ -32,9 +32,18 @@ use tiling::{RenderTargetKind};
 #[cfg(feature = "pathfinder")]
 use webrender_api::DevicePixel;
 
+const RENDER_TASK_SIZE_SANITY_CHECK: i32 = 16000;
 const FLOATS_PER_RENDER_TASK_INFO: usize = 8;
 pub const MAX_BLUR_STD_DEVIATION: f32 = 4.0;
 pub const MIN_DOWNSCALING_RT_SIZE: i32 = 128;
+
+fn render_task_sanity_check(size: &DeviceIntSize) {
+    if size.width > RENDER_TASK_SIZE_SANITY_CHECK ||
+        size.height > RENDER_TASK_SIZE_SANITY_CHECK {
+        error!("Attempting to create a render task of size {}x{}", size.width, size.height);
+        panic!();
+    }
+}
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
@@ -335,6 +344,7 @@ impl RenderTask {
     }
 
     pub fn new_readback(screen_rect: DeviceIntRect) -> Self {
+        render_task_sanity_check(&screen_rect.size);
         RenderTask {
             children: Vec::new(),
             location: RenderTaskLocation::Dynamic(None, Some(screen_rect.size)),
@@ -369,6 +379,8 @@ impl RenderTask {
 
         size.width += padding.horizontal();
         size.height += padding.vertical();
+
+        render_task_sanity_check(&size);
 
         RenderTask {
             children,
@@ -458,6 +470,8 @@ impl RenderTask {
             }
         }
 
+        render_task_sanity_check(&outer_rect.size);
+
         RenderTask {
             children,
             location: RenderTaskLocation::Dynamic(None, Some(outer_rect.size)),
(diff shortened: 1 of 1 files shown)


webrender/src/render_task.rs
```diff
@@ -32,10 +32,19 @@ use tiling::{RenderTargetKind};
...
+const RENDER_TASK_SIZE_SANITY_CHECK: i32 = 16000;
 const FLOATS_PER_RENDER_TASK_INFO: usize = 8;
 pub const MAX_BLUR_STD_DEVIATION: f32 = 4.0;
 pub const MIN_DOWNSCALING_RT_SIZE: i32 = 128;
 
+fn render_task_sanity_check(size: &DeviceIntSize) {
+    if size.width > RENDER_TASK_SIZE_SANITY_CHECK ||
+        size.height > RENDER_TASK_SIZE_SANITY_CHECK {
+        error!("Attempting to create a render task of size {}x{}", size.width, size.height);
+        panic!();
```
kvark: let's just have this message in `panic!`

webrender/src/render_task.rs
```diff
@@ -458,6 +470,8 @@ impl RenderTask {
             }
         }
 
+        render_task_sanity_check(&outer_rect.size);
```
kvark: would it be easier (and more robust) if we just had a constructor like `RenderTask::new_dynamic(size, kind, clear_mode, children)` where the check is done?

webrender/src/render_task.rs
nical: The reason I put it in error!() is that it gets logged in a special gfxCriticalError thing which I can search for in crash-stats. I don't know how to search through panic messages in crash-stats although it might be possible.

webrender/src/render_task.rs
nical: I have a mild preference towards keeping the struct initialization syntax instead of using `::new` methods when construction is simple because the latter isn't very easy to read due to the lack of named arguments.
My hope is that within a few weeks of this sanity check being in the wild we'll have data pointing us to the missing places where we need to be careful about render task sizes and we can then remove the check (or replace it with something a bit better where we check against the real maximum size or something that makes sense on a per-type-of-item basis).

kvark: > I have a mild preference towards keeping the struct initialization syntax instead of using ::new methods when construction is simple

I agree! In my book the reasoning no longer stands if there is any logic that needs to be done though, which is the case here.

> and we can then remove the check

I don't see why we'd want to remove the check ever. Better put it once and forever in a nice spot to protect us against eternal evils of infinite render tasks.

nical: My own book has it that it's not worth losing the readability for a single sanity check function call that is only there for debugging purposes, but that's not really worth arguing over, I'll add the constructor.

kvark: I don't understand where "losing the readability" comes from. What am I missing?

kvark: oh, I was misreading it as "reliability", silly me

gw3583:
@nical @kvark This explicit panic occurs during the crash tests in CI. It may not be the same as the crash we're looking for, but I imagine it should be an easy fix:

```
thread 'WRRenderBackend#3' panicked at 'explicit panic', gfx/webrender/src/render_task.rs:43:9
TEST-UNEXPECTED-FAIL | file:///Users/cltbld/tasks/task_1533871261/build/tests/reftest/tests/layout/base/crashtests/265973-1.html | application terminated with exit code 1
REFTEST PROCESS-CRASH | file:///Users/cltbld/tasks/task_1533871261/build/tests/reftest/tests/layout/base/crashtests/265973-1.html | application crashed [@ mozalloc_abort]
Return code: 1
No suite end message was emitted by this harness.
# TBPL FAILURE #
The reftest suite: crashtest ran with return status: FAILURE
[taskcluster:error] exit status 2
```

nical:
Link ot the failing try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=b313719e75e0e1f9169cd8f6c38cc227d7d21cb0&selectedJob=193217773

The crashtest is:

```
<html>
<head>
</head>

<body>
<iframe style="border-top-width: 31378748; border-bottom-right-radius: 23895784; ">
</body>
</html>
```
Looking into it.

PR #2961, opened 2018-08-09
Embed the PicturePrimitive inside BrushPrimitive.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2961)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -6,7 +6,7 @@ use api::{AlphaType, BorderRadius, BuiltDisplayList, ClipMode, ColorF};
 use api::{DeviceIntRect, DeviceIntSize, DevicePixelScale, ExtendMode};
 use api::{FilterOp, GlyphInstance, GradientStop, ImageKey, ImageRendering, ItemRange, ItemTag, TileOffset};
 use api::{GlyphRasterSpace, LayoutPoint, LayoutRect, LayoutSize, LayoutToWorldTransform, LayoutVector2D};
-use api::{PipelineId, PremultipliedColorF, PropertyBinding, Shadow, YuvColorSpace, YuvFormat, DeviceIntSideOffsets};
+use api::{PremultipliedColorF, PropertyBinding, Shadow, YuvColorSpace, YuvFormat, DeviceIntSideOffsets};
 use api::{BorderWidths, BoxShadowClipMode, LayoutToWorldScale, NormalBorder};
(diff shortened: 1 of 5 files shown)


webrender/src/prim_store.rs
```diff
@@ -1241,36 +1226,26 @@ impl PrimitiveStore {
...
-        let picture = PicturePrimitive::new_image(
-            PictureId(self.next_picture_id),
-            composite_mode,
-            is_in_3d_context,
-            pipeline_id,
-            reference_frame_index,
-            frame_output_pipeline_id,
-            apply_local_clip_rect,
-        );
+    pub fn get_pic(&self, index: PrimitiveIndex) -> &PicturePrimitive {
```
kvark: This looks like a regression in our internal APIs. Where we had separate type for picture index, we were able to associate it with a picture, now we pass more `PrimitiveIndex` around. I don't see a way around it if you want to move the picture under primitive metadata, but just pointing out at the issue.

webrender/src/prim_store.rs
```diff
@@ -1241,36 +1226,26 @@ impl PrimitiveStore {
...
+        let metadata = &self.cpu_metadata[index.0];
+        let brush = &self.cpu_brushes[metadata.cpu_prim_index.0];
+        match brush.kind {
+            BrushKind::Picture(ref pic) => pic,
+            _ => unreachable!("bug: not a picture!"),
```
kvark: nit: I think if there is any doubt that this is reachable, then `panic!` is more appropriate :)

webrender/src/prim_store.rs
gw3583: I agree - this is an internal regression. However, I think it should be temporary only - I'd like to embed more of the prim store types directly into the runs struct in the Picture (see rationale above) which should then remove (most of?) the indexing that occurs here, if all goes according to plan.

webrender/src/prim_store.rs
gw3583: Done

gw3583:
Fixed nit and rebased on top of the review comments for #2951. I'll rebase again once that is merged.

gw3583:
I did a fresh try run after rebasing, that looks green apart from unrelated intermittents:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=8d62530193f44ce1f131d85ce5791f66d6f9abec

@bors-servo r=kvark

Issue #2958, opened 2018-08-08
SVGs and PNGs that don't have any alpha should go into the opaque pass when possible


pcwalton:
This is hurting performance on my reveal.js presentation, because I have some big images that are opaque and it's causing overdraw.

kvark:
Isn't this a problem downstream (in Gecko/Servo)? The `ImageDescriptor` struct passed to `add_image` has [is_opaque](https://github.com/servo/webrender/blob/0634d020dd50ed8d93d99e51a7d057bb1e2ec2c9/webrender_api/src/image.rs#L83) field which determines if we can use the opaque path for rendering it.

pcwalton:
OK, I diagnosed this problem with the shiny new "show overdraw" view I wrote and found that the issue is localized to SVGs.

pcwalton:
This also happens with RGBA PNGs that don't actually have any alpha pixels in them.

pcwalton:
Have a Gecko patch for this for PNG, just need to clean it up.

kvark:
@pcwalton any reason to keep this open if this is a purely downstream issue?

pcwalton:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1485534 so we can close this.

PR #2963, opened 2018-08-10
Add a debugging feature to visualize overdraw.

pcwalton:
Lighter shades of orange indicate more overdraw.

This has already helped to diagnose a few issues (e.g. #2958).

r? @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2963)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -212,17 +212,18 @@ impl BatchKind {
 bitflags! {
     #[derive(Default)]
     pub struct DebugFlags: u32 {
-        const PROFILER_DBG      = 1 << 0;
-        const RENDER_TARGET_DBG = 1 << 1;
-        const TEXTURE_CACHE_DBG = 1 << 2;
-        const GPU_TIME_QUERIES  = 1 << 3;
-        const GPU_SAMPLE_QUERIES= 1 << 4;
-        const DISABLE_BATCHING  = 1 << 5;
-        const EPOCHS            = 1 << 6;
-        const COMPACT_PROFILER  = 1 << 7;
-        const ECHO_DRIVER_MESSAGES = 1 << 8;
-        const NEW_FRAME_INDICATOR = 1 << 9;
-        const NEW_SCENE_INDICATOR = 1 << 10;
+        const PROFILER_DBG          = 1 << 0;
+        const RENDER_TARGET_DBG     = 1 << 1;
+        const TEXTURE_CACHE_DBG     = 1 << 2;
+        const GPU_TIME_QUERIES      = 1 << 3;
+        const GPU_SAMPLE_QUERIES    = 1 << 4;
+        const DISABLE_BATCHING      = 1 << 5;
+        const EPOCHS                = 1 << 6;
+        const COMPACT_PROFILER      = 1 << 7;
+        const ECHO_DRIVER_MESSAGES  = 1 << 8;
+        const NEW_FRAME_INDICATOR   = 1 << 9;
+        const NEW_SCENE_INDICATOR   = 1 << 10;
+        const SHOW_OVERDRAW         = 1 << 11;
     }
 }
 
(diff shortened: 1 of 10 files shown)


gw3583:
Looks like there are some shader validation failures on CI:

https://ci.appveyor.com/project/servo/webrender/build/1.0.10141#L304

It might just be related to needing to pass the right defines to the validation code, if they've changed.

gw3583:
We'll need to resolve the shader validation errors on CI above. If we instead added an ```#ifdef``` in ```brush.glsl``` we could probably select the debug variant of each normal brush shader. This might simplify some of the complexity involved with the ```vecs_per_brush``` code?

We should do a try run too, to make sure that nothing breaks with any of the weird shader combinations. I can do that on Monday if you're not sure how to do a gecko try.

gw3583:
(if you think it's simpler to leave as is, that's fine too - the above is just what came to mind when looking through the code).

kvark:
A screnshot would be handy here

pcwalton:
@gw3583 Updated patch per your suggestion. This is way simpler :)

@kvark 

Here's a before and after:

screen shot 2018-08-10 at 12 01 47 pm

screen shot 2018-08-10 at 12 01 51 pm

webrender/src/renderer.rs
```diff
@@ -4013,6 +4020,47 @@ impl Renderer {
         }
         self.device.end_frame();
     }
+
+    // Sets the blend mode. Blend is unconditionally set if the "show overdraw" debugging mode is
+    // enabled.
+    fn set_blend(&self, mut blend: bool) {
+        if !self.debug_flags.contains(DebugFlags::SHOW_OVERDRAW) {
```
kvark: shouldn't this be the other way around? `if self.debug_flags.contains(DebugFlags::SHOW_OVERDRAW) { blend = true; }`

webrender/src/renderer.rs
```diff
@@ -4013,6 +4020,47 @@ impl Renderer {
         }
         self.device.end_frame();
     }
+
+    // Sets the blend mode. Blend is unconditionally set if the "show overdraw" debugging mode is
```
kvark: shouldn't it only affect the final (screen) pass?

webrender/src/shade.rs
```diff
@@ -403,13 +404,8 @@ pub struct Shaders {
...
     // Brush shaders
-    brush_solid: BrushShader,
-    brush_image: Vec<Option<BrushShader>>,
-    brush_blend: BrushShader,
-    brush_mix_blend: BrushShader,
-    brush_yuv_image: Vec<Option<BrushShader>>,
-    brush_radial_gradient: BrushShader,
-    brush_linear_gradient: BrushShader,
+    brush: BrushShaders,
+    brush_debug_overdraw: BrushShaders,
```
kvark: a brush shader struct contains 3 versions:
```rust
struct BrushShader {
    opaque: LazilyCompiledShader,
    alpha: LazilyCompiledShader,
    dual_source: Option<LazilyCompiledShader>,
}
```
Would it make sense to add a 4th for the debug overdraw instead of duplicating the whole thing?

webrender/res/brush_debug_overdraw.glsl
```diff
@@ -0,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
kvark: is this shader still used?

webrender/src/renderer.rs
pcwalton: Doh!

webrender/src/renderer.rs
pcwalton: Could you elaborate?

pcwalton:
Comments addressed.

webrender/src/renderer.rs
kvark: We render a frame in multiple passes. All but the last one go into off-screen targets. Since we are only displaying the screen overdraw, we don't want the blending overrides to affect rendering in any other passes than the last one.

pcwalton:
Comment addressed.

webrender/src/renderer.rs
pcwalton: OK, changed it.

webrender/src/renderer.rs
```diff
@@ -2304,7 +2308,7 @@ impl Renderer {
 
             self.device.disable_scissor();
             self.device.disable_depth();
-            self.device.set_blend(false);
+            self.set_blend(false, true);
```
kvark: uh, that's a bit confusing when reading the code.
can have an `enum TargetKind` or something like this?

kvark:
Wonderful, thank you!
@bors-servo r+

PR #2964, opened 2018-08-10
Ensure that non-local clips are defaulted to be in the clip mask.

gw3583:
If we can't calculate a local space rect for the clip, due to
perspective weirdness, just assume that it should be in the
clip mask, for now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2964)


gw3583:
This fixes the issue discussed in https://github.com/servo/webrender/pull/2951/comment#issuecomment-411959968.

Try run looks good now:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=bcadc346ed0d6f5248e4b9866672b4cc8e69d1eb

r? @kvark

kvark:
>  just assume that it should be in the clip mask, for now.

Would be good to have some sort of an assert or at least a reminder to do that

@bors-servo r+

PR #2965, opened 2018-08-10
Embed PrimitiveMetadata inside Primitive struct.

gw3583:
This is another small step to moving primitives to be stored inside Picture structs, which will simplify some of the picture caching work.

While making that change, I also tidied up some of the prim_store code by moving several of the methods into the Primitive struct. I've done each of these as a separate commit to make it easier to review.

There's no functional changes here - just moving code around to make the picture caching work easier. There's also quite a few array indexing operations removed, which might be a small performance win.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2965)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -17,9 +17,9 @@ use internal_types::{FastHashMap, SavedTargetIndex, SourceTexture};
 use picture::{PictureCompositeMode, PicturePrimitive, PictureSurface};
 use plane_split::{BspSplitter, Clipper, Polygon, Splitter};
 use prim_store::{BrushKind, BrushPrimitive, BrushSegmentTaskId, DeferredResolve};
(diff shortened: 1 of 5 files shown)


gw3583:
r? @kvark 

This *looks* like a big PR, but most of it is just moving methods around as a result of the first commit. I tried to separate each method change into a single commit to make it easier to review.

Try run here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9ee64481c801ff2df8b85fa01e915a598fac9e13

There's a few orange items, but I think they are all unrelated - probably best to take a quick look to make sure though.

kvark:
@gw3583 feel free to r=me if having no desire to address the nits :)

gw3583:
Addressed all comments. Merging now since this is prone to rebase issues.

@bors-servo r=kvark

PR #2968, opened 2018-08-13
Optimize how local rects are calculated during culling.

gw3583:
Apply similar logic to the clip code to minimize the
amount of work that is needed when building local
rects for a picture. This change is quite a
significant optimization - in most cases it halves
the amount of work done here, by skipping the extra
rect calculation for non-3d rendering context
pictures. It also saves some rect x matrix transforms
in various cases.

This is a bit more preparation work for rasterizing
pictures in difference coordinate spaces. There are a
few other miscellaneous improvements too:

 * Rename some fields to be more appropriate.
 * Reduce number of matrix inverse calls.
 * Remove reference_frame_stack, which is no longer
   required.
 * Remove BuiltDisplayList from PrimitiveContext, which
   will be helpful in future patches.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2968)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -670,7 +670,7 @@ impl AlphaBatchBuilder {
                             // Push into parent plane splitter.
                             debug_assert!(picture.surface.is_some());
                             let transform = &ctx.transforms
-                                .get_transform(picture.reference_frame_index);
+                                .get_transform(picture.original_spatial_node_index);
 
                             match transform.transform_kind {
                                 TransformedRectKind::AxisAligned => {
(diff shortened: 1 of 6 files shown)


gw3583:
r? @kvark 

Try run (still a few pending jobs, but looks fine so far):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=939f56d0f0d0bc7f16d7fbd6980a91c87ec1fc54

gw3583:
Addressed review comments, thanks!

@bors-servo r=kvark

PR #2969, opened 2018-08-13
support color masking of images

lsalzman:
This is to help support the necessary fixes for Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1479196

Tofu glyphs will be drawn out of an atlas, and to avoid the need to have to generate a new atlas for every single color or opacity combination, this just allows push_image to specify a color which can then be used for masking or opacity.

Most of the plumbing is already there, just some parameters need to be added to support this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2969)

### DIFF
--- a/examples/blob.rs
+++ b/examples/blob.rs
@@ -17,7 +17,7 @@ use rayon::prelude::*;
 use std::collections::HashMap;
 use std::sync::Arc;
 use webrender::api::{self, DisplayListBuilder, DocumentId, PipelineId, RenderApi, Transaction};
-use webrender::api::{DeviceUintRect, DeviceUintPoint};
+use webrender::api::{ColorF, DeviceUintRect, DeviceUintPoint};
 
 // This example shows how to implement a very basic BlobImageHandler that can only render
 // a checkerboard pattern.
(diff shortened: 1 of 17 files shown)


gw3583:
This basically looks good, but I'm somewhat confused about the shader color mode change on some of the batching code. If I'm reading it correctly it looks like it *was* wrong previously, and is now right? But if so, how come we don't have test failures from those? I feel like I'm mis-reading part of the patch :)

We'll also need a try run before merging.

gw3583:
@lsalzman Ah that makes sense - so it *was* broken, but not in any user-visible way previously :hankey:

gw3583:
r=me once we have a try result posted here.

gw3583:
That looks like a lot of orange on the try - but it's also not the try configuration I normally use, so maybe those are expected.

I leave it to your wise judgment :)

@bors-servo delegate+

lsalzman:
Cleaner try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=d453838d0586955f94608413a8533b6770f708be

PR #2973, opened 2018-08-16
Fix bug where brush mask kind changes during scrolling.

gw3583:
Calculate the brush clip mask kind during clip chain instance
building, instead of during segment building.

If this changes during a scroll of the same display list, detect
that and rerun the logic that assigns the clip mask kind to
each segment.

This doesn't actually re-segment, just updates the clip mask kind,
which is fine since the segments are based only on clips in the
same spatial node (so unable to change relative to each other
during scrolling).

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1483568.

Unfortunately we don't have enough scroll testing infrastructure
in wrench to add a regression test for this yet.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2973)


gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=5e8f37dae9ce5517bf9ed44aa6bbfd0174a67ea9

r? @kvark

PR #2974, opened 2018-08-16
Simplify determining if a batch can be merged / needs scissor.

gw3583:
Previously, we checked if the bounding box of any child primitives
extended outside the rectangle of the allocated task rect. However,
there is a simpler way to calculate this.

If the allocated size of the render task is >= the unclipped size
of the picture bounding rect, no scissor is needed, since we know
that local clip rects will take care of ensuring nothing is
drawn outside the task boundary.

This is an optimization, but the main benefit is removing one more
piece of code that relies on knowledge of screen / device rects,
which simplifies the ongoing work to be able to rasterize in other
coordinate systems.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2974)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -263,7 +263,6 @@ impl OpaqueBatchList {
 pub struct BatchList {
     pub alpha_batch_list: AlphaBatchList,
     pub opaque_batch_list: OpaqueBatchList,
-    pub combined_bounding_rect: DeviceIntRect,
 }
 
 impl BatchList {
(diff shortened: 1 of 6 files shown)


gw3583:
Also removed support for dynamic tasks with an unknown size, since that's no longer used.

r? @kvark 

Try run here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=31d78e6ce4d91bfedf2716794b5e9656714f285d

Still has a few jobs to finish, but looks good so far:
```R1``` and ```R8``` are new PASSes from a previous PR.
```R5```, ```Wd```, ```wpt3```, ```mda2``` appear to be unrelated intermittents.

gw3583:
Fixed up a compile error in the pathfinder feature code path.

PR #2975, opened 2018-08-16
Simplify how local rects are accumulated for 3d contexts.

gw3583:
Pictures in a 3d rendering context are re-parented during
flattening, to ensure correct ordering during plane splitting
when un-transformed content is present.

Previously, we determined the spatial node index based on
the re-parented hierarchy, and this complicated local rect
calculation for items inside a 3d rendering context. For
these, we needed to retain the original spatial node index
and do extra calculations.

Now, we determine the correct spatial node index during
flattening, and store it inside each primitive directly
(rather than via the primitive runs inside a picture).

This greatly simplifies the process of accumulating a
local rect for a picture as we recurse through the
picture tree - allowing pictures inside a 3d context
to be processed just as any normal primitives are.

This simplifies some upcoming work to rasterize pictures
in local space, and is also a small optimization that
reduces the number of local rects we need to build and
transform.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2975)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -458,12 +458,8 @@ impl AlphaBatchBuilder {
 
         // Add each run in this picture to the batch.
         for run in &pic.runs {
-            let transform_id = ctx
-                .transforms
-                .get_id(run.spatial_node_index);
             self.add_run_to_batch(
                 run,
-                transform_id,
                 ctx,
                 gpu_cache,
                 render_tasks,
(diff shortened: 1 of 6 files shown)


gw3583:
r? @kvark 

Try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=79dabb3040fc72e6dbc73a0069d0b1ff1658b99c

Still has a few jobs to finish, but looks good so far.
```R1``` and ```R8``` are PASS results from a previous PR.
```Wr1``` and ```wpt3``` are unrelated intermittent failures.

PR #2977, opened 2018-08-20
Use get_relative_transform for clip <-> prim space conversions.

gw3583:
This expands the functionality in get_relative_transform to work
for both primitive and clip space conversions.

This code path is not super efficient (it could be improved), but
on the majority of sites it never runs, and even on cases where it
does run, the coordinate system tree is typically very shallow. We
can optimize this function if it ever shows up in a profile.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2977)


gw3583:
r? @kvark 

Try run here (looks good I think, just a few unrelated intermittents):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=ffd9a4eb3c27b46449058123fec949ff4d3cdea1

I changed the expected result for one of the wrench tests - it is an incorrect test that came from Servo, and the item should actually be clipped in this case.

kvark:
Thank you!
@bors-servo r+

PR #2979, opened 2018-08-21
move font addition/deletion requests off the main thread

lsalzman:
This is to resolve downstream Gecko issue https://bugzilla.mozilla.org/show_bug.cgi?id=1484366

The main insight here is that before any FontContext is accessed, its mutex is locked first. If we want to do font addition/deletion off the main thread, we don't actually have to wait for the entire font addition/deletion to complete. We just have to lock all the font contexts to which the requests are dispatched, so that no other thread can come in and access them while we're still processing the requests.  We can then proceed to let the font processing happen in parallel with subsequent work. In the worst case, a subsequent FontContext access may block, but that just puts us back to the current status quo, which should be rare.

I did some profiling of various threading strategies for doing this in parallel. It seems that kicking off a bunch of jobs for each FontContext can be reliably slower for quicker loading fonts, as the thread overhead starts adding up. The cheapest strategy was just to kick off one thread, lock all contexts on that thread, signal the condition, then do all subsequent processing in that same thread. 

As discussed with Glenn offline, In the future we probably want to move towards some sort of message queue, but that would be a substantial restructuring of FontContext, whereas this will at least be a quick and easy stop-gap measure to hide the latency of adding/removing fonts from the main thread.

### DIFF
--- a/webrender/src/glyph_rasterizer/mod.rs
+++ b/webrender/src/glyph_rasterizer/mod.rs
@@ -15,7 +15,7 @@ use rayon::ThreadPool;
 use std::cmp;
 use std::hash::{Hash, Hasher};
 use std::mem;
-use std::sync::{Arc, Mutex, MutexGuard};
+use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 use std::sync::mpsc::{channel, Receiver, Sender};
 
 #[cfg(feature = "pathfinder")]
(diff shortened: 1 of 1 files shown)


gw3583:
Apologies, I didn't get to this today. Will look at it first thing tomorrow unless someone else gets there first.

PR #2980, opened 2018-08-22
Correctness and optimizations for clips in different coord systems.

gw3583:
For clips that are in a different coordinate system, it's not
generally feasible to get a correct mapping into a different
coordinate system - so handle these clips in world space. For
now, the world space is actually screen space. However, it is
trivial to modify this so that the world space is the coordinate
system that we are rasterizing this picture in.

As an optimization, if we encounter a primitive that is large,
and has clips from a different coord system, but no local clips,
then segment it into a uniform grid. This allows large primitives
that would not otherwise be segmented to opt in to segmenting,
which can significantly reduce the size of the allocated clip
masks for that primitive.

As a further optimization, we are now able to generate a custom
clip chain instance for each segment, rather than using the
clip chain from the primitive. This allows us to build a more
optimized clip chain, based on the segment rectangle, which can
often remove redundant clips from the mask for each segment, or
even determine that this segment is not affected by any clips
at all.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2980)


gw3583:
r? @kvark

Try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a9d93837c673748ce01d0f45ccfc150d5f4fb036

Fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1483610

Some explanation / double checking is warranted for the try run:

```[2] [wpt8] [wpt6]``` - unrelated intermittent

```[clipping-4-canvas.html] [R2]``` max difference 8/1 - marked fuzzy on several other backends - should fuzz for WR too.

```[1081185-1.html] [R8 + R2]``` This was *fixed* by #2974, and then marked as fixed during the WR update. However, that fix was actually incorrect and just masking the existing bug that causes the failure. Then, #2975 merged overnight, which fixes the bug in #2974 but re-exposes the bug which causes this test to fail. This patch doesn't actually alter the result of this test - it occurs on master due to the above reason. We should mark this test as FAIL again during the next update - it will be fixed by some of the follow up work to these current patch sets.

gw3583:
No idea what all those purple items are on the try or why it seems to be triggering dozens of re-runs by itself?

gw3583:
Well that was weird - the try job looked like it was in some weird loop where it just kept automatically respawning dozens of jobs. I cancelled the try run - I think all the important tests had completed?

sotaroikeda:
Sorry, I added  a lot of talos tasks to check if talos performance is improved. To check performance, we need to run multiple talos tasks. The purple items often happens when many linux talos tasks are added.

gw3583:
@sotaroikeda Ah, that makes sense, thanks for the explanation! :+1:

gw3583:
Added another try job which has the same results but I didn't cancel the run:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=741d3c12cd7cb29e864fda2b4837a57fdde407d5

gw3583:
Added a follow up commit that has a few isolated parts from the patch I'm working on locally. The extra commit shouldn't have any noticeable effect, but I pushed a try run just in case:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=91b0112bfa8957e12d9522a3612d3e4128f44a06

gw3583:
@kvark Thanks, addressed those comments, I think. Once you're happy with the changes, I'll do another try run before merging, just to make sure I didn't break anything with those changes.

gw3583:
Pending try with the current review comments addressed:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=f0ca06207944b6c45f84cd3b6bd15b7ed07eb464

gw3583:
Latest try run looks good.

kvark:
Those R8 and R2s are expected failures?

gw3583:
@kvark Yep - there's some details on why here https://github.com/servo/webrender/pull/2980#issuecomment-414946095

gw3583:
Looks like a network error on the windows CI machine, hopefully intermittent.

@bors-servo retry

jrmuizel:
CRL or OCSP is broken?

@bors-servo retry

gw3583:
@jrmuizel Yea, seems to be affecting other projects using appveyor too.

jrmuizel:
https://github.com/rust-lang/cargo/issues/2464 seems like an old related issue.

Issue #2934, opened 2018-07-27
Blur filter not clipped properly


Pryx:
I've been noticing for some time weird blur under Facebook videos when running Firefox with Webrender. I've been able to track down the issue and create a minimal testcase:
```
<!DOCTYPE html>
<html>
<head>
	<title>Test</title>
</head>
<body>
	<div style="height:200px;width:200px;overflow:hidden">
		<img src="http://via.placeholder.com/200x200" style="filter:blur(20px)">
	</div>
</body>
</html>
```

Side by side comparison with Opera: 
![side by side](https://user-images.githubusercontent.com/5222912/43343885-2f96c768-91e8-11e8-985f-03fcec2b7128.PNG)

nical:
We create a statcking context which refers to the correct clip chain and has a blur filter applied to it, however the clip attached to the stacking context is applied to the content of the stacking context before the blur is applied.
@glennw is this the correct semantic of the stacking context clip? if so I guess we should wrap it in another stacking context that has the clip and probably make it so that the blur's stacking context doesn't get the clip attached.

My guess is that the intent was for the clip to be applied to the stacking context (rather than the source of its filter) but since it doesn't matter for most filters we need to handle blurs differently. We don't seem to have any code attempting to apply clips to a picture primitive though.

nical:
Equivalent yaml testcase:

```
---
root:
  items:
    - type: stacking-context
      bounds: [100, 100, 300, 300]
      filters: blur(10)
      complex-clip:
        rect: [200, 200, 100, 100]
      items:
      - image: "reftests/filters/firefox.png"
        bounds: 20 20 256 256
```

Darkspirit:
https://bugzilla.mozilla.org/show_bug.cgi?id=1459065
A blur breaks out of an iframe: [Testcase](https://bugzilla.mozilla.org/attachment.cgi?id=9003281)

emilio:
Regressed by 5e422431.

gw3583:
As way of background, I *think* the following is true:
 * A long time ago, WR supported clips only on primitives, and not on stacking contexts.
 * To work around this, Gecko has some fairly complex logic to convert gecko clip chains to primitive clip chains that include any clips from parents that would normally apply to the stacking context.
[...]

PR #2982, opened 2018-08-23
Fix the yaml reader so that clip ids from complex-clip are correctly represented.

emilio:
This fixes the YAML test-case from #2934. I assume similar changes should be
made to Gecko to pass the right clip node ID around.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2982)

### DIFF
--- a/wrench/reftests/filters/filter-blur-clip-ref.yaml
+++ b/wrench/reftests/filters/filter-blur-clip-ref.yaml
@@ -1,0 +1,15 @@
+---
+root:
+  items:
+  - type: clip
+    id: 2
+    bounds: [200, 200, 100, 100]
+    complex:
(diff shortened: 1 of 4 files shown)


kvark:
This is somewhat hacky, reminds me of https://github.com/servo/webrender/pull/2875
I'd very much prefer us to come up with a good design on how the clips are specified and processed by stacking context and iframe items.

gw3583:
@emilio Is still still relevant / correct based on our discussions today?

emilio:
This is correct, but it's probably only an interim stage. Ideally the clip node id on stacking context shouldn't be needed (see #2934) so this would be a temporary thing anyway.

emilio:
Let's close this for now, we can always resurrect it.

PR #2986, opened 2018-08-27
Fix a recent bug which resulted in redundant clips being drawn.

gw3583:
This fixes a performance regression where primitives with
segments would allocate and draw clip masks for the edge
segments.

It's probably not a noticeable regression on most sites.

Also updated rounded rects reftest to have one primitive which
is large enough to invoke clip segment logic.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2986)


gw3583:
Try run here looks good (still a few jobs to complete):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7aaef426d8d992970ec9dce7cbbd5a64cb35c8fc

webrender/src/prim_store.rs
```diff
@@ -2211,7 +2211,8 @@ impl Primitive {
 
             match segment_clip_chain {
                 Some(segment_clip_chain) => {
-                    if segment_clip_chain.clips_range.count == 0 {
+                    if segment_clip_chain.clips_range.count == 0 ||
+                       (!segment.may_need_clip_mask && !segment_clip_chain.has_non_local_clips) {
```
kvark: hmm, shouldn't this be `segment_clip_chain.clips_range.count == 0 || !segment.may_need_clip_mask || segment_clip_chain.has_non_local_clips`?

gw3583: I don't think so.

The ```segment.may_need_clip_mask``` field (perhaps confusingly) refers to whether that segment might need a clip mask due to it's local segment.

So this condition says a segment can be drawn opaque if there are no relevant clip masks at all, or if the segment builder didn't have any local clips and there's no global clips that affect this segment.

It is subtle though, so I could be making a mistake here - could you explain your reasoning why it should be the above?

kvark: I think `may_need_clip_mask` just needs to be renamed for clarity. The code is good.

kvark:
@bors-servo r-
hmm, what about those R1 failures?

gw3583:
Those R1 failures are expected due to a previous PR - they need a slight bit of fuzziness added. I think that has actually already landed now, it's just that this try run was done before that WR update landed.

gw3583:
R1 passed in the re-triggered try.

@bors-servo r=kvark

PR #2989, opened 2018-08-27
Add a low priority scene builder thread.

nical:
This goes on top Kats' work from [bug 1477819](https://bugzilla.mozilla.org/show_bug.cgi?id=1477819) and replaces the pair of rasterizer thread by a single low-priority scene builder thread which coordinates with the normal scene builder thread.

The low priority scene builder is an additional thread that can be used to perform some work before forwarding the transaction to the "normal" scene builder thread.
This lets us choose to rasterize very expensive content blobs without blocking the scene builder allowing updates of the UI display list.
Since all low priority scenes request get forwarded to the regular scene builder as well, any operation that affects both threads (such as flushes or shutting down) can simply be sent to the low priority thread.

The flow between threads looks like this:

```
+------------------+     +------------------+
| LowPSceneBuilder | --> |   SceneBuilder   |
+------------------+     +------------------+
              ^           ^     |
              |           |     |
              |           |     v
         +-----------------------+      +-----------+
api ---> |     RenderBackend     | ---> |  Renderer |
         +-----------------------+      +-----------+

```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2989)

### DIFF
--- a/webrender/src/scene_builder.rs
+++ b/webrender/src/scene_builder.rs
@@ -27,6 +27,7 @@ pub struct Transaction {
     pub request_scene_build: Option<SceneRequest>,
     pub blob_requests: Vec<BlobImageParams>,
     pub blob_rasterizer: Option<Box<AsyncBlobImageRasterizer>>,
+    pub rasterized_blobs: Vec<(BlobImageRequest, BlobImageResult)>,
     pub resource_updates: Vec<ResourceUpdate>,
     pub frame_ops: Vec<FrameMsg>,
     pub set_root_pipeline: Option<PipelineId>,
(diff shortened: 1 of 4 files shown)


webrender/src/scene_builder.rs
```diff
@@ -150,11 +152,14 @@ impl SceneBuilder {
                     build_scene(&self.config, request)
                 });
 
-                let rasterized_blobs = blob_rasterizer.as_mut().map_or(
+                let mut more_rasterized_blobs = blob_rasterizer.as_mut().map_or(
                     Vec::new(),
                     |rasterizer| rasterizer.rasterize(&blob_requests),
                 );
 
+                rasterized_blobs.append(&mut more_rasterized_blobs);
```
kvark: this looks weird. Isn't `rasterized_blobs` just `Vec`? Can't the argument just be `more_rasterized_blobs` without `&mut`?

nical:
> So the idea here is that LP scene builder has the same API as the regular one, but it only cares about blobs and then forwards the request to build the scene to the real (high priority) builder?

Yes. If there are other expensive operations that we'd want to split out of the regular scene building we can (and should) make them happen in the low-priority scene builder as well, but in the short term I need large content blobs to not block the UI.

The low priority queue is opt-in, so for widgets that don't contains web pages the low and high priority senders feed the same queue (which means no-reordering of transactions).

kvark:
Frankly speaking, I'm not a fan of ad-hoc threading model that the PR extends. I think we should eventually have a more flexible job scheduling system that avoids blocking and respects dependencies. So I suppose this is an intermediate solution for us to ship something.
@bors-servo r+

nical:
> I think we should eventually have a more flexible job scheduling system that avoids blocking and respects dependencies.

I've been thinking about things like this a lot lately. There is definitely things we can do to model the flow and ordering of things using data instead of implicitly through threads control flow and I'd also like a more graph-like scheduling system.
That said I think that it'll take time to get something like that right, and also I think'd it'd be hard to nail the requirements at the first attempt without having brought webrender to a shippable state (we figured out we needed this extra level of asynchronicity just a few weeks ago).

It'll be interesting whenever we get to that.

PR #2990, opened 2018-08-28
Provide direct access to DL creation

pyfisch:
Create normalizing constructors for gradients.
Make push_item and push_iter public.

Servo wants to add items to the DisplayListBuilder directly as it already uses webrender display items internally. This allows servo to skip the push_* functions for various simple items. 

This makes the API a bit more unsafe than it already is. Should these functions carry a warning/be unsafe?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2990)

### DIFF
--- a/webrender_api/src/display_list.rs
+++ b/webrender_api/src/display_list.rs
@@ -16,7 +16,7 @@ use time::precise_time_ns;
 use {AlphaType, BorderDetails, BorderDisplayItem, BorderRadius, BorderWidths, BoxShadowClipMode};
 use {BoxShadowDisplayItem, ClipAndScrollInfo, ClipChainId, ClipChainItem, ClipDisplayItem, ClipId};
 use {ColorF, ComplexClipRegion, DisplayItem, ExtendMode, ExternalScrollId, FilterOp};
-use {FontInstanceKey, GlyphInstance, GlyphOptions, GlyphRasterSpace, Gradient};
+use {FontInstanceKey, GlyphInstance, GlyphOptions, GlyphRasterSpace, Gradient, GradientBuilder};
 use {GradientDisplayItem, GradientStop, IframeDisplayItem, ImageDisplayItem, ImageKey, ImageMask};
 use {ImageRendering, LayoutPoint, LayoutPrimitiveInfo, LayoutRect, LayoutSize, LayoutTransform};
 use {LayoutVector2D, LineDisplayItem, LineOrientation, LineStyle, MixBlendMode, PipelineId};
@@ -947,7 +947,12 @@ impl DisplayListBuilder {
         index
     }
 
-    fn push_item(&mut self, item: SpecificDisplayItem, info: &LayoutPrimitiveInfo) {
+    /// Add an item to the display list.
+    ///
+    /// NOTE: It is usually preferable to use the specialized methods to push
+    /// display items. Pushing unexpected or invalid items here may
+    /// result in WebRender panicking or behaving in unexpected ways.
+    pub fn push_item(&mut self, item: SpecificDisplayItem, info: &LayoutPrimitiveInfo) {
         serialize_fast(
             &mut self.data,
             &DisplayItem {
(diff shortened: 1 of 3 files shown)


webrender_api/src/display_item.rs
```diff
@@ -865,3 +865,123 @@ impl ExternalScrollId {
...
+}
+
+impl RadialGradient {
+    /// Create a normalized radial gradient.
+    ///
+    /// The stops are normalized in-place.
+    pub fn new(
+        center: LayoutPoint,
+        radius: LayoutSize,
+        stops: &mut Vec<GradientStop>,
```
kvark: It feels to me that the concept you really want to expose is `GradientBuilder`. This object would be able to manage `Vec<GradientStop>` internally and produce `Gradient/RadialGradient` objects.

webrender_api/src/display_list.rs
```diff
@@ -947,7 +947,7 @@ impl DisplayListBuilder {
         index
     }
 
-    fn push_item(&mut self, item: SpecificDisplayItem, info: &LayoutPrimitiveInfo) {
+    pub fn push_item(&mut self, item: SpecificDisplayItem, info: &LayoutPrimitiveInfo) {
```
kvark: This should still be memory safe, just easier to panic by pushing unexpected items. So I think no `unsafe` is needed, just add some cautionary documentation on top.

webrender_api/src/display_item.rs
pyfisch: I am not entirely sure how the `GradientBuilder` API should look like. Can you write down the type and function signatures if you have time?

kvark: sure!
```rust
pub struct GradientBuilder {
  stops: Vec<GradientStop>,
  // something else?
}
impl GradientBuidler {
  fn normalize(&mut self) {...}
  pub fn gradient(&mut self, points: Range<LayoutPoint>, extend_mode: ExtendMode) -> Gradient {...}
  pub fn radial_gradient(&mut self, center: LayoutPoint, radius: LayoutSize, extend_mode: ExtendMode) -> RadialGradient {..}
}
```

pyfisch: How do I put stops "into" the builder and how can I get the vector "out" again?

kvark: well, by writing code/methods doing so :)
If this is unclear of confusing to you, we can proceed as is. I'm just thinking that this code is asking for a builder like that, given how it considers the ownership of `Vec<GradientStop>` at each step.

pyfisch: > well, by writing code/methods doing so :)

Sure. It's just that I am not very good at (Rust) API design. I will just add those methods as I think they work best.

pyfisch:
Added a GradientBuilder.

kvark:
Nice, thank you! The code looks great to me. I'd just like one extra change - move out the gradient builder into a separate source module.

pyfisch:
Done :)

webrender_api/src/gradient_builder.rs
```diff
@@ -0,0 +1,149 @@
+use {ExtendMode, Gradient, GradientStop, LayoutPoint, LayoutSize, RadialGradient};
```
kvark: please add the license header, same as other files have

webrender_api/src/gradient_builder.rs
```diff
@@ -0,0 +1,149 @@
...
+                    // but this matches what Gecko and Blink does
+                    stops.push(GradientStop { color: last.color, offset: 0.0, });
+                    stops.push(GradientStop { color: last.color, offset: 1.0, });
+
+                    (0.0, 1.0)
+                }
+            }
+        }
+    }
+}
```
kvark: missing newline at the end

pyfisch:
Yep!

kvark:
@pyfisch would you mind squashing the commits into one?

pyfisch:
Remove trailing whitespace 😒

PR #2991, opened 2018-08-28
Perspective interpolation fix for brush_blend

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1486113
This is a follow-up to #2939

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2991)


kvark:
@gw3583 please check the new commit where the reftest is added. FYI, the old result was:
screen-space-blit-old

PR #2994, opened 2018-08-29
Reduce doc.render() calls if possible

sotaroikeda:
The pull request is for reduce cpu usage by reducing doc.render() calls if possible. It is especially for reduce cpu usage during video playback [Bug 1478566](https://bugzilla.mozilla.org/show_bug.cgi?id=1478566)

It also have an effect of improving talos performance in general.

https://treeherder.mozilla.org/perf.html#/compare?originalProject=try&originalRevision=00e262c79f6716ea4815ca83749da1b17d8e56de&newProject=try&newRevision=7cb9aff96b7b2a573e8f6724d86737344200adb4&framework=1

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2994)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -86,6 +86,8 @@ struct Document {
     // The scene with the latest transactions applied, not necessarily built yet.
     // what we will send to the scene builder.
     pending: SceneData,
+    // Calll render() on next generate_frame.
+    render_on_generate_frame: bool,
 
     view: DocumentView,
 
@@ -143,6 +145,7 @@ impl Document {
                 scene: Scene::new(),
                 removed_pipelines: Vec::new(),
             },
+            render_on_generate_frame: false,
             view: DocumentView {
                 window_size,
                 inner_rect: DeviceUintRect::new(DeviceUintPoint::zero(), window_size),
@@ -204,10 +207,15 @@ impl Document {
                     }
                 };
 
+                let scrolled = self.scroll_nearest_scrolling_ancestor(delta, node_index);
+
                 let should_render =
                     should_render &&
-                    self.scroll_nearest_scrolling_ancestor(delta, node_index) &&
+                    scrolled &&
                     self.render_on_scroll == Some(true);
+
+                self.render_on_generate_frame |= scrolled && !should_render;
+
                 DocumentOps {
                     scroll: true,
                     render: should_render,
@@ -229,13 +237,18 @@ impl Document {
             }
             FrameMsg::SetPan(pan) => {
                 self.view.pan = pan;
+                self.render_on_generate_frame = true;
                 DocumentOps::nop()
             }
             FrameMsg::ScrollNodeWithId(origin, id, clamp) => {
                 profile_scope!("ScrollNodeWithScrollId");
 
-                let should_render = self.scroll_node(origin, id, clamp)
-                    && self.render_on_scroll == Some(true);
+                let scrolled = self.scroll_node(origin, id, clamp);
+
+                let should_render = scrolled &&
+                    self.render_on_scroll == Some(true);
+
+                self.render_on_generate_frame |= scrolled && !should_render;
 
                 DocumentOps {
                     scroll: true,
@@ -583,10 +596,12 @@ impl RenderBackend {
             }
             SceneMsg::SetPageZoom(factor) => {
                 doc.view.page_zoom_factor = factor.get();
+                doc.render_on_generate_frame = true;
                 DocumentOps::nop()
             }
             SceneMsg::SetPinchZoom(factor) => {
                 doc.view.pinch_zoom_factor = factor.get();
+                doc.render_on_generate_frame = true;
                 DocumentOps::nop()
             }
             SceneMsg::SetWindowParameters {
@@ -597,6 +612,7 @@ impl RenderBackend {
                 doc.view.window_size = window_size;
                 doc.view.inner_rect = inner_rect;
                 doc.view.device_pixel_ratio = device_pixel_ratio;
+                doc.render_on_generate_frame = true;
                 DocumentOps::nop()
             }
             SceneMsg::SetDisplayList {
@@ -700,6 +716,13 @@ impl RenderBackend {
         // 2^64 scenes ought to be enough for anybody!
         self.last_scene_id += 1;
         self.last_scene_id
+    }
+
+    pub fn set_render_on_generate_frame(&mut self) {
+        // Set render_on_generate_frame for any existing documents.
+        for (_, doc) in &mut self.documents {
+            doc.render_on_generate_frame = true;
+        }
     }
 
     pub fn run(&mut self, mut profile_counters: BackendProfileCounters) {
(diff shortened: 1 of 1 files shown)


webrender/src/render_backend.rs
```diff
@@ -702,6 +718,13 @@ impl RenderBackend {
         self.last_scene_id
     }
 
+    pub fn set_render_on_generate_frame(&mut self) {
```
NiLSPACE: As a complete outsider this seems confusing to me. If a method starts with `set` I assume I can change it to what I specify, but it always sets it to true. Wouldn't `enable_render_on_generate_frame` make more sense?

gw3583:
@nical Could you take a quick look over this?

zptan:
Sotaro will take PTO in 31/8 ~ 7/9...

webrender/src/render_backend.rs
```diff
@@ -204,10 +207,15 @@ impl Document {
...
 
+                let scrolled = self.scroll_nearest_scrolling_ancestor(delta, node_index);
+
                 let should_render =
                     should_render &&
-                    self.scroll_nearest_scrolling_ancestor(delta, node_index) &&
+                    scrolled &&
                     self.render_on_scroll == Some(true);
+
+                self.render_on_generate_frame |= scrolled && !should_render;
```
kvark: could you explain the logic here please?

webrender/src/render_backend.rs
```diff
@@ -1042,6 +1078,12 @@ impl RenderBackend {
             return;
         }
 
+        if !transaction_msg.resource_updates.is_empty() {
+            // Set render_on_generate_frame for any existing documents
+            // if resource_updates is not empty, since the updates could affects to another documents.
```
kvark: "could affects to another documents"

kvark: Why do we draw the connection between resource updates and frame generation here?

webrender/src/render_backend.rs
mattwoodrow: The comment just above this about double borrowing is probably contributing to this being more complex than it should be.

That said, I think we could rename the earlier usage of should_render to found_scroll_node, and then make this code look like:

let mut should_render = false;
if (found_scroll_node && self.scroll_nearest_scrolling_ancestor(delta, node_index) {
  if (self.render_on_scroll == Some(true)) {
    should_render = true;
  } else {
    self.render_on_generate_frame = true;
  }
}

webrender/src/render_backend.rs
mattwoodrow: It's not changing whether we generate a frame, it's just changing if we call render() at that time.

Before this patch we unconditionally called render() before generating a frame, now we only do so if something relevant has changed that would a new render. I assume that a resource update is one of those things, but not 100% sure.

mattwoodrow:
> I feel a bit uneasy about the increased complexity of the internal state. Could you go into a bit more detail on what use case this change is meant to optimize and why?

Currently when playing video, we get two render() calls. Once when we update the video frame, and again during generate frame.

This patch makes the second call to render() conditional, and only does it if we actually need (we didn't render earlier, or relevant changes have been made since).

webrender/src/render_backend.rs
kvark: (heads up - I edited your code a bit)
the suggestion sounds good to me

webrender/src/render_backend.rs
kvark: do you mean "called render() *after* generating a frame"?

kvark:
@mattwoodrow 

>  Once when we update the video frame, and again during generate frame.

If all Gecko needs to do when playing a video is to update the contents, why do we even go the "generate frame" route?

kvark:
@mattwoodrow I agree, my confusion comes from this sentence:
> Once when we update the video frame, and again during generate frame.
> This patch makes the second call to render() conditional, and only does it if we actually need (we didn't render earlier, or relevant changes have been made since).

Since V-sync triggered render makes more sense to leave untouched, it should be the *first* call to render() that is conditional (currently triggered by the image update, if I understand correctly), not the second.

mattwoodrow:
> Since V-sync triggered render makes more sense to leave untouched, it should be the first call to render() that is conditional (currently triggered by the image update, if I understand correctly), not the second.

We might have to wait for Sotaro to get back, I am running out of answers :)

My reading of the patch is that it's doing the opposite, doing a render() on the image update, and skipping the one on vsync/generate-frame.

nical:
We need to do the render at vsync time and skip the other one, otherwise the frame is rendered with the previous animated properties which we sample and send with the vsync transaction.

nical:
I've been experimenting with building frames only when strictly necessary. I think that we should be able to figure out the minimum amount of work needed by looking at what's in the transaction in the render backend.

One tricky aspect is that scrolling depends on having valid hit testing information while also invalidating the frame.
Since building the hit testing information happens while building the frame, we end up with a

FrameBuild <- HitTest <- FrameBuild dependny chain that breaks when we try to remove frame builds when no rendering is requested.
To fix this we can make store a flag that says whether the hit test info is up to date, and build the frame before (to get hit test info) and after (to render with new scroll offsets) handling the scroll message. This is a bit silly but technically not worse than what we have now, and the frame build happening before can be replaced with a simpler version that only builds the hit testing info as a followup optimization.

kvark:
@nical what would helpful here to resolve the dependency roundrips is to not do a full frame rebuild in order to get the hit test info, and instead do some minimal processing there. This is #2807

sotaroikeda:
>Since V-sync triggered render makes more sense to leave untouched, it should be the first call to render() that is conditional (currently triggered by the image update, if I understand correctly), not the second.

We need to update hittest on document render when display list is updated. And by [Bug 1478566 Comment 9](https://bugzilla.mozilla.org/show_bug.cgi?id=1478566#c9), render document during display update resulted in better scroll performance.

sotaroikeda:
> Currently when playing video, we get two render() calls. Once when we update the video frame, and again during generate frame.

On windows, we do not get  render() for updating the video frame by  [Bug 1477608](https://bugzilla.mozilla.org/show_bug.cgi?id=1477608) if video frame format is same.

sotaroikeda:
> On windows, we do not get render() for updating the video frame by Bug 1477608 if video frame format is same.

Then the content does not have apz change nor animation nor display list update, we do not need to call  render() during generating frame.

webrender/src/render_backend.rs
sotaroikeda: Thanks @mattwoodrow, it is what I intended!

webrender/src/render_backend.rs
sotaroikeda: Thanks for the advice:) Yes, it makes more sense!

webrender/src/render_backend.rs
sotaroikeda: On windows, we do not get render() for updating the video frame by [Bug 1477608](https://bugzilla.mozilla.org/show_bug.cgi?id=1477608) if video frame format is same. Then the content does not have apz change nor animation nor display list update, we do not need to call render() during generating frame.

sotaroikeda:
> Then the content does not have apz change nor animation nor display list update, we do not need to call render() during generating frame.

So for me  #2807 a bit orthogonal problem. This pull request is for skip render()  during generating frame if possible.

kvark:
> We need to update hittest on document render when display list is updated

Hittest metadata shouldn't need to be changed if we just replaced the external ID of an existing image item to a new one: dimensions, formats, and positions all stay the same. This is what the "first" frame-build (of the 2) is about, if I understand correctly.

> So for me, #2807 an orthogonal problem. This pull request is for skip/suppress render() during generating frame if possible.

If we don't need to build a frame for updating hit-testing, there are no longer going to be 2 calls to render(). This is not a blocker for the PR, but I find it highly related.

sotaroikeda:
> I have something like this partially implemented, I'll put a PR up soon-ish.

Thanks!

sotaroikeda:
This was replaced by #3043.

PR #2995, opened 2018-08-29
Various improvements towards correct handling of 3d transforms.

gw3583:
* Pictures only calculate a local rect if they are not a pass-through primitive (i.e. if they have a surface).
* Change batching code to use world rects. This avoids the need to do int conversions, and avoids the need to offset each bounding rect.
* Change prim metadata to store world rects in floats, rather than integer screen rects. This integrates better with the plane splitting code.
* Change clipping code to project to world space for complex clips.
* Introduce SpaceMapper, a struct for efficiently converting between arbitrary coordinate systems.
* Introduce Picture coordinate space, which is used to hold the local rect for a given Picture which has a render surface.
* Prim metadata now only stores the clipped world rect. Most of the time the unclipped world rect isn't needed. It's only needed for Picture primitives that are drawing to a surface - so we calculate it on demand in Picture::prepare_for_render().

The major change here is that pictures now only calculate a local rect if they have a surface. Previously, all pictures tried to calculate a local rect. This caused problems because we were trying to map rects with partial transforms, and losing information as we then only considered the 2d projection within each picture. Now, we only calculate a local rect for pictures that actually have a surface. This fixes several reftests.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2995)


gw3583:
r? @kvark

This is, I hope, the last major change we need in order to be able to rasterize pictures in different coordinate spaces. This will make caching feasible for all picture surface types, and make it possible to fix the remaining 3d transform issues. Apologies for the size of the patch!

Try run is:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d23a9348cbf9a08fadb700a34c85dbe4c28bec6e

There's a lot of orange items, some discussion below:

Unrelated intermittents:
```ts_paint_heavy```
```tp6_google_heavy```
```nav2_test_document_open.html```
```update.https.html```

FAIL -> PASS
	```mask-layer-3.html```
	```1081185-1.html```
	```transform3d-preserve3d-011.html```
	```object-position-svg-001o.html``` (not sure if caused by the existing intermittent PASS, or now permanent)

PASS -> FAIL
```img-novb-height-meet-1.html``` and ```img-novb-height-slice-1.html``` - these are already marked as fuzzy for WR and skiaContent. I think it's probably fine to increase the fuzz here.
```long-chain.html``` - fail on Linux only - now passes on Windows/OSX. Already has fuzziness on WR and all other platforms. I think it's fine to increase the fuzziness here.
```async-scrolling/perspective-scrolling-1.html``` and ```async-scrolling/perspective-scrolling-4.html``` - discussion below.

Most of the new failures are fine to fuzz, I think. I'm still not 100% sure what is causing the ```perspective-scrolling``` failure. I *think* it is actually a race condition in Gecko or the async scene code in WR, but I'm not certain yet. What leads me to think this is:
 * I changed the reference image to also include ```preserve-3d``` and the tests still fail.
 * That means the only difference between the tests is that one sets scrollTop via JS, and one sets async-scroll reftest property.
 * I logged out the state of the clip-scroll tree for each test when the screenshot is taken:

```
Update RF SpatialNodeIndex(0) [I]
Update other SpatialNodeIndex(1) (0.0,0.0)
Update RF SpatialNodeIndex(2) [I]
Update other SpatialNodeIndex(3) (0.0,0.0)
Update other SpatialNodeIndex(4) (0.0,0.0)
Update RF SpatialNodeIndex(5) [1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 24.00621, 0.0, 1.0]
Update other SpatialNodeIndex(6) (0.0,-100.0)
Update RF SpatialNodeIndex(7) [1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, -400.0, -400.0, 1.0, -1.0, 100.0, 100.0, 0.0, 1.0]
```

```
Update RF SpatialNodeIndex(0) [I]
Update other SpatialNodeIndex(1) (0.0,0.0)
Update RF SpatialNodeIndex(2) [I]
Update other SpatialNodeIndex(3) (0.0,0.0)
Update other SpatialNodeIndex(4) (0.0,0.0)
Update RF SpatialNodeIndex(5) [I]
Update other SpatialNodeIndex(6) (0.0,0.0)
Update RF SpatialNodeIndex(7) [1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, -400.0, -500.0, 1.0, -1.0, 100.0, 100.0, 0.0, 1.0]
```

As you can see from above, the incoming offset in spatial node 6 is different between the tests - one has an offset while one doesn't. I *think* this explains why the test fails, and suggests the problem lies earlier than the changes in this patch could cause. But I want to do a bit more investigation first to try and confirm what is going on here.

gw3583:
(TL;DR of the above is: apart from review comments / changes, I need to do a little bit more investigation of a couple of the try results, to either confirm that they are unrelated to this patch, or fix them first before landing).

gw3583:
OK, I think I have fixed the ```perspective-scrolling-1.html``` test. Kicking off a try run shortly.

gw3583:
It was an issue with not including the accumulated scroll offset when defining a new coordinate system.

Current try run is here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=cf31254cd5f2e4e13d92c8a09a19dc5021df9db6&selectedJob=196555936

Still a few jobs to go, but I think it looks good now, despite all the orange items (I'll follow up with some details in a comment shortly about the differences).

gw3583:
Linux:

[1] [2] [mda1] [h1] [h2] [tp6] [mda2] [wpt3] unrelated failures

[R1] New PASS in mask-layer-3.html
[R4] New PASS in blur-cap-large-radius-on-software.html
[R8] New PASS in 1081185-1.html
[Wr1] New PASS in transform3d-preserve3d-011.html
[R2] New FAIL in img-novb-height-meet-1.html and img-novb-height-slice-1.html [see below]
[R7] New FAIL in long-chain.html [see below]

OSX:

[R1] New PASS in 1081185-1.html
[R2] New PASS in long-chain.html and blur-cap-large-radius-on-software.html

Windows:
[not completed yet - I believe they will be the same differences as above, but will confirm]

Discussion:

So the failures are in img-novb-height-meet-1.html, img-novb-height-slice-1.html and long-chain.html (on Linux). In each of these cases the results are very close, and already have fuzziness on WR and/or other backends. With signoff from @jrmuizel, I think these are fine to fuzz, which just leaves new PASS results.

gw3583:
Addressed most of the comments - there are a couple of questions that might need additional changes, otherwise I think this is ready to go now.

PR #2997, opened 2018-08-30
Fix backface visibility in nested stacking contexts.

gw3583:
Picture primitives should not consider the backface visibility
flag - they are containers for primitives only. Instead, if all
the primitives inside a picture are backface culled, then the
entire picture will be empty and also culled.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1442916.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2997)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -732,10 +732,7 @@ impl<'a> DisplayListFlattener<'a> {
                 unreachable!("Should have returned in parent method.")
             }
             SpecificDisplayItem::PushShadow(shadow) => {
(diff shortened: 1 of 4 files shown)


gw3583:
r? @kvark 

Try run looks good (a couple jobs remaining still):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=fcc299da568f0ce64cede984969de03c5dfce4b1

kvark:
Try looks good to me.
@bors-servo r+

gw3583:
It's not clear to me from the logs why this failed - timeout?

@bors-servo retry

PR #2992, opened 2018-08-28
Enforce async scene building

nical:
Make async scene building mandatory. There are some simplifications we can get out of doing this that I want to take advantage of for the async tab switch work.
This is a first step that enforces that all transactions which invalidate the scene must go through the scene builder thread. More stuff coming.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2992)

PR #2998, opened 2018-08-30
Move the pending scene data structure to the SceneBuilder thread

nical:
This PR reorganizes the code in the render backend and scene builder such that:

 - The render backend now only manages a single scene per document: the one it can use for frame building.
 - The scene builder now manages a scene per document and uses it to generate the `FrameBuilder`.
 - The render backend sends to the scene builder thread the events it needs to keep the Scene data structure up to date.
 - Transactions have been consolidated around 3 structs: `TransactionMsg` is what the API sends to the render backend, `Transaction` is what the render backend creates and sends to the scene builder thread(s), and `BuiltTransaction` is what the scene builder sends back to the render backend after scene building. These three structs represent the same thing but at different stages of the pipeline.
 - A *lot* of code was simplified (for example no more calling `update_document` twice with different parameters before and after scene building to have it do different things)

This change is needed in order to get the low priority scene building in #2989 to work without requiring extra handshakes between the scene builder and the render backend threads.

This is rebased on top of #2992.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2998)

### DIFF
--- a/webrender/src/scene_builder.rs
+++ b/webrender/src/scene_builder.rs
@@ -30,7 +30,7 @@ pub struct Transaction {
     pub resource_updates: Vec<ResourceUpdate>,
     pub frame_ops: Vec<FrameMsg>,
     pub set_root_pipeline: Option<PipelineId>,
-    pub generate_frame: bool,
+    pub build_frame: bool,
     pub render: bool,
 }
 
@@ -62,7 +62,7 @@ pub struct BuiltTransaction {
     pub removed_pipelines: Vec<PipelineId>,
     pub scene_build_start_time: u64,
     pub scene_build_end_time: u64,
-    pub generate_frame: bool,
+    pub build_frame: bool,
     pub render: bool,
 }
 
@@ -92,7 +92,7 @@ pub struct LoadScene {
     pub font_instances: FontInstanceMap,
     pub view: DocumentView,
     pub config: FrameBuilderConfig,
-    pub generate_frame: bool,
+    pub build_frame: bool,
 }
 
 pub struct BuiltScene {
@@ -242,8 +242,8 @@ impl SceneBuilder {
 
             let txn = Box::new(BuiltTransaction {
                 document_id: item.document_id,
-                generate_frame: true,
-                render: item.generate_frame,
+                build_frame: true,
+                render: item.build_frame,
                 built_scene,
                 resource_updates: Vec::new(),
                 rasterized_blobs: Vec::new(),
@@ -322,7 +322,7 @@ impl SceneBuilder {
 
         Box::new(BuiltTransaction {
             document_id: txn.document_id,
-            generate_frame: txn.generate_frame || built_scene.is_some(),
+            build_frame: txn.build_frame || built_scene.is_some(),
             render: txn.render,
             built_scene,
             resource_updates: replace(&mut txn.resource_updates, Vec::new()),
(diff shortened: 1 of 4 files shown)


nical:
I am so sorry for this wall of code. Most of it is moving things around. I think that the code is quite a bit easier to understand after this change than before but the diff view doesn't do it justice. You can ignore the first 5 commits which are from #2992.

webrender/src/scene_builder.rs
```diff
@@ -3,47 +3,116 @@
...
 use resource_cache::FontInstanceMap;
 use render_backend::DocumentView;
 use renderer::{PipelineInfo, SceneBuilderHooks};
 use scene::Scene;
 use std::sync::mpsc::{channel, Receiver, Sender};
+use std::mem::replace;
 use time::precise_time_ns;
 
+/// Represents the work associated to a transaction before scene building.
+pub struct Transaction {
```
nical: Whether splitting the data in `Transaction` into a lot of vectors the best is up for debate, here is a bit of context:
I realize that having a bunch of small vectors means a bunch of small allocations, however:
 - This representation made it easy to do a few of things in this PR such as knowing whether the transaction needs triggers a scene build and whether it can skip the scene builder by simply looking at which vectors are empty.
 - It will simplify the problem of merging transaction whenever we get to implementing that. Right now if scene building takes a long time, requests can accumulate faster than they are processed which leads to growing delay between user input and rendering which is awful. I want to be able to take several transactions and merge them into one that accumulates the important states while discarding redundant or out of date changes. This is a easier to do if things are organized this way than with a single vector.

nical:
I added a commit that simplifies the terminology around generating frames and rendering (and fixes a bug I caused because of the associated confusion)

Before this PR:

 - `render` meant building a frame with the frame builder but not necessarily asking the renderer to render it.
 - `composite` meant asking the renderer to render. 

After this PR:

 - `generate_frame` means asking the frame builder to build a frame.
 - `render` means asking the renderer to render a frame.

nical:
try push looks good so far: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e606386b9e899fcde16d8344fdc8c5ddaaf9e6e3

The push has all of the commits since the last WR update so I can't take credit for the unexpected passes. There are also a few failures with rendering differences within fuzzing range and I don't think these can be the result of this PR.

nical:
> so on the high level - why not make all mutation of the document to go through a transaction?

The way I picture things in my mind, they all do. There are 3 structs that represent the transaction at different stages of the pipeline TransactionMsg -> Transaction -> BuiltTransaction

Some things get applied between TransactionMsg and Transaction (that's the ones you are referring to in this comment), others between Transaction and BuiltTransaction (scene building, blob rasterization), others after BuiltTransaction (scrolling, frame building for example).

> heads up - still not addressed?

Sorry I got mixed up, it was actually not a mistake. It's correct to move the blob rasterizer out of the transaction here because we want to move it into the render backend (it is uniquely owned). Using `take()` here helped getting things to build because, probably something about partially moving out a boxed struct making the compiler unhappy.

sotaroikeda:
This caused regression of [Bug 1489471](https://bugzilla.mozilla.org/show_bug.cgi?id=1489471) and [Bug 1489189](https://bugzilla.mozilla.org/show_bug.cgi?id=1489189).

PR #2999, opened 2018-08-31
Force all calls to project_rect to include a bounding rect.

gw3583:
This greatly improves the accuracy of the clipper code when
required to calculate the world rect for a primitive that
crosses the near plane.

This patch by itself doesn't fix anything specific, but when
combined with the fixes that are in progress in the
plane-split crate, it will allow correct world rect calculations
for primitives that cross the near plane in various edge cases.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2999)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -24,7 +24,7 @@ use std::f32;
 use std::sync::Arc;
 use tiling::{Frame, RenderPass, RenderPassKind, RenderTargetContext};
 use tiling::{ScrollbarPrimitive, SpecialRenderPasses};
-use util;
+use util::{self, MaxRect};
 
 
 #[derive(Clone, Copy, Debug, PartialEq)]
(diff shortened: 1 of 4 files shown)


gw3583:
r? @kvark 

*Mea culpa*

Pending try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9eaec0a34c022601684bbe70b11f6d76d6d8a7c2

I expect the try run to have some orange items due to the previous PRs that have recently landed, so I'll comment on the results once the try completes.

gw3583:
Try looks good:

It matches the current baseline https://treeherder.mozilla.org/#/jobs?repo=try&revision=d25ddc2775a1146d2dad8b63fa8d011f49a6222e&selectedJob=196793766 against WR master, except for one new PASS  in R1 [1157984-1.html].

webrender/src/frame_builder.rs
```diff
@@ -108,14 +108,24 @@ impl PictureState {
...
             ref_spatial_node_index,
             clip_scroll_tree,
         );
+        let pic_bounds = map_pic_to_world.unmap(
+            &world_rect,
+        ).unwrap_or(PictureRect::max_rect());
+
+        let map_local_to_pic = SpaceMapper::new(
+            ref_spatial_node_index,
+            pic_bounds,
```
kvark: looks like the `SpaceMapper` abstraction is paying off nicely here :+1:

webrender/src/prim_store.rs
```diff
@@ -161,6 +166,21 @@ impl<F, T> SpaceMapper<F, T> where F: fmt::Debug {
         }
     }
 
+    pub fn unmap(&self, rect: &TypedRect<f32, T>) -> Option<TypedRect<f32, F>> {
```
kvark: I think this should be renamed. It sounds like the mapping was reversible in the first place and we just undo it, but in reality `inverse_rect_footprint` is clearly irreversible.

kvark:
This resolves my concern from #2995 review. Thank you!
@bors-servo r+

PR #3000, opened 2018-08-31
Add requested/actual composite mode to Picture primitive.

gw3583:
Previously, the composite_mode for a Picture was set during
display list flattening, and then not changed.

In the near future, we want to be able to decide during frame
build what the composite mode of a picture is. For example, if
we encounter a picture with a perspective transform, we may
decide to rasterize it in local space. Equally, if we encounter
a picture in a 3d rendering context, but detect there is no
perspective present and no plane splitting needed, we may
choose to skip drawing this to a surface and draw directly into
the parent.

This patch doesn't contain any functional changes - it just
introduces the concept of a requested and actual composite
mode, and updates surrounding code for this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3000)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -490,6 +490,9 @@ impl AlphaBatchBuilder {
                             );
 
             let source_task_id = pic
+                .raster_config
+                .as_ref()
+                .expect("BUG: no raster config")
                 .surface
                 .as_ref()
                 .expect("BUG: unexpected surface in splitting")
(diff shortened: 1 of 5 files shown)


gw3583:
r? @kvark 

```
This patch doesn't contain any functional changes - it just
introduces the concept of a requested and actual composite
mode, and updates surrounding code for this.
```

Hopefully that makes this patch easy to review, despite the number of lines changed.

Pending try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=fb07964ca7f1f0eb591c8a508a6feb5e5022506f

gw3583:
Addressed review comments and rebased.

kvark:
PR 3000 yo!
@bors-servo r+

PR #3001, opened 2018-08-31
Fix order of compositing operations with a single stacking context.

gw3583:
A rare case is that a stacking context has both a mix-blend-mode
and a filter list on it.

The CSS compositing specification says that filters should be
applied first, and then any compositing.

Fixing this will also simplify some upcoming optimizations that
make mix-blend-mode much faster in commonly used scenarios.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3001)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -1046,36 +1046,6 @@ impl<'a> DisplayListFlattener<'a> {
             *self.picture_stack.last().unwrap()
         };
 
-        // For each filter, create a new image with that composite mode.
-        for filter in composite_ops.filters.iter().rev() {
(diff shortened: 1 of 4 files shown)


gw3583:
r? @kvark or anyone

(I will kick off a try run shortly).

gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c0277d254dc03ba7a0bfebb3bc4d82c3b331e869

Looks good so far.

kvark:
@gw3583 
> Fixing this will also simplify some upcoming optimizations that
make mix-blend-mode much faster in commonly used scenarios.

I'm curious about the details here :)

PR #3002, opened 2018-08-31
allow blob images to query font instance data

lsalzman:
This feature is to help solve downstream Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1480615

We need to be able to recreate ScaledFonts in Gecko from WR's font instance data so that blob images and WR can line up their font usage. The only thing missing for doing this was a way to get the necessary font instance data back out. This just adds a simple query to BlobImageResources to do that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3002)


kvark:
Looks harmless :)
@bors-servo r+

Issue #3004, opened 2018-09-01
Consider removing `flat varying` for ANGLE


djg:
Using `flat varying` between vertex shader and fragment shader causes ANGLE to generate a geometry shader per draw prim category for those programs.

Discussing the issue with @gw3583, it was suggested to experiment removing flat in all cases or just special casing it for ANGLE.

jrmuizel:
I was curious why this geometry shader was needed and found: https://chromium-review.googlesource.com/c/angle/angle/+/309155 and https://bugs.chromium.org/p/angleproject/issues/detail?id=754
"OpenGL ES3 does not use the same 'provoking vertex' for flat shading as Direct3D 11. See Table 2.12 of the ES3 spec, versus the 'leading vertex' illustrated in http://msdn.microsoft.com/en-us/library/windows/desktop/bb205124(v=vs.85).aspx"

gw3583:
The way we use ```flat``` attributes is an optimization hint in most cases, and (I think) easily worked around in other cases. So we can probably just not use ```flat``` at all (at least in the case of ANGLE).

kvark:
@jrmuizel thanks for those links. This is just sad... I think they should give us an option to say "I don't give a heck about provoking vertex". And since they are at it, also write down some documentation about the performance pitfalls.

nical:
I documented this in the [platform/driver bag of sadness](https://github.com/servo/webrender/wiki/Driver-issues#3004-slow-flat-varying-implementation-in-angle).

pcwalton:
Note that there is a `glProvokingVertex()` function in GL 4.x (extension in 3.x I think) that is designed to solve this problem: https://www.khronos.org/opengl/wiki/Primitive#Provoking_vertex

 I don't know if ANGLE supports it, though.

djg:
Talking this issue over with @gw3583 and asking my friendly GPU expert, `flat` is a possible improvement on unified shader architectures. Ideally, we'd like ANGLE not to use geometry shaders and to keep the `flat` specifier.

djg:
Discussion with @gw3583, we decided not to move ahead with this.

PR #3007, opened 2018-09-03
Remove flat varying

djg:
This addresses Issue #3004

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3007)


jrmuizel:
```
thread 'validate_shaders' panicked at 'Shader compilation failed: cs_border_segment
ERROR: 0:599: 'mod' : no matching overloaded function found
ERROR: 0:599: '/' : wrong operand types - no operation '/' exists that takes a left-hand operand of type 'in highp float' and a right operand of type 'const int' (or there is no acceptable conversion)
', webrender/tests/angle_shader_validation.rs:143:13
```

djg:
> thread 'validate_shaders' panicked at 'Shader compilation failed: cs_border_segment
> ERROR: 0:599: 'mod' : no matching overloaded function found
> ERROR: 0:599: '/' : wrong operand types - no operation '/' exists that takes a left-hand operand of type 'in highp float' and a right operand of type 'const int' (or there is no acceptable conversion)
>', webrender/tests/angle_shader_validation.rs:143:13

Because `2` is not the same as `2.0` 🤦‍♂️

PR #3008, opened 2018-09-03
Fix framebuffer size in multiwindow example

zakorgy:
#2910 previously fixed the framebuffer size in the boilerplate file. Since the multiwindow example doesn't use the boilerplate, the framebuffer size didn't get fixed in this example.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3008)


gw3583:
@bors-servo r+

Thanks!

PR #3009, opened 2018-09-03
Provide Gecko with ImageRendering property when locking NativeTexture external images

### LINKED ISSUE #2993: ImageRendering property is not updated on NativeTexture external images on Windows and MacOS(SkiaGL)
When investigating https://bugzilla.mozilla.org/show_bug.cgi?id=1452979 with the help of @mstange, we concluded that the issue is produced within WR. When using canvas elements that have a CSS image-rendering property, Gecko creates the texture as a native D3D11Texture and initialises it using linear filtering. Neither Gecko nor WR change this filtering ever after creating the texture, thus resulting in a filtering that is always the same as on creation, see https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/gfx/webrender_bindings/RenderD3D11TextureHostOGL.cpp#111.

Right now, the ImageRendering property of an image is correctly applied when it's placed in the correct texture cache, excluding any textures that are not cached. WR should be able to apply filtering to external images, e.g. when showing the same video twice with different filtering settings. Looking at https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/gfx/webrender/src/batch.rs#1635, the ImageRequest objects already holds the correct ImageRendering property. Therefore, it should be possible to apply this filtering property. 

I would like to work on this, if possible :) My proposal is to extend the DeferredResolve struct with an ImageRendering item (see https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/gfx/webrender_api/src/display_item.rs#578) and add it to the deferred_resolves list. The lock call at https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/gfx/webrender/src/renderer.rs#3456 could then use this ImageRendering item to provide the correct filter at https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/gfx/webrender_bindings/RenderD3D11TextureHostOGL.cpp#111 or https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/gfx/webrender_bindings/RenderD3D11TextureHostOGL.cpp#133

hwinnemoe:
fixes #2993

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3009)

### DIFF
--- a/examples/frame_output.rs
+++ b/examples/frame_output.rs
@@ -50,7 +50,7 @@ impl webrender::OutputImageHandler for OutputHandler {
 }
 
 impl webrender::ExternalImageHandler for ExternalHandler {
-    fn lock(&mut self, _key: ExternalImageId, _channel_index: u8) -> webrender::ExternalImage {
+    fn lock(&mut self, _key: ExternalImageId, _channel_index: u8, _rendering: ImageRendering) -> webrender::ExternalImage {
         webrender::ExternalImage {
             uv: TexelRect::new(0.0, 0.0, 1.0, 1.0),
             source: webrender::ExternalImageSource::NativeTexture(self.texture_id),
--- a/examples/frame_output.rs
+++ b/examples/frame_output.rs
@@ -50,7 +50,12 @@ impl webrender::OutputImageHandler for OutputHandler {
 }
 
 impl webrender::ExternalImageHandler for ExternalHandler {
-    fn lock(&mut self, _key: ExternalImageId, _channel_index: u8, _rendering: ImageRendering) -> webrender::ExternalImage {
+    fn lock(
+        &mut self,
+        _key: ExternalImageId,
+        _channel_index: u8,
+        _rendering: ImageRendering
+    ) -> webrender::ExternalImage {
         webrender::ExternalImage {
             uv: TexelRect::new(0.0, 0.0, 1.0, 1.0),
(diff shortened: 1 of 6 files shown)


hwinnemoe:
Try push of WIP, also includes additional fixes: https://treeherder.mozilla.org/#/jobs?repo=try&revision=4fc42b3e4e77699bff309491005f76ff1d3572fe

gw3583:
Looks like the examples need to be updated:

```
error[E0050]: method `lock` has 3 parameters but the declaration in trait `webrender::ExternalImageHandler::lock` has 4
  --> examples\texture_cache_stress.rs:64:62
   |
64 |     fn lock(&mut self, _key: ExternalImageId, channel_index: u8) -> webrender::ExternalImage {
   |                                                              ^^ expected 4 parameters, found 3
   |
   = note: `lock` from trait: `fn(&mut Self, webrender::<unnamed>::ExternalImageId, u8, webrender::<unnamed>::ImageRendering) -> webrender::ExternalImage<'_>`
```

kvark:
@gw3583  We've got to start tracking the breaking changes more careful as WR gets published.
@bors-servo r=gw3583

PR #3010, opened 2018-09-03
Be consistent about how the algorithm we use to modulate colors.

emilio:
In particular, if the color is black, every browser does its best to infer the
color.

Fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1488294

This should require a couple test updates in Gecko I presume.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3010)


gw3583:
@emilio Could you do a try run and paste the link?

emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=85f6a936bf2a8512f87d25fded0a8c05db7cae88 is the try run.

I expect a couple table tests will start failing, that's because right now we're painting both the test and the reference with black borders, which is wrong.

Marking them as expected failures should be fine. They should be fixed over with the work from https://bugzilla.mozilla.org/show_bug.cgi?id=1487407.

emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=51c7f3991d72e4a2323706b3c2ad2433bc1cee85 should be a faster (but linux-only) try run.

emilio:
> https://treeherder.mozilla.org/#/jobs?repo=try&revision=85f6a936bf2a8512f87d25fded0a8c05db7cae88 is the try run.

So, quick report on the failures... A bunch of the table-bordercollapse failures are expected, and should be progressions (before we should be rendering black borders instead of groove and co.).

Most of the failures are single-pixel borders. I missed that Gecko has the condition that if it needs two colors for a border but doesn't have two pixels, it'll just render a solid border with the unmangled color.

Can be seen in the following page:

```html
<!doctype html>
<style>
  div {
    border: 1px groove black;
  }
</style>
<div style="border: 1px groove black">Should not have lighter border</div>
<div style="border: 10px groove black">Should have lighter border</div>
```

The latest commit should address this. The failure in layout/reftests/bugs/186317-1.html is of the same nature.

A bit more worrying is the failure in `sticky-legend-1.html`. Looks like the rendering is correct, but we're getting the antialiasing slightly differently along the border. That's because (I presume) the weird position: sticky clipping, or something of that sort. I should be able to reduce it, but: @gw3583 / @jrmuizel / @staktrace have you seen something like that before?

emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=0e3afa2058fd3f1ed19af3a2fd9eec0e9e817704 is a more updated try run which should only include the bordercollapse failures and the sticky legend one.

emilio:
r? @kvark or @gw3583 

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1488403 to fix the sticky legend reftest to not depend on the AA we do around the groove border, so assuming that gets accepted this should be green on try. I couldn't reproduce that failure locally, but I think I understand the root cause (see the patch there).

We may want to represent groove borders as two different border segments to avoid doing distance AA across them, but it's not clear it's better to me, and in any case it's not the point of neither this change or the test.

emilio:
(or @nical, if you're familiar with this code)

webrender/res/cs_border_segment.glsl
```diff
@@ -102,22 +102,41 @@ vec2 get_outer_corner_scale(int segment) {
...
+        return vec4(color.rgb * light_scale, color.a);
+    }
+    return vec4(color.rgb * dark_scale, color.a);
 }
 
 vec4[2] get_colors_for_side(vec4 color, int style) {
     vec4 result[2];
-    const vec2 f = vec2(1.3, 0.7);
+
+    bool is_black = color.rgb == vec3(0.0, 0.0, 0.0);
```
kvark: This may not be portable. IIRC, vector comparison produces vector of booleans. We should be using `dot(color.rgb,color.rgb)` or `all(color.rgb == vec3(0.0, 0.0, 0.0))`

nical: I could be wrong but I think that  `some_vec3 == other_vec3` returns a boolean in glsl and that the [equal function](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/equal.xhtml) returns a vector of boolean.

kvark: Indeed. I just recalled a [driver bug](https://www.opengl.org/discussion_boards/showthread.php/169135-vector-equality) with those, hence my radar fired up the red light :)

kvark: Also note the relevant record in our own driver bug treasury: https://github.com/servo/webrender/wiki/Driver-issues#1728-1368---shader-compiler-bug-when--is-used-inline-as-function-arg

PR #3011, opened 2018-09-03
Add simplier border shader for solid borders.

### LINKED ISSUE #3003: Consider splitting up cs_border_segment shader
Profiling shader cache utilization on Windows shows `cs_border_segment` spends 10x the amount of time in the Intel graphics driver than any of the other "start up" shaders.

Discussing with @gw3583 splitting out the most common, simple border shader for the 99% common styles (i.e. solid), and then a complex shader for the other border styles.

Additionally, `web_render` used to draw borders as part of the main scene, so having one border shader was important for batching. However, borders are drawn into the texture cache at the start of the frame, so they can be reused (they are then just drawn as images in the main frame). So, batching doesn't matter in this case, and potentially, the border shader could be split up into 7, or more, simple shaders, one for each border style.

djg:
This fixes #3003

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3011)

### DIFF
--- a/webrender/res/cs_border_solid.glsl
+++ b/webrender/res/cs_border_solid.glsl
@@ -1,0 +1,223 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include shared,ellipse
+
+// For edges, the colors are the same. For corners, these
+// are the colors of each edge making up the corner.
+flat varying vec4 vColor0;
+flat varying vec4 vColor1;
+
+// A point + tangent defining the line where the edge
+// transition occurs. Used for corners only.
+flat varying vec4 vColorLine;
+
+// x = segment, z = edge axes, w = clip mode
+flat varying ivec4 vConfig;
+
+// xy = Local space position of the clip center.
+// zw = Scale the rect origin by this to get the outer
+// corner from the segment rectangle.
+flat varying vec4 vClipCenter_Sign;
+
+// An outer and inner elliptical radii for border
+// corner clipping.
+flat varying vec4 vClipRadii;
+
+// Reference point for determine edge clip lines.
+flat varying vec4 vEdgeReference;
+
+// Stores widths/2 and widths/3 to save doing this in FS.
+flat varying vec4 vPartialWidths;
+
+// Clipping parameters for dot or dash.
+flat varying vec4 vClipParams1;
+flat varying vec4 vClipParams2;
+
+// Local space position
+varying vec2 vPos;
+
+#define SEGMENT_TOP_LEFT        0
+#define SEGMENT_TOP_RIGHT       1
+#define SEGMENT_BOTTOM_RIGHT    2
+#define SEGMENT_BOTTOM_LEFT     3
+#define SEGMENT_LEFT            4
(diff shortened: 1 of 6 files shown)


gw3583:
@djg Could you do a try run and paste the link?

djg:
@gw3583 I can. I think. :-)

djg:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=e408a5e4f74ff9908b879e4913f777fe2509d353&selectedJob=197291382

Looks like so reftest failures to cleanup.

kvark:
@djg there is a bunch of unexpected passes (:tada: ), and all the failures are marked as intermittent.

webrender/res/cs_border_solid.glsl
```diff
@@ -0,0 +1,223 @@
...
+            p = vec2(1.0, 0.0);
+            break;
+        case SEGMENT_BOTTOM_RIGHT:
+            p = vec2(1.0, 1.0);
+            break;
+        case SEGMENT_BOTTOM_LEFT:
+            p = vec2(0.0, 1.0);
+            break;
+        default:
+            // Should never get hit
```
kvark: why would it never be hit? the caller function has a `switch (segment)` after calling this, which handles `SEGMENT_TOP` and other non-corner segments. Either they are allowed here, or the other switch needs to be simplified as well.

webrender/src/tiling.rs
```diff
@@ -684,8 +688,13 @@ impl TextureCacheRenderTarget {
                 }
 
                 let instances = mem::replace(&mut task_info.instances, Vec::new());
-
-                self.border_segments.extend(instances);
+                for instance in instances.into_iter() {
```
kvark: nit: `into_iter()` is not needed here

webrender/src/tiling.rs
```diff
@@ -27,6 +27,8 @@ use texture_allocator::GuillotineAllocator;
 use webrender_api::{DevicePixel, FontRenderMode};
 
 const MIN_TARGET_SIZE: u32 = 2048;
+const SOLID_SOLID: i32 = ((BorderStyle::Solid as i32) << 8) | ((BorderStyle::Solid as i32) << 16);
```
kvark: `STYLE_SOLID`?

djg:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9a6cd6951e21d025d791e7aeef0ef7897c55649a&selectedJob=197318957 is just the changes from this patch and not all of webrender. The `transform-3d` reftest failure goes away.

webrender/src/tiling.rs
djg: I called it `SOLID_SOLID` because it matches both `style0` and `style1` being `BorderStyle::Solid`

webrender/res/cs_border_solid.glsl
djg: I copied this code from `cs_border_segment` and that's where the comment comes from. @gw3583 is this comment still valid? From the calling code, it appears `get_outer_corner_scale()` is called for all segments, so `default` handles `SEGMENT_TOP`, `SEGMENT_LEFT`, `SEGMENT_BOTTOM`, and `SEGMENT_RIGHT`.

webrender/src/tiling.rs
kvark: oh, I see now. Yes, `STYLE_SOLID_SOLID` sounds good to me

PR #3012, opened 2018-09-04
Remove flat varying

djg:
This addresses Issue #3004

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3012)

### DIFF
--- a/webrender/res/base.glsl
+++ b/webrender/res/base.glsl
@@ -35,3 +35,19 @@
     precision highp float;
     #define varying in
 #endif
+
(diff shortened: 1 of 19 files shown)


gw3583:
Nice! Could you kick off a try run?

djg:
> Nice! Could you kick off a try run?

Yup. Just sorting out local changes.

djg:
Discussion with @gw3583, we agreed that it was better to keep `flat varying` because it's a win for fragment shader performance.

PR #3020, opened 2018-09-05
Add external event notifications in transactions.

nical:
This allows gecko to attach a callback to the transaction and run it after the scene is built to notify the tab switcher that the work is done.

The first commit adds a drain_filter utility modeled after currently unstable Vec::drain_filter. The idea is to be able to get a vector of operations, and each stage of the pipeline can decide to move operations out of the the array and handle them or let them for the next stages to consume.
Hopefully we can use this to merge back the frame_ops, scene_ops and the other vectors in the `Transaction` struct.
This was previously inconvenient because of how some operations need to move elements out of the vector but while we need to keep the array alive at the same time, and drain_filter makes that less painful.

This PR is based on top of #2989

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3020)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -9,6 +9,7 @@ use api::{DeviceIntPoint, DevicePixelScale, DeviceUintPoint, DeviceUintRect, Dev
 use api::{DocumentId, DocumentLayer, ExternalScrollId, FrameMsg, HitTestFlags, HitTestResult};
 use api::{IdNamespace, LayoutPoint, PipelineId, RenderNotifier, SceneMsg, ScrollClamping};
 use api::{ScrollLocation, ScrollNodeState, TransactionMsg, ResourceUpdate, ImageKey};
+use api::ExternalEvent;
 use api::channel::{MsgReceiver, Payload};
 #[cfg(feature = "capture")]
 use api::CaptureBits;
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -9,7 +9,7 @@ use api::{DeviceIntPoint, DevicePixelScale, DeviceUintPoint, DeviceUintRect, Dev
 use api::{DocumentId, DocumentLayer, ExternalScrollId, FrameMsg, HitTestFlags, HitTestResult};
 use api::{IdNamespace, LayoutPoint, PipelineId, RenderNotifier, SceneMsg, ScrollClamping};
 use api::{ScrollLocation, ScrollNodeState, TransactionMsg, ResourceUpdate, ImageKey};
-use api::ExternalEvent;
+use api::{NotificationRequest, Checkpoint};
 use api::channel::{MsgReceiver, Payload};
 #[cfg(feature = "capture")]
 use api::CaptureBits;
@@ -45,6 +45,7 @@ use std::u32;
 #[cfg(feature = "replay")]
 use tiling::Frame;
 use time::precise_time_ns;
+use util::drain_filter;
 
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
@@ -803,7 +804,7 @@ impl RenderBackend {
             resource_updates: transaction_msg.resource_updates,
             frame_ops: transaction_msg.frame_ops,
             rasterized_blobs: Vec::new(),
-            notifications: Vec::new(),
+            notifications: transaction_msg.notifications,
             set_root_pipeline: None,
             build_frame: transaction_msg.generate_frame,
             render_frame: transaction_msg.generate_frame,
@@ -876,7 +877,7 @@ impl RenderBackend {
         document_id: DocumentId,
         resource_updates: Vec<ResourceUpdate>,
         mut frame_ops: Vec<FrameMsg>,
-        notifications: Vec<ExternalEvent>,
+        mut notifications: Vec<NotificationRequest>,
         mut build_frame: bool,
         mut render_frame: bool,
         frame_counter: &mut u32,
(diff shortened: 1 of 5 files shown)


webrender/src/util.rs
```diff
@@ -508,3 +508,67 @@ pub fn world_rect_to_device_pixels(
     let device_rect = rect * device_pixel_scale;
     device_rect.round_out()
 }
+
+/// Run a first callback over all elements in the array. If the callback returns true,
```
kvark: nit: "the first callback"

webrender/src/util.rs
```diff
@@ -508,3 +508,67 @@ pub fn world_rect_to_device_pixels(
...
+/// The element is removed from the array and moved to a second callback.
```
kvark: nit: "the element"

webrender/src/util.rs
```diff
@@ -508,3 +508,67 @@ pub fn world_rect_to_device_pixels(
...
+///
+/// This is a simple implementation waiting for Vec::drain_filter to be stable.
+/// When that happens, code like:
+///
+/// ```
+/// let filter = |&mut op| {
```
kvark: can we have that in unit tests instead of (or in addition to) the documentation?

kvark: I don't think you can even write `|&mut op|` so checking these by the compiler would help ;)

webrender/src/util.rs
```diff
@@ -508,3 +508,67 @@ pub fn world_rect_to_device_pixels(
...
+///     match op {
+///         Foo => { foo(); }
+///         Bar => { bar(); }
+///         Baz => { unreachable!(); }
+///     }
+/// }
+/// ```
+///
+/// See https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter
+pub fn drain_filter<T, F1, F2>(
```
kvark: nit: can we rename `F1 -> Filter` and `F2 -> Action`?

webrender/src/util.rs
```diff
@@ -508,3 +508,67 @@ pub fn world_rect_to_device_pixels(
...
+    mut action: F2,
+)
+where
+    F1: FnMut(&mut T) -> bool,
+    F2: FnMut(T)
+{
+    let mut i = 0;
+    while i != vec.len() {
+        if filter(&mut vec[i]) {
+            action(vec.remove(i));
```
kvark: This is quadratic asymptotic complexity, or rather O(N*M) where M is the number of filter matches. Can we do better?

webrender_api/src/api.rs
```diff
@@ -171,6 +175,10 @@ impl Transaction {
         self.merge(resources);
     }
 
+    pub fn notify(&mut self, event: ExternalEvent) {
```
kvark: can we have  some documentation here on what does this mean? When will the notification be fired, and why we need it.

webrender/src/resource_cache.rs
```diff
@@ -554,8 +554,17 @@ impl ResourceCache {
                 }
                 _ => {}
             }
+        }
 
-            match update {
+        drain_filter(
+            updates,
+            |update: &mut ResourceUpdate| match *update {
```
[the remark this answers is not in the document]
nical: I thought so too, then I was battling some build errors and ended up putting that and thought it was what helped. Turns out this wasn't what fixed my trouble since it indeed compiles when I remove the argument type now.

webrender/src/util.rs
nical: We can but should we? We'll get the the optimized version for free when the std implementation stabilizes and I am certain this will not show up in profiles.
To be honest it also rubs me the wrong way and I'd probably have made more of an effort if we weren't getting the better implementation through std eventually.

kvark: ok, that's fine by me, assuming there is a limited array size on the input

nical: Right now the size of resource updates vector is 0 the vast majority of times, 3 while playing a video. occasionally 1 when loading an image, etc.

webrender_api/src/api.rs
```diff
@@ -171,6 +175,18 @@ impl Transaction {
         self.merge(resources);
     }
 
+    // Note: Gecko uses this to get notified when a transaction that contains
+    // potentially long blob rasterization or scene build is ready to be rendered.
+    // sp that the tab-switching integration can react adequately when tab
+    // switching takes too long. For this use case what matters is that the
+    // notification doesn't fire before scene building.
```
kvark: it's not clear from the description if it fires after blob rasterization or after the scene building

webrender_api/src/api.rs
```diff
@@ -171,6 +175,18 @@ impl Transaction {
         self.merge(resources);
     }
 
+    // Note: Gecko uses this to get notified when a transaction that contains
+    // potentially long blob rasterization or scene build is ready to be rendered.
+    // sp that the tab-switching integration can react adequately when tab
```
kvark: "sp"

webrender/src/render_backend.rs
```diff
@@ -1028,9 +1029,11 @@ impl RenderBackend {
             self.result_tx.send(msg).unwrap();
         }
 
-        for evt in notifications {
-            self.notifier.external_event(evt)
-        }
+        drain_filter(
+            &mut notifications,
+            |n| { n.when() == CheckPoint::AfterFrameBuilding },
```
kvark: nit: `CheckPoint` should be [a single word](https://en.wikipedia.org/wiki/Checkpoint)

webrender_api/src/api.rs
```diff
@@ -1161,3 +1167,48 @@ pub trait RenderNotifier: Send {
     }
     fn shut_down(&self) {}
 }
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum CheckPoint {
+    AfterSceneBuilding,
```
kvark: Maybe `SceneBuilt` and `FrameBuilt`?

nical:
@kvark are we good to go?

kvark:
@nical yes, if you are confident that a try push isn't needed here :)

PR #3021, opened 2018-09-05
cs_border_segment: Don't expand the radius for very small dots.

emilio:
Otherwise we'd do antialiasing between the dots and it'd really look like a
solid line.

This is important for non-hidpi screens, for outlines and such in particular. I
can fix this also in the fragment shader after applying the other clips if you
think it's better, but this felt a bit less hacky.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1434006.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3021)


kvark:
> thread 'main' panicked at 'size=1920×1080 ws=1916×1053', wrench\src\reftest.rs:496:9

We need to reduce the reftest image size

emilio:
@bors-servo r=kvark

Thanks for the hint! I cropped it a bit.

kvark:
heh, did your crop blurred it? :)
> REFTEST TEST-UNEXPECTED-FAIL | reftests\border\small-dotted-border.yaml == reftests\border\small-dotted-border.png | image comparison, max difference: 128, number of differing pixels: 196

emilio:
Hmm, not really, in fact I even re-run the PNG command in (headless) wrench...

kvark:
@emilio I assume reftests pass for you locally? (`cd wrench && script/headless.py reftest`)

emilio:
Yes of course

emilio:
Well, to be honest there's a very small off-by-one in some text-related reftests. But I think those are not relevant.

emilio:
@bors-servo r=kvark

 * Disabled the test on windows, as many other border reftests... Do we know why?

PR #3022, opened 2018-09-05
cs_border_solid: Simplify a bit more the simple border shader.

emilio:
We shouldn't need to care about clip modes either, since those are only for
dotted / dashed borders.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3022)

### DIFF
--- a/webrender/res/cs_border_solid.glsl
+++ b/webrender/res/cs_border_solid.glsl
@@ -13,7 +13,7 @@ flat varying vec4 vColor1;
 // transition occurs. Used for corners only.
 flat varying vec4 vColorLine;
 
-// x = segment, z = edge axes, w = clip mode
+// x = segment, z = edge axes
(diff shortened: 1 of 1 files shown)


emilio:
r? @kvark or @gw3583 

cc @djg

webrender/res/cs_border_solid.glsl
```diff
@@ -186,29 +168,6 @@ void main(void) {
     bool in_clip_region = all(lessThan(vClipCenter_Sign.zw * clip_relative_pos, vec2(0.0)));
     float d = -1.0;
```
kvark: nit: later on, we don't need `d = max(d, d_radii)`

emilio: Nice catch, thanks :)

PR #3028, opened 2018-09-05
Rewrite the blur shader to remove the variable-length loop and to use the texture filtering hardware more effectively.

### LINKED ISSUE #2821: Implement blur filter with dual kawase technique
See https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-26-50/siggraph2015_2D00_mmg_2D00_marius_2D00_notes.pdf for details of quality / performance improvements.

pcwalton:
This new shader performs Gaussian *resampling* instead of regular convolution.
It samples in between texels to reduce the number of taps.

The speed is about the same as the existing technique. It is nevertheless the
fastest blur method that I could come up with. In particular, it exceeds the
performance of the Kawase and dual Kawase blur techniques. The speed comes from
working at lower resolution and incurring fewer downsampling and upsampling
passes. As is often the case, ALU performance does not really seem to be the
limiting factor; it's mostly memory bandwidth, which is why downsampling is so
important. Further improvements should come from not doing the downsampling at
all and instead rendering the blurred content at low resolution to begin with.

Closes #2821.

Some reftests are currently failing; do not merge yet. I just wanted to get this here to discuss. cc @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3028)


pcwalton:
Just one reftest is failing and needs investigation. (It may just be a harmless minor blur algorithm difference, but I want to be sure.)

kvark:
Is there anything special going on during upsampling in order to preserve the missing detail?

pcwalton:
OK, this should be ready for review, though we should do a Gecko try run as well. r? @gw3583

kvark:
Launched a try in https://treeherder.mozilla.org/#/jobs?repo=try&revision=94b30b8a56e652bf8e653d1165f33b7e7acc6a3f

kvark:
> There's no (well, negligible) missing detail because Gaussian blur is a low-pass filter. (The Fourier transform of a Gaussian convolution is itself a Gaussian with the characteristic bell curve shape.) That's why downsampling works in the first place. Because the output of a Gaussian blur is effectively bandlimited, the Nyquist-Shannon sampling theorem says that we can preserve fewer samples in order to perfectly reconstruct the signal.

We aren't perfectly reconstructing the signal, if I understand correctly, it's still an approximation.
Some light reading with pictures for those interested - http://chemaguerra.com/downsampling-and-gaussian-filters/

kvark:
Weirdly Github doesn't show the CI icon for your latest commit. Could happen if you changed the date and/or committed from a different time zone?

webrender/res/cs_blur.glsl
```diff
@@ -5,131 +5,120 @@
...
 
 struct BlurTask {
     RenderTaskCommonData common_data;
-    float blur_radius;
+    vec2 coefficients;
+    int direction;
 };
 
-BlurTask fetch_blur_task(int address) {
+BlurTask fetchBlurTask(int address) {
```
kvark: please stick to the existing naming convention in the shaders

webrender/res/cs_blur.glsl
```diff
@@ -5,131 +5,120 @@
...
-    gl_Position = uTransform * vec4(pos, 0.0, 1.0);
+    gl_Position = uTransform * vec4(position, 0.0, 1.0);
 }
+
 #endif
 
 #ifdef WR_FRAGMENT_SHADER
 
-#if defined WR_FEATURE_COLOR_TARGET
+#define SUPPORT     4
```
kvark: it's not clear what `SUPPORT` here means

webrender/res/cs_blur.glsl
```diff
@@ -5,131 +5,120 @@
...
-    gauss_coefficient.y = exp(-0.5 / (vSigma * vSigma));
-    gauss_coefficient.z = gauss_coefficient.y * gauss_coefficient.y;
-
-    float gauss_coefficient_sum = 0.0;
-    SAMPLE_TYPE avg_color = original_color * gauss_coefficient.x;
-    gauss_coefficient_sum += gauss_coefficient.x;
-    gauss_coefficient.xy *= gauss_coefficient.yz;
+    bool vertical = vVertical != 0;
+    vec2 axisCoord = vertical ? vUv.yx : vUv.xy;
+    float start = floor(axisCoord.x - float(SUPPORT)) + 0.5;
```
kvark: if the flooring changes our start position, wouldn't the computed Gaussian factors no longer match?

webrender/src/clip.rs
```diff
@@ -274,7 +274,7 @@ impl ClipNode {
                 // Quote from https://drafts.csswg.org/css-backgrounds-3/#shadow-blur
                 // "the image that would be generated by applying to the shadow a
                 // Gaussian blur with a standard deviation equal to half the blur radius."
-                let blur_radius_dp = (info.blur_radius * 0.5 * device_pixel_scale.0).round();
+                let blur_radius_dp = (info.blur_radius * device_pixel_scale.0).round();
```
kvark: comment above needs to be addressed as well

webrender/src/render_task.rs
```diff
@@ -536,62 +554,55 @@ impl RenderTask {
...
-                downscaling_src_task_id,
-                adjusted_blur_target_size,
-            );
-            downscaling_src_task_id = render_tasks.add(downscaling_task);
+        // Look up some info.
+        let full_size = render_tasks[src_task_id].get_dynamic_size();
+        let uv_rect_kind = render_tasks[src_task_id].uv_rect_kind();
+
+        // Determine the number of downsampling passes that we'll use.
+        let downsampling_passes = (30 as u32).saturating_sub((radius as u32).leading_zeros());
```
kvark: number of downsample passes = 30 - (leading 0 bits of radius)?
where does 30 come from?

webrender/src/renderer.rs
```diff
@@ -2938,30 +2917,18 @@ impl Renderer {
...
+        // Blurs are rendered using a two-pass separable implementation, with a fixed support size.
+        if !target.blurs.is_empty() {
             let _timer = self.gpu_profile.start_timer(GPU_TAG_BLUR);
 
             self.set_blend(false, framebuffer_kind);
             self.shaders.cs_blur_rgba8
                 .bind(&mut self.device, projection, &mut self.renderer_errors);
 
-            if !target.vertical_blurs.is_empty() {
+            if !target.blurs.is_empty() {
```
kvark: we already check for this condition in an enclosing block

webrender/res/cs_blur.glsl
pcwalton: I made that change to improve consistency. Do function names have `snake_case` and variable names have CamelCase? I never was able to figure it out.

kvark: Look at `prim_shared.glsl` for examples. It's pretty much the Rust naming convention:
  - variables and functions are `snake_case`
  - type names are `CamelCase`

webrender/res/cs_blur.glsl
pcwalton: The computed Gaussian factors are based on `offset`, which is relative to the un-floored `axisCoord.x`.

pcwalton:
I should mention that we shouldn't be downsampling any more than the old code did.

I noticed a bunch of test failures on try (thanks for kicking off the try run!)—I'll investigate those.

pcwalton:
Given all the test failures on try, I'm going to try separating the new shader from the downscaling changes to try to isolate the source of the problems.

gw3583:
@pcwalton Do we still want this kept open?

gw3583:
@pcwalton #3122 has landed now. Should we keep this open and rebase on top of those changes, or would it be best to close this and open a new PR when ready?

PR #3034, opened 2018-09-08
WIP: Update base64

Eijebong:


atouchet:
base64 0.9.3 is now out.

kvark:
@pyfisch why is this WIP?

Eijebong:
@kvark Because it would block updating wr in servo if it was merged while hyperup isn't ready to land

gw3583:
@Eijebong What's the status of this?

Eijebong:
@gw3583 No real changes unfortunately; Tests are green on servo's side, need to make it mergeable but it still requires a ton of work (and I'm lacking motivation these days :/)

Eijebong:
Ok, ready, let's do this :tada:

PR #3035, opened 2018-09-09
Remove an unused VAO.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3035)


gw3583:
Minor fix I noticed while working on the 3d transforms WIP.

PR #3036, opened 2018-09-10
Ensure transactions that request a frame always get forwarded to the renderer.

nical:
Gecko depends on transactions that request frames to be passed all the way to the renderer, even if the frame can't be built or doesn't need to be built (nothing visibly changed). If one of these transaction is dropped before getting to the renderer, Gecko might wait for the notification that this transaction was rendered forever and not send new frames, and nothing ever gets rendered again.
This type of race tends to manifest itself on windows a lot more than on other platforms, probably because of the way the GPU process affects timings.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3036)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -908,6 +908,7 @@ impl RenderBackend {
         profile_counters: &mut BackendProfileCounters,
         has_built_scene: bool,
     ) {
+        let requested_frame = render_frame;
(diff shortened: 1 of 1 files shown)


nical:
The first commit in this PR is already reviewed and landing in Gecko, it should fix the Windows bustage. The second commit makes extra sure we don't skip `update_document`.

PR #3037, opened 2018-09-10
Remove render_on_scroll.

nical:
It was used by tests before we had transactions but it isn't useful anymore.

This commit should not change the current behavior. Turns out we weren't taking advantage of the knowledge of whether something actually scrolled for anything other than the render_on_scroll mechanism.
This is useful information, though, that should help with deciding if/when we need to re-build the frame, so I left it visible in unused `_scrolled` variables. We should take advantage of these in followup work to remove redundant frame building.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3037)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -97,12 +97,6 @@ struct Document {
     // A set of pipelines that the caller has requested be
     // made available as output textures.
     output_pipelines: FastHashSet<PipelineId>,
-    // A helper switch to prevent any frames rendering triggered by scrolling
-    // messages between `SetDisplayList` and `GenerateFrame`.
-    // If we allow them, then a reftest that scrolls a few layers before generating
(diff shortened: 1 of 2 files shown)


kvark:
If the tests still work, we are good :)
Thank you!
@bors-servo r+

gw3583:
I think that Servo may still be relying on this behavior - so we might need to write up some brief details on what Servo will need to do with transactions to emulate this behavior (if anything).

nical:
> I think that Servo may still be relying on this behavior - so we might need to write up some brief details on what Servo will need to do with transactions to emulate this behavior (if anything).

A quick search for `render_on_scroll` on servo's github repository didn't yield anything but in any case, if servo uses it, the only thing to add is a call to `generate_frame` on the transactions that contain scrolling operations before sending them.

Issue #2828, opened 2018-06-19
Border arrows are rendered incorrectly


freaktechnik:
Making arrows with just one border side set to actually be visible is broken:

https://getbootstrap.com/docs/3.3/components/#dropdowns

kvark:
Could you specify more precisely where the issue is visible on the page? Comparing our rendering with Chromium, I could see that the shadow is a little bit denser, but nothing striking as broken.

freaktechnik:
The "arrow" with the carret class points to the top right instead of down for me. Though looking at it working when I take a screenshot, this may be a platform issue and I should raise this on bugzilla.

Firefox screenshot:
![image](https://user-images.githubusercontent.com/640949/41614609-3e2b8686-73f9-11e8-91c3-a4849483b37a.png)

Native screenshot:
![image](https://user-images.githubusercontent.com/640949/41614638-4aea1874-73f9-11e8-9b5b-fe6c58c20a6e.png)

kvark:
Looks fine on my Linux Nightly.

freaktechnik:
This is on the latest linux nightly, too. With an nVidia GPU.

The "firefox screenshot does not match native screenshot" thing was exactly the same with the text shadow issue.

staktrace:
The "firefox screenshot" doesn't use webrender, so none of the webrender-codepath bugs will appear in those screenshots.

staktrace:
(Also, I can reproduce the problem on macOS nightly)

gw3583:
I wonder if this will be fixed by https://github.com/servo/webrender/pull/2864 ?

Darkspirit:
Nope. Still broken on Debian Testing, GTX 1060.
mozregression --repo [try](https://bugzilla.mozilla.org/show_bug.cgi?id=1473284#c3) --launch [8579dbe6d0db61290f9a34073ba6401b52214fe4](https://treeherder.mozilla.org/#/jobs?repo=try&revision=8579dbe6d0db61290f9a34073ba6401b52214fe4) --pref gfx.webrender.all:true startup.homepage_welcome_url:'https://getbootstrap.com/docs/3.3/components/#dropdowns'

kvark:
I can repro (now?) on Linux. Took a capture from a reduced test case locally.

cpearce:
Minimized test case: https://cpearce.github.io/html-test-cases/webrender-issue-2828-testcase.html

Something like:

```
      <style>
        .caret {
            border-top:4px dashed;
[...]

PR #3039, opened 2018-09-10
Make dashed-border rendering more similar to Gecko / other browsers.

emilio:
This mostly focuses on non-cornered dashed borders.

The worst case of this algorithm is the one where the borders are rounded and the space between corners (the solid segment) is small but non-zero.

Thus I'm a bit on the fence on the fourth commit... I could preserve the current algorithm for the rounded corner cases.

The tl;dr is that this makes a dashed segment always start with half a dash, which matches other browsers (modulo a TODO where other browsers start with the whole segment, and border-radius when each browser does something completely different so I suspect we may be a bit more lenient).

I think we should do this for now if it doesn't cause problems, though it's trivial to extend it to do the same as Gecko for discontinued borders, etc. It's just a bit more annoying because instead of passing a flag I need to pass the offset around in a new float value.

I ended up shifting the brush in `brush_picture.glsl`, where the rest of the repeating code is handled. That being said, I've done the transformation in local space, but there may be a better way to do it.

There are some cleanups to the border code that I can do if this is a decent / acceptable approach.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3039)


gw3583:
It's probably not ideal that we need some border-specific code / flags in the common brush.glsl code, which I guess is what you were talking about on IRC?

What we actually do for dashes in corners now
https://github.com/servo/webrender/blob/9a2fbef7d29af4ce3cf394767913f09026a459e4/webrender/src/border.rs#L1096
is generate an instance for each dash / dot. We do this for corners because the size of the dash / dot varies with each instance as traced around the corner.

I think what we probably want to do is something similar for dots / dashes in the border edge segments. This would give us complete control over specifying a local rect for each of the dashes, without affecting the common brush code.

Previously, we wanted to do it in the shader to avoid generating a lot of instances per frame, but we now cache border segments to the texture cache, and re-use them between frames, so this isn't really a concern now.

Having said all that, given our current deadlines and bug list - it's probably fine to merge a solution like this and do the work above as a follow up later on - thoughts?

emilio:
Try is green and this of course fixes #2828 and such.

Need to cleanup and remove the now unused flags and such which I'll do tomorrow, but @gw3583 does this look reasonable to you?

emilio:
A little bit of a lie: It's green modulo some panic which is fixed in a commit here already.

Previous try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=706f34643c938dbeafaf4db23d332b4bba64592c&selectedJob=198528498

New one: https://treeherder.mozilla.org/#/jobs?repo=try&revision=d28f4117d92954f19cf509d5aef4e2542fd71983

PR #3040, opened 2018-09-10
Check preserve-3d when computing stacking-context backface-visibility

Gankro:
this appears to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1489937

try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=386d8adb469cb458af4b5359de63c9a067345e22

(that some other should-be-fine commits snuck onto)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3040)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -1075,6 +1075,22 @@ impl<'a> DisplayListFlattener<'a> {
             );
         }
 
+        // preserve-3d's semantics are to hoist all your children to be your siblings
+        // when doing backface-visibility checking, so we need to grab the backface-visibility
+        // of the lowest ancestor which *doesn't* preserve-3d, and AND it in with ours.
+        //
+        // No this isn't obvious or clear, it's just what we worked out over a day of testing.
+        // There's probably a bug in here, but I couldn't find it with the examples and tests
+        // at my disposal!
+        let ancestor_is_backface_visible =
+            self.sc_stack
+                .iter()
+                .rfind(|sc| sc.transform_style == TransformStyle::Flat)
+                .map(|sc| sc.is_backface_visible)
+                .unwrap_or(is_backface_visible);
+
+        let is_backface_visible = is_backface_visible && ancestor_is_backface_visible;
+
         // Push the SC onto the stack, so we know how to handle things in
         // pop_stacking_context.
         let sc = FlattenedStackingContext {
--- a/wrench/reftests/transforms/complex-preserve-3d.yaml
+++ b/wrench/reftests/transforms/complex-preserve-3d.yaml
@@ -1,0 +1,24 @@
+# the root sc should dominate backface-visibility and hide the rect
+
+---
+root:
+  items:
+    -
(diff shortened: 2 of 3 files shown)


Gankro:
New impl, new try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=454e91fbffc1f642362eab3df8edc5136dd2cf61

Writing a test now (since nothing in gecko's suite is robust enough to catch this issue, sadly)

webrender/src/display_list_flattener.rs
```diff
@@ -1029,6 +1029,21 @@ impl<'a> DisplayListFlattener<'a> {
             None
         };
 
+        // preserve-3d's semantics are to hoist all your children to be your siblings
```
kvark: this doesn't sound correct to me

webrender/src/display_list_flattener.rs
```diff
@@ -1029,6 +1029,21 @@ impl<'a> DisplayListFlattener<'a> {
...
+        // when doing backface-visibility checking, so we need to grab the backface-visibility
+        // of the lowest ancestor which *doesn't* preserve-3d, and AND it in with ours.
+        //
+        // No this isn't obvious or clear, it's just what we worked out over a day of testing.
+        // There's probably a bug in here, but I couldn't find it with the examples and tests
+        // at my disposal!
+        let mut ancestor_is_backface_visible = is_backface_visible;
+        for sc in self.sc_stack.iter().rev() {
```
kvark: nit:
```rust
let ancestor_is_backface_visible = self.sc_stack
  .iter()
  .rev()
  .find(|sc| sc.transform_style == TransformStyle::Flat)
  .map(|sc| sc.is_backface_visible)
  .unwrap_or(is_backface_visible);
```

webrender/src/display_list_flattener.rs
```diff
@@ -1029,6 +1029,21 @@ impl<'a> DisplayListFlattener<'a> {
             None
         };
 
+        // preserve-3d's semantics are to hoist all your children to be your siblings
+        // when doing backface-visibility checking, so we need to grab the backface-visibility
+        // of the lowest ancestor which *doesn't* preserve-3d, and AND it in with ours.
+        //
+        // No this isn't obvious or clear, it's just what we worked out over a day of testing.
```
kvark: I'm sorry, I don't understand why we are looking for the nearest Flat ancestor. Is there an actual reason backing this logic, or just empirical evidence?

webrender/src/display_list_flattener.rs
```diff
@@ -1029,6 +1029,21 @@ impl<'a> DisplayListFlattener<'a> {
             None
         };
 
+        // preserve-3d's semantics are to hoist all your children to be your siblings
+        // when doing backface-visibility checking, so we need to grab the backface-visibility
+        // of the lowest ancestor which *doesn't* preserve-3d, and AND it in with ours.
```
kvark: I think the proper thing to do would be to grab the closest ancestor that is still `Preserve3D` but *whose parent* is `Flat`

PR #3041, opened 2018-09-11
Avoid allocating clip tasks for clip rects in same coordinate system.

gw3583:
If a clip chain only ends up with clip rects that are from the
same coordinate system, there is no need to allocate a clip mask,
as the clip can be implemented via the local clip rect in the
primitive vertex shader.

This is an important optimization for reducing clip task
allocations on common pages (although they won't actually be
drawn into the clip mask anyway).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3041)


gw3583:
r? @kvark

Try run looks good (matches current baseline):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=8d16b486151b583205ff8372aab80840cdeaf066

PR #3042, opened 2018-09-11
Refactor flattener for mix-blend-mode optimizations.

gw3583:
This patch doesn't contain any user visible changes. It does
change the display list flattener to append pictures to their
parent picture during pop_stacking_context instead of during
push_stacking_context.

This not only makes the code clearer, it also makes it possible
to introduce some optimizations to mix-blend-mode in the
future. Specifically, we want to detect mix blend mode, and
insert a new picture node in the tree which re-parents the
existing stacking context, in some cases.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3042)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -107,9 +107,6 @@ pub struct DisplayListFlattener<'a> {
     /// A stack of stacking context properties.
     sc_stack: Vec<FlattenedStackingContext>,
 
-    /// A stack of the current pictures.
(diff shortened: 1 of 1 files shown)


gw3583:
r? @kvark 

Try run looks good (matches current baseline):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=9b3210e94dcd5d330a623a58ded5d75059941738&selectedJob=198517446

gw3583:
@kvark Thanks! Comments addressed and pushed an updated commit.

PR #3043, opened 2018-09-11
Track the validity of the frame and hit tester to avoid redundant work.

nical:
This lays the ground work for keeping track of the validity of the current frame and hit tester to allow only re-building them when something invalidates them.

This initial commit only skips redundant frame builds (typically when a scene is built and the vsync triggered transactions that follows doesn't scroll or use async animations). But after we separate building the frame and the hit tester, we'll be able to improve upon this by lazily building the frame and avoid double frame builds during scrolling and animations as well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3043)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -956,8 +956,9 @@ impl RenderBackend {
             render_frame = false;
         }
 
-        // If we don't generate a frame it makes no sense to render.
-        debug_assert!(build_frame || !render_frame);
+        if doc.frame_is_valid {
+            build_frame = false;
+        }
 
         let mut frame_build_time = None;
         if build_frame && doc.has_pixels() {
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -105,6 +105,11 @@ struct Document {
     /// Properties that are resolved during frame building and can be changed at any time
     /// without requiring the scene to be re-built.
     dynamic_properties: SceneProperties,
+
+    /// Track whether the last built frame is up to date or if it will need to be re-built
+    /// before rendering again.
+    frame_is_valid: bool,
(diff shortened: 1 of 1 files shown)


nical:
try push: https://treeherder.mozilla.org/#/jobs?repo=try&selectedJob=198630924&revision=f8c819bbfcb0f2576a028645fe911d18f348dcce

Also worth exploring as a followup: query the visibility of updated images so that we can skip frame building and rendering altogether if something animates off screen.

nical:
cc @sotaroikeda @kvark

webrender/src/render_backend.rs
```diff
@@ -105,6 +105,11 @@ struct Document {
     /// Properties that are resolved during frame building and can be changed at any time
     /// without requiring the scene to be re-built.
     dynamic_properties: SceneProperties,
+
+    /// Track whether the last built frame is up to date or if it will need to be re-built
+    /// before rendering again.
+    frame_is_valid: bool,
+    hit_tester_is_valid: bool,
```
kvark: if it's not valid, can we just have `hit_tester == None`?

nical: If we do this we need to first process all messages that use the hit tester before setting it to `None`. for example if you have a transaction that both updates and uses the hit tester, right now the rule is that hit testing operations use the previous hit tester and we update the hit tester at the end.
So it'll be a bit more fiddly but I think it's a good idea.

nical: Alternatively, I'm leaning towards replacing the `Option<HitTester>` with a bare `HitTester` containing a valid boolean flag. This would let us recycle it's allocations instead of reallocating each time. Thoughts?

kvark: everything for performance! :)

PR #3046, opened 2018-09-12
Handle scale + offset as same coordinate system.

gw3583:
Apart from simple translations, the most common transform that
is encountered in WR is a scale + offset combination. This is
used especially heavily on animations, and hi-dpi screens.

Previously, WR would consider this a different coordinate
system, which means falling back to clip masks for many
cases (e.g. a rectangle in root coordinate system that has
a partial overlap with a primitive in a scaled coord system).

However, these scale + offset transforms still result in an
axis-aligned rectangle between the two coordinate systems. Thus,
we can actually consider them to be the same coordinate system.

The outcome of this is that we can apply clip rectangles as
local clip rects in the vertex shader. This greatly reduces
the number of clip masks that we allocate in such situations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3046)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -17,7 +17,7 @@ use prim_store::{ClipData, ImageMaskData, SpaceMapper};
 use render_task::to_cache_size;
 use resource_cache::{ImageRequest, ResourceCache};
 use std::{cmp, u32};
-use util::{extract_inner_rect_safe, pack_as_float, project_rect, recycle_vec};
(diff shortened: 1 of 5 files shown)


gw3583:
I'll kick off another try run shortly, to make sure I didn't break anything when addressing those comments.

gw3583:
Try run with those changes is still good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d5bfdda6179728168db31995542e8e1398465550

gw3583:
@kvark Ah, fair enough - I'll make that change as a follow up. Thanks!

PR #3048, opened 2018-09-12
clip text against the local clip rect after snapping rather than before

lsalzman:
This partially fixes Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1477625

Previously, we took the glyph rect within the atlas and then added the glyph offset with subpixel offset and all onto it, and critically, clipped it first, then snapped it second.

The problem is the glyph rect in the atlas represents a conservative bounding box of the glyph. So if you add an offset to it, which might cause it to go outside the clip and snip it. Then the snapping pushes it back inside the bounds, so that the snipped portion is actually still inside the clip, leaving the glyph wrongly snipped.

The way to fix this, while not quite as pretty as the previous code, is to ensure that we snap the glyph rect into place in device space first, then transform it back to local space to perform the clipping. This way, snapping is first, clipping is second. I've managed to keep the amount of math comparable for performance, and maybe the snapping code is even a bit simpler than before due to necessarily snapping the glyph rect origin rather than individual corners.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3048)


lsalzman:
Try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=f5934ff5de4d1d9ff2acaf9a96d9a680feda1129

webrender/res/ps_text_run.glsl
```diff
@@ -61,64 +61,82 @@ TextRun fetch_text_run(int address) {
...
 
-#if defined(WR_FEATURE_GLYPH_TRANSFORM)
-    bool remove_subpx_offset = true;
-#else
-    bool remove_subpx_offset = transform.is_axis_aligned;
-#endif
+#ifndef WR_FEATURE_GLYPH_TRANSFORM
     // Compute the snapping offset only if the scroll node transform is axis-aligned.
-    if (remove_subpx_offset) {
+    if (transform.is_axis_aligned) {
```
kvark: can we please not change the block structure/indentation by the preprocessor branches? let's have a boolean to check in `if` uniformly

webrender/res/ps_text_run.glsl
```diff
@@ -61,64 +61,82 @@ TextRun fetch_text_run(int address) {
...
+#endif
+        // Transform from local space to device space.
+        float device_scale = uDevicePixelRatio / transform.m[3].w;
+        mat2 device_transform = mat2(transform.m) * device_scale;
```
kvark: are we working within an assumption that `transform` is non-perspective here?

webrender/res/ps_text_run.glsl
```diff
@@ -61,64 +61,82 @@ TextRun fetch_text_run(int address) {
...
+    RectWithSize local_rect = transform_rect(glyph_rect, local_transform);
+
+    // Select the corner of the glyph's local space rect that we are processing.
+    vec2 local_pos = local_rect.p0 + local_rect.size * aPosition.xy;
+
+    // If the glyph's local rect would fit inside the local clip rect, then select a corner from
+    // the device space glyph rect to reduce overdraw of clipped pixels in the fragment shader.
+    // Otherwise, fall back to clamping the glyph's local rect to the local clip rect.
+    local_pos = rect_inside_rect(local_rect, local_clip_rect) ?
+                    local_transform * (glyph_rect.p0 + glyph_rect.size * aPosition.xy) :
```
kvark: would it be straightforward to just clamp that against `local_clip_rect` unconditionally?

webrender/res/ps_text_run.glsl
```diff
@@ -61,64 +61,82 @@ TextRun fetch_text_run(int address) {
...
+        vec2 device_text_pos = device_transform * text_offset + transform.m[3].xy * device_scale;
+        snap_offset = floor(device_text_pos + 0.5) - device_text_pos;
+
+        // Snap the glyph offset to a device pixel, using an appropriate bias depending
+        // on whether subpixel positioning is required.
+        vec2 device_glyph_offset = device_transform * glyph_offset;
+        snap_offset += floor(device_glyph_offset + snap_bias) - device_glyph_offset;
+
+        // Transfrom from device space back to local space.
+        mat2 local_transform = inverse(device_transform);
```
kvark: for an axis-aligned transform, this should be equal to `transpose`

lsalzman:
Added a reftest

webrender/res/ps_text_run.glsl
kvark: We can always just have stuff declared but not defined *outside* of the block, and only used inside:
```glsl
vec4 some_value;
if (condition) {
   some_value = xxx;
}
if (condition) {
   yyy = some_value;
}
```
The compiler can see that we aren't using uninitialized variables, and we don't have to sacrifice the block structure/control flow for it.

webrender/res/ps_text_run.glsl
kvark: clamping `local_transform * (glyph_rect.p0 + glyph_rect.size * aPosition.xy)` would result in overdraw? I thought that's the tighter bound alrady

kvark:
@lsalzman just to clarify, the test is failing without your changes, right?

lsalzman:
> @lsalzman just to clarify, the test is failing without your changes, right?

Of course

webrender/res/ps_text_run.glsl
lsalzman: I thought you meant to only clamp against the local clip rect. I guess that's possible.

lsalzman:
Did suggested cleanups.

kvark:
Thanks, shipit!
@bors-servo r+

PR #3050, opened 2018-09-12
Don't cache pictures if they contain an external image

mattwoodrow:
Fixes bug https://bugzilla.mozilla.org/show_bug.cgi?id=1482727

Videos can be updated externally without a scene build, so we can't cache Pictures that contain them.

This detects when a picture contains an external image, and disables caching.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3050)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -24,7 +24,7 @@ use prim_store::{BorderSource, Primitive, PrimitiveDetails};
 use render_task::{RenderTaskAddress, RenderTaskId, RenderTaskTree};
 use renderer::{BlendMode, ImageBufferKind, ShaderColorMode};
 use renderer::BLOCKS_PER_UV_RECT;
-use resource_cache::{CacheItem, GlyphFetchResult, ImageRequest, ResourceCache};
+use resource_cache::{CacheItem, GlyphFetchResult, ImageRequest, ResourceCache, ImageProperties};
 use scene::FilterOpHelpers;
 use std::{f32, i32};
 use tiling::{RenderTargetContext};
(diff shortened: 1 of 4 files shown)


webrender/src/batch.rs
```diff
@@ -1604,6 +1604,34 @@ impl Primitive {
...
+            }
+            PrimitiveDetails::Brush(BrushPrimitive { kind: BrushKind::YuvImage{ yuv_key, .. }, .. }) => {
+                Some(yuv_key[0])
+            }
+            PrimitiveDetails::Brush(_) |
+            PrimitiveDetails::TextRun(..) => {
+                None
+            }
+        };
+        if let Some(image_key) = image_key {
```
gw3583: One last nit - we could probably write this as something like:

```
        image_key.map_or(true, |image_key| {
            match resource_cache.get_image_properties(image_key) {
                Some(ImageProperties { external_image: Some(_), .. }) => {
                    false
                }
                Some(_) |
                None => {
                    true
                }
            }
        })
```

Which would remove the need for the last true and any early return statements. Then after squashing this should be good to merge, thanks!

gw3583:
@bors-servo r+

Thanks!

webrender/src/batch.rs
```diff
@@ -1604,6 +1604,33 @@ impl Primitive {
             }
         }
     }
+
+    pub fn is_cacheable(
+        &self,
+        resource_cache: &mut ResourceCache
```
kvark: this shouldn't be `&mut`

webrender/src/batch.rs
```diff
@@ -1604,6 +1604,33 @@ impl Primitive {
...
+            PrimitiveDetails::Brush(_) |
+            PrimitiveDetails::TextRun(..) => {
+                return true
+            }
+        };
+        match resource_cache.get_image_properties(image_key) {
+            Some(ImageProperties { external_image: Some(_), .. }) => {
+                false
+            }
+            Some(_) |
```
kvark: nit: could just be `_ => true`

webrender/src/prim_store.rs
```diff
@@ -1653,6 +1653,10 @@ impl PrimitiveStore {
                     Some(pic_rect)
                 };
 
+                if !pic_state_for_children.is_cacheable {
+                  pic_state.is_cacheable = false;
```
kvark: why aren't children cachable if the parent is not? If a picture contains an external image and another picture, that other picture might as well be cached, no?

mattwoodrow: The intention is the reverse: If a picture contains a picture, and that inner picture contains an external image, then neither picture is cacheable.

kvark:
Thanks @mattwoodrow !
@bors-servo r=gw3583,kvark

PR #3052, opened 2018-09-13
Speed up updating the hit-test tree

nical:
The first two commits take advantage of what was put in place in #3043 to avoid building the frame when all we need is the hit tester.
The next two commits pick some low hanging optimization fruits in the hit tester to paper over a few minor talos regressions from the first commits.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3052)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -291,7 +291,9 @@ impl ClipScrollTree {
         // TODO(gw): This is an ugly borrow check workaround to clone these.
         //           Restructure this to avoid the clones!
         let mut state = state.clone();
-        let node_children = {
+        let mut node_children: SmallVec<[SpatialNodeIndex; 4]> = SmallVec::new();
+
+        {
             let node = match self.spatial_nodes.get_mut(node_index.0) {
                 Some(node) => node,
                 None => return,
@@ -307,8 +309,8 @@ impl ClipScrollTree {
             }
 
             node.prepare_state_for_children(&mut state);
-            node.children.clone()
-        };
+            node_children.extend_from_slice(&node.children);
+        }
 
         for child_node_index in node_children {
             self.update_node(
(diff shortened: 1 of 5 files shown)


nical:
Each commit can be reviewed independently.

Talos results from avoid extra frame builds: https://treeherder.mozilla.org/perf.html#/compare?originalProject=try&originalRevision=c8a1ae299661&newProject=try&newRevision=c851a94a67beadc74362309cce18f73b2ee1f2cd&framework=1&showOnlyComparable=1

The biggest win is unsurprisingly displaylist_mutate with 16%-21% improvement.

Talos results from adding the next two commits: https://treeherder.mozilla.org/perf.html#/compare?originalProject=try&originalRevision=c851a94a67be&newProject=try&newRevision=1d59b6336976240b25709da2b97fb3581e460230&framework=1

Notice an unexpectedly spectacular 37% improvement in tp6_youtube, just from removing some small allocations.

Overall talos results from the 4 commits: https://treeherder.mozilla.org/perf.html#/compare?originalProject=try&originalRevision=c8a1ae299661&newProject=try&newRevision=1d59b6336976240b25709da2b97fb3581e460230&framework=1&showOnlyComparable=1

There are some regressions which I can't explain, for example 3.3% tp5o_scroll which shouldn't have been affected much or if anything should have improved due to the reduced number of allocations when updating the clip scroll tree. The regression is small compared to the winds overall.

This replaces #2936.

kvark:
I launched a try push with just your changes (before the review fixes), and it has failures: https://treeherder.mozilla.org/#/jobs?repo=try&revision=cf544e432dbce2b2375c3d99ceafd6f4baa4c1b1&selectedJob=199154816
Please have a look, we can't land this in the meantime.

kvark:
Oh snap, I thought the CI has stopped because of a network error on one of the targets, but it got ignored.

zptan:
@nical These commits have been reverted in #3061, could you please tell us if you are working on the fix?

nical:
I am working on the fix. It turns out these commits uncover a race condition in the code that handles blob image updates so I have to fix that one before I can re-land this, but that's my current priority.

PR #3054, opened 2018-09-13
border: Get rid of BorderCornerClipSource.

emilio:
Instead, just add functions to add the instances directly, avoiding the
intermediate allocation and deindenting a bunch of the code.

This should have no behavior change.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3054)


emilio:
r? @kvark or anyone else?

I'm sorry, github is terrible at flagging indentation changes. :(

emilio:
> nit: could use vector operations?

I couldn't do this:  no implementation for `{float} - euclid::TypedVector2D<f32, webrender_api::DevicePixel>`

emilio:
(Rest of comments should be addressed)

PR #3049, opened 2018-09-12
Gracefully avoid generating empty border segments

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1489644
Also related to https://github.com/servo/webrender/issues/2715#issuecomment-419948326

The change makes us avoid doing meaningless work and then stumbling upon unexpected data. Good for both performance, stability, and the health of the users ;)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3049)

PR #3056, opened 2018-09-13
border: Avoid generating empty segments.

emilio:
This is a bit less invasive and should do the job.

I've confirmed that all the negative sizes came from edges, when a huge radii
with radius > rect.size - cw_border.width - ccw_border.width caused avail_width to be negative.

Still we're creating a gazillion empty borders which are useless, so prevent
that too by checking in add_edge_segment and add_corner_segment.

I've confirmed that this passes the reftests that are failing in #3049, but waiting for try just to double-check.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3056)

### DIFF
--- a/webrender/src/border.rs
+++ b/webrender/src/border.rs
@@ -4,7 +4,7 @@
 
 use api::{BorderRadius, BorderSide, BorderStyle, BorderWidths, ColorF};
 use api::{ColorU, DeviceRect, DeviceSize, LayoutSizeAu, LayoutPrimitiveInfo, LayoutToDeviceScale};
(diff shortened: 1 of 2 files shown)


kvark:
@emilio ideally, the first commit should be left with me as an author. If you rebased on the PR, you'd have it right.

emilio:
So I was seeing the same failures on our automation, and I debugged again and I understand the issue now. Will file an issue, but there's no point in keeping this open for now.

The TL;DR: is that nothing the output of `BorderRenderTaskInfo::new` can depend on anything that isn't in our `BorderCacheKey`, as of right now. Will mention why in a second.

PR #3059, opened 2018-09-14
Tidy up the scene flattening interface.

gw3583:
Now that display list flattening is always done by the scene
building thread, there is never a valid old frame builder
that can be recycled. Update the interface to make this
clearer.

Also slightly tidy up how picture ids are generated. This is
a temporary hack only - they won't be needed once we are
caching pictures based on their content.

This also removes the last current use or recycle_vec and
associated methods. I've removed them for now to avoid
warnings, but hopefully in the future we can find a way
to use these again in some places to avoid so many allocations
per frame!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3059)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -19,7 +19,7 @@ use render_task::to_cache_size;
 use resource_cache::{ImageRequest, ResourceCache};
 use std::{cmp, u32};
 use std::os::raw::c_void;
-use util::{extract_inner_rect_safe, pack_as_float, project_rect, recycle_vec, ScaleOffset};
+use util::{extract_inner_rect_safe, pack_as_float, project_rect, ScaleOffset};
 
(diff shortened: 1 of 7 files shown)


gw3583:
r? @nical 

Just some bits and pieces I came across while prototyping picture caching changes.

gw3583:
Pending try (shouldn't be any changes from this patch):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=836046b753e16a014e113260e525694413248eb4

gw3583:
Rebased and kicked off another try run after the hit-test PR was reverted.

Pending:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=77a8d1c255c4055f23fc50b6c5fd16c6c63a1edf

PR #3060, opened 2018-09-14
Only use serde when feature="serialize | deserialize" are active

fschutt:
Since #[derive(Serialize, Deserialize)] are already behind these feature flags, there isn't much to do, it's simply unnecessary to require serde when neither serialize or deserialize features are active.

Together with https://github.com/servo/app_units/pull/44, this will eliminate serde from webrender entirely in `--no-default-features` builds.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3060)


fschutt:
Sorry, made the PR way too early. Closing for now, there is stuff in lib.rs that isn't going to pass the tests.

PR #3062, opened 2018-09-14
Move app_units::Au into api::units, re-export Au

### LINKED ISSUE #3045: Re-export app_units::Au or disable serde for app_units
In order to create a `AddFontInstance` struct, you have to use an `app_units::Au`. This is very annoying because:

- `app_units` depends on serde and num_traits, both very compile-time heavy crates that are completely unused, yet they still need to be compiled in
- This is the only place in the entire **public** webrender API where `Au` is used
- `app_units` isn't re-exported, so other crates have to add the exact version that webrender uses to their Cargo.toml manually.

So my proposal is to create a function that creates an `Au` from an i32 value, this way you don't have to re-export the `Au`, i.e.:

```rust
pub fn webrender_create_au_from_px(value: i32) -> Option<Au> {
    let value = value * AU_PER_PX;
    if value < MIN_AU || value > MAX_AU {
        None
    } else {
        Some(Au(value))
    }
}
```

And please make `serde` and `num-traits` in the `app-units` crate optional - for a regular webrender application, these crates go completely unused because serde serialization is only used for the debugger, but these crates still have to be compiled, leading to bad compile times.

fschutt:
Fixes #3045

Now we can update app_units in one place (the webrender_api crate) instead of having to track the version of app_units in every tool. Also any dependencies that build upon webrender now don't have to worry about tracking what version of app_units webrender is using.

Also created a new `create_app_units_from_px` which does the required validation from pixel values to Au.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3062)


fschutt:
It seems to fail because of a "license check issue" on Linux - I don't understand that, I didn't change anything about licenses. Maybe a style lint issue?

gw3583:
It's not clear to me what that means either - you might be able to get more info by running that step locally? (the taskcluster / appveyor config files should have the command info).

fschutt:
The reason I couldn't run in offline was because it runs `servo-tidy` and I don't know where to find that `servo-tidy` script.

kvark:
This one? https://github.com/servo/servo/tree/master/python/tidy/servo_tidy

kvark:
@fschutt maybe this is what it's really complaining about:
> ./webrender_api/src/units.rs:166: no newline at EOF

fschutt:
Alright, seems to have fixed itself. Rebased on master to resolve the merge conflicts.

fschutt:
I think this can be closed, since `Au` is now exported under `units::Au`.

PR #3064, opened 2018-09-15
Prep work for retaining clip nodes in GPU cache between display lists.

gw3583:
This will also allow us to quickly check clip nodes for equality (part of picture caching work) in the future.

Commit notes:

```
Remove clip item range.

Instead, there is a single clip node index stored in each clip
chain node. This makes the clip chain nodes slightly longer in
some use cases (most clip nodes only contain a single clip).
```

```
Store the spatial node index in the clip chain node.

This means the ClipNode is independent of positioning, which
makes it simple to retain in the GPU cache between frames and
display lists.
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3064)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1777,10 +1777,10 @@ impl ClipBatcher {
         transforms: &mut TransformPalette,
     ) {
         for i in 0 .. clip_node_range.count {
-            let (clip_node, flags) = clip_store.get_node_from_range(&clip_node_range, i);
+            let (clip_node, flags, spatial_node_index) = clip_store.get_node_from_range(&clip_node_range, i);
 
             let clip_transform_id = transforms.get_id(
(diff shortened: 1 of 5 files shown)


gw3583:
r? @kvark or anyone

These changes shouldn't have any visible effect - they are just re-arranging some data structures to make it easier to deal with clips for picture caching and retaining nodes in the gpu cache.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7c37fbb548199ef231418e388f2e9b4f378ec934

gw3583:
Try run looks better now:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2acd6b91951445970e34ff985fc1b9084e921ce9

gw3583:
Wow, that comment above mentions ```clip``` 11 times. :fearful:

PR #3065, opened 2018-09-17
Replace BorderWidths with SideOffsets2D

### LINKED ISSUE #3053: Remove BorderWidths in favor of SideOffsets2D
For both borders, margins (and others) information (usually lengths) needs to be stored for the four sides of a rect. This is done with the SideOffsets2D type in servo-layout. But in Webrender the BorderWidths struct is used for borders instead. This is a bit annoying as the field order is different left/top/right/bottom vs top/right/bottom/left, BorderWidths lacks convenience methods to do calculations or test stuff (border_widths.is_zero()?). Additionally servo-layout needs to convert SideOffsets2D<Au> into two different types with f32 as the value.

I understand that offset does not immediately convey width but I think the other benefits outweigh the more general name.

pyfisch:
closes #3053

While I am at it, should I also replace BorderWidthsAu and remove the border outset?

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -228,12 +228,7 @@ impl Example for App {
             color: ColorF::new(0.0, 0.0, 1.0, 1.0),
             style: BorderStyle::Groove,
         };
-        let border_widths = BorderWidths {
-            top: 10.0,
-            left: 10.0,
-            bottom: 10.0,
-            right: 10.0,
-        };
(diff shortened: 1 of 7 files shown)


nical:
Looks great! @bors-servo r+

> While I am at it, should I also replace BorderWidthsAu and remove the border outset?

That would be nice.

nical:
@jrmuizel heads up there will be a few things to change in the API for next WR update although it should straightforward since BorderWidths and SideOffsets have the same members.

PR #3066, opened 2018-09-17
Add debug commands to simulate slow scene building.

nical:
I found a race condition that was only reliably reproducible by artificially slowing down scene building. Adding this to the debug commands to make it easier in the future. It would be useful to follow up with debug options to simulate slow frame builds and slow rendering.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3066)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -817,6 +817,16 @@ impl RenderBackend {
                         self.resource_cache.clear(mask);
                         return true;
                     }
+                    DebugCommand::SimulateLongSceneBuild(time_ms) => {
+                        self.scene_tx.send(SceneBuilderRequest::SimulateLongSceneBuild(time_ms)).unwrap();
+                        return true;
+                    }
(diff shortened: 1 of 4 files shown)


webrender/src/render_backend.rs
```diff
@@ -817,6 +817,16 @@ impl RenderBackend {
                         self.resource_cache.clear(mask);
                         return true;
                     }
+                    DebugCommand::SimulateLongSceneBuild(time_ms) => {
+                        self.scene_tx.send(SceneBuilderRequest::SimulateLongSceneBuild(time_ms)).unwrap();
+                        return true;
+                    }
+                    DebugCommand::SimulateLongLowPrioritySceneBuild(time_ms) => {
```
kvark: should just be a bool parameter of `SimulateLongSceneBuild`

nical: I almost did that but eventually we'll have SimulateLongFrameBuilder and SimulateLongRender as well and then having the SB ones separate instead of a bool feels more natural (totally splitting hairs here).

PR #3067, opened 2018-09-17
border: Don't always share instances across borders if the instances depend on the available border size.

### LINKED ISSUE #3058: Need to be less aggressive at border caching.
This is what is preventing us from landing #3049 or #3056. This is of course already broken on trunk, see the following page for sadness: https://crisal.io/tmp/wr-border-caching.html

[Here's a YAML thingie which also shows the issue](https://github.com/servo/webrender/files/2380794/out.txt), though less exacerbated because there's less garbage in the GPU memory I guess.

The upshot is that as of right now nothing based on something like the available size should change the output of segments, which means that the optimization we were trying to land in those PRs breaks so hard.

Now, doing that for solid borders (and anything non-dashed-or-dotted, really) might be feasible (we'd still need to decide if we want to pay the price of generating those zero-size instances, but I suspect it might be possible). However for those it's not.

I had a very hacky patch which fixed all those issues, which looks something like that (note that I haven't verified that the first change is needed):

```diff
diff --git a/gfx/webrender/src/prim_store.rs b/gfx/webrender/src/prim_store.rs
index 5b65764173a3..3e07b343f16a 100644
--- a/gfx/webrender/src/prim_store.rs
+++ b/gfx/webrender/src/prim_store.rs
@@ -2845,17 +2845,17 @@ impl Primitive {
             // TODO(gw): When drawing in screen raster mode, we should also incorporate a
             //           scale factor from the world transform to get an appropriately
             //           sized border task.
             let world_scale = LayoutToWorldScale::new(1.0);
             let mut scale = world_scale * frame_context.device_pixel_scale;
             let max_scale = BorderRenderTaskInfo::get_max_scale(&border.radius, &widths);
             scale.0 = scale.0.min(max_scale.0);
             let scale_au = Au::from_f32_px(scale.0);
-            let needs_update = scale_au != cache_key.scale;
+            let needs_update = scale_au != cache_key.scale || true;
             let mut new_segments = Vec::new();
 
             if needs_update {
                 cache_key.scale = scale_au;
 
                 *task_info = BorderRenderTaskInfo::new(
                     &self.metadata.local_rect,
                     border,
@@ -2863,17 +2863,17 @@ impl Primitive {
                     scale,
                     &mut new_segments,
                 );
             }
 
             *handle = task_info.as_ref().map(|task_info| {
                 frame_state.resource_cache.request_render_task(
                     RenderTaskCacheKey {
-                        size: DeviceIntSize::zero(),
+                        size: task_info.size,
                         kind: RenderTaskCacheKeyKind::Border(cache_key.clone()),
                     },
                     frame_state.gpu_cache,
                     frame_state.render_tasks,
                     None,
                     false,          // todo
                     |render_tasks| {
                         let task = RenderTask::new_border(
```

In any case, we need to figure out how we want to make this caching work, and if we want to do something different for solid borders than for dashed borders (I guess so).

/cc @glennw

emilio:
I've added a big comment to BorderRenderTaskInfo with my current understanding
of the invariants this code is trying to preserve.

Please sanity-check it, but I think this patch should both be sound and also
prevent the scrolling-franzine issue
(https://bugzilla.mozilla.org/show_bug.cgi?id=1489644).

Fixes #3058

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3067)


webrender/src/border.rs
```diff
@@ -528,10 +528,63 @@ pub struct BorderSegmentInfo {
     widths: DeviceSize,
 }
 
+bitflags! {
+    /// Whether we depend in the available size for the border (effectively in
```
kvark: did you mean "depend on" instead of "depend in"?

webrender/src/border.rs
```diff
@@ -894,9 +965,33 @@ impl BorderRenderTaskInfo {
...
+        local_size: &LayoutSize,
+        scale: LayoutToDeviceScale,
+    ) -> DeviceIntSize {
+        let mut size = DeviceIntSize::zero();
+        if self.available_size_dependence.is_empty() {
+            return size;
+        }
+
+        let device_size = (*local_size * scale).to_i32();
+        if self.available_size_dependence.intersects(AvailableSizeDependence::VERTICAL) {
```
kvark: nit: use `contains`?

emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=f7214c7ebbc3611712383bf1bf982e5304f1019b

webrender/src/border.rs
```diff
@@ -894,9 +965,33 @@ impl BorderRenderTaskInfo {
...
+    /// dependence computed earlier.
+    #[inline]
+    pub fn cache_key_size(
+        &self,
+        local_size: &LayoutSize,
+        scale: LayoutToDeviceScale,
+    ) -> DeviceIntSize {
+        let mut size = DeviceIntSize::zero();
+        if self.available_size_dependence.is_empty() {
+            return size;
```
kvark: why aren't we returning the `*local_size * scale` here?

webrender/src/border.rs
emilio: Yup, will fix.

webrender/src/border.rs
emilio: Will do.

webrender/src/border.rs
emilio: Because for borders where we don't depend on any size we just want the same value for all of them, so they hit the cache and we don't build instances unnecessarily.

emilio:
Comments addressed.

emilio:
Reftests and such look all green so far, yay :)

webrender/src/border.rs
kvark: what about a particular width/height then? shouldn't we put 0.0 into width if the key is only height-sensitive?

emilio: That's what we do.

emilio: Note that we return `size` instead of `device_size`. I'm open to suggestions about how to make that clear.

emilio:
@bors-servo r=kvark

@gw3583 feel free to take a quick skim at the changes just to sanity-check, if you want :)

PR #3069, opened 2018-09-17
GPU cache debug visualization

kvark:
This is a barebones visualization path for GPU cache updates. The idea is to assign different tags/colors for different types of the GPU cache blocks, so that we can see how bad the updates are fragmented upon scrolling and/or scene updates.

wr-gpu-cache-debug

r? @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3069)


gw3583:
@kvark Just the one question, then r=me.

kvark:
@gw3583 thanks for the quick look! Should be good now ;)

PR #3072, opened 2018-09-18
Document webrender_api::image

bholley:
Added docs to most things, stubbed out a few, and added deny(missing_docs).

r? @gw3583 or @kvark

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3072)

### DIFF
--- a/webrender_api/src/image.rs
+++ b/webrender_api/src/image.rs
@@ -253,8 +253,6 @@ pub struct RasterizedBlobImage {
 #[derive(Clone, Debug)]
 pub enum BlobImageError {
     Oom,
-    InvalidKey,
-    InvalidData,
     Other(String),
 }
 
--- a/webrender_api/src/image.rs
+++ b/webrender_api/src/image.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#![deny(missing_docs)]
+
 extern crate serde_bytes;
 
 use font::{FontInstanceKey, FontInstanceData, FontKey, FontTemplate};
@@ -10,13 +12,18 @@ use {DevicePoint, DeviceUintPoint, DeviceUintRect, DeviceUintSize};
 use {IdNamespace, TileOffset, TileSize};
 use euclid::size2;
 
+/// An opaque identifier describing an image registered with WebRender.
+/// This is used as a handle to reference images, and is used as the
+/// hash map key for the actual image storage in the `ResourceCache`.
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub struct ImageKey(pub IdNamespace, pub u32);
 
 impl ImageKey {
+    /// Placeholder Image key, used to represent None.
     pub const DUMMY: Self = ImageKey(IdNamespace(0), 0);
 
+    /// Mints a new ImageKey. The given ID must be unique.
     pub fn new(namespace: IdNamespace, key: u32) -> Self {
         ImageKey(namespace, key)
     }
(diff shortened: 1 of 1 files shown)


webrender_api/src/image.rs
```diff
@@ -74,17 +114,35 @@ impl ImageFormat {
...
+    /// Format of the image data.
     pub format: ImageFormat,
+    /// Width and length of the image data, in pixels.
     pub size: DeviceUintSize,
+    /// The number of bytes from the start of one row to the next. Lazily
+    /// computed via compute_stride() and cached here when the image is
+    /// tile.
+    ///
+    /// FIXME(bholley): The computation of this looks pretty cheap. Do we have
+    /// evidence that this is hot enough that it needs caching?
```
emilio: This doesn't look like a cache, it's an override for when the stride is not width times bpp.

webrender_api/src/image.rs
```diff
@@ -222,44 +298,64 @@ pub trait BlobImageHandler: Send {
 
 /// A group of rasterization requests to execute synchronously on the scene builder thread.
 pub trait AsyncBlobImageRasterizer : Send {
+    /// Rasterize the requests.
     fn rasterize(&mut self, requests: &[BlobImageParams]) -> Vec<(BlobImageRequest, BlobImageResult)>;
 }
 
 
+/// FIXME: document
```
emilio: Maybe just allow() here to avoid having the comment below?

webrender_api/src/image.rs
nical: Yep, what Emilio said. Different source of images have different alignment constraints for rows, so the stride isn't always equal to width * bpp.

webrender_api/src/image.rs
```diff
@@ -222,44 +298,64 @@ pub trait BlobImageHandler: Send {
...
 #[derive(Copy, Clone, Debug)]
 pub struct BlobImageParams {
+    /// FIXME: document
     pub request: BlobImageRequest,
+    /// FIXME: document
     pub descriptor: BlobImageDescriptor,
+    /// FIXME: document
```
nical: /// An optional sub-rectangle of the image to avoid re-rasterizing the entire image when only a portion is updated.
///
/// If set to `None` the entire image is rasterized.

webrender_api/src/image.rs
```diff
@@ -222,44 +298,64 @@ pub trait BlobImageHandler: Send {
...
 pub type BlobImageData = Vec<u8>;
 
+/// Result type for blob raserization.
 pub type BlobImageResult = Result<RasterizedBlobImage, BlobImageError>;
 
+/// Metadata (but not storage) for a blob image.
 #[repr(C)]
 #[derive(Copy, Clone, Debug)]
 pub struct BlobImageDescriptor {
+    /// Bounding rectangle of the image, in device pixels.
```
nical: I'd rephrase this as:

/// Size in device pixels of the blob's output image.

webrender_api/src/image.rs
```diff
@@ -222,44 +298,64 @@ pub trait BlobImageHandler: Send {
...
     pub size: DeviceUintSize,
+    /// FIXME - Document me.
     pub offset: DevicePoint,
```
nical: Huh! it's unused. We should probably remove it.

webrender_api/src/image.rs
```diff
@@ -222,44 +298,64 @@ pub trait BlobImageHandler: Send {
...
 pub enum BlobImageError {
+    /// Out of memory.
     Oom,
-    InvalidKey,
-    InvalidData,
+    /// Other failure, embedding-specified.
     Other(String),
 }
 
+/// FIXME: document
```
nical: A key identifying blob image rasterization work requested to the blob image rasterizer.

webrender_api/src/image.rs
```diff
@@ -222,44 +298,64 @@ pub trait BlobImageHandler: Send {
...
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct BlobImageRequest {
+    /// Unique handle to the image.
     pub key: ImageKey,
+    /// Tiling offset.
+    /// FIXME - document tiling better.
```
nical: /// Tile offset in number of tiles, if applicable.
///
/// If set to `None` it means the image is not tiled.

webrender_api/src/image.rs
```diff
@@ -74,17 +114,35 @@ impl ImageFormat {
...
+    /// Width and length of the image data, in pixels.
     pub size: DeviceUintSize,
+    /// The number of bytes from the start of one row to the next. Lazily
+    /// computed via compute_stride() and cached here when the image is
+    /// tile.
+    ///
+    /// FIXME(bholley): The computation of this looks pretty cheap. Do we have
+    /// evidence that this is hot enough that it needs caching?
     pub stride: Option<u32>,
+    /// Byte offset used for tiling. FIXME: Better documentation.
```
nical: /// Offset in bytes of the first pixel in the buffer that holds the image.

webrender_api/src/image.rs
bholley: It's used by wrench.

PR #3075, opened 2018-09-18
Introduce structure interner, and use for clip interning.

gw3583:
This patch introduces a new data structure, that allows generic
structures to be interned. This works similarly to how normal
string interning works, however it is specialized to the
thread model for WR (explained at the top of intern.rs).

The effect of this change is that clip nodes are de-duplicated
and persisted between both frames and display lists.

This has three primary benefits:
 * Since they are de-duplicated, the handle for an interned
   structure uniquely identifies it. This is very useful for
   future use where we want to be able to quickly and cheaply
   compare if the contents of a cached picture matches that
   of a new display list.
 * Since they are persisted between display lists, the GPU
   cache handles for the nodes remain valid. This means far
   fewer GPU cache update patches for types that are interned.
 * Since they are de-duplicated by content hash value, there
   are fewer clips overall used by the frame builder.

The plan in the future is to extend this to other primitive types,
as well as gradient stops, text runs etc. This will allow us to
very quickly check if a cached picture remains valid, even in the
presence of a completely new display list.

This adds a small amount of overhead to the scene builder thread,
(extra hashing) but reduces the CPU time in the render backend and
compositor threads, which is also a good tradeoff.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3075)

### DIFF
--- a/webrender/src/intern.rs
+++ b/webrender/src/intern.rs
@@ -7,6 +7,7 @@ use std::fmt::Debug;
 use std::hash::Hash;
 use std::marker::PhantomData;
 use std::mem;
+use std::ops;
 use std::u64;
 
 /*
@@ -39,6 +40,11 @@ use std::u64;
 
  */
 
+/// The epoch is incremented each time a scene is
+/// built. The most recently used scene epoch is
+/// stored inside each item and handle. This is
+/// then used for cache invalidation (item) and
+/// correctness validation (handle).
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
 #[derive(Debug, Copy, Clone, PartialEq)]
@@ -48,12 +54,12 @@ impl Epoch {
     pub const INVALID: Self = Epoch(u64::MAX);
 }
 
-// A list of updates to be applied to the data store,
-// provided by the interning structure.
+/// A list of updates to be applied to the data store,
+/// provided by the interning structure.
 pub struct UpdateList<S> {
-    // The current epoch of the scene builder.
+    /// The current epoch of the scene builder.
     epoch: Epoch,
-    // The additions and removals to apply.
+    /// The additions and removals to apply.
     updates: Vec<Update<S>>,
 }
 
@@ -81,8 +87,8 @@ pub struct Update<S> {
     kind: UpdateKind<S>,
 }
 
-// The data item is stored with an epoch, for validating
-// correct access patterns.
+/// The data item is stored with an epoch, for validating
+/// correct access patterns.
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
 struct Item<T> {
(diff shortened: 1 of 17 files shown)


gw3583:
r? @kvark

This does seem to give a noticeable reduction in the number of GPU cache rows on several pages that I tested on. Of course, the major wins wouldn't come until we use it for more than just clips.

The main remaining bit of work (apart from review comments) is that I haven't hooked up the interner and data store to the capture / replay code yet.

Try run looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=32256ade54c961eabec197c380bb2eeda194f8d1&selectedJob=199879097

(lots of blue and orange, but they are unrelated as far as I can tell - still some jobs pending though)

gw3583:
And also r? @nical too, since this touches scene builder thread and such things.

webrender/src/render_backend.rs
```diff
@@ -267,8 +275,12 @@ impl Document {
                 &mut resource_profile.texture_cache,
                 &mut resource_profile.gpu_cache,
                 &self.dynamic_properties,
+                &mut self.clip_data_store,
             );
-            self.hit_tester = Some(frame_builder.create_hit_tester(&self.clip_scroll_tree));
+            self.hit_tester = Some(frame_builder.create_hit_tester(
```
nical: My only worry is creating a dependency between frame building and building the hit tester. We really need to be able to build the latter independently. Here we are first mutably passing the clip data store to the frame builder and then using it to build the hit tester.
Upon inspecting [the code that mutates the data store](https://github.com/servo/webrender/pull/3075/files#diff-a1e61c23c68078e351d27b526664bf56R449), it looks like only GPU thingies are modified so it should be safe to create the hit tester independently, but I'd feel better if we moved the line that creates the hit tester above the `FrameBuilder::build` call to make sure dependencies don't creep in.

webrender/src/intern.rs
```diff
@@ -0,0 +1,288 @@
...
+// scene builder thread, and handles hashing and interning
+// unique data structures. It also manages a free-list for
+// the items in the data store, which is synchronized via
+// an update list of additions / removals.
+#[cfg_attr(feature = "capture", derive(Serialize))]
+#[cfg_attr(feature = "replay", derive(Deserialize))]
+pub struct Interner<S : Eq + Hash + Clone + Debug, M> {
+    // Uniquely map an interning key to a handle
+    map: FastHashMap<S, Handle<M>>,
+    // List of free slots in the data store for re-used.
```
nical: nit: "for re-use"?

webrender/src/render_backend.rs
gw3583: Yep, sounds good.

gw3583:
@kvark Thanks for the review! Still need to implement the capture functionality, and any follow ups from the review that remain above.

gw3583:
@kvark Added one more commit with the capture / replay integration. I think it's possible to have races where the interner / data store get out of sync, but maybe it's really unlikely in practice. What do you think? (I'm going to do some testing locally to see if I can ever get them out of sync).

gw3583:
@kvark I did some testing with the simple capture / replay integration and it does seem to work. I couldn't reproduce any out of sync issues, so it may be fine as is.

I also rebased it, so I think this is ready to go once you check over the last commit.

Kicked off a try run (pending) too, to make sure I didn't break anything with the review comment fixes:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=f8027b1a08ffbc4c0e8fc4dbed58b79a9ec2f59c

gw3583:
(rebased again)

gw3583:
Try run still looks good :tada:

gw3583:
Rebased again - just removed the modified box shadow reftest image, which is no longer different with this change after @pcwalton 's box shadow border radius fix.

PR #3079, opened 2018-09-18
Store a queue of rasterized blobs instead for non-tiled images.

nical:
This fixes a race-condition that happens if we don't upload rasterized blobs right after scene building which blocks the work on avoiding redundant frame builds.

The idea is that instead of keeping a single rasterized blob per image, we store a queue which lets us have several partial updates of a blob in flight (previously the second update would clobber the first and miss the updated area of the first update).
To prevent the queue from growing indefinitely, we make it so that older elements are discarded if a newer one covers them, and re-rasterize everything if the queue gets big to put it back to a single update.

I expect that most of the time for visible items the queue will stay at 1 (or rarely 2) updates even if the blob is updating constantly because we tend to render at least as often as we build scenes.
For items that are off-screen and thus not getting uploaded, the queue will alternate between 1, 2 and 3 elements, and we'll force the rasterization of the whole image each time the queue gets to 3 to force the queue back to 1 element.

The case of animated off-screen blobs is pretty bad because they generate rasterization constantly and don't end up on screen, but that's already the case today.
We could (as a followup) try to just discard the elements of the queue when it gets big (3 elements or more) and skip rasterization of the blob image until we see it.
This means the next time we see it we'll risk jank since it will trigger the synchronous rasterization during frame building, but at least we'd have avoided rasterizing the blobs during the time it was animating off-screen.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3079)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -32,6 +32,7 @@ use profiler::{ResourceProfileCounters, TextureCacheProfileCounters};
 use render_backend::FrameId;
 use render_task::{RenderTaskCache, RenderTaskCacheKey, RenderTaskId};
 use render_task::{RenderTaskCacheEntry, RenderTaskCacheEntryHandle, RenderTaskTree};
+use smallvec::SmallVec;
 use std::collections::hash_map::Entry::{self, Occupied, Vacant};
 use std::collections::hash_map::ValuesMut;
 use std::{cmp, mem};
@@ -100,7 +101,7 @@ enum State {
 /// Post scene building state.
 enum RasterizedBlob {
     Tiled(FastHashMap<TileOffset, RasterizedBlobImage>),
-    NonTiled(RasterizedBlobImage),
+    NonTiled(Vec<RasterizedBlobImage>),
 }
 
 /// Pre scene building state.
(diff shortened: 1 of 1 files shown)


nical:
This PR might be easier to understand by reading the commits individually. It looks kinda complicated and I'm sorry about that. The alternative was in my opinion also complicated and more invasive.
I'm still testing it out in conjunction with the frame building stuff.

nical:
This try push has the patches from this PR plus the backed out ones that prevent redundant frame building
https://treeherder.mozilla.org/#/jobs?repo=try&revision=8f436bfb2b982214738e512042a460c73bc96432
It looks good so far.

nical:
Yep I haven't addressed the nits yet, coming soon.
I don't think of SmallVec as hack. I wish rust was better at expressing them but it's probably my favorite data structure.
Persisting a vector however (when in the mast majority of cases the size will be 1) feels like a more convoluted solution to the problem.

kvark:
My general concern about `SmallVec` is that it makes it more difficult to reason about performance. If you increase the scene complexity gradually, you don't want the frame rate to jump rapidly. In case of `SmallVec`, there is a threshold upon crossing which you end up paying for the heap allocation on every use. With a persistent temporary vector, there is no such problem.

This particular case may not be that important, I'm just expressing the general idea.

kvark:
shipit
@bors-servo r+

PR #3080, opened 2018-09-18
Call `ensure_no_corner_overlap()` to legalize border radii in `ClipItem::new_box_shadow()`.

### LINKED ISSUE #2581: Box-shadow has wrong corner radius in this testcase
It looks correct without a blur, but as soon as you add 1px of blur, things start to go wrong.

http://tests.themasta.com/box-shadow-playground/#blue_red_outset_0_1_265_265_250x250_0x0_0x0_250x250_350_0_1_-80

This is also broken in non-webrender Firefox, but in different and exciting ways.

pcwalton:
If no blur is present, `DisplayListFlattener::add_box_shadow()` calls
`ClipItem::new_rounded_rect()` in its fast path, which calls
`ensure_no_corner_overlap()`. However, in the slow path, which is used when blur is present, `ClipItem::new_box_shadow()` is called instead. Prior to this patch, `ClipItem::new_box_shadow()` neglected to call
`ensure_no_corner_overlap()`. This is why the bug manifested itself only when blur was present.

Closes #2581.

r? @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3080)


pcwalton:
This caused `no-stretch.yaml`, or #2467, to fail.

pcwalton:
It looks like this actually improves the rendering of `no-stretch.yaml` a bit, by comparing with Google Chrome's rendering. @gw3583, do you agree?

pcwalton:
Here's `no-stretch.yaml`.

Old (before this patch):

no-stretch-old

New (with this patch):

no-stretch-new

Google Chrome:

screen shot 2018-09-18 at 12 32 22 pm

gw3583:
Pending try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=aed9b249289ca99e102536b7ff84131e171ec53b

gw3583:
Try looks good. :rocket: 

@bors-servo r+

PR #3082, opened 2018-09-18
 Snap clips properly accounting for the prim transform. 

emilio:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1491555.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3082)


emilio:
Need to still update some reftests, but this is my proposed solution to that bug.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=ddb918141fbf2c766fc4e668aabd1c32d17522ff

gw3583:
This change looks good to me, thanks! I think the CI failures are that the new reftest images need to be cropped to previous size.

emilio:
The reftest was a bit fuzzy on windows, so given we don't have a good fuzzy-if() mechanism I just skipped it there...

@bors-servo r=gw3583

emilio:
@bors-servo r+

 * trying to get bors unstuck

emilio:
err,

@bors-servo r-

PR #3086, opened 2018-09-19
Bring back texelFetch from clip surface

kvark:
This PR reverts #3073, leaves the snapping offsets around, and fixes the concern spotted in https://github.com/servo/webrender/pull/3073#issuecomment-422891829

Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2fce2addd97b66f044ecc83a46c7f45b53dc5381&selectedJob=200247696

WIP because: needs @gw3583 to check if it fixes their test on Nvidia GPU.
cc @jrmuizel

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3086)


gw3583:
This still seems to result in garbage / corruption on the machine I'm testing on (Ubuntu 18.04, XPS 9550 with nVidia proprietary driver enabled).

kvark:
Thanks @gw3583 !
It does however fix a few cases in the try run ("unexpected" passes), so we are looking at 2 different issues, one of which is covered by this PR.

gw3583:
@kvark It looks like there is one new PASS, which is the skiprect test - but that was also fixed by #3073. Or are there other tests I missed?

kvark:
Yes, that's the one I meant. We had two issues with this code, and #3073 fixed both of them, while this PR only addresses one so far.
Please consider this blocked until I have a chance to look at those NVidia results.

gw3583:
Ah, I understand - sounds good, thanks!

kvark:
I looked at the bug some more, could reproduce somewhat similar behavior on native GL implementation of an AMD card. However, Wrench+Angle stubbornly refused to replay the D3D11 captures of the bug: everything is loaded and sent to the API for rendering, but Angle doesn't produce any actual draw calls, just the state switches :imp: 

My guess so far is that the bounds checks fail because of near-infinite values involved. We might be able to solve it, I just need to be able to replay it on Angle and play with it in RenderDoc...

kvark:
Problematic page for my own reference: https://bugzilla.mozilla.org/attachment.cgi?id=9009686

kvark:
@gw3583 I figured out what the *second* problem was. It's not a driver bug, it's just that the clip layer index we use in the pixel shader is interpolated, and even though all vertices output it as 1.0, it's still accumulating the interpolation precision issues (since it's a non-flat varying), which are especially visible at high perspective.

Your workaround was effective because a non-fetch version receives a float for the layer index, and it rounds up the float for us. I changed this PR to do the same for the fetch, which I confirmed on an NV machine to fix the issue :tada: Please take a look!

webrender/res/prim_shared.glsl
```diff
@@ -251,21 +243,18 @@ float do_clip() {
...
-
     // finally, the slow path - fetch the mask value from an image
-
-    // TODO(gw): texelFetch here fails on some nVidia hardware in
-    //           some cases. For now, just use texture()
-    //           unconditionally.
-    mask_uv = clamp(mask_uv, vClipMaskUvSampleBounds.xy, vClipMaskUvSampleBounds.zw);
-    return texture(sCacheA8, vec3(mask_uv, vClipMaskUv.z)).r;
+    // Note the Z getting rounded to the nearest integer because the variable
+    // is still interpolated and becomes a subject of precision-cauesed
```
leeoniya: "precision-caused"

nical:
@kvark should I remove the entry from the driver issues wiki page or is it enough of a gotcha that it should stay there for later reference?

kvark:
@nical too late, my friend :)

gw3583:
@kvark Nice fix! Do we need a try on this or is it good to me? r=me when you're ready to merge it.

kvark:
@gw3583 it's only difference from the previous try push is rounding of the layer coordinate, with should be entirely safe.
@bors-servo r=gw3583

PR #3090, opened 2018-09-20
Don't move pending_properties to current_properties in flush_pending_…

hiikezoe:
…updates

Basically add_properties is called along with set_properties, but sometimes
add_properties is called without set_properties, once it happens in a single
transaction, the DynamicProperties which was previously set by set_properties
will not be incorporated into the dynamic properties because it was cleared
in the previous transaction, thus a flicker will happen.  E.g. a transform
animation value set by set_properties isn't rendered.

To avoid the flicker this change stops moving pending_properties in
flush_pending_updates and thus only set_properties clears the
penrind_properties.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3090)

### DIFF
--- a/webrender/src/scene.rs
+++ b/webrender/src/scene.rs
@@ -58,8 +58,8 @@ impl SceneProperties {
     pub fn flush_pending_updates(&mut self) -> bool {
         let mut properties_changed = false;
 
-        if let Some(pending_properties) = self.pending_properties.take() {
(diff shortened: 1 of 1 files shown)


hiikezoe:
This PR fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1481313

nical:
Nice find! @bors-servo r+

PR #3092, opened 2018-09-20
Avoid redundant frame builds (take two).

nical:
#3052 rebased with fixes from #3079.
The [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=8f436bfb2b982214738e512042a460c73bc96432) of the version before rebase looked good but I'll make another one to be on the safe side.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3092)

### DIFF
--- a/webrender/src/clip_scroll_tree.rs
+++ b/webrender/src/clip_scroll_tree.rs
@@ -253,10 +253,14 @@ impl ClipScrollTree {
         &mut self,
         pan: WorldPoint,
         scene_properties: &SceneProperties,
-    ) -> TransformPalette {
-        let mut transform_palette = TransformPalette::new(self.spatial_nodes.len());
+        mut transform_palette: Option<&mut TransformPalette>,
+    ) {
         if self.spatial_nodes.is_empty() {
-            return transform_palette;
+            return;
+        }
+
+        if let Some(ref mut palette) = transform_palette {
+            palette.allocate(self.spatial_nodes.len());
         }
 
         self.coord_systems.clear();
@@ -282,7 +286,9 @@ impl ClipScrollTree {
             };
 
             node.update(&mut state, &mut self.coord_systems, scene_properties);
-            node.push_gpu_data(&mut transform_palette, node_index);
+            if let Some(ref mut palette) = transform_palette {
+                node.push_gpu_data(palette, node_index);
+            }
 
             if !node.children.is_empty() {
                 node.prepare_state_for_children(&mut state);
@@ -293,8 +299,6 @@ impl ClipScrollTree {
                 );
             }
         }
-
-        transform_palette
     }
 
     pub fn finalize_and_apply_pending_scroll_offsets(&mut self, old_states: ScrollStates) {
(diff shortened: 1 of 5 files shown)


zptan:
`error: unused std::result::Result which must be used
   --> webrender/src/scene_builder.rs:233:21
    |
233 |                     self.api_tx.send(ApiMsg::WakeUp);
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D unused-must-use` implied by `-D warnings`
    = note: this `Result` may be an `Err` variant, which should be handled
error: Could not compile webrender.
`

nical:
Note that there's an extra commit which wasn't in the original PR (so it should be reviewed as new stuff).

It makes sure APZ is sampled when we skip rendering (it shouldn't matter to users but some tests that scroll programmatically appear to be sensible to that), and makes sure the ExternalEvent messages don't skip ahead of transactions.

nical:
I'll do the try push right after the next wr update.

nical:
___
*[webrender/src/frame_builder.rs, line 346 at r2](https://reviewable.io/reviews/servo/webrender/3092#-LMwaKwf6XIFG289XY6h:-LMwoVexE8S5vIk8Jc_s:b-kij39l) ([raw file](https://github.com/servo/webrender/blob/1f3661952ff4b96a0f6e482fe54ebc008cd6690d/webrender/src/frame_builder.rs#L346)):*

it's also a bad thing to create an empty object only for it to be completely overwritten by the function you provide it with `&mut`. Semantically, the function should return the palette, but only if we ask it to do so. So the `Some()` is basically a bool here, and `allocate` is basically a *real* constructor. I find this obscure, hence the proposal to at least initialize it properly where it's constructed.
</blockquote></details>

I can hardly disagree more. Creating an empty object and asking a function to fill it is a super common programming pattern and it is totally fine.
If you flip it around by passing a boolean parameter to decide whether the function should build the palette, the only way to be sure of what the function actually does without reading it. By having the palette an optional out parameter you make it structurally impossible for the function to do the wrong thing. Kats' initial implementation was returning an `Option<TransformPalette>` and when building the hit tester only he ended up adding an assertion that the returned option was None " 'cause you never know".

nical:
___
*[webrender/src/frame_builder.rs, line 346 at r2](https://reviewable.io/reviews/servo/webrender/3092#-LMwaKwf6XIFG289XY6h:-LMwxQs_1-dGv43J4C22:b-jkokow) ([raw file](https://github.com/servo/webrender/blob/1f3661952ff4b96a0f6e482fe54ebc008cd6690d/webrender/src/frame_builder.rs#L346)):*

> Creating an empty object and asking a function to fill it is a super common programming pattern and it is totally fine.

A common pattern is to mutably work with something, yes, but that's not exactly what you are doing. `allocate()` drops everything and completely re-initializes the object. The whole internal API contract you have now is: "oh, give me something maybe, and I'll throw it out" - doesn't sound quite elegant to me.

So I'm not proposing to pass a boolean parameter, I know this is going to be ugly as well (for the reasons you specified). I'm saying that *semantically* you are just providing a boolean.

I think my (super picky) concern could be resolved by just changing the `allocate` method to work additively instead of destructively. This way, the internal API contract is: "give me something maybe, and I'll add more stuff to it", which is sound.
</blockquote></details>

My understanding of this code is that if we were to pass a non-empty palette, overwriting it would be the correct thing to do, because the logic rely on the size being a certain number. If we were to append stuff we'd still overwrite the beginning of the buffer and leave the appended stuff in the end unitialized, because the offsets are already decided before we build the palette.
I think that the API contract here is really "give me something maybe, and I'll make it the current transform palette".

nical:
Try already looks good: https://treeherder.mozilla.org/#/jobs?repo=try&author=nsilva@mozilla.com
I wanted to avoid landing it at the same time as #3111 in case there is any other regression.

zptan:
I guess that. Just a reminder, I don't mean to push, :-)

PR #3093, opened 2018-09-20
Construct split planes with help of the inversed transform

### LINKED ISSUE #3070: Transform 3D precision issues
We have inconsistent plane splitting results (presumably) because of the thresholds/precision.
Original bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1491088

kvark:
Fixes #3070
Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=5eb3a20402b513ccfa2d526b14f32ed707821e92

Meat of the fix is within `plane-split` 0.13.1 update, which we use in this PR in order to construct consistently transformed polygons for splitting.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3093)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -973,7 +973,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "plane-split"
-version = "0.13.0"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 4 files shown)


kvark:
Try looks good :)
Also, rebased now.
@gw3583 please take a look

gw3583:
@bors-servo r+

@kvark Looks good, thanks! I wonder if there is a way to add a wrench reftest for this as a follow up?

kvark:
@gw3583 great idea! this shouldn't be too hard :)

PR #3095, opened 2018-09-20
Intersect the update rect with each tile's dirty rect when updating an image.

nical:
Otherwise each tile gets a dirty rect even if the tile isn't affected. This confuses the logic that requests/upload tiles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3095)


nical:
try push https://treeherder.mozilla.org/#/jobs?repo=try&revision=266d75942682baaa9602c63eacc08e5ad26ab6b5

gw3583:
Try run looks good.

@bors-servo r=aosmond

PR #3099, opened 2018-09-21
Use raster roots for text shadows.

gw3583:
For shadows, establish a new rasterization root. This ensures
that text shadows are rasterized in local space, which in
turn makes them suitable to be cached between frames and
display lists when the content doesn't change.

The actual caching will be done as a follow up.

Also change the API from GlyphRasterSpace to RasterSpace, since
in the future we'll use it to signify that the entire stacking
context should be rasterized in local space, not just the glyphs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3099)


gw3583:
r? @kvark 

Try run here:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c3902b4c5df90fe36ec31e3228bf6ed66673d0a7

Looks good so far - didn't build on mac because I didn't update the vendored deps correctly, but it should be OK since it's good on Linux / Windows.

Compared to current baseline against master:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c79849adeec4766874c1f2f8a355056aa997eb8f

kvark:
Great, :shipit: :)

gw3583:
Added 3 more pixels of fuzziness on mac for one test.

@bors-servo r=kvark

PR #3100, opened 2018-09-21
Some simple but important gradient optimizations.

gw3583:
If all the gradient stop colors have an alpha of 0, then skip
adding that gradient to the frame builder, since it can't
affect any pixels. I have observed Gecko sending these in large
numbers on some pages.

Also, check for opacity on the stop colors and try to draw
gradients in the opaque pass where possible. This can have
a large performance win on many sites, especially sites that
use gradient backgrounds.


gw3583:
Also opened https://bugzilla.mozilla.org/show_bug.cgi?id=1493068

kvark:
Thanks for opening the Gecko issue!
@bors-servo r+

PR #3101, opened 2018-09-21
Make the Checkpoint enum repr(u32).

nical:
Forgot to do that but I want it accessible across ffi.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3101)


nical:
r? anyone

PR #3105, opened 2018-09-21
support Windows variation fonts

lsalzman:
This uses the new variation font support in DWrote so that we can fix Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1441323

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3105)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -334,15 +334,6 @@ version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "dwrite-sys"
-version = "0.2.0"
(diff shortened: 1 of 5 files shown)


lsalzman:
This is because Pathfinder is pulling in an older version of dwrote. I think the only way to fix that is for him to update the deposit. Nothing I can do on this side.

lsalzman:
Rather him = @pcwalton

pcwalton:
I'm on it.

pcwalton:
Updated the `webrender` branch of Pathfinder to the latest version of `winapi`. Let me know if you run into any problems with it.

PR #3107, opened 2018-09-22
border: Add a way to avoid anti-aliasing border corners.

emilio:
I need this for table border-collapsing
(https://bugzilla.mozilla.org/show_bug.cgi?id=1487407).

I only need solid borders and such, so for now I restricted the implementation
to that since it's way simpler and allowed me to only modify the border_solid
shader.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3107)


webrender_api/src/display_item.rs
```diff
@@ -262,6 +262,11 @@ pub struct NormalBorder {
     pub top: BorderSide,
     pub bottom: BorderSide,
     pub radius: BorderRadius,
+    /// Whether to do anti-alias on the border corners.
```
gw3583: nit: Whether to apply anti-aliasing ....

emilio:
> What should we do in the case that the border is a transformed (e.g. rotated) element.

I don't think gecko does that, and that'd cause seams on the corners which we don't really want to. So I think respecting the no-aa is fine for now at least.

emilio:
I fixed the nit. I assume this is r=you @gw3583? (Not sure, so given I'm still waiting for some Gecko changes to get review I'd rather wait).

Gankro:
it's a bit weird that you use a bool in the wr API, but in the gecko bindings introduced an enum to represent this?

Gankro:
for reference: https://phabricator.services.mozilla.com/differential/changeset/?ref=155872&whitespace=ignore-most

emilio:
Yeah, I prefer avoiding boolean arguments, but inside WR it's not used as an argument anywhere.

@bors-servo r=kvark

In any case if it becomes confusing later I can change it to make it an enum everywhere.

PR #3110, opened 2018-09-24
Introduce PrimitiveInstance in picture runs.

gw3583:
The goal of this work is to start separating the data stored in the Primitive struct into instance and template data.

The idea is that Pictures will store PrimitiveInstance structures, instead of just primitive indices. These instance structures will eventually contain information relevant to this specific instance of a primitive (for example, the combined local clip rect, the clip task ID etc), and the ID of the interned template data. 

What remains in the Primitive structure will then be treated as primitive template data. It will be hashed and interned, in the same way that clip nodes are. This will mean:

- We can quickly compare if the primitive in a picture is the same (by comparing the interned primitive ID, and a small amount of per-instance data). This can also be done during scene building, rather than frame building, for most primitive types.
- The GPU cache handles (and contents) for primitive templates are persisted between display lists, significantly reducing the amount of GPU cache updates.
- Pages that contain many duplicated primitive template structures will be more efficient, since they'll be de-duplicated by the interning process in the scene builder.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3110)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -19,7 +19,7 @@ use picture::{PictureCompositeMode, PicturePrimitive, PictureSurface};
 use plane_split::{BspSplitter, Clipper, Polygon, Splitter};
 use prim_store::{BrushKind, BrushPrimitive, BrushSegmentTaskId, DeferredResolve};
 use prim_store::{EdgeAaSegmentMask, ImageSource, PrimitiveIndex};
-use prim_store::{PrimitiveMetadata, VisibleGradientTile};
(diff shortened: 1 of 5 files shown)


gw3583:
r? @kvark or @nical 

Pending try (not checked yet):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=793c6a58ee92131d3dca5cc0394a805e63f1f4ba

gw3583:
There shouldn't be any functional changes here - but I've split it up into smaller self contained commits that may be easier to review individually.

PR #3108, opened 2018-09-22
Temporarily revert #3079 for a crash workaround.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3108)

PR #3111, opened 2018-09-24
Store a queue of rasterized blobs instead for non-tiled images (take two)

nical:
The 3 commits from #3079 which were reverted in #3108 plus a commit that fixes the crash that motivated the backout.

Bugzilla entry: https://bugzilla.mozilla.org/show_bug.cgi?id=1493177

In the texture cache we have an automatic and a manual eviction policy. The manual eviction policy ensures blob images don't get automatically evicted from the cache which could otherwise race with the asynchronous rasterization. The Texture cache was missing a tiny piece of code to ensure the eviction policy was respected in for shared cache entries (it does work for normal cache entries). I bet border images make uses of these shared entries. While scrolling the image eventually gets discarded and next time we see it we request the image. The problem is that when we decide whether a blob image is missing, we only look at whether there is an entry in the rasterized_blob_image map, but we don't check whether that entry contains any actual data to upload in its queue.

Before the regressing commit, the bug would occur without crashing, but we wouldn't necessary upload all of the image (only the last available dirty region). With that commit we end up requesting something we think we have but don't have and later panic with an empty handed upload request.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3111)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -979,7 +979,7 @@ impl ResourceCache {
             let request: BlobImageRequest = request.into();
             let missing = match (self.rasterized_blob_images.get(&request.key), request.tile) {
                 (Some(RasterizedBlob::Tiled(tiles)), Some(tile)) => !tiles.contains_key(&tile),
-                (Some(RasterizedBlob::NonTiled(..)), None) => false,
+                (Some(RasterizedBlob::NonTiled(ref queue)), None) => queue.is_empty(),
                 _ => true,
             };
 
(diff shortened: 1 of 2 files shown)


jrmuizel:
FWIW, I was thinking about this approach some more and think it would probably be better to just keep a copy of the whole rasterized blob around in memory instead of having the accumulating queue of updates

nical:
This is blocking a bunch of other things, so if we want to pursue another approach, let's first land it and then implement something else so that the other work can land.

I'm rather neutral about whether we should retain the whole blob in memory and paint over it or do this queud thing. In principle the former feels cleaner. It might be harder to implement because it isn't clear which thread can own the retained rasterized blobs. Also in theory it involves copying more memory but I'm not sure whether that really matters.

jrmuizel:
I agree that we should land this first.

kvark:
@nical looks reasonable! Are you confident to land this without a try push?

nical:
I have both the confidence and the [try push](https://treeherder.mozilla.org/#/jobs?repo=try&author=nsilva@mozilla.com) (although the latter has commits from another PR in addition to this one).

nical:
What's up @bors-servo ?

staktrace:
Mac CI is busted. Some sort of expired cert problem when the worker tries to upload the log to taskcluster.

Issue #3116, opened 2018-09-24
Gecko's displaylist_mutate talos test is slow


mattwoodrow:
Looking at profiles, it appears that both the scene builder and render backend thread take a very long time, mostly within resizing vectors.

This test has a large number of elements (10k), and they're mostly the same every frame, so is testing the overhead of unchanged elements.

I'm going to have a go at preallocating things where possible, and using SmallVec for the ones that are usually small.

gw3583:
@mattwoodrow made a lot of allocation improvements here. There's still work we can do here, but we can open specific bugs next time we profile it.

PR #3117, opened 2018-09-25
Improve scene building and frame building performance by reducing allocations

mattwoodrow:
This roughly triples our FPS on the displaylist_mutate talos test, and goes a long way towards resolving #3116.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3117)

### DIFF
--- a/webrender/src/border.rs
+++ b/webrender/src/border.rs
@@ -11,7 +11,7 @@ use ellipse::Ellipse;
 use euclid::SideOffsets2D;
 use display_list_flattener::DisplayListFlattener;
 use gpu_types::{BorderInstance, BorderSegment, BrushFlags};
-use prim_store::{BrushKind, BrushPrimitive, BrushSegment};
+use prim_store::{BrushKind, BrushPrimitive, BrushSegment, BrushSegmentVec};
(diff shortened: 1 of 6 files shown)


gw3583:
Looks good! r=me once CI passes.

gw3583:
Oh, some compile errors in release builds. Must be missing some debug attributes, I guess:

```
error[E0609]: no field `initialized` on type `&mut segment::SegmentBuilder`
   --> webrender/src/segment.rs:212:18
    |
212 |             self.initialized = true;
    |                  ^^^^^^^^^^^
error[E0609]: no field `initialized` on type `&mut segment::SegmentBuilder`
   --> webrender/src/segment.rs:408:28
    |
408 |         debug_assert!(self.initialized);
    |                            ^^^^^^^^^^^
error[E0609]: no field `initialized` on type `&mut segment::SegmentBuilder`
   --> webrender/src/segment.rs:549:18
    |
549 |             self.initialized = false;
    |                  ^^^^^^^^^^^
error: aborting due to 3 previous errors

```

pcwalton:
Note that there is the `arrayvec` crate if we want to both allocate on the stack and ensure that a vector *never* grows beyond a certain capacity.

PR #3122, opened 2018-09-25
Use the texture filtering hardware to reduce the number of sample operations in the blur shader.

pcwalton:
Linear filtering allows us to evaluate the formula

    k₀c₀ + k₁c₁                 (Equation 1)

where c₀ and c₁ are the colors of adjacent texels and k₀ and k₁ are arbitrary
factors (in this case, the results of evaluating the Gaussian function) with a
single lookup. Linear filtering evaluates the following expression for some t
between 0 and 1:

    lerp(c₀, c₁, t)

It can be shown algebraically that Equation 1 is equivalent to:

                 ⎛           k₁  ⎞
    (k₀ + k₁)lerp⎜c₀, c₁, ───────⎟
                 ⎝        k₀ + k₁⎠

Which can be readily evaluated by letting `t = k₁/(k₀ + k₁)` and performing a
texture lookup.

This results in some minor visual differences because we always compute an even
number of pixels on each side (i.e. the *support* is now always rounded up to
the nearest even number of pixels), whereas before we might have evaluated an
odd number of pixels. All differences should therefore be minor quality
improvements.

Don't merge this yet; I want to run tests and a try build.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3122)


pcwalton:
Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=6920b63af9336557d1937429b59706ee4f6cf5d1

mstange:
If then new blur is just as "correct" as the software reference blur, it's probably not worth making them match. It would be better to spend the time on making sure that WebRender is used for both the test and the reference ([by making SVG filters go through WebRender](https://bugzilla.mozilla.org/show_bug.cgi?id=1485512).)

pcwalton:
Much better now. Just one test has imperceptible differences, probably due to rounding.

pcwalton:
Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=6f72ba3298e9ca75a6cd4bc86e5be4946fcda3e6

pcwalton:
Just a few minor fuzziness differences on try. This looks good.

r? @gw3583

kvark:
Windows reftests are failing though.

gw3583:
@pcwalton It looks like this was basically ready to go but then got overlooked. If that's the case, I guess rebasing and doing a try would be good to make sure it's still fine, and then we can get this merged?

pcwalton:
It was ready to go but I wasn't sure about try results. Let me try again.

gw3583:
@pcwalton Did you get a chance to kick off another try for this?

pcwalton:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=bea495521277680f7db1a04b5d7457ca66016696

gw3583:
From a quick look at the try run:

```1``` and ```2``` failures look like unrelated intermittents.
```R1``` and ```R3``` on OSX are new passes (less fuzzy than the current annotations).
```R7``` on Linux is a new fail - the max pixel difference is the same as the current reftest annotations, but there are more pixels that need to be marked fuzzy.

I think these changes are fine, but we should get @staktrace or @jrmuizel to double check. Also looks like it needs a rebase to fix the conflicting reftest image.

**Comment by staktrace** [comment]
> From a quick look at the try run:
> 
> `1` and `2` failures look like unrelated intermittents.

Yes although I don't see these failures anywhere else so I'd like to see another try push rebased on central to confirm. And this try push was actually based on mozilla-beta rather than mozilla-central, so it may not even be accurate in terms of what it shows.

> `R1` and `R3` on OSX are new passes (less fuzzy than the current annotations).
> `R7` on Linux is a new fail - the max pixel difference is the same as the current reftest annotations, but there are more pixels that need to be marked fuzzy.

Agreed, with the caveat that these results are from mozilla-beta. If we get the same results on mozilla-central that's fine.

**Comment by gw3583** [comment]
@staktrace Thanks for double checking those.

@pcwalton Looks like we need a try run against m-c, and to fix the rebase conflict above, then this should hopefully be good to go.

**Comment by gw3583** [comment]
Try looks good (although will get @staktrace to double check again). Just need to resolve that conflicting png above then this is ready, I think.

**Comment by staktrace** [comment]
Yup, try push looks good to me (I can adjust the reftest numbers on those two tests when I merge it to m-c).

**Comment by pcwalton** [comment]
Patch updated.

**Comment by gw3583** [comment]
One of the reftest fails on Linux CI and will need fuzziness of ..... 1 pixel .... max difference 1 :man_facepalming: 

```
REFTEST TEST-UNEXPECTED-FAIL | reftests/text/two-shadows.yaml == reftests/text/two-shadows.png | image comparison, max difference: 1, number of differing pixels: 1
```

PR #3124, opened 2018-09-26
Add memory reporters and eviction for cached textures

bholley:
Reviewed by Glenn in https://bugzilla.mozilla.org/show_bug.cgi?id=1493196

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3124)

### DIFF
--- a/webrender/src/device/gl.rs
+++ b/webrender/src/device/gl.rs
@@ -487,6 +487,12 @@ impl Texture {
         self.last_frame_used == frame_id
     }
 
+    /// Returns true if this texture was used within `threshold` frames of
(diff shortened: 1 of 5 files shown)


bholley:
@bors-servo r=gw3583

tidy

bholley:
On that note, how do I run tidy locally?

gw3583:
You should be able to do something like:

```
virtualenv venv
./venv/bin/activate.sh
pip install servo-tidy PyYAML voluptuous
servo-tidy
```

Probably missing something but that should roughly work - run it from the root WR dir.

PR #3126, opened 2018-09-26
Round blur support up to the nearest even number of pixels.

pcwalton:
I'm pretty confident this is correct--it's either just as accurate or more
accurate than the current implementation--so it'll let me sort out which test failures, if any, from #3028 and #3122 are true regressions.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3126)


pcwalton:
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=829f974e0204b435f968a6956ae4a0e5f37e7f9c

pcwalton:
How do I mark various failing Gecko tests as fuzzy?

mstange:
With a Gecko patch that modifies the `reftest.list` files in the directory of the test.

pcwalton:
@mstange Does that need to land before or after the corresponding patch lands in the WR repository?

gw3583:
I generally get @jrmuizel or @staktrace to sign off on the fuzziness increases being OK here in a comment, and then they update them as part of the WR update.

mstange:
It needs to land together with the WebRender update that pulls that change into Gecko. So just attach the patch to the [wr-future-update](https://bugzilla.mozilla.org/show_bug.cgi?id=wr-future-update) bug and Jeff or Kats will include it.

pcwalton:
Gotcha, thanks.

pcwalton:
I verified all test differences are imperceptible.

PR #3128, opened 2018-09-26
Add capability to permit namespace id allocation by client

sotaroikeda:
RenderApiSender::create_api() does sync ipc with ApiMsg::CloneApi message. It blocks clinet thread. The thread is sometimes blocked more than 10ms. It is not good :( It is nice if client side could allocate namespace id. It could make clone api faster with gecko.

[Bug 1493327](https://bugzilla.mozilla.org/show_bug.cgi?id=1493327)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -400,6 +400,7 @@ pub struct RenderBackend {
     recorder: Option<Box<ApiRecordingReceiver>>,
     sampler: Option<Box<AsyncPropertySampler + Send>>,
     size_of_op: Option<VoidPtrToSizeFn>,
+    namespace_alloc_by_client: bool,
 
     last_scene_id: u64,
 }
@@ -419,6 +420,7 @@ impl RenderBackend {
         recorder: Option<Box<ApiRecordingReceiver>>,
         sampler: Option<Box<AsyncPropertySampler + Send>>,
         size_of_op: Option<VoidPtrToSizeFn>,
+        namespace_alloc_by_client: bool,
     ) -> RenderBackend {
         RenderBackend {
             api_rx,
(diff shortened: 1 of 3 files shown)


webrender_api/src/api.rs
```diff
@@ -864,6 +867,23 @@ impl RenderApiSender {
             next_id: Cell::new(ResourceId(0)),
         }
     }
+
+    /// Creates a new resource API object with a dedicated namespace.
+    /// Namespace id is allocated by client.
+    ///
+    /// The function could be used only when RendererOptions::namespace_alloc_by_client is true.
+    /// When the option is true, create_api() could not be used to prvent namespace id conflict.
```
kvark: "prvent" typo

webrender/src/render_backend.rs
```diff
@@ -400,6 +400,7 @@ pub struct RenderBackend {
     recorder: Option<Box<ApiRecordingReceiver>>,
     sampler: Option<Box<AsyncPropertySampler + Send>>,
     size_of_op: Option<VoidPtrToSizeFn>,
+    namespace_alloc_by_client: bool,
```
kvark: is there any reason not to *always* allocate it on the client side?

webrender_api/src/api.rs
sotaroikeda: Thanks.

webrender/src/render_backend.rs
sotaroikeda: If client always allocate namespace id(32bit), the id allocation is not easy when RenderApiSender::create_api() is called from several different processes.

In gecko's use case, create_api() is always called by single process, then the id allocation is easy by client side.

kvark:
It feels quite insecure atm. The only check we do is for existing documents. We don't check if there are any API senders with the custom namespace, and equally we don't check if a namespace was used as a custom one when we advance the namespace counter... That being said, the use-case seems strong, and we can make it more robust as a follow-up.

@bors-servo r+

PR #3131, opened 2018-09-26
Revised backface visibility and 3D plane ordering

### LINKED ISSUE #3114: Backface visibility is not propagated properly
This is a WR tracking bug for the following Gecko issues:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1492030
- https://bugzilla.mozilla.org/show_bug.cgi?id=1454706

kvark:
Fixes #3114

## Backface visibility logic changes

  - the property is not propagated hierarchically
  - the property is inherited only by the pictures directly spawned from a primitive
  - backface-visibility is determined in the coordinate space of the containing block. For 3D-participating contexts, this is the parent of the 3D sub-tree root.

~~This isn't obvious, and neither it is guaranteed to be correct. It's just a rather simple scheme that makes us behave properly on the issues listed in #3114. Feedback is welcome!~~
Update: this should be correct. [Gecko changes](https://github.com/servo/webrender/pull/3131#issuecomment-426479546) are also needed though.

~~Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=f25ab9510acbc46b3a3c4a8a03eb23c7b75fc4ae~~

cc @Gankro @mstange (and thanks for fruitful discussion!)

## Plane splitting re-architecture

All the interactions with `plane_split` crate have been moved from batching code to `prepare_xxx_for_render`. The code is better structured now, with most of the plane splitting logic in one place inside a (relatively small) `picture` module, organized as helper functions. Batching still does a bit of work to actually make draw instances from the ordered planes, but it's not longer a computational work, it's just shuffling data.

## Plane split logic changes

Whenever we iterate the children of a stacking context that participates in plane splitting, and we encounter a child which is another stacking context, we break the immediate children sequence and organize a new picture from them. Any following immediate children will be going to another picture, therefore, and the order or rendering is preserved this way.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3131)


kvark:
AAAnd we have a fail in /css/css-transforms/transform3d-backface-visibility-006.html 
So this new model needs more thought (marking as WIP)

Darkspirit:
https://codepen.io/desandro/pen/LmWoWe
Regression: Not only the text flickers, now there is even a red & blue flicker on the back side.

kvark:
Worked quite a bit more on the issue, made it function in accordance to the [draft spec](https://drafts.csswg.org/css-transforms-2/#backface-visibility).
Pending try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=3ca6af02e87b720df1f145b7662900eaa6ce712c

emilio:
Not sure how much the draft spec is web-compatible, probably worth checking with @mattwoodrow.

Gankro:
The spec has historically been inconsistent with the implementations, which is why Matt and I didn't bother looking at it

kvark:
I've been looking at [backface-visibility-2](https://searchfox.org/mozilla-central/source/layout/reftests/transform-3d/backface-visibility-2.html) myself and with @mstange , and it looks like the failure is caused by Gecko not properly propagating the backface visibility property.

Supposedly, error happens here:
https://searchfox.org/mozilla-central/rev/6ddb5fb144993fb5de044e2e8d900d7643b98a4d/layout/painting/nsDisplayList.cpp#8767

The front face element is red color and has the backface visibility disabled. It's parented to the "card" div that establishes a preserve-3d context, and that is rotated backwards.

Observed behavior: Gecko spawns a (flat) stacking context (SC) around it, which gets backface visibility enabled. Since this SC serves as a preserve-3d hierarchy boundary, and the element still has backface visibility disabled, it turns out to be invisible (regardless of the transforms).

Expected behavior: either Gecko specifies the element as is (without wrapping into SC), or the backface-visibility property gets moved up to the SC wrap.

@staktrace would you happen to know how (or if?) this should be fixed on Gecko side? Attaching the display lists dump generated by Gecko and passed to WR. Alternatively, I'll keep poking at it.

[dl-dump.txt](https://github.com/servo/webrender/files/2438885/dl-dump.txt)

staktrace:
@kvark - I'm not totally following your explanation. What I see in the display list dump is this (simplified):

```
div id=card     --> produces rotation SC with is_backface_visible = true
  div id=front  --> produces rect with is_backface_visible = false
  div id=back   --> produces rotation SC with is_backface_visible = false
                --> also produces rect with is_backface_visible = false
```

This seems like it matches the HTML exactly, since the backface-visibility:hidden is set on two inner divs and not the card div. Are you saying that we should be propagating the `is_backface_visible=false` to the outer SC? If so, what happens to other elements that may live inside it and that have `is_backface_visible=true`?

staktrace:
... although I may have oversimplified things. There's a lot of reference frames and duplicate stacking contexts as well.

mstange:
`div id=front` also produces a stacking context in line 30 of the dump:

```
        GenericDisplayItem { item: PushStackingContext(PushStackingContextDisplayItem { stacking_context: StackingContext { transform_style: Flat, mix_blend_mode: Normal, clip_node_id: None, raster_space: Screen } }), clip_and_scroll: ClipAndScrollInfo { scroll_node_id: Spatial(4, PipelineId(1, 7)), clip_node_id: None }, info: PrimitiveInfo { rect: TypedRect(0.0×0.0 at (0.0,0.0)), clip_rect: TypedRect(0.0×0.0 at (0.0,0.0)), is_backface_visible: true, tag: None } }
```

This stacking context has `is_backface_visible: true`, which is the problematic part here.

kvark:
Third try push - https://treeherder.mozilla.org/#/jobs?repo=try&revision=e1c69444dcd1e9d574b3d6b1d796613bc26c081f

kvark:
Suggested Gecko patch by @staktrace :
```diff
diff --git a/layout/painting/nsDisplayList.cpp b/layout/painting/nsDisplayList.cpp
--- a/layout/painting/nsDisplayList.cpp
+++ b/layout/painting/nsDisplayList.cpp
@@ -8686,16 +8686,24 @@ nsDisplayTransform::ShouldBuildLayerEven
 bool
 nsDisplayTransform::CreateWebRenderCommands(
   mozilla::wr::DisplayListBuilder& aBuilder,
   mozilla::wr::IpcResourceUpdateQueue& aResources,
   const StackingContextHelper& aSc,
   WebRenderLayerManager* aManager,
   nsDisplayListBuilder* aDisplayListBuilder)
 {
+  if (mNoExtendContext) {
+    // This is a wrapper nsDisplayTransform that separates descendants from
+    // the 3D context of the ancestors. WR doesn't need a stacking context for
+    // this so we can just generate the WR items for the descendants.
+    return mStoredList.CreateWebRenderCommands(
+        aBuilder, aResources, sc, aManager, aDisplayListBuilder);
+  }
+
   // We want to make sure we don't pollute the transform property in the WR
   // stacking context by including the position of this frame (relative to the
   // parent reference frame). We need to keep those separate; the position of
   // this frame goes into the stacking context bounds while the transform goes
   // into the transform.
   LayoutDevicePoint position;
Matrix4x4 newTransformMatrix = GetTransformForRendering(&position);
```

kvark:
Investigated a bit more, found a problem with WR side code. Fixed, rebased, and pushed here.
Local testing shows everything to be great :crossed_fingers: 
New try push, including the Gecko patch above: https://treeherder.mozilla.org/#/jobs?repo=try&revision=eceaeb8eb63881833e6c752ec92ecd04788a82ec

kvark:
Here is what's going on with `split-non-ortho1.html`. It has non-transformed items (inside preserve-3D context) mixed with transformed ones, but transformation is only within the 2D plane (so nothing to plane split or sort, really). The expectation of the test is that the order elements are provided in DOM is preserved then when rendering them.

In the old (Gecko!) code, each of the elements would spawn a separate stacking context, and WR is able to preserve the order of these when plane-splitting.

In the new Gecko code, some elements no longer have a separate stacking context, so WR groups them into the same picture, which participates in ordering and plane splitting as a whole... thus, the drawing order gets broken.

TL;DR; the proposed Gecko patch unveils a problem in WR 3D ordering that was hidden before. Either the patch needs to be rewritten (i.e. by propagating backface-visibility into the spawned contexts), or WR needs to be fixed w.r.t. ordering in this case. I'd prefer the latter, and I'm looking into it, but it's unfortunate that this very PR is not quite strongly related to this blocking fix.

gw3583:
@kvark Interesting problem! A couple of comments:

* There is code in WR that *supposedly* deals with having un-transformed content inside a 3d rendering context in addition to planes that undergo plane splitting. I had to implement this to get some of the WPT tests passing. It's quite likely it's broken in some (perhaps most) cases though! It's a bit subtle to describe in a GH comment, but perhaps ping me tomorrow and we can discuss the way it currently is meant to work, and why it's failing in this case.

* I'm working on a patch in preparation for picture caching at the moment that does alter how we push / pop stacking contexts during flattening. It's probably not relevant here, but just raising it now so we can include that in the discussion mentioned above.

kvark:
@gw3583 thanks for the info! Let's talk this over tonight on IRC.

In the meantime, the way our 3D stacking contexts are treated, I think, should be the following:
  - in `push_stacking_context` if the relative transform to the previous one doesn't have Z component, then just really flatten it into the parent, don't spawn a new plane for splitting
  - if the relative transform does have Z, then "break the picture" as in - close the one containing all untransformed items, and then open a new one if any show up after that stacking context.

That scheme would preserve the ordering with regards to the most 3D cases (outside of obscure ones where there is an inversely transformed context inside a transformed one, and the app expects the order to be preserved ... yikes!).

gw3583:
@kvark I certainly haven't looked into this in detail, so I don't have a full understanding of the problem yet. That said, when I implemented the picture code for this, it seemed like it should match the specification and not require any workarounds like you've mentioned above.

I think you already ran into this btw, but the draft spec is quite different from what browsers implement. The spec I used as a reference is https://drafts.csswg.org/css-transforms-2/#transform-functions (linked from the MDN docs at https://developer.mozilla.org/en-US/docs/Web/CSS/transform). I had a quick read again and it seems like we should be matching the specification as far as I can tell (most likely I'm missing something though, since we clearly have an issue here!).

The other thing to mention regarding your comment above - I don't think we can change any behavior in ```push_stacking_context``` that is based on the transform values. This is because the flattening is done once per scene and doesn't know the future value of the transforms. For example, property animation could be used to change the transform completely without a new scene being built. So, *if* we do need to do something like this, it'll probably need to be in the picture preparation code.

Next time we're online, let's discuss this in detail - I also have some other semi-related thoughts on how to optimize plane-splitting in the common case, that may end up factoring into this stuff.

kvark:
Rebased (basically re-writing most of the code), added a new reftest for the 3D ordering problem, and made some preparations for the implementation. Working on it...

kvark:
@Gankro correct. I'll shout out when it's ready for looks ;)

kvark:
New try push (take 5): https://treeherder.mozilla.org/#/jobs?repo=try&revision=5fcfc040f2fd24165d4d3dff7182333a9130f0bb

kvark:
I forgot to include the Gecko patch in this^ one. Here is an updated try push (take 6):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=dae7575337569f60cfe7fa5585be818f59f5a068

The PR should be ready for review now! Although much depends on the try push outcome.

kvark:
It doesn't fix https://3dtransforms.desandro.com/card-flip (must be something else in play), but HackerOne is good.

kvark:
Looks like there are 2 issues: one with not properly respecting opacity filters, and another in `transform3d-sorting-004.html`. Switching back to WIP until this is sorted (pun!).

kvark:
The opacity is a legitimate failure, should be fixed now. However, the [sorting-004](https://searchfox.org/mozilla-central/rev/eef79962ba73f7759fd74da658f6e5ceae0fc730/testing/web-platform/tests/css/css-transforms/transform3d-sorting-004.html#10) looks like a Gecko bug:
the case has a wrapping `<div>` that makes the transformed Z of the child divs irrelevant. However, it doesn't appear that Gecko provides this div in the display list. Looking at "scene-1-0.ron" I see that it's missing, hence the test fails.

[scene-1-0.ron.zip](https://github.com/servo/webrender/files/2489130/scene-1-0.ron.zip)

The PR now includes a wrench test for the same behavior, and it passes (as expected).
@staktrace any idea how Gecko patch should be adjusted to not skip that stacking context?

staktrace:
@kvark I'm not sure I know that answer to that. I suspect that the gecko display list might not have any items for that div since it doesn't seem to have anything special about it (in terms of styles). If you can get a gecko+WR display list dump it would probably help confirm that. I'm not sure how gecko handles the "special behaviour" of that div in terms of flattening the contents in the non-webrender scenario, @mattwoodrow might be able to provide more info on that.

gw3583:
(Just a quick skim of the easy bits for now)

mattwoodrow:
> @kvark I'm not sure I know the answer to that. I suspect that the gecko display list might not have any items for that div since it doesn't seem to have anything special about it (in terms of styles). If you can get a gecko+WR display list dump it would probably help confirm that. I'm not sure how gecko handles the "special behaviour" of that div in terms of flattening the contents in the non-webrender scenario, @mattwoodrow might be able to provide more info on that.

I had a look at the .ron file, and it seems like the coloured rectangles (red and lime) are contained within a  PushStackingContext(transform_style: Flat).

Is that not what is needed? Is it because PushReferenceFrame is outside of PushStackingContext?

gw3583:
@kvark I took a *very quick* look at the implementation here and it looks great! I'll take a more detailed look once the remaining issues are resolved, but in general this seems like the right approach to me. It should also give us the benefit of being able to skip plane splitting in cases where it doesn't affect the output (as a follow up if needed). :1st_place_medal:

kvark:
@mattwoodrow yes, they both are withing Flat stacking contexts, this is fine. What is expected is that between those and the Preserve3D parent, there is another Flat parent, and I'm not seeing this in the capture. Here is this case translated into the reftest (which is a part of the PR, it tests this behavior):
```yaml
+# The "preserve-3d" context has a "flat" child, which contains two other stacking contexts
+# that have their Z modified. The Z should not affect their drawing order, since they
+# are hidden from the 3D hierarchy by the flat parent.
+---
root:
  items:
    - type: stacking-context
      bounds: [0, 0, 100, 100]
      transform-style: preserve-3d
      items:
        - type: stacking-context # <--- missing from Gecko output
          bounds: [0, 0, 100, 100]
          items:
            - type: stacking-context
              bounds: [0, 0, 100, 100]
              transform: translate(0, 0, 10)
              items:
                - type: rect
                  bounds: [0, 0, 100, 100]
                  color: red
            - type: stacking-context
              bounds: [0, 0, 100, 100]
              transform: translate(0, 0, -10)
              items:
                - type: rect
                  bounds: [0, 0, 100, 100]
                  color: green
```

kvark:
@staktrace this DL dump?  [dl.txt](https://github.com/servo/webrender/files/2489922/dl.txt)

staktrace:
That's the DL for the parent (chrome) process. Set gfx.webrender.dl.dump-content to true and you should get the content process DL with the generated WR DL items interleaved.

But I guess what Matt is saying is that the "flat" gets carried down to wrap the children individually instead of wrapping all the children combined, which might be what you're expecting.

kvark:
@staktrace here is a proper one :crossed_fingers:  [dl.txt](https://github.com/servo/webrender/files/2489966/dl.txt)

> But I guess what Matt is saying is that the "flat" gets carried down to wrap the children individually instead of wrapping all the children combined, which might be what you're expecting.

I'm expecting a single "flat" child of "preserve-3d", then 2 "flat" children beneath it, each carrying a solid rect (of different color).

staktrace:
Here's one which has the interleaved WR DL items: [dl-better.txt](https://github.com/servo/webrender/files/2492093/dl-better.txt) that I got from the latest macOS nightly

It shows that there is a stacking context just like you expect (on line 22) that is Flat and contains the other per-rect reference frame items. Do you have any local changes that might be affecting this?

(Also getting that dump was entirely too hard, I'll try and make it easier)

staktrace:
> @staktrace yes, I have your patch, which is likely causing the issue here.

Ah yeah that would explain it. Unless you need the patch for other testcases you can drop it. If you need it for other testcases we can figure out a way to modify it to accomodate all the necessary scenarios.

kvark:
@staktrace I needed the patch for the backface visibility to work (see comments above), since Gecko doesn't seem to be propagating the backface visibility correctly for the stacking context wrapping an item. Let's discuss this again and figure out a proper Gecko patch.

kvark:
After a bit of back-and-forth discussion with @staktrace  (and @mattwoodrow - thanks as well!), it appears that the best outcome is achieved by not patching Gecko at all, which I suppose is good news :)
Here is a fresh try push with all the recent changes that has no Gecko tweaks:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7b4743a04e494eba6c1d193fb517dadcc92373eb

kvark:
@staktrace bummer, there is a few failures in the latest try push:
  - opacity-preserve3d-1.html
  - opacity-preserve3d-4.html
  - transform3d-sorting-004.html

I looked at "opacity-preserve3d-1" and I see that the "div" element that has both opacity and preserve-3d is translated into 2 different stacking contexts by Gecko. As a result, we apply the opacity, while the test expects us not to.

kvark:
That last try is also incorrect. The rebasing is tricky because a small part of the change went in separately in #3206. When I rebase on earlier code, I'm losing the shader side changes, but everything compiles without errors :/

Fixed try push is here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=b02514f0ae0689a9d69023f32dd2ea7ee7caa924

kvark:
That try is looking great, at last! r? @gw3583

gw3583:
@kvark OK, this looks good to me! There are a few nits in the review to solve, but that's all that remains. The other question is whether we want to merge this so close to the beta merge date for Gecko. Do we have any thoughts on possible risk / whether we should hold off on merging this until some point later in the week?

kvark:
Thanks for the review @gw3583 !

> The other question is whether we want to merge this so close to the beta merge date for Gecko. Do we have any thoughts on possible risk / whether we should hold off on merging this until some point later in the week?

AFAIK, Beta is branching out today, and in order for this code to ship there we'd need another WR update in FF. I'd like this code to stay a bit in Nightly before shipping. Confirmed with @staktrace on IRC that we aren't planning on a WR update before the branch. Any further hotfixes can be cherry-picked, so the PR isn't blocking us to do anything. Let's proceed :steam_locomotive: 

@bors-servo r=gw3583

PR #3134, opened 2018-09-27
Remove the scroll bar APIs.

gw3583:
It's much better if applications draw scrollbars by adding normal
primitives to the display list. This means:

* Better control of how they look visually.
* Can make them clickable / draggable.
* Less code complexity in WR.
* Better control of scrollbar behavior (where enabled etc).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3134)


gw3583:
This reduces code complexity for some other ongoing work (since scroll bar codes relies on global primitive indices).

However, let's not merge this until we have a resolution to my questions in https://github.com/servo/servo/issues/21817.

Gankro:
Seems great, r=me when you're ready to land

PR #3137, opened 2018-09-27
Collect clip nodes for any surface, not just raster roots.

gw3583:
The existing code collects ancestor clips whenever we encounter
a rasterization root. This patch changes the behavior to
collect ancestor clips for any picture that has a surface.

This is a subtle but important step towards picture caching. It
ensures that any clips that are positioned by ancestors of the
surface positioning node are collected and applied when the
surface is composited into the parent, rather than on the
items in the surface itself. (The CSS specifications ensure
that this will give the equivalent result).

What this means is that a picture can be cached and re-used
on a subsequent frame, even if the clips in ancestor nodes
have moved relative to the positioning of the cached picture,
since those clips are only used during composition and do
not affect the contents of the picture itself.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3137)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -422,20 +422,20 @@ impl ClipStore {
         &self.clip_node_instances[(node_range.first + index) as usize]
     }
 
-    // Notify the clip store that a new rasterization root has been created.
+    // Notify the clip store that a new surface has been created.
     // This means any clips from an earlier root should be collected rather
     // than applied on the primitive itself.
-    pub fn push_raster_root(
+    pub fn push_surface(
         &mut self,
-        raster_spatial_node_index: SpatialNodeIndex,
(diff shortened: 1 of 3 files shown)


gw3583:
r? @kvark 

I did a Linux-only try run previously with this change, and it looks good (just has some expected failures from the recent blur changes):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=426f4aa5dcd07ec3e7238e863dfee571081ae0be

I have also kicked off a full try run (pending):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a02e214f9a6ee5d101dd2708536e6a245340a6cb

kvark:
> Try run looks good, I think - appears to be the usual assortment of unrelated intermittent failures mixed with the expected differences due to the recent blur changes

wow, the usual assortment is pretty scary

PR #3139, opened 2018-09-28
Add document and frame resources.

gw3583:
We were previously passing the clip interner and data store
around directly. In the near future we'll have more interning
structures (e.g. text runs, gradients etc). With this change
it's easy to introduce new interning / data store types without
re-plumbing things through all the WR types.

This change shouldn't have any functional effect, it's just
refactoring for future changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3139)


gw3583:
@nical I assume that should include an r+ ?

nical:
Woops! @bors-servo r+

PR #3141, opened 2018-09-28
Sanitize shadow radii.

nical:
We sanitize blur radii but not shadows. This PR prevents large shadow radii from causing crashes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3141)


nical:
Fixes more of https://bugzilla.mozilla.org/show_bug.cgi?id=1493917

PR #3142, opened 2018-09-28
Switch to generic-worker for mac CI

staktrace:
Don't merge this yet, still testing. Also I want @jrmuizel to set up generic-worker on the toronto macmini as well before merging this otherwise we'll be back down to one worker, and doing that setup is blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1494996

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3142)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -128,7 +128,7 @@ tasks:
       owner: '{{ event.head.user.email }}'
       source: '{{ event.head.repo.url }}'
     provisionerId: 'localprovisioner'
-    workerType: 'kats-webrender-ci-osx'
+    workerType: 'webrender-ci-osx'
     extra:
       github:
         events:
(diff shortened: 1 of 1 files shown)


staktrace:
@jrmuizel you should have access to the google drive for the webrender-ci mozillians group now, which has a document on how to set up the generic-worker instance on the toronto macmini. The first set of steps are the same as before, just follow the steps from "Download the worker binaries" onwards. For now we'll want to run both generic-worker and taskcluster-worker in parallel until all the pending PRs and WIP branches that people have pick up this change. Note that generic-worker uses a separate per-task folder for scratch space so there shouldn't be any collisions if both workers are running tasks concurrently.

staktrace:
@jrmuizel says the generic-worker instance is up and running on the toronto macmini, so this should be good to merge now.

PR #3147, opened 2018-10-01
Pass a Box instead of Arc for the notification handler.

nical:
It's a bit of a detail but this lets us use `NotificationHandler`s that are `Send` but not `Sync`. We aren't currently taking advantage of the reference counting so the change is pretty straightforward.

The clone impl doesn't do an actual clone which might look odd but it's not particularly worse than the previous situation since these handlers are meant to be signaled once and only once, and since they are only used for synchronization purposes, it is okay to lose them in the case where clone happens (in wrench), just like they are skipped during serialization/deserialization.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3147)

### DIFF
--- a/webrender_api/src/api.rs
+++ b/webrender_api/src/api.rs
@@ -11,7 +11,6 @@ use std::fmt;
 use std::marker::PhantomData;
 use std::os::raw::c_void;
 use std::path::PathBuf;
-use std::sync::Arc;
 use std::u32;
 use {BuiltDisplayList, BuiltDisplayListDescriptor, ColorF, DeviceIntPoint, DeviceUintRect};
 use {DeviceUintSize, ExternalScrollId, FontInstanceKey, FontInstanceOptions};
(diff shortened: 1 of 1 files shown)


nical:
r? anyone

webrender_api/src/api.rs
```diff
@@ -1269,34 +1268,41 @@ pub trait NotificationHandler : Send + Sync {
...
-            done: false,
         }
     }
 
     pub fn when(&self) -> Checkpoint { self.when }
 
     pub fn notify(mut self) {
-        self.handler.notify(self.when);
-        self.done = true;
+        if let Some(handler) = self.handler.take() {
```
kvark: no need for `take` here

webrender_api/src/api.rs
```diff
@@ -1269,34 +1268,41 @@ pub trait NotificationHandler : Send + Sync {
...
     fn drop(&mut self) {
-        if !self.done {
-            self.handler.notify(Checkpoint::TransactionDropped);
+        if let Some(ref mut handler) = self.handler {
+            handler.notify(Checkpoint::TransactionDropped);
+        }
+    }
+}
+
+impl Clone for NotificationRequest {
```
kvark: let's have some comments on why this is needed and why it's safe to clone handler as `None`. Maybe we can have a more specific method instead of using the generic `clone()`?

webrender_api/src/api.rs
nical: `take` turns the handler into `None` which replace the `done` boolean in preventing the notification to be signaled multiple times (in particular during drop).

webrender_api/src/api.rs
nical: I added a comment. I'm neutral about whether or not we should have a more specific method (ideally we wouldn't have to clone the messages at all). If we don't use `Clone` we'll have to manually implement the new thing for a lot of types which can be tedious or maybe the custom derive mechanism (which I don't know much about) can help. Anyway, I'd prefer that this be done separately, There's some Gecko stuff I'm working on that need this PR.

PR #3149, opened 2018-10-01
Temporarily force enable raster in screen-space.

gw3583:
There are some robustness issues with the current code that
handles local space raster mode. Specifically, it doesn't
correctly support adjusting the scale when content requests
a local space picture that is larger than the maximum render
task size.

Until that's fixed, we'll temporarily revert to rasterizing
everything in screen space (except perspective rasterization
roots). This is no big deal as we don't yet take advantage
of this to cache the local space pictures anyway.

Once we add support for handling scale on local space pictures,
then we can remove this hack.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3149)


gw3583:
Try run looks good.

PR #3150, opened 2018-10-02
Pass device pixel scale via picture render task data.

gw3583:
As a follow up, we will be able to include a per-picture
local scaling factor. This can be used for drawing pictures
in local space at a specific scale factor.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3150)

### DIFF
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -42,20 +42,6 @@ ImageBrushData fetch_image_data(int address) {
     return data;
 }
 
-#ifdef WR_FEATURE_ALPHA_PASS
-vec2 transform_point_snapped(
-    vec2 local_pos,
-    RectWithSize local_rect,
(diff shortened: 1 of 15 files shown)


gw3583:
r? @kvark 

*Pending* try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=8b3ba71cedc7c3af0efcaf1e4879e3c339a8e146

gw3583:
Try run looks good.

gw3583:
Filed https://github.com/servo/webrender/issues/3157 and rebased.

I agree with the testing comment. I did a fair amount of manual testing in wrench with difference ratios, and it seems OK, but I'm not sure how good our CI coverage is. Once I start to leverage this for rasterizing pictures at different scales, we can definitely get some good CI coverage via that.

PR #3153, opened 2018-10-02
Forward the ClearNamespace messages to the scene builder

nical:
This prevents the `ClearNamespace` message from racing ahead of resource updates that are going through the scene builder. Currently gecko avoids the race condition by flushing the whole pipeline before sending the clear namespace message, which is very expensive.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3153)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -661,6 +661,10 @@ impl RenderBackend {
                     SceneBuilderResult::ExternalEvent(evt) => {
                         self.notifier.external_event(evt);
                     }
+                    SceneBuilderResult::ClearNamespace(id) => {
+                        self.resource_cache.clear_namespace(id);
+                        self.documents.retain(|doc_id, _doc| doc_id.0 != id);
+                    }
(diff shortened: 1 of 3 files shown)


nical:
The commit also clears some blob image related data structures that weren't cleared previously, as well as the documents on the scene builder.
I also added `SceneBuilder::send` that does the pirouette of waking up the render backend thread after putting a message in the result queue which is easy to forget.

nical:
Try push with this + the gecko patch to remove the sync flush https://treeherder.mozilla.org/#/jobs?repo=try&revision=a9456f153a36d064d9ef0a163b7814b3d3adce63

nical:
The try push looks good with the exception of `layout/reftests/css-animations/continuation-opacity.html` which is https://bugzilla.mozilla.org/show_bug.cgi?id=1489327. The frequency is worrying, though. I'll look into it some more but since this test is failing in a timing sensitive way, so it's most likely to fluctuate with each of the synchronization stuff I'm going to land.

PR #3154, opened 2018-10-02
Fix ordering of transforms on a reference frame.

staktrace:
The updated comment explains why the new ordering is correct. Previously
the scroll offset transform was being applied after the local reference
frame transform, instead of before. A more detailed explanation of why
the old code was incorrect and the new code is correct can be found at
https://bugzilla.mozilla.org/show_bug.cgi?id=1415272#c23

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3154)

### DIFF
--- a/webrender/src/spatial_node.rs
+++ b/webrender/src/spatial_node.rs
@@ -256,12 +256,20 @@ impl SpatialNode {
                     .pre_mul(&source_transform.into())
                     .pre_mul(&info.source_perspective);
 
-                // The transformation for this viewport in world coordinates is the transformation for
-                // our parent reference frame, plus any accumulated scrolling offsets from nodes
-                // between our reference frame and this node. Finally, we also include
-                // whatever local transformation this reference frame provides.
+                // In order to compute a transformation to world coordinates, we need to apply the
+                // following transforms in order:
+                //   state.parent_accumulated_scroll_offset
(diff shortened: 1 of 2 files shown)


staktrace:
r? @kvark or @gw3583 

I wasn't sure how much detail to put in the code comment vs commit message, let me know if you want me to move more of the explanation from the bug into the code.

staktrace:
I did a try push at https://treeherder.mozilla.org/#/jobs?repo=try&bugfiler=&group_state=expanded&revision=fd156919846cdc951c47d94e2c8df7b0df64fa47 - one unexpected-pass which we can update the annotation for, rest is green.

kvark:
Awesome, thanks!
@bors-servo r+

kvark:
>  "\nSchema Validation Failed!\nRejecting Schema: https://schemas.taskcluster.net/docker-worker/v1/payload.json#\nErrors:\n  * data.env['RUST_BACKTRACE'] should be string"

Out of a sudden!

staktrace:
Weird, I guess somebody deployed stricter checking somewhere. I put up https://github.com/servo/webrender/pull/3156 that should fix it.

gw3583:
@staktrace Is it feasible to add a wrench test to exercise this? It'd be great to avoid regressing something this subtle :)

PR #3158, opened 2018-10-02
Add the option to share Shaders between Renderers and precache what we need

mattwoodrow:
This is a significant startup and window opening performance win for gecko.

Most of the work done by Dan, I just cleaned things up and made sure it passed tests.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3158)

### DIFF
--- a/examples/common/boilerplate.rs
+++ b/examples/common/boilerplate.rs
@@ -164,7 +164,7 @@ pub fn main_wrapper<E: Example>(
         DeviceUintSize::new(size.width as u32, size.height as u32)
     };
     let notifier = Box::new(Notifier::new(events_loop.create_proxy()));
-    let (mut renderer, sender) = webrender::Renderer::new(gl.clone(), notifier, opts).unwrap();
+    let (mut renderer, sender) = webrender::Renderer::new(gl.clone(), notifier, opts, None).unwrap();
     let api = sender.create_api();
     let document_id = api.add_document(framebuffer_size, 0);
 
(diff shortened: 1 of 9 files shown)


mattwoodrow:
I'll upload the gecko side changes needed for this into bug 1495902

examples/common/boilerplate.rs
```diff
@@ -164,7 +164,9 @@ pub fn main_wrapper<E: Example>(
         DeviceUintSize::new(size.width as u32, size.height as u32)
     };
     let notifier = Box::new(Notifier::new(events_loop.create_proxy()));
-    let (mut renderer, sender) = webrender::Renderer::new(gl.clone(), notifier, opts).unwrap();
+    let upload_method = webrender::UploadMethod::PixelBuffer(webrender::VertexUsageHint::Stream);
```
gw3583: Let's make the ```upload_method``` argument to ```Device::new``` be ```Option<UploadMethod>``` and then inside the constructor, we could have something like:
```
let upload_method = upload_method.unwrap_or( ... pixel buffer / stream default ... );
```

That way clients don't need to specify an upload method unless they care.

examples/common/boilerplate.rs
```diff
@@ -164,7 +164,9 @@ pub fn main_wrapper<E: Example>(
...
+    let device = webrender::Device::new(gl.clone(), opts.resource_override_path.clone(), upload_method, None);
```
gw3583: Perhaps instead of ```Renderer::new``` taking a ```Device``` it could be:

```
fn new(device_and_shaders: Option<(Device, Shaders)>, ...)
```

Which would mean that you only need to create the device if you're also pre-creating the shaders. If ```None``` is supplied, we'd create the device internally as it was previously.

PR #3159, opened 2018-10-03
Async link support for shader compilation.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3159)


gw3583:
Marked [WIP] because @mattwoodrow needs to test this on talos and make sure it hasn't broken the startup performance improvements.

mattwoodrow:
Confirmed that this still works perfectly on Talos!

I filed bug 1496283 for the Gecko side changes needed.

PR #3160, opened 2018-10-03
Reduce number of calls to build clip chain instances.

gw3583:
A clip chain instance is built for each primitive. Then, during
segmentation we build a clip chain instance for each segment.

However, if the segmenting process only ends up with one segment
(the common case) we are effectively duplicating work by building
a clip chain instance which will be the same as the primitive
clip chain instance.

Detect this case and just use the primitive clip chain instance
if only one segment was produced.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3160)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -519,6 +519,58 @@ impl BrushSegment {
             brush_flags,
         }
     }
+
+    pub fn update_clip_task(
+        &mut self,
+        clip_chain: Option<&ClipChainInstance>,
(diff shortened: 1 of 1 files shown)


gw3583:
r? @kvark or @nical 

*Pending* try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a19df15e4c9f5ef30b578e07ee42dbe49a71b61e

gw3583:
Try run looks good (one new PASS from a previous PR).

PR #3162, opened 2018-10-03
Expose priority to the blob rasterizer.

nical:
This will help us refine the heuristics about when to install work in the thread pool in gecko's blob rasterizer.


nical:
r? anyone

webrender_api/src/image.rs
```diff
@@ -343,7 +343,11 @@ pub trait BlobImageHandler: Send {
 /// A group of rasterization requests to execute synchronously on the scene builder thread.
 pub trait AsyncBlobImageRasterizer : Send {
     /// Rasterize the requests.
```
Gankro: If there's anything interesting to say about the semantics of priority, this is the place to document it, eh?

Gankro:
r=me, feel free to land as-is if you feel there's nothing interesting to document where I noted.

webrender_api/src/image.rs
nical: The semantics are a bit specific to working around a surprising behavior in rayon which I hope is temporary. I'll add a hand-wavey comment about this flag telling the rasterizer to schedule work in such a way that low-priority work doesn't block high priority work.

PR #3170, opened 2018-10-05
Skip frame rendering if nothing is changed

sotaroikeda:
This pull request add a capability to skip frame rendering if nothing is changed. It is for fixing [Bug 1461239](https://bugzilla.mozilla.org/show_bug.cgi?id=1461239). The change was review by @nical in bugzilla.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3170)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -139,6 +139,7 @@ struct Document {
     /// before rendering again.
     frame_is_valid: bool,
     hit_tester_is_valid: bool,
+    rendered_frame_is_valid: bool,
 
     resources: FrameResources,
 }
@@ -169,6 +170,7 @@ impl Document {
             dynamic_properties: SceneProperties::new(),
             frame_is_valid: false,
             hit_tester_is_valid: false,
+            rendered_frame_is_valid: false,
             resources: FrameResources::new(),
         }
     }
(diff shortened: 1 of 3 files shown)


webrender/src/render_backend.rs
```diff
@@ -1068,11 +1074,19 @@ impl RenderBackend {
         // Avoid re-building the frame if the current built frame is still valid.
         let build_frame = render_frame && !doc.frame_is_valid;
 
+        // Request composite is true when we want to composite frame even when
+        // there is no frame update. This happens when video frame is updated under
+        // external image with NativeTexture or when platform requested to composite frame.
+        if invalidate_rendered_frame {
+            doc.rendered_frame_is_valid = true;
```
nical: This should set the it to false, no? (missed that in bugzilla)

webrender/src/render_backend.rs
```diff
@@ -1111,6 +1125,13 @@ impl RenderBackend {
             self.result_tx.send(msg).unwrap();
             profile_counters.reset();
         } else if requested_frame {
+
+            // If frame is not rebuilt since last frame rendering and composite frame is
+            // not requested, there is no need to render frame.
+            if render_frame && !doc.rendered_frame_is_valid {
```
nical: I think that the meaning of `rendered_frame_is_valid` is reversed throughout the commit (here we should skip rendering if `rendered_frame_is_valid` is `true`).

sotaroikeda: Oh, I made mistake during update the patch in bugzilla.

PR #3172, opened 2018-10-05
Fix scrolling example to re-render on scroll

staktrace:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3172)


staktrace:
r? anybody

PR #3173, opened 2018-10-05
Apply a basis change on the perspective transform.

staktrace:
When applying async scrolling transforms, the perspective matrix also
needs to be adjusted with a basis change. This is a version of #3154
that doesn't break async scrolling on 2D-transformed elements. Although
this passes the required test cases and displays correct behaviour,
the intuition behind why this works is not really clear to me; see
https://bugzilla.mozilla.org/show_bug.cgi?id=1415272#c23 onwards for
some discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3173)


staktrace:
Try push is at https://treeherder.mozilla.org/#/jobs?repo=try&bugfiler=&group_state=expanded&revision=b5b03f9334ff26c20354a0f967c04a232c6fd30c

PR #3176, opened 2018-10-08
Remove unused LocalClip enum from API.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3176)


gw3583:
I confirmed this isn't referenced in either Servo or Gecko. It's no longer needed since we use clip nodes for anything other than the local clip rect in the display list.

r? anyone

PR #3177, opened 2018-10-09
delete FontInstances from FontContexts when evicting a GlyphCache

lsalzman:
This is to address Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1495661

If we have a page that is animating the variation values within a font, this can cause continuous accumulation of new FontInstances in the FontContext. These will never get cleared out until we delete the FontKey from which they were generated. This may never happen until the tab or window is closed, even though Gecko itself has long ago deleted the corresponding ScaledFonts to and told WR to delete the FontInstanceKeys.

Due to the fact that zoom and device scale can cause a FontInstanceKey to no longer map to the FontInstance used to key the GlyphCache and the FontContext, there can be multiple FontInstances with sizes that can't be traced back to the original FontInstanceKey. This prevents us from doing a straightforward deletion based on the value immediately stored with the FontInstanceKey.

However, we already have a mechanism in place to deal with this in the GlyphCache, evicting GlyphKeyCaches that have also been evicted from the TextureCache, so that we have effective limiting in place of FontInstances that haven't been recently used. So all we need to do here is connect that to deleting the FontInstance in the FontContext when it is being evicted from the GlyphCache. In the worst cases this may cause a false positive that evicts something from the FontContext while it is still alive, but it can be rematerialized when necessary, and this should be relatively rare. This will also give us a natural hard limit on the number of backend API fonts we keep around in this scenario.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3177)

### DIFF
--- a/webrender/src/glyph_cache.rs
+++ b/webrender/src/glyph_cache.rs
@@ -4,7 +4,7 @@
 
 #[cfg(feature = "pathfinder")]
 use api::DeviceIntPoint;
-use glyph_rasterizer::{FontInstance, GlyphFormat, GlyphKey};
+use glyph_rasterizer::{FontInstance, GlyphFormat, GlyphKey, GlyphRasterizer};
 use internal_types::FastHashMap;
 use render_task::RenderTaskCache;
 #[cfg(feature = "pathfinder")]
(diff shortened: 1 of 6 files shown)


gw3583:
The change looks good. Do we know if this fixes the reported issue, or do we need to get it into gecko and see? And do we need a try run for this?

gw3583:
I'll add delegate to this so it can be merged if / when you hear back that it fixes the reported issue.

@bors-servo delegate+

PR #3181, opened 2018-10-10
Switch line decorations from clip mask to cached primitive.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3181)

### DIFF
--- a/webrender/res/cs_line_decoration.glsl
+++ b/webrender/res/cs_line_decoration.glsl
@@ -2,87 +2,57 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#include shared,clip_shared
+#include shared
 
 #define LINE_STYLE_SOLID        0
 #define LINE_STYLE_DOTTED       1
 #define LINE_STYLE_DASHED       2
 #define LINE_STYLE_WAVY         3
 
-varying vec3 vLocalPos;
-flat varying int vStyle;
+// Local space position
+varying vec2 vPos;
+
 flat varying float vAxisSelect;
+flat varying int vStyle;
 flat varying vec4 vParams;
-flat varying vec2 vLocalOrigin;
 
 #ifdef WR_VERTEX_SHADER
 
 #define LINE_ORIENTATION_VERTICAL       0
 #define LINE_ORIENTATION_HORIZONTAL     1
 
-struct LineDecorationData {
-    RectWithSize local_rect;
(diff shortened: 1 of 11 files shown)


gw3583:
This **isn't ready** for merge yet, but I'm putting it up now in case anyone has time to start looking over it.

Before it's ready for merge, I want to:
- [ ] Update the commit message and comments with some rationale for why this makes sense and how it relates to picture caching.
- [ ] Fix the 2 test failures in the try run (https://treeherder.mozilla.org/#/jobs?repo=try&revision=c4c4513da98fbe52a38ef1773bb9eb58d2ab366c).

gw3583:
I rebased, fixed up the test failures above, and did a full try which looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=89f1ccfd1b9912cccc995cdc1f56406d52bdc3b7

I then made a slight tweak to how we clip partial dots. I don't expect this to affect gecko but I have kicked off another try run just to be sure (pending):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c1c69920f9edae58b17295db80f31b7dea8c5048

I also updated the commit message with the following:
```
There's a number of reasons for doing this:
 - It's a performance improvement, since we use the render task
   caching system to draw line decorations once and re-use them.
   They can be re-used both between frames and also new display
   lists, since the caching is based on content. The cached
   image can also be shared between multiple text runs on the
   same page that use the same decoration.
 - Performance win by drawing fewer pixels. The code now draws
   only a single dash, dot or wavy line. The brush_image shader
   makes use of the repeat functionality to draw this across
   the width of the primitive.
 - Switching both line decorations and box shadows (future) to
   be a primitive effect rather than a clipping effect is going
   to simplify some of the picture caching work. Specifically,
   by ensuring that line and box-shadow decorations only exist
   at the primitive level and not in arbitrary clip chains, we
   can move a lot of the segment building and processing logic
   to be handled during primitive interning during scene
   building. This will make the code simpler as well as reduce
   the amount of work that's required during frame building.
```

So I think this is ready for review now, whoever of @kvark / @nical  / @Gankro wants to take a look.

gw3583:
The follow up try run looks good.

gw3583:
@kvark Thanks, pushed a commit to address those comments. I don't think this will be slower in any cases. Even in the case of different text colors, the cached version is just an alpha mask, and the color is applied during the brush_image compositing of the effect. Theoretically this could go in the alpha cache target, but I left it in the RGBA target, assuming it will result in better batching for the alpha pass since most images reference the RGBA texture.

PR #3182, opened 2018-10-10
 Add sRGB to linear and back conversion filters

nical:
Adds two filters implemented in brush_blend.glsl letting us convert between sRGB and linear color spaces, which we need to support a subset of the SVG filters through our CSS filters implementation.

Long term I'm not sure whether the best strategy would be to keep the extra passes to do the conversion or to add the input and output conversion directly in the other filters  through an extra flag. I suspect that it would be quite fast since the branch would be coherent.

In the mean time the current implementation is as simple as it gets, doesn't need any API change and is certainly faster than computing SVG blurs on the CPU.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3182)

### DIFF
--- a/webrender/res/brush_blend.glsl
+++ b/webrender/res/brush_blend.glsl
@@ -128,6 +128,23 @@ vec3 Brightness(vec3 Cs, float amount) {
     return clamp(Cs.rgb * amount, vec3(0.0), vec3(1.0));
 }
 
+// Based on the Gecko's implementation in
+// https://hg.mozilla.org/mozilla-central/file/91b4c3687d75/gfx/src/FilterSupport.cpp#l24
+// These could be made faster by sampling a lookup table stored in a float texture
+// with linear interpolation.
+
+vec3 SrgbToLinear(vec3 color) {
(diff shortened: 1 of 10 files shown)


nical:
r? anyone

webrender/res/brush_blend.glsl
```diff
@@ -154,6 +154,24 @@ Fragment brush_fs() {
...
+            vec3 c2 = pow(color / 1.055 + vec3(0.055 / 1.055), vec3(2.4));
+            color = mix(c1, c2, lessThanEqual(color, vec3(0.04045)));
+            break;
+        }
+        case 12: {
+            // Linear to sRGB
+            vec3 c1 = color * 12.92;
+            vec3 c2 = vec3(1.055) * pow(color, vec3(1.0 / 2.4)) - vec3(0.055);
+            color = mix(c1, c2, lessThanEqual(color, vec3(0.0031308)));
+            break;
```
pcwalton: nit: Perhaps these should be separate functions in order to match the style of the rest of the switch statement?

PR #3183, opened 2018-10-10
Stop reinitializing textures

bholley:
This was reviewed in [1]. @kvark, can you confirm that you don't have any additional feedback? Feel free to trigger landing if you don't.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1497665

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3183)


kvark:
My concerns have been addressed, and the only remaining one - about hashmap - is being discussed, but it's not a ton of work to bring back the vector indexation in case we need it.
@bors-servo r+

PR #3185, opened 2018-10-11
Ensure calling VertexDataTexture::update always guarantees texture() to succeed.

### LINKED ISSUE #3184: Reftests are panicking after #3183.
After #3183 running:

```
$ ../target/debug/wrench show reftests/text/blank.yaml
```

Panics with:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:355:21
```

Because in `bind_frame_data`, `frame.prim_headers.headers_float` is empty, and thus we don't initialize the texture at all. And we panic when calling `texture()` on the line below.

cc @bholley

emilio:
Alternative is stop assuming there's a texture, which we may or may not want to
do (happy to do that, but I know Bobby has in-flight patches for a bunch of the
texture stuff, so this probably is less breaking).

Fixes #3184

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3185)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -1324,13 +1324,17 @@ impl VertexDataTexture {
     }
 
     fn update<T>(&mut self, device: &mut Device, data: &mut Vec<T>) {
-        if data.is_empty() {
-            return;
-        }
-
         debug_assert!(mem::size_of::<T>() % 16 == 0);
         let texels_per_item = mem::size_of::<T>() / 16;
         let items_per_row = MAX_VERTEX_TEXTURE_WIDTH / texels_per_item;
+
+        // Ensure we always end up with a texture when leaving this method.
+        if data.is_empty() {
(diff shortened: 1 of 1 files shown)


webrender/src/device/gl.rs
```diff
@@ -1232,7 +1232,7 @@ impl Device {
                     0,
                     desc.external,
                     desc.pixel_type,
-                    pixels.map(texels_to_u8_slice),
+                    pixels.and_then(texels_to_u8_slice),
```
kvark: I don't think this is correct. If the slice is provided, we exactly X bytes of data. Just ignoring empty data slices and treating them as storage allocation without initialization is something we shouldn't need to do.

kvark:
@bors-servo r-
let's not create 0x0 textures :)

kvark:
Thanks!
@bors-servo r+

Issue #3087, opened 2018-09-19
Deformed and glichy terminal on stripe.com (at 280+% zoom)


Darkspirit:
Bugzilla: [bug 1492525](https://bugzilla.mozilla.org/show_bug.cgi?id=1492525)
URL: https://stripe.com/terminal
Probably regressed by: #3030

Using try build from [bug 1492389 comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1492389#c1) (which fixed [bug 1492016](https://bugzilla.mozilla.org/show_bug.cgi?id=1492016)):
[Screencast](https://bugzilla.mozilla.org/attachment.cgi?id=9010327)

Darkspirit:
With latest Nightly it's no longer deformed like in the screencast for me, so the mentioned regression range might be obsolete. But there is still some flicker on the left side of the purple terminal if you zoom in.

emilio:
This was fixed by #3187.

PR #3187, opened 2018-10-11
Text + perspective clipping fix.

emilio:
Much in the spirit of #2991.

Working on a test now, but it's taking annoyingly hard because all the capture stuff is broken for fonts here :(


emilio:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1494685.

mstange:
Does this also fix the bug reported in https://github.com/servo/webrender/issues/3087#issuecomment-423595274 ?

emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=aec12a535893af5891242026d1e2f27f071f55f2

emilio:
@bors-servo r-

 * want to wait to add a test-case.

kvark:
@bors-servo r-
let's wait for the try to finish, then r=me

emilio:
@mstange it does.

emilio:
Hmm, needs some work, I think I just need to account for the snap_offset, but will double-check.

emilio:
@bors-servo try

 * Yeah, that fixes it locally.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=efe5e3a2c559b0b07a3ee81aa6d884d336179db7

emilio:
There are some tests failing on automation, but the text tests on my machine are way more unreliable, so it's going to be fun...

(C'mon, even my own test is failing on automation, and I just took the screenshot myself :/)

```
	reftests/text/synthetic-bold-transparent.yaml == reftests/text/synthetic-bold-transparent-ref.yaml
	reftests/text/white-opacity.yaml == reftests/text/white-opacity.png
	reftests/text/snap-text-offset.yaml == reftests/text/snap-text-offset-ref.yaml
	reftests/text/snap-clip.yaml == reftests/text/snap-clip-ref.yaml
	reftests/text/perspective-clip.yaml == reftests/text/perspective-clip.png
```

emilio:
Ah, the snap offset can be in local space for untransformed glyphs... That's somewhat messy :(

emilio:
Ok, I understood what this code is trying to do. This should be the right thing :)

emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c3d29b194814cceef79aefaf8ccd8d732237d9a6

kvark:
So you ended up not fixing any more snapping offsets?

emilio:
Nope, the snapping offset bit was wrong, because glyphs are snapped before getting clamped (#3048), plus the offset may not be in the coordinate space we want (it may be transformed to local space if transforms are disabled).

So this is the right thing to do. I have a patch that moves normal brush shaders to snapping before clamping to see what effect it has on the reftests that WR fails right now due to snapping / rounding differences.

@bors-servo r=kvark

PR #3192, opened 2018-10-11
Make compute_tile_range crashes more verbose.

jrmuizel:
These are happening in the wild. Let's figure out what the unexpected
values are.

See bug 1498385.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3192)

### DIFF
--- a/webrender/src/image.rs
+++ b/webrender/src/image.rs
@@ -237,12 +237,12 @@ pub fn compute_tile_range(
     let t0 = point2(
         f32::floor(visible_area.origin.x as f32 * tw),
         f32::floor(visible_area.origin.y as f32 * th),
-    ).cast::<u16>();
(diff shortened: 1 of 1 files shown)


webrender/src/image.rs
```diff
@@ -237,12 +237,12 @@ pub fn compute_tile_range(
     let t0 = point2(
         f32::floor(visible_area.origin.x as f32 * tw),
         f32::floor(visible_area.origin.y as f32 * th),
-    ).cast::<u16>();
+    ).try_cast::<u16>().unwrap_or_else(|| panic!("compute_tile_range bad values {:} {:}", visible_area, tile_size));
```
emilio: Huh, don't you need `{:?}`? Or then wouldn't it be just `{}`?

jrmuizel: Yeah. Quite.

PR #3193, opened 2018-10-12
Remove two unused arguments from snapping code.

emilio:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3193)


emilio:
Err, nvm, it has another caller I need to cleanup, one second...

emilio:
That should do.

PR #2731, opened 2018-05-04
Fall back to RGBA format when BGRA isn't supported by GLES.

jdm:
This makes Servo builds run and display content in the Android emulator. The colours are incorrectly swapped as one might expect; I'm less clear on how to actually swap the underlying texture bytes to match the reported format. Fixes https://github.com/servo/servo/issues/20476.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/2731)

PR #3195, opened 2018-10-12
Switch to immutable texture storage and invalidate unused framebuffers

bholley:
Reviewed in https://bugzilla.mozilla.org/show_bug.cgi?id=1496168


jrmuizel:
```
thread 'main' panicked at 'assertion failed: texture.target == gl::TEXTURE_2D || texture.target == gl::TEXTURE_RECTANGLE ||
    texture.target == gl::TEXTURE_EXTERNAL_OES', webrender/src/device/gl.rs:1671:9
```

staktrace:
@bholley @kvark any ideas of the top of your head why this PR would result in all the glyphs and images turning into black rectangles on Android? https://bugzilla.mozilla.org/show_bug.cgi?id=1499785

bholley:
I'll comment in the bug.

kvark:
@staktrace re-reading the code, maybe we are using the wrong external format now?
Old code had `gl::BGRA_EXT` or `gl::RGBA`. The new code has `gl::BGRA` always as the external format. Although, this wouldn't turn them black, I suppose.

bholley:
> Old code had gl::BGRA_EXT or gl::RGBA. The new code has gl::BGRA always as the external format. 

I think the old code was nonsensical. IIUC, the external format describes the arrangement of pixels delivered by the application, which wouldn't change based on whether GL_EXT_texture_format_BGRA8888 is supported by the driver. ImageFormat::BGRA is going to have pixels in that order.

> Although, this wouldn't turn them black, I suppose.

Yeah, it would just swap around the colors. And that's exactly what jdm reported in #2731. :-)

I'm pretty sure the black means that the textures aren't getting created at all. Driver logs are probably the next step.

PR #3197, opened 2018-10-15
Cache the current batch for both opaque and alpha batch lists.

gw3583:
In the (near) future, we will want to support adding segments
where the texture may be different between segments (it is
currently assumed to be the same for all segments in a single
primitive). To do this efficiently, we need to be able to add
segments while switching batch only when necessary, so we now
cache the current batch in the batch lists, and re-use it if
the batch keys are compatible.

This is also a general optimization for opaque batches - this
patch allows the previously cached batch to be re-used for
different opaque primitives. This can avoid searching the
batch array when sequences of the same opaque primitive are
added.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3197)

### DIFF
--- a/webrender/res/brush.glsl
+++ b/webrender/res/brush.glsl
@@ -9,7 +9,7 @@ void brush_vs(
     int prim_address,
     RectWithSize local_rect,
     RectWithSize segment_rect,
-    ivec3 user_data,
+    ivec4 user_data,
     mat4 transform,
     PictureTask pic_task,
     int brush_flags,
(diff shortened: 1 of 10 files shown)


gw3583:
r? @nical or @kvark 

*Pending* try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=b0cfe8c974d0bc17a07125e0b0a5039c916242f4

This is about 7 steps removed from picture caching, but will make some necessary changes for that easier.

Specifically, we'll be able to make borders use a separate render task per edge / segment, which will allow segment building to take place during scene building, which will allow local clipping to be calculated during scene building, which will be important for some of the picture caching code I'm working on. :grimacing:

gw3583:
Added a follow up commit that adds the basic structure to using per-segment user data for the texture cache address in ```brush_image```.

Kicked off a try run with that new commit:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2c0b065affa4ddb78ce3eb15e4a7c5b090cd453e

PR #3200, opened 2018-10-15
Open code get_font_from_descriptor() to get more information from crashes

jrmuizel:
This should help with bug 1455848

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3200)


kvark:
CI is not happy, otherwise good to go

PR #3201, opened 2018-10-16
Build border segments during scene building.

gw3583:
This patch changes border primitives to be able to
build their segments during scene building, rather
than frame building.

The main complexity was that we previously had to
do this during frame building to calculate the
device-space texel offsets for the segments. Now,
we supply this information as a local space stretch
rect for segment images. This allows the segments
for a border (and all primitives) to be treated
as immutable after first construction.

This is a CPU optimization (more so once the border
primitives are interned between display lists), but
most importantly it means that all segments can
now be built during scene building.

This in turn will allow a lot of the general segment
building and local space clips to be handled during
scene building and primitive interning, which is
both a performance win and makes possible some of
the future picture caching changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3201)


gw3583:
r? @emilio

Since you're the last one who touched the border code, but
feel free to hand off if there's parts you would prefer
not to review :)

I *think* this handles the cases you fixed previously with
the size dependence by including the local space size of
the dashed / dotted stretch size in the cache key for
edges where appropriate. But if that's not sufficient
let me know!

A Linux-specific try looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=3c531836c2f5dd05f47ccd8ebb1d15cf30c22d53

I also kicked off a *pending* try on all platforms with the last couple tweaks I made:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4b389587c86c7053fc09180c948e8d39f3427013

gw3583:
Looks like there is one crash test failure, caused by an overflow - investigating now.

gw3583:
So far I'm not able to reproduce that overflow failure locally - will try again to reproduce tomorrow. I expect it to be something minor though, so this should still be ready for review in the meantime.

gw3583:
Ah, I finally managed to reproduce it - looks like somehow the ```total_size``` field can be < 0.0 in some very specific / random cases. I'll track down how this can happen and catch it earlier.

gw3583:
Gecko sometimes supplies a display list with a dashed border style where the size of the primitive is negative. I pushed a fix which catches this before the code that was overflowing.

Try run here now looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2146b8bb2e63028d57163c308b0e586d80c64552&selectedJob=205743952

gw3583:
@kvark Thanks for the review! Addressed all review comments, I think. So this should be good to go, unless @emilio wants to do a quick check over first to make sure I didn't break all his caching fixes for dashed and dotted borders?

webrender/src/border.rs
```diff
@@ -1233,31 +952,51 @@ fn add_corner_segment(
...
 
-    border_segments.push(BorderSegmentInfo {
-        task_rect,
-        segment,
-        radius,
-        widths,
-    });
+    if image_rect.size.width <= 0. || image_rect.size.height <= 0. {
+        return;
+    }
```
emilio: Just double-checking, it's fine to avoid pushing the border segment here since we request each segment to the render task cache separately, right? That's much nicer :)

The reason the previous code couldn't do that was because we'd just cache all the instances for all the segments at the same time, which meant IIRC that if any of the segments was skipped, another border with an actual border size but the same cache key could reuse the wrong instances...

emilio:
I think it looks fine, sorry for the lag here! I had actually added those comments earlier today, but somehow forgot to publish the review request.

webrender/src/batch.rs
```diff
@@ -1426,14 +1485,65 @@ fn get_image_tile_params(
     }
 }
 
+/// Either a single texture / user data for all segments,
+/// or a list of one per segment.
+enum SegmentDataKind {
+    Shared(SegmentInstanceData),
+    Instanced(SmallVec<[SegmentInstanceData; 4]>),
```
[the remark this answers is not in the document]
gw3583: Good point - I changed it to 8. It's possible it might show up in profiles - we can rearrange this code a bit if it ever does.

webrender/src/border.rs
gw3583: Yep, that's right - the brush / batch code didn't previously support a render task / texture per-segment, but we can take advantage of that now :)

PR #3203, opened 2018-10-16
Make the pinch-zoom API a frame message instead of a scene message.

### LINKED ISSUE #3202: Adjusting the pinch zoom should be a frame operation, not a scene operation
Pinch zooming, like async scrolling, should be done on frames and should not require a scene rebuild. Right now both `set_page_zoom` and `set_pinch_zoom` are scene ops.

staktrace:
Fixes #3202.

There might be other issues that are exposed by this, not sure yet. But it's a start.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3203)

### DIFF
--- a/examples/scrolling.rs
+++ b/examples/scrolling.rs
@@ -153,18 +153,30 @@ impl Example for App {
                 ..
             } => {
                 let offset = match key {
-                    winit::VirtualKeyCode::Down => (0.0, -10.0),
-                    winit::VirtualKeyCode::Up => (0.0, 10.0),
-                    winit::VirtualKeyCode::Right => (-10.0, 0.0),
-                    winit::VirtualKeyCode::Left => (10.0, 0.0),
-                    _ => return false,
+                    winit::VirtualKeyCode::Down => Some((0.0, -10.0)),
+                    winit::VirtualKeyCode::Up => Some((0.0, 10.0)),
+                    winit::VirtualKeyCode::Right => Some((-10.0, 0.0)),
+                    winit::VirtualKeyCode::Left => Some((10.0, 0.0)),
+                    _ => None,
+                };
+                let zoom = match key {
+                    winit::VirtualKeyCode::Key0 => Some(1.0),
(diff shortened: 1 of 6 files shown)


nical:
I think that it makes sense to move the pinch zoom to a frame message since `DisplayListFlattener::create_frame_builder` doesn't seem to access it. We should probably split the `DocumentView` struct into things that scene building needs and things that only matter for frame building.

staktrace:
Note that `create_frame_builder` does use the `accumulated_scale_factor()` here: https://searchfox.org/mozilla-central/rev/c9272ef398954288525e37196eada1e5a93d93bf/gfx/webrender/src/display_list_flattener.rs#212 which includes the pinch zoom value. I don't really understand the purpose of that though.

nical:
D'oh! You are right I spoke too fast. This affects `resolved_transform` in `SpatialNodeType::ReferenceFrame` which itself looks like it isn't affecting scene building (I think?). the `resolved_transform` member doesn't seem to need to be persisted at all. I tried removing the member and adapting the code in `SpatialNode::update_transform` to use a variable on the stack instead of overwriting the removed member and it compiles.

staktrace:
Ah yes, I remember seeing that `resolved_transform` thing when I was working on the parallax bug. I think the only place it's used as a member variable is in `print_node` and I don't know how important it is there, seems like just debugging output.

kvark:
> I tried removing the member and adapting the code in SpatialNode::update_transform to use a variable on the stack instead of overwriting the removed member and it compiles.

Should we do it first (or as a part of this PR) then before merging?

staktrace:
Sure, if you're ok with me taking it out of the `print_node` code I can add it to this PR.

staktrace:
Patches updated ^

kvark:
@bors-servo delegate+
(leaving a chance for a try push, if needed)

kvark:
Try isn't needed (as per IRC discussion).
@bors-servo r+

PR #3208, opened 2018-10-17
Add border segments to the opaque render pass where possible.

gw3583:
Use the new per-segment instance data to allow specifying a
per-segment opacity value. This is used for border segments
to specify which border segments are opaque.

Any time we move segments or primitives to the opaque pass
is typically a significant GPU time win due to the reduced
overdraw. It's also a CPU optimization during batching.


gw3583:
This is a decent GPU time win on some pages, and a small win on others.

r? @nical or @kvark

PR #3212, opened 2018-10-17
Shrink the sizes of the shared cache textures.

bholley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1499601


gw3583:
Looks like this has caused some CI failures:

```
REFTEST INFO | 319 passing, 2 failing
	reftests/boxshadow/box-shadow-stretch-mode-x.yaml == reftests/boxshadow/box-shadow-stretch-mode-x.png
	reftests/boxshadow/inset-mask-region.yaml == reftests/boxshadow/inset-mask-region.png
Traceback (most recent call last):
```

I haven't checked, but it might be sub-pixel differences caused by them being in a different allocation location in the texture cache. In which case, re-generating the PNG references would be fine.

bholley:
I can probably find a ten minute chunk to get back to this tonight. Are there instructions anywhere to regenerate the pngs?

gw3583:
@bholley You can run ```./script/headless.py png <yaml name>``` and that should generate a new reference image.

gw3583:
(I'll pull this branch and re-generate them later this afternoon if you don't get to it).

PR #3213, opened 2018-10-18
Store picture primitives in a vec, and refer by index in prims.

gw3583:
This is prep work for changing the structure of the prepare_prims
pass.

Currently, the code is (roughly):
 - Traverse primitive tree, recursing through pictures.
  - When visible, call prepare_prim to update GPU cache, clips etc.
 - Assign render tasks to passes.
 - Assign primitive lists for each picture to batches.

In future, it will instead be:
 - Traverse picture tree, but not individual primitives.
  - Determine surface allocation, plane-splitting, visibility.
 - For each visible picture, prepare_prims and add to batches.

This will simplify some existing code, as well as making possible
some requirements for picture caching (e.g. determining the scale
to rasterize a picture at before creating the child render tasks).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3213)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -507,7 +507,7 @@ impl AlphaBatchBuilder {
         for poly in splitter.sort(vec3(0.0, 0.0, 1.0)) {
             let prim_instance = &pic.prim_instances[poly.anchor];
             let prim_index = prim_instance.prim_index;
-            let pic_metadata = &ctx.prim_store.primitives[prim_index.0].metadata;
+            let prim = &ctx.prim_store.primitives[prim_index.0];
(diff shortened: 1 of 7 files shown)


gw3583:
r? @kvark 

This is a start point to what we were discussing earlier. A few points:
 * The code is going to be a bit messy in the interim in places - hopefully between the changes we discussed and your plane-splitting changes, it should be a lot tidier afterwards.
 * If this looks like it'll be difficult to rebase your plane-splitting patches over, we can just leave this open and I'll add further commits to this as I work through it.

gw3583:
@kvark Addressed most of those comments - I'm a little unsure about the second one though. I don't think there should be any leaks, but perhaps I'm misunderstanding the comment?

kvark:
Closing in favor of #3218

PR #3215, opened 2018-10-18
Avoid capturing platform-specific font data

kvark:
This PR avoids the capturing to carry or load the platform-specific font instance data. It's less ideal for the case where the replay is done on the same platform, but makes captures more robust and portable overall. Ideally, we'd want to still save it but ignore on other platforms, but this would be less trivial (TODO).
r? anyone

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3215)


kvark:
thanks!
@bors-servo r=gw3583

PR #3216, opened 2018-10-18
Retain texture cache entries if fewer than 32

squarewave:
Noticed this when looking at how the caching interacts with
documents. Correct me if I'm wrong but this looks unintended - at
the very least it doesn't seem to match the comment.

Additionally, this moves the sorting inside the > 32 clause, since
there's no point in sorting if we're going to retain them all.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3216)

### DIFF
--- a/webrender/src/texture_cache.rs
+++ b/webrender/src/texture_cache.rs
@@ -611,19 +611,20 @@ impl TextureCache {
             }
         }
 
-        // Sort by access time so we remove the oldest ones first.
-        eviction_candidates.sort_by_key(|handle| {
-            let entry = self.entries.get(handle);
-            entry.last_access
-        });
-
         // We only allow an arbitrary number of unused
         // standalone textures to remain in GPU memory.
         // TODO(gw): We should make this a better heuristic,
         //           for example based on total memory size.
         if eviction_candidates.len() > 32 {
+            // Sort by access time so we remove the oldest ones first.
+            eviction_candidates.sort_by_key(|handle| {
+                let entry = self.entries.get(handle);
+                entry.last_access
+            });
             let entries_to_keep = eviction_candidates.split_off(32);
             retained_entries.extend(entries_to_keep);
+        } else {
+            retained_entries.extend(eviction_candidates.drain(..));
         }
 
         // Free the selected items


gw3583:
I think you're right - it doesn't match the original intended behavior or the comment.

Having said that, changing this now to be less aggressive in dropping textures might increase our memory usage a bit. @bholley thoughts on whether we should land this now / hold off until beta merge / do some profiling / investigation of the performance / memory tradeoffs this changes?

It should certainly reduce our CPU compositor time + texture upload time, at the cost of some extra memory usage. I think it's probably worth having, although having the threshold be dropped to 8 or 16 might be safer for now?

bholley:
I think we should probably wait until after the merge, and possibly also until we settle on a larger strategy for texture cache memory management. That's on my todo list for next week, so I can report back.

nical:
> I think this is basically about plain old images above the 512 width/height threshold.

This is also my understanding.

> I didn't get a chance to track down which images get requested in a given frame

We request images when they are visible on screen (so the smaller of the two view/displayports which I will continue confusing until the end of times).

> Thoughts

Uploads can cause jank so I wouldn't want to be overly aggressive at evicting images if they are off screen for a single frame, but allocating too much GPU memory can put us in even worse situation. So Ideally we would be progressively more aggressive at evicting images the more memory is allocated, up to a point where anything off-screen gets evicted right away (but we only get there in really bad cases).
I don't think "32 images" is very meaningful threshold so if we can use an approximation in bytes I'd be happier.

bholley:
We discussed this extensively on IRC [1], and have a good plan. I'll write up some patches.

[1] https://mozilla.logbot.info/gfx/20181030#c15541654

gw3583:
@bholley @squarewave What do we want to do with this PR? Close it and incorporate it into the plan referenced above?

bholley:
Yeah, I'm working on eviction policy in my local tree, just keep getting distracted with things. It's near-done though, so we can go ahead and close this PR.

bholley:
Closing in favor of https://bugzilla.mozilla.org/show_bug.cgi?id=1504115

Issue #2852, opened 2018-06-27
Support tiled clip images


kvark:
When we request image for a clip source that is an image, we never supply a tile:
```rust
                ClipSource::Image(ref mask) => {
                    resource_cache.request_image(
                        ImageRequest {
                            key: mask.image,
                            rendering: ImageRendering::Auto,
                            tile: None, // <---------- HERE
                        },
                        gpu_cache,
                    );
                }
```
This breaks the resource cache expectations that a tiled image would only be requested with specific tiles.
cc @nical @aosmond

nical:
Drawing the tiles into an an intermediate surface and using the latter as the clip source would make sense here.

gw3583:
I haven't thought about this too deeply, but I don't think we need an intermediate surface. I think to support this, we need to:

(1) Determine which of the image tiles overlap the clip mask bounds (we should be able to reuse some of the current tile visibility code for this).
(2) In the clip batcher, loop over each of those tiles and draw them into the clip mask at the appropriate location.

nical:
I think that the clip mask and what I called the intermediate surface are the same thing in my head (I haven't touched the masking code much so I'm a bit fuzzy about the terminology)

emilio:
Will take a look at this.

PR #3219, opened 2018-10-19
Skip clip masks with tiled blobs, instead of crashing.

gw3583:
We don't currently support tiled blobs for clip masks (see
issue #2852). This patch detects that case, and skips applying
a clip mask in that case. It's not ideal, but it's better
than crashing inside the resource cache, which is the current
behavior.

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -145,11 +145,14 @@ impl From<ClipItemKey> for ClipNode {
                 )
             }
             ClipItemKey::ImageMask(rect, image, repeat) => {
-                ClipItem::Image(ImageMask {
-                    image,
-                    rect: LayoutRect::from_au(rect),
-                    repeat,
-                })
+                ClipItem::Image(
+                    ImageMask {
+                        image,
+                        rect: LayoutRect::from_au(rect),
+                        repeat,
+                    },
+                    true,
+                )
             }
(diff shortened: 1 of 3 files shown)


gw3583:
r? @nical or @jrmuizel 

I think this workaround is OK - the only issue would be if there are any situations in the resource cache where we currently do handle limited cases of clip mask images with tiling that this would regress. Otherwise, it should at least stop the crash for now, until we implement the proper solution in #2852.

*Pending* try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=fdf14a03c74120efb6f08dfcd7046aab63bb552d

gw3583:
cc @Gankro @mattwoodrow We'll need to fix #2852 before we consider enabling tiling for all blobs :disappointed:

gw3583:
The try run looks good.

webrender/src/clip.rs
```diff
@@ -760,7 +773,7 @@ impl ClipItemKey {
 pub enum ClipItem {
     Rectangle(LayoutRect, ClipMode),
     RoundedRectangle(LayoutRect, BorderRadius, ClipMode),
-    Image(ImageMask),
+    Image(ImageMask, bool),
```
nical: Please make the boolean a named field or add a comment, It takes a bit of looking around to find what it is for (batch.rs is the only place where the meaning is explicit, and it's not where I'd look first),

emilio:
I added the comment.

@bors-servo r=nical

PR #3220, opened 2018-10-19
Properly support tiling clip mask images.

### LINKED ISSUE #2852: Support tiled clip images

emilio:
Fixes #2852.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3220)


emilio:
r? @gw3583 / @nical / @kvark 

I'd prefer not to land it yet, since the code is basically copy-pasted from image tiling, and it could get some cleanup.

Just want to confirm this approach is the right way to fix this.

I also need to write some reftests for this, but testing of blob images is none as of right now... So I'll add some Gecko reftests to the bug.

If this an acceptable solution I'll clean up and such before get it landed.

emilio:
I think this should be basically ready for review. I added a bunch of cleanup commits that made me feel less bad about copying the image tiling code, and a bunch of tests (will add reftests that check the blob image stuff in Gecko after this lands).

Once thing that may be problematic is that if I understand correctly this I'm drawing the whole visible area for each tile just to `discard` the pixels afterwards... I think there must be a way to do better, but not sure how... Also that wouldn't necessarily need to be part of this PR I guess.

In any case... r? :)

emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a1029ae05c0e88b6cfe84f27120ba25e45b0b558 has this patch.

gw3583:
This basically looks good to me (a couple of minor nits). It'd be good to get @nical to take a look over the image tiling changes, if possible, as well.

nical:
Looks good to me!

emilio:
Thanks!

@bors-servo r=gw3583,nical

gw3583:
Trying to wake up bors

PR #3221, opened 2018-10-20
Make batch caching based on z id rather than prim index.

gw3583:
This is a necessary change for picture caching, because in the
future pictures won't have a primitive index. It is also a
better way of expressing the intent of the batch caching anyway
(that the overlap detection should be invoked next time we have
a primitive with a different z value).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3221)


gw3583:
Rebased on top of the backface visibility changes.

@bors-servo r=kvark

emilio:
(trying to wake up bors, again)

gw3583:
Manually merging since bors is asleep. I manually rebased this on the last manual merge and ran the reftests locally.

PR #3224, opened 2018-10-23
Support local data during interning.

gw3583:
This doesn't currently store any data, but it adds the support
in the interner for future use.


gw3583:
This doesn't achieve anything useful with the new functionality, just yet. But I'm trying to land isolated patches where possible to make reviews easier. If you'd prefer, we can wait until the follow up that makes use of this functionality is ready, before reviewing / merging this.

r? @kvark or @nical

nical:
Looks reasonable. @bors-servo r+

PR #3225, opened 2018-10-23
Shrink render targets

bholley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1501073


bholley:
On IRC, kvark said the reftest wasn't something we should worry about.

@bors-servo r=kvark

gw3583:
@bholley Instead of marking it fuzzy, could you re-generate the PNG? I know it takes a bit of time, but it's always good to get a notification when some pixel rendering changes, even if we then decide it's just subpixel accuracy.

bholley:
Unfortunately it's not a PNG. We discussed this more here: https://mozilla.logbot.info/gfx/20181023#c15510918

One improvement would be to update the reftest harness, since dbaron recently added support for more precise fuzziness ranges [1], which would at least tell us if the test stopped being fuzzy.

[1] https://groups.google.com/forum/#!topic/mozilla.dev.platform/By3AkwbLhCo

PR #3226, opened 2018-10-24
Don't update max_size for tasks drawn into the texture cache

bholley:
Previously, we sized all render target textures using max_size. This
changed, and now we size them based on the union of the allocator's used_rect
for each slice.

At present, the code updates max_size for every render task. But it's
only used to size the per-slice allocators, which in turn are only used
for RenderTaskLocation::Dynamic. Accounting for this fixes a panic
(wherein check_ready can fail because the texture is smaller than
max_size) and also allows more-efficient render target texture sizing.

Differential Revision: https://phabricator.services.mozilla.com/D9608

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3226)


bholley:
@bors-servo r=gw3583

pathfinder.

PR #3228, opened 2018-10-24
Move Primitive methods to PrimitiveInstance.

gw3583:
This is a small change towards reducing implementation details
on the Primitive struct. Eventually, Primitive will be removed
and we'll just have PrimitiveInstance and PrimitiveTemplate.

The PrimitiveTemplate struct will be the interned and shared
data for primitives that are the same across display lists.

The PrimitiveInstance struct currently contains information
calculated in prepare_prims and then used by batching. Once
we move the prepare prims code from the first (picture tree
pass) into the batching pass, most of the storage in the
PrimitiveInstance can be removed and used immediately.

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -18,7 +18,7 @@ use picture::{Picture3DContext, PictureCompositeMode, PicturePrimitive, PictureS
 use prim_store::{BrushKind, BrushPrimitive, BrushSegmentTaskId, DeferredResolve};
 use prim_store::{EdgeAaSegmentMask, ImageSource};
 use prim_store::{VisibleGradientTile, PrimitiveInstance};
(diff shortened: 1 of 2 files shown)


gw3583:
r? @nical or @kvark 

This probably looks like bike-shedding, but there's method to the madness. I'm hoping to be able to land some of the more substantial work I have for picture caching locally in the next few days.

gw3583:
This shouldn't have any functional effect at all, but I'll kick off a try run later tonight to be on the safe side.

gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=8cd6fc109eb4ab854b9fba715744ed3da10c7bd2

gw3583:
Try looks good.

@bors-servo r=nical

PR #3229, opened 2018-10-25
Introduce picture traversal pass.

gw3583:
This introduces a new pass during frame building. It's a very
quick pass, since we only traverse the picture tree. This pass
must not access individual primitives.

In the future, the prepare_prims pass will be split into two
parts. Some of the work done in that pass will be moved to this
picture traversal pass (e.g. plane splitting), while the rest
(e.g GPU cache updates) will be moved to the batching pass.

The outcome of the first pass will determine whether a picture
has a valid cached surface, and thus decide whether the second
primitive preparation and batching pass is required to build
a new surface cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3229)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -23,7 +23,7 @@ use gpu_types::BrushFlags;
 use hit_test::{HitTestingItem, HitTestingRun};
 use image::simplify_repeated_primitive;
 use internal_types::{FastHashMap, FastHashSet};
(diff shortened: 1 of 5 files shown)


gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=cbc178b2c07a0ce2e5ca28507257f21890adc751

gw3583:
Looks like this has a couple of failures on try that I need to investigate.

gw3583:
Try run isn't quite finished but looks good so far.

PR #3230, opened 2018-10-25
Scale box-shadow to maximum of 2048 device pixels.

djg:
Fixes https://bugzil.la/1490007

`box-shadow` weren't being clipped nor clamped to the maximum dimension of resource cache. This PR scales box shadows in the same manner that borders are scaled. The scaling is to the maximum dimension of the resource cache and not to 2048 like borders.

Increased the render task sanity check size from 16000 to 16384.

Reftest is mostly to catch that we don't regress and hit the sanity check panic.

This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3230)


djg:
I spoke with @nical and @gw3583 and have changed the size clamp from the maximum supported texture size to 2048. This will affect the quality of rendering, but Glenn was of the opinion that the case that triggered the `panic!` is a synthetic pathological case that will very rarely happen in the real world, so the quality degradation is acceptable.

djg:
In the test case from https://bugzil.la/1490007, I added rendering of a `<div>` to understand how the rounded-rect is rendered in webrender.  I noticed that the rendering of rect looks correct compared to the `box-shadow` applied to it.

Rendering the round-rect, I verified that the huge values are passed through to rendering, yet the code doesn't `panic!`
https://github.com/servo/webrender/blob/62af01cdf4f09f8f403e4f66f067e8db7ae42329/webrender/src/clip.rs#L391-L395

`rect` was `{ x= -19158, y= 164, w= 20200, h= 200 }` and `radius` was `{ top-left = { w= 20200, h= 200 }, ... }`

Yet when it comes to creating the render task for the rect in `fn new_mask(...)`, the `outer_rect` is used:
https://github.com/servo/webrender/blob/20214bf371002786cfd8daabf58c815bf1ee84a3/webrender/src/render_task.rs#L571-L579

which in my test case was the clipping rectangle for the visible area, `{ x= 0, y=478, w= 3040, h= 584 }`.

The `box-shadow` render task is instead trying to render the entire rounded-rect into a texture before applying blur, which is causing the `panic!`.  It seems possible that rendering quality could be improved for this case by only rendering the clipped part into the shadow, but as @gw3583 said, having huge border radius is very unlikely in practice.

kvark:
@djg code looks ready to me.
Do you have concerns left? Could you launch a try push?

djg:
@kvark no concerns from me. 

https://treeherder.mozilla.org/#/jobs?repo=try&revision=570acf1b68318b04a12500825da22b71f0ae103c&selectedJob=209075983

kvark:
Thanks!
@bors-servo r+

djg:
> error: Connection to server timed out
> error: Could not compile `osmesa-src`.

:-(

PR #3233, opened 2018-10-25
Prepare opacity consistently

kvark:
This is a follow-up to #3232 
It makes opacity computation more functional, having us to consider it equally important for all primitive types. It's not supposed to change any logic.
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=6e4ca6f1b473522e93d5fd11e64c563c2bdf50dd

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3233)


bholley:
This seemed to have slightly improved performance [1]. Was that at all expected?

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1502344#c4

PR #3236, opened 2018-10-28
Refactor chase tracing to work with primitive instance IDs.

gw3583:
In the process of interning primitives, the PrimitiveIndex
will become less useful, eventually being removed since a
primitive instance will contain just a reference to an interned
primitive template.

Having the chase ID based on the PrimitiveIndex is complicating
some of that refactoring, so this patch switches it to create
a unique ID per primitive instance, which is then used by the
chase tracing code. The primitive ID only exists when the
tracing code is compiled in.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3236)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -31,10 +31,27 @@ use renderer::{MAX_VERTEX_TEXTURE_WIDTH};
 use resource_cache::{ImageProperties, ImageRequest, ResourceCache};
 use scene::SceneProperties;
 use std::{cmp, fmt, mem, ops, usize};
+#[cfg(debug_assertions)]
+use std::sync::atomic::{AtomicUsize, Ordering};
 use util::{ScaleOffset, MatrixHelpers};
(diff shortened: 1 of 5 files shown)


gw3583:
r? @kvark 

Does this seem reasonable to you?

webrender/src/prim_store.rs
```diff
@@ -1620,29 +1645,37 @@ impl PrimitiveInstance {
...
         }
     }
+
+    #[cfg(debug_assertions)]
+    pub fn is_chased(&self) -> bool {
+        PRIM_CHASE_ID.load(Ordering::SeqCst) == self.id.0
+    }
+
+    #[cfg(not(debug_assertions))]
+    pub fn is_chased(&self) -> bool {
```
emilio: May be worth to `#[inline]` so that we ensure the compiler can remove the dead code.

kvark:
Reviewed as a part of #3244. I think we can close this one.

PR #3238, opened 2018-10-29
 Introduce primitive clustering and use for some basic cases.

gw3583:
Primitive clusters are used to group primitives within a picture
into similar categories during scene building.

Initially, these are only used to group primitives by spatial node
and by backface flags. In future, these will include spatial
clustering, which will allow removal of much of the per-primitive
bounding rect and culling code during frame building.

In this patch each primitive belongs to one cluster only, but
when spatial clusters are introduced, a primitive will belong
to one or more cluster(s). Clusters are processed during the
initial picture traversal, allowing the picture bounding rect to
be found without touching all primitives.

There's a couple of goals of this work:
- Reduce the amount of comparison work we need to do to check
  if a cached picture surface is valid.
- Allow cached picture surfaces to be drawn in tiles, to
  handle cases where only a small portion of a picture
  contents has changed.

There's also a number of other small related changes:
- Store offscreen surface information outside the picture primitives.
  - Simplifies borrow check issues on picture traversal.
  - Eventually the surface structures will be retained between display lists, and hold cached surface details.

- Build minimal local clip rect for primitives during scene building.
  - Used to allow quick calculation of picture bounding rects.
  - Can take advantage of this to reduce work during frame building (not done yet).

- Picture bounding rect calculation is done during initial picture traversal pass.
  - Much faster than previously, as we only need to calculates bounds for each cluster.

 - Backface visibility is done once per cluster now, rather than per-primitive.
 - Intern the primitive local rect and clip rects for future use.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3238)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -480,7 +480,7 @@ impl AlphaBatchBuilder {
         let task_address = render_tasks.get_task_address(task_id);
 
         // Add each run in this picture to the batch.
-        for prim_instance in &pic.prim_instances {
+        for prim_instance in &pic.prim_list.prim_instances {
             self.add_prim_to_batch(
                 prim_instance,
                 ctx,
@@ -598,7 +598,7 @@ impl AlphaBatchBuilder {
                             // Convert all children of the 3D hierarchy root into batches.
                             Picture3DContext::In { root_data: Some(ref list), .. } => {
                                 for child in list {
-                                    let prim_instance = &picture.prim_instances[child.anchor];
+                                    let prim_instance = &picture.prim_list.prim_instances[child.anchor];
                                     let pic_primitive = &ctx.prim_store.primitives[prim_instance.prim_index.0];
 
                                     let clip_task_address = prim_instance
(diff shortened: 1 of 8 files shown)


gw3583:
(This includes the commit from https://github.com/servo/webrender/pull/3236).

Looks like there is one failure on try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=1d65d6b0005b662956871ca809ce97962081f982

I'll investigate that in the morning, but that should hopefully only be a minor fix needed.

nical:
> In future, these will include spatial
clustering, which will allow removal of much of the per-primitive
bounding rect and culling code during frame building.

Exciting!

gw3583:
I realized that some of the existing picture logic relies on the calculated picture rect being the ```unclipped``` size. So I updated the patch to not include the local clip rect in the culling rect, for now. This doesn't have any impact on performance, since it matches previous behavior. As a follow up, we can fix this logic and start to take advantage of building the local clip rect during scene building.

The try run with this fix still has a few jobs to run, but looks good so far, so this is ready for review now.
https://treeherder.mozilla.org/#/jobs?repo=try&revision=00db2fce1b954143a03d0d9b687bbf140d752563&selectedJob=208488140

emilio:
The OSX R2 crashes look related to webrender:

```
17:53:39     INFO -  3  XUL!mozilla::layers::APZSampler::~APZSampler() [Mutex.h:00db2fce1b954143a03d0d9b687bbf140d752563 : 60 + 0x5]
```

called by:

```
17:53:39     INFO -  7  XUL!webrender::render_backend::RenderBackend::update_document [render_backend.rs:00db2fce1b954143a03d0d9b687bbf140d752563 : 1039 + 0x1a]
```

May not be related to this PR though.

staktrace:
APZSampler::~APZSampler crashes might be related to https://bugzilla.mozilla.org/show_bug.cgi?id=1502495 - I've starred the failure as such.

gw3583:
Closing since this is reviewed as part of #3244.

PR #3239, opened 2018-10-29
Don't use BGRA8 unless GL_EXT_texture_storage is available

jamienicol:
https://bugzilla.mozilla.org/show_bug.cgi?id=1499785

We recently changed to use the glTexStorage* family of functions,
along with sized internal format types, such as BGRA8. Unfortunately,
the GL_EXT_texture_format_BGRA8888 extension does not provide BGRA8 as
a valid format unless the GL_EXT_texture_storage extension is also
available, which is not usually the case on GLES3 as glTexStorage*
are built in.

Unfortunately this means we must fall back to using glTexImage* rather
than glTexStorage* when dealing with BGRA data. Longer term it will
make sense to ensure the provided data is in RGBA format, as the
benifits of immutable storage are desirable.

### DIFF
--- a/webrender/src/device/gl.rs
+++ b/webrender/src/device/gl.rs
@@ -736,6 +736,16 @@ impl Drop for SharedDepthTarget {
     }
 }
 
+/// Describes the support for the glTexStorage* family of functions.
+/// Various combinations of gl/gles version and extensions can mean either
+/// not supported, supported but not for BGRA8 format, or full support.
+#[derive(PartialEq, Debug)]
+enum TexStorageSupport {
+    None,
+    NonBGRA8,
+    Full,
+}
+
 pub struct Device {
     gl: Rc<gl::Gl>,
     // device state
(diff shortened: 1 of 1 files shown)


webrender/src/device/gl.rs
```diff
@@ -736,6 +736,16 @@ impl Drop for SharedDepthTarget {
     }
 }
 
+/// Describes the support for the glTexStorage* family of functions.
+/// Various combinations of gl/gles version and extensions can mean either
+/// not supported, supported but not for BGRA8 format, or full support.
+#[derive(PartialEq, Debug)]
+enum TexStorageSupport {
+    None,
```
nox: Why is this here? You can just use `Option<TexStorageSupport>` elsewhere.

jamienicol: I can't claim to be an expert on idiomatic rust, but my logic is that the level of support for glTexStorage is a tri-state: No support, support for non-bgra, and full support. In terms of how we use the variable, there is nothing special about the None case, nor in common about the two would-be Some cases. So using an Option feels less conceptually correct, and would require messier matching code, in my opinion.

jamienicol:
Okay, the latest version restructures the logic and adds some more comments so hopefully it's a bit more obvious. And I fixed it so we do indeed use TexStorage when BGRA is not supported (ie the internal format is RGBA).

I also renamed TexStorageSupport to TexStorageUsage, because I think it makes more sense for that last case: TexStorage is still not technically supported for BGRA internal formats, but we are using it for  our BGRA textures (because they have RGBA internal formats). Then I could happily rename "None" to "Never" to avoid that issue.

kvark:
@jamienicol could you launch a try?

jamienicol:
`supports_bgra` is the condition which affects our decision. The assertions just make sure we're not in a situation we can't handle. It just so happens that they correlate (because until we support RGBA, we can't properly handle `!supports_bgra` on gles) but logically bgra support is what we're testing for.

Try run looks good: https://treeherder.mozilla.org/#/jobs?repo=try&selectedJob=209780494&revision=019bb574894c88bb74c22e17c13b3b2de231d05f

kvark:
Thanks!
@bors-servo r+

PR #3240, opened 2018-10-29
Fix srgb/linear conversion filters.

nical:
I made a *very* silly mistake in the srgb/linear conversion filters of swapping c1 and c2 (see the patch). The mistake was symetrical so it passed the tests, but the filter ended up merely quantifying the input color space into 256/13 values which isn't very useful and makes most color matrix filter applied in linear space totally wrong (as soon as the resulting colors fall out of the 0..256/13 range).

So here is the fix. I was going to introduce the `if_then_else` macro in another thing I'm working on but it's a good fit here because the argument positions are less prone to the error I made.

r? anyone

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3240)

### DIFF
--- a/webrender/res/shared.glsl
+++ b/webrender/res/shared.glsl
@@ -113,6 +113,22 @@
             return 0.0;
         return 0.5 + dist * (0.8431027 * dist * dist - 1.14453603);
     }
+
+    /// Component-wise selection.
+    ///
+    /// The idea of using this is to ensure both potential branches are executed before
+    /// selecting the result, to avoid observable timing differences based on the condition.
+    ///
+    /// Example usage: color = if_then_else(LessThanEqual(color, vec3(0.5)), vec3(0.0), vec3(1.0));
+    ///
+    /// The above example sets each component to 0.0 or 1.0 independently depending on whether
+    /// their values are below or above 0.5.
+    ///
+    /// This is written as a macro in order to work with vectors of any dimension.
+    ///
+    /// Note: Some older android devices don't support mix with bvec. If we ever run into them
+    /// the only option we have is to polyfill it with a branch per component.
+    #define if_then_else(cond, then_branch, else_branch) mix(else_branch, then_branch, cond)
 #endif
 
 //======================================================================================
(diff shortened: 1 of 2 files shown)


nical:
Some gecko reftests will cover this as soon as I [enable using SVG color matrix filters in WR](https://bugzilla.mozilla.org/show_bug.cgi?id=1468183). We can still add a reftest that tests the exact values of the srg-to-linear conversion in the webrender repo but I'm not too worried about that.

emilio:
Ok, wfm.

@bors-servo r+

webrender/res/shared.glsl
```diff
@@ -113,6 +113,22 @@
...
+    /// Example usage: color = if_then_else(LessThanEqual(color, vec3(0.5)), vec3(0.0), vec3(1.0));
+    ///
+    /// The above example sets each component to 0.0 or 1.0 independently depending on whether
+    /// their values are below or above 0.5.
+    ///
+    /// This is written as a macro in order to work with vectors of any dimension.
+    ///
+    /// Note: Some older android devices don't support mix with bvec. If we ever run into them
+    /// the only option we have is to polyfill it with a branch per component.
+    #define if_then_else(cond, then_branch, else_branch) mix(else_branch, then_branch, cond)
```
kvark: d'oh, why don't we just use `step` instead of `lessThanEqual` so that `bvec` isn't involved?
I also don't think we should be having the one-line wrappers around GLSL functions, fwiw

nical: I wasn't aware that one can use step with bvec. It's not clear from the [glsl docs](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/step.xhtml) that you could do that whereas it is explicitly specified for [mix](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/mix.xhtml).
We do want component-wise selection here so bvec is needed.

I stand by wrapping the single line function because the function will most likely be used all over the place to avoid timing attacks, using mix directly is error prone (the bug this PR is fixing) and this might need to be polyfilled if we support low end android devices.

kvark: What I'm saying is: we don't need `bvec` at all. We can just use `mix` and `step` without triggering the bugtrap GLSL booleans are.

nical: Not sure I follow, you want to do it manually for each individual channel?

kvark: I'm saying that instead of doing this:
```glsl
mix(c1, c2, lessThanEqual(color, vec3(0.04045)));
```
we could be doing this:
```glsl
mix(c1, c2, step(color, vec3(0.04045)));
```
And it would not need any wrappers or messing around `bvec`

PR #3243, opened 2018-10-29
Assert that depth targets are bound when depth features are used

bholley:
Dzmitry asked for this during review on another PR.

Differential Revision: https://phabricator.services.mozilla.com/D9954

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3243)

### DIFF
--- a/webrender/src/device/gl.rs
+++ b/webrender/src/device/gl.rs
@@ -748,6 +748,10 @@ pub struct Device {
     default_read_fbo: FBOId,
     default_draw_fbo: FBOId,
 
+    /// Track depth state for assertions. Note that the default FBO has depth,
+    /// so this defaults to true.
+    depth_available: bool,
+
     device_pixel_ratio: f32,
     upload_method: UploadMethod,
 
@@ -927,6 +931,8 @@ impl Device {
             program_mode_id: UniformLocation::INVALID,
             default_read_fbo: FBOId(0),
             default_draw_fbo: FBOId(0),
+
+            depth_available: true,
 
             max_texture_size,
             renderer_name,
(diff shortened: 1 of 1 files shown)


bholley:
> error[E0658]: use of unstable library feature 'libc': use `libc` from crates.io (see issue #27783)

bholley:
@kvark needs a clobber or something?

bholley:
Huh. I guess we're not gating on the OSX jobs then?

staktrace:
We do gate on OSX. All the tests passed on the merge commit though: https://tools.taskcluster.net/groups/elQ-bVgVQiubofBZo1sL5A

bholley:
Huh. I wonder what the difference is with the job I was looking at (one of the ones displayed at the bottom of the PR while it's pending). That one had compile bustage in a dependent crate (rand IIUC).

gw3583:
Somehow we seem to have a compile error intermittent - I've seen it on three PRs now, I think. :astonished:

gw3583:
(I wonder if there is a different rustc version on some CI machines or something like that?)

staktrace:
Looks like the failure was on the CI for the PR (as opposed to the merge commit that homu makes): https://tools.taskcluster.net/groups/cXFabj8VShOh073dXzZe5A

In this case, yes, the failure is non-blocking. FWIW the failed run happened on the wr-toronto-1 mac mini and I'm not sure offhand which version of rust is on that, if that's the problem.

staktrace:
According to https://tools.taskcluster.net/provisioners/localprovisioner/worker-types/webrender-ci-osx/workers/wr-toronto/wr-toronto-1 the worker doesn't seem to be failing everything, so there's some nondeterminism coming from somewhere.

PR #3244, opened 2018-10-30
 Store picture indices directly in primitive instances.

gw3583:
This removes the use of the BrushPrimitive for pictures, allowing
direct access to the picture index from the PrimitiveInstance.

Although there appears to be a lot of lines changed, it's mostly
just shuffling code around, there shouldn't be any functional
changes.

The motivation for this patch is to:
 - Remove use of primitive store for pictures (since these are
   being progressively removed and switched to primitive instances
   and interned primitive templates).
 - Most importantly, restructure the code such that we can add
   a primitive instance kind for primitives that only rely
   on the interned template, and not a primitive index. This will
   allow us to start interning primitives by type incrementally.

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -31,10 +31,27 @@ use renderer::{MAX_VERTEX_TEXTURE_WIDTH};
 use resource_cache::{ImageProperties, ImageRequest, ResourceCache};
 use scene::SceneProperties;
 use std::{cmp, fmt, mem, ops, usize};
+#[cfg(debug_assertions)]
+use std::sync::atomic::{AtomicUsize, Ordering};
 use util::{ScaleOffset, MatrixHelpers};
 use util::{pack_as_float, project_rect, raster_rect_to_device_pixels};
 use smallvec::SmallVec;
 
+/// Counter for unique primitive IDs for debug tracing.
+#[cfg(debug_assertions)]
+static NEXT_PRIM_ID: AtomicUsize = AtomicUsize::new(0);
+
+#[cfg(debug_assertions)]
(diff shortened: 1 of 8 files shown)


gw3583:
This one is unfortunately difficult to review because (a) it includes #3238 and #3236, so looks bigger than it actually is and (b) has a large line diff count itself, due to re-shuffling and indentation changes.

r? anyone (it might make sense to wait until #3238 and #3236 are reviewed and merged, and then I'll rebase this PR on top of those to remove those commits).

gw3583:
*Pending* try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=20c585079ec3f79ab11b4b471f6365957f9e6628

gw3583:
Some of the changes in this patch are a bit messy. It should become quite a bit tidier as we start to intern primitives, and split the prepare_prims pass completely from the picture traversal pass. But, if there's something in here that seems particularly untidy or problematic, let me know.

gw3583:
@kvark @emilio Thanks for the reviews! I think all comments are addressed now, and the try run looks good.

webrender/src/prim_store.rs
```diff
@@ -2180,7 +2180,7 @@ impl PrimitiveStore {
             //           an index buffer / set of primitive instances
             //           that we sort into render order.
             let mut in_visible_cluster = false;
-            for ci in prim_instance.cluster_range.first .. prim_instance.cluster_range.last {
+            for ci in prim_instance.cluster_range.start .. prim_instance.cluster_range.end {
```
kvark: nit: `for ci in prim_instance.cluster_range.clone()`

kvark:
Thanks, shipit!
@bors-servo r+

PR #3245, opened 2018-10-30
webrender_api: Pass &[FilterOp] instead of Vec<FilterOp>.

waywardmonkeys:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3245)

### DIFF
--- a/examples/alpha_perf.rs
+++ b/examples/alpha_perf.rs
@@ -37,7 +37,7 @@ impl Example for App {
             None,
             TransformStyle::Flat,
             MixBlendMode::Normal,
-            Vec::new(),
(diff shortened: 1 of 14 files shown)


nical:
Nice! The taskcluster failure looks unrelated I'll try to retrigger when appveyor is done.

emilio:
(There's no point in retriggering taskcluster since bors does a whole new run)

staktrace:
```
error: Connection to server timed out
error: Could not compile `osmesa-src`.
```

This definitely seems like an sccache intermittent. Somebody should tell bors to retry.

PR #3250, opened 2018-10-30
Properly evict cached blob tiles when they are out of the visible area.

nical:
This fixes [bug 1494173](https://bugzilla.mozilla.org/show_bug.cgi?id=1503220). Before this PR we would discard the rasterized blob image but not the cache entries (which didn't mesh well with the manual eviction policy, so tiles would stick around until the image key is deleted).
This aggressively removes the cached tile entry as well as the texture cache entry as soon as the tile is out of the visible area.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3250)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -1202,6 +1202,20 @@ impl ResourceCache {
         );
 
         tiles.retain(|tile, _| { tile_range.contains(tile) });
+
+        let texture_cache = &mut self.texture_cache;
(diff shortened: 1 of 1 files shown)


nical:
cc @mstange

nical:
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=971667c0b8f6720d2d5faf135a66e9234cbc3aa0

nical:
Try looks good. r? anyone who knows about resource_cache

PR #3252, opened 2018-10-30
Fix opacity of Clear brushes

kvark:
Follow-up to #3233
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1502585
Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2319bbf4ce04d1af7a785567c09d43884b8c3ca3
cc @staktrace 
r? anyone (once the try is green)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3252)

### DIFF
--- a/webrender/src/prim_store.rs
+++ b/webrender/src/prim_store.rs
@@ -3135,7 +3135,7 @@ impl PrimitiveInstance {
                         }
                         PrimitiveOpacity::from_alpha(opacity_binding.current * color.a)
                     }
-                    BrushKind::Clear => PrimitiveOpacity::opaque(),
+                    BrushKind::Clear => PrimitiveOpacity::translucent(),
                 }
             }
         };


staktrace:
r+ if try is green

mstange:
Does this make Clear brushes go into the transparent pass? If so, that would be unfortunate: whatever is behind a Clear brush will never be visible, so ideally we should avoid drawing it.

Clear is opaque in the sense that "whatever is under me will never reach the screen". It is not opaque in the sense "after I draw, alpha will be 1".

kvark:
Forgot to thank @mstange for hinting about the clear brushes, and now they are saying we need a different fix :) oh well. I believe the `translucent()` is what the old behavior was like, and it would be nice to fix the regression first before trying to optimize it. And yes, I agree that ideally we'd want to the clear brushes to be done in the opaque pass, just with the color masked out (and only alpha written). Let's take care of this as a follow-up.

@bors-servo r=staktrace

kvark:
Filed #3253 for the follow-up.

gw3583:
Once a force push occurs, I think bors might need a new r+.

@bors-servo r=staktrace

kvark:
@gw3583 the tricky thing is when a force push is done for a commit that was already approved. Bors shouldn't require another r+ there

PR #3254, opened 2018-10-31
Port line decorations to be interned primitives.

gw3583:
Line decorations make a good target for the first primitive to
port to the interning system, because (a) they don't rely
on segments, (b) they are used in text shadows, which are an
important caching opportunity (c) they are a simple primitive
to port due to lack of opacity bindings etc.

Most of this patch is setting up the general framework for
porting other primitives to be interned, and acting as a
proof of concept.

The benefits of this patch are:
 - prim_data_handle is a stable value that uniquely identifies
   the same line decoration primitive, across both frames and
   different display lists.
 - the gpu_cache_handle for the line decoration primitive
   template is also retained and valid across both frames and
   different display lists.

This gives us a very quick and efficient way to compare the
contents of a cached surface (comparing prim_data_handle plus
the state of the transforms and clips, which are also interned).

It also has the potential to significantly reduce GPU cache updates
when a new display list is processed. This currently results in
all the existing GPU cache handles being orphaned, but more and
more of these will remain valid as other primitive types are
ported to the interning system.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3254)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use api::{AlphaType, ClipMode, DeviceIntRect, DeviceIntSize, LineStyle};
+use api::{AlphaType, ClipMode, DeviceIntRect, DeviceIntSize};
 use api::{DeviceUintRect, DeviceUintPoint, ExternalImageType, FilterOp, ImageRendering};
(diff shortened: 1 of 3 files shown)


gw3583:
r? @kvark or anyone who knows this code.

Pending try (but looks good so far):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=cf245be316d677ff02e82d323a0b7a4652381609

gw3583:
Try looks good :+1:

gw3583:
@kvark Thanks, fixed up the nits and followed up on those questions :)

PR #3255, opened 2018-10-31
Properly track whether a new scene has been built to fix the debug indicator

### LINKED ISSUE #3247: The new scene indicator doesn't work
The debugging indicator behind the pref "gfx.webrender.debug.new-scene-indicator" should animate when scenes are built but doesn't move anymore.

nical:
I broke the new scene indicator when when removing redundant frame builds. Now that we don't necessarily send a new frame to the renderer when the scene is built, the information is lost so we have store it somewhere.

r? anyone

Fixes #3247.


nical:
The OSX taskcluster failure is unrelated.

PR #3257, opened 2018-11-01
Apply opacity bindings in vertex shader, rather than CPU.

gw3583:
Previously, opacity bindings were calculated on the CPU and
stored in the GPU cache as part of the primitive color. This
required the GPU cache location to be invalidated when the
opacity binding changed.

This complicates primitive interning, where we want to be able
to share the GPU cache data for interned primitives, and not
need to update it during frame building.

Instead, provide the opacity for rectangles and images as part
of the primitive instance header, and apply in the vertex
shader.

This will simplify porting rectangles and images to use
primitive interning.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3257)

### DIFF
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -146,6 +146,9 @@ void brush_vs(
 #ifdef WR_FEATURE_ALPHA_PASS
     vTileRepeat = repeat.xy;
 
+    float opacity = float(user_data.z) / 65535.0;
+    image_data.color *= opacity;
+
     switch (color_mode) {
         case COLOR_MODE_ALPHA:
         case COLOR_MODE_BITMAP:
--- a/webrender/res/brush_image.glsl
+++ b/webrender/res/brush_image.glsl
@@ -26,6 +26,10 @@ flat varying vec2 vTileRepeat;
 
 #ifdef WR_VERTEX_SHADER
 
+// Must match the AlphaType enum.
+#define BLEND_MODE_ALPHA            0
+#define BLEND_MODE_PREMUL_ALPHA     1
+
 struct ImageBrushData {
     vec4 color;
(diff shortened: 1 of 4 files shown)


gw3583:
r? @kvark or @nical 

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=52f7de460d5df426de6081ff7e0f57b41a406a7d

kvark:
Shaders aren't building any more: https://tools.taskcluster.net/groups/WwadJX38Sx6vDksws5nnPw/tasks/BnVdw5MNQqqmoYeFWJwL4Q/runs/0/logs/public%2Flogs%2Flive.log#L2145

gw3583:
Fixed the shader compile errors - needed a float cast.

gw3583:
Try run looks good so far. There is an off-by-one alpha fuzziness in two tests in R1, however the same fuzziness exists in D2D backend, according to the reftest.list

gw3583:
Good point about the premultiplied alpha blend - I'll check up on that.

gw3583:
Pushed a fix for blend modes in image shader. Good catch!

As a follow up, we should port all the batching code that uses an image shader to use a common function for getting the batch params, and add a wrench test for images with alpha blend mode (I'll open issues for these).

New pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=97ee08f71c2078113d68798aecb8697d721374e7

gw3583:
New try run looks good. Just need a review of the follow up commit that fixes the blend mode case in the image shader. The only failure is a 1/255 pixel difference in a couple of tests, but these are also fuzzy in D2D, so we can just mark them as fuzzy.

cc @staktrace

gw3583:
Rebased :+1:

PR #3259, opened 2018-11-01
Revert "Add border segments to the opaque render pass where possible."

gw3583:
This reverts commit 75ecc8181ae2435e26cd112c12426e5fcf949cfb.

This patch causes regressions in some of the talos performance tests. After initial investigation, it's not clear why this is or how to resolve them. Since it's a minor optimization, revert this patch for now, until we have more time to investigate. (see https://bugzilla.mozilla.org/show_bug.cgi?id=1500101 for details).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3259)


gw3583:
r? anyone

PR #3260, opened 2018-11-01
Port text runs to be interned primitives.

gw3583:
This makes the GPU cache for text runs stable across display
lists, which drastically reduces GPU cache updates on real
world pages when a new scene is built.

It also makes the prim_data_handle for text runs stable
across frames and display lists, making it very quick
to compare if a text run is equivalent.

Measurements of GPU cache rows (per profiler overlay):

  Wikipedia front page:  82 -> 47
  nytimes.com:           78 -> 56

I heaven't measured, but this should also reduce CPU compositor
times a bit, due to the reduced amount of updates to the GPU cache.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3260)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -543,6 +543,116 @@ impl AlphaBatchBuilder {
             .map_or(OPAQUE_TASK_ADDRESS, |id| render_tasks.get_task_address(id));
 
         match prim_instance.kind {
+            PrimitiveInstanceKind::TextRun { ref run, .. } => {
+                let subpx_dir = run.used_font.get_subpx_dir();
+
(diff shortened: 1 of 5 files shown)


gw3583:
r? @kvark 

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=fa289cf9813f624061f0a94bb08976a1dd536a0f

gw3583:
Try run looks good :tada:

gw3583:
@kvark Thanks, all comments addressed, I think.

kvark:
thank you!
@bors-servo r+

PR #3265, opened 2018-11-02
Add new notification checkpoint FrameTexturesUpdated.

aosmond:
FrameTexturesUpdated notifications should be issued when there are
either no pending texture updates, or all of the pending texture updates
have been applied. This is useful for listeners who would like to
perform an action after updating a resource. In particular this is
useful for Gecko to learn as soon as possible when it can release a
surface after it has swapped the external image ID of an image key such
that it points to a new buffer. Currently we rely upon FrameBuilt for
this, but future changes for recycling previously displayed surfaces for
animated images are more timing sensitive.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3265)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -2155,6 +2155,13 @@ impl Renderer {
                     }
                 }
                 ResultMsg::AppendNotificationRequests(mut notifications) => {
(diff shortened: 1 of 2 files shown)


aosmond:
@kvark r?

PR #3269, opened 2018-11-05
Clean up standalone texture cache eviction

bholley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1504115

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3269)

### DIFF
--- a/webrender/src/device/gl.rs
+++ b/webrender/src/device/gl.rs
@@ -26,12 +26,37 @@ use std::ptr;
 use std::rc::Rc;
 use std::slice;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT, Ordering};
 use std::thread;
 
 #[derive(Debug, Copy, Clone, PartialEq, Ord, Eq, PartialOrd)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
 pub struct FrameId(usize);
+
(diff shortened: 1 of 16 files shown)


bholley:
@staktrace FYI, there are some unexpected passes that need to be fixed when merging this:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=f918178b21ad388502acf0256d941831b187e042

staktrace:
> @staktrace FYI, there are some unexpected passes corresponding to annotations that need to be fixed when merging this:
> 
> https://treeherder.mozilla.org/#/jobs?repo=try&revision=f918178b21ad388502acf0256d941831b187e042

This try push shows intermittent failures on all platforms. On Linux `layout/reftests/bugs/836844-1.html` seems to `UNEXPECTED-PASS` but not always. On macOS, `layout/reftests/svg/filters/css-filters/blur-cap-large-radius-on-software.html` was `UNEXPECTED-PASS` on opt but not debug, and on windows `layout/reftests/bugs/501627-1.html` was also intermittent. I've retriggered some jobs to make sure, but I don't think this PR should have gotten merged without (a) a patch to actually annotate these failures and (b) ensuring that said patch produces a green try run.

bholley:
Per IRC discussion - we're going to do the wr-update, mark the handful of tests that are now intermittently passing as random-if, and I'll investigate in a followup.

staktrace:
For the record, additional try pushes didn't show the same intermittents or reftest changes that the try push earlier in this thread showed. Not sure what happened, but there appears to be no need to mark anything random-if or change any annotations.

PR #3270, opened 2018-11-05
Use picture caching for text shadows.

gw3583:
This patch allows text shadows to be cached and persisted both
between frames and new display lists, in the texture cache.

Previously, text shadows could be cached between frames, but would
be flushed and redrawn when a new display list was processed.

The primitive and clip interning system is used to determine when
the contents of a cached picture is valid.

For now, this is only applied in very limited scenarios, in order
to ensure the core functionality is correct (specifically, blur
filters that contain glyphs and line decorations, which covers
most text shadows). In the future, this will be expanded to cover
(a) more primitive types (b) more surface / filter effects and
(c) at a finer granularity to allow partial picture caching (tiles).

As part of this change, we can remove all references to the global
scene id, as this is no longer required.

This brings the GPU time in the cloud ad [1] from 43 ms down to
8 ms for me. This is still higher than we want, due to the very high
number of cached text shadows that get composited. This should be
resolved as we support nested picture caching.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1436193

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3270)


gw3583:
r? @kvark 

Initial try run looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=037fa9053b15dfb9876f55378e9cc5921d8ce56d

I also kicked off a pending try on all platforms:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=e71266f050b3a2fb150073e6f99ac2f9a739f2ed

I'm not sure how good our CI testing is for things that cached and then invalidated. It seems to work fine to me from some manual browsing, but it'd be worth some sanity checks of the cache invalidation logic to see if it seems sound to you.

gw3583:
Looks like there is a failing crash test on OSX, but I haven't been able to reproduce locally yet.

gw3583:
Both of the crash test failures on OS X are explicit panics due to render task size. However, one of the tests doesn't appear to have any text shadows at all, so it may be something unrelated to this patch. Still investigating.

gw3583:
Ah, it seems that Gecko sometimes provides an extremely large text shadow in the chrome on the first frame or two, and this patch was not catching that. Updated the patch and kicked off a new try:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=05153326984fe8dad10b663ee41829781789a1b1

Hopefully that should resolve those crash test failures.

gw3583:
Rebased. The latest try run looks good - just one new PASS in ```blur-cap-large-radius-on-software.html```

gw3583:
Rebased.

kvark:
@gw3583 sorry about the delay, reviewing it now.

gw3583:
Pending re-try https://treeherder.mozilla.org/#/jobs?repo=try&revision=90937def6efff37f31da64797fd7c777e39747f7

gw3583:
The updated try run looks good.

gw3583:
Kicked off another try run with those changes:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=ce79c55379c611432197d584d2942dbb767d6db7

gw3583:
We should be fine to land this now, since https://github.com/servo/webrender/pull/3285 is in the queue and also in gecko inbound.

@bors-servo r=kvark

PR #3271, opened 2018-11-05
Port Clear primitives to be interned.

gw3583:
This should be a straightforward conversion, since these primitives
don't use segments.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3271)


gw3583:
r? anyone

PR #3272, opened 2018-11-05
Store tile dirty rects in tile space.

nical:
Before this PR each tile stores its own dirty rect relative to the entire image. This is somewhat confusing because this field is used to track partial uploads correctly and uploads reason about dirty rects relative to the tiles so it has to do some conversions. Storing relative to the tile is also truer to the "device space" the rects are tagged with.
Last but not least the image-relative coordinates will make even less sense for tiled blob images as we start treating them as virtually infinite planes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3272)

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -196,21 +196,21 @@ pub fn intersect_for_tile(
 
 fn merge_dirty_rect(
     prev_dirty_rect: &Option<DeviceUintRect>,
-    dirty_rect: &Option<DeviceUintRect>,
-    descriptor: &ImageDescriptor,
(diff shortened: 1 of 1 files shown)


nical:
try looks good, r? anyone

Gankro:
r=me with docs updated

Gankro:
Ah, I see you delete the commented code in #3277 so should be fine..?

staktrace:
@nical This PR is not happy-making in Gecko: https://treeherder.mozilla.org/#/jobs?repo=try&group_state=expanded&revision=f39162e828cb0a91e51504ef813201573bebd918

PR #3273, opened 2018-11-05
Optimize segment builder no-ops.

gw3583:
If we don't end up adding any clip regions to the segment builder,
then there is no point in going through the process of creating
events, sorting, and emitting segments.

Instead, detect this case and early out with a single segment.

This takes the backend time on the displaylist_mutate test on
my machine from 10.8ms to 8.2ms, which is quite a significant
improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3273)

### DIFF
--- a/webrender/src/segment.rs
+++ b/webrender/src/segment.rs
@@ -177,7 +177,8 @@ pub struct SegmentBuilder {
     items: Vec<Item>,
     inner_rect: Option<LayoutRect>,
     bounding_rect: Option<LayoutRect>,
-    
+    has_interesting_clips: bool,
+
     #[cfg(debug_assertions)]
     initialized: bool,
 }
@@ -190,6 +191,7 @@ impl SegmentBuilder {
             items: Vec::with_capacity(4),
             bounding_rect: None,
             inner_rect: None,
+            has_interesting_clips: false,
             #[cfg(debug_assertions)]
             initialized: false,
         }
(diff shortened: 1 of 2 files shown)


gw3583:
r? anyone

gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a95b2cd98619df26a6c66d79e808912aa3719963

gw3583:
Try run looks good.

webrender/src/segment.rs
```diff
@@ -413,140 +423,152 @@ impl SegmentBuilder {
             None => return,
         };
 
-        // First, filter out any items that don't intersect
-        // with the visible bounding rect.
-        self.items.retain(|item| item.rect.intersects(&bounding_rect));
+        if self.has_interesting_clips {
```
nical: nit: I have a slight preference for avoiding right drift by doing:

```rust
if early_out_cond {
    // early out code
}

// most of the stuff
```

I think that avoiding lots of nested scope makes the code generally easier to read and less intimidating.

gw3583: In general I agree with you. I didn't do that in this case because there is a piece of code at the end of the method (that resets the initialization state) that much run regardless of which branch is taken. Although, having said that, I just realized that can be moved above the early out in this case.

gw3583: Pushed an update to resolve this.

PR #3275, opened 2018-11-06
Record time spent uploading in RendererStats

mattwoodrow:
For gecko bug 1503405, we want to know the amount of time we spent doing texture uploads when drawing a frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3275)


bholley:
Hm, do we actually want to include gpu cache upload time? It is technically texture upload, but it's really more about sending parameters for the shaders, which is pretty separate from the images and such in the texture cache. If we're going to measure it, maybe we should measure it separately?

nical:
+1 for measuring both separately. They are stressing the same bottleneck but they are different things to investigate.

mattwoodrow:
Yeah that's probably a good idea, updated the pull to store the two values separately.

Need to think about the best way to report these with gecko telemetry.

PR #3277, opened 2018-11-06
Introduce the DirtyRect type

nical:
This is Alexis' [DirtyRect patch](https://phabricator.services.mozilla.com/D9559) completed and rebased on top of #3272.

It turns `Option<Rect>` into an explicit type which is a lot easier to understand as `None` previously meant different things in different areas of the code.


nical:
Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e3bde69a576c86d71a88e80c8c77b4a295562642 I added some fixes to get wrench to build after the try push but the gecko visible parts are in there.

Gankro:
SGTM, but I wrote most of the second commit so I'm not sure if my vote counts 😅

(signed off on the first one in the other PR)

nical:
New try push with the latest changes https://treeherder.mozilla.org/#/jobs?repo=try&revision=f91e139cb384a0f827ce46d8a69f7f9425e466f2

kvark:
There are R5 failures in the try.

nical:
> There are R5 failures in the try.

These are fuzzable and appeared after I rebased the webrender parts so I suspect they come from another commit that doesn't have the corresponding gecko change in the push. There is also an unexpected pas that I can't take credit for.
I just rebased the gecko parts again, I'll land this if the next try push comes back green.

Issue #3248, opened 2018-10-30
Webrender CI intermittently fails


staktrace:
The macOS debug build is intermittently failing in webrender CI. All the failures appear to be on the wr-toronto-1 worker, and always fail on the `rand` crate compilation. I looked at a few good vs bad logs, and in the good cases the command that's run is this:

```
sccache rustc --crate-name rand /Users/worker/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.4.2/src/lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="libc"' --cfg 'feature="std"' -C metadata=036fd69d32b51127 -C extra-filename=-036fd69d32b51127 --out-dir /Users/worker/generic-worker/generic-worker-taskdir/task_1540896746/webrender/target/debug/deps -L dependency=/Users/worker/generic-worker/generic-worker-taskdir/task_1540896746/webrender/target/debug/deps --extern libc=/Users/worker/generic-worker/generic-worker-taskdir/task_1540896746/webrender/target/debug/deps/liblibc-3a24e28cdb54211e.rlib --cap-lints allow --deny warnings
```

and in the bad case it's this:

```
sccache rustc --crate-name rand /Users/worker/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.4.2/src/lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="libc"' --cfg 'feature="std"' -C metadata=4885fbb33c82b8cc -C extra-filename=-4885fbb33c82b8cc --out-dir /Users/worker/generic-worker/generic-worker-taskdir/task_1540884279/webrender/target/debug/deps -L dependency=/Users/worker/generic-worker/generic-worker-taskdir/task_1540884279/webrender/target/debug/deps --cap-lints allow --deny warnings
```

Notably, the bad case is missing the `--extern libc=...` bit. I don't yet know why this only happens intermittently and only on wr-toronto-1.

staktrace:
For reference you can find logs by going to https://tools.taskcluster.net/provisioners/localprovisioner/worker-types/webrender-ci-osx/workers/wr-toronto/wr-toronto-1 and clicking on the job name.

staktrace:
In the bad case cargo apparently doesn't realize libc is a dependency of rand, and tries to compile rand without compiling libc first. I strongly suspect this is an sccache bug.

[...]

PR #3278, opened 2018-11-06
Turn on sccache logging

staktrace:
This should help track down #3248, per discussion with @luser on IRC.

r? anyone

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3278)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -149,6 +149,9 @@ tasks:
             git checkout {{event.head.sha}}
             source $HOME/servotidy-venv/bin/activate
             servo-tidy
+            sccache --stop-server || true
+            mkdir -p ../artifacts
+            RUST_LOG=sccache=debug SCCACHE_ERROR_LOG=$PWD/../artifacts/sccache.log sccache --start-server
             export RUST_BACKTRACE=full
             export RUSTFLAGS='--deny warnings'
             export PKG_CONFIG_PATH="/usr/local/opt/zlib/lib/pkgconfig:$PKG_CONFIG_PATH"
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -149,13 +149,21 @@ tasks:
             git checkout {{event.head.sha}}
             source $HOME/servotidy-venv/bin/activate
             servo-tidy
+            sccache --stop-server || true
```
luser: You might want to stick a `--stop-server` invocation after your test steps as well, since sccache outputs stats after shutting down and those are useful to have in build logs.

luser:
This looks good to me.

.taskcluster.yml
```diff
@@ -149,13 +149,21 @@ tasks:
...
+            export RUST_LOG=sccache=debug
```
luser: Actually, after looking at [a build log on this PR](https://tools.taskcluster.net/groups/Ej71TayqSVufyFY8ENrgaA/tasks/Tdq8F9kqQ2yDgKbtHcuaHg/runs/0/logs/public%2Flogs%2Flive.log), you probably don't want to export this, just put it in the `--start-server` line. Otherwise you get log spew every time cargo invokes sccache.

luser:
[Here's sccache.log](https://taskcluster-artifacts.net/Tdq8F9kqQ2yDgKbtHcuaHg/0/public/sccache.log) from the OS X debug tests task.

staktrace:
Updated patch to address both comments

staktrace:
@bors-servo r=luser

(Having only one mac CI machine is really slowing things down, and apparently even that one is showing some errors, so getting this in sooner will help us debug what's going on)

PR #3281, opened 2018-11-06
Change add_split_plane to not rely on prim instance.

gw3583:
This will make it easier to move the plane splitting code to the
initial picture traversal pass, instead of during primitive
preparation pass.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3281)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -483,7 +483,7 @@ pub struct PicturePrimitive {
 
     /// The spatial node index of this picture when it is
     /// composited into the parent picture.
-    spatial_node_index: SpatialNodeIndex,
+    pub spatial_node_index: SpatialNodeIndex,
 
     /// The local rect of this picture. It is built
     /// dynamically during the first picture traversal.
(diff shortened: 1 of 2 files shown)


gw3583:
r? @kvark

I _think_ that these changes are safe to make, but you know the details of this code much better, so I wanted to make a small PR for this to ensure it doesn't regress anything. Specifically:

 * Don't include local clip rect in the polygon splitting. This seems more logical anyway, and any surface allocations get clipped to the screen size during picture preparation. Equally, when we start to cache these, we'll want the option to draw them without this clip rect applied anyway.
 * For the optional bounds, just supply the world rect of the scene instead of the clipped primitive world rect. This in theory is slightly less accurate, but in practice it probably doesn't matter? Do you know of any test cases / situations where this would cause a problem?

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=27991a703827b7cc93bc2a7cf53390837545132c

kvark:
The code looks good, but I'm confused about this comment:
> Don't include local clip rect in the polygon splitting

The local rect is still defining the polygon bounds in the new code.

gw3583:
Yep, the comment just means that the local _clip_ rect from the polygon isn't applied at splitting time, but at surface allocation time. Thanks!

@bors-servo r=kvark

PR #3284, opened 2018-11-07
Revert "Store tile dirty rects in tile space."

nical:
Reverts #3272. It had a green try but the rebase apparently didn't go well as the webrender update is on fire.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3284)


staktrace:
r+, thanks

PR #3285, opened 2018-11-08
Reduce shared texture cache to 32 layers.

bholley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1505664

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3285)


bholley:
Note that I landed this directly to inbound, and told the sheriffs to be sure it's in the next nightly.

PR #3286, opened 2018-11-08
Extract a method to return the dumped display list as a string.

staktrace:
For proper android support we need to dump this to logcat instead of
stderr, so it seems better to just have a function to return the dumped
string and we can send it wherever we want in the embedding application.
This also changes the print_display_list back to the original semantics
(prior to servo/webrender#2893) where it just dumps the whole display
list to stdout.

Reviewed by @kvark at https://phabricator.services.mozilla.com/D11195
Related bug in bugzilla: https://bugzil.la/1505375

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3286)


staktrace:
@bors-servo retry

sccache flakiness.

PR #3288, opened 2018-11-08
Remove PictureId and PictureIdGenerator.

gw3583:
They are no longer required since we are comparing the content
of pictures when choosing to cache them.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3288)


gw3583:
r? anyone

kvark:
D'oh, it's the second time we are removing those! I barely got used to having them :)

PR #3289, opened 2018-11-09
Separate brush segment descriptors from clip mask instances.

gw3583:
As part of interning primitives, we can move the segmenting
process to occur when a primitive is newly interned (which is
a significant optimization opportunity). We can do this since
segmentation only occurs on clip nodes with the same spatial
node as the primitive, thus the segmentation cannot be changed
due to scrolling or transform animation.

However, the presence of a clip mask per segment may still
change each frame, due to clips from other positioning nodes.

To handle this, we need to split the per-frame clip mask instance
information from the brush segment descriptors.

This is also a necessary step to allow borders to be interned,
which rely on pre-generated segment descriptors.

At the same time, make the size of a RenderTaskId smaller in
release builds, and tidy up FrameId::invalid().

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3289)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -15,10 +15,10 @@ use gpu_types::{PrimitiveInstanceData, RasterizationSpace, GlyphInstance};
 use gpu_types::{PrimitiveHeader, PrimitiveHeaderIndex, TransformPaletteId, TransformPalette};
 use internal_types::{FastHashMap, SavedTargetIndex, TextureSource};
 use picture::{Picture3DContext, PictureCompositeMode, PicturePrimitive, PictureSurface};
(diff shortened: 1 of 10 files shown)


gw3583:
This also seems to give a small but noticeable performance win on some of the talos tests.

Try run is not complete but looks good so far:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4e7481e1ced995bf0119a53472481a487ab28044

gw3583:
Mentioned in the comments but repeating here too - the work here to maintain per-instance data between ```prepare_prims``` and batching should be temporary. In the near-ish future we should be able to unify those passes, and avoid storing any of this data.

gw3583:
Try run looks good.

gw3583:
@kvark Those comments all made sense and have been fixed. Also rebased.

PR #3291, opened 2018-11-09
Use DeviceInt* instead of DeviceUint*.

nical:
We currently use a mix of both which can be untidy at places and is error prone when doing intermediate computations that may go through negative values.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3291)

### DIFF
--- a/direct-composition/src/main_windows.rs
+++ b/direct-composition/src/main_windows.rs
@@ -34,7 +34,7 @@ fn main() {
 
     let mut clicks: usize = 0;
     let mut offset_y = 100.;
-    let size = api::DeviceUintSize::new;
+    let size = api::DeviceIntSize::new;
     let mut rects = [
         Rectangle::new(&composition, &notifier, factor, size(300, 200), 0., 0.2, 0.4, 1.),
         Rectangle::new(&composition, &notifier, factor, size(400, 300), 0., 0.5, 0., 0.5),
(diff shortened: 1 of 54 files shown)


nical:
Sorry for the antisocial patch. The change is completely mechanical but it ended touching quite a bit of code and will likely conflict with everything else.

As discussed in the gfx daily, the idea is to use signed integers everywhere instead of a mix of signed and unsigned integers for device pixels. This removes an old TODO in the code that was already complaining about the mess of mixing all kinds of scalar types. This will also help avoid the kind of mistakes that got #3272 backed out (intermediate computations potentially using negative values). For reference, gecko uses signed integers for all of its integer point/rect/size types.

I'm working on the corresponding gecko patch since this affects the API.

nical:
Green build-only try push: https://treeherder.mozilla.org/#/jobs?repo=try&selectedJob=210593072&revision=d142cdb56ff05159e42748b10a163204ec141a76

There was a prior push that passed the tests on linux but didn't build everywhere before I did the build fixes: https://treeherder.mozilla.org/#/jobs?repo=try&selectedJob=210593072&revision=5e79214b6ec27704581198ce0b7d7b2b2feb813e

Unfortunately this PR bitrots faster than I can test it on the CI, but since it is only changing tons of `u32`s into `i32`s I'm confident that it's in a good shape now.

staktrace:
@nical can you put the Gecko side patch (properly formatted please) somewhere that I can pull?

nical:
The Gecko patch is up at https://bugzilla.mozilla.org/show_bug.cgi?id=1507140

Issue #3198, opened 2018-10-15
All examples have no content


gnzlbg:
I've tried running a couple of examples (`image_resiz`, `basic`, `animation`, ...) on MacOS 10.14 and the CLI output is:

```
OpenGL version 4.1 INTEL-12.0.34
Shader resource path: None
Device pixel ratio: 1
Loading shaders...
Entering event loop
```

A window does open, but it has no content in any of the examples.

kvark:
This is a lower-level issue, see https://github.com/tomaka/glutin/issues/1069
We can keep this open until it's fixed upstream and we update the dependencies.

atouchet:
glutin 0.19 is now out which should include a fix for this.

alexanderuv:
@kvark  looks like this has been fixed upstream. Do you know when dependencies will be updated for Servo?

gw3583:
cc @jdm

jdm:
@alexanderuv Servo is updated as needed. You're welcome to submit a PR.

atouchet:
glutin has finally been updated in WebRender. Is this fixed now?

MarcoPolo:
Still looks broken.
<img width="1002" alt="Screen Shot 2019-09-25 at 9 01 42 PM" src="https://user-images.githubusercontent.com/594035/65657344-dcfc5d80-dfd7-11e9-9efc-2a9b418380d6.png">

macos 10.14.6

PR #3295, opened 2018-11-10
Update to glutin 0.19 and winit 0.18 issue #3198 macOS 10.14 fix

martinenzinger:
fixes the empty window issue on macOS 10.14 

https://github.com/tomaka/winit/pull/685

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3295)


atouchet:
Doesn't cargo.lock still need to be updated?

martinenzinger:
@atouchet `Cargo.lock` should not be manually changed. Here is a link explaining what the `.lock` files are meant to do: [Cargo.toml vs Cargo.lock](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html).
If the examples still don't work on Mojave, you can try cloning [martinenzinger/webrender](https://github.com/martinenzinger/webrender) into a new folder.

martinenzinger:
@gw3583  In the v0.19 release of glutin the `GlContext` trait is different. I guess this would require changes I, having seen the code for the first time a week ago, can not carry out. What do the other people think about the issue being examples are not working and a different version of glutin for the wrench debugging as a temporary solution?

On the other hand, the pull request is linked in the issue, and in the meantime, Mojave users can find an easy fix when they search for issues.

https://docs.rs/glutin/0.19.0/glutin/trait.GlContext.html
https://github.com/tomaka/glutin/releases/tag/v0.19.0

kvark:
@martinenzinger if Cargo.lock is in Git, we need to update it here. If you want to raise the discussion of *why* we keep it in Git, that would be a different issue.

> Merge remote-tracking branch 'upstream/master' 

Also, please rebase instead of merging from upstream.

PR #3297, opened 2018-11-11
Revert "Change add_split_plane to not rely on prim instance."

gw3583:
This reverts commit cbad1c4ee3143315baa5bfefa68695e561ea9c7f.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3297)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -396,7 +396,7 @@ pub struct PicturePrimitive {
 
     /// The spatial node index of this picture when it is
     /// composited into the parent picture.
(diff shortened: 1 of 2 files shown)


gw3583:
This is causing a regression on nightly (see https://bugzilla.mozilla.org/show_bug.cgi?id=1506327 for details). I can see why it's occurring, but the proper fix is slightly involved, so for now we will just revert this and follow up with a working patch later.

r? @kvark or anyone

kvark:
@gw3583 

> I can see why it's occurring, but the proper fix is slightly involved

Can you provide more details please?

PR #3300, opened 2018-11-12
Extract nine patch segment generation into create_nine_patch_segments.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3300)


gw3583:
The patches I'm working on towards border interning are getting quite large, so I'm trying to split it up into smaller chunks. This patch is a straightforward move of code from flattening into a reusable function. We will need this later since the segment generating when interning only happens when a unique primitive is created, not every time a new scene is built.

r? @kvark or anyone

gw3583:
Added a follow up comment that reduces the size of PrimitiveInstance, and ensures that interned primitives don't use the gpu cache handle that (was) in the instance.

Pending try with both patches:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=0348b4407434029ec50b94b3ab0e0babc44573dd

gw3583:
Try run looks good.

PR #3304, opened 2018-11-13
Use trace-level debugging for sccache

staktrace:


staktrace:
Should help with debugging https://github.com/mozilla/sccache/issues/328

Issue #2638, opened 2018-04-09
Mac CI is broken


jrmuizel:
I seem to have to broken it to cause osmesa not to build:

```
make[4]: /Users/tcworker/.cargo/git/checkouts/osmesa-src-bf84ce2b80210dd7/1cf8c2f/mesa-src/bin/install-sh: Permission denied
make[4]: /Users/tcworker/.cargo/git/checkouts/osmesa-src-bf84ce2b80210dd7/1cf8c2f/mesa-src/bin/install-sh: Permission denied
make[4]: *** [os/.deps/.dirstamp] Error 1
make[4]: *** Waiting for unfinished jobs....
make[4]: *** [pipebuffer/.deps/.dirstamp] Error 1
make[3]: *** [all-recursive] Error 1
make[2]: *** [all-recursive] Error 1
make[1]: *** [all] Error 2
make: *** [all-recursive] Error 1
thread 'main' panicked at 'failed with: exit code: 2', /Users/tcworker/.cargo/git/checkouts/osmesa-src-bf84ce2b80210dd7/1cf8c2f/build.rs:58:9
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:68
             at src/libstd/sys_common/backtrace.rs:57
             at src/libstd/panicking.rs:381
   2: std::panicking::default_hook
             at src/libstd/panicking.rs:397
   3: std::panicking::begin_panic
             at src/libstd/panicking.rs:577
   4: std::panicking::begin_panic
             at src/libstd/panicking.rs:538
   5: std::panicking::try::do_call
             at src/libstd/panicking.rs:522
   6: build_script_build::run
             at ./build.rs:58
   7: build_script_build::main
             at ./build.rs:45
   8: std::rt::lang_start::{{closure}}
             at /Users/travis/build/rust-lang/rust/src/libstd/rt.rs:74
   9: std::panicking::try::do_call
             at src/libstd/rt.rs:59
             at src/libstd/panicking.rs:480
  10: panic_unwind::dwarf::eh::read_encoded_pointer
             at src/libpanic_unwind/lib.rs:101
  11: std::sys_common::bytestring::debug_fmt_bytestring
             at src/libstd/panicking.rs:459
             at src/libstd/panic.rs:365
[...]

PR #3308, opened 2018-11-13
Switch to upstream servo/osmesa-src.

emilio:
Make the mac hack for #2638 specific for CI, via `CARGO_MAKEFLAGS`.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -139,6 +139,8 @@ tasks:
           - master
     payload:
       maxRunTime: 3600
+      env:
+        CARGO_MAKEFLAGS: '-j1' # See #2638
       command:
         - - /bin/bash
           - '--login'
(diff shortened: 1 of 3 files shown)


jrmuizel:
This has the same Permission denied problem
```
make[4]: /Users/tcworker/.cargo/git/checkouts/osmesa-src-bf84ce2b80210dd7/e17b23b/mesa-src/bin/install-sh: Permission denied
```

gw3583:
Could it be something like the +x file permission bit is somehow not set on that script?

emilio:
It'd be really weird that that only happens when building in parallel mode.

Looks like OSX make doesn't really respect MAKEFLAGS, but I'll find a different solution.

@bors-servo r-

emilio:
Let's see how that goes.

emilio:
https://tools.taskcluster.net/groups/RlTzFrmjSfyDJykTgmkftw is green on OSX.

emilio:
(I just squashed the commits in the run above).

r? @gw3583 or @jrmuizel

PR #3309, opened 2018-11-14
[WIP] Add spatial clustering to pictures.

gw3583:
This improves performance by reducing the amount of per-primitive
work that is required each frame. It also groups primitives that
are near each other, which will be helpful for efficiently
determining which parts of a cached picture are valid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,6 +80,11 @@ version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "bit-vec"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bitflags"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


gw3583:
This is a [WIP] for now, because:
 - There are a couple of test failures on try.
 - It (very slightly) regresses performance on dl_mutate on talos, which is far from ideal!

I'll investigate these tomorrow, but posting for now in case anyone has comments on the general approach.

gw3583:
@pcwalton Not a big deal, but seeing this on CI:

```
./Cargo.lock:1: duplicate versions for package `bit-vec`
	The following packages depend on version 0.4.4 from 'crates.io':
		pathfinder_partitioner
	The following packages depend on version 0.5.0 from 'crates.io':
		webrender
```

Would it be possible to get pathfinder_partitioner updated to use the newer bit-vec and published?

pcwalton:
@gw3583 On it.

pcwalton:
@gw3583 https://github.com/pcwalton/pathfinder/commit/5db0ecf8b051700200ff0d5d5590798c62db53e3

PR #3311, opened 2018-11-14
[WIP] Remove explicit clip ID from stacking context

kvark:
This is one of the preparatory changes towards #3251

With this in, we should be able to remove this (hacky) code path in Gecko:
https://searchfox.org/mozilla-central/rev/d850d799a0009f851b5535580e0a8b4bb2c591d7/gfx/layers/wr/ClipManager.cpp#224

Currently, we provide a clip ID explicitly when pushing a stacking context. This change makes us use the clip ID on the clip/scroll stack instead, like with every other item. The implementation comes with a few quirks, namely the introduction of "push_clip_id_override" and "push_scroll_id_override" into the API. These aren't conflicting with #3251 and will be removed by it (together with the whole clip/scroll stack).

WIP because: still working on the Gecko side

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3311)

### DIFF
--- a/examples/alpha_perf.rs
+++ b/examples/alpha_perf.rs
@@ -34,7 +34,6 @@ impl Example for App {
 
         builder.push_stacking_context(
             &info,
-            None,
             TransformStyle::Flat,
             MixBlendMode::Normal,
             &[],
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -59,7 +59,6 @@ impl Example for App {
         let info = LayoutPrimitiveInfo::new(bounds);
         builder.push_stacking_context(
             &info,
-            None,
             TransformStyle::Flat,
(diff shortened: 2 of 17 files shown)


kvark:
Thinking about this more and looking at the failing reftests, this isn't going to work. With the current implementation we need a separate clip ID inside the SC: the client code typically pushes a reference frame on the C/S stack, then pushes a stacking context, and it expects all the inner items *to still be clipped* by the clip associated with the scroll ID of the reference frame (yeah... see https://github.com/servo/webrender/pull/3251#issuecomment-439063422). So we currently have separate clipping semantics for SC inner/outer context, and we shouldn't be hacking the C/S stack semantics to avoid that separate field (instead, we should just kill C/S stack entirely, which is done in #3251).

PR #3313, opened 2018-11-15
Reduce need for full world rect calculation for primitives.

gw3583:
The fewer places that rely on a full world rect, the easier it
is to do clustered culling, where we accept / reject primitives
in groups.

The main use for world rects of primitivs is overlap calculations
during batching. Instead, switch to use the surface relative rect
for batching. This is safe since we know that when we merge batches
from different surfaces, they will never be overlapping in the
allocated render target.

Other changes:
 * Refactor the initial picture traversal to use a state object
   that maintains an internal stack of picture / surface info.
   This is easier to reason about, and will be helpful once we
   start using this to pass information about caching state.
 * Use world rect rather than clipped prim world rect for bounds
   during plane-splitting. This was landed previously but
   backed out due to an unrelated bug in that patch.
 * Change get_raster_rects to not calculate the transform, since
   most uses of this method don't require it.
 * Change conservative tiling calculations to use the world rect
   for bounds instead of clipped prim rect. All we're trying to
   do here is reject tiles outside the viewport, so this simplifies
   the code and removes the need for the primitive world rect in
   one more location.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3313)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -4,7 +4,7 @@
 
 use api::{AlphaType, ClipMode, DeviceIntRect, DeviceIntSize};
 use api::{DeviceUintRect, DeviceUintPoint, ExternalImageType, FilterOp, ImageRendering};
-use api::{YuvColorSpace, YuvFormat, WorldRect, ColorDepth};
+use api::{YuvColorSpace, YuvFormat, ColorDepth, PictureRect};
 use clip::{ClipDataStore, ClipNodeFlags, ClipNodeRange, ClipItem, ClipStore};
 use clip_scroll_tree::{ClipScrollTree, ROOT_SPATIAL_NODE_INDEX, SpatialNodeIndex};
 use glyph_rasterizer::GlyphFormat;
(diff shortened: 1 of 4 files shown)


kvark:
In try push:
  - wpt6 isn't currently recognized as an intermittent, although it is likely
  - wr1's UNEXPECTED_FAIL also isn't marked as an intermittent, but this one doesn't look entirely unrelated

As for the crash, the "known" one, as far as I'm aware is the failed "invalid parent!" expectation. Your case is slightly different, not to mention that I didn't expect the old one to show up on CI (seemed deterministic, cc @aosmond ).

gw3583:
Thanks for the double check!

The wpt6 run definitely looks unrelated.

The Wr1 fail is one I see regularly, and it looks like there are intermittent bugs filed for it.

The crash that I'm talking about is https://bugzilla.mozilla.org/show_bug.cgi?id=1502717 - it's an intermittent array out of bounds error, which is the same as occurred here (and only occurred in one of the test configs).

PR #3315, opened 2018-11-15
Don't ignore the clip on the stack for reference frames and clips

kvark:
Similar to #3311, this is a prelude to #3251, with the main purpose of getting rid of Gecko's hack in https://searchfox.org/mozilla-central/rev/d850d799a0009f851b5535580e0a8b4bb2c591d7/gfx/layers/wr/ClipManager.cpp#224

It also makes sense for WR in isolation (for the current clip/stack model, at least): in those spots addressed, we used to ignore the clip ID on the stack, only to re-interpret (smuggle!) the scroll ID as a clip. Now we are actually using the clip ID more consistently.

Gecko try (that disables the hack in addition to having the WR change):
https://treeherder.mozilla.org/#/jobs?repo=try&selectedJob=212038686&revision=c10e5d839e024ba451a6622fa611c3525457f3ff
(Looks green! :tada: )

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -377,7 +377,7 @@ impl<'a> DisplayListFlattener<'a> {
 
         debug_assert!(info.clip_id != info.scroll_frame_id);
 
-        self.add_clip_node(info.clip_id, clip_and_scroll_ids.scroll_node_id, clip_region);
+        self.add_clip_node(info.clip_id, clip_and_scroll_ids, clip_region);
 
         self.add_scroll_frame(
             info.scroll_frame_id,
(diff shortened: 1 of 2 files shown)


gw3583:
I _think_ this makes sense. Does this have any impact on the number of clip chains that will end up being assigned to stacking context (pictures). (i.e. is the extra intermediate surface generation at [1] going to be triggered more with this change?).

[1]

https://github.com/servo/webrender/blob/9ec08a4cf0ce6762a98ddd4df34611dbf0694703/webrender/src/display_list_flattener.rs#L1046

kvark:
The `clipping_node` argument in `push_stacking_context` is unchanged, so there should be no extra intermediate surfaces. Speaking of which, perhaps we should annotate more reftests with the expected number of passes/surfaces in order to prevent this type of regressions.

kvark:
Added a few checks to the reftests in the new commit. Confirmed locally that they pass with and without the new code.
Thanks for the review!
@bors-servo r=gw3583

PR #3316, opened 2018-11-15
Ensure coord system ID is propagated for non-invertible nodes.

gw3583:
An invariant required by the clip-scroll tree is that any child
spatial node must have a coordinate system ID >= of the parent
node's coordinate system ID.

This patch changes the update_transform traversal to ensure that
even when a non-invertible transform is detected, the coordinate
system ID from the parent is still propagated.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1502717.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3316)

### DIFF
--- a/webrender/src/spatial_node.rs
+++ b/webrender/src/spatial_node.rs
@@ -273,36 +273,38 @@ impl SpatialNode {
                 self.world_content_transform = self.world_viewport_transform;
 
                 info.invertible = self.world_viewport_transform.is_invertible();
-                if !info.invertible {
-                    return;
+
+                if info.invertible {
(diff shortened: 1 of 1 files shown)


gw3583:
r? @kvark 

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=69fe7dcace0390952ad07adcbf695ebd029c8618

gw3583:
Try looks good.

kvark:
Looks straightforward. Thank you for fixing that nasty issue!
@bors-servo r+

PR #3317, opened 2018-11-16
don't disable subpixel positioning for a text run when it is requested

lsalzman:
This resolves Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1506053

Gecko is generating metrics with DWrite assuming subpixel positioning. So if it requests WR use subpixel position, we have to honor it, otherwise the rendered glyphs will not match up. It's not really tenable to just disable subpixel positioning on the metrics side, since that requires using entirely different render modes (GDI vs. ClearType) which may significantly impact appearance.

The line that disabled subpixel positioning did not even previously exist till I accidentally added it in https://github.com/servo/webrender/commit/58ccf522b660c01ee0ae8fe6900491f8d088e4e5

This just undoes that mistake.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3317)


gw3583:
```
REFTEST INFO | 238 passing, 1 failing
	reftests\text\large-glyphs.yaml != reftests\text\blank.yaml
WARN: rx::HLSLCompiler::compileToBinary(228): 
```

gw3583:
@lsalzman Looks like a failing reftest on CI.

Issue #3324, opened 2018-11-19
Add scratch data storage for prepare_prim to PrimitiveStore


djg:
Talking with @gw3583 on IRC:

> So the data in the prim_store is mutated during prepare_prims and then read/used during batching when the glyphs are rasterized and allocated in the texture cache. Instance data like that can be thought of as "scratch data" this prim instance needs to request during prepare_prims and then use during batching.
> So, another way to think about that scratch data, is:
> We could probably just have a few fixed arrays in prim_store, e.g. one for all glyph_keys, and each text run instance just writes to that array, and stores in itself a (index, count) of what it wrote in that buffer, and then reads them during batching.
> So, for instance, we might end up with:
```rust
struct PrimStore {
   glyph_key_buffer: Vec<GlyphKey>,
}
```
> and each text run instance would have:
```rust
glyph_keys: (GlyphKeyIndex, u32)
```
> with:
```rust
pub struct GlyphKeyIndex(u32)
```
> For a bit of extra type safety

PR #3325, opened 2018-11-19
Glyph key scratch store

djg:
Implement scratch storage of GlyphKeys for #3324

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3325)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -595,7 +595,7 @@ impl AlphaBatchBuilder {
                 );
             }
             PrimitiveInstanceKind::TextRun { run_index, .. } => {
-                let run = &ctx.prim_store.text_runs[run_index.0];
+                let run = &ctx.prim_store.text_runs[run_index];
                 let subpx_dir = run.used_font.get_subpx_dir();
 
                 // The GPU cache data is stored in the template and reused across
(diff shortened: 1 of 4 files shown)


djg:
Try run result for first version: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2a1353c60b3cedfea396330beab65b825170b821&selectedJob=212544428

gw3583:
It'd be interesting to get some separate profiles of the change to use a scratch buffer vs. using the chunked storage. It may well be that a normal ```Vec``` with a reasonable ```reserve``` estimate up front is actually faster than the chunked storage?

Running ```benchmarks/text-rendering.yaml``` is a reasonable stress test of drawing a lot of text that is easy to profile within wrench. You'd need to run with ```-r``` though to replay the display list each time you hit right-arrow.

kvark:
@djg try looks good, and my concerns are addressed (for the most part). Could you check that benchmark mentioned by @gw3583 before we go?

djg:
I'll run the benchmark that @gw3583 mentioned once I get my current interning work to a good point.

@gw3583, I was thinking that sizing the initial storage to ~64KB

gw3583:
(Discussed on IRC - I was initially misreading how the storage type works. This should be a simple win, and we'll extend it with recycle functionality as a follow up).

PR #3326, opened 2018-11-19
Use euclid 0.19.3+ exported from webrender_api.

nical:
Extracted this from #3277 to make the diff less scary.

Use euclid 0.19.3 and rather than manage the same dependency in both webrender and webrender_api, just publicly reexport in webrender_api.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3326)

### DIFF
--- a/webrender/src/lib.rs
+++ b/webrender/src/lib.rs
@@ -159,7 +159,6 @@ extern crate dwrote;
 extern crate app_units;
 extern crate bincode;
 extern crate byteorder;
-extern crate euclid;
 extern crate fxhash;
 extern crate gleam;
 extern crate num_traits;
@@ -202,4 +201,5 @@ pub use renderer::{RendererStats, SceneBuilderHooks, ThreadListener, ShaderPreca
 pub use renderer::MAX_VERTEX_TEXTURE_WIDTH;
 pub use shade::{Shaders, WrShaders};
 pub use webrender_api as api;
+pub use webrender_api::euclid;
 pub use resource_cache::intersect_for_tile;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,7 +307,7 @@ dependencies = [
 name = "direct-composition"
 version = "0.1.0"
 dependencies = [
- "euclid 0.19.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "euclid 0.19.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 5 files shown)


nical:
r? anyone

webrender/src/lib.rs
```diff
@@ -202,4 +201,5 @@ pub use renderer::{RendererStats, SceneBuilderHooks, ThreadListener, ShaderPreca
 pub use renderer::MAX_VERTEX_TEXTURE_WIDTH;
 pub use shade::{Shaders, WrShaders};
 pub use webrender_api as api;
-pub use resource_cache::intersect_for_tile;
+pub use webrender_api::euclid;
+pub use resource_cache::intersect_for_tile;
```
kvark: nit: newline

webrender/src/lib.rs
```diff
@@ -202,4 +201,5 @@ pub use renderer::{RendererStats, SceneBuilderHooks, ThreadListener, ShaderPreca
 pub use renderer::MAX_VERTEX_TEXTURE_WIDTH;
 pub use shade::{Shaders, WrShaders};
 pub use webrender_api as api;
-pub use resource_cache::intersect_for_tile;
+pub use webrender_api::euclid;
```
kvark: is this meant to be `pub`?

nical: yeah, I can't think of any reason to not make it public since users of webrender need to keep in sync with webrender's version of euclid. This way it's easier to just use `webrender::euclid` or `webrender_api::euclid` (depending on the situation) and not have to worry about having the right version the Cargo.toml.

In my opinion we should do this with all of the dependencies that are both in webrender and webrender_api, it's simpler to manage.

kvark: I don't think it's possible to work with `webrender` but not link to `webrender_api`. Since the latter already exposes `euclid`, it would be more strict to just keep it the only crate that re-exports dependencies.

nical: I'm not sure what would be the advantage from being stricter here. For all consumers of WebRender except servo, the `webrender` crate is the only dependencies and the api is treated as module of it. It's rather common to expose public dependencies at the root of the crate. I wouldn't lose sleep over having to look for euclid and other dependencies in `webrender::api` rather than the root if you feel strongly about it but I'm curious about what the pros of that would be.

kvark: ah, right, they access it through `webrender::api`. Ok, that's fine then

PR #3327, opened 2018-11-19
Use an intermediate renderbuffer when blitting between texture arrays…

jamienicol:
… on adreno devices

When the texture cache is grown we allocate a larger texture array
then copy each layer of the old array in to the new one. To perform
this copy we use use glBlitFramebuffer, with the read and draw
framebuffers bound to the correct layer in the old and new textures.

Unfortunately on Adreno devices this does not work. glBlitFramebuffer
always writes to the 0th layer of the texture bound to the draw
framebuffer. glCopyTexSubImage3D also does not work correctly - it
always reads from the 0th layer of the texture bound to the read
framebuffer.

As a workaround, we use glBlitFramebuffer to blit from the old texture
in to a temporary renderbuffer, then use glCopyTexSubImage3D to copy
from that renderbuffer to the new texture.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3327)

### DIFF
--- a/webrender/src/device/gl.rs
+++ b/webrender/src/device/gl.rs
@@ -857,6 +857,11 @@ pub struct Device {
     /// format, we fall back to glTexImage*.
     texture_storage_usage: TexStorageUsage,
 
+    // Whether we are able to use glBlitFramebuffers with the draw fbo
+    // bound to a non-0th layer of a texture array. This is buggy on
+    // adreno devices.
+    can_blit_to_texture_array: bool,
+
     // GL extensions
     extensions: Vec<String>,
 }
@@ -1024,6 +1029,10 @@ impl Device {
             )
         };
 
+        // Due to a bug on adreno devices, blitting to an fbo bound to
+        // a non-0th layer of a texture array is not supported.
+        let can_blit_to_texture_array = !renderer_name.starts_with("Adreno");
+
         // Explicitly set some global states to the values we expect.
         gl.disable(gl::FRAMEBUFFER_SRGB);
         gl.disable(gl::MULTISAMPLE);
@@ -1066,6 +1075,7 @@ impl Device {
             frame_id: GpuFrameId(0),
             extensions,
             texture_storage_usage,
+            can_blit_to_texture_array,
         }
     }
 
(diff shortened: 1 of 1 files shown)


leeoniya:
good candidate for https://github.com/servo/webrender/wiki/Driver-issues ?

gw3583:
@leeoniya Yep, definitely.

gw3583:
Compile errors on CI (I didn't check the platforms / features causing it):

```
error[E0609]: no field `width` on type `&device::gl::Texture`
    --> webrender\src\device\gl.rs:1641:46
     |
1641 |                                          src.width as _, src.height as _);
     |                                              ^^^^^
error[E0609]: no field `height` on type `&device::gl::Texture`
    --> webrender\src\device\gl.rs:1641:62
     |
1641 |                                          src.width as _, src.height as _);
     |                                                              ^^^^^^
error[E0609]: no field `width` on type `&device::gl::Texture`
    --> webrender\src\device\gl.rs:1668:57
     |
1668 |                                               0, 0, src.width as _, src.height as _);
     |                                                         ^^^^^
error[E0609]: no field `height` on type `&device::gl::Texture`
    --> webrender\src\device\gl.rs:1668:73
     |
1668 |                                               0, 0, src.width as _, src.height as _);
     |                                                                         ^^^^^^
error[E0609]: no field `0` on type `&_`
    --> webrender\src\device\gl.rs:1673:48
     |
1673 |             self.gl.delete_renderbuffers(&[rbo.0]);
     |                                                ^
error: aborting due to 5 previous errors
For more information about this error, try `rustc --explain E0609`.
Caused by:

```

jamienicol:
Texture.width/height got changed to Texture.size recently, but that hadn't made it to gecko yet.

jamienicol:
I have updated the wiki page to include this bug

webrender/src/device/gl.rs
```diff
@@ -1621,12 +1631,47 @@ impl Device {
...
         debug_assert!(dst.layer_count >= src.layer_count);
 
-        // Note that zip() truncates to the shorter of the two iterators.
+        // If we are unable to blit directly between the texture arrays
+        // then we must create an intermediate renderbuffer.
+        let temp_buffer = if src.layer_count > 1 && !self.can_blit_to_texture_array {
+            let rbo = RBOId(self.gl.gen_renderbuffers(1)[0]);
+            let fbo = FBOId(self.gl.gen_framebuffers(1)[0]);
+            self.gl.bind_renderbuffer(gl::RENDERBUFFER, rbo.0);
+            self.gl.renderbuffer_storage(gl::RENDERBUFFER, gl::RGBA8,
```
kvark: how are we sure `gl::RGBA8` is compatible with whatever texture we are blitting?

webrender/src/device/gl.rs
```diff
@@ -1621,12 +1631,47 @@ impl Device {
...
+            None
+        };
+
         let rect = DeviceIntRect::new(DeviceIntPoint::zero(), src.get_dimensions().to_i32());
-        for (read_fbo, draw_fbo) in src.fbos.iter().zip(&dst.fbos) {
+        for (i, read_fbo) in src.fbos.iter().enumerate() {
+            // Blit from the src to the dst (or the intermediate buffer if necessary)
             self.bind_read_target_impl(*read_fbo);
-            self.bind_draw_target_impl(*draw_fbo);
+            self.bind_draw_target_impl(match &temp_buffer {
```
kvark: nit: remove `&`

webrender/src/device/gl.rs
```diff
@@ -1621,12 +1631,47 @@ impl Device {
...
+                Some((_, fbo)) => *fbo,
+                None => dst.fbos[i],
+            });
             self.blit_render_target(rect, rect);
+
+            // Copy from the intermediate buffer to the dst texture if necessary
+            if let Some((_, fbo)) = &temp_buffer {
```
kvark: here and below: when matching we don't need the reference, instead we can avoid the fields moved out by marking them as `Some((_, ref fbo))`. IIRC, recent Rust inserts some of those `ref` automatically, but I'm not a big fan of that

webrender/src/device/gl.rs
```diff
@@ -1621,12 +1631,47 @@ impl Device {
...
+                self.bind_read_target_impl(*fbo);
+                self.bind_texture(DEFAULT_TEXTURE, dst);
+                self.gl.copy_tex_sub_image_3d(dst.target, 0, 0, 0, i as _,
```
kvark: note that we have 2 bugs here covered by the same flag: inability to blit to layer != 0 and inability to copy from layer != 0. The flag name itself only reflects the first bug, which is a bit unclear.

staktrace:
For posterity: this patch fixes the issues on my Xperia XZ1 Compact (which has an Adreno 540) but doesn't fix the problems on my Xperia Z3 Compact (which has an Adreno 330).

kvark:
@staktrace do we know what other issues Adreno 330 has in addition to these?

staktrace:
It's a little hard to tell which of the visible artifacts are caused by this issue and which are caused by other issues. Certainly fixing this issue on Adreno 540 results in much better behaviour, so either all the visible artifacts on the 330 are caused by this problem, or there's multiple problems that only affect the 330. I don't have a good sense for how likely that is.

Manishearth:
https://github.com/servo/servo/issues/22232 is also fixed by this

Manishearth:
@kvark any progress on this being merged? it fixes a rather annoying Servo bug too.

jdm:
@Manishearth see #3338.

PR #3329, opened 2018-11-19
Don't look up uniforms in create_program since we also do it during l…

mattwoodrow:
…ink_program.

Looks like this is flushing async compiling, and snuck in during a rebase :(

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3329)

### DIFF
--- a/webrender/src/device/gl.rs
+++ b/webrender/src/device/gl.rs
@@ -1892,13 +1892,11 @@ impl Device {
             sources,
         });
 
-        let u_transform = self.gl.get_uniform_location(pid, "uTransform");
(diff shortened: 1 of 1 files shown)


webrender/src/device/gl.rs
```diff
@@ -1892,13 +1892,10 @@ impl Device {
...
         });
 
-        let u_transform = self.gl.get_uniform_location(pid, "uTransform");
-        let u_mode = self.gl.get_uniform_location(pid, "uMode");
-
         let program = Program {
             id: pid,
-            u_transform,
-            u_mode,
+            u_transform: 0,
```
emilio: Maybe add a comment here pointing to `link_program`?

PR #3331, opened 2018-11-20
Ensure that coord system id is propagated for non-invertible nodes.

gw3583:
When a spatial node is encountered that has an invalid matrix, we
need to ensure that the coord system ID is always propagated. This
avoids array indexing errors that can occur during calls to
get_relative_transform.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1507323.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3331)

### DIFF
--- a/webrender/src/spatial_node.rs
+++ b/webrender/src/spatial_node.rs
@@ -191,8 +191,12 @@ impl SpatialNode {
         true
     }
 
-    pub fn mark_uninvertible(&mut self) {
+    pub fn mark_uninvertible(
+        &mut self,
+        state: &TransformUpdateState,
+    ) {
         self.invertible = false;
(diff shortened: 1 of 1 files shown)


gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c98ea88d3b345d3764a7d81a3fa8a03d0c173bef

emilio:
(Is there any chance to land a crashtest of some sort for this?)

gw3583:
@emilio Yea - it's a bit tricky in wrench, since it relies on animation. But we should be able to either do a wrench rawtest or a browser crashtest.

gw3583:
Try run looks good.

@bors-servo r=kvark

PR #3332, opened 2018-11-21
Picture caching support, part 1 (of 2).

gw3583:
This patch introduces a picture composite mode that caches the
surface in a grid of tiles in the texture cache. This is enabled
by setting the picture composite mode to PictureCompositeMode::TileCache.

Basic overview:
 * During initial picture traversal, any tile caches encountered
   are pushed / popped onto the PictureUpdateState stack.
 * If a tile cache is present, and if the transforms have changed
   since last frame build, then the primitive instances are walked,
   mapping each prim instance to a set of tiles in the cache. This
   allows dependencies for each tile (transforms, images, animated
   property bindings etc) to be set. Although this does involve an
   extra pass of primitive instances, it's only used when the tile
   cache mappings (the relative transforms) have changed, so it's
   done quite rarely. In addition, in the future we can move much
   of the second primitive traversal (clip chain building) etc to
   also happen here, which will mean that doesn't get done each
   frame build.
 * During prim prepare, a picture with tile cache enabled checks
   the dependencies on each tile, to see if it needs to be invalidated.
   It builds a dirty rect that is the union of the set of dirty
   tiles for this picture. Right now, we support only a single
   dirty rect for simplicity, but this can easily be expanded in
   the future to support a number of regions as required. This
   dirty rect is then passed via the PictureContext to any child
   primitives, and is used as the culling rect for primitive, clip
   mask and child picture allocations.
 * The picture preparation step also maintains a list of texture
   cache handles per tile, and records a series of blit commands
   for the renderer to execute after drawing the picture. These
   blits are executed to cache the picture output in texture
   cache tiles.
 * On subsequent frames, if the tile(s) are not invalid, they are
   drawn as a series of images with the brush_image shader.

Other changes included in this patch series:
 * Make reftest failure messages for draw/color targets more useful.
 * Make animation example more useful for picture caching debugging.
 * Fix a texture cache bug where UV rect kind isn't updated if the size of the texture cache update is the same.
 * Add a method to resource cache to check if an image is dirty.
 * Shader changes to support using prim header for segment rect and/or stretch size.
 * Add method to scene properties to get a current float value by id.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3332)

### DIFF
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -24,13 +24,73 @@ use euclid::Angle;
 use webrender::api::*;
 
 struct App {
-    property_key: PropertyBindingKey<LayoutTransform>,
+    property_key0: PropertyBindingKey<LayoutTransform>,
+    property_key1: PropertyBindingKey<LayoutTransform>,
+    property_key2: PropertyBindingKey<LayoutTransform>,
     opacity_key: PropertyBindingKey<f32>,
-    transform: LayoutTransform,
     opacity: f32,
+    angle0: f32,
+    angle1: f32,
+    angle2: f32,
+}
+
+impl App {
+    fn add_rounded_rect(
+        &mut self,
+        bounds: LayoutRect,
+        color: ColorF,
+        builder: &mut DisplayListBuilder,
+        property_key: PropertyBindingKey<LayoutTransform>,
+    ) {
+        let filters = [
+            FilterOp::Opacity(PropertyBinding::Binding(self.opacity_key, self.opacity), self.opacity),
+        ];
+
+        let reference_frame_id = builder.push_reference_frame(
+            &LayoutPrimitiveInfo::new(LayoutRect::new(bounds.origin, LayoutSize::zero())),
+            Some(PropertyBinding::Binding(property_key, LayoutTransform::identity())),
+            None,
+        );
+
+        builder.push_clip_id(reference_frame_id);
+
+        builder.push_stacking_context(
+            &LayoutPrimitiveInfo::new(LayoutRect::zero()),
+            None,
+            TransformStyle::Flat,
+            MixBlendMode::Normal,
+            &filters,
+            RasterSpace::Screen,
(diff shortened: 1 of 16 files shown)


gw3583:
This patch doesn't actually ever set a picture to use TileCache
mode! In general, I dislike the idea of landing a patch where
the code isn't enabled, however I think it makes sense
to make an exception in this case, for a number of reasons:

 * I have verified manually that this is working in all wrench
   reftests and prototyped in Gecko successfully. There should
   be a fairly small window of time between merging this and
   when we enable it.
 * Until more of the primitive interning work lands, this is
   only of benefit during scrolling of the same display list.
   Ideally we'd enable the tile cache mode once primitives are
   being interned.
 * This patch is already very large, and landing it now gives
   some time for the code changes that are run in non-tile-cache
   mode to settle, while avoiding rebase pain. It should also
   minimize the chance of any regressions for now, while I
   continue to work on part 2 (primitive interning).
 * The sooner I can get these parts reviewed and a sanity
   check on the general approach, the better.
 * There's a small performance regression here, due to the
   extra tile assignment pass. This is noticeable because it's
   being rebuilt every time a new DL arrives. Once the prim
   interning is in place, this will not occur, which should
   help performance a lot of some of the talos tests.

If you feel strongly that we shouldn't land this in a disabled
state, let me know and we can discuss.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=80ab89d33da4d5433bb5cb8c675d4519a73b9029

r? @kvark @nical and anyone else interested.

gw3583:
Try looks good.

gw3583:
@emilio and @kvark Thanks for the reviews! I think that's all comments addressed.

gw3583:
Rebased on top of the recent opacity bindings change.

kvark:
:shipit: @bors-servo r+

PR #3335, opened 2018-11-21
Add PrimitiveScratchBuffer implementation.

gw3583:
This splits some information from the PrimitiveStore into the new
scratch buffer type.

The scratch buffer is specifically for storing information that
is only read / written during frame building, as opposed to other
items in the primitive store which is created during scene building.

Since the scratch buffer information is not moved between threads,
it's easy to retain it between both frame builds and also when a
new display list is sent by the scene builder thread.

When a new scene arrives, the recycle method allows the storage
of the scratch buffer to possibly be shrunk, if it is now
consuming much more memory than was used on the previous frame.

NOTE: This initial commit only uses the scratch buffer for clip
      mask instances, but there will be more fields moved to make
      use of the retained scratch buffer soon.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3335)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -617,7 +617,7 @@ impl AlphaBatchBuilder {
                     transform_id,
                 };
 
-                let glyph_keys = &ctx.prim_store.glyph_keys[run.glyph_keys_range];
+                let glyph_keys = &ctx.scratch.glyph_keys[run.glyph_keys_range];
 
                 ctx.resource_cache.fetch_glyphs(
                     run.used_font.clone(),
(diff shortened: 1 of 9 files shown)


gw3583:
r? @nical or @kvark or anyone

I don't expect this to have measurable effect on dl_mutate and friends, but it should help in the future as more fields make use of it.

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=ed36d570f1bf761d4dd589fc23694582610bc6f2

gw3583:
Added a follow up commit that moves the glyphs array to the retained scratch buffer.

gw3583:
Added a third commit:

```
Record the used size of the prim store vecs that can't be retained
after each scene build, and use this as a capacity estimate for
those arrays next time we build a scene.
```

Pending try with all those patches:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=295fa451a53f7b4480eedc192906d4cd9285cd5e

PR #3336, opened 2018-11-22
Switch opacity bindings to array, and optimize size of BrushPrimitive.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3336)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -16,7 +16,7 @@ use gpu_types::{PrimitiveHeader, PrimitiveHeaderIndex, TransformPaletteId, Trans
 use internal_types::{FastHashMap, SavedTargetIndex, TextureSource};
 use picture::{Picture3DContext, PictureCompositeMode, PicturePrimitive, PictureSurface};
 use prim_store::{BrushKind, BrushPrimitive, DeferredResolve};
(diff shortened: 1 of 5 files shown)


gw3583:
r? @nical or @kvark or anyone

There are two patches here.

The first moves opacity binding from being inline in the brush primitive to being a separate array. This makes things easier once I intern images and rectangles.

The second patch makes the border segments be a vec instead of a smallvec. I noticed that the size of the BrushPrimitive was very large, which is fixed by this. This is a temporary fix, it won't be needed once the primitives are interned, but it is a big enough win that it's worth having for now. It drops the size of each ```BrushKind``` from 1632 bytes down to 168 bytes. This drops the ```dl_mutate``` talos test from ~5300 to ~4700, which is a significant win.

gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7160b048de4bc64e76bee67a8b8401b5ec46217d

gw3583:
Try run looks good.

nical:
It'd be nice to sotre segments in a separate vec (or segmented vec) in the primitive store or some per-cluster data structure and refer to them by ids like this PR does with opacity bindings.

@bors-servo r+

PR #3339, opened 2018-11-22
Add default deserializer for frame_id

kvark:
Since frame ID is debug-only, it's not serialized when capturing from a release-built Firefox, making us unable to replay it with debug Wrench. This PR fixes this (broken recently by #3289).

r? anyone

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3339)


kvark:
oops, gotta gate for replay feature
@bors-servo r=emilio

PR #3340, opened 2018-11-22
Add unit test to detect size changes in critical structures.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3340)


nox:
When (I don't remember if that's already the case) `std::mem::size_of` is finally usable in constants, we will be able to write those tests as assertions (using `std::mem::transmute`) that fail to compile if false.

PR #3342, opened 2018-11-23
Port normal (css) style borders to be interned primitives.

gw3583:
This is mostly a straightforward conversion of normal borders
to make use of primitive interning.

It also adds support during batching and clip task creation for
interned primitives that make use of segments. This is needed
for porting the rest of the primitives, and should make the
conversion of the other primitive types simple.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3342)


gw3583:
Added a follow up commit that also ports image borders to be interned.

New pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=02be9f170cef1af94e62ffdb8fce438d354836fa

gw3583:
Some of these changes are quite messy, with duplicated code in places. This could definitely do with a tidy / clean up pass once all the primitive types are ported and we remove the BrushPrimitive type.

gw3583:
Squashed and rebased on the picture caching commit.

Try run looks good, I think. There are a couple of crash test failures (with known intermittents), and a border failure in R1 on OSX. The border failure seems to be text related, I think it's probably due to one of the other recent commits in master?

gw3583:
@kvark Thanks for the review! I think I answered / addressed all the comments.

Going to merge now as you mentioned there's no blockers, and I want to avoid rebasing it too often. Please just comment here if there's further follow ups based on those review comments.

@bors-servo r=kvark

Issue #2236, opened 2017-12-18
Capture: support tiled blob images


kvark:
Fetching a tiled blob seems to require a separate (carefully formed) request per tile. Then we'll need to re-shuffle the data received, since it's not going to occupy a continuous block in the whole image raw.

PR #3344, opened 2018-11-23
Basic support for capturing tiled images

kvark:
Related (but doesn't completely fix) to https://github.com/servo/webrender/issues/2236
Only ones that actually take a single tile are supported atm, and we also try not to crash on others. This is strictly an improvement over the current behavior (that just panics).

r? anyone


kvark:
rebased
@bors-servo r=nical

PR #3350, opened 2018-11-26
Port image primitives to use interning.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3350)


gw3583:
Thanks @nical and @kvark. Rebased this on master and addressed the review comments above.

PR #3352, opened 2018-11-26
Port linear gradients to be interned primitives.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3352)

### DIFF
--- a/webrender/src/border.rs
+++ b/webrender/src/border.rs
@@ -1128,169 +1128,171 @@ pub fn build_border_instances(
     instances
 }
 
-pub fn create_nine_patch_segments(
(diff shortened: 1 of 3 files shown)


gw3583:
Try run looks good so far. Also seems to be about another 2% win in dl_mutate (https://treeherder.mozilla.org/perf.html#/graphs?series=try,1660472,1,1&selected=try,1660472,407537,652574374).

gw3583:
Rebased and fixed one of the review comments. I agree with the other two suggestions too - although would prefer to do those as follow ups, to avoid rebase issues with other open patches.

PR #3349, opened 2018-11-24
Add primitive interning for YUV images.

gw3583:
Also restructure and tidy up some of the code that handles
segment building + interned primitives. This reduces some of
the code duplication and will simplify porting the remaining
primitives to interning.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3349)

PR #3353, opened 2018-11-26
Port radial gradients to interning, remove all legacy prim structures.

gw3583:
This ports the last legacy primitive to be interned (radial gradients), and removes all old primitive structures. It also removes PrimitiveContainer, which is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3353)


gw3583:
r? @kvark or anyone else.

Thus completes the conversion of all primitives to interning, which gives us a stable unique ID for primitives based on content.

This patch set is huge, and I apologize to anyone reviewing it. It's mostly shuffling code around, and also includes the patches in #3349, #3350, #3352, which can be reviewed separately, if easier.

gw3583:
Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=218615242ebba06de5d4e6fbb6045546808cf806

gw3583:
Try run looks good so far, and a small gain in dl_mutate, which is a bonus.

PR #3354, opened 2018-11-26
Fix panic when updating a non-tiled "multi" image cache entry.

nical:
In  #3277 I introduced the expectation that if a cached image entry is `ImageResult::Multi(..)` then it must be tiled and we can `unwrap` an optional tiling related variable.

It turns out (from [bug 1509643](https://bugzilla.mozilla.org/show_bug.cgi?id=1509643)) that this assumption doesn't always hold true. This change restores the previous behavior and also invalidates the entire entry if it is tiled but we don't have information about the tile size. I don't know for sure whether this case can happen in practice but I'd rather be conservative about it since invalidation bugs tend to be hard to track down.

### DIFF
--- a/webrender/src/resource_cache.rs
+++ b/webrender/src/resource_cache.rs
@@ -794,11 +794,10 @@ impl ResourceCache {
                 entry.dirty_rect = entry.dirty_rect.union(dirty_rect);
             }
             Some(&mut ImageResult::Multi(ref mut entries)) => {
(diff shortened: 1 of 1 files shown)


nical:
try push https://treeherder.mozilla.org/#/jobs?repo=try&revision=c32a56e7df07d5b5eda0b5f9ecd247c207e09315

PR #3359, opened 2018-11-27
Allow picture caching to work across scenes (display lists).

gw3583:
This patch hooks up the primitive interning work with the tile
caching work.

Since we have a uid that guarantees uniqueness of the content of
primitives and clip nodes, we can use that to construct a
(reasonably) efficient hash key for the content of a tile.

Along with some additional information (such as the state of the
transforms used by the tile), we can construct a hash key for
the content of a tile that is stable across new display lists,
even if the shape of the clip-scroll tree changes in ways that
don't affect the tile.

This patch takes advantage of that to retain tiles when a new
scene is built, where the content of a tile would result in
the same output.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3359)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1001,7 +1001,7 @@ impl AlphaBatchBuilder {
                                         let tile = &tile_cache.tiles[i as usize];
 
                                         // Check if the tile is visible.
-                                        if !tile.is_visible {
+                                        if !tile.is_visible || !tile.in_use {
                                             continue;
                                         }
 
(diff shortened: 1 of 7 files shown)


gw3583:
r? @kvark or @nical

This gives us the ability to cache picture tiles between both frames and scenes.

This is a somewhat more interesting patch than all the previous primitive interning changes :)

Note that it's still not actually hooked up to create pictures with tile caches enabled (in master), so this shouldn't have any effect on current gecko.

Nonetheless, pending try run to be safe:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=0432779ca9a8f9ddd7fb6b5c932b7f72dba6a2b8

gw3583:
@kvark Thanks for the review! Addressed those comments, and rebased on current master. Try run looks good (the failures are expected given previous PRs pending update in m-c).

PR #3360, opened 2018-11-27
[WIP] Avoid copies of large picture primitives during flattening

kvark:
Addresses some of the bigger #3358 parts related to `PicturePrimitive`. Adds the basic `push_with` helper to vectors that avoid copy (kudos to @jrmuizel ).
TODO:
- [ ] verify that the copies are gone (need the tool hosted/opened somewhere)
- [ ] rebase
- [ ] Gecko try push with talos jobs

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3360)

PR #3362, opened 2018-11-27
Avoid picture primitive copies via VecHelper

kvark:
This is a successor of #3360 that avoids the borrow checker dance via RAII 
Addresses part of #3358

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3362)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -372,7 +372,7 @@ impl FrameBuilder {
         let mut profile_counters = FrameProfileCounters::new();
         profile_counters
             .total_primitives
-            .set(self.prim_store.prim_count);
+            .set(self.prim_store.prim_count());
 
         resource_cache.begin_frame(stamp);
         gpu_cache.begin_frame(stamp.frame_id());
--- a/webrender/src/util.rs
+++ b/webrender/src/util.rs
@@ -8,11 +8,52 @@ use euclid::{Point2D, Rect, Size2D, TypedPoint2D, TypedRect, TypedSize2D, Vector
 use euclid::{TypedTransform2D, TypedTransform3D, TypedVector2D, TypedScale};
 use num_traits::Zero;
 use plane_split::{Clipper, Polygon};
-use std::{i32, f32, fmt};
+use std::{i32, f32, fmt, ptr};
 use std::borrow::Cow;
+
 
 // Matches the definition of SK_ScalarNearlyZero in Skia.
 const NEARLY_ZERO: f32 = 1.0 / 4096.0;
+
+/// A typesafe helper that separates new value construction from
+/// vector growing, allowing LLVM to ideally construct the element in place.
+#[must_use]
+pub struct Allocation<'a, T: 'a> {
+    vec: &'a mut Vec<T>,
+    index: usize,
+}
+
+impl<'a, T> Allocation<'a, T> {
+    // writing is safe because alloc() ensured enough capacity
+    // and `Allocation` holds a mutable borrow to prevent anyone else
+    // from breaking this invariant.
+    #[inline(always)]
+    pub fn init(self, value: T) -> usize {
+        unsafe {
+            ptr::write(self.vec.as_mut_ptr().add(self.index), value);
+            self.vec.set_len(self.index + 1);
+        }
+        self.index
+    }
+}
+
+pub trait VecHelper<T> {
+    fn alloc(&mut self) -> Allocation<T>;
+}
+
+impl<T> VecHelper<T> for Vec<T> {
+    fn alloc(&mut self) -> Allocation<T> {
+        let index = self.len();
+        if self.capacity() == index {
+            self.reserve(1);
+        }
+        Allocation {
+            vec: self,
+            index,
+        }
+    }
+}
+
 
 // Represents an optimized transform where there is only
 // a scale and translation (which are guaranteed to maintain
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -34,7 +34,7 @@ use spatial_node::{StickyFrameInfo};
 use std::{f32, mem};
 use std::collections::vec_deque::VecDeque;
 use tiling::{CompositeOps};
-use util::{MaxRect};
+use util::{MaxRect, VecHelper};
 
 #[derive(Debug, Copy, Clone)]
 struct ClipNode {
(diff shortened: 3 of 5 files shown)


webrender/src/util.rs
```diff
@@ -8,12 +8,50 @@ use euclid::{Point2D, Rect, Size2D, TypedPoint2D, TypedRect, TypedSize2D, Vector
...
-use std::{i32, f32, fmt};
+use std::{i32, f32, fmt, ptr};
 use std::borrow::Cow;
 
+
 // Matches the definition of SK_ScalarNearlyZero in Skia.
 const NEARLY_ZERO: f32 = 1.0 / 4096.0;
 
+/// A typesafe helper that separates new value construction from
+/// vector growing, which allows LLVM to elide the value copy.
```
jrmuizel: "and ideally construct the element in place"

webrender/src/util.rs
```diff
@@ -8,12 +8,50 @@ use euclid::{Point2D, Rect, Size2D, TypedPoint2D, TypedRect, TypedSize2D, Vector
...
+#[must_use]
+pub struct Allocation<'a, T: 'a> {
+    vec: &'a mut Vec<T>,
+    index: usize,
+}
+
+impl<'a, T> Allocation<'a, T> {
+    #[inline(always)]
+    pub fn init(self, value: T) -> usize {
```
jrmuizel: Maybe add:
// writing is safe because alloc() ensured enough capacity and Allocation holds a mutable borrow to prevent anyone else from breaking this invariant.

webrender/src/frame_builder.rs
```diff
@@ -338,7 +338,7 @@ impl FrameBuilder {
         let mut profile_counters = FrameProfileCounters::new();
         profile_counters
             .total_primitives
-            .set(self.prim_store.prim_count);
+            .set(self.prim_store.prim_count());
```
jrmuizel: It'd be nice if this only ran if profiling was on.

webrender/src/util.rs
emilio: Are you sure this actually avoids the memcpy / memmove? At least for `self` arguments a bit back we couldn't work around it like this, see https://github.com/rust-lang/rust/issues/42763.

kvark:
I'm not able to use @jrmuizel 's tool that find memcopies, since IR validation fails on `webrender` crate. However, I tried to [reproduce this in playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2015&gist=24c9f7a7e5a9619b23c6f5040688f0fb), and in both cases (`vec.push(x)` and `vec.alloc().init(x)`) the compiler created the data on the stack first and then issued a single memcpy, so it doesn't appear to be fixing the copy.

Is my playground experiment incorrect? Otherwise, what assumptions are wrong?

jrmuizel:
If you use an enum containing an array instead of an array and initialize
with a small enum variant the copy is elided in your example

kvark:
@jrmuizel indeed, I confirmed it with the playground experiment.
Therefore, we should proceed with this PR.

webrender/src/util.rs
kvark: Confirmed in [playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2015&gist=9aef0043bcb9a15155ae97119681f00d). In `fn foo()` changing the `push(xxx)` to `alloc().init(xxx)` removes the `memcpy`.

jrmuizel:
It looks like this may not have worked:
```
memcpy of 1592 in _ZN9webrender22display_list_flattener20DisplayListFlattener20pop_stacking_context17hc7e8ce6ec929e9eaE @
  write<webrender::picture::PicturePrimitive> /rustc/6f93e93af6f823948cc13d2938957757c6486d88/src/libcore/ptr.rs:739
  init<webrender::picture::PicturePrimitive> webrender/src/util.rs:33
  pop_stacking_context webrender/src/display_list_flattener.rs:1220
memcpy of 1592 in _ZN9webrender22display_list_flattener20DisplayListFlattener20pop_stacking_context17hc7e8ce6ec929e9eaE @
  write<webrender::picture::PicturePrimitive> /rustc/6f93e93af6f823948cc13d2938957757c6486d88/src/libcore/ptr.rs:739
  init<webrender::picture::PicturePrimitive> webrender/src/util.rs:33
  pop_stacking_context webrender/src/display_list_flattener.rs:1187
memcpy of 1592 in _ZN9webrender22display_list_flattener20DisplayListFlattener21push_stacking_context17h500d71065754d89aE @
  write<webrender::picture::PicturePrimitive> /rustc/6f93e93af6f823948cc13d2938957757c6486d88/src/libcore/ptr.rs:739
  init<webrender::picture::PicturePrimitive> webrender/src/util.rs:33
  cut_flat_item_sequence webrender/src/display_list_flattener.rs:2207
  push_stacking_context webrender/src/display_list_flattener.rs:994
memcpy of 1592 in _ZN9webrender22display_list_flattener20DisplayListFlattener20pop_stacking_context17hc7e8ce6ec929e9eaE @
  write<webrender::picture::PicturePrimitive> /rustc/6f93e93af6f823948cc13d2938957757c6486d88/src/libcore/ptr.rs:739
  init<webrender::picture::PicturePrimitive> webrender/src/util.rs:33
  pop_stacking_context webrender/src/display_list_flattener.rs:1158
memcpy of 1592 in _ZN9webrender22display_list_flattener20DisplayListFlattener12flatten_item17hfd2b2318f8c75547E @
  write<webrender::picture::PicturePrimitive> /rustc/6f93e93af6f823948cc13d2938957757c6486d88/src/libcore/ptr.rs:739
  init<webrender::picture::PicturePrimitive> webrender/src/util.rs:33
  pop_all_shadows webrender/src/display_list_flattener.rs:1564
  flatten_item webrender/src/display_list_flattener.rs:789
memcpy of 1592 in _ZN9webrender22display_list_flattener20DisplayListFlattener20pop_stacking_context17hc7e8ce6ec929e9eaE @
  write<webrender::picture::PicturePrimitive> /rustc/6f93e93af6f823948cc13d2938957757c6486d88/src/libcore/ptr.rs:739
  init<webrender::picture::PicturePrimitive> webrender/src/util.rs:33
  pop_stacking_context webrender/src/display_list_flattener.rs:1117
```

I wonder if llvm gets sad the enum variant is too big.

webrender/src/util.rs
jrmuizel: I filed https://github.com/rust-lang/rust/issues/56333 for a problem contributing to the first example not working well in the playground.

jrmuizel:
It looks like SmallVec is contributing to the sadness:

```rust
#![crate_type = "lib"]

extern crate smallvec;

use smallvec::SmallVec;

#[derive(Default)]
pub struct L {
    a: SmallVec<[f64; 16]>,
    b: SmallVec<[f64; 16]>,
    c: SmallVec<[f64; 16]>
}

pub struct Allocation<T> {
    f: *mut T,
}
use std::ptr;

impl<T> Allocation<T> {
    pub fn init(self, value: T) {
        unsafe { ptr::write(self.f, value) };
    }
}

#[inline(never)]
pub fn bar(a: Allocation<L>) {
    a.init(L{a: SmallVec::new(), b: SmallVec::new(), c: SmallVec::new()});
}
```
compiles to
```asm
playground::bar:
	pushq	%rbx
	subq	$720, %rsp
	movq	%rdi, %rbx
	xorps	%xmm0, %xmm0
	movaps	%xmm0, 288(%rsp)
	movaps	%xmm0, (%rsp)
	movaps	%xmm0, 144(%rsp)
	leaq	432(%rsp), %rdi
	movq	%rsp, %rsi
	movl	$144, %edx
	callq	memcpy@PLT
	leaq	576(%rsp), %rdi
	leaq	144(%rsp), %rsi
	movl	$144, %edx
	callq	memcpy@PLT
	leaq	288(%rsp), %rsi
	movl	$432, %edx
	movq	%rbx, %rdi
	callq	memcpy@PLT
	addq	$720, %rsp
	popq	%rbx
	retq
```

webrender/src/util.rs
jrmuizel: It looks like SmallVec also causes this to not work. I filed https://github.com/rust-lang/rust/issues/56356 about that.

jrmuizel:
The smallvec example now generates better code with Nightly:

```asm
playground::bar: # @playground::bar
# %bb.0:
	subq	$440, %rsp              # imm = 0x1B8
	xorps	%xmm0, %xmm0
	movaps	%xmm0, (%rsp)
	movaps	%xmm0, 144(%rsp)
	movaps	%xmm0, 288(%rsp)
	movq	%rsp, %rsi
	movl	$432, %edx              # imm = 0x1B0
	callq	*memcpy@GOTPCREL(%rip)
	addq	$440, %rsp              # imm = 0x1B8
	retq
```

However it looks like there's still an extra memcpy

jrmuizel:
I filed a follow up at https://github.com/rust-lang/rust/issues/58082

PR #3365, opened 2018-11-28
Fix a bug in invalidating tiles from a previous scene.

gw3583:
Any retained tiles from a previous scene were being dropped if not
used after building the dirty region for the first tile cache.

This works in simple applications, but in Gecko, where multiple
tile caches exist, it means that we were dropping any retained
tiles that other tile caches may want, after the first tile cache
updated.

Instead, only drop previous scene tiles after all tile caches
have had a chance to retain tiles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3365)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -278,6 +278,14 @@ impl FrameBuilder {
             &mut retained_tiles,
         );
 
+        // If we had any retained tiles from the last scene that were not picked
+        // up by the new frame, then just discard them eagerly.
+        // TODO(gw): Maybe it's worth keeping them around for a bit longer in
+        //           some cases?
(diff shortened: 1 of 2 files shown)


gw3583:
Found while integrating the picture caching into gecko.

r? @kvark or anyone

kvark:
@gw3583 it would be good to have some kind of test coverage for the tile cache. Perhaps, we could have reftests where you provide 2 yamls (with some attributes to select the cached tile-ability of stacking contexts) and then specify the expectation on the number of surfaces drawn?

@bors-servo r+

gw3583:
Yea, I've been thinking a bit about the best way to test it. I have some ideas, could be a good discussion topic for next week too.

PR #3366, opened 2018-11-28
Reduce data copies during internation

kvark:
Excessive copying in `fn intern()` is something we noticed yesterday with @jrmuizel .
This PR attempts to improve them in two ways (each in a separate commit):
  1. reduce the `Update` enum size by moving the data out
  2. adopt entry-like API to accommodate the common pattern of `if index == v.len() { v.push(xxx) } else { v[index] = xxx; }` without panics between element construction and actual assignment. It builds upon the `VecHelper` work of #3362.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3366)

### DIFF
--- a/webrender/src/intern.rs
+++ b/webrender/src/intern.rs
@@ -142,16 +142,10 @@ impl<S, T, M> DataStore<S, T, M> where S: Debug, T: From<S>, M: Debug {
         for update in update_list.updates {
             match update.kind {
                 UpdateKind::Insert => {
-                    let data = data_iter.next().unwrap();
-                    let item = Item {
-                        data: T::from(data),
+                    self.items.entry(update.index).set(Item {
+                        data: T::from(data_iter.next().unwrap()),
                         epoch: update_list.epoch,
-                    };
-                    if self.items.len() == update.index {
-                        self.items.push(item)
-                    } else {
-                        self.items[update.index] = item;
-                    }
+                    });
                 }
                 UpdateKind::Remove => {
                     self.items[update.index].epoch = Epoch::INVALID;
@@ -161,6 +155,7 @@ impl<S, T, M> DataStore<S, T, M> where S: Debug, T: From<S>, M: Debug {
                 }
             }
         }
+        debug_assert!(data_iter.next().is_none());
     }
 }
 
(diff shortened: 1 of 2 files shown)


kvark:
@gw3583 yep, will do

Also need to confirm with the playground, since in this case (unlike with Allocation) we still have a branch separating an item constructor from the target location, even though it's not a panicing one.

kvark:
Interestingly, a [Playground experiment](https://play.rust-lang.org/?version=stable&mode=release&edition=2015&gist=8448a38419a461eb1a4376e0e6c38771) shows a reduction of LLVM copy calls from 4 to 0 when using this change. Seems pretty ridiculous, you are encouraged to play with it and maybe tell what I'm doing wrong :)

jrmuizel:
Here's what my tool can tell us about where the copies are coming from:
```
memcpy of 402 in _ZN5entry3foo17h9b5dcf419aa0590cE @
  foo entry.rs:78
memcpy of 402 in _ZN5entry3foo17h9b5dcf419aa0590cE @
  write<entry::Thing> /rustc/400c2bc5ed292f77c49693320f4eda37bb375e90/src/libcore/ptr.rs:736
  push<entry::Thing> /rustc/400c2bc5ed292f77c49693320f4eda37bb375e90/src/liballoc/vec.rs:1003
  foo entry.rs:78
memcpy of 402 in _ZN5entry3foo17h9b5dcf419aa0590cE @
  foo entry.rs:80
memcpy of 402 in _ZN5entry3foo17h9b5dcf419aa0590cE @
  foo entry.rs:80
```

4 copies does seem excessive, but it is facts.

kvark:
@jrmuizel thanks for confirmation! It's going to be a big win then :)
I'm currently having trouble getting sane talos comparisons, but I'm fairly sure this is safe to land (given the playground results so far).
Green try: https://treeherder.mozilla.org/#/jobs?repo=try&selectedJob=214670152&revision=951e31ce0aa13802a0b12211f85013f57e99dad0

jrmuizel:
It would be worth comparing before/after on webrender.ll. You should be able to avoid the verification errors by using rust nightly. (That fixed it for me)

jrmuizel:
I suspect we're not seeing the improvement we expect because of https://github.com/rust-lang/rust/issues/56191. @nikic has a patch to llvm that might improve things here.

kvark:
Got 5% improvement in dl_mutate from that: https://treeherder.mozilla.org/perf.html#/alerts?id=18030
:tada:

PR #3367, opened 2018-11-28
Add a RendererOption to configure if picture caching is enabled.

gw3583:
Although not hooked up internally yet, this provides a hook so
that the gecko pref can be added.

We'll use this to provide a temporary way to compare performance
and also to allow picture caching to be disabled temporarily if
we encounter any bad regressions.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3367)

### DIFF
--- a/webrender/src/renderer.rs
+++ b/webrender/src/renderer.rs
@@ -4771,6 +4771,7 @@ pub struct RendererOptions {
     pub chase_primitive: ChasePrimitive,
     pub support_low_priority_transactions: bool,
     pub namespace_alloc_by_client: bool,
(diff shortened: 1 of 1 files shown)


gw3583:
(or @kvark or anyone really)

kvark:
@gw3583 it would be great to have 3 options here: force-enabled, force-disabled, and manual (specified per-SC). Your words got stuck in my head about the fact that picture caching is going to hide some of our performance issues, and this is worrying me. So I'd like to be able to constantly run without it to see how good our baseline is.

PR #3370, opened 2018-11-29
Advance frame ID per frame build, rather than per new scene.

gw3583:
This fixes the way texture eviction works with picture caching
when scrolling or property animation is occurring.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3370)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -410,6 +410,9 @@ impl Document {
         let accumulated_scale_factor = self.view.accumulated_scale_factor();
         let pan = self.view.pan.to_f32() / accumulated_scale_factor;
 
+        // Advance to the next frame.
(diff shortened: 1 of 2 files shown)


gw3583:
r? @bholley and @nical 

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=bf2b5206bda430809faf9184889e9075df49b56a

gw3583:
Try run looks good, I think. @bholley anything we need to do with eviction heuristics before landing this, or can they be tuned after?

bholley:
We should land the updated heuristics at the same time. I just finished rebasing my instrumentation, but still need to measure/analyze. Have some meetings coming up but will try to get to that this afternoon.

PR #3372, opened 2018-11-29
Make sure our blobs are still alive when pushing to missing_blob_images

jrmuizel:
This should also help with bug 1492241.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3372)


mattwoodrow:
It might also be worth querying blob_image_rasterizer to see if the template is in the HashMap within that, since that's the lookup that's failing.

PR #3374, opened 2018-11-29
Extend the scope of render target allocation strategy acrosss layers

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1509672

Technically consists of 3 parts:
  - cleanup bits, don't affect performance
  - spread the dynamic target allocation bins across the layers. Previously, we had each layer with it's own allocator, and we were only ever trying to allocate from the last one. This scheme was weak against cases where we reach the ideal maximum target size, since guillotine allocation of small bits forced us to spawn more and more layers... reaching almost 150 in the target case. New scheme uses the layers more efficiently, reducing the layers to just 3 (.. 50x reduction :P). Pending [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=27464c7d62f0f05a0bc96a7133b55e9706d3d449).
  - when we reach the ideal max size, also round up the requested size to 256. This makes allocations *within a layer* to be more robust against small inputs. With this change, the number of layers drops to just 2 (.. 75x reduction lol), which appears to be minimal for this case. Pending [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=1f4593fd68455842a7b12f396d0abbdf887a11a0).

The page scrolls smoothly with those changes, on my GTX TI 1050 at least.

r? @gw3583 

Note: the bugzilla issue also suffers from poor batching, I'm going to look at it separately.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3374)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1556,6 +1556,7 @@ dependencies = [
  "pathfinder_path_utils 0.2.0 (git+https://github.com/pcwalton/pathfinder?branch=webrender)",
  "plane-split 0.13.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "png 0.12.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "rayon 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "ron 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,8 +8,8 @@ environment:
   TARGET: x86_64-pc-windows-msvc
 
 install:
-  - ps: Start-FileDownload "https://static.rust-lang.org/dist/rust-1.27.0-${env:TARGET}.msi"
-  - msiexec /passive /i "rust-1.27.0-%TARGET%.msi" ADDLOCAL=Rustc,Cargo,Std INSTALLDIR=C:\Rust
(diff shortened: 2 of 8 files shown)


kvark:
Both tries look green. Appveyor used an older Rust version, which is updated in the last commit.
OSX Release TC bots are being updated to 1.30 by @staktrace as we speak.
This is ready for reviews!

gw3583:
Seems sane! Do we need to do any profiling / testing before merging?

kvark:
@gw3583 I visited a few websites with RT debug display enabled, and the improvement is real. On cnn.com we are down from 3 large (~3K x 2K) slices to just one. On HN, similarly, we only have a single slice with everything instead of 3.

Looks like this is going to be a major win on common sites, not just extreme cases. :tada:

gw3583:
@kvark Cool, ship it! :rocket:

staktrace:
The mac builders should have 1.30 now

zptan:
OS X debug tests failed:
```
...
sccache --stop-server || true
Stopping sccache server...
error: couldn't connect to server
caused by: Connection refused (os error 61)
mkdir -p ../artifacts
RUST_LOG=sccache=trace SCCACHE_ERROR_LOG=$PWD/../artifacts/sccache.log sccache --start-server
TRACE:sccache::cmdline: parse
TRACE:sccache::commands: Command::StartServer
Starting sccache server...
TRACE:sccache::commands: run_server_process
TRACE:sccache::cmdline: parse
TRACE:sccache::commands: Command::InternalStartServer
error: Timed out waiting for server startup
...
```

kvark:
@gw3583 I've added  a bit of code for validating texture allocator correctness, at least at test time.
@bors-servo r=gw3583

heftig:
With the latter try build the issue is gone for me. Thanks! :+1:

PR #3375, opened 2018-11-30
Use a hash digest rather than complete source to key shaders

bholley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1510490


bholley:
@staktrace This will need coordiland.

PR #3376, opened 2018-11-30
Make eager texture cache eviction not quite so eager.

gw3583:
Texture cache entries can be evicted at the start of
a frame, or at any time during the frame when a cache
allocation is occurring. This means that entries tagged
with eager eviction may get evicted before they have a
chance to be requested on the current frame. Instead,
advance the frame id of the entry by one before
comparison. This ensures that an eager entry will
not be evicted until it is not used for at least
one complete frame.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3376)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -103,7 +103,7 @@ impl FrameId {
     }
 
     /// Advances this FrameId to the next frame.
-    fn advance(&mut self) {
+    pub fn advance(&mut self) {
         self.0 += 1;
     }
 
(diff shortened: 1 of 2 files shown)


gw3583:
r? @kvark or @nical or @bholley 

(I discussed this with @bholley on IRC and the idea seems correct - just need to make sure I didn't mess up the implementation).

gw3583:
Pending try for completeness:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=972478647c2a441ffac854e59b3a1c5c74ce86ef

gw3583:
Try run looks good, I think.

kvark:
> sccache --stop-server || true
Stopping sccache server...
error: couldn't connect to server


We need to make sscache more fault-tolerant?
@bors-servo retry

PR #3377, opened 2018-11-30
Remove early returns from brush_fs shader functions

zakorgy:
We generate SPIR-V binaries from the shaders and use them in our WR which has `gfx-rs` as backend. For some reason the Vulkan implementation on the Windows try server machines crashes if we have this early return in the `brush_blend.glsl` shader, but works fine with discard.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3377)

### DIFF
--- a/webrender/res/brush_blend.glsl
+++ b/webrender/res/brush_blend.glsl
@@ -148,10 +148,6 @@ vec3 LinearToSrgb(vec3 color) {
 Fragment brush_fs() {
     vec4 Cs = texture(sColor0, vUv);
 
-    if (Cs.a == 0.0) {
-        return Fragment(vec4(0.0)); // could also `discard`
-    }
-
     // Un-premultiply the input.
     float alpha = Cs.a;
     vec3 color = Cs.rgb / Cs.a;
--- a/webrender/res/brush_mix_blend.glsl
+++ b/webrender/res/brush_mix_blend.glsl
@@ -208,16 +208,14 @@ Fragment brush_fs() {
     vec4 Cb = textureLod(sPrevPassColor, vBackdropUv, 0.0);
     vec4 Cs = textureLod(sPrevPassColor, vSrcUv, 0.0);
 
-    if (Cb.a == 0.0) {
-        return Fragment(Cs);
-    }
-    if (Cs.a == 0.0) {
-        return Fragment(vec4(0.0));
+    // The mix-blend-mode functions assume no premultiplied alpha
+    if (Cb.a != 0.0) {
+        Cb.rgb /= Cb.a;
     }
 
-    // The mix-blend-mode functions assume no premultiplied alpha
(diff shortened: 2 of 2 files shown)


leeoniya:
add to https://github.com/servo/webrender/wiki/Driver-issues ?

zakorgy:
Unfortunately I didn't notice that the `brush_mix_blend` shader also has early returns and also crashes on the try server with Vulkan. In this case the related tests are just marked with unexpected-fail and the job runs further. I will look into that and update this PR according to that.

zakorgy:
Also I can update https://github.com/servo/webrender/wiki/Driver-issues after I figured out `brush_mix_blend`.

kvark:
@zakorgy I take it this is still WIP then?

zakorgy:
@kvark Yes, this is.

nical:
@zakorgy the right solution (unfortunately) is probably to avoid both discard and early returns for anything that depends on the input color(s) to prevent timing attacks. We'd carry on with the computations and mask the result at the end.

zakorgy:
I have update the PR according to what @nical proposed. Also I have found, that there are two similar early returns which depends on the color(s) in [cs_blur](https://github.com/servo/webrender/blob/master/webrender/res/cs_blur.glsl#L113) and [pf_vector_cover](https://github.com/servo/webrender/blob/master/webrender/res/pf_vector_cover.glsl#L61). Should I take care those too?

kvark:
@zakorgy let's not touch these for now
@bors-servo r+

webrender/res/brush_blend.glsl
```diff
@@ -148,10 +148,6 @@ vec3 LinearToSrgb(vec3 color) {
 Fragment brush_fs() {
     vec4 Cs = texture(sColor0, vUv);
 
-    if (Cs.a == 0.0) {
-        return Fragment(vec4(0.0)); // could also `discard`
-    }
-
     // Un-premultiply the input.
     float alpha = Cs.a;
     vec3 color = Cs.rgb / Cs.a;
```
emilio: There's nothing preventing division by zero here right?

staktrace:
(For context, this patch seems to be causing https://bugzilla.mozilla.org/show_bug.cgi?id=1513697)

webrender/res/brush_blend.glsl
kvark: yeah, looks like a problem

nical: I think that we need to compute the result (even with the division by zero) and select the result at the end depending on whether the alpha was zero using `mix` or some other trick that dxc is unlikely to rewrite into an early return or early branch (and check that dxc indeed doesn't introduce a branch).
If division by zero is a real issue even when masked out, we can always temporarily set the alpha to 1 and still replace the color at the end of the shader as long as we don't introduce a shortcut.

PR #3378, opened 2018-11-30
Fix cache synchronization crash with blob images.

aosmond:
On the first pass of prepare_transaction:
1) Resource update contains only an AddBlobImage, but it is not visible according to viewport_tiles (set in SetBlobImageVisibleArea)
2) We call ResourceCache::create_blob_scene_builder_requests but since the image is not visible, it doesn't create any BlobImageParams entries (containing the blob image key) to be stored in txn.blob_requests
3) Since the transaction has no blob image updates, we can take the shortcut to call RenderBackend::update_document directly
4) We didn't call ResourceCache::set_blob_rasterizer with the updated txn.blob_rasterizer.

On the second pass of prepare_transaction:
1) Resource update contains only an SetBlobImageVisibleArea
2) We don't call ResourceCache::create_blob_scene_builder_requests since there is no Add/UpdateBlobImage entries.
3) Since the transaction has no blob image updates, we can take the shortcut to call RenderBackend::update_document directly
4) We are missing the blob we added in the previous step in ResourceCache::request_image, and it adds it to missing_blob_images.
5) We try to process missing_blob_images in ResourceCache::rasterize_missing_blob_images with the old blob_rasterizer.
6) Crash.

Potentially fixes bug 1492241.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3378)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -1111,6 +1111,9 @@ impl RenderBackend {
         }
 
         if !transaction_msg.use_scene_builder_thread && txn.can_skip_scene_builder() {
+            if let Some(rasterizer) = txn.blob_rasterizer.take() {
+                self.resource_cache.set_blob_rasterizer(rasterizer);
+            }
             self.update_document(
                 txn.document_id,
                 replace(&mut txn.resource_updates, Vec::new()),


aosmond:
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=5a9f64062defc8cc21e36383d9826a0ec06a0bda

aosmond:
@kvark review? also can you retrigger the failing taskcluster job? Looks like it was an internal error, but it won't accept my auth token for the retrigger...

jrmuizel:
Does this require an empty visible area to trigger?

kvark:
I believe this is safe to merge, especially given that we came up with this change rationally, and the try is green. :crossed_fingers: for the crash to be resolved
@bors-servo r+

jrmuizel:
Here's a try to find empty rects: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9484bdaa41dbb0b0a328ee605fd41d712349ef99

mattwoodrow:
Isn't it possible that there is a slow scene build in progress when this code runs?

When that scene build finishes, it will set the blob rasterizer again, to one that was created before this?

PR #3379, opened 2018-12-01
Add picture caching support.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3379)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -30,8 +30,8 @@ use render_backend::{DocumentView};
 use resource_cache::{FontInstanceMap, ImageRequest};
 use scene::{Scene, ScenePipeline, StackingContextHelpers};
 use scene_builder::DocumentResources;
-use spatial_node::{StickyFrameInfo, ScrollFrameKind};
-use std::{f32, mem};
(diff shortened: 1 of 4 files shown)


gw3583:
r? @kvark

This patch has 3 commits - the first two are simple bug fixes and functionality, and shouldn't be controversial.

The last commit has a couple of horrible hacks in it to enable the first version of picture caching (when the renderer option is enabled, which is off by default), where the code relies on knowledge of what the Gecko display list looks like.

I don't like this change at all, but I think that getting it merged while we work out a better solution (during the all-hands) has a number of benefits:
 * It's off by default, so shouldn't affect anyone.
 * It lets us dogfood the first iteration of picture caching internally by flipping a preference, which allows us to find bugs, compare performance etc.

gw3583:
@kvark Thank you for the quick review! :+1:

pcwalton:
@gw3583 Did you get a chance to profile this by any chance? I'm curious what kinds of perf improvements we saw, if any. (I wouldn't expect much on most sites, based on our experience, but it'd be good to confirm.)

gw3583:
@pcwalton Nothing detailed yet - anecdotally it does seem like quite a win on some sites, but I need to do some proper profiling next week.

PR #3381, opened 2018-12-03
Remove most use of the local prim rect by primitive templates.

gw3583:
The current picture caching code invalidates cached regions more
than expected, due to how gecko handles scroll offsets.

If a display list is received, then a scroll event occurs, and then
a new display list is sent, the local rects will contain different
coordinates, since the scroll offsets are baked into the local
rects rather than in the scroll / reference frames. This results
in primitives being interned with a different value, and cached
surfaces being incorrectly invalidated.

Since the way this works is non-trivial to change in Gecko, we need
to handle this in WR.

If we change primitive and clip keys and templates to store their
rectangles in a true local primitive space (where the origin is
always zero) then the interning will work as expected. To do this,
we'll store the origin of a primitive in the instance, and only
the size in the template.

A side benefit of this approach is that test cases like dl_mutate
will become much faster, since the only difference between the
thousands of rectangles is the origin. This will effectively
allow true instancing of primitives with common data.

This is the first step towards that - by removing reliance of the
primitive local rect from the primitive template code. The rest
will be done as a series of incremental patches.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3381)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1107,7 +1107,7 @@ impl AlphaBatchBuilder {
 
                                         let instance = BrushInstance {
                                             prim_header_index,
-                                            segment_index: 0,
+                                            segment_index: INVALID_SEGMENT_INDEX,
                                             edge_flags: EdgeAaSegmentMask::empty(),
                                             brush_flags: BrushFlags::empty(),
                                             clip_task_address,
@@ -1189,7 +1189,7 @@ impl AlphaBatchBuilder {
                                         let shadow_instance = BrushInstance {
                                             prim_header_index: shadow_prim_header_index,
                                             clip_task_address,
-                                            segment_index: 0,
+                                            segment_index: INVALID_SEGMENT_INDEX,
                                             edge_flags: EdgeAaSegmentMask::empty(),
                                             brush_flags: BrushFlags::empty(),
                                             user_data: shadow_uv_rect_address,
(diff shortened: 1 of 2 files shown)


kvark:
>If a display list is received, then a scroll event occurs, and then
a new display list is sent, the local rects will contain different
coordinates, since the scroll offsets are baked into the local
rects rather than in the scroll / reference frames.

Why isn't a scroll just changing a transform property binding without re-sending the DL or affecting the local rects?

webrender/src/prim_store.rs
```diff
@@ -1192,12 +1185,6 @@ impl PrimitiveTemplateKind {
...
             }
-            PrimitiveTemplateKind::LineDecoration { .. } => {
-                request.write_segment(
-                    prim_rect,
-                    [0.0; 4],
-                );
-            }
             PrimitiveTemplateKind::LinearGradient { ref brush_segments, .. } |
             PrimitiveTemplateKind::RadialGradient { ref brush_segments, .. } => {
                 for segment in brush_segments {
```
kvark: is it fine to still be writing the full rects for the segments?

webrender/src/batch.rs
```diff
@@ -586,7 +586,7 @@ impl AlphaBatchBuilder {
                 };
 
                 let instance = PrimitiveInstanceData::from(BrushInstance {
-                    segment_index: 0,
+                    segment_index: INVALID_SEGMENT_INDEX,
```
kvark: Is this unrelated? Are we changing this segment index at some later point?

gw3583:
> > If a display list is received, then a scroll event occurs, and then
> > a new display list is sent, the local rects will contain different
> > coordinates, since the scroll offsets are baked into the local
> > rects rather than in the scroll / reference frames.
> 
> Why isn't a scroll just changing a transform property binding without re-sending the DL or affecting the local rects?

It is. However, if you scroll in a way that adjusts the scroll transform (as above), then do something that triggers Gecko to send a new DL (e.g. hover over an element that highlights), then Gecko will send a new DL where the current scroll offsets are baked into the local primitive rects, rather than a scroll frame. :disappointed:

PR #3382, opened 2018-12-04
Pass AsyncImageBlobRasterizer through the stack explicitly

kvark:
Similar to #3378, this PR attempts to address our failures in bug 1492241.
If we aren't sure that the blob image rasterizer, being a state of the resource cache, is the right one to build the scene, then perhaps it shouldn't be a state in the first place. This is what this PR is doing. It's not guaranteed to fix anything, just bringing more clarity to the code, hopefully helping us to rule out one of the possible causes.

r? @nical cc @aosmond 

Gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=726e01545625a20153893dbb22509b70294651bb

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3382)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use api::{AsyncBlobImageRasterizer};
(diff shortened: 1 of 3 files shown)


nical:
Sorry I think there's something wrong @bors-servo r- let's talk about it and reland if I misunderstood this.

kvark:
Closing in favor of #3387

PR #3383, opened 2018-12-04
Store clip node templates in local space.

gw3583:
This change makes clip item keys independent of the local space
origin of the node. This ensures that even when Gecko sends a
new display list with clip nodes that have different scroll offsets
the interning logic results in the same uid for the clip node,
avoiding unnecessary invalidations of cached picture tiles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3383)


gw3583:
This is a ~~hack~~ minimal effort to get clip nodes interning correctly even when Gecko supplies a new display list with clip nodes that have scroll offsets applied.

If you have ideas on tidier ways to implement this, let me know.

r? @kvark

gw3583:
Try run looks good.

kvark:
Closing in favor of #3384

PR #3384, opened 2018-12-04
Switch segments to be defined relative to the local primitive position.

gw3583:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3384)


gw3583:
r? @kvark 

This is a very small patch - it just looks big because it includes #3383.

gw3583:
Try run looks good.

PR #3389, opened 2018-12-06
Add display list items for marking content caching region.

gw3583:
This is an experimental API - we may need to change how this
works, depending on how we integrate it into Gecko.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3389)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -975,6 +975,10 @@ impl<'a> DisplayListFlattener<'a> {
             SpecificDisplayItem::PopAllShadows => {
                 self.pop_all_shadows();
             }
+            SpecificDisplayItem::PushCacheMarker(_marker) => {
+            }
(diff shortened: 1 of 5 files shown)


gw3583:
This provides the API hooks for marking the region(s) of a display list that should be cached, allowing Gecko to specify where the main scroll root for the content region exists.

It doesn't actually _do_ anything with those markers yet, this initial patch is just to see if the API seems reasonable, and allow the Gecko integration work to be done.

Does this seem like all we would need @mstange @mattwoodrow @jrmuizel ?

gw3583:
(The scroll root is supplied by the normal WR mechanism, with push/pop clip id methods. We could explicitly make it include the scroll root id if preferred?)

kvark:
We've talked about this with @gw3583 a bit, and this appears to be an acceptable interim solution. We can't get this caching scope bound to any existing item, and will in future revise this based on the results.

PR #3392, opened 2018-12-07
Extract Linear and Radial Gradients.

djg:
In support of #3385, extract gradients into separate interned stores.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3392)

Issue #3385, opened 2018-12-05
Separate Interning of Primitive Data


djg:
Profiling of `displaylist_mutate` has shown a lot of time spent in `memcpy` or `realloc` during scene building surrounding adding primitives. Split apart `PrimitiveKeyKind` so to not pay the cost of large enum variants for small, common primitives.

- [x] LineDecoration
- [x] NormalBorder
- [x] ImageBorder
- [ ] Rectangle
- [x] YuvImage
- [x] Image
- [x] LinearGradient
- [x] RadialGradient
- [x] TextRun
- [x] Picture

PR #3393, opened 2018-12-07
Separate interning Image & YuvImage

djg:
In support of #3385, extract Image and YUVImage. This patch is based up PR #3392 and requires that to land first.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3393)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -551,10 +551,10 @@ impl AlphaBatchBuilder {
             render_tasks,
         ).unwrap_or(OPAQUE_TASK_ADDRESS);
 
-        let prim_data = &ctx.resources.as_common_data(&prim_instance);
+        let prim_common_data = &ctx.resources.as_common_data(&prim_instance);
         let prim_rect = LayoutRect::new(
             prim_instance.prim_origin,
-            prim_data.prim_size,
+            prim_common_data.prim_size,
         );
 
         match prim_instance.kind {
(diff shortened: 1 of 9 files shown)


djg:
Try & talos: https://treeherder.mozilla.org/#/jobs?repo=try&revision=27cbd66f358a757efa654b7b6804e95610b207a0

gw3583:
Could you rebase this to remove the patches that have landed now?

djg:
> Could you rebase this to remove the patches that have landed now?

Done

djg:
It appears that the `osx-release` taskcluster job failed to start. I've restarted it.

gw3583:
That last commit seems a bit more complex (esp. the ```cache``` fn) than I was thinking it would be, but I haven't looked in detail yet - I might just be missing something.

The general idea of the GpuCacheRequest being a by-value type is that it should work around the borrowck issues that way. I'll pull the branch today and see if I can provide some more useful comments than the vague details above, sorry!

djg:
> That last commit seems a bit more complex (esp. the `cache` fn) than I was thinking it would be, but I haven't looked in detail yet - I might just be missing something.
> 
> The general idea of the GpuCacheRequest being a by-value type is that it should work around the borrowck issues that way. I'll pull the branch today and see if I can provide some more useful comments than the vague details above, sorry!

@gw3583, in the previous incarnation of the code there was a split borrow of the prim fields and the borrowck was happy. It's unhappy with:

```
if Some(foo) = gpu_cache.bar(&self.common.gpu_handle) {
    ...
    self.write_prim_gpu_blocks(...);
    ...
}
```

because it can't see that `fn write_prim_gpu_blocks()` doesn't access `self.common`.

I welcome a second set of 👀 on the code.

gw3583:
This is roughly what I was thinking [1] - does that seem like it should achieve the same outcome and work for the rest of the prim types?

[1] https://github.com/gw3583/webrender/commit/a6910bceb599458169fab746f1ef5a3fad91c5df

djg:
@gw3583 Are we good to proceed with these changes?

gw3583:
Yup, I think so - do we need another try / talos run or should this be good to go?

djg:
Try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=5b13864afb0efce11588f4427603cc5c73cbf31c

PR #3395, opened 2018-12-09
Store glyphs in a normalized local space.

gw3583:
This prevents text runs being interned more often than they
should be when scrolling in Gecko.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3395)


gw3583:
r? @kvark or anyone

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d3787131d06be879f1fbe6a542e17766be1c6d30

PR #3396, opened 2018-12-09
Mark picture cached tiles as opaque.

gw3583:
We know that all picture cache tiles are opaque, for now, due to
the background color of the content frame being opaque. This
gives a good performance win by putting these tiles into the
opaque pass.

If we ever decide we want to support non-opaque tiles, we can
add the logic to handle this correctly at that time, making use
of the segment batching logic to draw the cached tiles.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3396)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1040,7 +1040,7 @@ impl AlphaBatchBuilder {
 
                                         let key = BatchKey::new(
                                             kind,
-                                            non_segmented_blend_mode,
+                                            BlendMode::None,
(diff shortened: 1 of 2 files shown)


staktrace:
sccache failure, https://github.com/mozilla/sccache/issues/344

@bors-servo retry

staktrace:
@bors-servo retry

If this keeps happening I might just disable sccache because it's hurting more than helping right now.

staktrace:
@bors-servo retry

Landed sccache disabling in #3402 so it should work better now.

PR #3397, opened 2018-12-09
Add a trait for internable types that helps debugging.

gw3583:
This allows selective logging of interning operations based
on specifics of the primitive or clip.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3397)


gw3583:
@djg I rebased this and updated to import the type directly to be consistent with the other types.

PR #3400, opened 2018-12-12
More CI script refactoring/cleanup

staktrace:
Mostly Windows test commands getting extracted, but also one tweak for the docker image to ensure the docker setup is more robust.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3400)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,3 +17,4 @@ build: false
 
 test_script:
   - cmd.exe /c ci-scripts\windows-tests.cmd
+  - cmd.exe /c ci-scripts\windows-pathfinder.cmd
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,15 +16,4 @@ install:
 build: false
 
 test_script:
-  - cd webrender_api
-  - cargo test --verbose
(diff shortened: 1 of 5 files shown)


appveyor.yml
```diff
@@ -17,3 +17,7 @@ build: false
 
 test_script:
   - cmd.exe /c ci-scripts\windows-tests.cmd
+  # Pathfinder builds servo-freetype-sys as a dependency which has additional
+  # build requirements. So we do it in a separate .cmd file that we skip over
+  # in Firefox CI for now.
```
staktrace: I'll move this comment to the commit message for the patch. I already have a solution in the works that will let us run the pathfinder build in Firefox CI but it may take an extra day or two to propagate the necessary patches from upstream repos. We won't need to touch this file again when it starts working in Firefox CI so this comment will be obsolete at that point.

staktrace:
> Looks good, though my cmd-foo is not the best, if it does what it seems it does it looks fine :-)

It should :) I tested some failure cases as well to make sure it errors out properly.

@bors-servo r=emilio

PR #3401, opened 2018-12-12
spatial-node: Flatten transforms properly.

### LINKED ISSUE #3394: Flatten the transforms of non-preserve-3d contexts
When a transform (or perspective) is specified on on a reference frame for a stacking context that's not `Preserve3D`, we are currently still applying the transform unchanged and ignore the Z component. The problem case arises when there are multiple 3D transforms stacked up, since we'd innocently multiply the transformation matrices, and Z coordinates would start to affect the positioning, which is undesirable for `Flat` style contexts.

What we should be doing is erasing/resetting the Z row of the transformation if its style is `Flat`.

cc @mattwoodrow @emilio

emilio:
This fixes Gecko bug 1512537.

Fixes #3394

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3401)


emilio:
r? @kvark 

Try is https://treeherder.mozilla.org/#/jobs?repo=try&revision=7220781e3ee3438ab0ec5dab5e501c262b00f0f4

staktrace:
> Looks like the taskcluster failure may be a flake:

Yeah that's https://github.com/mozilla/sccache/issues/344

PR #3403, opened 2018-12-12
Set SCCACHE_IDLE_TIMEOUT=0 when starting the sccache server

luser:
This will prevent the sccache server from shutting down due to a lack of activity which can manifest as build failures. This is effectively a workaround for this sccache issue: https://github.com/mozilla/sccache/issues/204 .

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3403)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -133,7 +133,7 @@ tasks:
             servo-tidy
             sccache --stop-server || true
             mkdir -p ../artifacts
-            RUST_LOG=sccache=trace SCCACHE_ERROR_LOG=$PWD/../artifacts/sccache.log sccache --start-server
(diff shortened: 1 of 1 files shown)


staktrace:
Thanks, this is good to know, but we're also still hitting https://github.com/servo/webrender/issues/3248#issuecomment-443070317 so at the moment I'd still like to leave sccache disabled until it's more reliable.

(Also note that the .taskcluster.yml file has two different jobs which use sccache so both of them would need this fix applied.)

gw3583:
@staktrace @luser Do we want to merge this (after a rebase) or should this be closed?

staktrace:
Unless we know how to resolve https://github.com/servo/webrender/issues/3248#issuecomment-443070317 I don't think it's worth re-enabling sccache. So I'd lean towards closing this for now.

gw3583:
OK, feel free to reopen if we want to reconsider landing this.

PR #3404, opened 2018-12-12
Fix handling of zero regions by texture allocator

kvark:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1513017

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3404)

### DIFF
--- a/webrender/src/texture_allocator.rs
+++ b/webrender/src/texture_allocator.rs
@@ -35,7 +35,7 @@ impl FreeListBin {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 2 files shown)


kvark:
@gw3583 ideally, yes. I suggest we continue the investigation of the actual cause after we put this fire down ;) Speaking of which, this version is confirmed to fix the bug, it's ready for the review.

kvark:
Pending Gecko try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=a571d97ed59e0341e27380c378b9b33353be7cb1

kvark:
One trivial fixup (derive PartialEq for a slice index)
@bors-servo r=gw3583

PR #3405, opened 2018-12-12
Optimize the allocation of rasterized blobs a bit

kvark:
Follow-up to #3398
Aimed to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1513521
Pending try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=26ea31fd69e8ed47167f33afefbd711424f0250a

I think it's harmless either way, waiting to see if talos results are back to normal.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3405)

### DIFF
--- a/webrender/src/scene_builder.rs
+++ b/webrender/src/scene_builder.rs
@@ -75,8 +75,13 @@ impl Transaction {
 
     fn rasterize_blobs(&mut self, is_low_priority: bool) {
         if let Some((ref mut rasterizer, _)) = self.blob_rasterizer {
-            let rasterized_blobs = rasterizer.rasterize(&self.blob_requests, is_low_priority);
-            self.rasterized_blobs.extend(rasterized_blobs);
+            let mut rasterized_blobs = rasterizer.rasterize(&self.blob_requests, is_low_priority);
+            // try using the existing allocation if our current list is empty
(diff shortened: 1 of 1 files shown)


kvark:
Try is green, talos shows some good wins: https://treeherder.mozilla.org/perf.html#/comparesubtest?framework=1&newProject=try&newRevision=26ea31fd69e8ed47167f33afefbd711424f0250a&newSignature=6e4abd046a0f519e61e3bf63050d34467c391e19&originalProject=try&originalRevision=a571d97ed59e0341e27380c378b9b33353be7cb1&originalSignature=6e4abd046a0f519e61e3bf63050d34467c391e19

gw3583:
r=me, but can we add a comment explaining why the code is faster this way than using extend, so that someone doesn't accidentally regress it?

kvark:
@gw3583 my understanding is that `append` versus `extend` here is rather secondary, we'll only hit it upon getting the "missing" rasterized blobs, which as a concept should go away soon (I hope!). The win here is from just taking existing vec if ours is not yet initialized. And that's pretty obvious in the code.

kvark:
Added a comment as requested.
@bors-servo r=gw3583

PR #3407, opened 2018-12-13
Support picture caching where content rect moves between display lists.

gw3583:
Gecko bakes scroll offsets into new display lists. Consider the
following example:

(a) WR receives a new display list.
(b) Scroll the display list, generating frames but not a new scene.
(c) Generate a new scene, due to hover / click etc.

In this case, the local coordinates of the primitives (and thus the
enclosing content rect) will be different between (a) and (c), due to
the current scroll offsets being baked into the local primitive rects.

Since this is non-trivial to change in Gecko, we need to ensure that
picture caching handles this case correctly.

There are two parts to this:

(a) Interned primitives must be stored and keyed in a true primitive
    local space, removing the local origin. Patches for this have
    already landed, and provide an optimization in cases where the
    same primitive template is instanced multiple times.
(b) Correctly create cache tiles even if the enclosing content rect
    has different coordinates. This is addressed by this patch.

We need to ensure that when generating tiles for a given content rect
the number of tiles (and local offsets) is the same for identical content,
regardless of the location of the content rect. To do this, we need to know
the local bounding rect of the enclosing content, and make tiles placed
relative to that. This moves the tile cache update pass to be done
*after* the initial picture traversal, so we know the content bounding
rect.

This also means we can reconfigure the tile grid once during
update_transforms, rather than checking after adding each primitive
whether we need to re-allocate the tile grid.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3407)

### DIFF
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -115,6 +115,8 @@ impl Example for App {
     }
 
     fn on_event(&mut self, win_event: winit::WindowEvent, api: &RenderApi, document_id: DocumentId) -> bool {
+        let mut rebuild_display_list = false;
+
         match win_event {
             winit::WindowEvent::KeyboardInput {
                 input: winit::KeyboardInput {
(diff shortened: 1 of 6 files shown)


gw3583:
There's still a few glitches and incorrect invalidations (which I'll be working on tomorrow), and there seems to be some GPU stalls due to how often tiles get added / removed from the texture cache (got a patch in progress to fix / work around this). Apart from that it seems to be working quite well now.

gw3583:
@kvark Thanks, comments addressed.

PR #3412, opened 2018-12-13
fall back to aliased glyph rasterization if cleartype rasterization fails

lsalzman:
This resolves Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1492443

ClearType rasterization may fail with various fonts above a certain glyph size. Skia addressed this in two ways. First, it had a much lower glyph size limit compared to us after which it switched over to drawing text as paths - 256 vs. our 1024. Secondly, if ClearType rasterization failed, it would switched over to aliased glyph rasterization which seemingly succeeds even in the cases that would otherwise fail.

So to bring us more in line with Skia, I have dropped our size limit down to 512, which should put us below the point where most fonts seem to fail, without making large glyphs too blurry. Secondly, I added the aliased glyph rasterization fallback as noted above, which even if rasterization somehow fails even below the 512 size limit (which it did not in my testing), should still let us show something.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3412)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -362,7 +362,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "dwrote"
-version = "0.6.2"
+version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "lazy_static 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


gw3583:
The change looks reasonable. @lsalzman assures me it doesn't need a try run :crossed_fingers: 

@bors-servo r+

PR #3413, opened 2018-12-13
Picture caching fixes and clean ups.

gw3583:
From the commits:

```
Store primitive instance origins in the tile descriptors.

This is necessary to uniquely identify a primitive instance, now
that we store primitive templates in a primitive-local space.
```

```
Switch default tile size to 1024 x 256.

This seems better suited for most pages I've tested on so far.

It also ensures that cache tiles are allocated as standalone
textures, rather than layers inside the texture cache. The current
texture cache grow / shrink logic was not handling the picture
cache allocation strategy very well, resulting in a lot of GPU
spikes as the texture cache arrays were being resized.

In future, we might consider a specialized cache texture for these
tiles that is, for example, 4096 x 4096. This would allow better
batching of draw calls for tiles when drawing the final scene.
```

```
Tidy up how tile rect coords are calculated. 
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3413)

### DIFF
--- a/webrender/src/display_list_flattener.rs
+++ b/webrender/src/display_list_flattener.rs
@@ -156,17 +156,6 @@ pub struct DisplayListFlattener<'a> {
     /// The root picture index for this flattener. This is the picture
     /// to start the culling phase from.
     pub root_pic_index: PictureIndex,
-
-    /// TODO(gw): This is a complete hack that relies on knowledge of
-    ///           what the Gecko display list looks like. It's used
-    ///           for now to work out which scroll root to use to
-    ///           create the picture cache for the content. It's only
-    ///           ever used if picture caching is enabled in the
-    ///           RendererOptions struct. We will need to work out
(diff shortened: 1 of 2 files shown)


gw3583:
Added a follow up commit:

```
Improve how display list flattening defines tile cache pictures.

This is both a bug fix (when opening devtools with picture caching
enabled) and also a much less hacky way to define where tile
cache picture(s) exist (although still not ideal).

Now, we only create a tile cache for top level (content) iframes.

When creating the tile cache for an iframe, skip any preceding
or trailing primitives that are not part of the main scroll root.
```

gw3583:
(merging now, assuming that the tick of approval means you're not too concerned about those nits).

@bors-servo r=kvark

bholley:
This improved displaylist_mutate: https://bugzilla.mozilla.org/show_bug.cgi?id=1514180#c4

PR #3417, opened 2018-12-14
Clear clip_node_instances on new frame builds.

### LINKED ISSUE #3416: ClipStore.clip_node_instances grows without bound in some cases
Start firefox
Load https://mozilla.staktrace.com/anim.html
Open a new window and load about:memory. Keep the anim.html window in the background
Periodically hit the measure button on about:memory
Watch the clip-stores memory reporter grow continuously

Switching to the anim.html window or otherwise triggering a repaint on it will free the accumulated memory. I suspect that [this code](https://searchfox.org/mozilla-central/rev/fd32b3a6fa3eff1468311f6fcf32b45c117136df/gfx/wr/webrender/src/clip.rs#678) gets run per generated frame for a given scene, which causes `clip_node_instances` to keep growing until a new scene build happens.

staktrace:
The clip node instances are rebuilt for each frame built from a given
scene, but are not cleared until a new scene is built. So if a given
scene builds many frames, the clip node instances just accumulate and
eat up memory.

Fixes #3416.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3417)

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -696,6 +696,11 @@ impl ClipStore {
         })
     }
 
+    pub fn clear_old_instances(&mut self) {
+        self.clip_node_instances.clear();
+    }
+
+
     /// Walk the clip chain of a primitive, and calculate a minimal
     /// local clip rect for the primitive.
     #[allow(dead_code)]
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -407,6 +407,7 @@ impl FrameBuilder {
             scene_properties,
             Some(&mut transform_palette),
         );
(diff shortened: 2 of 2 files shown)


staktrace:
This fixes the issue for me locally but I don't know if it breaks correctness elsewhere

staktrace:
Try push in progress at https://treeherder.mozilla.org/#/jobs?repo=try&group_state=expanded&revision=a197b27a72f4e5ac3dc9799a0e0518ecd232c951

staktrace:
Try push is green. r? @kvark

kvark:
:+1: 
@bors-servo r+

PR #3421, opened 2018-12-14
batch: Use the right clip rect for drop shadows.

emilio:
We don't really want to translate the clip, and furthermore, the picture clip
rect for these pictures is always the max rect, so it's not really what we want.

  https://searchfox.org/mozilla-central/rev/49e78df13e7a505827a3a86daae9efdf827133c6/gfx/wr/webrender/src/display_list_flattener.rs#1472

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1496019

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3421)


emilio:
s/ @glennw / @gw3583 there :)

emilio:
Pending try run is https://treeherder.mozilla.org/#/jobs?repo=try&revision=498dbdd146036d29d60b39e254f72d055dea725d

emilio:
... which looks green, with an unexpected pass even.

PR #3424, opened 2018-12-17
Fix a picture caching race condition.

gw3583:
Sometimes, the render backend produces frames faster than they
can be drawn by the renderer. The renderer has a code path for
this situation that drops frames in the render queue if there
are more recent frames to draw.

However, if the frame contains render tasks that will be persisted
in the texture cache, that frame must be executed in order to
ensure the texture cache is updated.

This patch extends that check to also consider render tasks that
will be blitted to the texture cache as a result of picture
caching.

This case occurs very rarely in real world usage, but _does_
occur quite commonly when running reftests on CI where the
time spent building simple frames is often much less than the
time for the software rasterizer to draw them.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3424)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -490,8 +490,9 @@ impl FrameBuilder {
                 &mut z_generator,
             );
 
-            if let RenderPassKind::OffScreen { ref texture_cache, .. } = pass.kind {
+            if let RenderPassKind::OffScreen { ref texture_cache, ref color, .. } = pass.kind {
                 has_texture_cache_tasks |= !texture_cache.is_empty();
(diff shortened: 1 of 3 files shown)


gw3583:
Added a follow up commit that fixes an issue where opacity bindings on pictures were not being included in a tile's dependencies.

kvark:
Leaving in case you want to launch a try ;) Feel free to proceed when ready please.

gw3583:
I don't *think* it'll be an issue in practice, but it's something to keep an eye out for certainly. Thanks for the quick review - try shouldn't be necessary since this still only affects the non-default render path for now.

@bors-servo r=kvark

PR #3425, opened 2018-12-17
make SubpixelOffset::quantize less fragile with unexpected values

lsalzman:
This resolves Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1509229

When we encounter unexpected values due to floating point precision issues, we can crash. Rather than try to belabor things trying to do anything fancy, it makes most sense to just zero out the subpixel offset in those cases rather than leave the code brittle with crashes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3425)


kvark:
I'm curious for what inputs did we previously get anything outside of 0..8 range?

PR #3426, opened 2018-12-17
Include picture composite mode in primitive key for pictures.

gw3583:
We need to ensure that a display list that is otherwise the same
except for a different compositing mode is detected by picture
caching code as invalid.

For example, if a picture has brightness(0.5) and then the next
display list is exactly the same but with brightness(0.6).

This patch adds a not particularly elegant key for picture
composite modes that is stored in the interned primitive data
for a picture, ensuring these cases are correctly detected.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3426)

### DIFF
--- a/webrender/src/prim_store/mod.rs
+++ b/webrender/src/prim_store/mod.rs
@@ -4,8 +4,8 @@
 
 use api::{AlphaType, BorderRadius, ClipMode, ColorF, PictureRect, ColorU, LayoutVector2D};
 use api::{DeviceIntRect, DeviceIntSize, DevicePixelScale, DeviceRect, LayoutSideOffsetsAu};
-use api::{FilterOp, ImageKey, ImageRendering, TileOffset, RepeatMode};
-use api::{LayoutPoint, LayoutRect, LayoutSideOffsets, LayoutSize};
+use api::{FilterOp, ImageKey, ImageRendering, TileOffset, RepeatMode, MixBlendMode};
+use api::{LayoutPoint, LayoutRect, LayoutSideOffsets, LayoutSize, PropertyBindingId};
 use api::{PremultipliedColorF, PropertyBinding, Shadow, YuvColorSpace, YuvFormat};
 use api::{DeviceIntSideOffsets, WorldPixel, BoxShadowClipMode, NormalBorder, WorldRect, LayoutToWorldScale};
 use api::{PicturePixel, RasterPixel, ColorDepth, LineStyle, LineOrientation, LayoutSizeAu, AuHelpers};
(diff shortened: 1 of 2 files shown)


gw3583:
Try run looks good.

gw3583:
I do think it's something worth considering in the future as we make the caching support more granular.

kvark:
What is the motivation for caching the post-composition contents?

gw3583:
The point of this change is so that a nested child picture inside a tile cache which had a composite mode is correctly detected as invalidating that tile when its parameters change. For example, imagine you have a picture / tile cache, and that has [image, opacity [image], rectangle] - we need to know that the opacity filter has changed to redraw that tile, just as we need to know if any primitive / clip has changed for that cached surface.

webrender/src/prim_store/mod.rs
```diff
@@ -342,17 +342,122 @@ pub struct PrimitiveSceneData {
...
+    Color,
+    Luminosity,
+}
+
+impl From<Option<PictureCompositeMode>> for PictureCompositeKey {
+    fn from(mode: Option<PictureCompositeMode>) -> Self {
+        match mode {
+            Some(PictureCompositeMode::MixBlend(mode)) => {
+                match mode {
+                    MixBlendMode::Normal => PictureCompositeKey::Identity,
```
kvark: why don't we include a variant like `MixBlend(MixBlendMode)` into `PictureCompositeKey` instead of duplicating the variants?

webrender/src/prim_store/mod.rs
```diff
@@ -342,17 +342,122 @@ pub struct PrimitiveSceneData {
...
+                    MixBlendMode::Exclusion => PictureCompositeKey::Exclusion,
+                    MixBlendMode::Hue => PictureCompositeKey::Hue,
+                    MixBlendMode::Saturation => PictureCompositeKey::Saturation,
+                    MixBlendMode::Color => PictureCompositeKey::Color,
+                    MixBlendMode::Luminosity => PictureCompositeKey::Luminosity,
+                }
+            }
+            Some(PictureCompositeMode::Filter(op)) => {
+                match op {
+                    FilterOp::Blur(value) => PictureCompositeKey::Blur(Au::from_f32_px(value)),
```
kvark: similarly, it looks like we should turn `FilterOp<T>` to be generic and just have a variant like `Filter(FilterOp<Au>)` in the composite key

webrender/src/prim_store/mod.rs
gw3583: Duh, I should have done that! Will do next time I'm in this area.

webrender/src/prim_store/mod.rs
gw3583: I wasn't sure if that would break the public API / bindgen bindings, so I tried to leave the public API as is.

PR #3427, opened 2018-12-17
Properly handle nested sticky nodes

staktrace:
It's possible to have a nested sticky node (i.e. a scrollframe that
contains a sticky node descendant which contains another sticky node
descendant). In this case, we want the scroll offset from the outer
sticky node to be used in the computation for the inner sticky node in
order to produce correct behaviour.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1495962, which
includes Gecko reftests that exercise this scenario.

r? @kvark or @emilio

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3427)

### DIFF
--- a/webrender/src/spatial_node.rs
+++ b/webrender/src/spatial_node.rs
@@ -481,6 +481,9 @@ impl SpatialNode {
                 // offsets actually adjust the node position itself, whereas scroll offsets
                 // only apply to contents inside the node.
                 state.parent_accumulated_scroll_offset += info.current_offset;
+                // We want nested sticky items to take into account the shift
+                // we applied as well.
+                state.nearest_scrolling_ancestor_offset += info.current_offset;
             }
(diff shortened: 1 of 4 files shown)


staktrace:
I posted a try link in bug 1495962 which includes the new reftests. Here it is for posterity:

https://treeherder.mozilla.org/#/jobs?repo=try&group_state=expanded&revision=9248c1317a942a91216c4cfde0bc58f71a71f85b

kvark:
Could you port one of them as a reftest here please?

staktrace:
Ok, will do

staktrace:
Updated patch to include a reftest. I didn't "port" the gecko ones but just wrote one from scratch. Verified it fails without the patch and passes with the patch.

staktrace:
(Sorry, I accidentally left a copy-paste comment in the new reftest. Removed now)

@bors-servo r=kvark

PR #3429, opened 2018-12-18
Separate interning borders

djg:
In support of #3385, extract normal and image borders. This patch is based up PR #3393 and requires that to land first.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3429)


gw3583:
#3393 has landed now, so it should be possible to rebase this and remove those commits.

gw3583:
One question - can either ignore or do as a follow up, if that's easier.

Apart from that, looks good to me - do we want a try / talos run before merging? (I suspect this might have a decent perf win on dl_mutate, given the reduction in primitive key size).

@bors-servo delegate+

djg:
I have try / talos scheduled. You beat me to it: https://treeherder.mozilla.org/#/jobs?repo=try&revision=0ab2321c567ca56aa43f1614d7dccf94a816d62e

PR #3430, opened 2018-12-18
Fix a couple of picture caching reftest issues with subpx offsets.

gw3583:


gw3583:
Fix a couple of subtle reftest issues that picture caching was hitting.

r? @kvark

PR #3432, opened 2018-12-18
ensure lcd filter padding is applied to cbox when rasterizing glyphs with FreeType

lsalzman:
This resolves Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1508163

We don't apply the lcd filter padding when querying the cbox to rasterize the glyph, even though we apply the padding when querying the bounds. This causes subpixel AA text to get shifted by the amount of the padding due to the padding not being accounted for in the offset.

The solution is to apply the padding in both cases so it properly cancels out.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3432)


lsalzman:
Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=134623e2d9a58e9a2bc525184240f1efe707671e

PR #3433, opened 2018-12-18
Remove redundant cluster range, as an optimization.

gw3583:
Initially I was planning to have primitives belong to multiple
clusters as part of the picture caching work, however this turned
out to not be required.

This patch removes support for that, which reduces the size of
PrimitiveInstance, and removes the prim cluster map from the
PrimitiveList.

Also reduce the size of the clip task index field as an additional
small struct size optimization.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3433)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -31,7 +31,7 @@ use scene::{FilterOpHelpers, SceneProperties};
 use scene_builder::DocumentResources;
 use smallvec::SmallVec;
 use surface::{SurfaceDescriptor, TransformKey};
-use std::{mem, ops};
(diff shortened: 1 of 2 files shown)


gw3583:
r? @kvark or @nical or anyone

Try run looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=6c3fd05236a16c8832b3870cdc2c4a541ec3b091&selectedJob=217618325

gw3583:
Rebased to fix the conflict around some ```use``` statements.

@bors-servo r=kvark

PR #3434, opened 2018-12-18
More version pinning of docker packages

staktrace:
Installing with pip is also unsafe in that it will just use the latest
version and so the resulting behaviour will be dependent on when the
docker image is built. This switches to use debian packages instead of
pip where possible, and locks a particular version for the remaining
python package.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3434)

### DIFF
--- a/ci-scripts/docker-image/setup.sh
+++ b/ci-scripts/docker-image/setup.sh
@@ -25,28 +25,22 @@ apt-get install -y \
     libx11-dev \
     pkg-config \
     python \
+    python-mako \
     python-pip \
+    python-setuptools \
+    python-voluptuous \
+    python-yaml \
     software-properties-common
 
-# Build freetype with subpixel rendering enabled
-curl -sSfL -o ft.tar.bz2 \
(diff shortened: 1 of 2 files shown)


staktrace:
Updated patch addresses @glandium's comments from https://phabricator.services.mozilla.com/D14405. I'll merge it now to avoid the extra timezone-induced latency and I've asked him to do a post-review of the updated patch.

ci-scripts/docker-image/setup.sh
```diff
@@ -25,28 +25,22 @@ apt-get install -y \
...
-    include/freetype/config/ftoption.h
-./configure
-make
-make install
-
-# Replace the system libfreetype with the one we just built
-cd /usr/lib/x86_64-linux-gnu/
-rm -f libfreetype.so.6
-ln -s /usr/local/lib/libfreetype.so.6
+# Get freetype 2.8 with subpixel rendering enabled. The SNAPSHOT_ARCHIVE
```
glandium: Maybe add a comment why 2.8, so that someone else doesn't have to rediscover why next time :)

PR #3436, opened 2018-12-18
Fix out of bounds access of clip tasks in rare conditions.

gw3583:
In rare cases, a picture primitive may have a surface (e.g. due
to opacity) on one frame, and then become a pass-through primitive
on the next frame (e.g. opacity is animated to 1).

In these cases, it was possible that the primitive instance for
the picture has a clip task index set from the previous frame
that is not correctly reset. This can cause an out of bounds
access of the scratch buffer clip instances array during
batching.

To ensure this can't happen, reset the clip task index for the
primitive instance to invalid as early as possible (at the same
time the primitive instance visibility is reset).

### DIFF
--- a/webrender/src/prim_store/mod.rs
+++ b/webrender/src/prim_store/mod.rs
@@ -1591,6 +1591,12 @@ impl PrimitiveInstance {
         }
     }
 
+    // Reset any pre-frame state for this primitive.
+    pub fn reset(&mut self) {
+        self.bounding_rect = None;
+        self.clip_task_index = ClipTaskIndex::INVALID;
+    }
+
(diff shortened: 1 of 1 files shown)


gw3583:
r? @kvark or anyone

Pending try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=44cdc2846ae50f93952f187efd702eac7a63b3e9

This is a fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1507017

gw3583:
Try run looks good.

kvark:
We should also make those clip task indices to be generational under debug, so that we can panic on using a stale index.
@bors-servo r+

PR #3438, opened 2018-12-19
Reduce size of spatial node index

### LINKED ISSUE #3431: Reduce the size of SpatialNodeIndex
It's currently a ```usize``` and is embedded inside ```PrimitiveInstance```. We could safely reduce this to a ```u32``` which would be a noticeable performance optimization for test cases with large numbers of primitive instances.

djg:
Fixes #3431.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3438)


zptan:
@kvark r?

djg:
I've re-written the patch to not use type safe index nor the `Vec` wrapper.

PR #3439, opened 2018-12-20
Optimize pop_stacking_context.

djg:
Profiling showed that the `parent_sc.primitives.extend(...)` was a hot spot in
scene building.  Further investigation showed that `parent_sc.primitives` is empty
a majority of the time, so copy the storage instead of copying all the elements.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3439)


djg:
An easy review, @gw3583

PR #3442, opened 2018-12-21
add support for FreeType variation fonts

lsalzman:
This resolves Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1442693

Linux/Android is the last platform we don't support variation fonts on. This fixes that.

Some dlsym hackery is required because most older versions of FreeType don't actually implement the necessary functions, while on Android we assume we are statically linking against a recent FreeType.


lsalzman:
Try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e84b52254bd8a9ea7b7d2b1b1554290da51b9f14

lsalzman:
Fixed Android try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=4116f15ac77a8b367fee216fa23349280e7ad2fc

webrender/src/platform/unix/font.rs
```diff
@@ -38,16 +44,141 @@ const FT_LOAD_TARGET_MONO: FT_UInt   = 2 << 16;
...
+struct VariationFace(FT_Face);
+
+impl Drop for VariationFace {
+    fn drop(&mut self) {
+        unsafe { FT_Done_Face(self.0) };
+    }
+}
+
+fn new_ft_face(font_key: &FontKey, lib: FT_Library, file: &FontFile, index: u32) -> Option<FT_Face> {
+    unsafe {
```
jrmuizel: Maybe scope the 'unsafe' tighter?

lsalzman: Will maybe do that in a follow-up PR.

PR #3444, opened 2018-12-23
Render cached render tasks in a separate pass. 

emilio:
This prevents us from culling out the picture that contains that render task and
making other pictures that hit the cache and need it from finding that the task
was not drawn.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1503373, and similar issues with line decorations and such.

I think it also allows us to get rid of the `SpecialRenderPasses` that pathfinder uses, but I've left a FIXME instead for now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3444)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -435,6 +435,10 @@ impl FrameBuilder {
                                                        &mut render_tasks,
                                                        texture_cache_profile);
 
+        // TODO(emilio): Now that cached render tasks know how to create these
+        // extra passes, are the special render passes needed? i.e., does
+        // anything depend on the alpha -> color ordering? If not, seems we
+        // could just ditch them.
         let mut passes = vec![
             special_render_passes.alpha_glyph_pass,
             special_render_passes.color_glyph_pass,
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -440,7 +440,30 @@ impl FrameBuilder {
             special_render_passes.color_glyph_pass,
         ];
 
+        {
+            let passes_start = passes.len();
+            let mut required_pass_count = 0;
+            for cacheable_render_task in &render_tasks.cacheable_render_tasks {
+                render_tasks.max_depth(
+                    *cacheable_render_task,
+                    0,
+                    &mut required_pass_count,
+                );
+            }
+            for _ in 0 .. required_pass_count {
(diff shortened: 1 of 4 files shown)


emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=cd65b25e4fb3e1c76888dd3c0a907654d97842cf

emilio:
@gw3583 not sure if this approach makes sense to you, I think it's less complex than what I thought it would be, so I may be missing something.

emilio:
Try looks fine.

gw3583:
@emilio Nice work, that is indeed a bit simpler than I thought it would be too :+1: 

@bors-servo r+

kvark:
@emilio could you clarify this part please?
>This prevents us from culling out the picture that contains that render task and
making other pictures that hit the cache and need it from finding that the task
was not drawn.

emilio:
@kvark sure. Imagine we have two different surfaces, and both contain a border segment with a particular cache key (style / color / width / etc.).

When creating the render tasks for the first one, before this patch, we go through `request_render_task` and insert the render task in the current surface:

```
surfaces[surface_index.0].tasks.push(target_to_cache_task_id);
```

And the second one just hits the cache, so doesn't create a new render task.

But it may be the case that we end up not drawing that surface (for it being clipped out, or offscreen, or what not). In that case, the second surface will still use the texture reserved for that render task, but the render task hasn't ran! Which means that we'll just render whatever was in that GPU memory.

You can see the test-cases in https://bugzilla.mozilla.org/show_bug.cgi?id=1503373 for examples of this.

kvark:
@emilio thanks for explanation! It does make it clear what the issues is, and it's pretty severe.

I am concerned though about the fix a bit. Ideally, I'd think that two surfaces refer to the same task, and visibility of pictures is resolved before assigning tasks to passes. So the first picture wouldn't register this task dependency but the second one would, and everything would work good. At the same time, if both pictures are invisible, the task still doesn't get rendered.

With your fix, do we always draw all the cached tasks even if they aren't needed this frame?

emilio:
> With your fix, do we always draw all the cached tasks even if they aren't needed this frame?

Yes, though most of the time it'd be in the texture cache already, so I assumed it shouldn't be a huge deal.

kvark:
@emilio I see. My concern is if there is something changing every frame but not visible, and we end up re-rendering it to the cache anyway.

PR #3445, opened 2018-12-24
render_task: Assign passes in a single pass.

emilio:
Right now creating render passes is a three-step process:

 * You first go through the render task tree and figure out the max depth.
 * You allocate as many passes as needed.
 * You traverse the whole tree again assigning to those passes.

This patch makes it so that assign_to_passes is responsible to grow the passes
vector with offscreen passes lazily as needed, and its caller is responsible to
reverse the array so that the passes end up in order.

Thus so that we can build the pass list in a single pass (no pun intended) as we
traverse down the tree.

This also removes the SpecialRenderPasses because now that cached render tasks
know how to create passes as needed I'm pretty sure it's not needed. There's no
dependency between the two passes I could think of, looking at the code, and all
the tasks use color render targets, so I think they're unnecessary.

And in any case they should be easy to add back, should pathfinder really needed
them.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3445)

### DIFF
--- a/webrender/src/frame_builder.rs
+++ b/webrender/src/frame_builder.rs
@@ -27,7 +27,6 @@ use spatial_node::SpatialNode;
 use std::{f32, mem};
 use std::sync::Arc;
 use tiling::{Frame, RenderPass, RenderPassKind, RenderTargetContext};
-use tiling::{SpecialRenderPasses};
 
 
 #[derive(Clone, Copy, Debug, PartialEq)]
@@ -86,7 +85,6 @@ pub struct FrameBuildingState<'a> {
     pub clip_store: &'a mut ClipStore,
     pub resource_cache: &'a mut ResourceCache,
     pub gpu_cache: &'a mut GpuCache,
-    pub special_render_passes: &'a mut SpecialRenderPasses,
     pub transforms: &'a mut TransformPalette,
     pub segment_builder: SegmentBuilder,
     pub surfaces: &'a mut Vec<SurfaceInfo>,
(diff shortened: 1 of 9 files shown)


emilio:
Try is closed due to infra issues. I pushed regardless just in case it works so I don't need to do it in the morning:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=af56103f937386bda6d3189f30d4652399548a8f

Otherwise I'll push in the morning and ensure this is green before landing.

emilio:
(Assuming it's r+'d of course :P)

emilio:
Try looks good.

kvark:
> This also removes the SpecialRenderPasses because now that cached render tasks
know how to create passes as needed I'm pretty sure it's not needed.

I'm slightly confused by this (haven't gone through the code yet). The tasks should be creating passes in the first place, since passes are higher level concept.

>  There's no dependency between the two passes I could think of, looking at the code, and all
the tasks use color render targets, so I think they're unnecessary.

Please clarify this part. Things are separated into passes, conceptually, so that *with a pass* there are no dependencies. But dependencies between passes are naturally expected. Also, we do have alpha render targets (as opposed to color), and even pathfinder's special tasks have alpha, as far as I see.

emilio:
> I'm slightly confused by this (haven't gone through the code yet). The tasks should be creating passes in the first place, since passes are higher level concept.

So these passes were passes specific to pathfinder, where all the glyphs were rendered before-hand. This was, I think, working around the bug I fixed in #3444 (which is much more likely to occur with glyphs). Now the more generic code that I added in that PR should handle it, by creating passes as necessary.

> Please clarify this part. Things are separated into passes, conceptually, so that with a pass there are no dependencies. But dependencies between passes are naturally expected. Also, we do have alpha render targets (as opposed to color), and even pathfinder's special tasks have alpha, as far as I see.

Yes, I understand that. My point is that I don't think there's any dependency between `special_render_passes.alpha_glyph_pass` and `special_render_passes.color_glyph_pass`. And as such there shouldn't be any reason for their existence, they should just be merged with the rest of the `cacheable_render_tasks` passes.

kvark:
Thanks for clarification!
@bors-servo r+

emilio:
> shouldn't we leave unimplemented! here, given that the original passes no longer exist?

No, because this should use the more generic mechanism described above.

PR #3446, opened 2018-12-24
Make picture caching more performant, and fix some bugs.

gw3583:
One the remaining issues with enabling picture caching is that
it was a noticeable performance regression on sites that don't
benefit from picture caching, due to the extra render targets
and blits. Instead, we now draw to the main framebuffer and
blit those results into tiles for caching, as appropriate.

There is still a bit of work to be done here (specifically,
detecting when a tile isn't opaque, and disabling the cache blit
if we decide the tile is changing too often), but this is in
general a much faster solution than before. It's also faster
on pages that *do* benefit from picture caching, removing
one extra blit step during cache creation.

This also changes how we handle the way gecko supplies different
display lists with different local coordinate systems. Now, we
cache based on the world rect of the tile, and verify the content
with the new ComparableVec type. This is much more robust and
fixes a number of existing bugzilla bugs related to picture caching.

This patch also removes much of the hashing required, which is a
performance win on the sites I tested (even with the vec comparison
that we now use instead). We do currently update prim deps each
scroll frame, which can be improved / optimized in the future.

Finally, this implementation only allocates and stores tiles for
the current viewport, which is much better for performance and
scaling of content where the picture rect is very large.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3446)

### DIFF
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -41,10 +41,18 @@ impl App {
         color: ColorF,
         builder: &mut DisplayListBuilder,
         property_key: PropertyBindingKey<LayoutTransform>,
+        opacity_key: Option<PropertyBindingKey<f32>>,
     ) {
-        let filters = [
-            FilterOp::Opacity(PropertyBinding::Binding(self.opacity_key, self.opacity), self.opacity),
-        ];
+        let filters = match opacity_key {
+            Some(opacity_key) => {
+                vec![
+                    FilterOp::Opacity(PropertyBinding::Binding(opacity_key, self.opacity), self.opacity),
+                ]
+            }
+            None => {
+                vec![]
+            }
+        };
 
         let reference_frame_id = builder.push_reference_frame(
             &LayoutPrimitiveInfo::new(LayoutRect::new(bounds.origin, LayoutSize::zero())),
@@ -89,8 +97,8 @@ impl App {
 }
 
 impl Example for App {
-    const WIDTH: u32 = 1024;
-    const HEIGHT: u32 = 1024;
+    const WIDTH: u32 = 2048;
+    const HEIGHT: u32 = 1536;
 
     fn render(
         &mut self,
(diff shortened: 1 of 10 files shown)


gw3583:
This is not quite ready for merge yet - I need to tidy it up a bit and add a couple small remaining bits to the patch (mostly mentioned in TODO comments). But it makes picture caching much faster than it was, and fixes a number of visual glitches that are bugzilla bugs. It seems like a much more solid approach, even though the current implementation is quite messy.

kvark:
Could you clarify this part please?
> Now, we cache based on the world rect of the tile, and verify the content
with the new ComparableVec type.

gw3583:
Rebased and added a try run to make sure this doesn't cause any issues with existing code paths (looks good so far). I'm working on the remaining bits and pieces for this as a follow up commit.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=a5f13a55c62b101c120cfe3bbb07625e0df3bbd9&selectedJob=219010851

Darkspirit:
Login and click to open a tweet.

RUST_BACKTRACE=1 mozregression --repo try --launch a5f13a55c62b101c120cfe3bbb07625e0df3bbd9 --pref gfx.webrender.all:true gfx.webrender.picture-caching:true -B debug -a https://www.twitter.com

>  1:37.78 INFO: Hit MOZ_CRASH(bug: no world clip rect) at libcore/option.rs:1008

Darkspirit:
Still partly invisible:
mozregression --repo try --launch a5f13a55c62b101c120cfe3bbb07625e0df3bbd9 --pref gfx.webrender.all:true gfx.webrender.picture-caching:true -a https://bug1515540.bmoattachments.org/attachment.cgi?id=9032606

Darkspirit:
New regression first appeared in this try build: Zoom in on https://bugzilla.mozilla.org/enter_bug.cgi?product=Core&component=Graphics%3A%20WebRender so that you have horizontal and vertical scrollbars: They're flickering or are even invisible then.

gw3583:
@Darkspirit Thanks for taking a look!

I pushed an update which fixes the crash / panic you were hitting. The scrollbars (and I think other artifact) are expected until I land the follow up TODOs for detecting opaque tiles and splitting the batches based on this.

I think that this is fine to merge as long as it doesn't have any bad effect on the non-picture caching path, since they are a known issue that I'm working on now as a follow up.

The try run looks OK - none of those issues are ones that occur on the non-picture caching path, is that right?

Thanks!

Darkspirit:
Haven't seen other problems. Your latest commit doesn't have a try build, but we'll get one in wr-future-update and I'll test that too.

@bors-servo r+

PR #3447, opened 2018-12-24
Update sha2 to 0.8

Eijebong:
This will dedupe sha2 in m-c

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3447)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,11 +34,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arrayref"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "arrayvec"
 version = "0.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


emilio:
Err, right...

@bors-servo r+

staktrace:
@Eijebong While I'm not opposed to bumping dependencies I just wanted to point out that this does not dedup anything in m-c and in fact adds a bunch of new stuff.

https://hg.mozilla.org/try/rev/1968173d59dea55133ae1e82c769f09d73762686

Eijebong:
@staktrace Oh right, I did my checks with a fully updated lockfile. `cargo update -p lalrpop -p lalrpop-util` will update stuff to dedupe sha2.

staktrace:
Doing that also bumps the regex crate to a new version which includes a 100k+ test file, and requires build peer approval to land. So if you want to do that please do it as a separate bug. When I do the WR update I'm just going to take this PR without the lalrpop update.

Eijebong:
For reference https://bugzilla.mozilla.org/show_bug.cgi?id=1516337

PR #3452, opened 2018-12-28
Reduce GPU cache memory usage

bholley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1505449

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3452)

### DIFF
--- a/webrender/src/gpu_cache.rs
+++ b/webrender/src/gpu_cache.rs
@@ -132,7 +132,7 @@ impl GpuCacheHandle {
 // A unique address in the GPU cache. These are uploaded
 // as part of the primitive instances, to allow the vertex
 // shader to fetch the specific data.
-#[derive(Copy, Debug, Clone)]
+#[derive(Copy, Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(feature = "capture", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
 pub struct GpuCacheAddress {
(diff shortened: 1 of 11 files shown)


bholley:
Huh, weird. Windows Firefox try push looks green, and all the standalone reftests are green on mac and linux, but appveyor shows 56 failed reftests where we just render blank.

Trying to reproduce locally. In the mean time, I'll do a retry here just to be sure.

@bors-servo retry

@staktrace have you seen anything like this before?

staktrace:
@bholley No I haven't seen this before. I kicked off a [try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=c4eec6516d64ae6320f0e3169996039157656dd2) to see if this failure reproduces in the "standalone WR tests on windows in firefox CI" setup - although I don't know if that result will really help you much either way.

kvark:
@bholley let's also fire a PR for half of those commits to start bisection on CI?

bholley:
@kvark suggested it might be because we're creating zero-sized textures, but that's not the issue here (there's code above to pad to at least one row of data).

The issue seems to be that something breaks when the texture is too-tightly-scoped to the actual size of the data. I can make the problem go away by simply adding 1 to needed_height when passing it to create_texture (or just .max(2)-ing it).

So my theory is that there's a bug in the shaders that consume these textures that trigger an out-of-bounds access on the texture that doesn't affect rendering, but trips an error on whatever driver version our reftests get run on. Thoughts @kvark?

I just dropped that patch from the queue for now to get the rest landed. It's a significant memory win though, so I want to try to get it relanded.

kvark:
@bholley great to hear that you found the cause! Next thing would be to narrow down the search by finding out what texture exactly needs at least 2 of the height.

bholley:
This turned out to be https://bugs.chromium.org/p/angleproject/issues/detail?id=3039

kvark:
Good old Angle...

PR #3454, opened 2018-12-30
Fix picture caching on sites with multiple scroll roots.

gw3583:
For now, we simply disable picture caching when we encounter a
display list with content that has multiple scroll roots (other
than the root scroll node) - this still handles the majority of
sites. In future, we can expand this to support more cases.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3454)


gw3583:
r? @kvark 

This is a fix / workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1514490.

PR #3455, opened 2018-12-31
Follow ups from the recent picture caching optimizations.

gw3583:
This patch contains a number of fixes and improvements to
the picture caching changes that landed recently to improve
the general performance of picture caching. Specifically:

* Support setting scissor rect for the dirty rect that is
  being drawn. This is a performance win, and also simplifies
  the batching logic to not have to adjust the z_id for the
  tiles that are being drawn.
* Support alpha batch containers having more than one batch list.
  Each batch list supports an optional scissor rect, and a list
  of blits to run after drawing that batch list. This ensures
  that items batched after the cacheable content are not drawn
  into the cache tiles.
* The pending tile blits no longer need to be stored in the render
  task or surface information struct. Instead, they are retrieved
  directly from the picture's tile cache struct during batching.
* Include the local clip rect of the picture when drawing tiles,
  to ensure that the tiles don't write to the z-buffer where the
  scroll bars for a content frame is (they typically come earlier
  in the display list than the content, relying on clipping rather
  than render order).
* Include the tile relative position of clip vertices - this ensures
  that tiles are correctly invalidated in cases where only the
  relative position of a clip node changes between frames (there are
  a couple of reftests that verify this).
* Handle the case of a zero-sized clip mask correctly, to avoid
  trying to allocate a zero sized texture if only one clip task
  exists in a pass.
* Skip pushing / popping clip node collector for tile cache surfaces.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3455)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -20,7 +20,7 @@ use prim_store::{VisibleGradientTile, PrimitiveInstance, PrimitiveOpacity, Segme
 use prim_store::{BrushSegment, ClipMaskKind, ClipTaskIndex};
 use prim_store::image::ImageSource;
 use render_backend::FrameResources;
-use render_task::{RenderTaskAddress, RenderTaskId, RenderTaskTree};
+use render_task::{RenderTaskAddress, RenderTaskId, RenderTaskTree, TileBlit};
 use renderer::{BlendMode, ImageBufferKind, ShaderColorMode};
 use renderer::BLOCKS_PER_UV_RECT;
 use resource_cache::{CacheItem, GlyphFetchResult, ImageRequest, ResourceCache, ImageProperties};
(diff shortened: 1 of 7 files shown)


gw3583:
r? @kvark 

There is still one more known picture caching issue I'm working on locally, but these changes are reasonable to review / land while I continue work on that.

Try run (pic caching disabled) looks good:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=8e341ba75771ad6ba315e170e1be78575071a4b1

Try run (pic caching enabled) looks good (one new pass in R6):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4b0bb3b0faf4864ab1c34e5dde8edddc2802eb53

gw3583:
(some of these changes are a bit subtle - let me know if they don't make sense and we can discuss)

Darkspirit:
This fixes [bug 1516695](https://bugzilla.mozilla.org/show_bug.cgi?id=1516695) and [bug 1516786](https://bugzilla.mozilla.org/show_bug.cgi?id=1516786).

gw3583:
We want to get this landed as soon as possible, so I got @bholley to take a quick sanity check and we'll get @kvark to do a detailed follow up review next week. Merging now since both try runs above are green.

@bors-servo r=bholley

PR #3458, opened 2019-01-03
Remove dead code

staktrace:
The first patch here provides all the actual benefit. Not sure if you want the second one or not. I can drop it if it's undesirable.

r? @gw3583

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3458)


gw3583:
Seems reasonable, we can always retrieve this from the history if / when we make use of it for future optimizations, thanks!

@bors-servo r+

PR #3459, opened 2019-01-03
Fix picture caching on pages with fixed position clips.

gw3583:
Previously, the code used a clip node collector to build an
overall (clipped) bounding rect for the tile cache.

This is needed to ensure that the tiles don't write into the
z-buffer where primitives don't exist on the tiles, so that
primitives earlier in the render order (e.g. scroll bars)
still draw correctly in cases where the content overflows the
scroll region and is clipped.

The clip node collector code handles common cases, but it fails
when there are fixed position clips on small items within the
primitive instance list.

Instead, build the overall clipped world rect and per-primitive
required rects to handle partial tiles. This is quite a
conservative approach, for now. In future we should be able
to optimize these changes quite significantly.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1517117

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -23,7 +23,7 @@ use render_task::to_cache_size;
 use resource_cache::{ImageRequest, ResourceCache};
 use std::{cmp, u32};
 use std::os::raw::c_void;
-use util::{extract_inner_rect_safe, project_rect, ScaleOffset, MaxRect};
+use util::{extract_inner_rect_safe, project_rect, ScaleOffset};
 
 /*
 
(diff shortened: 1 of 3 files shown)


gw3583:
r? @bholley or @kvark

This fixes a number of picture caching glitches on various sites. It's a fairly conservative approach, but I'd rather get the correctness issues fixed and then apply more clever optimizations once enabled.

Pending try (caching disabled):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a4f08722b5be28da75a52578fbdefe3ddfbf6aaf

Pending try (caching enabled):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d8c9665e759bef2f042ef9218d6122dc68e59017

There are a couple of known intermittent failures on try with caching enabled. I know what these are caused by, and they are unrelated to this patch. So it should be fine to merge this after review, since (a) caching is still disabled by default and (b) I'm working on fixes for these now.

bholley:
@bors-servo r+

This is a rubber-stamp since kvark is out, he should give it a real review when he gets back.

Darkspirit:
As bugspam indicated, this also fixes [bug 1516839](https://bugzilla.mozilla.org/show_bug.cgi?id=1516839), [bug 1516753](https://bugzilla.mozilla.org/show_bug.cgi?id=1516753), [bug 1515540](https://bugzilla.mozilla.org/show_bug.cgi?id=1515540) and [bug 1512784](https://bugzilla.mozilla.org/show_bug.cgi?id=1512784).

gw3583:
@Darkspirit Thank you for re-testing all those bugs against this try! That was going to be my first job this morning :grinning:

PR #3460, opened 2019-01-03
Compute the layout space tiling origin.

nical:
The tiling origin is computed in image::tiles instead of being provided to the function.
In addition, the image rect in device space is exposed as function parameter. In a followup commit, callers will have to determine the correct image rect using the blob image's visible area.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3460)


gw3583:
I think we can close this now, and land any remaining parts via m-c.

PR #3461, opened 2019-01-03
[WIP] hook up the blob visible area in tile-decomposition

nical:
Store the blob visible areas in yet another hash map (this one being in the bag of post-scene-building state whereas `blob_image_templates` is pre-scene-building info), and pass it along to the tile decomposition code.

This PR needs #3460 and #3409.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3461)


gw3583:
I think we can close this now, and land any remaining parts via m-c.

PR #3463, opened 2019-01-04
Pre hash shaders at build time

bholley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1510491

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3463)


staktrace:
@bholley FYI try push is showing standalone CI failures because this PR didn't update the lockfile

staktrace:
> Hm, sorry about that. So I guess what happened is that those jobs were deployed to m-c more recently than my try push, and the WR CI jobs don't use this particular Cargo.lock file?

Those jobs were deployed to m-c on Dec 19, so somewhere after some of your try pushes but before other try pushes. Although you need to update your try syntax to actually select those jobs (syntax can be found [here](https://wiki.mozilla.org/Platform/GFX/Quantum_Render#Try_syntax)).

The WR CI jobs do use this Cargo.lock file, but they will just modify it in-place. On TreeHerder we run cargo test with `--frozen` to ensure it doesn't have network traffic during the tests, and that's what prevented the modification. (If we wanted to run WR/Github CI with `--frozen` we'd have to do some sort of vendoring thing to download all the crate dependencies without modifying the lockfile.. probably not worth the effort)

PR #3465, opened 2019-01-04
Skip evicting retained tiles that are currently not visible.

gw3583:
The code in pre_update selects a desired tile grid, based on the
currently visible world rect.

Any tiles retained after this point should try to retain their backing
texture, even if they are currently off-screen, since it's possible
they will be scrolled back into the visible rect soon.

Once it's determined they are far enough away from the visible
rect during pre_update, they will be dropped and evicted eagerly.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3465)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -100,6 +100,10 @@ pub struct OpacityBindingInfo {
     changed: bool,
 }
 
+/// A stable ID for a given tile, to help debugging.
+#[derive(Debug, Copy, Clone)]
+struct TileId(usize);
+
 /// Information about a cached tile.
 #[derive(Debug)]
 pub struct Tile {
@@ -118,11 +122,15 @@ pub struct Tile {
     /// cache handle can be used. Tiles are invalidated during the
     /// build_dirty_regions method.
     is_valid: bool,
+    /// The tile id is stable between display lists and / or frames,
+    /// if the tile is retained. Useful for debugging tile evictions.
+    id: TileId,
 }
 
 impl Tile {
     /// Construct a new, invalid tile.
     fn new(
+        id: TileId,
     ) -> Self {
         Tile {
             local_rect: LayoutRect::zero(),
@@ -131,6 +139,7 @@ impl Tile {
             handle: TextureCacheHandle::invalid(),
             descriptor: TileDescriptor::new(),
             is_valid: false,
+            id,
         }
     }
 
@@ -290,6 +299,8 @@ pub struct TileCache {
     /// scroll bars in gecko, when the content overflows under the
     /// scroll bar).
     world_bounding_rect: WorldRect,
+    /// Counter for the next id to assign for a new tile.
+    next_id: usize,
 }
 
 impl TileCache {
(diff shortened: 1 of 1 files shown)


webrender/src/picture.rs
```diff
@@ -520,8 +532,14 @@ impl TileCache {
                 let key = (px.round() as i32, py.round() as i32);
 
                 let mut tile = match old_tiles.remove(&key) {
-                    Some(tile) => tile,
-                    None => Tile::new(),
+                    Some(tile) => {
+                        tile
+                    }
```
bholley: This indentation is weird - does rustfmt require it? I'd think that this could still just be => tile, without braces.

webrender/src/picture.rs
```diff
@@ -118,11 +122,15 @@ pub struct Tile {
     /// cache handle can be used. Tiles are invalidated during the
     /// build_dirty_regions method.
     is_valid: bool,
+    /// The tile id is stable between display lists and / or frames,
+    /// if the tile is retained. Useful for debugging tile evictions.
+    id: TileId,
```
bholley: I assume there are few enough tiles in the system and they're moved infrequently enough that an extra word for debugging has negligible overhead?

webrender/src/picture.rs
```diff
@@ -520,8 +532,14 @@ impl TileCache {
...
                 let mut tile = match old_tiles.remove(&key) {
-                    Some(tile) => tile,
-                    None => Tile::new(),
+                    Some(tile) => {
+                        tile
+                    }
+                    None => {
+                        let id = TileId(self.next_id);
+                        self.next_id += 1;
+                        Tile::new(id)
```
bholley: I think it would be a bit cleaner to make next_id a method and then this could just be Tile::new(self.next_id()).

webrender/src/picture.rs
```diff
@@ -897,6 +915,20 @@ impl TileCache {
...
+            }
+
+            // Request the backing texture so it won't get evicted this frame.
+            // We specifically want to mark the tile texture as used, even
+            // if it's detected not visible below and skipped. This is because
+            // we maintain the set of tiles we care about based on visibility
+            // during pre_update. If a tile still exists after that, we are
+            // assuming that it's either visible or we want to retain it for
+            // a while in case it gets scrolled back onto screen soon.
+            resource_cache.texture_cache.request(&tile.handle, gpu_cache);
```
bholley: I think it would be clearer if this were in an else block, since it's a no-op in the already-evicted case (handled above), but that's not obvious to readers.

bholley: Also, it seems like what we really want here is EvictionPolicy::Manual, which would let us skip this work, and trigger faster eviction of the tiles we don't care about.

I won't insist you do this now, but add a TODO?

webrender/src/picture.rs
gw3583: Fixed.

webrender/src/picture.rs
gw3583: Yea, typically 10 - 20 tiles ever active.

webrender/src/picture.rs
gw3583: Done

PR #3467, opened 2019-01-04
Retain tiles just outside the viewport a bit longer.

gw3583:
Expand the size of the needed region that active tiles exist
over. This means that as tiles get scrolled just in / out of
the screen we retain them for a bit longer.

Since we only support a single dirty rect (for now), this is
helpful for avoiding the case of creating a large dirty rect
when something in the middle of the screen is animating while
scrolling. In future, once multiple dirty rects are supported,
it will be far less important to retain tiles outside the
viewport.

Also fix a subtle bug with the descriptor dependencies where
clipped out primitives would invalidate the tile due to changing
the length of the descriptor needed_rects array.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3467)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -113,6 +113,9 @@ pub struct Tile {
     pub local_rect: LayoutRect,
     /// The valid rect within this tile.
     valid_rect: WorldRect,
+    /// The currently visible rect within this tile, updated per frame.
+    /// If None, this tile is not currently visible.
+    visible_rect: Option<WorldRect>,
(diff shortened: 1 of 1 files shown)


gw3583:
This includes #3465, which is being reviewed separately. We can either wait until that's reviewed or close that and land both commits in this PR.

r? @bholley or @kvark 

Pending picture caching try:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=2153754475503d6cf80886877f5f9a133d228e85

PR #3468, opened 2019-01-04
Allow running headless.py in a debugger.

nical:
Set the environment variable DEBUGGER to either "gdb", "cgdb", "rust-gdb" or "rr" and run headless.py.
Example: `DEBUGGER=rr ./script/headless.py rawtest`

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3468)

### DIFF
--- a/wrench/script/headless.py
+++ b/wrench/script/headless.py
@@ -63,6 +63,13 @@ def use_rr():
     return debugger() == 'rr'
 
 
+def optimized_build():
+    if "OPTIMIZED" in os.environ:
+        opt = os.environ["OPTIMIZED"]
+        return opt not in ["0", "false"]
+    return True
+
+
 def set_osmesa_env(bin_path):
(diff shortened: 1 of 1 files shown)


nical:
While I'm at it I added a way to use debug builds by setting "OPTIMIZED" to either "0" or "false".

wrench/script/headless.py
```diff
@@ -67,11 +88,31 @@ def set_osmesa_env(bin_path):
...
+
+subprocess.check_call(build_cmd)
+
+set_osmesa_env(objdir)
+
+dbg_cmd = []
+if use_rr():
+    dbg_cmd = ['rr', 'record']
+elif use_gdb():
+    dbg_cmd = [debugger(), '--args']
```
emilio: Maybe something like the following would be worth it:

```suggestion
    dbg_cmd = [debugger(), '--args']
elif debugger():
    print("Unknown debugger: {}".format(debugger())
    sys.exit(1)
```

PR #3470, opened 2019-01-04
Fix blob leak panic in rawtests

nical:
When rawtests are run in debug mode we panic towards the end because some blobs were not manually evicted from the texture cache. Some of the tests were not fully cleaning up after themselves and this commit fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3470)


nical:
The failure I'm referring to is:

```
thread 'WRRenderBackend#0' panicked at 'Manual eviction requires cleanup', webrender/src/resource_cache.rs:225:9
```

r? anyone

PR #3473, opened 2019-01-04
Fix a panic that can occur when the cached picture is off-screen.

gw3583:
When a valid picture cache exists, that is followed by a display
list where the cached picture is completely off-screen, the code
would previous clear the existing tiles array. However, it wasn't
resetting self.tile_count, which could result in invalid tiles
being accessed and causing a panic.

Instead of trying to handle that explicitly, handle this case by
using a tile rect that covers the current screen. This means that
the tile array stays valid, and may mean that retained tiles can
be reused if that picture is subsequently scrolled back into view.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3473)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -458,20 +458,16 @@ impl TileCache {
             .map(&pic_rect)
             .expect("bug: unable to map picture rect to world");
 
-        // TODO(gw): Inflate the world rect a bit, to ensure that we keep tiles in
-        //           the cache for a while after they disappear off-screen.
+        // If the bounding rect of the picture to cache doesn't intersect with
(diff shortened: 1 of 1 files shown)


gw3583:
r? anyone

Pending try (pic caching enabled):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=5feddec2a09967743ead41f6418c2aeb4c63a489

This fixes a panic I encountered while trying to repro https://bugzilla.mozilla.org/show_bug.cgi?id=1517805 - it *might* also fix that bug itself, even though I can't currently repro it.

gw3583:
Closing since this is included in #3474

PR #3477, opened 2019-01-07
 Optimizations to picture caching validation and blits.

gw3583:
Reduce the size of blits done to the tile cache, by using only
the valid rect portion, rather than any clipped out parts.

Remove the per-primitive needed/current rect code, in favor
of a single valid pixel rect. This was an attempted fix for
some primitives not having a correct clip rect - but it turns
out this was caused by a bug in how we were accumulating the
clip rect for picture primitives. Now that this is resolved,
we can use the simpler / faster single valid rect method.

This also unblocks a couple of other changes coming - such as
only caching tiles if they have been the same for some time.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3477)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -112,10 +112,7 @@ pub struct Tile {
     /// The current local rect of this tile.
     pub local_rect: LayoutRect,
     /// The valid rect within this tile.
-    valid_rect: WorldRect,
-    /// The currently visible rect within this tile, updated per frame.
-    /// If None, this tile is not currently visible.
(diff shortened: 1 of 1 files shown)


gw3583:
(I think @bholley will be fine to rubber stamp this with proper review later, but I'll at least wait on the pending try run).

https://treeherder.mozilla.org/#/jobs?repo=try&revision=6cb4b383bd55b77152f5dceb894b289385b64e10

gw3583:
Hmm, this might need a little bit more thought first.

PR #3478, opened 2019-01-07
Fix some intermittent picture caching test failures.

gw3583:
Storing just the world position of the clip vertex can fail in
some edge cases. Instead, store the local origin and a list of
the relative transforms that affect primitive on this tile, which
is also a CPU optimization.

Also reduce the number of pixels blitted during tile creation
by only copying the visible part of the tile.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3478)


gw3583:
Try run here looks much better:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=0f4183b94ad0f23d7fb1fb06db786b60603434fd

PR #3480, opened 2019-01-07
picture: Only collect clip nodes for raster roots, not all surfaces.

emilio:
This is effectively a revert of 47242e13a6f88c17186afeaf2545796670e2d078.

The assumption that patch introduces does just not hold for fixed-pos clips,
unless we introduce more complexity to turn them into their own spatial nodes of
course.

Per https://bugzilla.mozilla.org/show_bug.cgi?id=1501111#c11 this is no longer
needed for picture caching, so let's use the clip node collectors when the
assumption they use hold (which is when the picture is a raster root, since
raster roots are reference frames, and reference frames are a containing block
for fixed-pos elements).

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1501111.

### DIFF
--- a/webrender/src/clip.rs
+++ b/webrender/src/clip.rs
@@ -497,22 +497,21 @@ impl ClipStore {
         &self.clip_node_instances[(node_range.first + index) as usize]
     }
 
-    // Notify the clip store that a new surface has been created.
-    // This means any clips from an earlier root should be collected rather
-    // than applied on the primitive itself.
-    pub fn push_surface(
-        &mut self,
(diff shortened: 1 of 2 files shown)


emilio:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=895697c9f64279f37a5060ff50e91e24f5f9096b

gw3583:
The try run seems to be missing some of the usual configs? But maybe they are not relevant in this case?

The change looks good to me.

PR #3482, opened 2019-01-08
Fix handling of world clip rects when APZ is occurring.

gw3583:
During APZ, it's possible to receive a new display with effectively
the same content, but with display items at different local
positions.

Handling this with world clip rects, while not invalidating tiles
unnecessarily as they change relative to scrolling tiles is
quite subtle. This patch fixes these by maintaining the region
of the primitive in its primitive-relative space that affects
the tile (similar to how primitive interning stores rects as
primitive relative).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3482)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -181,11 +181,20 @@ pub struct PrimitiveDescriptor {
     /// Uniquely identifies the content of the primitive template.
     prim_uid: ItemUid,
     /// The origin in local space of this primitive.
-    origin: LayoutPoint,
+    origin: WorldPoint,
     /// The first clip in the clip_uids array of clips that affect this tile.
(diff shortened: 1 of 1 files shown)


gw3583:
r? @kvark or @bholley

Try run with pic caching enabled looks good so far (still a few pending jobs):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=68daa0348f59d86c20a8379e62d78340ee2f1add

gw3583:
Try run looks good, the CI failure looks like an intermittent network issue.

gw3583:
Replied to comments and pushed a couple of fixups. But I pushed them to #3487, which includes this, so I'll close this patch and land it as part of #3487.

Issue #3399, opened 2018-12-11
Dashed borders should draw with a tight bounding rect.


gw3583:
When dashes are drawn we draw vertices using the bounds of the corner rect. This results in a lot of overdraw for corners with a lot of dashes. This is generally not an issue since the corners are cached and the dash count / corner size is typically quite low.

However, in some Gecko reftests, the combination of the above with a software rasterizer, can make this take so long that it times out the reftest.

To fix it, we should use a tight bounding rect for the vertices around a dash in order to drastically reduce overdraw.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1510076 for more information.

gw3583:
@nical fixed this.

PR #3483, opened 2019-01-08
Reduce the overdraw of border dashes.

nical:
Shrink the dash segments using a conservative rect computed in the vertex shader.
The dash segment rects could be made tighter (using a very simple and safe conservative approximation here).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3483)


nical:
The picture below shows the geometry in renderedoc, as you can see the rect is very conservative and could be made tighter.

screenshot from 2019-01-08 14-15-28

I think that it overshoots by a max of 30% (`2.0 - sqrt(2)/2`) a (still) conservative rect that would assume no curvature on the dashes. We could tweak this but I suspect that this PR is enough to fix #3399 and [bug 1510076](https://bugzilla.mozilla.org/show_bug.cgi?id=1510076). Before this PR, each segment covers the entire corner so the wins are already significant.

nical:
With this we can re-enable the 4 disabled crashtests of bug 1510076 and they pass: https://treeherder.mozilla.org/#/jobs?repo=try&revision=6c622c1d46307b5b1f46d1f7d34ecc8890e619ca&selectedJob=220591569

r? @kvark

webrender/res/cs_border_segment.glsl
```diff
@@ -227,6 +225,17 @@ void main(void) {
 
         vPos = vClipParams1.xy + radius * (2.0 * aPosition.xy - 1.0);
         vPos = clamp(vPos, vec2(0.0), aRect.zw);
+    } else if (clip_mode == CLIP_DASH_CORNER) {
+        vec2 center = (aClipParams1.xy + aClipParams2.xy) * 0.5;
+        // This is a gross approximation which works out because dashes don't have
+        // a strong curvature and we will overshoot by inflating the geometry by
+        // this amount on each side (sqrt(2) * length(dash) would be enough and we
+        // compute 2 * approx_length(dash)).
+        float dash_length = length(aClipParams1.xy - aClipParams2.xy);
```
kvark: is it possible to process width and height separately here (and get a tighter bound therefore)?

nical: I may be missing something but since we don't know the orientation of the dash, it's hard to consider the dash width and length separately. So we have to assume the longest side in any direction and potentially the worst orientation (45 degrees).

We could compute the orientation from `aClipParams1.xy`  and `aClipParams2.xy` which are at the start and end of the dash but that would add a lot of math. To get a really tight rect we'd need to consider the curvature of the dash in the computation which would make that a ton more complicated for a rather small amount of overdraw.

That's why I settled for a very conservative rect. Without extra information we could technically reduce the rect because, again, assuming dashes are rectangles with no curvature (which is not quite true), a sufficiently conservative bounding square would have each side equal to `sqrt(2) * max(width, length)` (sqrt(2) coming from the worst-case angle of 45 degrees) and this PR uses a factor of 2 instead of sqrt(2). The factor of two is implicit (it's just that we inflate the rect of that quantity on each side).

Because this formula doesn't account for the curvature we'd need some extra headroom to make sure we don't clip dashes with strong curvature but quantifying how much we need is not worth the complexity in my opinion. It's most likely less than the ~30% from this PR though. So if we are willing to accept a magic constant factor that has no mathematical ground, we should be safe multiplying by, say, 0.8 and leave only about 10% on each side for the curvature. I almost put that on the PR but I suspected an arbitrary factor would generate some debate.

I'm not sure real-world pages will benefit much from tighter bounding rects anyway, so I think we should stick to something simple and maybe come back and implement something more complicated if/when we have evidence that it can have an impact.

nical: So right now there isn't really a magic factor (other than since we are creating a square with edges at `±max(width, length)` from the center we get `2*max(width, length)` sides. Do you mean adding a factor like the 0.8 I previously mentioned or do you mean I should reword the existing comment?

PR #3486, opened 2019-01-08
use new dwrote API for making GlyphRunAnalysis fallible

lsalzman:
This resolves Firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=1424122

DWrote 0.7.0 now allows GlyphRunAnalysis calls to return errors, so we just need to adjust the code that uses them to appropriately fail as might occasionally happen, so that we no longer assert.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3486)

### DIFF
--- a/webrender/src/platform/windows/font.rs
+++ b/webrender/src/platform/windows/font.rs
@@ -253,7 +253,7 @@ impl FontContext {
         size: f32,
         transform: Option<dwrote::DWRITE_MATRIX>,
         bitmaps: bool,
-    ) -> (dwrote::GlyphRunAnalysis, dwrote::DWRITE_TEXTURE_TYPE, dwrote::RECT) {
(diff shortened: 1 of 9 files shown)


webrender/src/platform/windows/font.rs
```diff
@@ -355,7 +355,10 @@ impl FontContext {
         } else {
             None
         };
-        let (_, _, bounds) = self.create_glyph_analysis(font, key, size, transform, bitmaps);
+        let (_, _, bounds) = match self.create_glyph_analysis(font, key, size, transform, bitmaps) {
```
staktrace: nit: You could do `.ok()?;` on the `Result` instead of the match statement here

lsalzman: Made a substantial rewrite of GlyphRasterResult into a Result type to make this all much cleaner.

PR #3487, opened 2019-01-08
 Only cache tiles when they have had the same content > 2 frames.

gw3583:
This prevents tile caching doing redundant blits if the content
is changing every frame. This fixes a lot of the performance
regression on pages / benchmarks that have content which is
animated / changing every frame.

It still allows caching of tiles within those pages that are
static, giving the best of both worlds.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3487)

### DIFF
--- a/webrender/src/picture.rs
+++ b/webrender/src/picture.rs
@@ -80,6 +80,7 @@ pub struct TileIndex(pub usize);
 /// size for real world pages.
 pub const TILE_SIZE_WIDTH: i32 = 1024;
 pub const TILE_SIZE_HEIGHT: i32 = 256;
+const FRAMES_BEFORE_CACHING: usize = 2;
 
 #[derive(Debug)]
 pub struct GlobalTransformInfo {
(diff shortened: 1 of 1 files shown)


gw3583:
r? @kvark or @bholley

This includes #3482, which can be reviewed separately or as part of this PR.

This is a potential fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1518406, https://bugzilla.mozilla.org/show_bug.cgi?id=1514539 and https://bugzilla.mozilla.org/show_bug.cgi?id=1518407, and is just generally a sensible thing to do. Having said that, the reported performance regressions in those bugs are severe, and don't match what I see with picture caching, so still warrant further investigation.

Pending try (pic caching enabled):
https://treeherder.mozilla.org/#/jobs?repo=try&revision=bdca53d341eb9b2afd88b1f06f9db9091a74647c

gw3583:
Try run looks good.

gw3583:
Pushed a couple of fixes and addressed the comments in each PR.

Thanks for the review!

staktrace:
@gw3583 is this good to land now?

gw3583:
@staktrace I think so - there are a couple of comments from #3482 that need are not completely resolved, but I think it's fine to do them as follow ups.

@bors-servo r=kvark

PR #3488, opened 2019-01-09
Don't evict standalone textures when clearing shared cache.

gw3583:
When reclaiming shared texture cache memory, the final step that
can occur is to completely flush the shared cache in order to
shrink it.

However, this was calling clear() which also evicts all the
existing standalone textures, even ones that are in current use.

This results in picture caching tiles being evicted in some cases
when they shouldn't be. Instead, change the code to only clear
the shared texture cache in this code path.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3488)

### DIFF
--- a/webrender/src/freelist.rs
+++ b/webrender/src/freelist.rs
@@ -263,6 +263,7 @@ impl<T, M> FreeList<T, M> {
         slot.value.take().unwrap()
     }
 
+    #[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


gw3583:
Try run looks good.

bholley:
@bors-servo r+

We discussed on IRC, this is just a performance optimization rather than a correctness fix.

PR #3491, opened 2019-01-09
Clear blob_requests on LowPrioritySceneBuilder.

jrmuizel:
PR #3398 made it so we weren't clearing txn.blob_requests
on the low priority scene builder. I believe this change was
unintentional. Clearing blob_requests cuts the amount of
blob rasterization that we do in half on hixie-004.xml and
should fix the talos regression that we were seeing.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3491)


nical:
Yes indeed. @bors-servo r+

kvark:
@jrmuizel nicely spotted! I think this would make sense to do inside `rasterize_blobs()`, but that's minor.
Thanks for fixing this :1st_place_medal:

PR #3492, opened 2019-01-09
Clear non-rasterized blob image tiles during exceeding MAX_TILES_PER_REQUEST

sotaroikeda:
In create_blob_scene_builder_requests(), when  blob image tiles exceeds MAX_TILES_PER_REQUEST, some tiles are not rendered. It caused an incorrect blob painting. To avoid the problem, non-rasterized blob image tiles needs to be cleard.

The pull request does the clearing in set_blob_rasterizer(), since blob image rendering happens async way.

See [Bug 1494164](https://bugzilla.mozilla.org/show_bug.cgi?id=1494164).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3492)

### DIFF
--- a/webrender/src/render_backend.rs
+++ b/webrender/src/render_backend.rs
@@ -858,8 +858,8 @@ impl RenderBackend {
                         self.resource_cache.add_rasterized_blob_images(
                             replace(&mut txn.rasterized_blobs, Vec::new())
                         );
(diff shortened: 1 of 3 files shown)


sotaroikeda:
MAX_TILES_PER_REQUEST is 64, but the actual request became 286 (11*16) on my laptop in STR of [Bug 1494164](https://bugzilla.mozilla.org/show_bug.cgi?id=1494164).

PR #3493, opened 2019-01-09
Add basic picture caching debug overlay.

gw3583:
Add support for frame building to push debug rects and strings
that are displayed by the debug renderer.

Add a debug flag for toggling the picture caching debug overlay.

Add some basic information to the picture caching overlay:
 - The red rectangle is the current dirty rect.
 - The green rectangles show tile boundaries with per-tile stats.

In future, this will be expanded to include more information about
each cached tile, such as invalidation reasons.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3493)


gw3583:
r? @kvark or anyone

The current debug display (much more information yet to be added):

dirty_rect

webrender/src/picture.rs
```diff
@@ -1061,6 +1066,29 @@ impl TileCache {
             // Decide how to handle this tile when drawing this frame.
             if tile.is_valid {
                 self.tiles_to_draw.push(TileIndex(i));
+
+                #[cfg(feature = "debug_renderer")]
```
emilio: I think you can put the `#[cfg()]` just on top of the `if` and deindent it, here and below.

gw3583: Unfortunately that gives ```error: attributes are not yet allowed on `if` expressions```

PR #3495, opened 2019-01-09
Make static fontconfig linking on android optional

jdm:
The current setup breaks Servo's Magic Leap build, which is treated as an Android target but has an old version of fontconfig present.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3495)


staktrace:
Just a note/reminder that we are going to be making mozilla-central the canonical upstream for webrender basically after #3495 merges, so although this PR can be reviewed on github it would land on m-c first. @jdm if you want to file a bugzilla/phabricator for this patch that would be great, if not I'm happy to land the patch on m-c after it gets reviewed here (and after #3495 is synced over).

jdm:
Did you mean a different PR number?

jdm:
Opened 1518937.

staktrace:
Thanks!

PR #3500, opened 2019-01-10
don't snap text position in local raster space

lsalzman:
This resolves Firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=1512010

We don't want to snap text positions in raster space local as this is used to indicate that the text is animated. There are still legitimate non-transform cases where we want to snap (i.e. no scale/rotate, only translation), so that is why this patch goes through some pain to pass in the raster space and make sure we check it before skipping the snapping.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3500)


lsalzman:
Moved patch to Gecko side...

PR #3506, opened 2019-01-14

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3506)


staktrace:
@lsalzman I think these patches might have made `reftests/text/shadow.yaml == reftests/text/shadow-ref.yaml` intermittent on macOS. The test passed on the merge commit bors made [here](https://tools.taskcluster.net/groups/Cha9mKloRbWbu-BCB5Capw/tasks/YRFz7ITIQWqfkxlcnvBktw/runs/0) but failed on the tests that ran as part of the PR branch [here](https://tools.taskcluster.net/groups/cIs0GLqVR0avtQ5T4D5pkQ/tasks/JzQCgdeKTimbSQndNH1tig/runs/0) and now stuff landing on top of this are also hitting failures in #3507. Unfortunately we are not running macOS wrench reftests on treeherder (it was deemed nonblocking) so this didn't get caught earlier. If you can investigate that would be good.

PR #3507, opened 2019-01-14

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3507)


staktrace:
Weird. The test must be intermittent since there's no way these patches would affect it. Next patches that get merged to m-c should result in bors trying again, so let's wait and see what happens.

PR #3516, opened 2019-01-22

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3516)


nox:
I'm so damn confused about those syncs. https://bugzilla.mozilla.org/show_bug.cgi?id=1521518 is RESOLVED FIXED, why isn't the code in that sync?

nox:
Oh I unconfused myself, it landed in a previous sync, disregard me. 😂

PR #3522, opened 2019-01-26

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3522)


staktrace:
@jdm not urgent but bors seems stuck again

PR #3540, opened 2019-02-04
use switch statement for text snapping decision

### LINKED ISSUE #3538: Text rasterization regression
Compared to the 0.58.0 release on crates.io, current master causes a quality regression for text rendering on my OSX machine (Mojave). I've narrowed the regression down via git bisect, and the change that introduces the problem is 16a9870083622de45f60d5dac45e3d599a8c37dd.

I'm using `RasterSpace::Screen` for all the stacking contexts that contain text glyphs, and after the change, the glyph positioning is all over the place.

I've been trying to dig into this issue for a bit, and I'm not sure if it's a bug in the OSX OpenGL driver, but the `if (raster_space == RASTER_SCREEN)` condition in the text run shader behaves really weirdly. Playing around with it a bit, `raster_space == RASTER_SCREEN` is buggy, while `raster_space != RASTER_LOCAL` and `raster_space >= RASTER_SCREEN` fixes the issue.

I've verified that when the primitive header is built on the CPU, the third element in the user data vector is definitely `1` (`RasterizationSpace::Screen`), but maybe there is something wrong with the way that the header gets transmitted to the GPU?

I've uploaded screen grabs for comparison:

- [Good](https://user-images.githubusercontent.com/2677165/52171054-df404c00-27a9-11e9-94e7-b272e903de45.png) (pre 16a9870083622de45f60d5dac45e3d599a8c37dd)
- [Bad](https://user-images.githubusercontent.com/2677165/52171056-e0717900-27a9-11e9-98c4-377a45ed8a63.png) (post 16a9870083622de45f60d5dac45e3d599a8c37dd)

srijs:
Fixes #3538.

@lsalzman this implements your suggestion of switching from an `if` to a `switch` statement, and it really fixes the issue for some puzzling reason! I'd love to put a better description of the issue into the commit message, but this behaviour thoroughly confuses me. Can you offer more insight into why this particular change alleviates the problem?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3540)

### DIFF
--- a/webrender/res/ps_text_run.glsl
+++ b/webrender/res/ps_text_run.glsl
@@ -81,34 +81,39 @@ VertexInfo write_text_vertex(RectWithSize local_clip_rect,
     // Compute the snapping offset only if the scroll node transform is axis-aligned.
     if (remove_subpx_offset) {
         // Be careful to only snap with the transform when in screen raster space.
-        if (raster_space == RASTER_SCREEN) {
-            // Transform from local space to device space.
-            float device_scale = task.common_data.device_pixel_scale / transform.m[3].w;
-            mat2 device_transform = mat2(transform.m) * device_scale;
+        switch (raster_space) {
+            case RASTER_SCREEN: {
+                // Transform from local space to device space.
+                float device_scale = task.common_data.device_pixel_scale / transform.m[3].w;
+                mat2 device_transform = mat2(transform.m) * device_scale;
 
-            // Ensure the transformed text offset does not contain a subpixel translation
-            // such that glyph snapping is stable for equivalent glyph subpixel positions.
-            vec2 device_text_pos = device_transform * text_offset + transform.m[3].xy * device_scale;
-            snap_offset = floor(device_text_pos + 0.5) - device_text_pos;
+                // Ensure the transformed text offset does not contain a subpixel translation
+                // such that glyph snapping is stable for equivalent glyph subpixel positions.
+                vec2 device_text_pos = device_transform * text_offset + transform.m[3].xy * device_scale;
+                snap_offset = floor(device_text_pos + 0.5) - device_text_pos;
 
-            // Snap the glyph offset to a device pixel, using an appropriate bias depending
-            // on whether subpixel positioning is required.
-            vec2 device_glyph_offset = device_transform * glyph_offset;
-            snap_offset += floor(device_glyph_offset + snap_bias) - device_glyph_offset;
+                // Snap the glyph offset to a device pixel, using an appropriate bias depending
+                // on whether subpixel positioning is required.
+                vec2 device_glyph_offset = device_transform * glyph_offset;
+                snap_offset += floor(device_glyph_offset + snap_bias) - device_glyph_offset;
 
-            // Transform from device space back to local space.
-            local_transform = inverse(device_transform);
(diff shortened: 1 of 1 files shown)


nical:
Looks like this should be documented in the [driver issues](https://github.com/servo/webrender/wiki/Driver-issues) wiki page.

staktrace:
Thanks for the patch! https://bugzilla.mozilla.org/show_bug.cgi?id=1525032 will merge this to m-c and then it will get synced back to webrender.

nical:
> It sounds very similar to https://github.com/servo/webrender/wiki/Driver-issues#2864---mac-glsl-compiler-bug-with-integer-comparisons, though not quite the same?

Looks very similar indeed, I added https://github.com/servo/webrender/wiki/Driver-issues#2864---mac-glsl-compiler-bug-with-integer-comparisons right next to it, not sure whether they are the same but it's good to keep a history of each workaround.

kvark:
What's bothering me here is that `local_transform` isn't always initialized. Maybe we are just hitting the UB in the shader because of that, and turning `if` into `switch` just affects what garbage turns out to be placed on `local_transform` space, thus luckily improving the results?

lsalzman:
local_transform is only ever used with WR_FEATURE_GLYPH_TRANSFORM, which is only ever used with RasterSpace::Screen

srijs:
FWIW, initializing `local_transform` to the identity matrix on line 74 also fixes the issue on my machine. Happy to push that change instead if you want to go that way.

lsalzman:
That's fairly surprising that even though local_transform is never used at
all that it caused that issue. In any case, keep the switch() I would
suggest, but also add in the initialization of mat2 local_transform =
mat2(1.0) just to make the compiler supremely happy, I suppose.

kvark:
@lsalzman are you 100% sure that `local_transform` wasn't used in this case?
@srijs  would you mind setting it to something crazy and seeing if it affects anything?

I'm not a fan of `switch` fix because that's something we can't explain. And the hypothesis that `local_transform` is used would provide a much cleaner explanation to the root of the problem :)

lsalzman:
Since this is technically an enum value, the switch is a more natural expression of the intent of the code anyway, so there is no harm in just leaving the switch.

srijs:
I've pushed the additional commit onto the branch.

@kvark using a bogus initializer for `local_transform` does not have any effect in my case, but given that I only use screen raster spaces, I wouldn't expect to hit the branch where `local_transform` is used uninitialized?

lsalzman:
WR_FEATURE_GLYPH_TRANSFORM will only get triggered if there is a font transform, which can only happen in RasterSpace::Screen. If RasterSpace::Local is used, then the font transform is identity, and WR_FEATURE_GLYPH_TRANSFORM will never be used. So the only possible way for the shader enum to not actually be RASTER_SCREEN is if there was either a driver bug with the uniform value, or a bug in our uniform code. However, given that this problem only manifests on one very limited configuration on macOS and only on specific hardware, and nowhere else, I think this is just a driver bug.

staktrace:
Sorry about closing the PR while discussion is still ongoing but the syncing process assumes that once the PR is approved it won't be modified. If the additional commit needs landing please open a new PR with that patch. Thanks!

PR #3547, opened 2019-02-10
always initialize local_transform to the identity matrix

srijs:
This is a follow-up on #3540.

/cc @kvark @lsalzman

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3547)


staktrace:
This will land in https://bugzilla.mozilla.org/show_bug.cgi?id=1526769 and then get mirrored back to this github repo.

PR #3568, opened 2019-02-24

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3568)


staktrace:
@gw Looks like the `blend-source-transform.yml` test is a little different on appveyor even though it passed on our taskcluster CI. Any thoughts on how we should resolve this? We could (a) try to investigate the difference and why it's happening, (b) mark it fuzzy, or even (c) turn off appveyor and just rely on m-c's testing.

staktrace:
Err meant to ping @gw3583

PR #3573, opened 2019-02-28

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3573)


staktrace:
Looks like rustc 1.33 decided `window` was unused even though it's used a couple of lines down. rustc 1.32 is perfectly happy with it. Woohoo.

nox:
The `Ok` type at that line is uninhabited.

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=97f7f50cc3e05ed90b9098dbc17cba1f

nox:
Note that I'm just stating the reason why it does that here, I don't think it's a good idea, as I've also used empty enums that way in other circumstances and that code probably warns now, I think an issue should be filed on [rust-lang/rust](https://github.com/rust-lang/rust).

staktrace:
For now I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1531512 to fix this on the WR side. I don't plan on filing an issue against rust-lang because I don't have a strong opinion about this, but feel free to do so if you think rustc should change behaviour.

nox:
https://github.com/rust-lang/rust/issues/58844

PR #3576, opened 2019-03-05

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3576)


staktrace:
I think we were using some ancient version of rustc on the mac builder. Updated it and it seems to do better in a test branch.

@bors-servo retry

PR #3592, opened 2019-03-15

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3592)


kvark:
huh, zero lines changed!

staktrace:
The thing that landed got backed out because it was broken. So yeah, no change.

PR #3610, opened 2019-04-06

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3610)


staktrace:
@jrmuizel is the Mac Mini still up?

PR #3636, opened 2019-05-03
The shadow stack changeset 

nical:
I'm putting it here because I think it's easier to get an overview of a long patch series here than on phabricator. I'll wait for https://phabricator.services.mozilla.com/D29777 to land before I move this over to mercurial, since it's unlikely to rebase cleanly.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3636)

### DIFF
--- a/webrender/src/batch.rs
+++ b/webrender/src/batch.rs
@@ -1298,6 +1298,7 @@ impl AlphaBatchBuilder {
                                             PrimitiveInstanceData::from(instance),
                                         );
                                     }
+                                    FilterOp::DropShadowStack(..) => { unimplemented!() } // TODO(nical)
                                     FilterOp::DropShadow(offset, ..) => {
                                         // Draw an instance of the shadow first, following by the content.
 
@@ -1409,7 +1410,8 @@ impl AlphaBatchBuilder {
                                             FilterOp::Sepia(..) => 6,
                                             FilterOp::Brightness(..) => 7,
                                             FilterOp::Opacity(..) => 8,
-                                            FilterOp::DropShadow(..) => 9,
+                                            FilterOp::DropShadow(..) |
+                                            FilterOp::DropShadowStack(..) => 9,
                                             FilterOp::ColorMatrix(..) => 10,
                                             FilterOp::SrgbToLinear => 11,
                                             FilterOp::LinearToSrgb => 12,
@@ -1433,6 +1435,7 @@ impl AlphaBatchBuilder {
                                             }
                                             // Go through different paths
                                             FilterOp::Blur(..) |
+                                            FilterOp::DropShadowStack(..) |
                                             FilterOp::DropShadow(..) => {
                                                 unreachable!();
                                             }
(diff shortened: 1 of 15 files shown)


nical:
@glennw @kvark how would you prefer the final review to look like once I move this to phrabricator, in small-ish patches like here or larger squashed ones ? I can squash there and keep the small patches here, although once I move to hg it's unlikely that I'll keep this PR in sync with review changes, etc.

kvark:
@nical I'd prefer if you used git cinnabar in the first place, so that Phabricator didn't associate with mercurial any more :)
Seriously though, I think Reviewable workflow would be *slightly* more convenient for me.

nical:
I added the PR to reviewable. we can start at least the initial review there and do the final formatting/approval in phabricator.

nical:
Heads up: having a non repr(C) type in FilterOp is causing a ton of badness on the C++ side due to relying on having the declaration generated by cbindgen, and I can't come up with a clean way around that.

So my plan is to change these patches in the following way:

- Remove ShadowStack from the public facing FilterOp API.
- Have a separate  internal FilterOp type in WebRender which will have a shadow stack variant and will be generated automatically during the PushShadow/PopAllShadows dance in display_list_flattener.

The upside of doing this is that Gecko (and servo, and other users) will get the benefits without opting into new APIs.
As an aside that'll let me box the color matrix in webrender's internal FilterOp and drastically slim down the enum and structures that embed it.

nical:
So we have two ways to render shadows:

## As a picture composition op

This code path rasterizes the picture, creates a mask from a blur of the picture and renders the picture and its shadow as two segments of an image brush, applying the shadow color at the last step.
The current shadow stack code adds support for efficient multiple shadows in this code path specifically, but Gecko isn't set up to take advantage of it at least in the use case we primarily care about (text with loads of shadows).

## Using PushShadow / PopAllShadows

This code path is different. We generate special shadow items which are a "tinted" version of the shadowed item, and blur that already colored. We then composite the blurred result using an image brush.
The idea of a shadow stack with per shadow colors isn't compatible with this.

## What now

It's important to get the PushShadow/PopAllShadows path to optimize large shadow stacks to fix issues like https://bugzilla.mozilla.org/show_bug.cgi?id=1536021

The simplest thing would probably be to generated a picture with a shaodw stack composition op when detecting shadow stacks in `pop_all_shadows` and mark that picture so that we only composite its shadows.


@gw3583 @kvark a few questions:

- Right now if we have several items in a Push/Pop shadow pair, they all get shadowed independently. Is it for correctness or simplicity? Would it be correct/desirable to push all of the items into the same picture and shadow that?
- Is it OK to not do sup-px aa for text shadows at all (even when the blur radius is zero)?
- Could we implement all shadows the same way (using the approach of the picture composition op where the color is applied at the end) and get rid of all of the shadow item code? I find the code already more complicated than I'd like and I'm about to make it worse.

nical:
@gw3583 I see "Make box shadow a first-class primitive" in the roadmap for Q3 could you elaborate on that?

PR #3664, opened 2019-06-05
Use dyn for trait objects.

waywardmonkeys:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3664)


staktrace:
I think the second patch on https://bugzilla.mozilla.org/show_bug.cgi?id=1556597 includes these changes as well as some as other stuff

PR #3687, opened 2019-06-26
Remove sha256 from dependencies

### LINKED ISSUE #3586: Remove SHA256 dependency (since it's only used for hashing shaders)?
Right now webrender has a dependency on sha256 which is only used once, to verify the unique-ness of an OpenGL source string (by hashing the vertex and the fragment shader together):

https://github.com/servo/webrender/blob/607ce5d4ee77ffaafc6dc418d9b0dc37836d6026/webrender/build.rs#L36

This is the only place where the sha256 dependency is used - only during build time, i.e. the performance of the hasher itself doesn't really matter, just that the resulting program hash is unique. In my fork of webrender, I've removed this dependency in favor of  `std::hash::DefaultHasher` (replacing the `[u32;8]` sha256 hash with a simple `u64`) and haven't noticed any impact on functionality (but reduced the dependencies by about 10 crates). So I wanted to ask if it would be possible to remove the `sha256` dependency, because `DefaultHasher` is good enough. Another alternative would be to ditch the hashing approach altogether (since webrenders vertex / fragment shaders are always the same, they don't change at runtime), and give each vertex + fragment shader a unique `VertexShaderId` / `FragmentShaderId`.

fschutt:
- The shaders are not hashed by their content anymore, only by their file name / their imports - i.e. a shader "a.glsl" that imports "b.glsl" get a unique ID based on the combination ["a.glsl", "b.glsl"] (instead of a SHA hash of the file contents of a.glsl and b.glsl). This also avoids possible collisions.
- The "program digest" is now just a shader ID, there was a check for checking whether the source of the loaded shader is equal to the source of the included shader (which I thought was fairly useless, since those shaders are always the same, so the check can never fail?). So I've removed that check, but I'm not sure if it had historical reasons.

This is a WIP PR, just to give an update, doesn't compile yet, closes #3586.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3687)


fschutt:
PR seems to be borked. I'll create a new one

PR #3743, opened 2019-08-28
Check key existence in hit_test::get_pipeline_root

### LINKED ISSUE #3739: Intermittent "no entry found for key" error in HitTester::get_pipeline_root
Under certain scenarios in Servo (see https://github.com/servo/servo/issues/23964, https://github.com/servo/servo/issues/21480) the `RenderBackend` thread might panick at [`HitTester::get_pipeline_root`](https://github.com/servo/webrender/blob/49ef7304c380ca0bd2a7590e7d7808487bbf5ef3/webrender/src/hit_test.rs#L509) with a `no entry found for key` message, due to the fact that at the time the function gets called `pipeline_id` is not in the `pipeline_root_nodes` hash map.

Having a check for containment would remove the panick. [`HitTester::get_pipeline_root`](https://github.com/servo/webrender/blob/49ef7304c380ca0bd2a7590e7d7808487bbf5ef3/webrender/src/hit_test.rs#L509) is only used in one single [place](https://github.com/servo/webrender/blob/49ef7304c380ca0bd2a7590e7d7808487bbf5ef3/webrender/src/hit_test.rs#L569), which makes it easy to change it return an `Option`. However, I'm not very familiar with webrender and whether this is an unrecoverable error or not.

georgeroman:
Fixes #3739

---
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3743)

### DIFF
--- a/webrender/src/hit_test.rs
+++ b/webrender/src/hit_test.rs
@@ -505,8 +505,12 @@ impl HitTester {
         result
     }
 
-    pub fn get_pipeline_root(&self, pipeline_id: PipelineId) -> &HitTestSpatialNode {
-        &self.spatial_nodes[self.pipeline_root_nodes[&pipeline_id].0 as usize]
+    pub fn get_pipeline_root(&self, pipeline_id: PipelineId) -> Option<&HitTestSpatialNode> {
+        if self.pipeline_root_nodes.contains_key(&pipeline_id) {
+            Some(&self.spatial_nodes[self.pipeline_root_nodes[&pipeline_id].0 as usize])
+        } else {
+            None
+        }
     }
 }
 
@@ -567,8 +571,11 @@ impl HitTest {
             .and_then(|id|
                 hit_tester
                     .get_pipeline_root(id)
-                    .world_viewport_transform
-                    .transform_point2d(point)
+                    .and_then(|pipeline_root| {
+                        pipeline_root
+                            .world_viewport_transform
+                            .transform_point2d(point)
+                    })
             )
             .unwrap_or_else(|| {
                 WorldPoint::new(self.point.x, self.point.y)


webrender/src/hit_test.rs
```diff
@@ -505,8 +505,12 @@ impl HitTester {
         result
     }
 
-    pub fn get_pipeline_root(&self, pipeline_id: PipelineId) -> &HitTestSpatialNode {
-        &self.spatial_nodes[self.pipeline_root_nodes[&pipeline_id].0 as usize]
+    pub fn get_pipeline_root(&self, pipeline_id: PipelineId) -> Option<&HitTestSpatialNode> {
+        if self.pipeline_root_nodes.contains_key(&pipeline_id) {
```
kvark: could we do `get().map()` instead?

webrender/src/hit_test.rs
```diff
@@ -505,8 +505,12 @@ impl HitTester {
         result
     }
 
-    pub fn get_pipeline_root(&self, pipeline_id: PipelineId) -> &HitTestSpatialNode {
-        &self.spatial_nodes[self.pipeline_root_nodes[&pipeline_id].0 as usize]
+    pub fn get_pipeline_root(&self, pipeline_id: PipelineId) -> Option<&HitTestSpatialNode> {
```
kvark: what are the conditions for which the pipeline root couldn't be found? I'm worried that by returning an `Option` we'd just be hiding an internal logic problem.

georgeroman: Unfortunately, I'm not sure what are the conditions under which this can generally happen as I'm not that familiar to the codebase.

kvark: More importantly, I wonder why Servo hits this and Gecko doesn't. Perhaps, it's something on Servo side to fix... We need to understand the problem better.

jdm: Yes, I'm digging in a bit in https://github.com/servo/servo/issues/23964 since it's easy to reproduce the panic.

jdm: Based on my investigations in that issue, I think silently suppressing this error is probably not the best solution.

PR #3770, opened 2019-10-04

moz-gfx:
This change is [Reviewable](https://reviewable.io/reviews/servo/webrender/3770)


jrmuizel:
Looks like it's failing a reftest on Windows.

PR #3779, opened 2019-10-22
Dedupe synstructure

atouchet:


kvark:
Ported to https://phabricator.services.mozilla.com/D50138

PR #3792, opened 2019-11-05

moz-gfx:


jrmuizel:
C:\fakepath(1753,25-29): error X4576: Input array signature parameter SV_Position cannot be indexed dynamically.

kvark:
@jrmuizel we need that on CI...

PR #3793, opened 2019-11-06
Bug 1575648 - Migrate from taskcluster.net to tools.community-tc.services.mozilla.com

petemoore:
See [bug 1575648](https://bugzil.la/1575648) for context.


petemoore:
> Submitting the task to Taskcluster failed. Details

This message is from the legacy taskcluster deployment, that no longer works with these changes. When this PR is merged we can remove the former github app so that only the new github app for the community deployment is active.

kvark:
Ported as https://phabricator.services.mozilla.com/D52103
try https://phabricator.services.mozilla.com/D52103

jdm:
This has merged from upstream.

PR #3796, opened 2019-11-07
Move external image handler API into webrender_api

jdm:
This reduces some unnecessary dependencies on webrender (vs webrender_api) in Servo and makes it more efficient to bisect during webrender upgrades.

### DIFF
--- a/example-compositor/compositor/src/main.rs
+++ b/example-compositor/compositor/src/main.rs
@@ -21,6 +21,73 @@ use webrender::api::units::*;
 #[cfg(target_os = "windows")]
 use compositor_windows as compositor;
 
+#[cfg(not(target_os = "windows"))]
+mod compositor {
+    use std::os::raw::{c_void, c_char};
+
+    pub struct Window;
+
+    pub fn create_surface(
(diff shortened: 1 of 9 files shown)


example-compositor/compositor/src/main.rs
```diff
@@ -21,6 +21,73 @@ use webrender::api::units::*;
 #[cfg(target_os = "windows")]
 use compositor_windows as compositor;
 
+#[cfg(not(target_os = "windows"))]
```
kvark: this change doesn't seem relevant?

jdm: It was relevant in that `cargo build` from the root directory is broken on non-windows platforms. In any case, I have submitted https://phabricator.services.mozilla.com/D52686 which only includes https://github.com/servo/webrender/pull/3796/commits/34d64d063f28f03ffa07e2110b046aeda4acb0c1.

atouchet:
Was this done in #3797?

kvark:
Yes, via https://phabricator.services.mozilla.com/D52686

PR #3805, opened 2019-11-28
Remove IPC API.

jdm:
This removes the use of ipc-channel. I would love to remove the Serialize/Deserialize lies as well, but last time I tried to do that proved harder than I anticipated.


kvark:
Why is this done?

jrmuizel:
```
./Cargo.lock:1: duplicates for `crossbeam-utils` are allowed, but only single version found
	The following packages depend on version 0.2.2 from 'crates.io':
		crossbeam-deque
		crossbeam-epoch
```

jrmuizel:
the ipc-channel usage isn't need anymore because of https://github.com/servo/servo/commit/564c16d75499ae45b3945f487e49587c6fd662c7

jdm:
Fixed the tidy issue.

kvark:
> error: Package `webrender_api v0.60.0 (/home/worker/webrender/webrender_api)` does not have these features: `ipc`

atouchet:
This was done in #3850.

PR #3809, opened 2019-12-02

moz-gfx:


jrmuizel:
@gw3583 reftests/text/split-batch.yaml is failing on MacOS

PR #3821, opened 2019-12-24
Format code using 'cargo fmt'

Atul9:


Atul9:
done, I have branched off from master and run `cargo fmt` and updated the PR

kvark:
@Atul9 unfortunately, landing your PR to upstream WebRender (which is not this repo, it's in Mozilla-Central) is problematic, since any change to WR (and there is a small team working on WR) would break it. I think what we can do is wait till the Friday night, and then for the mozilla inbound -> central sync, and then do the cargo fmt on the code from mozilla-central, and make it a PR in Phabricator directly, to be reviewed and landed on the weekend.
Also, all parties need to be notified and permission gained, so that we don't ruin anybody's changes... all in all, it feels like this is a task not worth contributing externally - we (as a team) are better suited to do it ourselves.

mrobinson:
This repository is just a mirror currently, so substantive changes to WebRender should be landed upstream: https://github.com/mozilla/gecko-dev/tree/24f9846b7d8ea0819edee7aafc51a39753f1619a/gfx/wr

PR #3827, opened 2020-01-10
Update smallvec

Manishearth:
Wrench still depends on an older smallvec, this can be fixed if we update winit and glutin, but it's a bunch of work.


atouchet:
This was done in #3914.

PR #3858, opened 2020-02-13
Update gleam to 0.8.0

fschutt:
Only updates gleam from 0.6.17 to 0.8.0 (there were some new, platform-specific functions added to the `Gl` trait but I don't think they matter?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,28 +2,31 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adler32"
-version = "1.0.2"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


fschutt:
In order to de-duplicate the dependencies, I have to remove the `sha2` crate first AND update `winit` to the latest version, which requires code changes to `wrench`. Not going to do that atm, don't have the time or patience.

kvark:
Why not `gleam 0.9`?

jrmuizel:
I ended updating gleam in https://bugzilla.mozilla.org/show_bug.cgi?id=1615694 which should be making it's way to this repo in a day or so.

jrmuizel:
This is merged now.

fschutt:
I mostly wanted gleam 0.8 because `glutin 0.23` works with gleam 0.8. Since this is merged now, I'll try working on my other PR instead.

Issue #3860, opened 2020-02-17
Multiwindow example panics when debug feature is enabled


SergeevPavel:
I try `cargo run --bin multiwindow --features=debug` and get
```
thread 'main' panicked at 'assertion failed: profiler.borrow().is_none()', /Users/pavel/.cargo/registry/src/github.com-1ecc6299db9ec823/thread_profiler-0.1.3/src/lib.rs:95:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
fish: 'cargo run --bin multiwindow --f…' terminated by signal SIGABRT (Abort)
```

As far as I understand the problem is in double initializaition of `thread_profiler` for compositor thread. What do you think about moving it under `wr_has_been_initialized` condition?

https://github.com/SergeevPavel/webrender/commit/7a2587acb45ba6552f86c9743ce31ceb616f364c

kvark:
Looks reasonable to me!

PR #3864, opened 2020-02-18
initailize thread profiler on Compositor thread only once

SergeevPavel:


SergeevPavel:
Resolve #3860

kvark:
Thank you!
ported to https://phabricator.services.mozilla.com/D63242

PR #3872, opened 2020-02-27

moz-gfx:


staktrace:
The version of rust on the mac and windows workers needs bumping. I'll take care of it.

staktrace:
Looks like a couple of failing reftests on windows. Is this a known intermittent thing?

jrmuizel:
Not to me.

staktrace:
The two failing tests are:

``
	reftests\text\raster_root_C_8192.yaml == reftests\text\raster_root_C_ref.yaml
	reftests\transforms\raster_root_B_8192.yaml == reftests\transforms\raster_root_B_ref.yaml
```

And when I pasted the output into reftest analyzer it looks like the failures are just fuzzy pixel differences. Patches in this PR are bug 1510030 by @hiikezoe , bug 1616995 by @jfkthame , bug 1596513 by @cbrewster , bug 1579235 by @gw3583 and bug 1613260 by @bpeersmoz . Any of you have ideas on what might have introduced some fuzz in these tests that shows up in appveyor but not Firefox Windows CI?

bpeersmoz:
That's my change from last night that introduced the reftests and the code
they're trying to test.

The fuzziness is normal in the sense that both images use different
resolutions so there is a lot of interpolation going on. If you look at
reftests.list in the commit you'll see that I had to give it fuzzy
parameters that are the max of what I got from try-run & running locally
with/without --angle.

If you've found another platform that exceeds these limits then please feel
free to increase the max difference and/or max count some more (since
you've already checked visually that the tests look legit qualitatively).

Note that even though the vast majority of pixels are off by a small
amount, a few outliers on the edge may be off by 100+ due to
lerping+blending.  This unfortunately leads to fuzzy statements like
(120,20000) which is borderline useless for catching actual regression if
and when it occurs. I've logged this bug to get this back under control:

https://bugzilla.mozilla.org/show_bug.cgi?id=1619393

Thanks for looking into this!
bert

staktrace:
Thanks! I'll put up a patch to increase the fuzz numbers. For `raster_root_C` it's just a couple more pixels, for `raster_root_B` it's more but still not a huge increase.

staktrace:
@jdm does bors need a kick?

staktrace:
Another fuzz difference in AppVeyor, put patch on https://phabricator.services.mozilla.com/D65513

PR #3875, opened 2020-03-07
Remove sha256 from dependencies (closes #3586)

### LINKED ISSUE #3586: Remove SHA256 dependency (since it's only used for hashing shaders)?

fschutt:
#3586 - `Sha256::new()` can be safely replaced by `DefaultHasher::new()` since its only used for *verifying* shaders if they use an different source file.

Note that there seems to have been a change in Rust Edition 2018 so that `Hasher::input` was renamed to `Hasher::write`. Not sure if this impacts the supported compiler versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1442,17 +1442,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "sha2"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "block-buffer 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "digest 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "fake-simd 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "opaque-debug 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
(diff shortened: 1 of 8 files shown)


fschutt:
Tests passed.

r? @kvark

kvark:
This landed upstream in https://phabricator.services.mozilla.com/D66064, closing

PR #3892, opened 2020-03-24

moz-gfx:


staktrace:
@bors-servo retry

Seems like appveyor does better outside of EST/PST working hours...

PR #3928, opened 2020-04-21

moz-gfx:


staktrace:
Looks like we need to update the docker image to include clang to make @jamienicol 's glsl thing work. I'm trying to do that

jamienicol:
Thanks @kats! For future reference, what did I break here and how do I avoid breaking it in the future?

staktrace:
So the webrender tests run in two configurations, since WebRender is used both as part of Firefox and as a standalone project. In Firefox CI they are run via the config in `taskcluster/ci/webrender/kind.yml`, but here on github for the standalone testing they are run via the configuration in `.taskcluster.yml`. Both invoke the `ci-scripts/linux-*-tests.sh` and similar scripts, so that the actual set of tests run in both environments is the same, but the setup of the environment is different. In Firefox CI the `webrender/kind.yml` file defines the dependencies on toolchains and base docker image (which is defined in-tree at `taskcluster/docker/webrender/`). The `.taskcluster.yml` file does similarly for standalone testing, but uses a different docker image and can't directly declare dependencies on firefox toolchains. So that latter thing is what broke the standalone testing.

In terms of avoiding it in the future, I guess the way to know would have been to clone this github repo, copy over your changes to a fork, and make a PR to let the run `.taskcluster.yml` file do its testing. I'm doing that in #3929 right now with additional changes to test my fixes. In general this sort of breakage is pretty rare, and should only happen if we introduce new dependencies into the build that aren't just rust crates. That was the case here.

(Aside: in theory you shouldn't need to create a PR, and should be able to run the taskcluster-github integration on your own fork directly but the documentation for that is missing, see bug 1619286).

jamienicol:
Cool, thanks. I'll be aware of the standalone webrender taskcluster configuration in the unlikely event I make a similar change again :)

(Just realised I tagged the wrong username in my last reply, sorry other kats!)

PR #3954, opened 2020-05-15

moz-gfx:


staktrace:
@jrmuizel is the mac worker still in the office?

PR #3957, opened 2020-05-21
Batch allocation refactor

kvark:


webrender/src/batch.rs
```diff
@@ -211,102 +369,116 @@ impl AlphaBatchList {
         // The bounding box of everything at this Z plane. We expect potentially
         // multiple primitive segments coming with the same `z_id`.
         z_bounding_rect: &PictureRect,
-        z_id: ZBufferId,
+        _z_id: ZBufferId,
```
gw3583, line 372: I guess we don't need this check anymore, but let's remove the param altogether if that's the case?

webrender/src/batch.rs
```diff
@@ -442,43 +633,53 @@ impl AlphaBatchContainer {
...
-                    self.alpha_batches[index].merge(other_batch);
-                    min_batch_index = index;
-                }
-                None => {
-                    self.alpha_batches.push(other_batch);
-                    min_batch_index = self.alpha_batches.len();
-                }
-            }
-        }
+        // Note: we are able to reorder the opaque batches as needed, theoretically.
```
gw3583, line 665: I don't think this will have any effect on fill rate. By definition, batch merging only occurs for non-overlapping regions of a target (separate render tasks). Does that have an effect on what we should do here by default?

kvark:
Thank you! I forgot to mention (here) that the latest code to review is in https://phabricator.services.mozilla.com/D76715

PR #3961, opened 2020-05-25
Fix mouse movement position tracking in scrolling example.

### LINKED ISSUE #3898: Hit test / scrolling does not work in scrolling example
I recently noticed that the `scrolling` example doesn't work anymore, no matter what I do. For some reason, the rectangles don't scroll anymore. The `animation` example still works (so transform / opacity keys still seem to work).

I only get this output on running the scrolling example:

```
OpenGL version 3.3 (Core Profile) Mesa 19.2.8
Shader resource path: None
Device pixel ratio: 1.25
Loading shaders...
Entering event loop
Hit test results:

Hit test results:

Hit test results:

Hit test results:

Hit test results:

Hit test results:

Hit test results:

Hit test results:

Hit test results:

Hit test results:

```

Theoretically there should be a list of rectangles that have been hit (i.e. the rectangles that need to be scrolled), but... nothing.

## How to reproduce

```
git clone https://github.com/servo/webrender
cd webrender/examples
cargo run --bin scrolling
```

Then move the mouse cursor above the rectangles and see if scrolling works.

## Additional info

- webrender version: https://github.com/servo/webrender/commit/02be1053a833a5490800260540adeb19611fcb0d
- rustc version:

```
rustc 1.41.0 (5e1a79984 2020-01-27)
binary: rustc
commit-hash: 5e1a799842ba6ed4a57e91f7ab9435947482f7d8
commit-date: 2020-01-27
host: x86_64-unknown-linux-gnu
release: 1.41.0
LLVM version: 9.0
```

Mornix:
Fixes #3898.

### DIFF
--- a/examples/common/boilerplate.rs
+++ b/examples/common/boilerplate.rs
@@ -235,9 +235,18 @@ pub fn main_wrapper<E: Example>(
         };
         match win_event {
             winit::WindowEvent::CloseRequested => return winit::ControlFlow::Break,
-            // skip high-frequency events
             winit::WindowEvent::AxisMotion { .. } |
-            winit::WindowEvent::CursorMoved { .. } => return winit::ControlFlow::Continue,
+            winit::WindowEvent::CursorMoved { .. } => {
+                custom_event = example.on_event(
+                        win_event,
+                        &mut api,
+                        document_id,
+                    );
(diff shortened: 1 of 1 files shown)


kvark:
Ported to https://phabricator.services.mozilla.com/D76716

PR #4002, opened 2020-07-01
Add a reftest to ensure that scroll frames can be defined and used in any order.

pcwalton:
r? @gw3583 

@mrobinson How's this?


gw3583:
Looks fine to me - would you be able to provide it as a patch to m-c (since this is a read-only mirror of m-c now)? If your environment isn't set up for that, let me know, and we can manually move this across to m-c.

jdm:
@gw3583 Would you mind moving this to m-c for us?

gw3583:
https://phabricator.services.mozilla.com/D84628

jdm:
Thanks!

PR #4168, opened 2020-11-20

moz-gfx:


staktrace:
```
thread 'validate_shaders' panicked at 'attempted to zero-initialize type `shaders::ffi::ShBuiltInResources`, which is invalid', /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/mem/mod.rs:622:9
```

staktrace:
@lsalzman @jamienicol any ideas on which patch caused this? I'm guessing it's not intermittent but a result of environment differences between this CI and firefox CI

jamienicol:
I don't think that could be from my patch

staktrace:
Oh, this is from the rust version bump. If I run [this command](https://searchfox.org/mozilla-central/rev/2efcda6dc74c63863fd8f04a6d9d7ac6b09c7eca/gfx/wr/ci-scripts/linux-debug-tests.sh#34-36) locally in gfx/wr on m-c tip, it works fine with rust 1.47 but fails with rust 1.48. I guess 1.48 is either stricter or buggy.

staktrace:
The backtrace points to [this line](https://github.com/servo/mozangle/blob/9539b4490316fe56223924cae6e4862b8260c16d/src/shaders/mod.rs#L137) in mozangle as the culprit. https://github.com/servo/mozangle/issues/43 appears to already been filed for this.

PR #4183, opened 2020-12-15

moz-gfx:


staktrace:
@gw3583 @lsalzman 

```
REFTEST TEST-UNEXPECTED-FAIL | reftests/text/shadow-transforms.yaml == reftests/text/shadow-transforms.png | image comparison, max difference: 2, number of differing pixels: 144 | 144 differences > 0 and <= 2 (allowed 135); 
```

Any idea which patch caused it? I'm guessing there's some subtle environment differences between this CI and Firefox CI.

staktrace:
Difference is 169 pixels on release job.

gw3583:
The recent change to use 2d textures instead of texture arrays causes this. It also causes an intermittent on m-c, but with less fuzziness - I've included fuzziness for this amount in https://phabricator.services.mozilla.com/D99825.

PR #4186, opened 2020-12-17
Start splitting the Renderer module

kvark:
This is a first piece of https://bugzilla.mozilla.org/show_bug.cgi?id=1675615

It makes a folder for the rendering code and moves out all the vertex state logic. Some parts turned out to be particularly great, like the exposed surface of the vertex modules is very small. The `RendererVAOs` is now nicer as well, since the `deinit()` logic is all visible, so hard to miss when adding/removing the types of VAO. Finally, there is `self.vaos[kind]` indexing as icing on the 🍰 .

@gw3583 I'd love to get your review and land this on the weekend, since otherwise this is risking to be a change that will be broken by every bit of code coming in WR. I think it's going to be easier to proceed afterwards: moving of `renderer.rs` is the main bit to conflict.


gw3583:
The patch looks good! Let's hold off on landing it until the render task graph changes land (https://bugzilla.mozilla.org/show_bug.cgi?id=1682365), but otherwise this should be fine to merge. Hopefully it should be a fairly minor rebase over the render task graph changes, most of that work is outside the renderer module. Create a phabricator patch and I'll approve it there?

kvark:
Ported over to https://phabricator.services.mozilla.com/D100148

kvark:
landed!

PR #4233, opened 2021-01-28
Use Cow<'static, [u8]> instead of Vec<u8> for ImageData

fschutt:
This is a small memory optimization in case the image is allocated / decoded at compile time (useful for embedded images). Since webrender never copies the image data but just reference it, this can avoid a couple of large allocations in the `&'static` case.

NOTE: This is a WIP PR to see if the tests run, I'm currently in the progress of doing the same thing for FontData, but I need to update dwrote-rs for that (since dwrote-rs uses `Arc<Vec<u8>>` instead of `Arc<Cow<'static, [u8]>>`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -547,6 +547,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "dwrote"
+version = "0.12.0"
+source = "git+https://github.com/fschutt/dwrote-rs?rev=7ca1b79032519ad4a9711b7ae04acb043aa949f0#7ca1b79032519ad4a9711b7ae04acb043aa949f0"
+dependencies = [
+ "lazy_static",
+ "libc",
+ "serde",
+ "serde_derive",
+ "winapi 0.3.8",
+ "wio",
+]
+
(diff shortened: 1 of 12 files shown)


kvark:
Interesting! Even if you embed images, wouldn't you get them compressed/encoded? I mean Png, Jpeg, and friends? In this case, you'd still decode them into a `Vec<>` before passing to WR.

If we are to have `Cow<'static, [u8]>`, let's define it as, say `type ByteData` or something.

fschutt:
Yes, but you can technically decode them at build time. GIMP / PS also has some tools to dump the raw bytes. It's more important for fonts.

kvark:
Ok. I think I would be fine with this change as long as `Cow` is behind a typedef to not expose this `static` everywhere.

mrobinson:
This repository is just a mirror, so substantive changes should be contributed upstream at https://github.com/mozilla/gecko-dev/tree/24f9846b7d8ea0819edee7aafc51a39753f1619a/gfx/wr.

PR #4234, opened 2021-01-28
Use Cow<'static, [u8]> for FontData instead of Vec<u8> 

fschutt:
Same idea as https://github.com/servo/webrender/pull/4233, but this PR is blocked until https://github.com/servo/dwrote-rs/pull/46 is merged.


mrobinson:
This repository is just a mirror currently, so please contribute all substantive changes upstream to Gecko.

PR #4317, opened 2021-03-25

moz-gfx:


staktrace:
`error: panic message is not a string literal`
`note: this is no longer accepted in Rust 2021`
`process didn't exit successfully: `rustc --crate-name wrench --edition=2018 ...`

If we're compiling with 2018 edition flags, why is it complaining about 2021 edition errors? Dunno! @kvark you've probably been following rust editions more than I have, any idea what's going on here?

staktrace:
> `error: panic message is not a string literal`
> `note: this is no longer accepted in Rust 2021`
> `process didn't exit successfully: rustc --crate-name wrench --edition=2018 ...`
> 

I'll take care of this in https://bugzilla.mozilla.org/show_bug.cgi?id=1701418

PR #4750, opened 2022-11-23
chore: fix animation example

keiya01:
The animation example was not working because of `duplicate key SpatialNodeUid`.


keiya01:
Ported to https://bugzilla.mozilla.org/show_bug.cgi?id=1803738

PR #4774, opened 2023-02-07
Fix broken links and other doc warnings

rMazeiks:


rMazeiks:
(I don't think this PR broke CI, as it only affects doc comments. If I'm reading the logs correctly, there was an issue with crates.io)

nical:
Hi, thanks for the PR and sorry for the delay. This repository is only a mirror, patches to webrender go in firefox's reprository (`gfx/wr/webrender/` folder in the repository).

There is some documentation about how to contribute here: https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html
Note that the process involves logging into bugzilla but this can be done without creating an account (using your github account instead).

PR #4824, opened 2023-10-19
Update mozangle

sagudev:
I know this should go via m-c but I just open PR to keep it noted until then.

### DIFF
--- a/wrench/Cargo.toml
+++ b/wrench/Cargo.toml
@@ -50,7 +50,7 @@ software = [ "swgl" ]
 
 [target.'cfg(target_os = "windows")'.dependencies]
 dwrote = "0.11"
-mozangle = { git = "https://github.com/sagudev/mozangle.git", branch = "bump", features = ["egl"] }
+mozangle = { git = "https://github.com/atbrakhi/mozangle.git", branch = "shared_object", features = ["egl"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 libc = "0.2"
--- a/wrench/Cargo.toml
+++ b/wrench/Cargo.toml
@@ -17,7 +17,7 @@ env_logger = { version = "0.10", optional = true, default_features = false }
 gleam = "0.15"
 glutin = "0.28"
 clap = { version = "3.1", features = ["yaml"] }
-glsl-lang = { version = "0.2", features = ["lexer-v2-full"] }
+glsl-lang = { version = "0.3", features = ["lexer-v2-full"] }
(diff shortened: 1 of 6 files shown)


wrench/Cargo.toml
```diff
@@ -17,7 +17,7 @@ env_logger = { version = "0.10", optional = true, default_features = false }
 gleam = "0.15"
 glutin = "0.28"
 clap = { version = "3.1", features = ["yaml"] }
-glsl-lang = { version = "0.2", features = ["lexer-v2-full"] }
```
sagudev, line 20: This is also needed due to bindgen and cc-rs version conflict.

wrench/Cargo.toml
```diff
@@ -50,7 +50,8 @@ software = [ "swgl" ]
 
 [target.'cfg(target_os = "windows")'.dependencies]
 dwrote = "0.11"
-mozangle = { version = "0.3.2", features = ["egl"] }
+#mozangle = { version = "0.4.0", features = ["egl"] }
```
sagudev, line 53: commented line should be removed

sagudev:
CI error is so weird. egl feature should only be enabled when target is windows. What is even weirder that dwrote (windows only dependency) is not included.

Cc @mrobinson

wrench/src/egl.rs
```diff
@@ -114,9 +114,10 @@ impl Context {
                 GlRequest::Specific(_, _) => return Err(CreationError::OpenGlVersionNotSupported),
                 GlRequest::GlThenGles { opengles_version, opengl_version } => {
                     if egl_version >= (1, 4) {
-                        if egl::BindAPI(ffi::egl::OPENGL_API) != 0 {
+                        /*if egl::BindAPI(ffi::egl::OPENGL_API) != 0 {
                             (Some(opengl_version), Api::OpenGl)
-                        } else if egl::BindAPI(ffi::egl::OPENGL_ES_API) != 0 {
+                        } else */
+                        if egl::BindAPI(ffi::egl::OPENGL_ES_API) != 0 {
```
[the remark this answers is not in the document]
sagudev: Ideally we should fallback to ES if no GL configs are available.

sagudev:
> CI error is so weird. egl feature should only be enabled when target is windows. What is even weirder that dwrote (windows only dependency) is not included.

`cargo test --all` causes problem within resolver, so either setting resolver to version 2 or running wrench unit test separately fixes the issue.

PR #4838, opened 2024-02-13
Update winit to 0.29, core-graphics to 0.23, and core-text to 20 for servo

Nopey:
These are the latest versions of each.

Although this webrender branch's test suite isn't healthy, this does build downstream in https://github.com/servo/servo/pull/31278

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,39 @@
 version = 3
 
 [[package]]
+name = "ab_glyph"
+version = "0.2.23"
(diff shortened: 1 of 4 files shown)


mrobinson:
The main branch in this this repository is a mirror. We cut release from the https://github.com/servo/webrender/tree/0.63 branch currently, but I think it doesn't make sense to update WebRender until Servo is ready to be updated.

mrobinson:
I'm going to close this one as it needs to, at least, be opened against our release branch.

mrobinson (approved):

PR #4848, opened 2024-03-13

moz-gfx:


sagudev:
This is obsolete per https://github.com/servo/webrender/pull/4849

PR #4856, opened 2024-08-28
Fix missing optimized shaders on OpenHarmony

jschwe:
We use GLES on OpenHarmony, so the shaders should be optimizied accordingly. 

Cargo guarantees that the `CARGO_CFG_xxx` variables are set. Using expect instead of matching improves the readability.

### DIFF
--- a/webrender/build.rs
+++ b/webrender/build.rs
@@ -128,9 +128,11 @@ fn write_optimized_shaders(
 
     // The full set of optimized shaders can be quite large, so only optimize
     // for the GL version we expect to be used on the target platform. If a different GL
-    // version is used we will simply fall back to the unoptimized shaders.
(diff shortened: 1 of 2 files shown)


mrobinson (approved):
Would you mind also submitting this to Gecko so that we will have this change the next time we rebase onto upstream WebRender?

jschwe:
I've added it to the tracking issue in the main servo repo so I don't forget. 
How often do we rebase onto newer webrender releases (approximately)?

mrobinson:
> I've added it to the tracking issue in the main servo repo so I don't forget. How often do we rebase onto newer webrender releases (approximately)?

It isn't done regularly now, but the hope is that in the future it could become more regular. So far I've done it twice.

PR #4857, opened 2024-09-05
fix: multiwindow example share events_loop

rockyzhengwu:
run example `cargo run --bin multiwindow ` on x11 will get panic with message :
`3414369 segmentation fault (core dumped)  cargo run --bin multiwindow` ,  this PR share the events_loop to fix this. 

maybe the best way is to upgrade winit to latest , winit `EventLoop`  can't be created twice.  https://github.com/rust-windowing/winit/blob/8db3e0e043198c02be997bb01cfd2ccac49e8053/src/event_loop.rs#L101

many line is auto formated by rustfmt ， what the rustfmt config to keep them not be changed ?

### DIFF
--- a/examples/multiwindow.rs
+++ b/examples/multiwindow.rs
@@ -41,17 +41,18 @@ impl RenderNotifier for Notifier {
         let _ = self.events_proxy.send_event(());
     }
 
-    fn new_frame_ready(&self,
-                       _: DocumentId,
-                       _scrolled: bool,
-                       composite_needed: bool,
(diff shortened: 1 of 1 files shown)


mrobinson:
Please submit this fix to the Gecko project. This repository is a downstream version of WebRender. We accept patches that help to improve WebRender on Servo, but only temporarily as we expect them all to go upstream to gecko-dev. 

https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html

PR #4858, opened 2024-09-24
Add RGBAF16 support

sagudev:
This will be needed to support rgba16float in gpucanvas.


mrobinson:
Is this still necessary?

sagudev:
Yes, yes it is, but it has low priority from me.

PR #4876, opened 2025-10-06
Update the sync location to reflect the new Firefox repository

mrobinson:


nicoburns (approved):
Looks good. It would be nice to copy the setup we have for Stylo: automated  daily syncs to the `upstream` branch, automatic reflection from `main` to named branches, etc.

PR #4877, opened 2025-12-22
Fetch `current_epoch` via `render_api`

shubhamg13:
This PR targets to access the `Scene`'s `PipelineInfo` using `render_api`. 

Aiming `render_api` as only interface used to communicate with `WebRender`. 
As of now `render_api` is already being used for most of communication except this one (and `update` ). 

In this PR

1. `pipeline_info` is moved to `SceneBuilderThread` along with `removed_pipelines`.
2. Connecting `PublishPipelineInfo` to `SceneBuilerThread`.
3. Along with making this `pipeline_info` synchronous as earlier we need to rely on calling of `renderer.update()` to `PublishPipelineInfo`. Now using event loop of `SceneBuilerThread` it is synchronous.

Add-ons
4. `FlushPipelineInfo` is also connected to `SceneBuilerThread` as required.

### DIFF
--- a/example-compositor/compositor/src/main.rs
+++ b/example-compositor/compositor/src/main.rs
@@ -498,7 +498,7 @@ fn main() {
             // as required.
             renderer.update();
             renderer.render(device_size, 0).unwrap();
-            let _ = renderer.flush_pipeline_info();
+            let _ = api.flush_pipeline_info();
 
             // Construct a simple display list that can be drawn and composited by DC.
             let mut txn = Transaction::new();
--- a/examples/common/boilerplate.rs
+++ b/examples/common/boilerplate.rs
@@ -312,7 +312,7 @@ pub fn main_wrapper<E: Example>(
 
         renderer.update();
         renderer.render(device_size, 0).unwrap();
-        let _ = renderer.flush_pipeline_info();
+        let _ = api.flush_pipeline_info();
         example.draw_custom(&*gl);
         windowed_context.swap_buffers().ok();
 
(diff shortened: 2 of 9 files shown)


shubhamg13:
@mrobinson @xiaochengh  Please take a look.

yezhizhen (commented):
Let's fix the CI first. Can you explain the motivation of "synchronous", and how is it achieved?

shubhamg13:
> Let's fix the CI first. Can you explain the motivation of "synchronous", and how is it achieved?

Please run the workflow again. I updated the PR description. PTAL.

xiaochengh:
Could you elaborate on the motivation of this PR?

Why do we prefer a sync api over an async one? Is there any issue with the current async API? Or do you just want to make it easier to call (which on its own might not justify the change though...)?

shubhamg13:
> Could you elaborate on the motivation of this PR?
> 
> Why do we prefer a sync api over an async one? Is there any issue with the current async API? Or do you just want to make it easier to call (which on its own might not justify the change though...)?

I updated the PR description. Please take a look.

mrobinson:
I'm also curious about the motivations here. I don't really great about making large architectural changes to WebRender. It's better to have fewer changes on top of the upstream version. The exception is for solving a problem that we are encountering. In that case, we should discuss with the folks at Mozilla about making our changes upstream or look at reworking the code in Servo.

mrobinson (commented):
Regarding this change: we try to keep our modifications to WebRender as minimal as possible and limited just to things that Servo needs. If you feel like this could be generally useful, a good step might be to submit it upstream to the Gecko repository and see if it's something that also makes sense for Firefox as well. If it lands upstream, it will also be available the next time we re-sync WebRender with upstream.

PR #4878, opened 2026-01-22
Prepare WebRender for crates.io release

nicoburns:
Changes made:

- Update `webrender` and in-repo crates that are dependencies of `webrender` to version 0.68
- Update these crates to use workspace depency versions similar to [stylo#249](https://github.com/servo/stylo/pull/249) (seeing as the feedback on that change was positive).

I have not bothered updating the versions of the other in-repo crates that are not dependencies of `webrender` (e.g. `wrench` and crates that are used by `wrench` like `swgl`) as we don't need to publish those and thus their version numbers don't matter, and I figured it would keep the diff a little smaller. But it would be easy to update those too if we wanted to.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
       - name: Install freetype
-        run: sudo apt install libfreetype-dev
+        run: sudo apt update && sudo apt install -y libfreetype-dev
       - name: Build wrench
         run: |
           cd wrench
--- a/webrender/Cargo.toml
+++ b/webrender/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "webrender"
-version = "0.68.0"
+version.workspace = true
 authors = ["Glenn Watson <gw@intuitionlibrary.com>"]
 license = "MPL-2.0"
 repository = "https://github.com/servo/webrender"
@@ -25,7 +25,7 @@ debugger = ["tiny_http", "serde_json", "url", "sha1", "base64", "api/debugger"]
 [build-dependencies]
 build-parallel = "0.1.2"
 glslopt = "0.1.11"
-webrender_build = { version = "0.0.2", path = "../webrender_build" }
+webrender_build = { workspace = true }
 
 [dependencies]
 bincode = "1.0"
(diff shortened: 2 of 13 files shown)


.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
       - name: Install freetype
-        run: sudo apt install libfreetype-dev
+        run: sudo apt update && sudo apt install libfreetype-dev
```
jschwe:
```suggestion
        run: sudo apt update && sudo apt install -y libfreetype-dev
```

webrender/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "webrender"
-version = "0.68.0"
+version.workspace = true
 authors = ["Glenn Watson <gw@intuitionlibrary.com>"]
 license = "MPL-2.0"
 repository = "https://github.com/servo/webrender"
```
jschwe, line 6: Should we perhaps update the description to mention this is servos fork of webrender? And perhaps remove authors, since it's not a required field (anymore) and not quite accurate?

jschwe:
And last comment, perhaps update the upstream URL in the readme to point to the new upstream at https://github.com/mozilla-firefox/firefox/

webrender/Cargo.toml
nicoburns: I think we probably don't need to mention that we're technically a fork. Even though Firefox/Gecko is upstream these days, they don't publish WebRender to crates.io, and if WebRender is like Stylo then their version can't easily be built outside of their tree (it depends on other C++ code in Gecko). So as far as crates.io consumers are concerned, Servo's version is the canonical version.

nicoburns:
> And last comment, perhaps update the upstream URL in the readme to point to the new upstream at https://github.com/mozilla-firefox/firefox/

I've done this. Although note that this PR is against the `0.68` branch (which is what we're using in Servo), not the `main` branch (which currently contains a very old version of WebRender, but is what Github shows as the README).

We should probably make the `main` branch the same as the 0.68 branch.

mrobinson (approved):

mrobinson:
I've amended this to not upgrade the versions of `wr_malloc_size_of` as the Mozilla project publishes the crate from upstream sometimes and the versions should not diverge.

mrobinson:
I have released WebRender 0.68 now, though I had to make some changes. I've merged them into this commit in the branch.

nicoburns:
@mrobinson Thanks!

Could we set the `main` branch to the contents of the `0.68` branch? I think it's very confusing for people when they hit the repo's main page and see a very old version. Looks like there may a couple of commits (https://github.com/servo/webrender/commit/c4bd5b47d8f5cd684334b445e67a1f945d106848 and https://github.com/servo/webrender/commit/e1c924ebad9ffdfe8c8c606aba77eb3f888c396a) that are only on `main` but not `0.68`, but they look like they should be easy enough to cherry-pick across.

mrobinson:
The issue is that because we are constantly resyncing with upstream, the main branch will have its history rewritten. Perhaps it is better to simply update the README in the main branch. In addition, I think we should look into merging our changes upstream and archiving this repository entirely.

nicoburns:
Updating the README (and perhaps deleting the code entirely) in the `main` branch would also work.

But this is effectively exactly the same issue as we have with Stylo, right? (except we sync less often and the diff is smaller). So I figured it could make sense to use the same solution here as we use there.

Getting rid of the repo entirely could make sense, except we probably don't want to have to coordinate with upstream to make releases? And it might be good to have a record of the code that was released?

(I also think we're also going to continue to need at least some Cargo.toml patches on top of upstream so as to strip out the Gecko-specific code (e.g. firefox-on-glean))

PR #4879, opened 2026-01-27
Clamp rect origin point and size after scaling

BudiArb:
Clamp rect origin point and size after scaling to prevent overflowing of f32 numbers


yezhizhen (commented):
Emm, I wonder if such change should be made in upstream instead?
cc @mrobinson

mrobinson:
Yes, let's try to get this upstream first, as it is not really related to Servo.

PR #4883, opened 2026-06-08
Move `FastTransform` to `webrender_api` crate

nicoburns:
`servo-malloc-size-of` and `servo-geometry` are currently depending on `webrender` because they need the `FastTransform` type. As these crates are low in the crate graph, this means that ~50 servo crates need to recompile when modifying webrender.

This PR, moves `FastTransform` into `webrender_api` which means that`servo-malloc-size-of` and `servo-geometry` can also depend only on `webrender_api` and the only servo crates that needs to depend on `webrender` are `servo-paint` and `servo`.

Servo PR: https://github.com/servo/servo/pull/45485


Loirooriol:
It's not a small patch, so I would like to know what @mrobinson thinks about this making future webrender syncs harder or such. Maybe it should land upstream, and then sync? Seems it has been a while.

nicoburns:
Ah, I had though the `FastTransform` type was already our downstream patch, but it looks like we just made it public. I can definitely send this upstream if that's preferable.

nicoburns:
I've sent this upstream as https://bugzilla.mozilla.org/show_bug.cgi?id=2046918. Maybe we can try and include it in the upcoming sync if it lands soon.

nicoburns:
I think the current state (without this PR) may actually be causing a relatively significant slowdown in clean build times for Servo, because it seems to be serializing the build of WebRender (including it's build script) and `script`/`script-bindings`, which both end up being part of the critical path in the build pipeline.

mrobinson:
We will likely re-sync and re-release soon due to https://github.com/servo/servo/issues/45728, so if you can land this upstream it should be included in the next release.

nicoburns:
I will try to land upstream soon. Will have to work out why it got rejected by CI last time (I'm currently having trouble with Mozilla try access)

PR #4886, opened 2026-06-13
Update freetype-sys to 0.23

jschwe:
- Split wrench into a different workspace, so we can avoid dependency unification. This is necessary since cargo doesn't allow two crate (versions) which link `freetype`.
- Because wrench depends on webrender and thus glyph_rasterizer, we need to disable the default feature static_freetype, so that wrench can still be built (with dynamic freetype).

This approachs avoids touching the unmaintained crate font-loader and the archived servo-fontconfig-sys

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,32 +114,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "85965b6739a430150bdd138e2374a98af0c3ee0d030b3bb7fc3bddff58d0102e"
 
 [[package]]
-name = "android_log-sys"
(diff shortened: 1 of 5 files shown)


mrobinson (commented):
Looks good, but please remove the new `Cargo.lock` file. I think it's fine to omit it from the repository.

mrobinson (approved):

mrobinson:
I've published the new version of `rust-freetype` now.

jschwe:
> Looks good, but please remove the new Cargo.lock file. I think it's fine to omit it from the repository.

Done. Should we go ahead and merge?

PR #4887, opened 2026-07-08
Bump the workspace version to 0.70.0

jschwe:
In preperation for the freetype-sys release.
This is technically a breaking change, since one can only link one freetype-sys version. We'd be fine in servoshell, but it could break third-party embedder builds, so better to be explicit and bump the version accordingly.

This PR should not be merged, and instead after approval pushed as the new 0.70 branch. 
We could also decide to publish as 0.69.1 and accept the potential downstream breackage, but I'd personally prefer to play it safe.
`webrender_api` wouldn't need a bump since it is unchanged, so we could revert that, but if we do a major bump i thought it would make things more consistent to stick with the same number.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1092,7 +1092,7 @@ dependencies = [
 
 [[package]]
 name = "glsl-to-cxx"
-version = "0.69.0"
+version = "0.70.0"
 dependencies = [
  "glsl",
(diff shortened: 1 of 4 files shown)


jschwe:
closing, since available now on 0.70 branch

